PR #10, opened 2018-08-20
Init component repository

### LINKED ISSUE #5: Release components on npm registry

### LINKED ISSUE #6: `NavigationItem` should hide the counter if it's 0

### LINKED ISSUE #7: Components should be either PascalCase or kebab-case
… according to https://vuejs.org/v2/guide/components-registration.html#Name-Casing.

### LINKED ISSUE #11: `NavigationItem` should allow simple click actions
In order to use this component for Mail, I need to add an entry to the app navigation for the expand/collapse folders feature. This entry neither points to an external URL, nor is it connected to the Vue router. It's a simple "button" with a click action.

Would that be possible, @skjnldsv?

skjnldsv:
:construction: WIP :construction: 

- [x] Popovermenu
- [x] Appnavigation
- [x] Datepicker
- [x] Webpack config
- [ ] Npm setup
- [ ] Tests
- [ ] Docs
- [ ] Github integration w/ drone
- [ ] ...?
- [x] Fix #7 
- [x] Fix #5 
- [x] Fix #11 
- [x] Fix #6


ChristophWurst:
Just so that we're on the same page: will this be released as a single npm package?

I kind of thought the components would be individual packages.

rullzer:
@skjnldsv see the question by @ChristophWurst

skjnldsv:
@ChristophWurst we wanted to at first but in the end this is really difficult to maintain :)
We decided to go for a single one, but you can import individual components like
`import { AppNavigation } from 'vue-components'`

ChristophWurst:
As discussed with @juliushaertl separate packages would mean more maintenance effort (separate npm/webpack configs) and we'd have dependencies of components inside the library anyway.

Let's give the single package approach a try and if it's problematic we can always split later on :+1:

PR #26, opened 2018-09-13
Change entry point, fix babel and update

skjnldsv:
Because we should be able to only import stuff using the js file.
Or am I wrong here?

Currently, if we do the usual `import { PopoverMenu } from 'nextcloud-vue'`, it will directly go onto the vue files. I think the goal when using webpack and stuff is to directly be able to only load a single js file and be done with it?

I'm confused a bit into all those settings.

This is how vue-material is doing for example 
https://github.com/vuematerial/vue-material/blob/383b25503e15f9d598ded3b3c0ca96b67ca856d6/package.json#L21-L22

I think the goal would be:
1. be able to import everything from a single file
    `VueMaterial from 'vue-material'`
2. be able to import individual components from individual js files (we need a new config)
    `import { MdButton, MdContent, MdTabs } from 'vue-material/dist/components'`
3. should we be able to import the source vue file directly?


skjnldsv:
Okay, I managed to end up with something interesting
``` js
import NextcloudVue from 'nextcloud-vue'
> {install: ƒ, AppNavigation: {…}, PopoverMenu: {…}, DatetimePicker: {…}}
```

BUT `import { AppNavigation } from 'nextcloud-vue'` is not working.
I'm guessing there is something strange or a missing setting.

Vue-Material is apparently not able to do that as well
`import { MdButton, MdContent, MdTabs } from 'vue-material/dist/components'` and not `'vue-material'`

ChristophWurst:
> Because we should be able to only import stuff using the js file.
> Or am I wrong here?

I don't see why we want to do that? Is that how Vue components usually work? I though one just pulls in Vue components as modules and uses the bundler of the app. But then you'd also need to make sure Babel and such are set up accordingly.

ChristophWurst:
https://cli.vuejs.org/guide/build-targets.html#library and https://medium.com/justfrontendthings/how-to-create-and-publish-your-own-vuejs-component-library-on-npm-using-vue-cli-28e60943eed3 could help

skjnldsv:
@ChristophWurst because with our current configuration the apps we were building were directly importing the vue file. :)
https://fr.vuejs.org/v2/cookbook/packaging-sfc-for-npm.html#Can%E2%80%99t-I-Just-Share-vue-Files-Directly

ChristophWurst:
> However, this excludes anyone who wishes to use the component directly in a browser via <script> tag, anyone who uses a runtime-only build, or build processes which don’t understand what to do with .vue files.

… is def what I don't want people to do with this lib. Imagine devs started loading their (outdated) versions of this lib in their apps with `script('…', '…');`? It would :boom: again.
IMO we really should get rid of that coding style and only allow this lib's code to be used as js modules.

skjnldsv:
@ChristophWurst that's a good question. You're right.
I have no idea then. Anyway without compilation, that's the only way to provide a module for another app, right?
So I guess it's the best way to still have a versatile option, right? :/

PR #55, opened 2018-09-27
Action component

### LINKED ISSUE #12: Action component
As discussed with @skjnldsv and @jancborchardt we should add a action component that will take a list of actions and render the action icon directly if there is only one. If multiple actions are added, a 3-dots menu with a popover menu should be shown.

### LINKED ISSUE #54: Linter fails
```
$ npm run lint

> nextcloud-vue@0.1.5 lint nextcloud-vue
> eslint --ext .js,.vue src tests


Oops! Something went wrong! :(

ESLint: 5.6.0.
No files matching the pattern "tests" were found.
Please check for typing mistakes in the pattern.
```

skjnldsv:
Fix #12 
Fix #54

### DIFF
--- a/src/components/Action/Action.vue
+++ b/src/components/Action/Action.vue
@@ -22,12 +22,13 @@
 
 <template>
 	<!-- if only one action, check if we need to bind to click or not -->
-	<a :tabindex="actions.length === 1" :href="actions.length === 1 && actions[0].href ? actions[0].href : '#'"
-		:class="[actions.length === 1 ? `${actions[0].icon} action-item--single` : 'action-item--multiple']"
-		v-on="actions.length === 1 && actions[0].action ? { click: actions[0].action } : {}">
+	<action :href="isSingleAction && actions[0].href ? actions[0].href : '#'"
+		:class="[isSingleAction ? `${actions[0].icon} action-item--single` : 'action-item--multiple']"
+		v-bind="mainActionElement()" class="action-item"
+		v-on="isSingleAction && actions[0].action ? { click: actions[0].action } : {}">
 
 		<!-- If more than one action, create a popovermenu -->
-		<template v-if="actions.length > 1">
+		<template v-if="!isSingleAction">
 			<span v-click-outside="closeMenu" tabindex="1" class="action-item__menutoggle icon-more"
 				@click="toggleMenu" />
 			<span :class="{ 'open': opened }" class="action-item__menu popovermenu">
@@ -35,7 +36,7 @@
 			</span>
 		</template>
 
-	</a>
+	</action>
 </template>
 
 <script>
@@ -77,6 +78,11 @@ export default {
 			opened: false
 		}
 	},
+	computed: {
+		isSingleAction() {
+			return this.actions.length === 1
+		}
+	},
 	mounted() {
 		// prevent click outside event with popupItem.
 		this.popupItem = this.$el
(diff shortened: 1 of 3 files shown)


src/components/Action/Action.vue
```diff
@@ -0,0 +1,93 @@
...
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+
+<template>
+	<!-- if only one action, check if we need to bind to click or not -->
+	<a :tabindex="actions.length === 1" :href="actions.length === 1 && actions[0].href ? actions[0].href : '#'"
```
ChristophWurst: May I suggest to move some logic into a computed property? It's hard to comprehend what's going on here otherwise.

src/components/Action/Action.vue
```diff
@@ -0,0 +1,93 @@
...
+		:class="[actions.length === 1 ? `${actions[0].icon} action-item--single` : 'action-item--multiple']"
+		v-on="actions.length === 1 && actions[0].action ? { click: actions[0].action } : {}">
+
+		<!-- If more than one action, create a popovermenu -->
+		<template v-if="actions.length > 1">
+			<span v-click-outside="closeMenu" tabindex="1" class="action-item__menutoggle icon-more"
+				@click="toggleMenu" />
+			<span :class="{ 'open': opened }" class="action-item__menu popovermenu">
```
ChristophWurst: Is `action-item__menu` a real class? What's that used for?

skjnldsv: Will be used for scoped design

src/components/Action/Action.vue
skjnldsv: I guess yes 🙈😂

src/components/Action/Action.vue
```diff
@@ -0,0 +1,127 @@
...
+	components: {
+		PopoverMenu
+	},
+	directives: {
+		ClickOutside
+	},
+	props: {
+		actions: {
+			type: Array,
+			required: true,
```
ChristophWurst: does `required: true` *and* a default make any sense? I guess the default will never be used.

skjnldsv: I usually use the default as an example, because let's face it, no one will ever use the default yes :stuck_out_tongue_closed_eyes:

ChristophWurst: I see. Let's move that into examples and/or documentation then :wink:

skjnldsv: Later then, on the jsDoc one?

ChristophWurst: Yes!

skjnldsv:
Bump for reviews then :rocket:

ChristophWurst:
Tested it as a replacement of the popovermenu in Mail's envelope component:

- [x] renders a single action nicely
- [ ] single action should have a tooltip
- [x] multiple actions render a popover menu
- [ ] popover menu positioning is a bit off
- [ ] popover menu entries should have a lower opacity (entries and icons are displayed black)

ChristophWurst:
> popover menu positioning is a bit off

## Current state
bildschirmfoto von 2018-11-02 09-18-42
## Locally fixed
bildschirmfoto von 2018-11-02 09-18-34

## Fix
bildschirmfoto von 2018-11-02 09-18-26

skjnldsv:
The popovermenu requires a fix on the server.
The right alignment is wrong :) (-5 instead of -2)
capture d ecran_2018-11-02_09-51-33

ChristophWurst:
> The popovermenu requires a fix on the server.

Fix it!

skjnldsv:
> > The popovermenu requires a fix on the server.
> 
> Fix it!

https://github.com/nextcloud/server/pull/12207

PR #64, opened 2018-10-12
Add avatar component with contacts menu

### LINKED ISSUE #14: Avatar
As listed in https://github.com/nextcloud/vue-components/issues/9, we need a standardized avatar component.

@juliushaertl @georgehrke @skjnldsv what should this look like?

IMO this should, for simple use cases, offer to display a colored single-letter avatar like the [image placeholder jQuery plugin in server](https://github.com/nextcloud/server/blob/2e25d22d1693e077ca324804ee208de2eb866040/core/js/placeholder.js), but as a Vue component. Additionally, the component should allow custom URLs to images (on external servers).

juliushaertl:
Fixes #14 

Example usage:

```
<!-- own avatar -->
<avatar user="admin" displayName="Julius" />

<!-- other users avatar -->
<avatar user="anexistinguser" displayName="An existing user" />

<!-- user to avatar image -->
<avatar :url="url" />

<!-- non-existing user -->
<avatar user="nonexistinguser" />
```

image

### DIFF
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -21,14 +21,14 @@
   -->
 
 <template>
-	<div v-tooltip="displayName" v-click-outside="closeMenu"
-		:class="{ 'icon-loading': loading, 'unknown': userDoesNotExist }"
+	<div v-tooltip="tooltip" v-click-outside="closeMenu"
+		:class="{ 'icon-loading': loadingState, 'unknown': userDoesNotExist }"
 		:style="avatarStyle"
 		class="avatardiv popovermenu-wrapper" @click="toggleMenu">
-		<img v-if="!loading && !userDoesNotExist" :src="avatarUrlLoaded">
+		<img v-if="!loadingState && !userDoesNotExist" :src="avatarUrlLoaded">
 		<div v-if="userDoesNotExist" class="unknown">{{ initials }}</div>
-		<div v-show="openedMenu" class="popovermenu">
-			<popover-menu :is-open="openedMenu" :menu="menu" />
+		<div v-show="contactsMenuOpenState" class="popovermenu">
+			<popover-menu :is-open="contactsMenuOpenState" :menu="menu" />
 		</div>
 	</div>
 </template>
@@ -55,7 +55,7 @@ export default {
 	props: {
 		/**
 		 * Set a custom url to the avatar image
-		 * either the url or user property must be defined
+		 * either the url, user or displayName property must be defined
 		 */
 		url: {
 			type: String,
@@ -63,7 +63,7 @@ export default {
 		},
 		/**
 		 * Set the user id to fetch the avatar
-		 * either the url or user property must be defined
+		 * either the url, user or displayName property must be defined
 		 */
 		user: {
 			type: String,
@@ -71,6 +71,9 @@ export default {
 		},
 		/**
 		 * Set a display name that will be rendered as a tooltip
+		 * either the url, user or displayName property must be defined
+		 * specify just the displayname to generate a placeholder avatar without
+		 * trying to fetch the avatar based on the user id
 		 */
 		displayName: {
 			type: String,
@@ -89,6 +92,13 @@ export default {
 		allowPlaceholder: {
 			type: Boolean,
 			default: true
+		},
+		/**
+		 * Disable the tooltip
+		 */
+		disableTooltip: {
+			type: Boolean,
+			default: false
 		}
 	},
 	data() {
@@ -101,14 +111,27 @@ export default {
 		}
 	},
 	computed: {
+		getUserIdentifier() {
+			if (this.isDisplayNameDefined) {
+				return this.displayName
+			}
+			if (this.isUserDefined) {
+				return this.user
+			}
+			return ''
+		},
 		isUserDefined() {
 			return typeof this.user !== 'undefined'
+		},
+		isDisplayNameDefined() {
+			return typeof this.displayName !== 'undefined'
 		},
 		isUrlDefined() {
 			return typeof this.url !== 'undefined'
 		},
-		isValid() {
-			return this.isUrlDefined || this.isUserDefined
+		shouldShowPlaceholder() {
+			return this.allowPlaceholder && (
+				this.userDoesNotExist)
 		},
 		avatarStyle() {
 			let style = {
@@ -118,19 +141,25 @@ export default {
 				fontSize: Math.round(this.size * 0.55) + 'px'
 			}
 
-			if (this.loadingState || !this.userDoesNotExist || !this.isUserDefined || !this.allowPlaceholder) {
+			if (!this.shouldShowPlaceholder) {
 				return style
 			}
 
-			const rgb = uidToColor(this.user)
+			const rgb = uidToColor(this.getUserIdentifier)
 			style.backgroundColor = 'rgb(' + rgb.r + ', ' + rgb.g + ', ' + rgb.b + ')'
 			return style
 		},
-		initials() {
-			if (!this.isUserDefined || !this.allowPlaceholder) {
-				return '?'
+		tooltip() {
+			if (this.disableTooltip) {
+				return false
 			}
-			return this.user.charAt(0).toUpperCase()
+			return this.displayName
+		},
+		initials() {
+			if (this.shouldShowPlaceholder) {
+				return this.getUserIdentifier.charAt(0).toUpperCase()
+			}
+			return '?'
 		},
 		menu() {
 			return this.contactsMenuActions.map((item) => {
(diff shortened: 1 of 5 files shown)


src/components/Avatar/Avatar.vue
```diff
@@ -0,0 +1,183 @@
...
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+
+<template>
+	<!-- TODO: Get rid of the inline css -->
+	<div v-tooltip="displayName" v-click-outside="closeMenu"
+		:class="{ 'icon-loading': loading, 'unknown': unknown }"
+		:style="{width: size + 'px', height: size + 'px'}"
```
juliushaertl: Ideally we should not have inline styles here, but we would loose the flexibility so specify any avatar size.

ChristophWurst: IMO totally valid in this case :+1:

src/components/PopoverMenu/PopoverMenuItem.vue
```diff
@@ -100,6 +101,15 @@ export default {
 			return this.item.key
 				? this.item.key
 				: Math.round(Math.random() * 16 * 1000000).toString(16)
+		},
+		iconIsUrl() {
+			try {
+				// eslint-disable-next-line no-unused-vars
```
ChristophWurst: if you don't use the object I guess you can just remove the `const url =` part and it should still do the same

src/components/Avatar/Avatar.vue
juliushaertl: Yes, just in case we might want to harden the server CSP to disallow inline script, but let's keep it for now.

src/components/Avatar/Avatar.vue
```diff
@@ -0,0 +1,176 @@
...
+			this.openedMenu = !this.openedMenu
+			if (this.openedMenu) {
+				this.fetchContactsMenu()
+			}
+		},
+		closeMenu() {
+			this.openedMenu = false
+		},
+		fetchContactsMenu() {
+			axios.post(OC.generateUrl('contactsmenu/findOne'), 'shareType=0&shareWith=' + this.user).then((response) => {
```
ChristophWurst: is `this.user` always save to use in the URL or should we wrap this with an [`encodeURIComponent`](https://www.w3schools.com/jsref/jsref_encodeuricomponent.asp)?

ChristophWurst:
While trying to test this in Mail I found a use case that is not yet supported: non-Nextcloud user but without an URL.

We don't always have a URL to an avatar. In those cases we still want to render an avatar with the sender's email address or name. This should more or less assemble what I've developed at https://github.com/nextcloud/mail/blob/b5ad1a0ddf3ca03d0b43e7c0fa835f4607279b34/js/components/Avatar.vue (simple placeholder with first letter).

ChristophWurst:
- [ ] Bug/limitation?: user of the component has to always specify the displayName in order to get a tooltip. If only the UID is available that won't be possible. Should we just use the `user` prop as fallback in that case?

juliushaertl:
> We don't always have a URL to an avatar. In those cases we still want to render an avatar with the sender's email address or name. This should more or less assemble what I've developed at nextcloud/mail:js/components/Avatar.vue@b5ad1a0 (simple placeholder with first letter).

Yep, let's include that as well. I'll look into that.

src/components/Avatar/Avatar.vue
```diff
@@ -0,0 +1,233 @@
...
+		url: {
+			type: String,
+			default: undefined
+		},
+		/**
+		 * Set the user id to fetch the avatar
+		 */
+		user: {
+			type: String,
+			default: undefined
```
skjnldsv: No required props?
I'm guessing there must be at least a minimal setup here? :)

src/components/PopoverMenu/PopoverMenuItem.vue
```diff
@@ -26,7 +26,8 @@
 		<a v-if="item.href" :href="(item.href) ? item.href : '#' "
 			:target="(item.target) ? item.target : '' "
 			rel="noreferrer noopener" @click="action">
-			<span :class="item.icon" />
+			<span v-if="!iconIsUrl" :class="item.icon" />
```
skjnldsv: I would like to not have this behavior as img are supposed to be deprecated inside popovermenus and navigation :)

src/components/Avatar/Avatar.vue
```diff
@@ -0,0 +1,233 @@
...
+		allowPlaceholder: {
+			type: Boolean,
+			default: true
+		}
+	},
+	data() {
+		return {
+			actions: [],
+			avatarUrlLoaded: null,
+			unknown: false,
```
skjnldsv: Could you comment a bit those vars?
I'm confused ;)

ChristophWurst: See https://github.com/nextcloud/vue-components/pull/64#issue-222429415

src/components/Avatar/Avatar.vue
juliushaertl: I couldn't find a way to define that either user or url must be defined. Any hints how that could be achieved are very welcome.

src/components/PopoverMenu/PopoverMenuItem.vue
juliushaertl: The reason is that the contacts menu api of the server returns urls for the action icons. We just don't have any icon class for those.

ChristophWurst:
Would it make sense to show a placeholder for unknown users with known `displayName` and `:allowPlaceholder="true"`? I see you already took the `$.imageplaceholder` logic from server (which I used at https://github.com/nextcloud/mail/blob/c1c875db07b59f500dd5ed695f772d8f979238f9/src/components/Avatar.vue#L18) to build the placeholder for known users. This should cover my use case from https://github.com/nextcloud/vue-components/pull/64#issuecomment-432617045.

juliushaertl:
> Would it make sense to show a placeholder for unknown users with known displayName and :allowPlaceholder="true"? I see you already took the $.imageplaceholder logic from server (which I used at nextcloud/mail:src/components/Avatar.vue@c1c875d#L18) to build the placeholder for known users. This should cover my use case from #64 (comment).

This should already work with :user and :allowPlaceholder="true". Is there a specific reason you might want to use displayname only instead of a user id?

ChristophWurst:
> Ah right, because using :user will always try to fetch the avatar first, which doesn't make sense for the mail app wink I'll add that as well.

Exactly. Thanks :v:

src/components/Avatar/Avatar.vue
skjnldsv: Yes, but ppl coming back to those vars won't understand anything :p 
Better have very understandable values or comments ;)

src/components/PopoverMenu/PopoverMenuItem.vue
skjnldsv: Then we should create the classes? :grin:

src/components/Avatar/Avatar.vue
skjnldsv: You can just use required: true. https://github.com/nextcloud/vue-components/blob/master/src/components/PopoverMenu/PopoverMenuItem.vue#L77
That way it will throw an error if not using the bare minimal? :woman_shrugging:

src/components/Avatar/Avatar.vue
ChristophWurst: One could also write a documentation for the lib :stuck_out_tongue:

juliushaertl:
@ChristophWurst 

> #64 (comment) I'd like to show a placeholder instead

`<avatar displayName="Example" />` will do that now, as discussed.

> The possibility to prevent the tooltip as it doesn't add much if the display name is already displayed next to the avatar

Combined with the above, you can also disable the tooltip with:

`<avatar displayName="FooBar" disableTooltip="true" />`

@skjnldsv I've tried to be a bit more descriptive with the variable names.

juliushaertl:
CI is also failing on master :/ https://travis-ci.com/nextcloud/vue-components/builds/89187769

ChristophWurst:
Thanks a lot :rocket: :confetti_ball: :taco: 

bildschirmfoto von 2018-10-25 18-46-43
:heavy_check_mark: works with display names only :rocket: 

bildschirmfoto von 2018-10-25 18-46-28
:heavy_check_mark: works with custom URLs :rocket:

examples/index.html
```diff
@@ -0,0 +1,18 @@
+<!doctype html>
```
skjnldsv: What to do with this file?

src/components/PopoverMenuButton/PopoverMenuButton.vue
```diff
@@ -0,0 +1,53 @@
+<template xmlns:>
```
skjnldsv: What is this component here btw? An error? :)

examples/index.html
juliushaertl: Ah sorry about that :see_no_evil: I accidentially added some files that were around in my folder. Fixed and rebased.

PR #69, opened 2018-10-26
Fix travis

skjnldsv:
Linting broken for a while.
I never noticed.... :man_facepalming:


skjnldsv:
Let's get this in so I can fix other prs

PR #70, opened 2018-10-26
Multiselect style and manual scoping

### LINKED ISSUE #30: Add multiselect
Add multiselect the same way we added the datepicker

skjnldsv:
Fix #30 
``` js
/**
 * Manually scope a vue component to be used with an external css file
 * The external css must have the root element set with [data-v-#{$scope_version}]
 * 
 * e.g. .multiselect[data-v-#{$scope_version}]
 * import ScopeComponent from 'Utils/ScopeComponent'
 * import Multiselect from 'vue-multiselect'
 * import './index.scss'
 * ScopeComponent(Multiselect)
 * 
 * @param {Object} Component the vue component
 */
```

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,7 +12,8 @@ module.exports = {
 		OC: true,
 		OCA: true,
 		Vue: true,
-		VueRouter: true
+		VueRouter: true,
+		SCOPE_VERSION: true
 	},
 	parserOptions: {
 		parser: 'babel-eslint',
--- a/src/components/AppNavigation/AppNavigationItem.vue
+++ b/src/components/AppNavigation/AppNavigationItem.vue
@@ -105,7 +105,7 @@
 </template>
 
 <script>
-import PopoverMenu from '../PopoverMenu/PopoverMenu'
+import PopoverMenu from 'Components/PopoverMenu/PopoverMenu'
 import ClickOutside from 'vue-click-outside'
 import Vue from 'vue'
 
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -38,7 +38,7 @@
 /* global OC oc_userconfig */
 
 import { VTooltip } from 'v-tooltip'
-import { PopoverMenu } from '../PopoverMenu'
+import { PopoverMenu } from 'Components/PopoverMenu'
 import ClickOutside from 'vue-click-outside'
 import axios from 'nextcloud-axios'
 import uidToColor from './uidToColor'
--- a/src/components/Multiselect/index.js
+++ b/src/components/Multiselect/index.js
@@ -22,11 +22,10 @@
 import Multiselect from 'vue-multiselect'
 import './index.scss'
 
+// Manually force the addition of a scope data for the css
 Multiselect.mounted = [
 	function() {
-		console.debug(this, this.$el)
-		// eslint-disable-next-line
-		this.$el.className += ' ' + SCOPE_VERSION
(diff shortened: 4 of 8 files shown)


skjnldsv:
@juliushaertl let me do a mixin for that before merging :)

dartcafe:
@skjnldsv 
I am confused. With 0.2.0 and some manual css, I got this result: 

before1
and
before2
After an update to 0.3.1, I get this: 
after1
after2

The code is here (0.2.0): https://github.com/nextcloud/polls/blob/develop-0.9/src/js/components/shareDiv.vue
and here (0.3.1): https://github.com/nextcloud/polls/blob/vue-components0.3.1/src/js/components/shareDiv.vue
My questions are: 

- how do I handle my own style to the multiselect again with the scoped design of the component? 
- And why do the seleted options appear as tags after the update? The html output also differs after the update to 0.3.1.

Additionally, I changed the hight of the drop down items, because, I use a custom template with avatars. I want them again to display properly.... 
after3

skjnldsv:
@dartcafe before, you needed to use the [`vue-multiselect`](https://github.com/nextcloud/server/blob/b7bd6bd68250612d975b63fcfbb0865224c93d7c/settings/src/components/userList.vue#L81) class to use the design nextcloud provided.
Now that the scope is implemented, we directly match nextcloud's style.

If you have multiple options selected AND use tagging, you need to use [a limit](https://github.com/nextcloud/server/blob/b7bd6bd68250612d975b63fcfbb0865224c93d7c/settings/src/components/userList/userRow.vue#L68) on the number of selected options. 
Then also use the scope to properly display the limit: 
```html
<span slot="limit" class="multiselect__limit" v-tooltip.auto="formatGroupsTitle(userSubAdminsGroups)">+{{userSubAdminsGroups.length-2}}</span>
```

Which we probably also should implement in nextcloud's version here. ^^

skjnldsv:
> how do I handle my own style to the multiselect again with the scoped design of the component?

You should not (theoretically) We need to work on the components here to have a working solution out of the box with every possibility. Nonetheless, we still need to support custom scope and such, so I don't really know yet :)

Maybe we should directly support different displays? and Instead of directly use the multiselect, create our own wrapper and slots with options (avatarList, tagging, etc etc)?
> And why do the seleted options appear as tags after the update? The html output also differs after the update to 0.3.1.

This is the display we have, but they're not tags, you cannot individually remove them. Though this is great looking when used with the limit (we should enforce that) :)

skjnldsv:
@dartcafe maybe an update from https://github.com/shentao/vue-multiselect ?

dartcafe:
0.2 and 0.3.1 are using vue-multiselect 2.1.0 as far as I can see. 
Comparing props, values and computed, the difference between both is, that in 0.3.1 the prop "showNoOptions" is missing (even, if I set it), but it seems not be involved to this. 

I am desperated, because I don't unterstand this. 

This is the HTML output of the version using vue-nextcloud 0.2.0
grafik
And this after the update to 0.3.1
grafik

skjnldsv:
@dartcafe is there a working poll version I can use?

dartcafe:
https://github.com/nextcloud/polls/tree/develop-0.9 is the working branch
https://github.com/nextcloud/polls/tree/vue-components0.3.1 is the branch with my problem
Both are working.

skjnldsv:
@dartcafe I'm working on a better Multiselect wrapping so we can add our own props.
Should be out today :)

skjnldsv:
capture d ecran_2018-11-08_16-36-37
Seems good to you @dartcafe ?

dartcafe:
Works as expected in 0.4.1, except the selection slot.

PR #75, opened 2018-10-28
Updating avatars after user change and support for group names

dartcafe:
I had some problems because the user of a user row was loaded asynchronous after the mounting of Avatar and so the avatar didn't get updated. Adding a watch to the username and moving the code from the mounted to a method did the job.

Additionally I got some problems, when creating an avatar for a group, which name was identically to a user's name. I added a prop to declare the user name ist not regarding a user. 

I hope, I found a right way to solve the problems.


dartcafe:
Sorry for hasseling with the base branch. Was not sure, which one has to be the right one.

skjnldsv:
> Sorry for hasseling with the base branch. Was not sure, which one has to be the right one.

Hum, it should be master since we merged the avatar component 🤔

dartcafe:
I am not happy with that rebasing. Will go the hard way...

PR #76, opened 2018-10-29
Adding avatar updates after user change and group support

dartcafe:
New try: 
I had some problems because the user of a user row was loaded asynchronous after the mounting of Avatar and so the avatar didn't get updated. Adding a watch to the username and moving the code from the mounted to a method did the job.

Additionally I got some problems, when creating an avatar for a group, which name was identically to a user's name. I added a prop to declare the user name ist not regarding a user.

I hope, I found a right way to solve the problems.

### DIFF
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -102,7 +102,7 @@ export default {
 		},
 		/**
 		 * Declares username is not a user's name, when true.
-		 * Prevents loading user's avatar from server and forces generating colored initials, 
+		 * Prevents loading user's avatar from server and forces generating colored initials,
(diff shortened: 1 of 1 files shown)


src/components/Avatar/Avatar.vue
```diff
@@ -99,6 +99,13 @@ export default {
 		disableTooltip: {
 			type: Boolean,
 			default: false
+		},
+		/**
+		 * Define username is not a user name
```
juliushaertl: I think we should add a note, that this will disable to load an avatar from the server. The current description is a bit confusing. :wink:

dartcafe:
done

Issue #91, opened 2018-11-05
Remove height attribute for .multiselect[...] .multiselect__content-wrapper li > span


dartcafe:
The height of the options in the dropdown is fixed to 20px and cuts out the template for the 'option' slot. Changing height to min-height?

https://github.com/nextcloud/vue-components/blob/6035e15ca2d87e74b3402cf4976ed650b5fa8eff/src/components/Multiselect/index.scss#L147

![grafik](https://user-images.githubusercontent.com/26707476/48024993-81687d80-e142-11e8-8cc8-092757cd64ec.png)

PR #92, opened 2018-11-05
Multiselect option height

dartcafe:
#91


skjnldsv:
Well, we need one at least for the other results?
Maybe just use min-height?

skjnldsv:
Fixed :)

Issue #81, opened 2018-10-31
Autocomplete component


ChristophWurst:
I'd like to see a component similar to jQuery UI's autocomplete. Mail currently uses that for recipient autocompletion.

@nextcloud/vue would that make sense for other apps as well or is this too specific?

ChristophWurst:
```
$ rg "autocomplete\(" .
core/js/sharedialogview.js
125:                    this.$el.find('.shareWithField').autocomplete("search");
435:                                            .autocomplete("option", "autoFocus", true);
565:                                            .autocomplete('search', $(e.target).val());
602:                                    .autocomplete('search', $(e.target).val());
632:                    $shareWithField.autocomplete('close');
633:                    $shareWithField.autocomplete('disable');
658:                                    $shareWithField.autocomplete('enable');
671:                                    $shareWithField.autocomplete('enable');
684:                                    $shareWithField.autocomplete('enable');
690:                                    $shareWithField.autocomplete('enable');
702:                            $shareWithField.autocomplete('enable');
761:                            $shareField.autocomplete({

```

georgehrke:
> @nextcloud/vue would that make sense for other apps as well or is this too specific?

In calendar there are a couple use-cases:
- [ ] autocomplete for sharing dialog
- [ ] autocomplete for an event's location (based on Contacts' addresses)
- [ ] autocomplete for entering an event's attendees

skjnldsv:
In Contacts we could suggest other contacts on the RELATED field :)

skjnldsv:
Basically you want a dataProvider for the multiselect?

ChristophWurst:
Ah right, multiselect. Yeah, I'll give that a go. The only downside is that we don't have colored avatars there (which the current Mail impl with jQuery UI's autocomplete has).

juliushaertl:
> Ah right, multiselect. Yeah, I'll give that a go. The only downside is that we don't have colored avatars there (which the current Mail impl with jQuery UI's autocomplete has).

We should be able to achieve that with https://vue-multiselect.js.org/#sub-custom-option-template

skjnldsv:
@ChristophWurst Yes, we should always use multiselect. The features are crazy available and it works really great! This is actually an issue on server, we need to use only one solution, this is a nightmare otherwise :)

Currently we have select2, multiselect, jquery autocomplete and now also vue-multiselect! :scream:

ChristophWurst:
Let's just use the multiselect component then :rocket:

[...]

PR #96, opened 2018-11-08
Wrap Multiselect

### LINKED ISSUE #91: Remove height attribute for .multiselect[...] .multiselect__content-wrapper li > span

skjnldsv:
After #70 
Fix #91
- [x] Create options standards for Multiselect
- [x] Slots
- [x] Style fixes

### Examples
Default style
``` html
<Multiselect :options="options" v-model="select" />
```
Default style multiple with auto width calculation (default option width is 150px)
You can set the min with `:tag-width="80"`
``` html
<Multiselect :options="options" v-model="select" :multiple="true" />`
```
Disable auto width calculation (always show max XXX entries)
``` html
<Multiselect :options="options" v-model="select" :multiple="true" :auto-limit="false" :limit="XXX" />`
```
Avatar select option
``` html
<Multiselect :options="options" v-model="select" label="displayName" track-by="user" :multiple="true" :user-select="true" />`
```

capture d ecran_2018-11-08_16-36-37

### DIFF
--- a/src/components/Multiselect/AvatarSelectOption.vue
+++ b/src/components/Multiselect/AvatarSelectOption.vue
@@ -1,0 +1,77 @@
+<template>
+	<span class="option">
+		<avatar :display-name="option.displayName" :user="option.user" :disable-tooltip="true"
+			class="option__avatar" />
+		<div class="option__desc">
+			<span class="option__desc--lineone">{{ option.displayName }}</span>
+			<span v-if="option.desc" class="option__desc--linetwo">{{ option.desc }}</span>
+		</div>
+		<span v-if="option.icon" class="icon option__icon" :class="option.icon"
+			title="Test0" />
+	</span>
+</template>
+
+<script>
+import Avatar from 'Components/Avatar'
+
+export default {
+	name: 'AvatarSelectOption',
+	components: {
+		Avatar
+	},
+	props: {
+		option: {
+			type: Object,
+			default() {
+				return {
+					desc: '',
+					displayName: 'Admin',
+					icon: 'icon-user',
+					user: 'admin'
+				}
+			},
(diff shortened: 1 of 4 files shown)


ChristophWurst:
> Default style multiple with auto width calculation (default option width is 150px)
> You can set the min with `:tag-width="80"`

What does that mean? Where do I need that?

> label="displayName" track-by="user"

Are these props on the data?

skjnldsv:
> What does that mean? Where do I need that?

It is the minimum width for the selected options. It will be calculated to create the limit popup.
If you set it to 80px, the min-width of the options will be to 80px. Since we use flex-shrink. Any more items that don't fit will be put in the `+xxx` popup :)
peek 09-11-2018 11-03

> Are these props on the data?

This is the normal props you can use on the multiselect lib to define which keys to use on the object options.
You will need options formatted like that:
```
{
	desc: 'This is the admin',
	displayName: 'Admin',
	icon: 'icon-user',
	user: 'admin'
}
```

ChristophWurst:
Idea for a future enhancement: a standard for the loading state (loading more data dynamically, https://github.com/shentao/vue-multiselect#asynchronous-dropdown)

skjnldsv:
> Idea for a future enhancement: a standard for the loading state (loading more data dynamically, https://github.com/shentao/vue-multiselect#asynchronous-dropdown)

Yes, That's what I was talking in your issue #81

PR #102, opened 2018-11-10
Highlighting selected option

dartcafe:


dartcafe:
grafik

skjnldsv:
How does it look on normal select (not the avatar design) ? :)

dartcafe:
grafik

dartcafe:
It seems, that the background needs !important

skjnldsv:
> It seems, that the background needs !important

What rule is overriding it on server?

dartcafe:
https://github.com/nextcloud/vue-components/blob/c793625e8780bc7c4527df8430beca288e7d730f/src/components/Multiselect/index.scss#L160

and corresponding setting in the server css.

BTW: shouldn't this changings be done in the server css as well?

skjnldsv:
> BTW: shouldn't this changings be done in the server css as well?

No, the idea is to scope this design and not use the one on server :)
We need to remove the multiselect-vue class and remove the old multiselect styling on server (or add a custom toggle.

dartcafe:
I just used the scoped style, so everything is fine, but the width. I want it 100% and override it with 
```
.multiselect {
    width: 100% !important;
    max-width: 100% !important;
}
```
div.multiselect:not(.multiselect-vue), select.multiselect:not(.multiselect-vue) is left over from server.css

skjnldsv:
> But `div.multiselect:not(.multiselect-vue), select.multiselect:not(.multiselect-vue)` is left over from server.css

Then I'll provide a pr tomorrow to make sure we don't have this :)
Let's get this in!

PR #104, opened 2018-11-12
Fix default items for app navigation

ChristophWurst:
Just see the template, the default has a (wrong? outdated?) structure.


skjnldsv:
We need the documentation ^^

PR #117, opened 2018-11-21
Use srcset for avatars

rullzer:
The browser can perfectly figure out itself what avatar to show.
As a bonus if you now start zooming in the browser will fetch a higer
resolution avatar if needed.

Signed-off-by: Roeland Jago Douma <roeland@famdouma.nl>


skjnldsv:
Should we then set the default size for the user avatar since we include bigger ones automatically?

rullzer:
The default size is already set to 32

PR #130, opened 2018-11-29
"active" class to popover-menu action entries

### LINKED ISSUE #128: allow to pass a "active" class to popover-menu action entries
We should have a way to set a `active` class to the popover-menu, so the selected item can be highlighted.
(as discussed with @juliushaertl)

violoncelloch:
this allows to highlight the selected entry based on the active class if item.active === true
fixes #128

Signed-off-by: Jonas Sulzer <jonas@violoncello.ch>

### DIFF
--- a/src/components/PopoverMenu/PopoverMenuItem.vue
+++ b/src/components/PopoverMenu/PopoverMenuItem.vue
@@ -68,7 +68,7 @@
 		</span>
 
 		<!-- If item.action is set instead, a button will be used -->
-		<button v-else-if="item.action" class="menuitem" :class="{'active':(item.active === true)}"
+		<button v-else-if="item.action" class="menuitem" :class="{'active':(item.active)}"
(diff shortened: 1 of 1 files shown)


src/components/PopoverMenu/PopoverMenuItem.vue
```diff
@@ -68,7 +68,8 @@
 		</span>
 
 		<!-- If item.action is set instead, a button will be used -->
-		<button v-else-if="item.action" class="menuitem" @click.stop.prevent="item.action">
+		<button v-else-if="item.action" class="menuitem" :class="{'active':(item.active === true)}"
```
skjnldsv: you could only check with `:class="{'active': item.active}"` :thinking:

violoncelloch: changed -- thank you!

PR #139, opened 2018-12-01
Multiselect fixes

skjnldsv:
- Border radius fixes
- Padding increase
- Removed triangle on single select

| Before | After |
|:---------:|:------:|
|capture d ecran_2018-12-03_18-16-36|capture d ecran_2018-12-03_18-17-35|
|capture d ecran_2018-12-03_18-19-44|capture d ecran_2018-12-03_18-19-16|

@jancborchardt I did not removed the user icon as this is actually a standard (you can use any icon you want like we do in the sharing with https://github.com/nextcloud/server/pull/11899)

### DIFF
--- a/src/components/Multiselect/index.scss
+++ b/src/components/Multiselect/index.scss
@@ -9,22 +9,39 @@
 	min-width: 160px; 
 	position: relative;
 	background-color: var(--color-main-background);
+
+	// active state, force the input to be shown, we don't want
+	// the placeholder or the currently selected options
 	&.multiselect--active {
 		/* Opened: force display the input */
 		input.multiselect__input {
 			opacity: 1 !important;
 			cursor: text !important;
+			// remove border radius on bottom opening
+			border-radius: var(--border-radius) var(--border-radius) 0 0;
 		}
 	}
+
+	// Remove radius on top opening
+	&.multiselect--active.multiselect--above {
+		input.multiselect__input {
+			border-radius: 0 0 var(--border-radius) var(--border-radius);
+		}
+	}
+
+	// disabled state background
 	&.multiselect--disabled,
 	&.multiselect--disabled .multiselect__single {
 		background-color: var(--color-background-dark) !important;
 	}
+
 	// loading state
 	&.icon-loading-small::after {
 		left: 100%;
 		margin-left: -24px;
 	}
+
+	// multiple selected options display
 	.multiselect__tags {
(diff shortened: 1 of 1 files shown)


ChristophWurst:
> Would you mind splitting these separate features for future PRs and provide before/after screenshots? That would be great!

Background: "Multiselect fixes" is just too generic. There will be many more of these fixes and knowing that has changed in a release of this lib will be hard to understand for anybody using this in their app.

skjnldsv:
> Background: "Multiselect fixes" is just too generic. There will be many more of these fixes and knowing that has changed in a release of this lib will be hard to understand for anybody using this in their app.

Commit comment is more detailed for this purpose ;)
Let me add some screenshots :)

ChristophWurst:
> Commit comment is more detailed for this purpose ;)

Yep, saw that, thanks! IMO still is a valid concern to have this separated as it also makes it possible to revert a single fix if it causes any issue.

skjnldsv:
@ChristophWurst updated op :)

jancborchardt:
@skjnldsv good stuff, thanks!

> did not removed the user icon as this is actually a standard

If I want to propose to only have an icon for when it's _not_ a single user (which is the default) should I just open an issue here?

(Cause right now with all entries having an icon, the icon becomes quite meaningless as it's almost always the same.)

skjnldsv:
I'd say this should be a 'common usage rule' then, because it make sense to have the icon for other things like an entry that represent a circle, a talk convo... (or god knows why :thinking:  ).
If I get you right, you mean that since the user is the default data item, we should not display the user icon but all the other ones are fine, correct? :wink: 

@ChristophWurst @juliushaertl I don't think we should prevent this specific icon (icon-user), but on the other hand, if we could be sure devs would follow this 'common sense usage rule', then we'd be in a dream world! Thoughts? :grin:

jancborchardt:
> If I get you right, you mean that since the user is the default data item, we should not display the user icon but all the other ones are fine, correct? :wink:

Exactly. :)

> I don't think we should prevent this specific icon (icon-user), but on the other hand, if we could be sure devs would follow this common sense usage rule, then we'd be in a dream world!

I also don’t think we should prevent it. But a good start would be:
- We don’t use it in our example apps or building blocks or in the docs
- In the docs we say something like "On the right you can show an icon signifying if it’s a group, a Talk conversation or something else special. If it is just a person, no icon is needed."

What do you think?

PR #140, opened 2018-12-01
Datepicker fixes

skjnldsv:
- [ ] Changed arrow style
- [ ] Improved days of the week readability
- [ ] Fixed padding and bigger clickable areas for accessibility
- [ ] Auto format input string with moment
- [ ] Embed the lang data from nextcloud
- [ ]


.babelrc.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
+	plugins: ['@babel/plugin-syntax-dynamic-import'],
```
ChristophWurst: not used

package.json
```diff
@@ -45,6 +47,7 @@
 	},
 	"devDependencies": {
 		"@babel/core": "^7.0.1",
+		"@babel/plugin-syntax-dynamic-import": "^7.0.0",
```
ChristophWurst: not used

src/components/DatetimePicker/DatetimePicker.vue
```diff
@@ -0,0 +1,78 @@
...
+
+	components: {
+		Datepicker
+	},
+
+	inheritAttrs: false,
+
+	computed: {
+		firstDay() {
+			return window.firstDay
```
ChristophWurst: who sets this global?

.babelrc.js
skjnldsv: Thanks, let me clean the leftovers :)

src/components/DatetimePicker/DatetimePicker.vue
skjnldsv: Nextcloud :)

ChristophWurst: How ugly :see_no_evil:

Can we at least externalize this to a helper/util module to hide this ugliness and make it more obvious where this is coming from?

skjnldsv: :cry: 
How would you do it?  :)

Issue #13, opened 2018-09-04
Moment


ChristophWurst:
As discussed at https://github.com/nextcloud/vue-components/issues/9#issuecomment-415686834, I'd like to see a component for the moment.js integration, a.k.a. *x seconds/minutes/hours ago thingy*.

The existing Vue components suggested didn't work as expected, esp. in regards to l10n. Moreover, they wouldn't work with our [live timestamp update magic](https://github.com/nextcloud/server/pull/1215) (cc @nickvergessen). Hence, I've created my own component. This is what it looks like:

```js
<template>
	<span class="live-relative-timestamp"
		  :data-timestamp="timestamp"
		  :title=title>{{ formatted }}</span>
</template>

<script>
	import moment from 'moment';

	if (typeof OC !== 'undefined') {
		moment.locale(OC.getLocale());
	}

	export default {
		name: "Moment",
		props: [
			'timestamp',
			'format'
		],
		computed: {
			title () {
				return moment.unix(this.timestamp / 1000).format(this.format || 'LLL');
			},
			formatted () {
				return moment.unix(this.timestamp / 1000).fromNow();
			}
		}
	}
</script>
```

It works well and can be easily integrated. The live updating also seems to work with this.

## Caveats

* Pulls in moment.js (obviously) but also lots of translations. Webpack combines some stuff, but it's still rather big. We basically have the same problem in server, but we'd now load the locals twice.
* Depends on a global var. I've added a check for it's existence, though.

ChristophWurst:
> How big is big? As we cache JS stuff pretty agressively.

~1/4 MB according to https://github.com/jmblog/how-to-optimize-momentjs-with-webpack#measurements.

skjnldsv:
1/4 like 0.25 or 1-4MB? ^^

ChristophWurst:
> 1/4 like 0.25

250kB :D

skjnldsv:
For a component that only display a relative date, this is crazy ^^"
If we only display by units like `x sec ago` `x days ago` `x hours ago`, shouldn't this be universal across languages?

skjnldsv:
Is there that many translations needed? I mean, it's like any other translation son our website?

nickvergessen:
the thing is, this will be included like 10 times in the end?
activity, notifications, announcements, talk, ...

> Is there that many translations needed? I mean, it's like any other translation son our website?

Well Days in long and short (Mon. Monday), same for months (dec. december), etc.

[...]

PR #146, opened 2018-12-04
Auto load locale from nextcloud for the Datepicker

### LINKED ISSUE #137: Datepicker fixes
- [x] Days of the week should have same colour as disabled days
- [x] More padding on the headers
- [x] Arrows with the same design as the original datepicker
- [ ] OK button with better wording
- [x] Wrap it
- [ ] Automatically use the provider locale from nc

skjnldsv:
Fix  #137

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
 		"@babel/polyfill": "^7.0.0",
 		"base64-inline-loader": "^1.1.1",
 		"md5": "^2.2.1",
+		"moment": "^2.22.2",
 		"nextcloud-axios": "^0.1.2",
 		"v-tooltip": "^2.0.0-rc.33",
 		"vue": "^2.5.16",
--- a/src/components/DatetimePicker/DatetimePicker.vue
+++ b/src/components/DatetimePicker/DatetimePicker.vue
@@ -24,7 +24,9 @@
 	<datepicker
 		v-bind="$attrs"
 		:minute-step="10"
+		:lang="lang"
 		:clearable="false"
+		:first-day-of-week="firstDay"
 		v-on="$listeners" />
 </template>
 
@@ -49,6 +51,28 @@ export default {
 		Datepicker
 	},
 
-	inheritAttrs: false
+	inheritAttrs: false,
+
+	computed: {
+		firstDay() {
+			return window.firstDay
+				? window.firstDay
+				: 0 // sunday as default
+		},
+		lang() {
+			// fallback to default in case of unavailable data
+			return {
+				days: window.dayNamesShort
+					? window.dayNamesShort			// provided by nextcloud
+					: ['Sun.', 'Mon.', 'Tue.', 'Wed.', 'Thu.', 'Fri.', 'Sat.'],
+				months: window.monthNamesShort
+					? window.monthNamesShort		// provided by nextcloud
+					: ['Jan.', 'Feb.', 'Mar.', 'Apr.', 'May.', 'Jun.', 'Jul.', 'Aug.', 'Sep.', 'Oct.', 'Nov.', 'Dec.'],
+				placeholder: {
+					date: 'Select Date' // TODO: Translate
+				}
+			}
+		}
+	}
 }
 </script>
--- a/src/utils/AsyncLoadLocale.js
+++ b/src/utils/AsyncLoadLocale.js
@@ -1,0 +1,41 @@
+
+/**
+ * @copyright Copyright (c) 2018 John Molakvoæ <skjnldsv@protonmail.com>
+ *
+ * @author John Molakvoæ <skjnldsv@protonmail.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+const AsyncLoadLocale = async function(locale) {
+	// default load e.g. 'fr-fr'
+	try {
+		await import('moment/locale/' + locale)
+		return locale // update local locale
+	} catch (e1) {
+		try {
+			console.error(e1)
+			await import('moment/locale/' + locale.split('-')[0]) // failure: fallback to 'fr'
+			return locale.split('-')[0] // update local locale
+		} catch (e2) {
+			console.error(e2)
+			return 'en' // last fallback to English
+		}
+	}
+}
+
+export default AsyncLoadLocale
--- a/package-lock.json
+++ b/package-lock.json
@@ -5047,14 +5047,12 @@
 				"balanced-match": {
 					"version": "1.0.0",
 					"bundled": true,
-					"dev": true,
-					"optional": true
+					"dev": true
 				},
 				"brace-expansion": {
 					"version": "1.1.11",
 					"bundled": true,
 					"dev": true,
-					"optional": true,
 					"requires": {
 						"balanced-match": "^1.0.0",
 						"concat-map": "0.0.1"
@@ -5074,8 +5072,7 @@
 				"concat-map": {
 					"version": "0.0.1",
 					"bundled": true,
-					"dev": true,
-					"optional": true
+					"dev": true
 				},
 				"console-control-strings": {
 					"version": "1.1.0",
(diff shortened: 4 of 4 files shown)


raimund-schluessler:
Will `moment` then be included twice if the app requires it as well? Or even be loaded three times by the server as well?

skjnldsv:
@raimund-schluessler I don't get your question :thinking:

raimund-schluessler:
The server ships `moment.js` in version 2.18.1 https://github.com/nextcloud/server/blob/master/package.json#L41
E.g. the contacts app requests `moment.js` in version 2.23.0 https://github.com/nextcloud/contacts/blob/master/package.json#L42
With this PR nextcloud-vue requires `moment.js` in version 2.22.2.

So my question is, will three different versions of `moment.js` be loaded when you open the contacts app? One from the server, and two with the `contacts.js` package (one directly, one in nextcloud-vue)?

Each `moment.js` version is at least 65 kB (gzip), more likely 225 kB (see https://bundlephobia.com/result?p=moment@2.23.0). This sounds like quite some overhead.

skjnldsv:
No, on contacts we use moment for the datepicker as well, so once this is merged, we will remove it from contacts :)
Though this is a good question, how should we do that. How can we make sure we don't load it multipel times. For calendar (for example) they'll probably have to also load it for the whole app. If we have the same version as nextcloud-vue, can webpack merge the two of those?

On another note, with dynamic loading, we can load the locale separately from moment core, so it will be moment (48.9KB) + locale (avg 1.5KB).
See #13

raimund-schluessler:
Maybe this might help https://github.com/webpack/docs/wiki/optimization#deduplication

skjnldsv:
> Maybe this might help https://github.com/webpack/docs/wiki/optimization#deduplication

> This option is redundant in webpack 2.

Seems perfect then!! Thanks @raimund-schluessler :hugs:

ChristophWurst:
My 2c: for independent, isolated code like the Vue components or an app bundle it should be possible add async loading of locales for moment. But in Server we just assume that they are ready on page load. I don't think we can fix that.

For that webpack deduplication to work I assume the versions have to be identical. Webpack must not deduplicate otherwise.

skjnldsv:
> But in Server we just assume that they are ready on page load. I don't think we can fix that.

What do you mean by that?
Couldn't we not load moment in our apps and use the server's version?

raimund-schluessler:
> Couldn't we not load moment in our apps and use the server's version?

We can. Tasks does not load its own moment version, just uses the server one.

ChristophWurst:
Sure you can use the server version, but this makes you dependent on server. And in the opposite direction updates of moment in server are more error prone as apps rely on it as well. This is the same issue with jQuery and the other dependencies.

Neither using the server version nor shipping a duplication with every app is ideal. But the latter at least offers isolation and easier maintenance.

skjnldsv:
So the best would be to have moment integrated with nextcloud-vue, then?

rullzer:
Yes. Deduplication from server is more important than the few kb.
Also note that this PR is a bit old. So in the future contacts and the vue-components will probably ship the same moment version so it will be merged.

src/utils/AsyncLoadLocale.js
```diff
@@ -0,0 +1,41 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+const AsyncLoadLocale = async function(locale) {
+	// default load e.g. 'fr-fr'
+	try {
```
juliushaertl: It would probably be nice to have something like this in the server main bundle as well.

skjnldsv: Async loading moment according to the user's locale?

dartcafe:
Any work on this? Date picker's localisation would be useful.

skjnldsv:
@dartcafe none so far. We should try to see what we can really do on this ! :thinking:

skjnldsv:
@ChristophWurst could we have locale and such in nextcloud-server for this one?

ChristophWurst:
> @ChristophWurst could we have locale and such in nextcloud-server for this one?

http://blog.wuc.me/nextcloud-server/modules/_l10n_.html#getlocale?

skjnldsv:
No, sorry, I meant window.monthNamesShort, window.dayNamesShort & window.firstDay

src/components/DatetimePicker/DatetimePicker.vue
```diff
@@ -49,6 +51,28 @@ export default {
...
 	},
 
-	inheritAttrs: false
+	inheritAttrs: false,
+
+	computed: {
+		firstDay() {
+			return window.firstDay
+				? window.firstDay
+				: 0 // sunday as default
```
daita: default should be 7

skjnldsv: as European, yes :)
Not for the rest of the world?

skjnldsv: @jancborchardt ^

daita: 0 is not working in fact, it returns an error; also, based on documentation:


Prop | Description | Type | Default
-- | -- | -- | --
first-day-of-week | set the first day of week | 1 - 7 | 7
-- | -- | -- | --

skjnldsv: oh damn

src/components/DatetimePicker/DatetimePicker.vue
```diff
@@ -49,6 +51,28 @@ export default {
...
+		},
+		lang() {
+			// fallback to default in case of unavailable data
+			return {
+				days: window.dayNamesShort
+					? window.dayNamesShort			// provided by nextcloud
```
korelstar: I suggest to document/validate this in/through the eslint plugin and not use the window variable:
https://github.com/nextcloud/eslint-plugin-nextcloud/blob/master/lib/environments/nextcloud.js
Then, it will be easier to track possible deprecation in the future.

src/components/DatetimePicker/DatetimePicker.vue
jancborchardt: > as European, yes :)
> Not for the rest of the world?

Monday as first day of the week is ISO standard. So while yeah, US, Canada, Australia and others don’t do that, we just need to pick something. :) Australia and Canada also adopted the metric system some time ago, so there’s hope. ;)

georgehrke:
What's the status here? :)

skjnldsv:
> What's the status here? :)

no idea, @ChristophWurst would have liked a better way of loading the locales from server, but there is still no packages like @nextcloud/lang or something!

I guess we should maybe get this in? What do you think @ChristophWurst ?

georgehrke:
> no idea, @ChristophWurst would have liked a better way of loading the locales from server, but there is still no packages like @nextcloud/lang or something!

Well, there is `@nextcloud/l10n`, which provides `getLocale()` and `getLanguage()`

src/utils/AsyncLoadLocale.js
```diff
@@ -0,0 +1,41 @@
...
+		await import('moment/locale/' + locale)
```
ChristophWurst: does this actually work when used in an app? Like, the `moment/locale` sits in side `node_modules`, which you don't have in production. Am I missing something?

skjnldsv:
> Well, there is `@nextcloud/l10n`, which provides `getLocale()` and `getLanguage()`

My bad, it was about window.firstDay, window.monthNamesShort and window.dayNamesShort :)

package.json
```diff
@@ -34,6 +34,7 @@
 		"@babel/polyfill": "^7.0.0",
 		"base64-inline-loader": "^1.1.1",
 		"md5": "^2.2.1",
+		"moment": "^2.22.2",
```
georgehrke: Should this be using `@nextcloud/moment` instead?

georgehrke:
> My bad, it was about window.firstDay, window.monthNamesShort and window.dayNamesShort 

I guess we could include a getFirstDay(), getMonthNamesShort(), etc. in `@nextcloud/l10n`?

@ChristophWurst

georgehrke:
https://github.com/nextcloud/nextcloud-l10n/pull/30

georgehrke:
Should we include the async loading of locales here or rather right away in `@Nextcloud/moment`?

skjnldsv:
@georgehrke even if you include it here, you will need your own implementatin of the dynamic moment loading on your webpack config :)

ChristophWurst:
> @georgehrke even if you include it here, you will need your own implementatin of the dynamic moment loading on your webpack config :)

Because of https://github.com/nextcloud/nextcloud-vue/pull/146#discussion_r333006574?

src/components/DatetimePicker/DatetimePicker.vue
skjnldsv: https://github.com/nextcloud/eslint-plugin/pull/111

raimund-schluessler:
Closing since https://github.com/nextcloud/nextcloud-vue/pull/2397 replaced it.

skjnldsv:
> Closing since #2397 replaced it.

Not really, this is about the display formatting, mine is about loading translations from the locale.

raimund-schluessler:
> > Closing since #2397 replaced it.
> 
> Not really, this is about the display formatting, mine is about loading translations from the locale.

@skjnldsv Sorry, I interpreted "supersedes #146" like that.

PR #150, opened 2018-12-06
Do not use default datepicker css

skjnldsv:
I dropped the use of the distant design because we had some conflict and I hate having to override design. Building our own is far better.

What I did:
1. copied the original style
2. cleaned it up
3. integrated our previous changes to it
4. increased the sizes of the buttons
5. used flex for better visual and reactivity
6. removed old buttons style so we can use nextcloud default

I took the liberty to properly implement the range picker. Even if nowhere else we use it YET, I think it looks very nice now.

This is the first step so we can then check the dark theme (we had some conflict before)

@nextcloud/designers @nextcloud/vue Any requests?

| Before | After |
|:---------:|:------:|
|peek 06-12-2018 19-01-old|peek 06-12-2018 19-01|
|peek 06-12-2018 19-03|peek 06-12-2018 19-05-new|
|peek 06-12-2018 19-07-old|peek 06-12-2018 19-06|


juliushaertl:
@skjnldsv I wanted to check out if cypress also offers some visual comparison, then we could use it for e2e testing and ui regression tests.

skjnldsv:
@juliushaertl it offers it, but during tests only, not sure if we can have a dedicated warning then :)
But there is a diff comparison between graphics iirc ;)

skjnldsv:
> The width of the hour and minute selection is a bit much, could be narrower? (Or even pick start and end time at the same time if possible depending on if app requests)

You mean the hover width? Like the blue area?

For the leading zero, there is nothing we can do on this one since this is coming from the library and we cannot override it nicely :)

PR #180, opened 2018-12-29
Set `opened` on item when toggling a nav item

weeman1337:
Changes on the open state are currently not reflected on the item.

### DIFF
--- a/src/components/AppNavigation/AppNavigationItem.vue
+++ b/src/components/AppNavigation/AppNavigationItem.vue
@@ -154,6 +154,7 @@ export default {
 		},
 		toggleCollapse() {
 			this.opened = !this.opened
+			this.item.opened = this.opened
 		},
 		cancelEdit(e) {
 			// remove the editing class


src/components/AppNavigation/AppNavigationItem.vue
```diff
@@ -154,6 +154,7 @@ export default {
 		},
 		toggleCollapse() {
 			this.opened = !this.opened
+			this.item.opened = this.opened
```
skjnldsv: Maybe we should not use a local opened state then and use a computed + getter/setter instead ?

juliushaertl: Yes, otherwise this will also fail if the item is stored in vuex and changed without using a mutation.

skjnldsv: @juliushaertl directly forwarding a data from a vux store is not the best idea anyway.
So you will also need to use a getter/setter on your app as well for that. :)
Though using one for here as well seems like a good idea to avoir a double state (this?opened AND this.item.opened)

ChristophWurst: With #195 apps will use this component directly, hence we can also just emit an event when the menu opens. Then apps can listen to that and update their state.

skjnldsv: Nice idea!
Let's use events and keep a local state then!

weeman1337:
Ah sorry. I would close this one as it wasn't the quick fix for the problem.
In the specific case it is the Deck navigation. There is now a custom navigation :/

When I find some time this needs more brain to make the nav items work in general..

PR #193, opened 2019-01-07
Datepicker dark mode

skjnldsv:
Require https://github.com/nextcloud/server/pull/13403

Created an automated font generation that gets applied to the scss.
That way we don't care about the dark theme, it gets automatically changed with the `color` attribute! :heart_eyes: 

capture d ecran_2019-01-07_13-41-45


ChristophWurst:
> #### Conflicting files
> `package-lock.json`
> `package.json`
> `src/assets/arrow-left-double.svg`
> `src/assets/arrow-left.svg`
> `src/assets/arrow-right-double.svg`
> `src/assets/arrow-right.svg`
> `src/components/DatetimePicker/DatetimePicker.vue`
> `src/components/DatetimePicker/index.scss`

:see_no_evil:

ChristophWurst:
> #### Conflicting files
> `package-lock.json`

well well well

hirschrobert:
could be related error https://github.com/nextcloud/contacts/commit/e1ec0b4521b640c7fc1fdfa4b136a63d315877b9#r31889831

raimund-schluessler:
> Require nextcloud/server#13403

This means, apps that use the datepicker component require >=NC15, although they would work perfectly fine on NC14 otherwise!?

ChristophWurst:
Reminds me that we should have a clear range of supported server versions in this library.

skjnldsv:
@raimund-schluessler yes! Actually that was always the plan. All of our apps migrations supports min 15 :)

raimund-schluessler:
> @raimund-schluessler yes! Actually that was always the plan. All of our apps migrations supports min 15 :)

Ok, fine with me.

PR #238, opened 2019-02-06
add props for navigation class and content class

### LINKED ISSUE #227: AppContent: "loading" in app-navigation
As far as I can see, there is no easy way to add the `loading` CSS class (loading animation) for the whole app-navigation when using the component `AppContent`.

Maybe, we should add a new parameter for passing CSS class names for app-navigation? Or is there another way to realize this with means of Vue?

Edit: same applies to template slot `content`.

korelstar:
As discussed in #227, I've added two new props for the component `AppContent` in order to set CSS classes for navigation and content.

Fixes #227.

### DIFF
--- a/src/components/AppContent/AppContent.vue
+++ b/src/components/AppContent/AppContent.vue
@@ -22,18 +22,18 @@
 
 <template>
 	<div id="content" :class="'app-' + appName">
-		<div v-if="!!$slots['navigation']"
+		<div v-if="$slots['navigation'] !== undefined"
 			id="app-navigation"
 			:class="navigationClass">
 			<slot name="navigation" />
 		</div>
-		<div v-if="!!$slots['content']"
+		<div v-if="$slots['content'] !== undefined"
 			id="app-content"
 			:class="contentClass">
 			<slot name="content" />
 		</div>
 		<slot />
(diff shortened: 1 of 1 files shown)


src/components/AppContent/AppContent.vue
```diff
@@ -23,11 +23,13 @@
 <template>
 	<div id="content" :class="'app-' + appName">
 		<div v-if="!!$slots['navigation']"
-			id="app-navigation">
+			id="app-navigation"
+			:class="navigationClass">
 			<slot name="navigation" />
 		</div>
 		<div v-if="!!$slots['content']"
```
korelstar: Due to that `v-if="!!$slots['content']`, the loading animation is only visible if the content is not empty. That's uncomfortable and I would like to remove this. Is there a need to keep this?

skjnldsv: @korelstar this is indeed an issue I mentioned with @ChristophWurst . :)

We should detect is slot is not declared, but not if empty indeed.
Because we still want to have app without app-navigation (for the mail account setup for example)

korelstar: I changed the check to `undefined`. Works for me now as intended. :-)

PR #260, opened 2019-02-20
Fix modal default width and previous/next event emit

skjnldsv:
- Do not emit if there is no previous or next action available
- Do not set a default width, let's leave that to the content width

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -102,10 +102,16 @@ export default {
 
 	methods: {
 		previous(data) {
-			this.$emit('previous', data)
+			// do not send the event if nothing is available
+			if (this.hasPrevious) {
+				this.$emit('previous', data)
+			}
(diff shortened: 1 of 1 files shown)


src/components/Modal/Modal.vue
```diff
@@ -201,7 +207,6 @@ export default {
 	height: 100%;
 	width: 100%;
 	.modal-container {
-		width: 70%;
 		max-width: 900px;
```
juliushaertl: Shouldn't we also remove the max-width/height then and leave this up to the implementing apps?

skjnldsv: No, this is actually nice to have a maximum standard. Just not a default size in case we want things smaller. But the maximum is a good thing to have :)

juliushaertl: Ok fine by me then :+1:

PR #262, opened 2019-02-21
Build separate entrypoints for each component

juliushaertl:
With this components can be imported individually and webpack can take care of just including the required code parts. 

```
import Avatar from 'nextcloud-vue/dist/Components/Avatar'
import Tooltip from 'nextcloud-vue/dist/Directives/Tooltip'
```

Unfortunately I'm not sure how compatible this is when an app doesn't use webpack, still needs some testing.

I had to remove the chunking since it didn't work properly when including it in another webpack bundle of an app, but with the current state, webpack seems to be able to deduplicate some parts of the code, even if both the full bundle and indivudal components are included at the same time.

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -47,17 +47,7 @@ module.exports = {
 		library: ['NextcloudVue', '[name]'],
 		umdNamedDefine: true
 	},
-	optimization: {
-		     splitChunks: {
-			   cacheGroups: {
-				vendor: {
-					test: /node_modules/,
-					chunks: "all",
-					filename: 'vendor[id].js',
-				}
-			  }
-		     }
-		},
+	
 	externals: {
 		vue: {
 			commonjs: 'vue',
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -1,4 +1,5 @@
 const path = require('path')
+const glob = require('glob')
 const { VueLoaderPlugin } = require('vue-loader')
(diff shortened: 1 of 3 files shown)


webpack.common.js
```diff
@@ -12,16 +12,52 @@ const md5 = require('md5')
 const PACKAGE = require('./package.json')
 const SCOPE_VERSION = JSON.stringify(md5(PACKAGE.version).substr(0, 7))
 
+function getComponents() {
+	const NcComponents = [
+		'Action', 'AppContent', 'AppNavigationItem', 'AppNavigationNew', 'AppNavigationSettings', 'Avatar', 'DatetimePicker', 'Modal', 'Multiselect', 'PopoverMenu'
```
juliushaertl: Should be dynamic

webpack.common.js
```diff
@@ -12,16 +12,52 @@ const md5 = require('md5')
...
+	];
+	let entries = {};
+	Object.values(NcComponents).forEach((component) => {
+		entries['Components/' + component] = path.join(__dirname, 'src', `components/${component}/index.js`)
+	})
+	return entries;
+}
+function getDirectives() {
+	const NcDirectives = [
```
juliushaertl: Should be dynamic

skjnldsv:
Looking very nice!

skjnldsv:
| Before | After |
|:---------:|:------:|
|||

PR #79, opened 2018-10-30
JSDoc init

skjnldsv:
- [ ] Do our components
- [ ] Fix default value display
- [ ] Host the docs somewhere

@nextcloud/vue

PR #272, opened 2019-02-23
Add first vuepress documentation

### LINKED ISSUE #17: Documentation
If possible : automated

juliushaertl:
I gave vuepress a try for building a documentation and this is the first WIP result http://vuecomponents.dev.bitgrid.net/code/components/Avatar/Avatar.html

The big benefit compared to #79 is that we can show actual examples of the component, since vuepress allows integrating components from markdown.

Note: Currently it is required to install vuepress locally, since there seems to be a conflict with the webpack build dependency versions of our repo.

close #17

### DIFF
--- a/documentation/.vuepress/components/VueCodeExample.vue
+++ b/documentation/.vuepress/components/VueCodeExample.vue
@@ -1,0 +1,42 @@
+<template>
+    <div class="code-example">
+        <strong>{{ title }}</strong>
+        <div class="code">
+            <slot></slot>
+        </div>
+        <strong>Rendered example</strong>
+        <div class="example-render">
+        <div>
+            <slot name="example"></slot>
+            <div ref="dynamicRendering"></div>
+        </div>
+        </div>
+    </div>
+</template>
+<style>
+.example-render {
+    border: 1px solid #aaa;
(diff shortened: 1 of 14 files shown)


skjnldsv:
wow, just wow!! @juliushaertl you're killing it!! :tada: :hugs:

skjnldsv:
@juliushaertl could we get an update on this?
Could be nice since we're getting bigger and bigger here :)

juliushaertl:
Ah yes, I'll have another look so we can get this done.

jancborchardt:
Good stuff, would indeed be great to have direct examples! :) One thing: The avatars are square – is this just because it’s a demo, or an issue in the component? Cause they should be circular. ;)

skjnldsv:
They're square: https://github.com/nextcloud/nextcloud-vue/blob/d3b54578ac7f78c5fff1885e978b758b95a5a8a9/src/components/Avatar/Avatar.vue#L324

Please submit a pr if needed (hiding as off topic both of our messages ;) )

skjnldsv:
@juliushaertl I rebased, but I still cannot get it to work :(

juliushaertl:
Next attempt in #439

PR #276, opened 2019-02-25
Add modal touch events

### LINKED ISSUE #263: Swipe support for modal component
The modal component we use for firstrunwizard and viewer should support swiping for touch interfaces like tablets and phones. :)

skjnldsv:
Fix #263

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -22,7 +22,7 @@
 
 <template>
 	<transition name="fade">
-		<div class="modal-mask">
+		<div ref="mask" class="modal-mask">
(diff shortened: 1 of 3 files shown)


src/components/Modal/Modal.vue
```diff
@@ -98,6 +100,15 @@ export default {
 	},
 	mounted() {
 		this.showModal = true
+
+		this.mc = new Hammer(this.$refs.mask)
```
ChristophWurst: :dancer:

skjnldsv: Yes ahah :dancing_women:

Issue #275, opened 2019-02-25
Allow to disable modal primary button styling


juliushaertl:
In the viewer the next button should just be a regular one instead of one with primary color.

skjnldsv:
@juliushaertl any particular reason? I kinda like the dual tone on this one :)
@jancborchardt shall we go for the same button design as default?

juliushaertl:
> @juliushaertl any particular reason? I kinda like the dual tone on this one :)

No real reason, it just looked a bit odd to me when testing ;)

skjnldsv:
Should we set the primary style on hover/focus? :)

skjnldsv:
:bellhop_bell: 
@ChristophWurst any opinions? :)

ChristophWurst:
Nope :man_shrugging:

jancborchardt:
The primary button styling should only be there in the first run wizard, so an optional styling.

For the picture viewer, "Next" is not necessarily the more important action, and also it takes way too much attention away from the actual content. In picture viewers, all the actions are usually monochrome.

skjnldsv:
Let's do it! :muscle:

skjnldsv:
Closed as this is disabled by default.

PR #277, opened 2019-02-26
Modal title and actions

skjnldsv:
- Add modal slideshow
- Add modal title
- Add modal actions
- Hide ui when the mouse is still

peek 26-02-2019 12-27

capture d ecran_2019-02-26_12-31-51


src/components/Action/Action.vue
```diff
@@ -82,6 +87,9 @@ export default {
 		},
 		firstAction() {
 			return this.actions[0]
+		},
+		moreIconClass() {
+			return `icon-more${this.forceWhite ? '-white' : ''}`
```
juliushaertl: How do we make sure this is also white on the dark theme then?

src/components/Modal/Modal.vue
```diff
@@ -154,13 +220,32 @@ export default {
...
+				this.clearView = true
+			}, 5000)
+		},
+		handleSlideshow() {
+			this.playing = true
+			if (this.hasNext) {
+				this.slideshowTimeout = setTimeout(() => {
+					this.next()
+					this.handleSlideshow()
+				}, 3000)
```
juliushaertl: It would be nice to have this dynamic :wink:

juliushaertl:
@skjnldsv Nice. However now we have two primary buttons, which seems a bit confusing as there shown't be two actions highlighted that way. I think next is definitely the main action here, while the slideshow start/stop is just an action.

src/components/Modal/Modal.vue
skjnldsv: you mean able to choose the time?

src/components/Action/Action.vue
skjnldsv: Good question! This is still an unsolved issue, right?

skjnldsv:
> @skjnldsv Nice. However now we have two primary buttons, which seems a bit confusing as there shown't be two actions highlighted that way. I think next is definitely the main action here, while the slideshow start/stop is just an action.

So, about #275, do you rather go with no primary at all? Or just the next? :)
I'm fine with both!

src/components/Modal/Modal.vue
juliushaertl: Yes.

src/components/Action/Action.vue
juliushaertl: For this one we could just ship the white more icon with the lib, since it will always be rendered on dark background.

skjnldsv: @juliushaertl shall we do the same with the close icon then?

skjnldsv:
@juliushaertl how about this?
capture d ecran_2019-02-26_14-56-58

src/components/Action/Action.vue
juliushaertl: Yes, at least until we have a proper other solution, I'd just include them if they are not needed to be inverted.

skjnldsv:
@juliushaertl done!
dev skjnldsv com_apps_vueexample_

src/components/Action/Action.vue
```diff
@@ -69,6 +69,10 @@ export default {
 					}
 				]
 			}
+		},
+		forceWhite: {
```
skjnldsv: I'm guessing that since we moved to the font icons, we can now set it on the parent component, so we might not need the props now :thinking:

jancborchardt:
Hehe, looking at backlog issue, my main feedback would have been what @juliushaertl already said about there being 2 primary buttons. :) Good stuff! :tada:

PR #278, opened 2019-02-27
Add a two-way prop 'open' to the action component

ChristophWurst:
Required for https://github.com/nextcloud/server/pull/14392.

This makes it possible to open/close the menu from the outside while also staying in sync with the internal state. This also still works without the property, just like it did before.

Usage:

```html
<Action v-if="!token.current"
    :actions="actions"
    v-bind:open.sync="actionOpen">
```

@skjnldsv as discussed :)

### DIFF
--- a/src/components/Action/Action.vue
+++ b/src/components/Action/Action.vue
@@ -69,11 +69,15 @@ export default {
 					}
 				]
 			}
+		},
+		open: {
+			type: Boolean,
+			default: false
 		}
 	},
 	data() {
 		return {
-			opened: false
+			opened: this.open
 		}
 	},
 	computed: {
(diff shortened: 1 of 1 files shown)


src/components/Action/Action.vue
```diff
@@ -90,10 +89,12 @@ export default {
 	},
 	methods: {
 		toggleMenu() {
-			this.opened = !this.opened
+			this.open = !this.open
```
skjnldsv: You cannot change a prop. This is against the vuejs guidelines :(
You need a local clone of the value

ChristophWurst: Yeah, I knew something felt wrong :)

ChristophWurst: Fixed.

PR #290, opened 2019-03-05
Make the app navigation (slot) a real component and add a spacer

ChristophWurst:
This makes it a lot easier for apps to create their own reusable navigation component.

Without this change, using an own component to make the navigation slot contents reusable resulted in an extra element wrapping the contents. This can easily break the design. I think the clean version is to just have a dedicated component as we have for many other things now.

### DIFF
--- a/src/components/AppNavigation/AppNavigation.vue
+++ b/src/components/AppNavigation/AppNavigation.vue
@@ -1,5 +1,5 @@
 <!--
- - @copyright Copyright (c) 2018 Christoph Wurst <christoph@winzerhof-wurst.at>
+ - @copyright Copyright (c) 2019 Christoph Wurst <christoph@winzerhof-wurst.at>
  -
  - @author Christoph Wurst <christoph@winzerhof-wurst.at>
  -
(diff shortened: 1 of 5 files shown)


ChristophWurst:
See https://github.com/nextcloud/mail/pull/1576 for how I used the code in an app.

src/components/AppNavigationSpacer/AppNavigationSpacer.vue
```diff
@@ -0,0 +1,36 @@
...
+</template>
+
+<script>
+export default {
+	name: 'AppNavigationSpacer'
+}
+</script>
+
+<style scoped>
+	li {
```
korelstar: Elements selectors should be avoided with `scoped`: https://vuejs.org/v2/style-guide/#Element-selectors-with-scoped-use-with-caution

ChristophWurst: Thanks! I didn't know that :)

src/components/AppNavigationSpacer/AppNavigationSpacer.vue
```diff
@@ -20,7 +20,7 @@
  -
  -->
 <template>
-	<li />
+	<li class="spacer" />
```
skjnldsv: Let's keep the same class structure? :see_no_evil: 
`app-navigation-spacer` ?

ChristophWurst: sure!

src/components/AppNavigation/AppNavigation.vue
```diff
@@ -0,0 +1,39 @@
...
+	<div id="app-navigation" :class="navigationClass">
+		<slot />
+	</div>
+</template>
+
+<script>
+export default {
+	name: 'AppNavigation',
+	props: {
+		navigationClass: {
```
korelstar: I think the property `navigationClass` is not needed anymore due to moving this element to it's own component. You can use `:class`, directly:

```
<app-navigation :class="{loading: loading}"> ... </app-navigation>
```

skjnldsv: Ah yes! You're right :see_no_evil: 
@ChristophWurst

ChristophWurst: Of course :see_no_evil: 

@korelstar could you submit a PR please? :)

korelstar: Yep, please see #296

PR #291, opened 2019-03-05
Fix navigation menu with two or more actions

ChristophWurst:
# Before 

bildschirmfoto von 2019-03-05 16-59-45

# After

bildschirmfoto von 2019-03-05 17-42-33

@skjnldsv as discussed


ChristophWurst:
This also fixes the bug that the menu couldn't be closed with a second click on the three dots icon.

PR #293, opened 2019-03-06
Prevent modal to go over the title

skjnldsv:
I'm not sure about this, it seems a good thing to do to avoid the content to go over the header content, but in the end i might look a bit shifted to the bottom. Any thoughts?

capture d ecran_2019-03-06_11-57-56

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -408,6 +408,9 @@ export default {
 	justify-content: center;
 	height: 100%;
 	width: 100%;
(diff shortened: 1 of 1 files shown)


ChristophWurst:
> I'm not sure about this, it seems a good thing to do to avoid the content to go over the header content, but in the end i might look a bit shifted to the bottom. Any thoughts?

Makes sense to me :+1:

skjnldsv:
@ChristophWurst then let's approve ;)

PR #296, opened 2019-03-08
remove obsolete prop 'navigationClass'

korelstar:
Component: `AppNavigation`

The property `navigationClass` is not needed anymore due to moving the app navigation element to it's own component. You can use `:class`, directly:

```
<app-navigation :class="{loading: loading}"> ... </app-navigation>
```

see https://github.com/nextcloud/nextcloud-vue/pull/290#discussion_r263565344


ChristophWurst:
Thanks!

PR #301, opened 2019-03-13
Add modal sizes

skjnldsv:
| big | medium | normal |
|:---------:|:---------:|:------:|
|dev skjnldsv com_apps_vueexample_|dev skjnldsv com_apps_vueexample_ (1)|dev skjnldsv com_apps_vueexample_ (2)|

I'm not convinced on the positions of the navigation. A bit of help @jancborchardt ? :wink:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3400,7 +3400,7 @@
     },
     "external-editor": {
       "version": "3.0.3",
-      "resolved": "https://registry.npmjs.org/external-editor/-/external-editor-3.0.3.tgz",
+      "resolved": "http://registry.npmjs.org/external-editor/-/external-editor-3.0.3.tgz",
       "integrity": "sha512-bn71H9+qWoOQKyZDo25mOMVpSmXROAsTJVVVYzrrtol3d4y+AsKjf4Iwl2Q+IuT0kFSQ1qo166UuIwqYq7mGnA==",
       "dev": true,
       "requires": {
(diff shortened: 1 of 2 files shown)


jancborchardt:
Nice! :)

- What’s this light 1px border around the image?
- I would instead go for: normal, large, and full.
  - normal is the smallest: Used for first run wizard
  - large is the middle one: Used for media files in the viewer
  - full takes the whole width or height, whatever comes first. It also has a full black background and _no_ border-radius. This is how the Gallery looks like right now, and it would be the full focus view when you switch to fullscreen.


------------------------

As said in the polishing issue about the controls:

> - Next & Previous buttons should be further to the sides, to be flush with X exit icon and not take focus away from the content
> - "Next" button shouldn’t be primary blue, takes too much attention from the images

skjnldsv:
> * What’s this light 1px border around the image?

quick mockup issue ;)



> * full takes the whole width or height, whatever comes first. It also has a full black background and _no_ border-radius. This is how the Gallery looks like right now, and it would be the full focus view when you switch to fullscreen.

what about the header?

jancborchardt:
> what about the header?

Just like gallery does it:
- Text has some dark text-shadow for visibility
- Control icons are white and have dark shadow too
- We could even have a short fade in the header from rgba(0,0,0,.3) to transparent, just like Netflix or Youtube does
- The controls fade out anyway

skjnldsv:
Please update review @jancborchardt :hugs: 
dev skjnldsv com_apps_files__dir=_ (1)

PR #307, opened 2019-03-16
AppNavigationItem: remove duplicate code

korelstar:
Supports both `action` and `href` in the same code snippet. However, this required a new JS-helper-method `callPreventStop` which stops the links to be processed if there is an action given.

This was done in the context of #246, which is now obsolete.


skjnldsv:
So, if there is an action, the link will be there but won't be fired on click, ritght?

korelstar:
Yes, like before, the action takes precedence (if both given).

skjnldsv:
I like it, it would also allow to open in a new tab if the link make sense.

PR #317, opened 2019-03-21
Reset timer on manual next

skjnldsv:
Fix https://github.com/nextcloud/viewer/issues/50

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -206,15 +206,28 @@ export default {
 	},
 
 	methods: {
+		// Events emitters
 		previous(data) {
 			// do not send the event if nothing is available
 			if (this.hasPrevious) {
+				// if data is set, then it's a user mouse event
+				// and not the slideshow handler, therefore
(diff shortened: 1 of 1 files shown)


src/components/Modal/Modal.vue
```diff
@@ -268,6 +275,35 @@ export default {
...
+			} else {
+				clearTimeout(this.slideshowTimeout)
+			}
+		},
+
+		/**
+		 * Reset the slideshow timer and keep going if it was on
+		 */
+		resetSlideshow() {
+			this.$el.querySelector('.progress-ring__circle').classList.remove('animated')
```
ChristophWurst: can't use use the `:class="{}` magic? or doesn't that work with svgs?

skjnldsv: Right, this is much nicer, I can use the next tick to trigger a `this.playing = !this.playing`
Vue is too clever, if you invert both on a function it won't change anything since it automatically detect that the change is worthless ^^

PR #320, opened 2019-03-24
AppNavigationNew: fix type of buttonClass

korelstar:
This allows for using also array or object syntax for CSS classes which is required, if the class has to be determined dynamically.


skjnldsv:
@korelstar don't forget to mention the nextcloud-vue group or at least me/ChristophWurst/juliushaertl when creating pr :wink:

korelstar:
@skjnldsv Okay, thanks for the hint. Didn't know about your working processes, so I thought that this is not necessary.

PR #321, opened 2019-03-24
AppNavigationItem: sync property for menu's open state

### LINKED ISSUE #309: Close an AppNavigationItem's PopoverMenu
If some of your app's `AppNavigationItem`s have at least two actions, they are shown in a `PopoverMenu`. Currently, it is not possible to close that menu automatically after triggering the action, since the respective switch is encapsulated in the  `AppNavigationItem` component. I'm not sure about which is the best solution to fix this. I see two possibilities:

1. Provide an interface to close the menu from outside the component. Unfortunately, there is no good way in Vue for calling a component's method from outside of that component.
2. Let `AppNavigationItem` close the `PopoverMenu` directly after calling the action. For this, a modified list of actions could be passed to `PopoverMenu`, where the actions are extended with the call to close the menu.

Any ideas on this?

korelstar:
Implements the idea of #278 for the component AppNavigationItem.

Example usage: https://github.com/nextcloud/notes/pull/290/commits/d4b7c439d01e5a66d11b7129d954369500f08de1

Closes #309.

### DIFF
--- a/src/components/AppNavigationItem/AppNavigationItem.vue
+++ b/src/components/AppNavigationItem/AppNavigationItem.vue
@@ -57,7 +57,7 @@
 				<!-- menu button if at least two actions -->
 				<li v-else-if="item.utils.actions && item.utils.actions.length > 1"
 					class="app-navigation-entry-utils-menu-button">
-					<button v-click-outside="hideMenu" @click="openedMenu = !openedMenu" />
+					<button v-click-outside="hideMenu" @click="toggleMenu" />
 				</li>
 			</ul>
 		</div>
@@ -109,11 +109,15 @@ export default {
 		item: {
 			type: Object,
 			required: true
+		},
+		menuOpen: {
+			type: Boolean,
+			default: false
 		}
 	},
 	data() {
 		return {
-			openedMenu: false,
+			openedMenu: this.menuOpen,
 			opened: !!this.item.opened
 		}
 	},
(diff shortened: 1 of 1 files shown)


src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -141,6 +148,11 @@ export default {
 	methods: {
 		hideMenu() {
 			this.openedMenu = false
+			this.$emit('update:menu-open', this.openedMenu)
```
skjnldsv: Does this work really?
I knew vue was auto-translating kebab-case to camelcase when using props, but not on events! :thinking:

korelstar: The [documentation](https://vuejs.org/v2/guide/components-custom-events.html#Event-Names) says:
> [...] For these reasons, we recommend you **always use kebab-case for event names**.

However, it looks that in this special case (using a prop with sync modifier), the update event can use both kebab-case or camelCase. And you can use both kebab-case or camelCase as HTML attribute when using the component. At least my test cases are working.

Therefore, I'm fine with changing the syntax. Please have a look at the added commit. This works fine for me in the notes app.

ChristophWurst: if the Vue docs recommend kebab-case then we should use it

skjnldsv: Agreed then! SOrry abou tthat :)
Thanks for the tip @korelstar !

PR #325, opened 2019-03-27
Force all modals to mount in body

skjnldsv:
That way, if any app uses it into a child element that is within #app-content, it will still be displayed above all the UI like any other modal! :grin: 
:tada:

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -196,6 +196,7 @@ export default {
 	},
 	mounted() {
 		this.showModal = true
+
 		// init clear view
 		this.handleMouseMove()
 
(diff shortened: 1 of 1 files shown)


src/components/Modal/Modal.vue
```diff
@@ -196,13 +196,17 @@ export default {
...
 		// init clear view
 		this.handleMouseMove()
 
 		this.mc = new Hammer(this.$refs.mask)
 		this.mc.on('swipeleft swiperight', e => {
 			this.handleSwipe(e)
 		})
+
+		// force mount the component to body
+		document.body.insertBefore(this.$el, document.body.lastChild)
```
ShGKme, line 209: @skjnldsv Do you remember why it was added **before** the last child, not as the last child? 😁

skjnldsv: I have no clue :see_no_evil:

PR #326, opened 2019-03-27
Allow custom atrributes to popovermenu entries

skjnldsv:
Allow us to provide custom attributes that will be applied to the element directly:
```js
{
	icon: 'icon-download',
	text: t('contacts', 'Download'),
	href: this.contact.url,
	attr: {
		download: this.contact.displayName + '.vcf'
	}
}
```
```html
<a href="/.../xxx.vcf" rel="noreferrer noopener" download="John doe.vcf">
	<span class="icon-download"></span>
	<span>Download</span>
</a>
```

@nextcloud/vue is this the way to go? Would you prefer individual properties like `item.download` instead?

### DIFF
--- a/src/components/PopoverMenu/PopoverMenuItem.vue
+++ b/src/components/PopoverMenu/PopoverMenuItem.vue
@@ -25,7 +25,8 @@
 		<!-- If item.href is set, a link will be directly used -->
 		<a v-if="item.href" :href="(item.href) ? item.href : '#' "
 			:target="(item.target) ? item.target : '' "
-			rel="noreferrer noopener" @click="action">
+			rel="noreferrer noopener" v-bind="item.attr"
+			@click="action">
 			<span v-if="!iconIsUrl" :class="item.icon" />
 			<img v-else :src="item.icon">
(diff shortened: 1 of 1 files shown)


ChristophWurst:
> @nextcloud/vue is this the way to go? Would you prefer individual properties like `item.download` instead?

🤷

skjnldsv:
@ChristophWurst because then we could directly remove the properties like`item.href` and put them into attr instead ! :thinking:

ChristophWurst:
We can do that. But tbh I'm not a big fan of the Object-type props we have for many components in this lib. This makes it just super hard to discover the available properties and the validation does also not work out of the box.

Therefore I would vote for an explicit `download` prop.

korelstar:
@skjnldsv  Do you have any other attributes in mind that could benefit from your approach?

skjnldsv:
> Therefore I would vote for an explicit `download` prop.

I agree, it feels weird then.
Let's go for download on `a`

skjnldsv:
> @skjnldsv Do you have any other attributes in mind that could benefit from your approach?

Not that I'm aware of :thinking:

PR #330, opened 2019-04-03
Fix bundling of mixins

ChristophWurst:


src/mixins/index.js
```diff
@@ -20,7 +20,7 @@
  *
  */
 
-import isMobile from './isMobile'
+import isMobile from './isMobile/index'
```
skjnldsv: not needed. index is the default :)
https://github.com/nextcloud/nextcloud-vue/blob/4e854e6cffe35ebe3df653a8a06f0e45274e40ce/src/directives/index.js#L23

ChristophWurst: fixed

ChristophWurst:
@skjnldsv MÖRGE

PR #333, opened 2019-04-03
Fix the isMobile mixin

ChristophWurst:
* Use a real method for data to have access to `this`
* Rename private methods with a _ prefix


src/mixins/isMobile/index.js
```diff
@@ -21,21 +21,23 @@
  */
 
 export default {
-	data: () => ({
```
skjnldsv: :man_facepalming:

PR #334, opened 2019-04-06
AppNavigationSettings: remove obsolete code; fix jiggle

korelstar:
I found a strange behavior when using the component `AppNavigationSettings`: When I click the first time on "Settings", the settings-details opens, but immediately closes again.  Clicking the next time, it works as expected.

Then, I tried to debug the code of this component and found a strange use of the boolean variable `open`: `:class="{open}"`. For me, this doesn't make any sense. So I removed it, and now the bug has disappeared.

But moreover, I found that the local variable `open` is not used anymore in this component, so it can be removed completely (handling open/close is done in nextcloud/server by `registerAppsSlideToggle`).

This looks like a big change, so please review this carefully. Maybe, I've overseen something.

### DIFF
--- a/src/components/AppNavigationSettings/AppNavigationSettings.vue
+++ b/src/components/AppNavigationSettings/AppNavigationSettings.vue
@@ -21,13 +21,10 @@
  -->
 
 <template>
-	<div id="app-settings"
-		v-click-outside="closeMenu"
-		:class="{open}">
+	<div id="app-settings">
 		<div id="app-settings-header">
 			<button class="settings-button"
-				data-apps-slide-toggle="#app-settings-content"
-				@click="toggleMenu">
+				data-apps-slide-toggle="#app-settings-content">
 				{{ title }}
 			</button>
 		</div>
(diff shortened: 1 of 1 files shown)


skjnldsv:
> :class="{open}"

This is a proper use of a destructuring object.
It will just add the 'open' class if open is true :)

This is actually not supposed to be removed. We do manage the settings state from the inside as we wanted to be independent from the weird old functions that manages it on server. If you remove this, contacts doesn work anymore for example :woman_shrugging: 

I think we might have to remove the `data-apps-slide-toggle` though, this is the one that is unnecessary :thinking: 
It might be conflicting with the legacy code we have in server. @ChristophWurst I think this might be similar to the navigation toggle, if the settings are not loaded yet when the legacy script is called, the settigs open/close handler gets registered twice (from here and from server) :)

korelstar:
Thanks for the clarification. Now I understand, why I'm seeing this jiggle on the first click: both methods are activated and they conflict (btw. you can reproduce this in your `vueexample`).

If you want to be independent from server, I will close this PR. But then, we should really remove `data-apps-slide-toggle`. This will fix the issue, too (I've just tested it).

However, the old toggle from server has a nice animation and the new one from vue has no animation (yet)!

skjnldsv:
> btw. you can reproduce this in your `vueexample`

Nice catch! :see_no_evil: I did not even noticed! :grin: 



> If you want to be independent from server, I will close this PR. But then, we should really remove `data-apps-slide-toggle`. This will fix the issue, too (I've just tested it).
> 
> However, the old toggle from server has a nice animation and the new one from vue has no animation (yet)!

Yes, let's do that and add the vue slideUp transition here :)

skjnldsv:
@korelstar any news? :)

korelstar:
I just started to work on the transition. But it looks that we can't use [vue transitions](https://vuejs.org/v2/guide/transitions.html) here, since visibility isn't done by vue means (`v-if` or `v-show`), but from server CSS (`display: none`). Hence, we have to add `v-show`, too. I'm experimenting a little bit with this and I will provide a new PR, soon.

korelstar:
see #343. But the animations still doesn't work :-(

PR #339, opened 2019-04-08
Add a navigation toggle and vue-slideout

ChristophWurst:
There is one glitch when the menu nav is hidden again for the first time by clicking the icon: it shows blank space instead of moving the content to the right. @skjnldsv could you have a look and work your magic? :)

### DIFF
--- a/src/components/AppContent/AppContent.vue
+++ b/src/components/AppContent/AppContent.vue
@@ -21,11 +21,72 @@
  -->
 
 <template>
-	<div id="app-content">
+	<div id="app-content" class="no-snapper" :style="opened ? 'transform: translateX(300px)' : ''">
+		<app-navigation-toggle @click="toggleNavigation" />
 		<slot />
 	</div>
 </template>
 
 <script>
-export default {}
+import Hammer from 'hammerjs'
+import AppNavigationToggle from '../AppNavigationToggle'
+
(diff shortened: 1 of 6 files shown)


ChristophWurst:
@skjnldsv what's the status here?

skjnldsv:
@ChristophWurst needs to find a fix for the server compatibility.
I would be in favour of NOT initiating it if the toggle is present OR if app-content is not here.

ChristophWurst:
You mean with https://github.com/nextcloud/server/pull/14972?

skjnldsv:
Yep :)

src/components/AppContent/AppContent.vue
```diff
@@ -21,11 +21,72 @@
...
+	},
+	mounted() {
+		this.mc = new Hammer(this.$el)
+		this.mc.on('swipeleft swiperight', e => {
+			this.handleSwipe(e)
+		})
+	},
+	unmounted() {
+		this.mc.off('swipeleft swiperight')
+		this.ms.destroy()
```
juliushaertl:
```suggestion
		this.mc.destroy()
```

skjnldsv: Thanks!! Nice catch!

PR #340, opened 2019-04-08
Implement sidebar component

skjnldsv:
@nextcloud/vue 

Closes https://github.com/nextcloud/server/issues/10289

Standardisation is minimal but looks very nice and allow quite a few specificity that looks really great!

| 1 | 2 |
|:------:|:-----:|
|dev skjnldsv com_apps_vueexample_|dev skjnldsv com_apps_vueexample_ (2)|
dev skjnldsv com_apps_vueexample_ (3)| |

``` vue
<app-sidebar v-show="show"
	title="andy-yeo-1387151-unsplash.jpg"
	subtitle="4,3 MB, last edited 19 days ago"
	background="/core/preview?fileId=54&x=1880&y=993&a=true"
	:actions="menu"
	:tabs="tabs"
	@close="closeSidebar">
	<template #primary-action>
		<button class="primary">Leave call</button>
	</template>
	<template #secondary-action>
		<input name="link-checkbox" id="link-checkbox" class="checkbox link-checkbox" value="1" type="checkbox">
		<label for="link-checkbox" class="link-checkbox-label">Share link</label>
	</template>
	<tab name="Chat" icon="icon-talk">
		this is the chat tab
	</tab>
	<tab name="Activity" icon="icon-activity">
		this is the activity tab
	</tab>
	<tab name="Comments" icon="icon-comment">
		this is the comments tab
	</tab>
</app-sidebar>
```

### DIFF
--- a/src/components/AppSidebar/AppSidebar.vue
+++ b/src/components/AppSidebar/AppSidebar.vue
@@ -26,7 +26,7 @@
 <template>
 	<transition name="slide-right">
 		<aside id="app-sidebar">
-			<header class="app-sidebar-header">
+			<header :class="{ 'app-sidebar-header--with-figure': hasFigure }" class="app-sidebar-header">
 				<!-- close sidebar button -->
 				<a href="#" class="icon-close" :title="t('core', 'close')"
 					@click="closeSidebar" />
@@ -325,6 +325,17 @@ $desc-vertical-padding: 18px;
 			margin: 10px;
 			max-height: 50px;
 			align-items: center;
+		}
+
+		// menu without figure needs to be fixed or
+		// it'll go over the close icon
+		&:not(.app-sidebar-header--with-figure) {
+			.app-sidebar-header__menu {
+				background-color: transparent;
+				top: 0;
+				right: 44px;
+				margin-top: 0;
+			}
 		}
 	}
 	.app-sidebar-tabs {
--- a/src/components/AppSidebar/AppSidebar.vue
+++ b/src/components/AppSidebar/AppSidebar.vue
@@ -57,7 +57,8 @@
 				</div>
 			</header>
 			<!-- tabs navigation -->
-			<nav class="app-sidebar-tabs__nav"
+			<nav v-if="hasMultipleTabs"
+				class="app-sidebar-tabs__nav"
 				@keydown.left.exact.prevent="focusPreviousTab"
 				@keydown.right.exact.prevent="focusNextTab"
 				@keydown.tab.exact.prevent="focusActiveTabContent"
@@ -80,7 +81,8 @@
 				</ul>
 			</nav>
 			<!-- tabs content -->
-			<div class="app-sidebar-tabs__content">
+			<div :class="{'app-sidebar-tabs__content--multiple': hasMultipleTabs}"
+				class="app-sidebar-tabs__content">
 				<slot :active-tab="activeTab" />
 			</div>
 		</aside>
@@ -135,6 +137,9 @@ export default {
 		},
 		hasFigure() {
 			return this.$slots['header'] || this.background
+		},
+		hasMultipleTabs() {
+			return this.tabs.length > 1
 		},
 		currentTabIndex() {
 			return this.tabs.findIndex(tab => tab.id === this.activeTab)
@@ -268,6 +273,8 @@ $desc-vertical-padding: 18px;
 				opacity: 1;
 			}
 		}
+
+		// header background
 		&__figure {
 			max-height: 250px;
 			height: 250px;
@@ -276,12 +283,14 @@ $desc-vertical-padding: 18px;
 			background-position: center;
 			background-repeat: no-repeat;
 		}
+
 		&__desc {
 			position: relative;
 			padding: #{$desc-vertical-padding} #{$desc-menu-right-margin * 4} #{$desc-vertical-padding} 10px;
 			&--with-star {
 				padding-left: 44px;
 			}
+			// titles
 			h3, h4 {
 				width: 100%;
 				white-space: nowrap;
@@ -289,6 +298,7 @@ $desc-vertical-padding: 18px;
 				overflow: hidden;
 				margin: 0;
 			}
+			// main title
 			h3 {
 				font-size: 16px;
 				padding: 0;
@@ -296,11 +306,13 @@ $desc-vertical-padding: 18px;
 					padding-top: 10px;
 				}
 			}
+			// subtitle
 			h4 {
 				font-size: 14px;
 				padding: 0;
 				opacity: .7;
 			}
+			// favourite
 			.app-sidebar-header__star {
 				display: block;
 				width: 44px;
@@ -310,6 +322,7 @@ $desc-vertical-padding: 18px;
 				top: $desc-vertical-padding - 12px; // aligned with main title
 				left: 0;
 			}
+			// main menu
 			.app-sidebar-header__menu {
 				position: absolute;
 				// aligned vertically in the middle
@@ -320,6 +333,8 @@ $desc-vertical-padding: 18px;
 				border-radius: 44px;
 			}
 		}
+
+		// sidebar action(s) slot
 		&__action {
 			display: flex;
 			margin: 10px;
@@ -380,8 +395,11 @@ $desc-vertical-padding: 18px;
 		}
 		&__content {
 			position: relative;
-			flex: 1 1 100%; // take full available height
-			> :not(section) {
+			// take full available height
+			flex: 1 1 100%;
+			// force the use of the tab component if more than one tab
+			// you can just put raw content if you don't use tabs
+			&--multiple > :not(section) {
 				display: none;
 			}
 		}
--- a/src/components/AppSidebar/AppSidebar.vue
+++ b/src/components/AppSidebar/AppSidebar.vue
@@ -1,31 +1,427 @@
 <!--
- - @copyright Copyright (c) 2019 Christoph Wurst <christoph@winzerhof-wurst.at>
- -
- - @author Christoph Wurst <christoph@winzerhof-wurst.at>
- -
- - @license GNU AGPL version 3 or any later version
- -
- - This program is free software: you can redistribute it and/or modify
- - it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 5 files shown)


ChristophWurst:
Instead of named slots *and* the passed array, couldn't there just be components  like `Tab`

```
<app-sidebar>
    <tab name="asldfj">
        ...
    </tab>
</app-sidebar>
```

Or am I missing something here? This could eliminate the duplication.

skjnldsv:
> Instead of named slots _and_ the passed array, couldn't there just be components like `Tab`

What would it do? the slot is just wrapped and we allow anything in the tabs, so what would the tab component do? :thinking:

ChristophWurst:
-> https://vuejsexamples.net/vue-tabs-component/ as discussed :)

skjnldsv:
Done!

src/components/AppSidebar/AppSidebar.vue
```diff
@@ -1,31 +1,308 @@
...
 <template>
-	<div id="app-sidebar">
-		<slot />
-	</div>
+	<transition name="slide-right">
+		<aside id="app-sidebar">
+			<header class="app-sidebar-header">
+				<a href="#" class="icon-close" :title="t('core', 'close')"
+					@click="closeSidebar" />
+				<div class="app-sidebar-header__figure" :style="{
```
korelstar: It would be nice if the header is optional: show it only if prop `background` and/or slot `header-background` is given.

src/components/AppSidebar/AppSidebar.vue
```diff
@@ -1,31 +1,308 @@
...
+					backgroundImage: `url(${background})`
+				}">
+					<slot class="app-sidebar-header__background" name="header-background" />
```
korelstar: Is this really about the background of the header? Seems to be more like the foreground of the header (?)

src/components/AppSidebarTab/index.js
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+import AppSidebarTab from './AppSidebarTab'
```
korelstar: the component has a wrong file name: `AppNavigationTab.vue` instead of `AppSidebarTab.vue`

src/components/AppSidebar/AppSidebar.vue
skjnldsv: @korelstar we want to enforce it on nextcloud :)
This is a decision that was made by the ui/ux designer because it actually lighten the sidebar quite a lot, and unify the overall design!

On the note, you could use undraw.co illustrations depending on the note content to show some nice visuals! We talked about it and thought that the way google is doing with calendar (create an event call `dentist` and see a nice header with a toothbrush) is a very nice way of putting a soft touch on helping the user quickly visualise what the data is about, but also is very pleasant to see :wink:

src/components/AppSidebar/AppSidebar.vue
skjnldsv: This is about the illustration on the top of the sidebar (file preview, nextcloud talk conversation... etc) :)
How do you think we should call it? :thinking:

src/components/AppSidebarTab/index.js
skjnldsv: Nice catch!

src/components/AppSidebar/AppSidebar.vue
korelstar: Yes, that would be nice for the user.

But in order to implement a dynamic image selection, there are many aspects to consider (how to load images from an external service while keeping privacy, how to find a usefull keyword from the user content, how to ensure that we don't transfer sensitive information to an external service, ...). Will there be a developer guide for all these aspects?

skjnldsv: Undraw is opensource, so we could create our service to match data to an image while keeping the images offline :)
> unDraw grants you an nonexclusive, worldwide copyright license to download, copy, modify, distribute, perform, and use the assets provided from unDraw for free, including for commercial purposes, without permission from or attributing the creator or unDraw



> Will there be a developer guide for all these aspects

We definitely should yes!

src/components/AppSidebar/AppSidebar.vue
korelstar: Maybe "teaser"/"header-teaser" or just "header"?

skjnldsv: Good for me :)

src/components/AppSidebar/AppSidebar.vue
korelstar: But wouldn't that mean that a Nextcloud installation has to ship thousands of (large) images?

skjnldsv: @korelstar those are svg, so I'd say maybe one hundred images at most? It should not be that big in the end I htink :thinking: 
We should not try to be overwhelming with tons of different images :)

@jancborchardt, @juliushaertl  thoughts?

juliushaertl: Yes, with those being svg files this shouldn't be a big issue. Also i did some testing and with proper svg optimization, we can reduce the size even further.

I don't think we should enforce using those illustrations, for calendar it makes sense to me, but for notes I don't see a proper use case.

skjnldsv: @juliushaertl @korelstar I'll tell you what, let's **try** to have illustrations as much as we can (as long as they're relevant, let's not blindly add stuff for the sake of it) but let's **not enforce it**. What do you think? :)

skjnldsv:
Which one do you prefer?

| Before | After |
|:---------:|:------:|
|dev skjnldsv com_apps_vueexample_ (12)|dev skjnldsv com_apps_vueexample_ (10)|
|dev skjnldsv com_apps_vueexample_ (6)|dev skjnldsv com_apps_vueexample_ (7)|
|dev skjnldsv com_apps_vueexample_ (8)|dev skjnldsv com_apps_vueexample_ (9)|

src/components/AppSidebar/AppSidebar.vue
korelstar: I think that this is a good tradeoff. It's a kind of pilot run and if it works, we can enforce it for all, later.

korelstar:
> Which one do you prefer? 1, 2 or 3?

Where is the difference, @skjnldsv ? Is it about one pixel margin between title and header-image?

skjnldsv:
@korelstar @ChristophWurst subtle, right? :D
This is about the paddingTop & paddingBottom of the title (and subtitle) block.

skjnldsv:
> This still looks 100% the same to me

open the right row in new tabs each and switch between them :see_no_evil:

jancborchardt:
I'd go for the one with most whitespace, I assume it's 3? :D

skjnldsv:
:bellhop_bell: :bellhop_bell: :bellhop_bell:  we've got a winner!!
You just got yourself a nice 🧸 :grin: 

_let's go with 3_

skjnldsv:
Okay, we're all good here!
here is the vuexample I'm using [vueexample.tar.gz](https://github.com/nextcloud/nextcloud-vue/files/3067515/vueexample.tar.gz)

I added some accessibility stuff according to the aria guidelines.
@nextcloud/accessibility I need some opinions on the tab/tab-content focus design feedback. When you're focusing the tabs, you can navigate with arrows and pageUp/Down. Pressing tab focus the tab content (aria guidelines) but we need a feedback for the user to know where the focus is. Same goes for the tabs navigation, we need to differentiate the focus and the selected tab.

Please test! :pray: :bowing_woman:

ChristophWurst:
- [ ] BUG: strange focus border on FF when sidebar content is in focus (I clicked the content element)
Bildschirmfoto vom 2019-04-11 09-59-45

skjnldsv:
@korelstar why would tabs be optional? 
The idea is also to try to improve the sidebar content. Currently we have a lot of apps that push a lot of thing at the same location and looks messy.
Ideally, we should regroup and tabify the data displayed in the sidebar! :)

Though I agree, you could also be too little informations to need 2 tabs or more.
Though a clear separation is needed, those are not the best looking visual :(

| 1 | 2 |
|:---------:|:------:|
|Capture d’écran_2019-04-11_16-11-21|Capture d’écran_2019-04-11_16-12-02|

korelstar:
In general, I agree totally, that tabs are useful. It just looks weird if I need one tab only. But you're right, the second variant isn't the best. Maybe you could just add more space between button 1 and the (tab) content?

jancborchardt:
Yes, @korelstar has a good point – if there is only 1 section of content, the tabs are not needed.

The tab content fields should also be self-explanatory (chat will have chat history and an input box, and detail inputs will have labels etc.) so we can leave it if there is only one tab.

skjnldsv:
@jancborchardt so, we don't change anything and we left it like the second screenshot?
I also agree that no tabs are needed in that case, but I also simply sees that the ui is not really great as is, without tabs. :wink:

jancborchardt:
@skjnldsv yup, I'd say 2nd screenshot is good!

(Of course with only 1 line of text it looks off, but that's also the case with the tab header. ;)

skjnldsv:
All done!
@korelstar if no tabs used, you can just put raw content without the tab component. :)

korelstar:
@skjnldsv 
> if no tabs used, you can just put raw content without the tab component. :)

No, this doesn't work, because you get an error "Error in mounted hook: "TypeError: this.tabs[0] is undefined"

But using a single tab is fine for me. This works as expected, now.

skjnldsv:
@korelstar ah, I might have forgot a check somewhere :)

skjnldsv:
@korelstar could you open a new issue with the full log stack? :)

korelstar:
Even better: I've provided a fix for that problem in #359. :-)

PR #342, opened 2019-04-09
Action and popovermenu accessibility

skjnldsv:
See aria guidelines https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20161214/examples/menu-button/menu-button-1/menu-button-1.html

Added:
- visual feedback on the menu
- can navigate through popovermenu with up/down pageUp/pageDown tab/shiftTab
- can close popovermenu with escape
- can open popovermenu with space and enter

Missing?
- A-Z navigation
- `tabindex="-1"` on all menu items

@nextcloud/accessibility @nextcloud/vue thoughts and reviews? :heart:

### DIFF
--- a/src/components/Action/Action.vue
+++ b/src/components/Action/Action.vue
@@ -126,7 +126,7 @@ export default {
 			this.opened = !this.opened
 			// focus first on menu open after opening the menu
 			if (this.opened) {
-				this.$nextTick(function() {
+				this.$nextTick(() => {
 					this.focusFirstAction()
 				})
 			}
--- a/src/components/Action/Action.vue
+++ b/src/components/Action/Action.vue
@@ -20,10 +20,14 @@
   -
   -->
 
+<!-- Accessibility guidelines:
+https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-actions.html -->
+
 <template>
 	<!-- if only one action, check if we need to bind to click or not -->
-	<a v-if="isSingleAction" href="#" :class="firstAction.icon"
-		class="action-item action-item--single" @click="firstAction.action ? firstAction.action : null" />
+	<a v-if="isSingleAction" :href="firstAction.href ? firstAction.href : '#'"
+		:class="firstAction.icon" class="action-item action-item--single"
+		@[firstActionEvent]="execFirstAction" />
 	<!-- more than one action -->
 	<div v-else
 		:class="{'action-item--open': opened}"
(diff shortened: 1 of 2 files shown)


src/components/Action/Action.vue
```diff
@@ -22,19 +22,31 @@
...
+		class="action-item action-item--single" @click="firstAction.action ? firstAction.action : null" />
+	<!-- more than one action -->
+	<div v-else
+		:class="{'action-item--open': opened}"
+		class="action-item"
+		@keydown.up.exact.prevent="focusPreviousAction"
+		@keydown.down.exact.prevent="focusNextAction"
+		@keydown.shift.tab.prevent="focusPreviousAction"
+		@keydown.tab.exact.prevent="focusNextAction"
+		@keydown.page-up.exact.prevent="focusFirstAction"
```
skjnldsv: oh man, this is so good!! :star2:

src/components/Action/Action.vue
```diff
@@ -100,16 +124,41 @@ export default {
...
+		focusFirstAction() {
+			this.focusIndex = 0
+			this.focusAction()
+		},
+		focusLastAction() {
+			this.focusIndex = this.$el.querySelectorAll('.focusable').length - 1
+			this.focusAction()
+		},
+		// exec the first action and prevent default
+		execFirstAction(e) {
```
skjnldsv: Strangely using `click.prevent` does not work. So I had to do that :woman_shrugging:

jancborchardt:
Nice nice!

> A-Z navigation

What do you mean by that?

> tabindex="-1" on all menu items

Why that? "-1" means they would be taken out of the tab order, which should be avoided. (Ideally there is no tabindex needed at all, because the HTML elements and order should be accessible.)

skjnldsv:
@jancborchardt read the aria docs :p

> A-Z, a-z | Moves focus to the next popup item that starts with that letter.If no popup menu item starts with the letter, focus does not move.

> |Role | Attribute | Element | Usage |
> | -- | -- | -- | -- | 
> | menuitem |   | li | The role="menuitem" attribute identifies the li element as an ARIA menuitem widget.Accessible name for the menuitem widget comes from child text content of the li element.The menuitem widget has tabindex="-1".|
> | | tabindex="-1" | li | Identifies each li[role=menuitem] as an interactive element on the page (e.g. can recived keyboard focus), but is not part of the tab order of the page.|

src/components/Action/Action.vue
```diff
@@ -100,16 +124,41 @@ export default {
 	methods: {
 		toggleMenu() {
 			this.opened = !this.opened
+			// focus first on menu open after opening the menu
+			if (this.opened) {
+				this.$nextTick(function() {
```
ChristophWurst: while it might be okay here, you should rather use arrow functions for this type of code because then you know `this` will always point to the context of the outer function. with `function` it could be something different.

skjnldsv: Vue automatically bind this to any function in nexttick :)
Otherwise it would not work :wink: 
Though arrow function looks nicer, let's change it!

PR #345, opened 2019-04-11
improve AppNavigationSettings slide-up transition

korelstar:
Thanks for adding the transition, @skjnldsv !

There are three things I would like to improve:

- [x] styleguide/eslint warning: put `v-show` to the front
- [x] make the transition phase a little bit longer (it was 200ms in server, and `var(--animation-quick)` is 100ms which is way too fast -- at least for me)
- [x] Hide the scroll-bar during transition phase. This looks weird if the content is large enough that the scroll-bar is shown when settings are opened. But it looks much more weirder if the content is small (typical use case) and the scroll-bar is not shown when settings are opened, but the scroll-bar appears during transition.

### DIFF
--- a/src/components/AppNavigationSettings/AppNavigationSettings.vue
+++ b/src/components/AppNavigationSettings/AppNavigationSettings.vue
@@ -81,8 +81,9 @@ export default {
 
 .slide-up-leave-active,
 .slide-up-enter-active {
-	transition-duration: var(--animation-quick);
+	transition-duration: 200ms;
 	transition-property: max-height, padding;
+	overflow-y: hidden !important;
 }
 
 .slide-up-enter,
--- a/src/components/AppNavigationSettings/AppNavigationSettings.vue
+++ b/src/components/AppNavigationSettings/AppNavigationSettings.vue
@@ -81,7 +81,7 @@ export default {
 
 .slide-up-leave-active,
 .slide-up-enter-active {
-	transition-duration: 200ms;
+	transition-duration: var(--animation-slow);
 	transition-property: max-height, padding;
 	overflow-y: hidden !important;
 }


src/components/AppNavigationSettings/AppNavigationSettings.vue
```diff
@@ -81,8 +81,9 @@ export default {
 
 .slide-up-leave-active,
 .slide-up-enter-active {
-	transition-duration: var(--animation-quick);
+	transition-duration: 200ms;
```
skjnldsv: @jancborchardt we should probably have this in a variable somewhere?

skjnldsv:
Ahah, sorry we were too fast :D

korelstar:
Like always ;-)

src/components/AppNavigationSettings/AppNavigationSettings.vue
jancborchardt: Yes, let's do it via variable. :) @korelstar could you adjust that?

korelstar: Yes, of course. I assume that we should put that variable to nextcloud/server (?). Should it be really "slow" or maybe "medium"?

skjnldsv: medium, and put slow at 300? :thinking:

korelstar: 300ms is already in use in https://github.com/nextcloud/server/blob/f2e9babaabeaecdf84bc7ae57cc7be6c7e16e431/apps/files/css/files.scss, so this sound reasonable.

korelstar: ... and many more: https://github.com/search?q=org%3Anextcloud+300ms&type=Code

jancborchardt: Sounds good! For simplicity ot might even be better to only have 2 animation speeds?

korelstar: @jancborchardt Yes, but then we have to decide which one to chose. There are also several usages of 200ms: https://github.com/search?q=org%3Anextcloud+200ms&type=Code
For the settings, I think 300ms is also fine. But I don't know if it is for the others.

skjnldsv: Let's go for 300ms and don't create a medium value :)

skjnldsv: https://github.com/nextcloud/server/pull/15107#issuecomment-483180903

PR #353, opened 2019-04-18
Center popoverby default and implement popovermenu variables

skjnldsv:
We should really change the way we do things here
We should split each action type into its own component.

QUESTIONS @nextcloud/vue :
- shall we move out from using actions objects and use slots with real components?
  I feel like it would be muuuuch cleaner
  ```vue
   <action>
      <action-button :class="icon-add">Add thing</action-button>
      <action-input :class="icon-user" @submit="onSubmit" @input="onInput"></action-input>
   <action>
  ```

### DIFF
--- a/src/assets/variables.scss
+++ b/src/assets/variables.scss
@@ -1,0 +1,27 @@
+/**
+ * @copyright Copyright (c) 2019 John Molakvoæ <skjnldsv@protonmail.com>
+ *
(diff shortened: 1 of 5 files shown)


ChristophWurst:
>  shall we move out from using actions objects and use slots with real components?
>  I feel like it would be muuuuch cleaner

Yes :raised_hands:

PR #358, opened 2019-04-19
Split actions into individual components

### LINKED ISSUE #357: A single input actions should render as popovermenu
This is lib is trying to be clever and renders action components with a single entry as button. This does not work if you want a text (or checkbox) input form. In that case we should always render the menu.

skjnldsv:
READY FOR REVIEWS: see [comment bellow](https://github.com/nextcloud/nextcloud-vue/pull/358#issuecomment-486322707)

- [x] Action is not Actions
- [x] ActionLink
- [x] ActionButton
- [x] ActionCheckbox
- [x] ActionInput
- [x] ActionText
- [x] ActionRouter
- [x] Actions bump
- [x] Actions accessibility fixes
- [x] Modal bump
- [x] Sidebar bump
- [ ] ~~AppNavigation dependency~~ to come later since it does not use the action component

### Goals
1. Be able to have custom data and events bound to the root element
2. Be able to just display text inside the component `<xxx>This is the action</xxx>`
3. Prevent slot from displaying anything else than text
4. Have an independent-from-server-and-or-ppopover style (aka should look perfect outside of a popovermenu)

### Breaking changes
- sidebar now uses two named slots (I'm open to suggestion for the names :see_no_evil:):
  1. `primary-actions` for whatever is displayed up front (button to start the call, datepicker for the calendar app...)
  2. `secondary-actions` for the actions menu 
- Modal now uses a named slot `actions`

### Enhancements
- Chose the Actions menu alignment with the `menu-align` prop
  ```js
  menu-align="left||right"
  ```

@nextcloud/vue 

Fix #357


src/components/ActionLink/ActionLink.vue
```diff
@@ -0,0 +1,161 @@
...
+
+<script>
+export default {
+	name: 'ActionLink',
+
+	props: {
+		href: {
+			type: String,
+			default: '#',
+			required: false
```
ChristophWurst: I would argue a link without href does not make sense. This should be required.

skjnldsv: No, because the default is `#`, so this is not required :)

ChristophWurst: how does an `ActionLink` make sense if you don't specify a link? What use case would that cover?

ChristophWurst: (the default should be removed as well)

skjnldsv: @ChristophWurst if there is no action nor link, so just text I think?
For only action I guess a button is better? No idea really :man_shrugging:

ChristophWurst: But an `a` to `#` has a certain semantic. This might be problematic for accessibility and if the click event is not prevented the browser changes the URL and therefore triggers a Vue router navigation.

My suggestion:
* If you want to render an `a` with an external link: ActionLink
* If you want to render a button with a js callback: ActionButton
* If you want to display text only: ActionText
* If you want to render a Vue router link: ActionRouterLink

skjnldsv: @ChristophWurst shall we really have empty text without actions? (ActionText)
And Router in popover as well? ^^

ChristophWurst: Yes, I use a text-only entry in Mail to display info about a folder. I think this is useful.

And vue router makes sense as well, e.g. for the *Edit* action of a Mail account. Having a router link makes it possible to open that link in a new tab. With callback and programmatic navigation this does not work.

ChristophWurst: Right now I even have to disable an eslint rule that complains that I'm using a sideeffect inside a computed property -> the programmatic navigation inside the callback of the *Edit* action of an account.

skjnldsv:
Okay everyone! We're good for review!
This is quite a change since I created a variable.scss file to help us stay organise and I was stupid enough to update our most common values instead of doing so in another pr :see_no_evil: 

BUT, fear no more my fellow vuers! I kept the commit history since I'm going to squash and merge to keep our repo clean (Hi @ChristophWurst)

You will find the vueexample app I used to play and test (messy :soccer:, but we don't care) :grin: 
[vueexample.tar.gz](https://github.com/nextcloud/nextcloud-vue/files/3113324/vueexample.tar.gz)


Please give also feedback about accessibility stuff and design (menu is fully navigable with keyboard and feedback should be on point :ok_hand: !

Thanks everyone!! :heart:

ChristophWurst:
- [ ] BUG: actions in content details list is not right-aligned with *…* button

Bildschirmfoto von 2019-04-25 10-41-39

For comparison, this is the old version: 
Bildschirmfoto von 2019-04-25 10-43-14

skjnldsv:
> * [ ]  BUG: actions in content details list is not right-aligned with _…_ button

This is a request by @jancborchardt to have menu centered by default now :)

ChristophWurst:
> This is a request by @jancborchardt to have menu centered by default now :)

I'm objected with this *bug requests* tbh

ChristophWurst:
Was this changed in this PR or another one?

skjnldsv:
> Was this changed in this PR or another one?

This pr! :)

Done, `menu-align="'left'||'right'"` is added

src/mixins/action.js
```diff
@@ -0,0 +1,60 @@
...
+	computed: {
+		isIconUrl() {
+			try {
+				return new URL(this.icon)
+			} catch (error) {
+				return false
+			}
+		},
+		text() {
+			return this.$slots.default[0].text
```
ChristophWurst: this will :boom: if you forget to fill the slot. Shall we warn in that case?

skjnldsv: @ChristophWurst nice idea yes!!

skjnldsv:
Done!

src/mixins/action.js
skjnldsv: Done

PR #381, opened 2019-05-08
AppNavigationItem: synced prop for open state

korelstar:
With this, the parent component has full control over the open state of a collapsible item. See also #180.

Please note: `item.opened` is not used anymore. Hence, this is a breaking change!


ChristophWurst:
> I don't get the difference. You already had the control on the parent? thinking

Only for the popovermenu, but for the collapsed children

PR #382, opened 2019-05-08
Add AppContentList and AppContentDetails components

ChristophWurst:
Implements
* https://docs.nextcloud.com/server/stable/developer_manual/design/content.html
* https://docs.nextcloud.com/server/stable/developer_manual/design/list.html (without the actual items, they will follow later)

These components were extracted from Mail. This is the bare minimum of functionality you would need.

<details>

Follow-up tasks:
- Find a common way to handle empty content lists (special `emtpy` slot, switched with prop?)
- Find a common way to display "new" items are available in a list. Maybe too specific, but we need that for mail
- Find (and hopefully enforce) a decent animation for new/removed items in the details list
- Find a common design for the details, e.g. to show a header that somewhat looks the same across all apps (with arbitrary content, of course -> slot?)

</details>

cc @nextcloud/vue

### DIFF
--- a/src/components/AppContentDetails/AppContentDetails.vue
+++ b/src/components/AppContentDetails/AppContentDetails.vue
@@ -1,0 +1,33 @@
+<!--
+ - @copyright Copyright (c) 2019 Christoph Wurst <christoph@winzerhof-wurst.at>
+ -
+ - @author Christoph Wurst <christoph@winzerhof-wurst.at>
+ -
+ - @license GNU AGPL version 3 or any later version
+ -
+ - This program is free software: you can redistribute it and/or modify
+ - it under the terms of the GNU Affero General Public License as
+ - published by the Free Software Foundation, either version 3 of the
+ - License, or (at your option) any later version.
+ -
+ - This program is distributed in the hope that it will be useful,
+ - but WITHOUT ANY WARRANTY; without even the implied warranty of
+ - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ - GNU Affero General Public License for more details.
+ -
+ - You should have received a copy of the GNU Affero General Public License
+ - along with this program. If not, see <http://www.gnu.org/licenses/>.
+ -
+ -->
+
+<template>
+	<div class="app-content-details">
+		<slot />
+	</div>
+</template>
+
+<script>
+export default {
+	name: 'AppContentDetails'
+}
+</script>
--- a/src/components/AppContentDetails/index.js
+++ b/src/components/AppContentDetails/index.js
@@ -1,0 +1,24 @@
+/*
+ * @copyright 2019 Christoph Wurst <christoph@winzerhof-wurst.at>
+ *
+ * @author 2019 Christoph Wurst <christoph@winzerhof-wurst.at>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+import AppContentDetails from './AppContentDetails'
+
+export default AppContentDetails
+export { AppContentDetails }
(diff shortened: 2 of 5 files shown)


jancborchardt:
Since it’s extracted from the Mail app, we need to adjust it so it uses the correct spacing.

The Mail app (on the left) seems to have wrong spacing at the moment, the Contacts app (on the right) has correct one. Note the first and second row being too far apart in Mail:
mail app spacing

jancborchardt:
(Also see the design docs for the proper spacing – maybe it’s a local issue with me in Mail?
list

ChristophWurst:
I think this is caused by the transition element wrapping the list, so unrelated to the change here. The simple components of this PR are 100% what's in the manual.

src/components/AppContentDetails/AppContentDetails.vue
```diff
@@ -0,0 +1,31 @@
...
+ -->
+
+<template>
+	<div class="app-content-details">
+		<slot />
+	</div>
+</template>
+
+<script>
+export default {}
```
skjnldsv: Component name

src/components/AppContentListItem/AppContentListItem.vue
```diff
@@ -0,0 +1,57 @@
...
+		<div v-if="$slots.menu"
+			class="app-content-list-item-menu">
+			<slot name="menu" />
+		</div>
+	</router-link>
+</template>
+
+<script>
+export default {
+	props: {
```
skjnldsv: Component name

src/components/AppContentListItem/AppContentListItem.vue
```diff
@@ -0,0 +1,57 @@
...
+ - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ - GNU Affero General Public License for more details.
+ -
+ - You should have received a copy of the GNU Affero General Public License
+ - along with this program. If not, see <http://www.gnu.org/licenses/>.
+ -
+ -->
+
+<template>
+	<router-link class="app-content-list-item" :to="to">
```
skjnldsv: We might have different uses than vue-router?

src/components/AppContentListItem/AppContentListItem.vue
```diff
@@ -0,0 +1,57 @@
...
+		<div v-if="$slots.icon"
+			class="app-content-list-item-icon">
```
skjnldsv: I would actually prefer the use of a prop for that. :wink:
If we want to support Avatar, then a slot for avatar only and add checks to make sure the component in the avatar slot have the name `avatar` :grin: 
Same goes for the texts, this is supposed to be text, so let's not add the option to have anything else. This goes for
- line-one
- line-two
- details

Actions make sense to have a slot for. Though could be default slot. :thinking:

jancborchardt:
@ChristophWurst aaah good find! @skjnldsv does the menu push the second row down cause of the padding?

Also in your screenshots note that the avatar/placeholder is not vertically centered in the row (like it is in the screenshots of the docs and Contacts). If I remember correctly that’s because of the reply and attachment icons – but it’s fine if they partly overlap the avatar, like the star does. Or maybe even better if they are on the same row as the subject, left of the 3-dot-menu.

skjnldsv:
> @ChristophWurst aaah good find! @skjnldsv does the menu push the second row down cause of the padding?

Most likely our Actions component needs a little tweak to fit our standards here yes :)

skjnldsv:
> Also in your screenshots note that the avatar/placeholder is not vertically centered in the row (like it is in the screenshots of the docs and Contacts).

I'm guessing it is because we request the default avatar at 32 and @ChristophWurst did not yet forced the use if 40px in the AVatar component :wink:

src/components/AppContentListItem/AppContentListItem.vue
ChristophWurst: So? What is your suggestion? Having a div as root element and optional link inside? Or let the app add the link?

skjnldsv: Either two props with :to and :href, or two components.
`AppContentListItemRouter.vue` `AppContentListItemLink.vue` ?

src/components/AppContentListItem/AppContentListItem.vue
ChristophWurst: Mail needs a line-two slot as it's adding icons for replied messages and ones with attachments. A prop would not allow that.

src/components/AppContentListItem/AppContentListItem.vue
ChristophWurst: let's add that in a follow-up PR.

src/components/AppContentListItem/AppContentListItem.vue
ChristophWurst: I'll use a custom component for Mail

skjnldsv:
What appended to the appContentListItem? :)

ChristophWurst:
I don't need the limited version right now: https://github.com/nextcloud/nextcloud-vue/pull/382#discussion_r288121435

skjnldsv:
But others could need it ;)

ChristophWurst:
Yes, that is why we'll add it in a follow-up PR 😉

PR #384, opened 2019-05-09
Add tooltip title, accessibility & css cleanup

skjnldsv:
Capture d’écran_2019-05-09_14-47-39

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -33,24 +33,37 @@
 						{{ title }}
 					</div>
 					<div class="icons-menu">
+						<!-- Actions menu -->
 						<Actions class="header-actions">
 							<slot name="actions" />
 						</Actions>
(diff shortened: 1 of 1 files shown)


src/components/Modal/Modal.vue
```diff
@@ -33,24 +33,37 @@
...
 					<div class="icons-menu">
+						<!-- Actions menu -->
 						<Actions class="header-actions">
 							<slot name="actions" />
 						</Actions>
-						<a v-if="hasNext && enableSlideshow" class="play-pause" @click="togglePlayPause">
+
+						<!-- Play-pause toggle -->
+						<button v-if="hasNext && enableSlideshow"
+							v-tooltip.auto="playPauseTitle"
```
korelstar: I haven't found this in the documentation, but I assume that `auto` is the default position. Hence, if it makes no difference, I would remove `.auto`.

skjnldsv: I'm not sure! :thinking:

skjnldsv:
So, let's get this in?

PR #392, opened 2019-05-13
Also allow relative url

skjnldsv:

### DIFF
--- a/src/components/ActionLink/ActionLink.vue
+++ b/src/components/ActionLink/ActionLink.vue
@@ -75,11 +75,11 @@ export default {
 			default: '#',
 			required: true,
 			validator: value => {
-				// href is either an anchor or a valid url
+				// href is either an anchor or a valid url starting with a scheme or a relative path
 				try {
 					return new URL(value)
 				} catch (error) {
-					return value.startsWith('#')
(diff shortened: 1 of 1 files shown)


src/components/ActionLink/ActionLink.vue
```diff
@@ -75,11 +75,11 @@ export default {
 			default: '#',
 			required: true,
 			validator: value => {
-				// href is either an anchor or a valid url
+				// href is either an anchor or a valid url starting with a scheme or an absolute path
 				try {
 					return new URL(value)
 				} catch (error) {
-					return value.startsWith('#')
+					return value.startsWith('#') || value.startsWith('/')
```
ChristophWurst: relative? :P

skjnldsv: how could we check that? :thinking:

ChristophWurst: no. a URL that start with `/` is *relative*. One that starts with a scheme or the host is *absolute*

ChristophWurst: The PR description (and the source code comment) should say *Also allow relative url*

skjnldsv: :see_no_evil:

PR #396, opened 2019-05-15
Add Nextcloud linting

ChristophWurst:


package.json
```diff
@@ -29,6 +29,7 @@
 		"dist"
 	],
 	"dependencies": {
+		"eslint-plugin-nextcloud": "^0.1.0",
```
korelstar: Shouldn't that be a dev dependency?

ChristophWurst: yes!

PR #398, opened 2019-05-16
Fix actions menu grey focus colour

skjnldsv:
Shadow was a bit too strong and not enough contrasted on dark bg :)

| White | Dark |
|:---------:|:------:|
|Peek 16-05-2019 08-05|Peek 16-05-2019 08-06|

### DIFF
--- a/src/components/Actions/Actions.vue
+++ b/src/components/Actions/Actions.vue
@@ -294,7 +294,8 @@ $arrow-margin: ($clickable-area - 2 * $arrow-width)  / 2;
 	&__menutoggle:active,
 	&.action-item--open {
 		border-radius: $clickable-area / 2;
-		background-color: var(--color-box-shadow);
+		// good looking on dark AND white bg
+		background-color: rgba(127, 127, 127, .25);
 		&,
 		.action-item__menutoggle {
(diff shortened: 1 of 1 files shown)


jancborchardt:
Great great stuff. One thing though, in Mail:
nextcloud mail more menu

Looks a bit strange with the border overlay. I’d even say let’s simply cut the border as the whitespace is enough. Will look much simpler, we already did it in the Android app (both Files app and Notes app) and it’s really nice.

skjnldsv:
@jancborchardt you mean overflow hide?

jancborchardt:
@skjnldsv no, the circle should stay round. :) I simply mean `.app-content-list .app-content-list-item { border-top:none;`:

Current, and then without lines – much simpler and the whitespace is enough to divide (just like we also don’t use divider lines in the app-navigation for example):
contact list current contact list no lines

What do you think?

skjnldsv:
Looks nice :+1: 
I'll try to use it a bit to see if usability is nice as well :)

jancborchardt:
Pull request for core open at https://github.com/nextcloud/server/pull/15691 but kinda needs some fixes in Mail / action menu component as mentioned in there. :)

PR #399, opened 2019-05-16
Fix action input width and randomid

skjnldsv:
@nextcloud/vue

### DIFF
--- a/src/components/ActionCheckbox/ActionCheckbox.vue
+++ b/src/components/ActionCheckbox/ActionCheckbox.vue
@@ -46,7 +46,8 @@ export default {
 	props: {
 		id: {
 			type: String,
-			default: 'checkbox-' + GenRandomId()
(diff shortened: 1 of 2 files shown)


src/components/ActionCheckbox/ActionCheckbox.vue
```diff
@@ -46,7 +46,12 @@ export default {
 	props: {
 		id: {
 			type: String,
-			default: 'checkbox-' + GenRandomId()
+			// we have to use computed/data
+			// or else GenRandomId will return the same string
+			// for all the ActionCheckbox instances
+			// most likely that the code gets executed on compile instead
```
ChristophWurst: this code is evaluated when the module is parsed. you could try using a factory function like `default: () => GenRandomId()`

skjnldsv: Ooooh! You're right, lemme try!

PR #400, opened 2019-05-16
Make action text reactive

skjnldsv:
Does anyone have a better solution?
@nextcloud/vue

### DIFF
--- a/src/mixins/actionGlobal.js
+++ b/src/mixins/actionGlobal.js
@@ -33,12 +33,27 @@ export default {
 		}
 	},
 
+	beforeUpdate() {
+		this.text = this.getText()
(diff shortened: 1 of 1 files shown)


skjnldsv:
@ChristophWurst though most unlikely we get a popovermenu update that often. :)

ChristophWurst:
> @ChristophWurst though most unlikely we get a popovermenu update that often. :)

Yes, but I don't know if any updates of parent components can trigger that method as well.

skjnldsv:
> Yes, but I don't know if any updates of parent components can trigger that method as well.

if anything changes the component itself, yes (show/hide, linked variable... props...) :)

PR #404, opened 2019-05-17
Fix #402

### LINKED ISSUE #402: Missing spacing on long name ellipsis in multiselect
nextcloud-vue v0.11.2.

With https://github.com/nextcloud/tasks/pull/412 there is a space missing in between the first and second part of the names in a multiselect, see
multiselect_ellipsis
("showfull" instead of "show full" and "hidethis" instead of "hide this").

~~This is because the space is cropped for these two entries, whereas it is correctly shown for the second string.~~

Edit: The string is split correctly, but the space ends up at the end of the first string, which is set to `white-space: nowrap;`, so the space is not shown there.

raimund-schluessler:


raimund-schluessler:
#403 was faster.

skjnldsv:
Can we delete this issue? :D
#404

raimund-schluessler:
> Damn, we can only delete issues 😢

Sorry to have spammed your PR list 😉

skjnldsv:
@raimund-schluessler that's ok, it was a bad joke to do anyway :grin:

PR #439, opened 2019-06-16
Build docs using vue-styleguidist

juliushaertl:
Next try to come up with some documentation. Unfortunately vuepress didn't fit that well, so this replaces #272 

I gave different tools a try and ended up with [https://vue-styleguidist.github.io/](vue-styleguidist), which generates documentation from code and allows interactive example rendering with live editing capabilities.

For preview I currently setup a build for this branch on netlify: https://nextcloud-vue-components.netlify.com

We could already get this in as a starting point, as it already allows browsing the existing components with props, events.


src/components/Avatar/Avatar.vue
```diff
@@ -20,6 +20,21 @@
   -
   -->
 
+<docs>
```
skjnldsv: Like it, so simple!!!

src/components/Actions/Actions.vue
```diff
@@ -91,6 +110,9 @@ const allowedChildren = [
 	'ActionText'
 ]
 
+/**
+ * @since 0.10.0
```
skjnldsv: Yeeeesss!!

skjnldsv:
A few travis errors :)

juliushaertl:
> A few travis errors :)

Fixed.

I also tried to add some documentation for the Actions component, props of all others as well some very basic examples for Modal and Avatar. I guess we can improve the actual docs step by step once this is merged.

I just encountered that netlify also generates a preview of the documentation for every pull request, that is then linked in the status check list: https://deploy-preview-439--nextcloud-vue-components.netlify.com/

docs/index.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Recommendations
+
+Some components require Nextcloud functionality to work, so it is currently recommended to extend Vue with the following:
+
+```js static
+Vue.prototype.t = window.t
+Vue.prototype.n = window.n
+Vue.prototype.OC = window.OC
+Vue.prototype.OCA = window.OCA
```
korelstar: For what is this needed?

If it's only for lint checks, extending [`plugin:nextcloud/recommended`](https://github.com/nextcloud/eslint-plugin-nextcloud) in eslint configuration should be better.

juliushaertl: Probably only t and OC are required, but they don't hurt as long as we don't ship translations with the components:
https://github.com/nextcloud/nextcloud-vue/search?l=Vue&q=t%28%27core%27
https://github.com/nextcloud/nextcloud-vue/search?q=getCurrentUser&unscoped_q=getCurrentUser

juliushaertl: Ah, OC probably also not needed since it is accessing the global object directly without the Vue object as a proxy.

korelstar: I don't like that app developers have to care about internals from the used library. We should add a [mixin like this](https://github.com/nextcloud/notes/blob/master/src/mixins/AppGlobal.js) to nc-vue and use it in those components which use `t` or similar. App developers then don't have to add this prototype code anymore. They can even use this mixin, too, when they want to us `t` in their app.

juliushaertl: Yes, sounds good. I just added it here as I encountered it in the docs build.

juliushaertl:
> Codacy/PR Quality Review — Not up to standards. This pull request quality could be better.

@skjnldsv Those seem to be only the minified server styles I've added for now, since otherwise the demo components look quite odd, as we still rely on the shipped CSS. I'd say those can be ignored.

docs/index.md
korelstar: Fine, I can do a PR for the mixin, later (in the next days, I think).

skjnldsv:
So, merging?

skjnldsv:
All good @juliushaertl?

juliushaertl:
Yep, good to go 🚢

PR #447, opened 2019-06-26
Revert "AppSidebar: fix tab validation: tabs are optional"

skjnldsv:
Reverts nextcloud/nextcloud-vue#445

### DIFF
--- a/src/components/AppSidebar/AppSidebar.vue
+++ b/src/components/AppSidebar/AppSidebar.vue
@@ -166,27 +166,22 @@ export default {
 	},
 	mounted() {
 		// Init tabs from $children
-		const tabs = this.$children.filter(comp => comp.$options.name === 'AppSidebarTab')
-		if (tabs.length === 0 || tabs.length === this.$children.length) {
-			this.tabs = tabs.reduce((tabs, tab) => {
-				if (!tab.name || typeof tab.name !== 'string') {
-					Vue.util.warn(`This tab is missing a valid name: ${tab.name}`, tab)
-					return tabs
-				}
-				if (!IsValidString(tab.id)) {
-					Vue.util.warn(`This tab is missing a valid id: ${tab.id}`, tab)
-					return tabs
-				}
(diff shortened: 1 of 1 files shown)


skjnldsv:
Hey @korelstar 
There was an issue with your fix. If you're not using the appsidebarTab component, then the tabs are always empty, so using reduce on an empty array will never check anything :)

skjnldsv:
Could you open again a fix so we can have a better look at it? :)
You're awesome! Tanks! :hugs:

korelstar:
What exactly didn't work and why should it work? It is intended, that non-AppSidebarTab's are not shown as tabs but as normal content.

skjnldsv:
Issue is that you can use other components as tabs, as long as you're providing the appropriate icon and name and id. Now if you check the tab component only, it will forbid it :)

(sorry about the late answer)

You can see the current pending pr that broke after your patch : https://github.com/nextcloud/server/pull/15719
:wink:

korelstar:
Okay, then I suggest to make the check as you wrote: a children is treated as tab if it has icon, id and name, otherwise it is treated as non-tab content. However, then your detailed warnings are not possible anymore.

The only other solution I see is to revert #420, since that enforces tabs, which is not intended.

skjnldsv:
I'll adapt your pr and 420 :)
Thanks for the help @korelstar ! :hugs: 

PS: are you coming at the conference this year? :rocket:

korelstar:
What about using different template slots for tabs and for non-tab-content? Then, it would be easy to check if and warn if not only one of them is used. I think that would provide a much clearer distinction between these two use cases. Unfortunately, this would be a breaking change (slot has to be used for tabs).

What do you think, @skjnldsv ?

korelstar:
> I'll adapt your pr and 420 :)

@skjnldsv Any news on this? There is still an error message when using no tabs in AppSidebar. Are you fine with slitting tabs and non-tab content in two different template slots? Then I could provide a PR.

PR #448, opened 2019-06-27
Add updateTabs function to appSidebar

skjnldsv:
Allow to trigger a ytab refresh from the parent
You can use `ref="sidebar"` and then `this.$refs.sidebar.updateTabs()` from the parent :)

I don't think we can watch over changed childrens. The `updated` method is triggered far too many times for other changes than $children change. :man_shrugging:


skjnldsv:
@chrisophwurst, bump?

PR #473, opened 2019-07-10
Fix actionInput margins

skjnldsv:
Was triggered even if the item was on the middle as any li could have been a far parent of the menu


juliushaertl:
@skjnldsv I'm just wondering, any idea why the list element was not using a scoped selector? That should already have avoided leaking to parent elements.

skjnldsv:
@juliushaertl it is using a scope selector :)
But the scope [data] is only applied to the last element, not all of them

korelstar:
Maybe the problem is due to this: https://vuejs.org/v2/style-guide/#Element-selectors-with-scoped-use-with-caution (see explanation)

The `li` doesn't use a class selector which is not recommended!

PR #486, opened 2019-07-17
Feature/470/split appnavigationitem into small components

### LINKED ISSUE #376: AppNavigationItem extensions necessary for Talk
* Counter should be possible to be String:
  - we use logic like `unreadMessages > 99 ? '99+' : unreadMessages` so it doesn't break with 3+ digit unread counts.
* Have multiple "counters":
  - Apart from the unread counter, we also have a counter `@` which signals that you were mentioned in a chat
* Counter should have assignable classes (at least primary, default grey, or none) (used for the @-mention to draw more attention to it)
* Separator in the actions
* iconUrl should be outside of the padding, just like the `icon` when it is using a CSS (see screenshot 2)
* Allow input as an option for the AppNavigationNew for search/filtering of the list as well as creating new entries
* Allow to have a favorite (generic icon) marker overlapping the `icon`/`iconUrl`

See screenshot for current example:
> Bildschirmfoto von 2019-05-02 15-22-24

icon vs. iconUrl:
> Bildschirmfoto von 2019-05-02 12-33-00

### LINKED ISSUE #470: Split AppNavigationItem into small components
Same as https://github.com/nextcloud/nextcloud-vue/pull/358 and after #25
https://github.com/nextcloud/nextcloud-vue/pull/195
Refs: https://docs.nextcloud.com/server/16/developer_manual/design/navigation.html#app-navigation-menu

- [x] Entry with counter
- [x] Use Actions components
- [x] Collapsible component
- [x] Undo (with transition: slide)
- [x] Edit (with transition: fade)
- [x] Pinned item
- [x] Nested AppNavigationItem

``` vue
<AppNavigationItem counter="test">
	This is the main entry
	<AppNavigationItem>Sub-element 1</AppNavigationItem>
	<AppNavigationItem>Sub-element 2</AppNavigationItem>
	<template #actions>
		<ActionButton @click="doSomething" />
	</template>
</AppNavigationItem>
```

ma12-co:
Fix #470 
Fix #376 

- [x] Replace the pinned script with pure css;
- [x] Fix the out of bounds popover menu;
- [x] Fix collapsing icon with long titles;
- [x] Fix overflow text of AppNavigationItem's title;
- [x] Emit click event from AppNavigationCaption;

#### Please use [this App.vue](https://gist.github.com/ma12-co/ec27beec385b016558073296e904cb24) in you vueexample app for reviewing, as the new components won't work with the old one!

### DIFF
--- a/src/assets/variables.scss
+++ b/src/assets/variables.scss
@@ -47,3 +47,11 @@ $opacity_full: 1;
 // menu round background hover feedback
 // good looking on dark AND white bg
 $action-background-hover: rgba(127, 127, 127, .25);
+
+// various structure data used in the 
+// `AppNavigation` component
+$header-height: 50px;
+$navigation-width: 300px;
+
+// mobile breakpoint
+$breakpoint-mobile: 1024px;
--- a/src/components/AppNavigation/AppNavigation.vue
+++ b/src/components/AppNavigation/AppNavigation.vue
@@ -2,6 +2,7 @@
  - @copyright Copyright (c) 2019 Christoph Wurst <christoph@winzerhof-wurst.at>
  -
  - @author Christoph Wurst <christoph@winzerhof-wurst.at>
+ - @author Marco Ambrosini <marcoambrosini@pm.me>
  -
  - @license GNU AGPL version 3 or any later version
  -
@@ -20,7 +21,7 @@
  -
  -->
 <template>
-	<div id="app-navigation">
+	<div id="app-navigation" class="vue">
 		<slot />
 	</div>
 </template>
@@ -30,13 +31,50 @@ export default {
 	name: 'AppNavigation'
 }
 </script>
-<style lang="scss">
+
+<style lang="scss" scoped>
+@import '../assets/variables.scss';
+
 #app-navigation {
 	will-change: transform;
 	transition: transform var(--animation-quick);
+	width: $navigation-width;
+	position: fixed;
+	top: $header-height;
+	left: 0;
+	z-index: 500;
+	overflow-y: auto;
+	overflow-x: hidden;
+	// Do not use vh because of mobile headers
+	// are included in the calculation
+	height: calc(100% - #{$header-height});
+	box-sizing: border-box;
+	background-color: var(--color-main-background);
+	-webkit-user-select: none;
+	-moz-user-select: none;
+	-ms-user-select: none;
+	user-select: none;
+	border-right: 1px solid var(--color-border);
+	display: flex;
+	flex-direction: column;
+	flex-grow: 0;
+	flex-shrink: 0;
 }
+
+//list of navigation items
+ul {
+	position: relative;
+	height: 100%;
+	width: inherit;
+	overflow-x: hidden;
+	overflow-y: auto;
+	box-sizing: border-box;
+	display: flex;
+	flex-direction: column;
+}
+
 // mobile view only
-@media only screen and (max-width: 768px) {
+@media only screen and (max-width: $breakpoint-mobile) {
 	// if navigation is shown
 	.nav-open #app-navigation {
 		transform: translateX(0);
--- a/src/components/AppNavigationCaption/AppNavigationCaption.vue
+++ b/src/components/AppNavigationCaption/AppNavigationCaption.vue
@@ -1,6 +1,6 @@
 <template>
-	<li class="app-navigation-caption--item">
-		{{ text }}
+	<li class="app-navigation-caption">
+		{{ title }}
 	</li>
 </template>
 
@@ -9,15 +9,16 @@
 export default {
 	name: 'AppNavigationCaption',
 	props: {
-		text: {
+		title: {
 			type: String,
 			required: true
 		}
 	}
 }
 </script>
+
 <style lang="scss" scoped>
-.app-navigation-caption--item {
+.app-navigation-caption {
 	font-weight: bold;
 	color: var(--color-text-maxcontrast);
 	line-height: $clickable-area;
@@ -27,10 +28,12 @@ export default {
 	text-overflow: ellipsis;
 	opacity: 0.7;
 	box-shadow: none !important;
-	pointer-events: none;
+	order: 1;
+	flex-shrink: 0;
 }
 
-.app-navigation-caption--item:not(:first-child) {
+// extra top space if it's not the first item on the list
+.app-navigation-caption:not(:first-child) {
 	margin-top: $clickable-area / 2;
 }
 </style>
--- a/src/components/AppNavigationCounter/AppNavigationCounter.vue
+++ b/src/components/AppNavigationCounter/AppNavigationCounter.vue
@@ -1,5 +1,5 @@
 <!--
- - @copyright Copyright (c) 2019 Marco Ambrosini <ma12co@pm.me>
+ - @copyright Copyright (c) 2019 Marco Ambrosini <marcoambrosini@pm.me>
  -
  - @author Marco Ambrosini <marcoambrosini@pm.me>
  -
@@ -25,26 +25,23 @@
 ### Normal Counter
 
 ```
-<AppNavigationCounter :highlighted="false">
-your number or string
-</AppNavigationCounter>
+<AppNavigationCounter>314+</AppNavigationCounter>
 ```
 
 ### Highlighted Counter (i.e. mentions)
 
 ```
-<AppNavigationCounter :highlighted="true">
-your number or string
-</AppNavigationCounter>
+<AppNavigationCounter :highlighted="true">@admin</AppNavigationCounter>
+<AppNavigationCounter :highlighted="true">314+</AppNavigationCounter>
 ```
 
 </docs>
 
 <template>
-	<li :class="{ highlighted }"
-		class="app-navigation-entry-utils-counter">
-		<span><slot /></span>
-	</li>
+	<div :class="{ 'app-navigation-entry__counter--highlighted': highlighted }"
+		class="app-navigation-entry__counter">
+		<slot />
+	</div>
 </template>
 
 <script>
@@ -62,25 +59,20 @@ export default {
 </script>
 
 <style lang="scss" scoped>
+.app-navigation-entry__counter {
+	overflow: hidden;
+	max-width: $clickable-area;
+	margin: 4px;
+	padding: 0 4px;
+	text-align: center;
+	text-overflow: ellipsis;
+	line-height: 1em;
 
-	.app-navigation-entry-utils-counter {
-		overflow: hidden;
-		text-align: right;
-		font-size: 9px;
-		line-height: $clickable-area;
-		padding: 0 12px; /* Same padding as all li > a in the app-navigation */
-
+	&--highlighted {
+		padding: 4px 6px;
+		color: var(--color-primary-text);
+		border-radius: 10px;
+		background-color: var(--color-primary);
 	}
-
-	.highlighted {
-		padding: 0;
-		text-align: center;
-		span {
-			padding: 2px 5px;
-			border-radius: 10px;
-			background-color: var(--color-primary);
-			color: var(--color-primary-text);
-		}
-	}
-
+}
 </style>
--- a/src/components/AppNavigationIconBullet/AppNavigationIconBullet.vue
+++ b/src/components/AppNavigationIconBullet/AppNavigationIconBullet.vue
@@ -1,0 +1,95 @@
+<!--
+  - @copyright Copyright (c) 2019 John Molakvoæ <skjnldsv@protonmail.com>
+  -
+  - @author John Molakvoæ <skjnldsv@protonmail.com>
+  -
+  - @license GNU AGPL version 3 or any later version
+  -
+  - This program is free software: you can redistribute it and/or modify
+  - it under the terms of the GNU Affero General Public License as
+  - published by the Free Software Foundation, either version 3 of the
+  - License, or (at your option) any later version.
+  -
+  - This program is distributed in the hope that it will be useful,
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+
+<docs>
+### Bullet
+This component is made to be used inside of the [AppNavigationItem](#AppNavigationItem) component slot: `icon`.
+
+```
+<AppNavigationIconBullet slot="icon" color="0082c9" />
+<AppNavigationIconBullet slot="icon" color="#0082c9" />
+
+```
+
+### AppNavigationItem example
+```
+<AppNavigationItem  title="Entry 2" :pinned="true">
+	<AppNavigationIconBullet slot="icon" color="0082c9" />
+</AppNavigationItem>
+```
+</docs>
+
+<template>
+	<div class="app-navigation-entry__icon-bullet" @click="onClick">
+		<div :style="{ backgroundColor: formattedColor }" />
+	</div>
+</template>
+
+<script>
+export default {
+	name: 'AppNavigationIconBullet',
+
+	props: {
+		color: {
+			type: String,
+			required: true,
+			validator: function(color) {
+				// #000, 000, #0082c9 and 0082c9
+				return /^#?([0-9A-F]{3}){1,2}$/i.test(color)
+			}
+		}
+	},
+
+	computed: {
+		formattedColor() {
+			if (this.color.startsWith('#')) {
+				return this.color
+			}
+			return '#' + this.color
+		}
+	},
+
+	methods: {
+		onClick(e) {
+			this.$emit('click', e)
+		}
+	}
+
+}
+</script>
+
+<style lang="scss" scoped>
+.app-navigation-entry__icon-bullet {
+	display: block;
+	// there is 2 margins
+	padding: $icon-margin + 1px;
+	div {
+		width: $icon-size - 2px;
+		height: $icon-size - 2px;
+		cursor: pointer;
+		transition: background 100ms ease-in-out;
+		border: none;
+		border-radius: 50%;
+	}
+}
+
+</style>
--- a/src/components/AppNavigationItem/AppNavigationIconCollapsible.vue
+++ b/src/components/AppNavigationItem/AppNavigationIconCollapsible.vue
@@ -1,0 +1,88 @@
+<!--
+ - @copyright Copyright (c) 2019 Marco Ambrosini <marcoambrosini@pm.me>
+ -
+ - @author Marco Ambrosini <marcoambrosini@pm.me>
+ -
+ - @license GNU AGPL version 3 or any later version
+ -
+ - This program is free software: you can redistribute it and/or modify
+ - it under the terms of the GNU Affero General Public License as
+ - published by the Free Software Foundation, either version 3 of the
+ - License, or (at your option) any later version.
+ -
+ - This program is distributed in the hope that it will be useful,
+ - but WITHOUT ANY WARRANTY; without even the implied warranty of
+ - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ - GNU Affero General Public License for more details.
+ -
+ - You should have received a copy of the GNU Affero General Public License
+ - along with this program. If not, see <http://www.gnu.org/licenses/>.
+ -
+ -->
+
+<template>
+	<!-- Button to expand or collapse children -->
+	<button
+		class="icon-collapse"
+		:class="{'icon-collapse--rotated':open}" @click="onClick" />
+</template>
+
+<script>
+export default {
+	name: 'AppNavigationIconCollapsible',
+
+	props: {
+		open: {
+			type: Boolean,
+			default: true
+		}
+	},
+
+	methods: {
+		onClick(e) {
+			this.$emit('click', e)
+		}
+	}
+
+}
+</script>
+
+<style lang="scss" scoped>
+@import '~Fonts/scss/iconfont-vue';
+
+.icon-collapse {
+	position: absolute;
+	z-index: 105; // above a, under button
+	width: 44px;
+	height: 44px;
+	margin: 0;
+	padding: 0;
+	transition: opacity var(--animation-quick) ease-in-out;
+	-webkit-transform: rotate(-90deg);
+	-ms-transform: rotate(-90deg);
+	transform: rotate(-90deg);
+	color: black;
+	border: none;
+	border-radius: 0;
+	outline: none !important;
+	background-color: transparent;
+	box-shadow: none;
+	font-size: 18px;
+
+	@include iconfont('triangle-s');
+
+	&:hover{
+		color: var(--color-primary);
+	}
+	&--rotated {
+		-webkit-transform: rotate(0deg);
+		-ms-transform: rotate(0deg);
+		transform: rotate(0deg);
+		color: var(--color-main-text);
+		&:hover{
+			color: var(--color-primary);
+		}
+	}
+}
+
+</style>
--- a/src/components/AppNavigationItem/AppNavigationItem.vue
+++ b/src/components/AppNavigationItem/AppNavigationItem.vue
@@ -2,6 +2,7 @@
   - @copyright Copyright (c) 2018 John Molakvoæ <skjnldsv@protonmail.com>
   -
   - @author John Molakvoæ <skjnldsv@protonmail.com>
+  - @author Marco Ambrosini <marcoambrosini@pm.me>
   -
   - @license GNU AGPL version 3 or any later version
   -
@@ -19,126 +20,328 @@
   - along with this program. If not, see <http://www.gnu.org/licenses/>.
   -
   -->
+<docs>
+
+# Usage
+
+### Simple element
+
+* With an icon:
+
+```
+<AppNavigationItem title="My title" icon="icon-category-enabled">
+
+```
+* With a spinning loader instead of the icon:
+
+<AppNavigationItem title="Loading Item" :loading="true" />
+
+### Element with actions
+Wrap the children in a template. If you have more than 2 actions, a popover menu and a menu
+button will be automatically created.
+
+```
+<AppNavigationItem title="Item with actions" icon="icon-category-enabled">
+	<template slot="actions">
+		<ActionButton icon="icon-edit" @click="alert('Edit')">
+			Edit
+		</ActionButton>
+		<ActionButton icon="icon-delete" @click="alert('Delete')">
+			Delete
+		</ActionButton>
+		<ActionLink icon="icon-external" title="Link" href="https://nextcloud.com" />
+	</template>
+</AppNavigationItem>
+```
+
+### Element with counter
+Just nest the counter into <AppNavigationItem> and add `slot="counter"` to it.
+
+```
+<AppNavigationItem title="Item with counter" icon="icon-folder">
+	<AppNavigationCounter slot="counter">
+		99+
+	</AppNavigationCounter>
+</AppNavigationItem>
+```
+
+### Element with children
+
+Wrap the children in a template with the `slot` property and use the prop `allowCollapse` to choose wether to allow or
+prevent the user from collapsing the items.
+
+```
+<AppNavigationItem title="Item with children" :allowCollapse="true">
+	<template>
+		<AppNavigationItem title="AppNavigationItemChild1" />
+		<AppNavigationItem title="AppNavigationItemChild2" />
+		<AppNavigationItem title="AppNavigationItemChild3"  />
+		<AppNavigationItem title="AppNavigationItemChild4"  />
+	</template>
+</AppNavigationItem>
+```
+### Editable element
+Add the prop `:editable=true` and an edit placeholder if you need it. By devault
+the placeholder is the previous title of the element.
+
+```
+<AppNavigationItem title="Editable Item" :editable="true"
+	editPlaceholder="your_placeholder_here" />
+```
+### Undo element
+Just set the `undo` and `title` props.
+
+```
+<AppNavigationItem :undo="true" title="Are you sure you want to delete?"  />
+
+```
+### Pinned element
+Just set the `pinned` prop.
+```
+<AppNavigationItem title="Pinned item" :pinned="true" />
+```
+
+</docs>
+
 <template>
-	<!-- Navigation item -->
-	<nav-element :id="item.id" v-bind="navElement(item)"
-		:title="item.title" :class="[{'icon-loading-small': item.loading, 'open': opened, 'collapsible': collapsible }, item.classes]">
-		<!-- Bullet -->
-		<div v-if="item.bullet" :style="{ backgroundColor: item.bullet }" class="app-navigation-entry-bullet" />
+	<!-- Navigation item, can be either an <li> or a <router-link> depending on the props -->
+	<nav-element v-bind="navElement"
+		:title="title"
+		:class="{
+			'app-navigation-entry--no-icon': !isIconShown,
+			'app-navigation-entry--opened': opened,
+			'app-navigation-entry--pinned': pinned,
+			'app-navigation-entry--editing' : editing,
+			'app-navigation-entry--deleted': undo,
+			'app-navigation-entry--collapsible': collapsible
+		}"
+		class="app-navigation-entry">
+		<!-- Icon and title -->
+		<a v-if="!undo && !editing"
+			class="app-navigation-entry-link" href="#"
+			@click="onClick">
 
-		<button v-if="collapsible" class="collapse" @click.prevent.stop="toggleCollapse" />
-
-		<a :class="item.icon" :href="(item.href) ? item.href : '#'"
-			@click="callPreventStop(simpleAction, $event)">
-			<img v-if="item.iconUrl" :alt="item.text" :src="item.iconUrl">
-			{{ item.text }}
+			<!-- icon if not collapsible -->
+			<!-- never show the icon over the collapsible if mobile -->
+			<div :class="{ 'icon-loading-small': loading, [icon]: icon && isIconShown }"
+				class="app-navigation-entry-icon">
+				<slot v-if="!loading" v-show="isIconShown" name="icon" />
+			</div>
+			<span class="app-navigation-entry__title">{{ title }}</span>
 		</a>
 
-		<!-- Popover, counter and button(s) -->
-		<div v-if="item.utils" class="app-navigation-entry-utils">
-			<ul>
-				<!-- counter -->
-				<AppNavigationCounter v-if="item.utils.counter" :highlighted="item.utils.counter_highlighted === true">
-					{{ item.utils.counter }}
-				</AppNavigationCounter>
-
-				<!-- first action if only one action -->
-				<li v-if="item.utils.actions && item.utils.actions.length === 1"
-					class="app-navigation-entry-utils-menu-button">
-					<button :class="item.utils.actions[0].icon" :title="item.utils.actions[0].text" @click="item.utils.actions[0].action" />
-				</li>
-
-				<!-- menu button if at least two actions -->
-				<li v-else-if="item.utils.actions && item.utils.actions.length > 1"
-					class="app-navigation-entry-utils-menu-button">
-					<button v-click-outside="hideMenu" @click="toggleMenu" />
-				</li>
-			</ul>
-		</div>
-
-		<!-- menu if at least two actions -->
-		<div v-if="item.utils && item.utils.actions && item.utils.actions.length > 1"
-			:class="{ 'open': openedMenu }" class="app-navigation-entry-menu">
-			<PopoverMenu :menu="item.utils.actions" />
-		</div>
-
+		<AppNavigationIconCollapsible v-if="collapsible" :open="opened" @click.prevent.stop="toggleCollapse" />
 		<!-- undo entry -->
-		<div v-if="item.undo" class="app-navigation-entry-deleted">
-			<div class="app-navigation-entry-deleted-description">
-				{{ item.undo.text }}
+		<div v-if="undo" class="app-navigation-entry__deleted">
+			<div class="app-navigation-entry__deleted-description">
+				{{ title }}
 			</div>
-			<button :title="t('settings', 'Undo')" class="app-navigation-entry-deleted-button icon-history" />
+		</div>
+
+		<!-- Counter and Actions -->
+		<div v-if="hasUtils" class="app-navigation-entry__utils">
+			<slot name="counter" />
+			<Actions menu-align="right" :open="menuOpen" @update:open="onMenuToggle">
+				<ActionButton v-if="editable && !editing" icon="icon-rename" @click="handleEdit" />
+				<ActionButton v-if="undo" icon="app-navigation-entry__deleted-button icon-history" @click="handleUndo" />
+				<slot name="actions" />
+			</Actions>
 		</div>
 
 		<!-- edit entry -->
-		<div v-if="item.edit" class="app-navigation-entry-edit">
-			<form @submit.prevent.stop="item.edit.action">
-				<input :placeholder="item.edit.text" type="text">
-				<input type="submit" value="" class="icon-confirm">
-				<input type="submit" value="" class="icon-close"
-					@click.stop.prevent="cancelEdit">
+		<div v-if="editing" class="app-navigation-entry__edit">
+			<form v-if="editing" @submit.prevent="handleRename">
+				<input v-model="newTitle" type="text" class="app-navigation-entry__edit-input"
+					:placeholder="editPlaceholder !== '' ? editPlaceholder : title">
+				<button type="submit" class="icon-confirm" />
+				<button type="submit" class="icon-close"
+					@click.stop.prevent="cancelEdit" />
 			</form>
 		</div>
 
-		<!-- if the item has children, inject the component with proper data -->
-		<ul v-if="item.children">
-			<app-navigation-item v-for="(child, key) in item.children" :key="key" :item="child" />
+		<!-- Children elements -->
+		<ul v-if="canHaveChildren && hasChildren" class="app-navigation-entry__children">
+			<slot />
 		</ul>
 	</nav-element>
 </template>
 
 <script>
-import { PopoverMenu } from 'Components/PopoverMenu'
 import ClickOutside from 'vue-click-outside'
-import { AppNavigationCounter } from 'Components/AppNavigationCounter'
+import Actions from '../Actions/Actions'
+import ActionButton from '../ActionButton/ActionButton'
+import AppNavigationIconCollapsible from './AppNavigationIconCollapsible'
+import isMobile from 'Mixins/isMobile'
 
 export default {
 	name: 'AppNavigationItem',
+
 	components: {
-		PopoverMenu,
-		AppNavigationCounter
+		Actions,
+		ActionButton,
+		AppNavigationIconCollapsible
 	},
 	directives: {
 		ClickOutside
 	},
+	mixins: [isMobile],
 	props: {
-		item: {
-			type: Object,
+		/**
+		 * The title of the element.
+		 */
+		title: {
+			type: String,
 			required: true
 		},
+		/**
+		* Refers to the icon on the left, this prop accepts a class
+		* like 'icon-category-enabled'.
+		*/
+		icon: {
+			type: String,
+			default: ''
+		},
+
+		/**
+		* Displays a loading animated icon on the left of the element
+		* instead of the icon.
+		*/
+		loading: {
+			type: Boolean,
+			default: false
+		},
+		/**
+		* Passing in a route will make the root element of this
+		* component a <router-link /> that points to that route.
+		* By leaving this blank, the root element will be a <li>.
+		*/
+		to: {
+			type: [String, Object],
+			default: ''
+		},
+		/**
+		* Pass in `true` if you want the matching behaviour to
+		* be non-inclusive: https://router.vuejs.org/api/#exact
+		*/
+		exact: {
+			type: Boolean,
+			default: false
+		},
+		/**
+		* Gives the possibility to collapse the children elements into the
+		* parent element (true) or expands the cildren elements (false).		*/
+		allowCollapse: {
+			type: Boolean,
+			default: false
+		},
+		/**
+		* Makes the title of the item editable by providing an `ActionButton`
+		* component that toggles a form
+		*/
+		editable: {
+			type: Boolean,
+			default: false
+		},
+		/**
+		* Only for 'editable' items, sets placeholder text for the editing form.
+		*/
+		editPlaceholder: {
+			type: String,
+			default: ''
+		},
+		/**
+		* Pins the item to the bottom left area, above the settings. Do not
+		* place 'non-pinned' `AppnavigationItem` components below `pinned`
+		* ones.
+		*/
+		pinned: {
+			type: Boolean,
+			default: false
+		},
+		/**
+		* Puts the item in the 'undo' state.
+		*/
+		undo: {
+			type: Boolean,
+			default: false
+		},
+		/**
+		 * The navigation collapsible state (synced)
+		 */
 		open: {
 			type: Boolean,
 			default: false
 		},
+		/**
+		 * The actions menu open state (synced)
+		 */
 		menuOpen: {
 			type: Boolean,
 			default: false
 		}
 	},
+
 	data() {
 		return {
+			newTitle: '',
 			opened: this.open,
-			openedMenu: this.menuOpen
+			editing: false
 		}
 	},
 	computed: {
 		collapsible() {
-			return this.item.collapsible && this.item.children && this.item.children.length > 0
+			return this.allowCollapse && !!this.$slots.default
 		},
-		simpleAction() {
-			if (this.collapsible && !this.item.action) {
-				return this.toggleCollapse
+
+		// is the icon shown?
+		// we don't show it on mobile if the entry is collapsible
+		// we show the collapse toggle directly!
+		isIconShown() {
+			return !this.collapsible || (this.collapsible && !this.isMobile)
+		},
+
+		// Checks if the component is already a children of another
+		// instance of AppNavigationItem
+		canHaveChildren() {
+			if (this.$parent.$options._componentTag === 'AppNavigationItem') {
+				return false
 			} else {
-				return this.item.action
+				return true
+			}
+		},
+		hasChildren() {
+			if (this.$slots.default) {
+				return true
+			} else {
+				return false
+			}
+		},
+		hasUtils() {
+			if (this.$slots.actions || this.$slots.counter || this.editable || this.undo) {
+				return true
+			} else {
+				return false
+			}
+		},
+		// This is used to decide which outer element type to use
+		// li or router-link
+		navElement() {
+			if (this.to) {
+				return {
+					is: 'router-link',
+					tag: 'li',
+					to: this.to,
+					exact: this.exactRoute
+				}
+			}
+			return {
+				is: 'li'
 			}
 		}
 	},
 	watch: {
 		open(newVal) {
 			this.opened = newVal
-		},
-		menuOpen(newVal) {
-			this.openedMenu = newVal
 		}
 	},
 	mounted() {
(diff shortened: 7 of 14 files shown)


src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -19,92 +19,71 @@
   - along with this program. If not, see <http://www.gnu.org/licenses/>.
   -
   -->
+<docs>
+
+### Props
```
korelstar: Props should be documented by JSDoc comment blocks in the script part, see https://vue-styleguidist.github.io/docs/Documenting.html#code-comments

src/assets/variables.scss
```diff
@@ -47,3 +47,78 @@ $opacity_full: 1;
...
+	@return darken($color, $value);
+}
+
+@function nc-lighten($color, $value) {
+	@return lighten($color, $value);
+}
+
+// SCSS variables
+// DEPRECATED, please use CSS4 vars
+$color-main-text: #222 !default; // Not #000 for better readability
```
korelstar: This will break theming. I think we should use CSS4 vars, like the comment indicates.

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -121,18 +112,35 @@ export default {
...
+			try {
+				return new URL(this.icon)
+			} catch (error) {
+				return false
+			}
+		},
+		// Checks if the component is already a children of another
+		// instance of AppNavigationItem
+		canHaveChildren() {
+			if (this.$parent.$options._componentTag === 'AppNavigationItem') {
```
korelstar: Does this work if I have a custom component which includes a `AppNavigationItem`?

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -19,92 +19,71 @@
...
 	directives: {
 		ClickOutside
 	},
 	props: {
+		title: {
+			type: String,
+			required: true
+
+		},
 		item: {
```
korelstar: The `item` will be finally removed completely, hmm?

src/components/AppNavigationItem/AppNavigationItem.vue
ma12-co: Sure, will do :)

src/assets/variables.scss
ma12-co: I'll look into it, thanks!

src/components/AppNavigationItem/AppNavigationItem.vue
ma12-co: I'm not sure I understand this question, could you give me an example please?

src/components/AppNavigationItem/AppNavigationItem.vue
ma12-co: That's the idea, yes!

ma12-co:
> This PR doesn't look like it is finished, but the tags and review requests say so. Hence, I did a fast review on the code, but I didn't test it in practice. This has to be done later.

Yep, sorry.. I just created it because I needed a bit of feedback/guidance (which you provided)! Thank you!

src/components/AppNavigationItem/AppNavigationItem.vue
ma12-co: Do you mean:

```
<AppNavigationItem />
    <template slot="children">
        <Custom>
            <AppNavigationItem />
        </Custom>
    </template>
<AppNavigationItem />
```

Because if that's the case it wouldn't work :(

src/assets/variables.scss
juliushaertl: @ma12-co As discussed the file should not be needed in the end but all occurences of the SCSS variables should be replaced with the corresponding CSS variables.

src/assets/functions.scss
```diff
@@ -0,0 +1,77 @@
...
+	.icon-#{$icon} {
+		@include icon-color($icon, $dir, $color-black, $version, $core);
+	}
+	.icon-#{$icon}-white,
+	.icon-#{$icon}.icon-white {
+		@include icon-color($icon, $dir, $color-white, $version, $core);
+	}
+}
+
+@mixin position($value) {
```
juliushaertl: Except for the position mixin, the rest is depending on the servers SCSS transpiler.

src/assets/variables.scss
ma12-co: right!

src/components/AppNavigationItem/AppNavigationItem.vue
ma12-co: ..Or we could keep it just for the navElement that you implemented..

src/components/AppNavigationItem/AppNavigationItem.vue
korelstar: What I meant was a custom component like this: https://github.com/nextcloud/notes/blob/fc7631eb9e7e89876f2570d18b872b89fb30832c/src/components/NavigationNoteItem.vue

But now, I think there is no problem with the scenario I had in mind.

However, I'm not sure, if this check is really needed. It's there in order to disallow a hierarchy with more than two levels, right?

ma12-co: In that case I think it'll work because the root template should not count.
Yes, it prevents `<AppNavigationItem />` children to have their own children by disabling the slot. We could also throw an error if somebody tries to go beyond those 2 levels...

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -113,6 +118,52 @@ export default {
...
+       	*/
+		isLoading: {
+			type: Boolean,
+			default: false
+		},
+		/**
+		* Gives the possibility to pass in an array of additional classes
+		* that will be applied to the root element.
+       	*/
+		classes: {
```
skjnldsv: The classes prop doesn't make sense, you can directly add classes on the component iself :)

src/components/AppNavigationItem/AppNavigationItem.vue
skjnldsv: > However, I'm not sure, if this check is really needed. It's there in order to disallow a hierarchy with more than two levels, right?

yes :)
We can still adapt this later on if this check doesn't fit your needs ;)

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -123,14 +174,36 @@ export default {
...
+			if (this.$parent.$options._componentTag === 'AppNavigationItem') {
+				return false
+			} else return true
+		},
+		// Checks if the component is inside a 'pinned' wrapper
+		// (used to toggle the class 'pinned' in the root element)
+		isPinned() {
+			if (this.$parent.$options._componentTag === 'AppNavigationPinnedItems') {
+				return true
+			} else return false
```
skjnldsv: Don't mix multiple if-else syntax styles ;)

```suggestion
			} else {
				return false
			}
```

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -172,20 +245,15 @@ export default {
...
-				}
-
+		navElement() {
+			if (this.route) {
 				return {
 					is: 'router-link',
 					tag: 'li',
-					to: item.router,
-					exact
+					to: this.route,
```
skjnldsv: I would rather have two AppNavigationItem with one mixin for things that are common to both.
AppNavigationItem and AppNavigationItemRouter ? What do you think?
That way this is cleaner?

src/components/AppNavigationPinnedItems/AppNavigationPinnedItems.vue
```diff
@@ -0,0 +1,64 @@
...
+<docs>
+
+#### Usage
+Wrap `<AppNavigationItem>` components with this `<AppNavigationPinnedItems />`
+component in order to pin them to the bottom left above the settings.
+
+### Example
+
+```
+<AppNavigationPinnedItems>
```
skjnldsv: Let's remove this then :)

src/components/AppNavigationItem/AppNavigationItem.vue
skjnldsv: Or I actually am wondering if using route or exact on the `AppNavigationItem` works?
If so it's easier to have directly a check for the route or exact prop and apply it there?

That way we could directly do `<AppNavigationItem route="contacts" :exact="true" />`

src/components/AppNavigationItem/AppNavigationItem.vue
ma12-co: Ok!

src/components/AppNavigationItem/AppNavigationItem.vue
ma12-co: done ;)

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -19,49 +19,22 @@
...
-	<nav-element :id="item.id" v-bind="navElement(item)"
-		:title="item.title" :class="[{'icon-loading-small': item.loading, 'open': opened, 'collapsible': collapsible }, item.classes]">
+	<!-- Navigation item, can be either an <li> or a <router-link> depending on the props -->
+	<nav-element v-bind="navElement()"
+		:title="title"
+		:class="[{'icon-loading-small': loading, 'open': opened, 'collapsible': collapsible }]"
+		@click.prevent.stop="">
 		<!-- Bullet -->
-		<div v-if="item.bullet" :style="{ backgroundColor: item.bullet }" class="app-navigation-entry-bullet" />
+		<div v-if="bullet !== ''" :style="{ backgroundColor: bullet }" class="app-navigation-entry-bullet" />
```
skjnldsv: It's better to have computeds for that, much cleaner
`haveBullet`

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -71,40 +44,74 @@
...
 					@click.stop.prevent="cancelEdit">
 			</form>
 		</div>
 
-		<!-- if the item has children, inject the component with proper data -->
-		<ul v-if="item.children">
-			<app-navigation-item v-for="(child, key) in item.children" :key="key" :item="child" />
+		<!-- Icon or image from url -->
+		<a :class="[ !isIconUrl && bullet === '' ? icon : '' ]" href="#"
+			@click="handleClick">
```
skjnldsv:
```suggestion
			@click="onClick">
```

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -158,40 +234,60 @@ export default {
...
-					exact
+					to: this.route,
+					exact: this.exactRoute
 				}
 			}
 			return {
 				is: 'li'
 			}
+		},
+		handleClick() {
```
skjnldsv:
```suggestion
		onClick(event) {
```

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -158,40 +234,60 @@ export default {
...
+			this.$emit('click')
```
skjnldsv:
```suggestion
			this.$emit('click', event)
```

src/components/AppNavigationItem/AppNavigationItem.vue
skjnldsv: Yes, we can do that.
have a to props and use it there, it will be better.
```suggestion
					to: this.to,
```

src/components/AppNavigationItem/AppNavigationItem.vue
korelstar: `v-if="bullet"` should be fine, too, I think.

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -201,6 +201,14 @@ export default {
 				return true
 			}
 		},
+		hasBullet() {
+			if (this.bullet) {
```
skjnldsv: You could just do a `return this.bullet !== '` :)

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -71,40 +44,74 @@
...
+			type: String,
+			required: true
+		},
+		/**
+		* Passing in a route will make the root element of this
+		* component a <router-link /> that points to that route.
+		* By leaving this blank, the root element will be a <li>.
+       	*/
+		to: {
+			type: String,
```
skjnldsv: Can also be an Object https://github.com/vuejs/vue-router/blob/edfa2c6a007c0ad5f9028d29141f60844d72f2f4/src/components/link.js#L8

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -71,40 +44,74 @@
...
+			default: ''
+		},
+		/**
+		* Pass in `false` if you want the matching behaviour to
+		* be inclusive: https://router.vuejs.org/api/#exact
+       	*/
+		exactRoute: {
```
skjnldsv:
```suggestion
		exact: {
```

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -71,40 +44,74 @@
...
+			type: Boolean,
+			default: true
```
skjnldsv:
```suggestion
			default: false
```
https://router.vuejs.org/api/#exact

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -113,6 +120,51 @@ export default {
...
+		},
+		/**
+		* Passing in a HEX color string in this prop will hide any
+		* icon or image and display a bullet with the selected
+		* colors instead.
+       	*/
+		bullet: {
+			type: String,
+			default: ''
```
skjnldsv: This could use a validator function :)
```suggestion
			default: '',
			validator: value => {
				return /^([a-f0-9]{3}){1,2}$/i.test(value)
			}
```

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -19,49 +19,22 @@
...
+<docs>
+###	Example usage
+</docs>
+
 <template>
-	<!-- Navigation item -->
-	<nav-element :id="item.id" v-bind="navElement(item)"
-		:title="item.title" :class="[{'icon-loading-small': item.loading, 'open': opened, 'collapsible': collapsible }, item.classes]">
+	<!-- Navigation item, can be either an <li> or a <router-link> depending on the props -->
+	<nav-element v-bind="navElement()"
```
skjnldsv:
```suggestion
	<nav-element v-bind="navElement"
```

skjnldsv: And move navElement to a computed :)

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -123,14 +178,57 @@ export default {
...
+		},
+		// This is used to decide which outer element type to use
+		// li or router-link
+		navElement() {
+			if (this.to) {
+				return {
+					is: 'router-link',
+					tag: 'li',
+					to: this.to,
+					exact: this.exactRoute
```
skjnldsv:
```suggestion
					exact: this.exact
```

src/components/AppNavigation/AppNavigation.vue
```diff
@@ -27,7 +27,26 @@
...
+
+	mounted: function() {
+		this.markFirstPinned()
+	},
+
+	updated: function() {
+		this.markFirstPinned()
+		// Removes the `first-pinned` class to the 'previous' first pinned
+		// element.
+		document.querySelectorAll('#app-navigation > ul > li.first-pinned')[1].classList.remove('first-pinned')
```
skjnldsv: will trigger an error if it doesn't exists

skjnldsv: Also won't work if for god knows why, you have more than 2 first-pinned :)

src/components/AppNavigation/AppNavigation.vue
```diff
@@ -27,7 +27,26 @@
...
+	},
+
+	methods: {
+		// Finds the first element with the class `pinned` and assigns the
+		// class first-pinned to id
+		markFirstPinned() {
+			document.querySelector('#app-navigation > ul > li.pinned').classList.add('first-pinned')
```
skjnldsv: will also trigger an error if it doesn't exists

src/components/AppNavigation/AppNavigation.vue
```diff
@@ -35,16 +35,21 @@ export default {
...
 
 	methods: {
 		// Finds the first element with the class `pinned` and assigns the
-		// class first-pinned to id
+		// class first-pinned to it. Then removes the first-pinned class from
+		// any other pinned elements.
 		markFirstPinned() {
-			document.querySelector('#app-navigation > ul > li.pinned').classList.add('first-pinned')
+			let first = document.querySelector('#app-navigation > ul > li.pinned')
+			first.classList.add('first-pinned')
```
skjnldsv: check if `first` exists :)

src/components/AppNavigation/AppNavigation.vue
```diff
@@ -35,16 +35,21 @@ export default {
...
 	},
 
 	methods: {
 		// Finds the first element with the class `pinned` and assigns the
-		// class first-pinned to id
+		// class first-pinned to it. Then removes the first-pinned class from
+		// any other pinned elements.
 		markFirstPinned() {
-			document.querySelector('#app-navigation > ul > li.pinned').classList.add('first-pinned')
+			let first = document.querySelector('#app-navigation > ul > li.pinned')
```
skjnldsv: const

src/components/AppNavigation/AppNavigation.vue
```diff
@@ -35,16 +35,21 @@ export default {
...
+			first.classList.add('first-pinned')
+			let all = document.querySelectorAll('#app-navigation > ul > li.pinned.first-pinned')
```
skjnldsv: const

skjnldsv:
Ready for reviews @nextcloud/vue :pray:

skjnldsv:
@juliushaertl we should probably create new sections on the docs for appnavigation and appsidebar.
How can we do that? :)

juliushaertl:
The component sections are defined here: 

https://github.com/nextcloud/nextcloud-vue/blob/master/styleguide.config.js#L37-L53

src/assets/variables.scss
```diff
@@ -47,3 +47,11 @@ $opacity_full: 1;
 // menu round background hover feedback
 // good looking on dark AND white bg
 $action-background-hover: rgba(127, 127, 127, .25);
+
+// various structure data
```
skjnldsv: ahah, please add proper comments like the one above :stuck_out_tongue_closed_eyes:

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -18,94 +18,265 @@
...
-				<input :placeholder="item.edit.text" type="text">
-				<input type="submit" value="" class="icon-confirm">
-				<input type="submit" value="" class="icon-close"
-					@click.stop.prevent="cancelEdit">
-			</form>
+		<!-- Counter and Actions -->
+		<div v-if="this.$slots.counter || this.$slots.actions || editable || undo" class="app-navigation-entry--utils">
+			<ul>
+				<slot name="counter" />
+				<Actions menu-align="center">
```
skjnldsv:
```suggestion
				<Actions menu-align="right">
```

src/components/AppNavigation/AppNavigation.vue
```diff
@@ -101,16 +102,67 @@ export default {
 			flex-wrap: wrap;
 			order: 1;
 			flex-shrink: 0;
+
+			/* Menu and submenu */
+			> li,
```
skjnldsv: This needs to go into the appnavigationitem

src/components/AppNavigation/AppNavigation.vue
```diff
@@ -101,16 +102,67 @@ export default {
...
-	}
-	&.hidden {
-		display: none;
 	}
 }
 // mobile view only
 @media only screen and (max-width: 768px) {
 	// if navigation is shown
-	.nav-open #app-navigation {
+	.nav-open #app__navigation {
```
skjnldsv: you cannot rename this one, you need to keep the app-navigation

skjnldsv:
@juliushaertl @georgehrke @ChristophWurst 
The app-navigation is conflicting HARD on all the properties because I used the `>` quite a lot and it have too much strength as a css selector.
I thin we can do this like that and push it for 17+

Capture d’écran_2019-07-29_16-03-51
Capture d’écran_2019-07-29_16-04-00

Thoughts?

src/components/AppNavigation/AppNavigation.vue
```diff
@@ -2,6 +2,7 @@
  - @copyright Copyright (c) 2019 Christoph Wurst <christoph@winzerhof-wurst.at>
  -
  - @author Christoph Wurst <christoph@winzerhof-wurst.at>
+ - @author Marco Ambrosini <marcoambrosini@pm.me>
```
skjnldsv: :+1:

src/components/AppNavigation/AppNavigation.vue
```diff
@@ -84,7 +85,20 @@ export default {
...
+	height: 100%;
+	width: inherit;
+	overflow-x: hidden;
+	overflow-y: auto;
+	box-sizing: border-box;
+	display: flex;
+	flex-direction: column;
+	// Gives some room to the actions and popover
+	// menus
+	padding-right: 6px;
```
skjnldsv: I think this needs to be fixed in the actions directly :)
https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/Actions/Actions.vue#L267

Maybe we should only adjust its position on align="center" :)

src/components/AppNavigationCounter/AppNavigationCounter.vue
```diff
@@ -25,26 +25,23 @@
 ### Normal Counter
 
 ```
-<AppNavigationCounter :highlighted="false">
-your number or string
-</AppNavigationCounter>
+<AppNavigationCounter">314+</AppNavigationCounter">
```
skjnldsv:
```suggestion
<AppNavigationCounter>314+</AppNavigationCounter>
```

src/components/AppNavigation/AppNavigation.vue
skjnldsv: Ah no, it actually help things. Without it it would be broken :)
https://github.com/nextcloud/nextcloud-vue/pull/474 works if the menu is not blocked by another div. The margin 10px is atually causing a bad shifting. We should maybe revert it to 0 and only add a margin-bottom of 10px.

ma12-co:
**collapsible navigation item:** @skjnldsv and me were talking about removing the icon on top of the triangle (as it appears [here](https://docs.nextcloud.com/server/16/developer_manual/_images/navigation-collapsible.gif)) to show that the item is collapsible without the need of  hovering on it. The problem is, once we remove it, the only viable option we've found to show the user that the triangle is clickable, is changing its color on hover. This is a very subtle effect compared to what we have right now, and it may not be enough.. so @jancborchardt, we need your take on this, what do you think? 

Screenshot_20190806_165942

jancborchardt:
@ma12-co @skjnldsv regarding the collapsible navigation item, it’s a bit confusing as you don’t know what is clickable with which effect. As we mostly have the actions on the right with the menu, for the case of a parent being both selectable and expandable (like in the News app) the opening triangle should be on the right instead of the left. And the left side should keep the icon.

What do you think?

skjnldsv:
> the opening triangle should be on the right instead of the left.

That's an interesting approach!
How would we handle too many items on the right? (action, counter, collapsible)
I think we should not put it in the actions menu as it will be very tedious to discover and to trigger. It should stay a one-click item :)
(I think that is not what you had in mind, but I wanted to clarify in case someone thought about this too :wink: )

> it’s a bit confusing as you don’t know what is clickable with which effect.

That is why we need you actually, currently there is not really great feedback on it. We tried with a grey circle like the actions, but it looks very weird.
We could check how other platforms do it?

ma12-co:
Now that I'm imagining this item with all sorts of things (counter, actions, edit, undo at once), actually I think that the current solution (folder on top) might be the best of the bunch..

jancborchardt:
> I think we should not put it in the actions menu as it will be very tedious to discover and to trigger. It should stay a one-click item :)

Yes, it should not be in a menu for sure.




> Now that I'm imagining this item with all sorts of things (counter, actions, edit, undo at once), actually I think that the current solution (folder on top and triangle on the right) might be the best of the bunch..

What do you mean by "folder on top and triangle on the right" as being the current solution? :)

ma12-co:
Sorry, I meant on the left ;)

jancborchardt:
@ma12-co ok, so folder on the left, and triangle on the right? But that’s not the current solution, but what I described in the comment at https://github.com/nextcloud/nextcloud-vue/pull/486#issuecomment-518743891 ? :)

(Or am I confused now?)

skjnldsv:
> (Or am I confused now?)

I'll join the club xD

ma12-co:
What I wanted to say is that the [current implementation](https://docs.nextcloud.com/server/16/developer_manual/_images/navigation-collapsible.gif) looks the most appropriate to me :)

jancborchardt:
> What I wanted to say is that the [current implementation](https://docs.nextcloud.com/server/16/developer_manual/_images/navigation-collapsible.gif) looks the most appropriate to me :)

Ok gotcha. The problem though is that this means clicking on the entry itself will _expand_ the subentries, not open the content?

Because that is not how it works in e.g. Files and News currently – there clicking on the entry itself does open the content (as expected) and only clicking on the triangle expands the subentries.

skjnldsv:
> Ok gotcha. The problem though is that this means clicking on the entry itself will _expand_ the subentries, not open the content?

currently you can have both, either the entry is clickeable/focusable and clicking just open the link, or it have nothing but text and expand the collapsible on click :)

ma12-co:
> currently you can have both, either the entry is clickeable/focusable and clicking just open the link, or it have nothing but text and expand the collapsible on click :)

This sounds good, we also have the option of both routing to the new page **and**  expanding the children at the same time with one click.. Wouldn't this work well in News?

jancborchardt:
> we also have the option of both routing to the new page and expanding the children at the same time with one click

I’d say this is unexpected behavior though. At least confusing, at worst annoying.

We should have this rather consistent, and the way it works in Files and News right now is pretty good, no?

ma12-co:
Right, so just to be clear, [this](https://cloud.ma12.co/index.php/s/KMM79Z7tgpHgbXj) is the conclusion? Sorry guys.. I'm slowly catching up :)

skjnldsv:
Yes, this is how we currently have it.

Can we focus and take a decision on the focus feedback for the triangle?
So we can move on? I feel like we lost our productivity on this ;)

skjnldsv:
Okay, let's recap once and for all, because we lost track of the initial goal.
Issues:
1. mobile don't know the entry is collapsible as we show the triangle on focus/hover, which the mobile do NOT have
2. Current focus feedback is just showing the triangle, this is not enough
3. Current triangle is too small, contrast is not good enough

Solutions for (1):
1. Move the triangle to the right
	- pros:
		- We can show the icon AND the triangle
	- cons:
		- Possibly not enough room if we have a counter and a menu
2. Always show the triangle
	- pros: 
		- simple
		- solves 2 too
	- cons:
		- loose the option to have an icon for this entry
3. Something else?

For solving 2&3, please advise some ideas! :bowing_woman:

ma12-co:
> For solving 2&3, please advise some ideas! bowing_woman

What do you think of this?

IMG_20190809_112828

skjnldsv:
I like it!
Though this rises a few issues:
1. we have a 44px width available for the icon (16px) and the triangle, your mockup (very nicely done btw) looks bigger. 
2. if you have lots of rows, the alignment of icons is off (vertical)

ma12-co:
> we have a 44px width available for the icon (16px) and the triangle, your mockup (very nicely done btw) looks bigger.

Thank you and yes, this solution would require extra space indeed

> if you have lots of rows, the alignment of icons is off (vertical)

google solves this by adding the extra padding to the non collapsible items too

okOUhQyCsMEIOzJEtqHmcOWP

jancborchardt:
Great explorations! :)

It’s a good mockup @ma12-co, I do share @skjnldsv’s concern about the width though. Note that Google doesn’t have any actions on the right side of the entry, so they have fewer issues with width here.

I’d say a good middle way would be to do it like we do right now, with the following exception to fix it on mobile:
- When an expandable entry is hovered/focused/active, the triangle icon shows instead of the folder icon. And in that case also clicking on that icon will expand/collapse it.

This makes sure that we don’t introduce something completely different just because we switch the underlying tech, but only fix the 1 issue at hand. What do you think?

ma12-co:
Thanks @jancborchardt :) 

> I’d say a good middle way would be to do it like we do right now, with the following exception to fix it on mobile:
> 
>     When an expandable entry is hovered/focused/active, the triangle icon shows instead of the folder icon. And in that case also clicking on that icon will expand/collapse it.

I don't understand, how would you hover/focus on it on mobile to reveal the collapse button?

jancborchardt:
> I don't understand, how would you hover/focus on it on mobile to reveal the collapse button?

On mobile of course it would only be on active, that is, when you opened the entry before, which might then mean that you want to do deeper filtering, which you can then do.

skjnldsv:
So
- Mobile:
	- no icon
	- only triangle
- Desktop
	- no-focus: icon shown, no triangle
	- focus: no icon, triangle shown

Is that what you mean @jancborchardt ?
If so, I would maybe suggest we completely drop the icon for desktop too.
If it's a collapsible, only show the triangle. Simple, clear and efficient :)

jancborchardt:
No, but like this:

- Desktop
	- not hover nor focus nor active: 📁 icon shown, no triangle
	- hover, focus or active: no icon, 🔻 triangle shown
- Mobile
	- not active: 📁 icon shown, no triangle
	- active: no icon, 🔻 triangle shown

skjnldsv:
> - Desktop
>   - not active: :file_folder: icon shown, no triangle

We don't want to do that :)
Mobile doens't have a focus and no one will know the icon hide a collapsible entry :)

jancborchardt:
> > * **Desktop**
> >   * not active: file_folder icon shown, no triangle
> 
> We don't want to do that :)
> Mobile doens't have a focus and no one will know the icon hide a collapsible entry :)

You quoted the part of desktop though (emphasis mine)? :D

ma12-co:
@jancborchardt  @skjnldsv : Call Monday Aug 12 9AM CET?

skjnldsv:
> You quoted the part of desktop though (emphasis mine)? :D

Ah, my bad. Though :active also doesn't exists in mobile either (I mean, it require a user touch, so it's not visible as is), so it's the same issue.

ma12-co:
Time for review @nextcloud/vue  :)

jancborchardt:
> Ah, my bad. Though :active also doesn't exists in mobile either (I mean, it require a user touch, so it's not visible as is), so it's the same issue.

I’m not talking about `:active`, but about the 1 entry which is selected / open / current at the moment. That is, more the class `.active`.

ma12-co:
(please use [this App.vue](https://gist.github.com/ma12-co/ec27beec385b016558073296e904cb24) in you vueexample app for testing, as the new components won't work with the old one!)

skjnldsv:
@jancborchardt let's implement this as the same behaviour as it's currently.
And see later how we can improve :)

src/components/AppNavigationItem/AppNavigationIconCollapsible.vue
```diff
@@ -0,0 +1,88 @@
...
+	background-color: transparent;
+	box-shadow: none;
+	font-size: 18px;
+
+	@include iconfont('triangle-s');
+
+	&:hover{
+		color: var(--color-primary);
+	}
+	&.rotated {
```
skjnldsv:
```suggestion
	&--rotated {
```

src/components/AppNavigationItem/AppNavigationIconCollapsible.vue
```diff
@@ -0,0 +1,88 @@
...
+ - You should have received a copy of the GNU Affero General Public License
+ - along with this program. If not, see <http://www.gnu.org/licenses/>.
+ -
+ -->
+
+<template>
+	<!-- Button to expand or collapse children -->
+	<button
+		class="icon-collapse"
+		:class="{'rotated':open}" @click="onClick" />
```
skjnldsv:
```suggestion
		:class="{ 'icon-collapse--rotated': open }" @click="onClick" />
```

src/components/AppNavigationIconBullet/AppNavigationIconBullet.vue
```diff
@@ -0,0 +1,95 @@
...
+### AppNavigationItem example
+```
+<AppNavigationItem  title="Entry 2" :pinned="true">
+	<AppNavigationIconBullet slot="icon" color="0082c9" />
+</AppNavigationItem>
+```
+</docs>
+
+<template>
+	<div class="app-navigation-entry__icon--bullet" @click="onClick">
```
skjnldsv:
```suggestion
	<div class="app-navigation-entry__icon-bullet" @click="onClick">
```
This is not a state :)

src/components/AppNavigationIconBullet/AppNavigationIconBullet.vue
```diff
@@ -0,0 +1,95 @@
...
+		onClick(e) {
+			this.$emit('click', e)
+		}
+	}
+
+}
+</script>
+
+<style lang="scss" scoped>
+.app-navigation-entry__icon--bullet {
```
skjnldsv:
```suggestion
.app-navigation-entry__icon-bullet {
```

src/components/AppNavigationCaption/AppNavigationCaption.vue
```diff
@@ -30,7 +31,7 @@ export default {
 	pointer-events: none;
 }
 
-.app-navigation-caption--item:not(:first-child) {
+.app-navigation-caption:not(:first-child) {
```
skjnldsv:
```suggestion
// extra top space if it's not the first item on the list
.app-navigation-caption:not(:first-child) {
```

jancborchardt:
> @jancborchardt let's implement this as the same behaviour as it's currently.
> And see later how we can improve :)

My only proposal to improve is for mobile:
For an entry/folder which has subitems and is currently selected/active (that is, its contents are shown in the `#app-content`): Do not show the folder icon, but instead the 🔻 triangle to indicate expandability/collapsibility.

skjnldsv:
> There is a design issue on Nextcloud 15.0.11

@korelstar This is nextcloud 17 only
I requested backport for 16
https://github.com/nextcloud/server/pull/16661

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -19,93 +20,253 @@
...
+```
+
+### Element with children
+
+Wrap the children in a template with the `slot` property and use the prop `allowCollapse` to choose wether to allow or
+prevent the user from collapsing the items.
+
+```
+<AppNavigationItem title="Item with children" :allowCollapse="true">
+	<template slot="children">
```
korelstar: This doesn't work. I think it must be `<template>`.

ma12-co: > This doesn't work. I think it must be <template>.

I don't get it, what do you mean exactly??

skjnldsv: Right, we remove the scoping of the children scope!

korelstar:
> @korelstar This is nextcloud 17 only
> I requested backport for 16
> [nextcloud/server#16661](https://github.com/nextcloud/server/pull/16661)

Please do also a backport for NC 15 :-)

skjnldsv:
> Please do also a backport for NC 15 :-)

https://github.com/nextcloud/server/pull/16936
https://github.com/nextcloud/server/pull/16937

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -19,93 +20,253 @@
...
+
+```
+<AppNavigationItem title="Editable Item" :editable="true"
+	editPlaceholder="your_placeholder_here" />
+```
+### Undo element
+Just set the `undo` and `undo-text` props.
+
+```
+<AppNavigationItem :undo="true" undo-text="Are you sure you want to delete?"  />
```
korelstar: As far as I can see, `undo-text` isn't also used anymore. You have to use `title`, hmm?

skjnldsv: Nope, seems like we lost this during a rebase :thinking:

skjnldsv: Yep, my bad @ma12-co https://github.com/nextcloud/nextcloud-vue/commit/1620213d2fed461d72e435de22e6fb8f1883b88a

ma12-co: Yes, I removed it and forgot to update the docs! Thanks :)

ma12-co: >Yep, my bad @ma12-co 

No, I removed it on purpose, I decided to just leave title prop only

skjnldsv: Ah! I thought I did a mistake! :+1:

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -158,40 +361,221 @@ export default {
...
+.app-navigation-entry {
+	position: relative;
+	min-height: $clickable-area;
+	width: 100%;
+	box-sizing: border-box;
+	display: flex;
+	flex-wrap: wrap;
+	padding-right: 6px;
+	order: 1;
+	flex-shrink: 0;
```
korelstar: These two CSS properties (`order` and `flex-shrink`) have to be applied to `AppNavigationCaption`, too. Otherwise, captions are not shown anymore!

korelstar:
Found another issue: I want to use an `AppNavigationItem` without icon but with a very long title (see example below). Issue: The text moves to the left and the right margin misses the three-dots abbreviation.

```
<AppNavigationItem  title="This is bullet item with a very text over more than one line" />
```

Screenshot

korelstar:
And one wish: Could you please add props to `AppNavigationCaption` for an icon and a click-event?

Before, I was able to use `AppNavigationItem` with the style of `AppNavigationCaption` in order to realize this. But now, due to the scoped styles, this is not possible anymore.

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -19,93 +20,253 @@
...
+			<div class="app-navigation-entry__deleted-description">
+				{{ title }}
 			</div>
-			<button :title="t('settings', 'Undo')" class="app-navigation-entry-deleted-button icon-history" />
+		</div>
+
+		<!-- Counter and Actions -->
+		<div v-if="hasUtils" class="app-navigation-entry__utils">
+			<slot name="counter" />
+			<Actions menu-align="right">
```
korelstar: The opened-state isn't synced! Please change this to
```
<Actions menu-align="right" open.sync="openedMenu">
```

korelstar:
One more thing: all actions from an `AppNavigationItem` have a HTML title (popover) with the item's title (not action title!). Seems to be weird. Is this intended?

korelstar:
In general, It appears that building/changing a navigation with a huge amount of items (~200) is very slow (~7 seconds on a slow notebook). This amount is not unusual in the notes app, because it's derived from the notes count and the captions for timeslots.

Maybe, the speed could be improved, e.g. by respecting some aspects from the vue style guide, e.g. https://vuejs.org/v2/style-guide/#Element-selectors-with-scoped-use-with-caution .

skjnldsv:
> One more thing: all actions from an `AppNavigationItem` have a HTML title (popover) with the item's title (not action title!). Seems to be weird. Is this intended?

this is from the actions component :)
Nothing related there then ;)
But yes, if an action is unique there is a tooltip for it.

> Maybe, the speed could be improved, e.g. by respecting some aspects from the vue style guide, e.g. [vuejs.org/v2/style-guide/#Element-selectors-with-scoped-use-with-caution](https://vuejs.org/v2/style-guide/#Element-selectors-with-scoped-use-with-caution) .

That is very interesting!!
Could you maybe profile the perfs a bit to see where we lack?
Indeed scoping is to be used cautiously, but I was always wondering how far that would go! :thinking: 

> And one wish: Could you please add props to `AppNavigationCaption` for an icon and a click-event?

Can't you have a click event on the component directly? Doesn't it work?
Also, current design standards aim to have no icons for a caption, do you have an example to give us? :)

ma12-co:
> In general, It appears that building/changing a navigation with a huge amount of items (~200) is very slow (~7 seconds on a slow notebook).

Can this be for the amount of `computed` logic we have in `aAppNavigationItem`? This code has to execute for each element at first load!

korelstar:
Regarding the speed, I will try to investigate further resp. provide an example.

--------

> > And one wish: Could you please add props to `AppNavigationCaption` for an icon and a click-event?
> 
> Can't you have a click event on the component directly? Doesn't it work?
> Also, current design standards aim to have no icons for a caption, do you have an example to give us? :)

No, `<AppNavigationCaption @click="...">` doesn't work.

I would like to use this in the notes app. Here, when selecting a single category, sub-categories are shown using the caption style, see screenshot below (the screenshot shows the current situation in Notes 3.0.2 using nextcloud-vue 0.12.2). The folder icon clarifies that this caption is about a sub-category.

**screenshot will follow in a few seconds**

korelstar:
### Performance

I did some test runs with [this App.vue](https://gist.github.com/korelstar/8d6fc42b8d69c2c84a70be298f39b451) (please use it within `vueexample` app). That app creates 200 `AppNavigationItem`s, each with three `ActionButton`s. It provides debug logs in the JavaScript console (please activate timestamps in your browser). After loading the app in your Nextcloud, please click the button "Toggle navigation" in order to show the 200 items and have a look at the console with debug logs.

Test results:
Intel Core i3 2120T: ~2,1 seconds
Intel N3160: ~5,4 seconds

Unfortunately, I have not the time to go into more analysis, currently.

skjnldsv:
> I would like to use this in the notes app. Here, when selecting a single category, sub-categories are shown using the caption style, see screenshot below (the screenshot shows the current situation in Notes 3.0.2 using nextcloud-vue 0.12.2). The folder icon clarifies that this caption is about a sub-category.

That's not what are captions for :)
Why don't you use tested items like we support?


> I did some test runs with [this App.vue](https://gist.github.com/korelstar/8d6fc42b8d69c2c84a70be298f39b451) (please use it within `vueexample` app).

I'll try to debug later :)

korelstar:
> That's not what are captions for :)
> Why don't you use tested items like we support?

Hmm. For what are captions then? I thought they exist in order to realize sections inside of the app navigation, i.e. items can be grouped together and complemented with a heading. Did I misunderstood that?

skjnldsv:
> Hmm. For what are captions then? I thought they exist in order to realize sections inside of the app navigation, i.e. items can be grouped together and complemented with a heading. Did I misunderstood that?

It's just titles of a navigation section :)
But your case is nested folders, that's what sub children are for :+1:

korelstar:
> It's just titles of a navigation section :)
> But your case is nested folders, that's what sub children are for 👍

That's not entirely true, since the notes app doesn't show the notes in nested folders. Instead, sub-sub-folders are shown directly in the list as "... / Sub-Folder / Sub-Sub-Folder". This is done in order to reduce complexity, since focus is on notes and not on folders. Hence, I'm not sure, if sub-children are the best way to realize this, because it could lead to misunderstandings. Captions instead, are more intuitive and look nicer ;-)

/cc @jancborchardt

skjnldsv:
> sub-sub-folders are shown directly in the list as "... / Sub-Folder / Sub-Sub-Folder". This is done in order to reduce complexity,

I'd say thats even more complex than before ;)
Also we don't support more than one level of nesting. This is not done randomly :)
Your navigation looks very confusing to me. How do you go up a folder?

Please comply to the design guidelines and only use one level.

korelstar:
> I'd say thats even more complex now (looking at your screenshot) 😕

We are using this for a long time in practice in both the Nextcloud-App and the corresponding Android-App. It's very useful! However, improvement proposals are always welcome :-)

> Also we don't support more than one level of nesting. This was not decided randomly :)

Yes, that's why I don't want to use sub-children, here.

> Your navigation looks very confusing to me. How do you go up a folder?

You don't need to "go up a folder". You select the main category by chosing the item "Categories". Then, you see all notes inside of that category. With this, you have easy access to all notes in that category. In order to make the access just more easier, notes are grouped by sub-category.

> Please comply to the design guidelines and only use one level.

I can't see where we don't comply to the design guidelines. We ARE using one level only.

I suggest that you install the notes app by yourself and write some notes in different categories and give it a try. :-)

skjnldsv:
> I can't see where we don't comply to the design guidelines. We ARE using one level only.

Yeaaah, I'd say that a bit cheating there!
Let's see this in another ticket, because we're just polluting this.
Same for the performances, it's related to the actions component, not this and I have already a fix for it.

korelstar:
> Let's see this in another ticket, because we're just polluting this.

Good idea. Thanks for your time, I will answer on the new ticket tomorrow (no more time for today): https://github.com/nextcloud/notes/issues/364

> Same for the performances, it's related to the actions component, not this and I have already a fix for it.

I'm not sure, if the actions component is the only problem. #569 is a good start, but that can't be all.

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -19,93 +20,253 @@
...
+			<div class="app-navigation-entry__deleted-description">
+				{{ title }}
 			</div>
-			<button :title="t('settings', 'Undo')" class="app-navigation-entry-deleted-button icon-history" />
+		</div>
+
+		<!-- Counter and Actions -->
+		<div v-if="hasUtils" class="app-navigation-entry__utils">
+			<slot name="counter" />
+			<Actions menu-align="right" open.sync="openedMenu">
```
skjnldsv:
```suggestion
			<Actions menu-align="right" :open.sync="openedMenu">
```

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -19,93 +20,253 @@
...
+			<div class="app-navigation-entry__deleted-description">
+				{{ title }}
 			</div>
-			<button :title="t('settings', 'Undo')" class="app-navigation-entry-deleted-button icon-history" />
+		</div>
+
+		<!-- Counter and Actions -->
+		<div v-if="hasUtils" class="app-navigation-entry__utils">
+			<slot name="counter" />
+			<Actions menu-align="right" :open.sync="openedMenu">
```
korelstar: Sorry, my fault: `openedMenu` isn't passed automatically to the parent component. We have to use `menuOpen` since there is a watcher on it.
```suggestion
			<Actions menu-align="right" :open.sync="menuOpen">
```

src/components/AppNavigationCaption/AppNavigationCaption.vue
```diff
@@ -1,6 +1,7 @@
 <template>
-	<li class="app-navigation-caption--item">
-		{{ text }}
+	<li class="app-navigation-caption"
+		@click="onClick">
```
korelstar: This doesn't work for me. Do we need a `<a>` tag?

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -158,40 +361,222 @@ export default {
...
+			height: $clickable-area;
+			display: flex;
+			flex: 0 0 $clickable-area;
+			justify-content: center;
+			align-items: center;
+		}
+	}
+
+	/* Second level nesting for lists */
+	.app-navigation-entry__children {
```
korelstar: This is the `<ul>` which contains flex elements. Therefore, we need to add
```suggestion
	.app-navigation-entry__children {
		display: flex;
		flex-direction: column;
```

Otherwise, there is a gap of 6px between the children items.

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -158,40 +361,222 @@ export default {
...
+<style lang="scss" scoped>
+@import '~Fonts/scss/iconfont-vue';
+
+.app-navigation-entry {
+	position: relative;
+	min-height: $clickable-area;
+	width: 100%;
+	box-sizing: border-box;
+	display: flex;
+	flex-wrap: wrap;
```
korelstar: Not needed. We don't want the item to be wrapped, hm?
```suggestion
```

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -158,40 +361,222 @@ export default {
...
+
+	/* Second level nesting for lists */
+	.app-navigation-entry__children {
+		flex: 0 1 auto;
+		width: 100%;
+		position: relative;
+
+		.app-navigation-entry {
+			display: inline-flex;
+			flex-wrap: wrap;
```
korelstar: Not needed. We don't want the item to be wrapped, hm?
```suggestion
```

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -19,93 +20,253 @@
...
+		*/
+		icon: {
+			type: String,
+			default: ''
+		},
+
+		/**
+		* Displays a loading animated icon on the left of the element
+		* instead of the icon.
+       	*/
```
korelstar:
```suggestion
		*/
```

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -19,93 +20,253 @@
...
+		* By leaving this blank, the root element will be a <li>.
+		*/
+		to: {
+			type: [String, Object],
+			default: ''
+		},
+		/**
+		* Pass in `true` if you want the matching behaviour to
+		* be non-inclusive: https://router.vuejs.org/api/#exact
+       	*/
```
korelstar:
```suggestion
		*/
```

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -19,93 +20,253 @@
...
+		exact: {
+			type: Boolean,
+			default: false
+		},
+		/**
+       	* Gives the possibility to collapse the children elements into the
```
korelstar:
```suggestion
		* Gives the possibility to collapse the children elements into the
```

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -19,93 +20,253 @@
...
+		* parent element (true) or expands the cildren elements (false).		*/
+		allowCollapse: {
+			type: Boolean,
+			default: false
+		},
+		/**
+		* Makes the title of the item editable by providing an `ActionButton`
+		* component that toggles a form
+       	*/
```
korelstar:
```suggestion
		*/
```

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -19,93 +20,253 @@
...
+		editable: {
+			type: Boolean,
+			default: false
+		},
+		/**
+		* Only for 'editable' items, sets placeholder text for the editing form.
+       	*/
```
korelstar:
```suggestion
		*/
```

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -19,93 +20,253 @@
...
+		editPlaceholder: {
+			type: String,
+			default: ''
+		},
+		/**
+		* Pins the item to the bottom left area, above the settings. Do not
+		* place 'non-pinned' `AppnavigationItem` components below `pinned`
+		* ones.
+       	*/
```
korelstar:
```suggestion
		*/
```

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -19,93 +20,253 @@
...
+		pinned: {
+			type: Boolean,
+			default: false
+		},
+		/**
+		* Puts the item in the 'undo' state.
+       	*/
```
korelstar:
```suggestion
		*/
```

src/components/AppNavigationCaption/AppNavigationCaption.vue
skjnldsv: Wait a second, `click.native` doesn't work from the parent?
According to the vuejs docs we should not need to emit it back like that on such simple component :(

src/components/AppNavigationItem/AppNavigationItem.vue
skjnldsv: Yes we do, because the children element is wrapped under the main line.
```
| icon - text  - counter - menu |
|           childrens           |
```

src/components/AppNavigationItem/AppNavigationItem.vue
skjnldsv: Should not be needed as the `app-navigation-entry` already have flex wrap

src/components/AppNavigationCaption/AppNavigationCaption.vue
ma12-co: > This doesn't work for me. Do we need a <a> tag?

I've been trying to do this yesterday and I haven't found  a way to make it work.. Tried the `<a>` as well..

skjnldsv: Fixed!
We indeed don't need to forward the click with an internal event, and if we leave the css `pointer-events: none` we'll definitely won't receive anything :grin: :see_no_evil:

src/components/AppNavigationItem/AppNavigationItem.vue
skjnldsv: yes, watcher and stuff are not used anymore.
Actions already manage its own internal state.

Though what you suggests is not valid in vue. You cannot edit a prop directly. You need to trigger an event to update it :)

ma12-co:
Thanks @korelstar for yr thorough review!

skjnldsv:
@korelstar all clear and rebased!
I reviewed again! All looks good to me :)

I'd say let's merge and individually fix the tiny issues we could find + do a proper aria pass on it!

skjnldsv:
> Good thing! Last remaining drawback is the slow performance compared to the situation before this PR. I strongly suggest to fix this before the next release.

Could you perform a lighthouse? Or a performance audit?

skjnldsv:
default brwoser tool. Either lighthouse or the profiling tab on chroomium or performance on firefox :)

korelstar:
I've created a new issue for the performance problem: #577

PR #499, opened 2019-07-23
Allow usage of custom-label prop

juliushaertl:


skjnldsv:
Hum, we already forward all the attributes: `v-bind="$attrs"` :thinking:

juliushaertl:
Hm, strange, let me check why that didn't work for me then :confused:

juliushaertl:
Nevermind, works fine if I also add a custom options slot content.

PR #502, opened 2019-07-26
Focus proper action on mouse move instead of removing focus

skjnldsv:
Peek 26-07-2019 11-18

Before we removed the focus to make sure we did not conflicted between the mouse and the keyboard navigation. But if you had a focus on an input and moving your mouse, you were losing the focus and had to click again, this is not good :)

Now we focus the nearest item and ignore if it is the currently focused item already :wink:


jancborchardt:
@skjnldsv nitpick: `Password protect`, second word lowercase. :)

skjnldsv:
@jancborchardt thanks :stuck_out_tongue_closed_eyes:

PR #510, opened 2019-07-29
Set defaultHtml setting of v-tooltip to false

juliushaertl:
Signed-off-by: Julius Härtl <jus@bitgrid.net>


juliushaertl:
I did a grep though the apps I've checked out and didn't find any, but of course there might be missing some.

skjnldsv:
Let's merge?

PR #513, opened 2019-08-02
Add docs to multiselect

skjnldsv:

### DIFF
--- a/src/components/Multiselect/Multiselect.vue
+++ b/src/components/Multiselect/Multiselect.vue
@@ -19,6 +19,88 @@
   - along with this program. If not, see <http://www.gnu.org/licenses/>.
   -
   -->
+<docs>
+## Multiselect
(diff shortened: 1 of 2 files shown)


src/components/Multiselect/Multiselect.vue
```diff
@@ -19,6 +19,88 @@
...
+		:user-select="true" />
+</template>
+
+<script>
+import Multiselect from '../index'
+export default {
+	data() {
+		return {
+			value: { id:2,label:'user 1' },
+			options: ['admin', 'user1', 'user2', 'guest', 'group1']
```
juliushaertl: Minor issue, when selecting an option in the demo then, it doesn't get shown in the field as the value doesn't have a label property.

skjnldsv: Thanks :)

PR #515, opened 2019-08-02
Add prop to disable menu on avatar & disable it on multiselect user-select view

skjnldsv:


src/components/Avatar/Avatar.vue
```diff
@@ -136,6 +136,13 @@ export default {
 			type: Boolean,
 			default: false
 		},
+		/**
+		 * Disable the menu
```
juliushaertl: Indentation looks odd :wink:

skjnldsv: Does it? it's ok here :thinking: 
Capture d’écran_2019-08-02_12-22-37

juliushaertl: Ah sorry for the noise, refined github was rendering that wrong

PR #519, opened 2019-08-02
Automatically bold data from search in user-select template

skjnldsv:
Protected against xss 
Capture d’écran_2019-08-02_17-04-42

### DIFF
--- a/src/components/Multiselect/AvatarSelectOption.vue
+++ b/src/components/Multiselect/AvatarSelectOption.vue
@@ -62,24 +62,23 @@ export default {
 </docs>
 
 <template>
-	<span class="option" v-bind="dataBind">
+	<span class="option">
 		<Avatar :display-name="displayName" :user="user"
 			:is-no-user="isNoUser"
 			:disable-menu="true" :disable-tooltip="true"
 			class="option__avatar" />
 		<div class="option__desc">
-			<span class="option__desc--lineone">
-				{{ displayName }}
-			</span>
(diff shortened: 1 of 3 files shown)


src/components/Multiselect/AvatarSelectOption.vue
```diff
@@ -130,12 +129,26 @@ export default {
...
 	},
 
 	methods: {
-		toKebabCase: (string) => {
+		toKebabCase(string) {
 			return string.replace(/([a-z0-9]|(?=[A-Z]))([A-Z])/g, '$1-$2').toLowerCase()
+		},
+		highlightPhrase(text) {
+			if (!this.search.length) return text
+			return text.replace(this.search, `<strong>${this.search}</strong>`)
```
juliushaertl: This only works for the first occurrence

skjnldsv: how so?

juliushaertl: > If pattern is a string, only the first occurrence will be replaced.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace

skjnldsv: Ah yes! sure :)

skjnldsv: Capture d’écran_2019-08-02_17-17-43
Fixed

PR #526, opened 2019-08-05
Allow properly settings avatars for guests as well

rullzer:
For https://github.com/nextcloud/text/issues/171


skjnldsv:
Merging while we wait for a cleanup of the avatar endpoint

PR #538, opened 2019-08-13
Correctly show option by label if it's a tag

### LINKED ISSUE #411: Raw JSON displayed when adding category to array of strings
When you use an array of strings and the tagging feature of the Multiselect component, it shows you raw JSON code in the dropdown after you searched for a category / want to add a new one.

This is how it looks like:
58124057-95bb2600-7c0d-11e9-86b0-ab42cbcff739

Corresponding demo code:
```
<template>
	<div>
		<Multiselect
			v-model="categories"
			:multiple="true"
			:searchable="true"
			:options="categories"
			:placeholder="'Select categories'"
			:taggable="true"
			:tag-placeholder="'Add this as a new category'"
			:close-on-select="false"
			class="multiselect-vue"
			@input="updateCategories"
			@tag="updateCategory"
		/>
	</div>
</template>

<script>
import { Multiselect } from 'nextcloud-vue'

export default {
	components: {
		Multiselect,
	},
	data: function() {
		return {
			categories: ['a', 'b', 'c'],
		}
	},
	methods: {
		/**
		 * Sets the categories of a task
		 *
		 * @param {Array} categories The new categories
		 */
		updateCategories: function(categories) {
			console.debug('Set categories to ' + categories)
		},

		/**
		 * Adds a category to the list of categories
		 *
		 * @param {String} category The name of the category to add
		 */
		updateCategory: function(category) {
			console.debug('Add categoriy ' + category)
		},
	}
}
</script>
```

The same code works just fine with the upstream `vue-multiselect`:
noerror

Code (only the import statement is changed):
```
<template>
	<div>
		<Multiselect
			v-model="categories"
			:multiple="true"
			:searchable="true"
			:options="categories"
			:placeholder="'Select categories'"
			:taggable="true"
			:tag-placeholder="'Add this as a new category'"
			:close-on-select="false"
			class="multiselect-vue"
			@input="updateCategories"
			@tag="updateCategory"
		/>
	</div>
</template>

<script>
import Multiselect from 'vue-multiselect'

export default {
	components: {
		Multiselect,
	},
	data: function() {
		return {
			categories: ['a', 'b', 'c'],
		}
	},
	methods: {
		/**
		 * Sets the categories of a task
		 *
		 * @param {Array} categories The new categories
		 */
		updateCategories: function(categories) {
			console.debug('Set categories to ' + categories)
		},

		/**
		 * Adds a category to the list of categories
		 *
		 * @param {String} category The name of the category to add
		 */
		updateCategory: function(category) {
			console.debug('Add categoriy ' + category)
		},
	}
}
</script>
```

Also see https://github.com/nextcloud/tasks/issues/416 where this was initially reported.

raimund-schluessler:
Closes #411

### DIFF
--- a/src/components/Multiselect/EllipsisedOption.vue
+++ b/src/components/Multiselect/EllipsisedOption.vue
@@ -42,13 +42,7 @@ export default {
 	},
 	computed: {
 		name() {
-			if (this.label) {
-				return this.option[this.label]
-			}
-			if (this.option.isTag) {
-				return this.option.label
-			}
-			return this.option
+			return this.$parent.getOptionLabel(this.option)
 		},
 		needsTruncate() {
 			return this.name && this.name.length >= 10
--- a/src/components/Multiselect/EllipsisedOption.vue
+++ b/src/components/Multiselect/EllipsisedOption.vue
@@ -45,6 +45,9 @@ export default {
 			if (this.label) {
 				return this.option[this.label]
 			}
+			if (this.option.isTag) {
+				return this.option.label
+			}
 			return this.option
 		},
 		needsTruncate() {


skjnldsv:
Reference https://github.com/shentao/vue-multiselect/blob/e7133beee554f996d2aa9f791ea6e911626f8a4f/src/multiselectMixin.js#L480-L491

Shall we add the other checks as well? :)
```js
	/**
     * Returns empty string when options is null/undefined
     * Returns tag query if option is tag.
     * Returns the customLabel() results and casts it to string.
     *
     * @param  {Object||String||Integer} Passed option
     * @returns {Object||String}
     */
    getOptionLabel (option) {
      if (isEmpty(option)) return ''
      /* istanbul ignore else */
      if (option.isTag) return option.label
      /* istanbul ignore else */
      if (option.$isLabel) return option.$groupLabel

      let label = this.customLabel(option, this.label)
      /* istanbul ignore else */
      if (isEmpty(label)) return ''
      return label
    },
```

raimund-schluessler:
> Shall we add the other checks as well? :)

Probably a good idea 👍

skjnldsv:
https://github.com/shentao/vue-multiselect/issues/739

raimund-schluessler:
> 
> 
> [shentao/vue-multiselect#739](https://github.com/shentao/vue-multiselect/issues/739)

Nice, I will have a look in the evening.

skjnldsv:
I guess you can just pass the getOptionLabel like I do here: https://github.com/nextcloud/nextcloud-vue/pull/539/files#diff-7cf0391745a74f4c74643ad9a34fa983R109 :)

raimund-schluessler:
The `getLabel` prop was removed again with https://github.com/shentao/vue-multiselect/pull/773. Don't really know how to use the function at the moment.

raimund-schluessler:
Correct me if I am wrong, but I don't think we have access to the `getOptionLabel` function, since it is not exposed as a prop here https://github.com/shentao/vue-multiselect/blob/master/src/Multiselect.vue#L134 anymore.

Do you know whether we can still use it somehow?

skjnldsv:
Ah damn you're right!
Though you can get it from the component itself :)
Either with the import module: `VueMultiselect.mixins.find(mixin => mixin.methods.getOptionLabel).methods.getOptionLabel`

Or with a ref `this.$refs.vuemultiselect.getOptionLabel` :)

PR #395, opened 2019-05-15
Trigger value save on year/month select

skjnldsv:
See https://github.com/nextcloud/nextcloud-vue/issues/137#issuecomment-481667668
Closes #137

@nextcloud/vue

PR #548, opened 2019-08-26
Fix year/month select on the datepicker

skjnldsv:
See https://github.com/nextcloud/nextcloud-vue/pull/395#issuecomment-518596537


src/components/DatetimePicker/DatetimePicker.vue
```diff
@@ -76,22 +77,24 @@ export default {
 
 	methods: {
 		handleSelectYear(year) {
-			if (this.value) {
+			const value = this.$refs.datepicker.currentValue
```
skjnldsv: this is a github glitch I think, the indent is good! :thinking:

Issue #356, opened 2019-04-19
Make it possible to have a disabled action


ChristophWurst:
This is something to add once @skjnldsv finished the action components rework. It would be great if actions with a callback could have a disabled state, e.g. because the operation is not allowed or something is still being loaded. It might also make sense for inputs like texts and checkboxes.

jancborchardt:
@jenniferpiperek and I were talking about this among other things and she proposed that when we show all items in full opacity, we can use opacity for disabled states. :)

Of course there should also still be notices on why something is disabled, e.g. as tooltip which will show on hover or on tap.

skjnldsv:
Hum, we use the .5/.7/1 opacity on a lot of others places on nextcloud.
I would keep the disabled state to no hover feedback + .5 opacity. :)

And with default cursor :)

jancborchardt:
Yes, and the idea of Jenni was that (I'm thinking except in the header bar) we don't use transparency so much anymore.

For example if you check out the conversations list in the Talk app navigation, all the chatd except the selected one look disabled in a way.

Similarly for the app-content-list component we don't use opacity but the full row background, which makes it much easier to see what is active. :)

ma12-co:
What about the blue border on the left, are we getting rid of it as well for disabled actions? 

![Screenshot_20190826_170147](https://user-images.githubusercontent.com/26852655/63700481-46dfe680-c823-11e9-8cf8-87051b600ec0.png)

skjnldsv:
Well, there should not be a blue border if there is no focusable class in an action. Like ActionText. It doesn't have a .focusable class, so the focus manager method in Actions don't apply any blue border to it :)

ma12-co:
Right, but toggling the `focusable` class doesn't affect the hover behavior, you still get the blue shadow on hover even if you don't have apply it. Should we add this to the focus manager in the `Actions` component?

skjnldsv:
Ah, I thought I removed it, well.
Since the mouse is also managing the focus, we shouldn't nee the hover state anymore :)

ma12-co:
working on this here  #550

ma12-co:
@nextcloud/vue @jancborchardt Of all the actions, which ones do you think should have the possibility to be disabled? (grayed out / not clickable / no hover feedback / tooltip)?

skjnldsv:
@ma12-co all of them except the text only one?

ma12-co:
Ok!

[...]

PR #550, opened 2019-08-27
feature/356/make-it-possible-to-have-a-disabled-action

ma12-co:
Added  modifier to the action mixin and bound the disabled class to the ActionButton component

#356 

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4627,10 +4627,13 @@
       }
     },
     "eslint-utils": {
-      "version": "1.3.1",
-      "resolved": "https://registry.npmjs.org/eslint-utils/-/eslint-utils-1.3.1.tgz",
-      "integrity": "sha512-Z7YjnIldX+2XMcjr7ZkgEsOj/bREONV60qYeB/bjMAqqqZ4zxKyWX+BOUkdmRmA9riiIPVvo5x86m5elviOk0Q==",
-      "dev": true
+      "version": "1.4.2",
+      "resolved": "https://registry.npmjs.org/eslint-utils/-/eslint-utils-1.4.2.tgz",
+      "integrity": "sha512-eAZS2sEUMlIeCjBeubdj45dmBHQwPHWyBcT1VSYB7o9x9WRRqKxyUoiXlRjyAwzN7YEzHJlYg0NmzDRWx6GP4Q==",
+      "dev": true,
+      "requires": {
+        "eslint-visitor-keys": "^1.0.0"
+      }
     },
     "eslint-visitor-keys": {
       "version": "1.0.0",


src/components/ActionButton/ActionButton.vue
```diff
@@ -42,7 +42,8 @@ If you're using a long text you can specify a title
 
 <template>
 	<li>
-		<button class="action-button focusable" :disabled="disabled" @click="onClick">
+		<button class="action-button focusable" @click="onClick"
+			:class="disabled ? 'action-button--disabled' : ''">
```
skjnldsv: cleaner?
```suggestion
			:class="{ 'action-button--disabled': disabled }">
```

src/components/Actions/Actions.vue
```diff
@@ -302,7 +302,7 @@ export default {
 		},
 		focusAction() {
 			// TODO: have a global disabled state for non input elements
-			const focusElement = this.$refs.menu.querySelectorAll('.focusable:not(:disabled)')[this.focusIndex]
+			const focusElement = this.$refs.menu.querySelectorAll(`.focusable:not([class$='disabled'])`)[this.focusIndex]
```
skjnldsv: I would rather use this in case the browser add the disabled class before the focusable.
```suggestion
			const focusElement = this.$refs.menu.querySelectorAll(`.focusable:not([class*='disabled'])`)[this.focusIndex]
```

skjnldsv: Also, there are other refs.menu.querySelectorAll bellow :)

src/components/ActionButton/ActionButton.vue
ma12-co: Much cleaner :)

skjnldsv:
Rebase and squash :)

ma12-co:
There's an issue to fix in the Actions component: even if the first Action is disabled, it gets focused when opening the menu. A fix for this would be removing the `focusable` class in each component when the disabled prop is passed in. What do you think @skjnldsv?

Screenshot_20190828_115329

skjnldsv:
> What do you think @skjnldsv?

An issue with `focusFirstAction`?

ma12-co:
> An issue with `focusFirstAction`?

I don't think it's an issue with the action component, I think that disabled items shouldn't have the `.focusable` class. After all, they're not focusable :) 
If we'd toggle the `focusable` class on/off when the action is enabled/disabled there'd be no need to touch the methods in the actions component. Everything would work as expected and `focusFirstAction` would focus on the first actually focusable (not disabled) action.

skjnldsv:
Sure then :)
Let's just remove the focusable state! :+1:

ma12-co:
@skjnldsv  Computed?? 
Screenshot_20190828_124959

skjnldsv:
`isFocusable` then :)

PR #565, opened 2019-08-30
Allow wrapped children in validation script

skjnldsv:
If you're passing a component as variable, you will have to wrap it in your app
Because you can import `ActionButton`, but if you do
```vue
<Actions>
	<component :is="ActionButton" />
</Actions>
<script>
import ActionButton from 'nextcloud-vue/dist/components/ActionButton'
```
Then importing make no sense as you need to have your own methods, provide a click handler, etc etc.

So you will need to wrap it into your own `myComponent` that uses `ActionButton`
```vue
<template>
	<ActionButton icon="icon-up" @click="copyNtoFN">
		{{ t('contacts', 'Copy to full name') }}
	</ActionButton>
</template>
<script>
import ActionButton from 'nextcloud-vue/dist/components/ActionButton'
export default {
	name: 'myComponent',
```
```vue
<Actions>
	<component :is="myComponent" />
</Actions>
<script>
import myComponent from './myComponent'
```
But then you'll end up with `myComponent` as direct children of `Actions` and that was forbidden before

Capture d’écran_2019-08-30_18-25-23

### DIFF
--- a/src/utils/ValidateChildren.js
+++ b/src/utils/ValidateChildren.js
@@ -29,9 +29,16 @@ import Vue from 'vue'
  */
 const ValidateChildren = (vm, allowed) => {
 	vm.$children.forEach((child, index) => {
-		if (allowed.indexOf(child.$options.name) === -1) {
+		const isChildren = allowed.indexOf(child.$options.name) !== 1
+		const isWrappedChildren = child.$children.length === 1
+			&& allowed.indexOf(child.$children[0].$options.name) === -1
+
+		// allow direct children action or wrapped action in case of using
+		// a fake component with :is="myActionWrapper" and some custom methods
+		// that you import  with import myActionWrapper from 'myActionWrapper'
+		if (!(isChildren || isWrappedChildren)) {
 			// warn
-			Vue.util.warn(`${child.$options._componentTag} is not allowed inside the ${vm.$options._componentTag} component`, vm)
+			Vue.util.warn(`${child.$options.name} is not allowed inside the ${vm.$options.name} component`, vm)
 
 			// cleanup
 			vm.$children.splice(index, 1)


skjnldsv:
> Honestly, I would remove that validation at all. There may be good reasons why to use another constellation, for which the new test will fail. So why be that restrictive?

I'm a bit thorn as well!
But Devs will do anything to implement their own stuff if they think it doesn't fit their cases. But we're talking about UIs stuff here. I spent already too many hours implementing standards to have a unified design on nextcloud (which was faaar from being the case 3 years ago)
And I've had some many people trying to work around stuff because they want to do it their own ways. https://blog.codinghorror.com/this-is-what-happens-when-you-let-developers-create-ui/

So I usually prefer to implement some harsh limits but do everything I can to keep the dialog open and implement missing features that devs requires.
But that way we have a unique way of doing so.

@jancborchardt any opinion? :)

jancborchardt:
I’m not sure I can judge on that – would rather hear from people who do write the code like @juliushaertl @georgehrke @ChristophWurst @ma12-co @marcelklehr @eneiluj, other app developers and well, @korelstar and @skjnldsv ;)

ma12-co:
We could move all these checks from the components themselves to some sort of 'UI compliance test', and require devs to pass it before publishing. This way we'd improve the performance of the apps while still enforcing these basic UI rules.

marcelklehr:
Generally, I think it's a good idea to have constraints on these things -- just like typed paramters on a function make it more predictable and robust, the same applies here, I think. It's better to have explicit rule for how something is to be used, than to generally allow anything and invite people to take advantage of some implicit behaviour. And, as @skjnldsv said, if there's a genuine usecase, things can be adjusted explicitly.

In this case, I'm not even sure if the example necessitates any changes to the constraints, as there are workarounds available (even if they're not that beautiful): you could for example apply a click listener that differentiates between the different possible components, or use v-if instead of dynamic components. Also, it might be worth thinking about pulling such a custom ActionButton into this library itself, instead of allowing all kinds of customizations.

PR #569, opened 2019-09-02
Fix actions events and performances

skjnldsv:
- Slow performances because all menus are hidden with v-show: switching to v-if
- We were using `$children`, so v-if don't display anything and since we create `this.actions` from `$children`, it was empty: now using `$slots.default` as the array is filled regardless of the template being shown or not.

### DIFF
--- a/src/components/Actions/Actions.vue
+++ b/src/components/Actions/Actions.vue
@@ -76,7 +76,8 @@ https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-action
 			:aria-expanded="opened"
 			@click.prevent="toggleMenu"
 			@keydown.space.exact.prevent="toggleMenu" />
-		<div ref="menu"
+		<div v-if="opened"
+			ref="menu"
 			:class="[`menu-${menuAlign}`, { 'open': opened }]"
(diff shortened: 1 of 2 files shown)


skjnldsv:
> Gives a small performance boost: from ~2,1 to ~1,5 seconds respective from ~5,4 to 4,4 seconds (reference see [#486 (comment)](https://github.com/nextcloud/nextcloud-vue/pull/486#issuecomment-526935407)).

How do you calculate this?
It seems different from what I get

korelstar:
I start the vueexample app with the customized `App.vue` (200 items, see link above) on two different computers in Firefox. The time period is calculated from the difference of the timestamps of the console debug output (on update, updated).

skjnldsv:
Already an improvement!
So after some testing, this is not coming from the scoping at all.
I'll try to check the performances potential later :)

skjnldsv:
Btw, it's much more faster in production mode. :)
```
08:38:43.707 App.vue:74 vue: beforeUpdate
08:38:43.955 App.vue:90 vue: updated
08:38:43.958 App.vue:92 vue: updated -> nextTick
```

Though this is quite the amount of items, not sure where to start to optimise :thinking:

PR #571, opened 2019-09-03
Create ColorPicker component

### LINKED ISSUE #138: Provide color picker
There should be a standardised color-picker in Nextcloud because it's needed in many places.
E.g.:
- calendar app
- mail app (cc @ChristophWurst)
- (Deck app? @juliushaertl)

This is the color-picker in the Calendar app (v1):
Safari, IE:
55d7a0ab-ac58-4ea7-8ccd-70f6baa5a3cd

Firefox, Chrome, Edge:
cbcfbb97-beb7-4d25-a678-40c04ee397f3

It provides a default set of colors plus a random color option in Safari, IE and a real color-picker for browsers that natively support it.
https://caniuse.com/#feat=input-color

The color-picker should also work inside the popover menu.

ma12-co:
Fix #138
design: https://www.figma.com/file/4th1eQVONZguVoDXMwnDJT/nc-color-picker?node-id=0%3A1
Screenshot_20190906_152526

Please use [this app.vue for testing](https://gist.github.com/ma12-co/5b3e6604ba6c5ba9c128adcf1f6f565f)


skjnldsv:
Fixed icon 
Capture d’écran_2019-09-04_17-35-56

src/components/ColorPicker/ColorPicker.vue
```diff
@@ -0,0 +1,190 @@
...
+	border-radius: 3px;
+	display: flex;
+	flex-direction: column;
+	justify-content: space-between;
+	align-content: flex-end;
+	overflow: hidden;
+	box-sizing: content-box !important;
+	&-simple {
+		display: grid;
+		grid-template-columns: repeat(4, 41px);
```
skjnldsv: 44px? :grin:

src/components/ColorPicker/ColorPicker.vue
```diff
@@ -0,0 +1,190 @@
...
+			&.more-settings {
+				@include iconfont('more');
+				margin-left: auto;
+			}
+		}
+	}
+}
+
+::v-deep .vc-chrome {
+	width: 164px;
```
skjnldsv: if the content is properly designed and have proper sizing and layout, youshould not need to have custom width. width: auto is enough.

skjnldsv: Same line 84 ;)

src/components/ColorPicker/ColorPicker.vue
ma12-co: $clickable-area? :)

src/components/ColorPicker/ColorPicker.vue
ma12-co: These are needed to overrule the defaults coming from the vue-color library, I'm afraid I cannot take them out

src/components/ColorPicker/ColorPicker.vue
ma12-co: Done!

ma12-co:
@skjnldsv I don't think there's the need to use the color generator algorithm here. If a user opens this it's because he wants to choose a color so IMO the simple picker should have a selection of nice hard-coded colors only. What do you think?

skjnldsv:
> IMO the simple picker should have a selection of nice hard-coded colors only

Yes, that is why the colour generator is here. It's generating the exact same colours everytime and that are the one we're using all over nextcloud :)
We could hardcode them, but it's easier to have a function for it that we can update later if we want to tune our nextcloud default colours ;)

ma12-co:
Hey @nextcloud/designers, this is the current state of things, the design has changed since the mock-up so I need some feedback:

- simple picker:

Screenshot_20190909_174846

- advanced picker: 

Screenshot_20190909_174815

for feedback and animations check out this video: https://cloud.ma12.co/index.php/s/KaRRTfP9iDExGZ5

What do u think?

skjnldsv:
- Alignment of the buttons is a bit off (uneven margins?)
- Centering of the button icons is off too (display flex, align-items & justify-content)
- Animation is great! :+1:

ma12-co:
thanks @skjnldsv!

> Alignment of the buttons is a bit off (uneven margins?)

This one is because of the hover state of the 'back' button:

Screenshot_20190909_175859

skjnldsv:
> This one is because of the hover state of the 'back' button:

I meant more like this:
64546869-bda0d780-d32b-11e9-9126-92a311c6bb3d

ma12-co:
got it @skjnldsv, it was a margin coming from the server's css. 
All fixed :)

skjnldsv:
Peek 09-09-2019 19-28

src/components/ColorPicker/ColorPicker.vue
```diff
@@ -38,7 +38,7 @@ You have to provide an initial HEX value to initialise it.
 
 ```
 <Popover>
-	<ColorPicker value="" @update:value="doSomething" />
+	<ColorPicker value="#someHexColor" @update:value="doSomething" />
```
skjnldsv: Maybe also add an example with a simple v-model?

jancborchardt:
Great stuff @ma12-co! :) Agree with @skjnldsv’s points:
- The button having a checkmark like on the colors is confusing. We should use text there only, either "Choose" or "Pick". We use "Choose" elsewhere like the file picker so I’d prefer that, and it’s also more commonly used in other software.
- The first color picker view with the color list also needs a primary confirm button. Otherwise I don’t know how to confirm the color (unless it gets confirmed by clicking on it directly and the popup closes – does Google do that?)
- So in the first picker view, the 3-dot icon could be on the bottom left instead, which would make the "go to details" and "go back" symmetric, you can just click in the same place.

ma12-co:
Thanks @jancborchardt!

>The button having a checkmark like on the colors is confusing. We should use text there only, either "Choose" or "Pick". 

I'll try this and we'll see how it looks :) 

>unless it gets confirmed by clicking on it directly and the popup closes

Nice catch! this is indeed the intended behavior but it's not properly implemented atm!

>So in the first picker view, the 3-dot icon could be on the bottom left instead, which would make the "go to details" and "go back" symmetric, you can just click in the same place.

makes sense :)

ma12-co:
@jancborchardt, everything updated, I'm not a fan of the 'choose' text instead of the checkmark button though, I think it takes up a lot of space and it's a bit invasive.. Here it is: 

Screenshot_20190910_164952

skjnldsv:
> I think it takes up a lot of space and it's a bit invasive

Agreed, shall we move to a white bg and primary border?

georgehrke:
> Agreed, shall we move to a white bg and primary border?

Or just emit the change event on change, not only on save and get rid of that button completely?

skjnldsv:
> Or just emit the change event on change, not only on save and get rid of that button completely?

I actually think it's nice to wait for a validation. On android I found both use cases:
1. the popup closes and the user see the changes instantly
2. the popup stays and the ui in the background updates so you get a preview of what you did.

I actually like the (2), like the calendar for example, you could update the whole ui accordingly and let the user change colours until it fits. I always have to click a few colours to see how well it matches with the other calendars that I have :)

It gives you a little better sense of control over the setting. ALso, on mobile, clicking outside cannot be the only way to quit a popup, so if you want to cancel, you'll need a button for it. Choose is fine if the user did not changed anything.

georgehrke:
@ma12-co Somehow I'm too stupid to test this. Could you push an example branch to https://github.com/skjnldsv/vueexample?

I added a `ColorPicker` component inside a `Popover`, also set `:open="true"` on the popover, but somehow i can't see anything and the corresponding div in the Chrome inspector is empty.

ma12-co:
Sorry @georgehrke, that's totally my fault. Docs are a misleading as we still don't have the popover component. And the colorpicker now works properly only with `v-model`
Please use [this app.vue for testing](https://gist.github.com/ma12-co/5b3e6604ba6c5ba9c128adcf1f6f565f)

skjnldsv:
@ma12-co please rebase and squash :)

ma12-co:
Done!

georgehrke:
🎉

PR #573, opened 2019-09-04
Create Popover component

### LINKED ISSUE #572: Create general purpose 'popover' component
It would be nice to have a general purpose popover component that we could use as a wrapper for Actions, DatetimePicker, ColorPicker, Contact search ?,  Settings?, Notifications?, ..

The behavior of these elements is pretty similar throughout so I think it would be nice to group the logic and not repeat ourselves, what  do you think?

ma12-co:
Fix #572

### DIFF
--- a/src/components/Popover/Index.js
+++ b/src/components/Popover/Index.js
@@ -1,0 +1,25 @@
+/**
+ * @copyright Copyright (c) 2019 Marco Ambrosini <marcoambrosini@pm.me>
+ *
+ * @author Marco Ambrosini <marcoambrosini@pm.me>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


ma12-co:
So this approach is In my opinion a very simple way to solve this using[ the v-tooltip library.](https://github.com/Akryum/v-tooltip#popover-component-reference) 
The component has 2 slots:

- trigger
The idea is to pass in a trigger which can be any HTML element (in the example below it's a span but it can be anything).  We can also automatically toggle it open/close with a prop, the trigger element is needed merely for the position of the popover. 

- content
whatever we want to put in the popover.

the code in the application would look kinda like this:
```
<Popover :open="open">
	<span slot="trigger"> I am the trigger </span>
	<Colorpicker>
</Popover>
```
With this we get standardization, automatic positioning and many more props to customise it!

I wanted to know your opinion @skjnldsv  @juliushaertl  before proceeding with tweaking styles and props.

juliushaertl:
Looks good to me. Also having the trigger slot optional makes sense :+1:

skjnldsv:
So, the no trigger docs option was working, but since clicking anywhere on the page closes the popovers, by the time you're reaching the example, it will be gone ^^



```vue
<template>
	<Popover :open="opened">
		<template>
			<h2>this is some content</h2>
		</template>
	</Popover>
</template>

<script>
export default {
	data() {
		return {
			opened: true
		}
	}
}
</script>
```

PR #587, opened 2019-09-12
Add ActionRadio

skjnldsv:
Bug:
- [x] event is not fired on the unchecked radio.

### DIFF
--- a/src/components/ActionRadio/ActionRadio.vue
+++ b/src/components/ActionRadio/ActionRadio.vue
@@ -23,26 +23,29 @@
 
 <docs>
 This component is made to be used inside of the [Actions](#Actions) component slots.
+Usually, you will provide a name prop to bind the radio together.
+So that only one of each name set can be selected at the same time.
 
 ```vue
 	<Actions>
-		<ActionCheckbox @change="alert('(un)checked !')">First choice</ActionCheckbox>
-		<ActionCheckbox value="second" @change="alert('(un)checked !')">Second choice</ActionCheckbox>
-		<ActionCheckbox :checked="true" @change="alert('(un)checked !')">Third choice (checked)</ActionCheckbox>
-		<ActionCheckbox :disabled="true" @change="alert('(un)checked !')">Second choice (disabled)</ActionCheckbox>
+		<ActionRadio @change="alert('(un)checked !')" name="uniqueId">First choice</ActionRadio>
+		<ActionRadio value="second" name="uniqueId" @change="alert('(un)checked !')">Second choice</ActionRadio>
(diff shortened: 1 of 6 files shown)


skjnldsv:
Bump :)

src/components/ActionRadio/ActionRadio.vue
```diff
@@ -0,0 +1,244 @@
...
+
+			if (this.$refs.radio.checked) {
+				/**
+				 * Emitted when the radio is checked
+				 * @type {Event}
+				 */
+				this.$emit('check', true)
+
+				// if part of a set, firing events on all the other
+				if (this.name) {
```
ChristophWurst:
```suggestion
				if (!this.name) {
```

and early return to lower the number of indentations

src/components/ActionRadio/ActionRadio.vue
```diff
@@ -0,0 +1,244 @@
...
+					const parentAction = GetParent(this, 'Actions')
+					if (parentAction) {
```
ChristophWurst:
```suggestion
					if (!parentAction) {
```

and early return

src/components/ActionRadio/ActionRadio.vue
```diff
@@ -0,0 +1,244 @@
...
+			 * @type {Event}
+			 */
+			this.$emit('change', event)
+
+			if (this.$refs.radio.checked) {
+				/**
+				 * Emitted when the radio is checked
+				 * @type {Event}
+				 */
+				this.$emit('check', true)
```
ChristophWurst: even if $emit is async I would argue that it should be fired *after* the `update:checked` was done, so we're 200% sure the listeners listening to change will get the new value when they read it

src/utils/GetChildren.js
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+/**
+ * Get the all the children matching the provided name
+ *
+ * @param {Object} context the context to search from (usualy this)
+ * @param {string} name the children name
+ * @returns {Object[]} the children components
+ */
+const GetChildren = function(context, name) {
```
ChristophWurst:
```suggestion
const getChildren = function(context, name) {
```

this is not c#

src/utils/GetParent.js
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+/**
+ * Get the first parent component matching the provided name
+ *
+ * @param {Object} context the context to search from (usualy this)
+ * @param {string} name the parent name
+ * @returns {Object} the parent component
+ */
+const GetParent = function(context, name) {
```
ChristophWurst:
```suggestion
const getParent = function(context, name) {
```

Issue #425, opened 2019-06-02
Actions: "active" style doesn't work


korelstar:
I would like to use the `actions` component with some `action` components. For some of them, the entry should reflect an enabled state (e.g. *fullscreen mode* is enabled/disabled). Therefore, I tried to pass the `active` class to the respective action component, e.g. `ActionButton`.

However, that class is directly removed, when opening the popover-menu from the `actions` component or when moving the mouse over some items (this can be observed using the browser's web developer tools).

Hence, currently it is not possible to set an *active state* for an `actions` item. It would be nice, if it would.

skjnldsv:
The active is just to force a hover/focus state.
You want a disable/enable state instead of the active, right?

korelstar:
Okay, I think I got the meaning of `active` wrong. In the App-Navigation, the CSS class `active` determines also the currently selected element (not just hover and focus). I thought this is the same, here.

However, I think "disabled" is also not the correct term, since the action should still be triggerable. Hence, this is different to #356.

As an alternative, I could use different labels for the same `ActionButton`, e.g. "Launch fullscreen mode" and "Close fullscreen mode" (these labels are quite long 😟) and/or colorize the icon in order to let the design emphasize the state. This doesn't require any changes in `nextcloud-vue`, but maybe we should have a common design speak for this.
/cc @jancborchardt

jancborchardt:
At least in current implementation (like Talk app notification level in 3-dot menu, and personal settings federation privacy settings) we use the `active` class for, well, the active element.

If there is not a specific reason otherwise, `.active` and hover/focus could be the same. That is, blue bar on the left and full opacity.

skjnldsv:
I'm not sure I really get the topic here :see_no_evil: 
could you  maybe illustrate it? Why not just change the icon for the full screen toggle? You know what state it is since it's a toggle, right?

korelstar:
Okay, here we go:

### App Navigation

In the app navigation, the currently selected item is shown with a blue bar at the left and full opacity. For this, the CSS class `active` is used. In the following screenshot, the item "Neueste" is active.
![Screenshot](https://user-images.githubusercontent.com/6277619/59107681-732d4a80-8939-11e9-89cb-c539b4726d90.png)

The same design is used when hovering an item from the app navigation. In the following screenshot, the item "Neueste" is active and I'm hovering the item "Favoriten":

[...]

PR #593, opened 2019-09-14
AppNavigation: fix style for CSS class "active"

korelstar:
Since  #486, `#app-navigation` has CSS class `vue` and Nextcloud server has the following style:

```
#app-navigation:not(.vue) > ul > li.active, #app-navigation:not(.vue) > ul > li a:hover, #app-navigation:not(.vue) > ul > li a:focus, #app-navigation:not(.vue) > ul > li a:active, #app-navigation:not(.vue) > ul > li a.selected, #app-navigation:not(.vue) > ul > li a.active, #app-navigation:not(.vue) > ul > li.active > a, #app-navigation:not(.vue) > ul > li a:hover > a, #app-navigation:not(.vue) > ul > li a:focus > a, #app-navigation:not(.vue) > ul > li a:active > a, #app-navigation:not(.vue) > ul > li a.selected > a, #app-navigation:not(.vue) > ul > li a.active > a {
	opacity: 1;
	box-shadow: inset 4px 0 var(--color-primary);
}
```

This means, that the `active` style from server is not applied anymore to `AppNavigationItem`. However, the component doesn't include it's own style for `active` CSS class. This is fixed with this PR.

### DIFF
--- a/src/components/AppNavigationItem/AppNavigationItem.vue
+++ b/src/components/AppNavigationItem/AppNavigationItem.vue
@@ -399,6 +399,7 @@ export default {
 	width: 100%;
 	min-height: $clickable-area;
 
+	&.active,
 	a:hover,
 	a:focus,
 	a:active {


skjnldsv:
Ah, we have an ongoing discussion about this in #425 :)
Which will also be applied to the navigation, so we'll wait for it to be integrated I think! @ma12-co ?

korelstar:
I'm aware of that. But it's unclear **when** this will be implemented. Currently, there is no way to identify the active item at all. Therefore, I think this intermediate quick-fix is helpful. Otherwise, testing an app using `master` is very annoying.

skjnldsv:
@korelstar ahah, sorry :)
Let's merge this then sure!

skjnldsv:
Ah no, too different, let's keep on 13.0.x

PR #597, opened 2019-09-16
fix check for non-tab-content

korelstar:
There was still an error message when using no tabs in `AppSidebar`. Since tabs should be optional, the validation has to be changed (see https://github.com/nextcloud/nextcloud-vue/pull/447).

My approach here is to find valid tabs (a tab is valid if it has `name`, `icon` and `id`) and then check if there is either tabs only or non-tab-content only. Otherwise, a warning is given.

Please check in detail, if this is sufficient for you.


korelstar:
Ouh :-(

What are the symptoms?

Do you get the warning "This tab is missing a valid name" if you use the master branch of nextcloud-vue?

korelstar:
I changed one thing. Now, the behavior should be the same like before, but less false-positive warnings.

Please try again :-)

PR #598, opened 2019-09-17
UserBubble component

violoncelloCH:
introduces a new UserBubble component to be used e.g. for @-mentions in social or talk
it features an optional popover to e.g. display the full username-handle for remote users in social

it should also be possible to link the UserBubble to the user's profile (social), but I'd like to have this optional, but didn't manage to do this, because the `<a>` element would need to wrap the whole bubble and therefore I can't just use `v-if` on it. Any idea @skjnldsv or @ma12-co ?

### DIFF
--- a/src/components/UserBubble/UserBubble.vue
+++ b/src/components/UserBubble/UserBubble.vue
@@ -21,11 +21,13 @@
 -->
 
 <template>
-	<Popover trigger="hover focus" :open="open" class="test"
-		@update:open="onOpenChange">
-		<div slot="trigger" class="user-bubble">
-			<Avatar v-bind="$attrs" :user="user" :size="16"
-				:disable-tooltip="true" :disable-menu="true" class="avatar" />
+	<Popover trigger="hover focus" :open="open" :disabled="popoverEmpty"
+		class="user-bubble-popover" @update:open="onOpenChange">
+		<div slot="trigger" v-bind="linkOrNot" class="user-bubble">
+			<slot name="avatar">
+				<Avatar v-bind="$attrs" :user="user" :size="16"
+					:disable-tooltip="true" :disable-menu="true" class="avatar" />
+			</slot>
 			<h6 class="user">
 				{{ displayName ? displayName : user }}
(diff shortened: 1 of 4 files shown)


ma12-co:
@violoncelloCH  In talk we can also mention a conversation so we need also the capability to display an icon instead of the avatar component in the bubble, could you add this?

violoncelloCH:
thanks for the suggestion @ma12-co !
I've implemented this as a slot (name=avatar), which is by default filled with the avatar... and I also added the possibility to conditionally have a link on the bubble and also the popover is now completely disabled if there is no content in it

src/components/UserBubble/UserBubble.vue
```diff
@@ -0,0 +1,146 @@
...
+.user-bubble-popover {
+	display: inline;
+	vertical-align: top;
+}
+
+.user-bubble {
+	display: flex;
+	height: 20px;
+	border-radius: 10px;
+	background-color: var(--color-background-dark);
```
skjnldsv:
```suggestion
	background-color: var(--color-background-dark);
	h6 {
		line-height: 20px;
	}
```

src/components/UserBubble/UserBubble.vue
```diff
@@ -0,0 +1,146 @@
...
+	</p>
+</template>
+```
+</docs>
+
+<template>
+	<Popover trigger="hover focus" :open="open" :disabled="popoverEmpty"
+		class="user-bubble-popover" @update:open="onOpenChange">
+		<div slot="trigger" v-bind="linkOrNot" class="user-bubble">
+			<slot name="avatar">
```
skjnldsv: Let's use a prop and detect if it's a url instead of an avatar
So it can be a css class or just a plain url

violoncelloCH: implemented!

src/components/UserBubble/UserBubble.vue
violoncelloCH: implemented!

PR #601, opened 2019-09-19
Allow auto match from value if useing track-by

skjnldsv:
Does it make sense? :)

### DIFF
--- a/src/components/Multiselect/Multiselect.vue
+++ b/src/components/Multiselect/Multiselect.vue
@@ -43,6 +43,40 @@ export default {
 </script>
 ```
 
+### Simple example with objects
+You can either use the exact object or the track-by key to match against your options
+
+```vue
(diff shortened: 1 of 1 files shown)


skjnldsv:
This allow us to use a key as value that will be automatically matched against the options objects list instead of having to pass the full object value

korelstar:
Looks good, but when experimenting with your new example in the documentation, I found this weird scenario:

- open drop-down
- click on "Option 1"
- see `Selected option: { "id": 1, "label": "Option 1" }"`
- open drop-down again
- click on "Option 1" again
- see `Selected option: `

For `:multiple="true"`, this seems to be fine, but not for single select.

skjnldsv:
@korelstar because I did not forbid the removal of selection on the docs?
`allowEmpty` is true by default :)

PR #606, opened 2019-09-24
Feature/595/editable title in app sidebar

georgehrke:
simple-input

### DIFF
--- a/src/components/AppSidebar/AppSidebar.vue
+++ b/src/components/AppSidebar/AppSidebar.vue
@@ -23,6 +23,68 @@
 <!-- Follows the tab aria guidelines
 	https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html -->
 
+<docs>
+
+	### General description
+
+	This component provides a way to include the standardised sidebar.
+	The standard properties like title, subtitle, starred, etc. allow to automatically
+	include a standard-header like it's used by the files app.
+
+	### Standard usage
+
+	```vue
+	<template>
+		<AppSidebar
+			:title="title"
+			:subtitle="subtitle">
+			<!-- Insert your slots and tabs here -->
+		</AppSidebar>
+	</template>
+	<script>
+		export default {
+			data() {
+				return {
+					title: 'cat-picture.jpg',
+					subtitle: 'last edited 3 weeks ago'
+				}
+			}
+		}
+	</script>
+	```
+
+	### Editable title
+	```vue
+	<template>
+		<AppSidebar
+			:title="title"
+			:title-editable="true"
+			:title-placeholder="titlePlaceholder"
+			:subtitle="subtitle"
+			@update:title="titleUpdate">
+			<!-- Insert your slots and tabs here -->
+		</AppSidebar>
+	</template>
+	<script>
+		export default {
+			data() {
+				return {
+					title: 'cat-picture.jpg',
+					titlePlaceholder: 'Filename',
+					subtitle: 'last edited 3 weeks ago'
+				}
+			},
+			methods: {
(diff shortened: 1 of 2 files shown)


src/components/AppSidebar/AppSidebar.vue
```diff
@@ -397,7 +424,7 @@ $desc-height: 46px;
 			position: relative;
 			padding: #{$desc-vertical-padding} #{$desc-menu-right-margin * 4} #{$desc-vertical-padding} $desc-vertical-padding / 2;
 			display: flex;
-			height: $desc-height / 2;
+			height: $desc-height * 0.5;
```
skjnldsv: Why? I feel like `/ 2` is more clear

src/components/AppSidebar/AppSidebar.vue
```diff
@@ -325,6 +336,22 @@ export default {
...
+			/**
+			 * Emitted on title events of the text field
+			 * @type {Event|Date}
+			 */
+			this.$emit('inputTitle', event)
+			/**
+			 * Emitted when the title value changes
+			 * @type {string|Date}
+			 */
+			this.$emit('updateTitle:value', event.target.value)
```
skjnldsv: if this is the only input editable here, let's just update:title, and emit an input & change event?

src/directives/Focus/index.js
```diff
@@ -0,0 +1,29 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+export const directive = {
+	inserted(el) {
+		el.focus()
```
skjnldsv: interesting! :rocket:

skjnldsv:
Don't forget the documentation ;)

src/components/AppSidebar/AppSidebar.vue
georgehrke: I changed it to `* 0.5`, because i also use `* 1.5` further down and i wanted it to be the same. I'm fine with changing it back though.

src/components/AppSidebar/AppSidebar.vue
georgehrke: Roger that!

skjnldsv:
> In that case the app could register an action to `Rename` and set `title-editable` dynamically.

yes, that what I had in mind too :)

georgehrke:
With style adjustments:

B4C4BBDC-3B7D-4AB9-B384-3FB4ACF4F1FA

skjnldsv:
And with subtitles?

georgehrke:
E9FB992D-AF33-470C-A0AB-C8E7575D320E

src/components/AppSidebar/AppSidebar.vue
```diff
@@ -23,6 +23,28 @@
...
+
+	This component provides a way to include the standardised sidebar.
+	The standard properties like title, subtitle, starred, etc. allow to automatically
+	include a standard-header like it's used by the files app.
+
+	### Usage
+
+	```
+	<AppSidebar
+		:title="title"
```
skjnldsv: Docs needs to work. Here title and subtitle does not exists as variables :)

Also could you also provide example for the regular one since we're at it? :p
We did not had the documentation back then! :grin:

skjnldsv:
Looks good!
But the documentation looks very weird. Shall we add other stuff? Like a background? Compact mode?

georgehrke:
>  Shall we add other stuff? Like a background? Compact mode?

Can we do that in a follow up PR? :)

skjnldsv:
Yes ;)

PR #614, opened 2019-09-30
Adjust font size of navigation elements

ma12-co:
As a result of the increment of the body's font-size to 15px, the navigation elements were too big. This pr brings the navigation item's titles down to 14px and the counter's text back down to 13px.

Before: 
Screenshot_20190930_103302

After (you can see how nav sizes and content size relate to each other):
Screenshot_20190930_105219

### DIFF
--- a/src/components/AppNavigationCounter/AppNavigationCounter.vue
+++ b/src/components/AppNavigationCounter/AppNavigationCounter.vue
@@ -67,6 +67,7 @@ export default {
 	text-align: center;
 	text-overflow: ellipsis;
 	line-height: 1em;
+	font-size: 0.8125rem; //13px with root font size set at 16px
 
 	&--highlighted {
 		padding: 4px 6px;
--- a/src/components/AppNavigationItem/AppNavigationItem.vue
+++ b/src/components/AppNavigationItem/AppNavigationItem.vue
@@ -398,6 +398,7 @@ export default {
 	box-sizing: border-box;
 	width: 100%;
 	min-height: $clickable-area;
+	font-size: 0.875rem; //14px with root font size set at 16px
 
 	&.active,
 	a:hover,


skjnldsv:
I'm not sure about this.
Playing with font size is usually considered a bad move :)
Maybe it's the time for us to move to dynamic width navigation (same as sidebar) 

DKGK-IBUEAE58Ll

ma12-co:
Why a bad move?? I think that creating a visual hierarchy with font sizes is very useful and powerful, obviously it has to be done with caution :) 
Like in that twitter example I can see 3 different font sizes (title/counters, text and button).

skjnldsv:
But it's not hierarchy here, al those elements have the same level of info.
It's fine for titles and subtitles, but I think we should rather emphasis on a different way here :)

skjnldsv:
One thing that would be good here for example, is default title like server, and subtitle at 14px :)

But for the counter, it's too much, 3 different sizes in one location that small, not a good idea I think.

Maybe subtitle + counter at 14px? What do you think @ma12-co ?

ma12-co:
> 3 different sizes in one location that small, not a good idea I think.

I was proposing just 2 font sizes, not 3: one for the text and a slightly smaller one for the counter. 

> One thing that would be good here for example, is default title like server, and subtitle at 14px :)

We could also do this but when checking out the competitors we saw that nobody went up to 15px anywhere in the navigation. Either 14 or 13. Plus, title and subtitle will be of 2 different colors anyway like in the contacts app.

What about everything in the navigation @14px and we adjust the counter's padding? This way we'd still have differentiation between main content and navigation ;)

ChristophWurst:
What concerns me more than the actual change: how's this compatible to Nextcloud server versions? Can we used the next nextcloud-vue with older stable releases? Is this a breaking design change? Would the design of apps on older versions suddenly change?

skjnldsv:
> What concerns me more than the actual change: how's this compatible to Nextcloud server versions? Can we used the next nextcloud-vue with older stable releases? Is this a breaking design change? Would the design of apps on older versions suddenly change?

That's why we should most likely just be compliant to the default font and make sure it is nicely displayed no matter what font size the server is providing :)

jancborchardt:
So, bolding the conversation name would be good here as an adjustment then. :)

Cause for unread state, we still have the counter.

PR #616, opened 2019-10-01
Create ListItem component

ma12-co:
This component will be used for talk's conversations, mail's envelopes and contact's list items. The goal here is to have a unified 'layout component' that you can wrap with your custom one and add your own logic on top of it.

### DIFF
--- a/src/components/ListItem/ListItem.vue
+++ b/src/components/ListItem/ListItem.vue
@@ -1,0 +1,502 @@
+<!--
+  - @copyright Copyright (c) 2021 Marco Ambrosini <marcoambrosini@pm.me>
+  -
+  - @author Marco Ambrosini <marcoambrosini@pm.me>
+  -
+  - @license GNU AGPL version 3 or any later version
+  -
+  - This program is free software: you can redistribute it and/or modify
+  - it under the terms of the GNU Affero General Public License as
+  - published by the Free Software Foundation, either version 3 of the
+  - License, or (at your option) any later version.
+  -
+  - This program is distributed in the hope that it will be useful,
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+-->
+<docs>
+
+# Usage
+```
+<ul>
+	<listItem
+		:title="'Title of the element'"
+		:bold="false"
+		:details="'One hour ago'"
+		:counter-number="44"
+		:counter-highlighted="true">
+		<template #icon>
+			<avatar size="44" user="janedoe" display-name="Jane Doe" />
+		</template>
+		<template #subtitle>
+			In this slot you can put both text and other components such as icons
+		</template>
+		<AppNavigationCounter #counter>
+			7
+		</AppNavigationCounter>
+		<template #actions>
+			<ActionButton>
+				Button one
+			</ActionButton>
+			<ActionButton>
+				Button two
+			</ActionButton>
+			<ActionButton>
+				Button three
+			</ActionButton>
+		</template>
+	</listItem>
+	<listItem
+		:title="'Title of the element'"
+		:bold="false">
+		<template #icon>
+			<avatar size="44" user="janedoe" display-name="Jane Doe" />
+		</template>
+		<template #subtitle>
+			In this slot you can put both text and other components such as icons
+		</template>
+		<AppNavigationCounter #counter>
+			7
+		</AppNavigationCounter>
+		<template #actions>
+			<ActionButton>
+				Button one
+			</ActionButton>
+			<ActionButton>
+				Button two
+			</ActionButton>
+			<ActionButton>
+				Button three
+			</ActionButton>
+		</template>
+	</listItem>
+	<listItem
+		:title="'Title of the element'"
+		:bold="false"
+		:details="'One hour ago'">
+		<template #icon>
+			<avatar size="44" user="janedoe" display-name="Jane Doe" />
+		</template>
+		<template #subtitle>
+			In this slot you can put both text and other components such as icons
+		</template>
+		<AppNavigationCounter #counter>
+			7
+		</AppNavigationCounter>
+	</listItem>
+</ul>
+
+```
+</docs>
+
+<template>
+	<!-- This wrapper can be either a router link or a `<li>` -->
+	<nav-element
+		class="list-item__wrapper"
+		v-bind="navElement">
+		<a
+			:id="anchorId"
+			ref="list-item"
+			:class="{ 'list-item--active' : active }"
+			href="#"
+			class="list-item"
+			:aria-label="linkAriaLabel"
+			@mouseover="handleMouseover"
+			@mouseleave="handleMouseleave"
+			@focus="handleFocus"
+			@blur="handleBlur"
+			@keydown.tab="handleTab"
+			@click="onClick"
+			@keydown.esc="hideActions">
+
+			<!-- @slot This slot is used for the avatar or icon -->
+			<slot name="icon" />
+
+			<!-- Main content -->
+			<div class="list-item-content">
+				<div class="list-item-content__main">
+
+					<!-- First line, title and details -->
+					<div class="line-one"
+						:class="{'line-one--bold': bold}">
+						<span
+							class="line-one__title">
+							{{ title }}
+						</span>
+						<span
+							v-if="hasDetails && !displayActions"
+							class="line-one__details">
+							{{ details }}
+						</span>
+					</div>
+
+					<!-- Second line, subtitle and counter -->
+					<div class="line-two"
+						:class="{'line-one--bold': bold}">
+						<span v-if="hasSubtitle" class="line-two__subtitle">
+							<!-- @slot Slot for the second line of the component -->
+							<slot name="subtitle" />
+						</span>
+
+						<!-- Counter -->
+						<span v-if="!displayActions" class="line-two__counter">
+							<AppNavigationCounter
+								v-if="counterNumber != 0"
+								:highlighted="counterHighlighted">
+								{{ counterNumber }}
+							</AppNavigationCounter>
+						</span>
+					</div>
+				</div>
+
+				<!-- Actions -->
+				<div
+					v-show="displayActions"
+					class="list-item-content__actions"
+					@click.prevent.stop="">
+					<Actions
+						ref="actions"
+						menu-align="right"
+						:aria-label="actionsAriaLabel"
+						@update:open="handleActionsUpdateOpen">
+						<!-- @slot Provide the actions for the right side quick menu -->
+						<slot
+							name="actions" />
+					</Actions>
+				</div>
+			</div>
+		</a>
+	</nav-element>
+</template>
+
+<script>
+import Actions from '../Actions'
+import AppNavigationCounter from '../AppNavigationCounter'
+
+export default {
+	name: 'ListItem',
+
+	components: {
+		Actions,
+		AppNavigationCounter,
+	},
+
+	props: {
+		/**
+		 * The details text displayed in the upper right part of the component
+		 */
+		details: {
+			type: String,
+			default: '',
+		},
+
+		/**
+		 * Title (first line of text)
+		 */
+		title: {
+			type: String,
+			required: true,
+		},
+
+		/**
+		* Pass in `true` if you want the matching behavior to
+		* be non-inclusive: https://router.vuejs.org/api/#exact
+		*/
+		exact: {
+			type: Boolean,
+			default: false,
+		},
+
+		/**
+		* The route for the router link.
+		*/
+		to: {
+			type: [String, Object],
+			default: '',
+		},
+
+		/**
+		 * Id for the <a> element
+		 */
+		anchorId: {
+			type: String,
+			default: '',
+		},
+
+		/**
+		 * Make title and subtitle bold
+		 */
+		bold: {
+			type: Boolean,
+			default: false,
+		},
+
+		/**
+		 * Toggle the active state of the component
+		 */
+		active: {
+			type: Boolean,
+			default: false,
+		},
+
+		/**
+		 * Aria label for the wrapper element
+		 */
+		linkAriaLabel: {
+			type: String,
+			default: '',
+		},
+
+		/**
+		 * Aria label for the actions toggle
+		 */
+		actionsAriaLabel: {
+			type: String,
+			default: '',
+		},
+
+		/**
+		 * If different from from 0 this component will display the
+		 * AppNavigationCounter component
+		 */
+		 counterNumber: {
+			 type: [Number, String],
+			 default: 0,
+		 },
+
+		/**
+		 * Toggles the highlighted state of the counter
+		 */
+		 counterHighlighted: {
+			 type: Boolean,
+			 default: false,
+		 },
+
+	},
+
+	data() {
+		return {
+			hovered: false,
+			focused: false,
+			displayActions: false,
+			menuOpen: false,
+		}
+	},
+
+	computed: {
+
+		hasDetails() {
+			return this.details !== ''
+		},
+
+		hasActions() {
+			return (!!this.$slots.actions)
+		},
+
+		// This is used to decide which outer element type to use
+		// li or router-link
+		navElement() {
+			if (this.to !== '') {
+				return {
+					is: 'router-link',
+					tag: 'li',
+					to: this.to,
+					exact: this.exact,
+				}
+			}
+			return {
+				is: 'li',
+			}
+		},
+
+		// Check if the subtitle slot is populated
+		hasSubtitle() {
+			return !!this.$slots.subtitle
+		},
+	},
+
+	watch: {
+
+		menuOpen(newValue) {
+			// A click outside both the menu and the root element hides the actions again
+			if (!newValue && !this.hovered) {
+				this.displayActions = false
+			}
+		},
+	},
+
+	methods: {
+
+		// forward click event
+		onClick(event) {
+			this.$emit('click', event)
+		},
+
+		handleMouseover() {
+			this.showActions()
+			this.hovered = true
+		},
+
+		showActions() {
+			if (this.hasActions) {
+				this.displayActions = true
+			}
+			this.hovered = false
+		},
+
+		hideActions() {
+			this.displayActions = false
+		},
+
+		/**
+		 * Show actions upon focus
+		 */
+		handleFocus() {
+			this.focused = true
+			this.showActions()
+		},
+
+		handleBlur() {
+			this.focused = false
+		},
+
+		/**
+		 * Hide the actions on mouseleave unless the menu is open
+		 */
+		handleMouseleave() {
+			if (!this.menuOpen) {
+				this.displayActions = false
+			}
+			this.hovered = false
+		},
+
+		/**
+		 * This method checks if the root element of the component is focused and
+		 * if that's the case it focuses the actions button if available
+		 * @param {Event} e the keydown event
+		 */
+		handleTab(e) {
+			if (this.focused && this.hasActions) {
+				e.preventDefault()
+				this.$refs.actions.$refs.menuButton.focus()
+				this.focused = false
+			} else {
+				this.displayActions = false
+				this.$refs.actions.$refs.menuButton.blur()
+			}
+		},
+
+		handleActionsUpdateOpen(e) {
+			this.menuOpen = e
+		},
+	},
+}
+</script>
+
+<style lang="scss" scoped>
+
+.list-item__wrapper{
+	position: relative;
+	width: 100%;
+}
+
+// listItem
+.list-item {
+	position: relative;
+	display: flex;
+	align-items: center;
+	flex: 0 0 auto;
+	justify-content: flex-start;
+	padding: 2px 2px 2px 8px;
+	height: 64px;
+	border-radius: 16px;
+	margin: 2px 0;
+	width: 100%;
+	cursor: pointer;
+	transition: background-color 200ms ease-in-out;
+	list-style: none;
+	&:hover,
+	&:focus  {
+		background-color: var(--color-background-hover);
+	}
+	&--active,
+	&:active,
+	&:active ~ .app-navigation-entry__utils {
+		background-color: var(--color-primary-light);
+	}
+
+	&-content {
+		display: flex;
+		flex: 1 1 auto;
+		justify-content: space-between;
+		padding: 0 8px;
+
+		&__main {
+			flex: 1 1 auto;
+			flex-direction: column;
+			width: 0;
+			margin: auto 0;
+		}
+
+		&__actions {
+			flex: 0 0 auto;
+			align-self: center;
+			justify-content: center;
+
+		}
+	}
+}
+
+.line-one {
+	display: flex;
+	align-items: center;
+	justify-content: space-between;
+	white-space: nowrap;
+	margin: 0 auto;
+	&--bold {
+		font-weight: bold;
+	}
+
+	&__title {
+		overflow: hidden;
+		flex-grow: 1;
+		cursor: pointer;
+		text-overflow: ellipsis;
+		color: var(--color-main-text);
+	}
+
+	&__details {
+		color: var(--color-text-lighter);
+		margin: 0 8px;
+	}
+}
+
+.line-two {
+	display: flex;
+	align-items: flex-start;
+	justify-content: space-between;
+	white-space: nowrap;
+	&--bold {
+		font-weight: bold;
+	}
+
+	&__subtitle {
+		overflow: hidden;
+		flex-grow: 1;
+		padding-right: 4px;
+		cursor: pointer;
+		white-space: nowrap;
+		text-overflow: ellipsis;
+		color: var(--color-text-lighter);
+	}
+
+	&__counter {
+		margin: 2px 4px 0 0;
+	}
+}
+
+</style>
--- a/src/components/index.js
+++ b/src/components/index.js
@@ -63,6 +63,7 @@ import PopoverMenu from './PopoverMenu'
 import RichContenteditable from './RichContenteditable'
 import SettingsSection from './SettingsSection'
 import UserBubble from './UserBubble'
+import ListItem from './ListItem'
 
 export {
 	ActionButton,
@@ -98,6 +99,7 @@ export {
 	Content,
 	DatetimePicker,
 	EmptyContent,
+	ListItem,
 	ListItemIcon,
 	Modal,
 	Multiselect,
--- a/src/components/AppNavigation/AppNavigation.vue
+++ b/src/components/AppNavigation/AppNavigation.vue
@@ -171,7 +171,7 @@ export default {
 	&__list {
 		position: relative;
 		height: 100%;
-		width: inherit;
+		width: 100%;
 		overflow-x: hidden;
 		overflow-y: auto;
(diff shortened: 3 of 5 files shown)


src/components/AppContentListItem/AppContentListItem.vue
```diff
@@ -0,0 +1,88 @@
...
+		class="app-content-list-item">
+		<!-- default slot for avatar or icon -->
+		<slot />
+		<div
+			class="app-content-list-item-line-one">
+			{{ title }}
+		</div>
+		<div
+			class="app-content-list-item-line-two">
+			{{ subtitle }}
```
ChristophWurst: what if there is none? Should we make the line-two optional?

src/components/AppContentListItem/AppContentListItem.vue
```diff
@@ -0,0 +1,88 @@
...
+		 * The details text displayed in the upper right part
+		 */
+		details: {
+			type: String,
+			default: ''
+		}
+	},
+	computed: {
+		hasDetails() {
+			if (this.details !== '' && !this.$slots.counter) {
```
ChristophWurst:
```suggestion
			return this.details !== '' && !this.$slots.counter
```
:wink:

src/components/AppContentListItem/AppContentListItem.vue
ChristophWurst: IMO it should be a slot. Otherwise Mail can't use it https://github.com/nextcloud/mail/blob/c1ef17b8b631bf14ef5a1bc61f2e330f45907b03/src/components/Envelope.vue#L19-L26

ChristophWurst (dismissed):

src/components/AppContentListItem/AppContentListItem.vue
ma12-co: > IMO it should be a slot. Otherwise Mail can't use it

Wait, are you talking about title and subtitle? Can't you do `:title="addresses"` and `:subtitle="data.subject"`? For the icon and actions there's indeed a slot.

ChristophWurst: Yes and no. The subtitle, for example, is italic for unread messages. How can I achieve that with a string prop?

skjnldsv: @ChristophWurst :drum: :drum:  CSS :grin:

ma12-co: Got it! but how many exceptions of this kind do we have across these apps? If it's just a few of them they could be just more props.

ChristophWurst: Or a slot :wink:

skjnldsv: A slot would be the easiest way  @ma12-co :)
Allowing text and whatever in it :)

ma12-co: >Allowing text and whatever in it :)

But isn't this supposed to standardize things across the apps? Why would we want to allow anything other than text in title and subtitle fields???

skjnldsv: Because there is a fine line between too much restriction and too much freedom :)
I'd say it's fine to allow a few icons here! Let's not recommend it and make sure only one line is displayed !

src/components/AppContentListItem/AppContentListItem.vue
```diff
@@ -114,7 +111,22 @@ export default {
 			return (this.details !== '' && !this.$slots.counter)
 		},
 		hasActions() {
-			return (this.$slots.actions)
+			return (!!this.$slots.actions)
+		},
+		// This is used to decide which outer element type to use
+		// li or router-link
+		navElement() {
+			if (this.to !== '') {
```
skjnldsv: :+1:

jancborchardt:
@ma12-co this will tackle the first 2 issues at https://github.com/nextcloud/mail/issues/2110, right?

ma12-co:
@jancborchardt it will, but this is still very basic and it will need quite a bit of work to support all the features required by mail, contacts and talk.

ma12-co:
Finally coming back to this. 
To give a bit of an update: there has been some discussion around completely removing the actions from the conversation component in talk, since all the settings for each conversation are now in a separate dialog and not all over the place anymore. We thought that that was a bit too much and that the actions there were handy, albeit not mandatory anymore. 
Finally we came up with a solution that de-clutters the UI and keeps the actions in the component but displays them only on hover and focus. 
The drawback to this is obviously the discover-ability of the button. In talk we can afford this because we treat this menu as a secondary, shortcut-like type of thing: all the actions contained in it are also present in the main settings dialog, which is discover-able. 
This is the proposed behavior:

https://user-images.githubusercontent.com/26852655/108831534-30eee580-75c2-11eb-8535-3c44dab33c53.mp4

So while I'm pretty confident that this will work for talk, I'm not sure about the other apps that use the two lines layout on the right side (e.g. mail). I'd be happy to see the tendency of unifying actions in the main content and stop showing these columns of 3 dot buttons in the navigation, but I'm not sure that's a suitable solution for every app.

Screenshot from 2021-02-23 10-45-25


So I'd like to know from you @jancborchardt @ChristophWurst  and @skjnldsv what are your thoughts on this. Should we have a toggle for these 2 types of layout? Should we have only one type?

skjnldsv:
I like the design! Looks really nice. As usual, the show-on-hover is strongly advised against for accessibility purpose. It's non-discoverable and incompatible with touch devices. :/

skjnldsv:
I like the design! Looks really nice.

> The drawback to this is obviously the discover-ability of the button. In talk we can afford this because we treat this menu as a secondary, shortcut-like type of thing: all the actions contained in it are also present in the main settings dialog, which is discover-able.

As usual, the show-on-hover is strongly advised against for accessibility purpose. It's non-discoverable and incompatible with touch devices. :/
The issue is not the keyboards, if it's focused, then it can be shown (granted, it's still bad as you have something that suddenly appear)

I'm afraid I can't agree with anything else that would always show the menu by default.
https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html

skjnldsv:
image

Because there is a lot more spacing now, I might suggest we add a bit more width to the Navigation section, so that showing the three-dot menu becomes less invasive.

ma12-co:
> I'm afraid I can't agree with anything else that would always show the menu by default.
https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html

If we commit to this to the letter, by extension all talk's messages should show the 3 dot menu permanently. So we'd have yet another column of those. I we take the spirit over the letter, I think there's room for discussion on when we might want do the hover/focus trick.

skjnldsv:
> If we commit to this to the letter, by extension all talk's messages should show the 3 dot menus permanently

Yep, I've said it numerous time too :shrug:

ma12-co:
> Yep, I've said it numerous time too **shrug**

Well it seems that we're about to have 2 buttons per message now, so that would be even more messy. 
To me a UI with 50+ buttons on display on a normal screen size is not something we should aim at. We'd probably meet those criteria, but it would create so much cognitive load on the user that it would be less usable and thus less accessible for everybody.

@jancborchardt can you please share your take here?

skjnldsv:
> To me a UI with 50+ buttons on display on a normal screen size is not something we should aim at. We'd probably meet those criteria, but it would create so much cognitive load on the user that it would be less usable and thus less accessible for everybody.

Not disagreeing here the design won't be as nifty. I don't make the accessibility guidelines nor am I directly concerned by this. But I'm sure that if any of us had accessibility issues or were using voice command all day long, we wouldn't not have this "debate".

jancborchardt:
So in the example of the menu for the individual Talk messages: This should show on hover or focus of message, but also on tap of the message → problem solved there for touchscreens.

For the navigation menus it's a bit different since of course it opens something on tap and changes the view.
So what we could do is to make sure that the actions in the menu are accessible exactly the same in a menu inside the content. Basically like we do for the Android Files app, where you can long-press on a file, or open it and go to the 3-dot menu in the header.

What do you think about those 2 fixes?

ma12-co:
Upon further reading of
> https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html
It seems that we'd only need to add the ability to dismiss the triggered hover state by pressing esc to meet the success criteria.. Or am I missing something?

skjnldsv:
> What do you think about those 2 fixes?

I'm honestly fine with any solution that take everything into account :)
I'm loving Marco's design, but I'm also loving people being able to use our software lol
It's annoying to have to find a middleground of course, and I would have loved to have a better solution to give. I'm just raising concerns here so it's not skipped :hugs:

skjnldsv:
@ma12-co maybe there is a way to detect touch screens?
If so always show on touch devices?

That doesn't cover the keyboard-only-no-mouse usage. But if the menu is focusable and appear on focus, that might be ok for keyboard with//or keyboards? :thinking:

jancborchardt:
> @ma12-co maybe there is a way to detect touch screens?
> If so always show on touch devices?

Not sure about that, cause:
- There are devices which have a touchscreen but people often use them with keyboard + mouse.
- The default UI will look different on different devices, which is not really normal (except when it’s different sizes)

skjnldsv:
> * There are devices which have a touchscreen but people often use them with keyboard + mouse.
> * The default UI will look different on different devices, which is not really normal (except when it’s different sizes)

Both fair points, you're right! :thinking:

ma12-co:
@jancborchardt and @juliushaertl  could you double check that we'd be ok accessibility wise with this?

ma12-co:
Ok, so after a call with @jancborchardt we decided that it's ok to go ahead with the hoverable thing since it meets the success criterion for WCAG 2 Level AA accessibility. We should all keep in mind that all the actions that are to be found in this menu should be present in the main content for the given navigation element.

jancborchardt:
@ma12-co hoverable and focusable though. :) The actions being also available in the main content is to take care that it works on touch-only interfaces.

src/components/AppContentListItem/AppContentListItem.vue
```diff
@@ -0,0 +1,366 @@
...
+
+		// This is used to decide which outer element type to use
+		// li or router-link
+		navElement() {
+			if (this.to !== '') {
+				return {
+					is: 'router-link',
+					tag: 'li',
+					to: this.to,
+					exact: this.exactRoute,
```
skjnldsv, line 170:
```suggestion
					exact: this.exact,
```

src/components/AppContentListItem/AppContentListItem.vue
```diff
@@ -0,0 +1,366 @@
...
+	border-radius: 10px;
+
+	&:not(.app-navigation-entry__counter--highlighted) {
+		background-color: var(--color-background-darker);
+	}
+
+	span {
+		padding: 2px 6px;
+	}
+}
```
skjnldsv, line 360: It feels a bit weird, no?
To have custom css for an appnavigationCounter within an appCOntentListItem ?

src/components/AppContentListItem/AppContentListItem.vue
```diff
@@ -0,0 +1,366 @@
...
+			default: '',
+		},
+
+		/**
+		 * Bold title and subtitle
+		 */
+		bold: {
+			type: Boolean,
+			default: false,
+		},
```
skjnldsv, line 136: Used for unread rooms I assume?
Shall this be managed by the app ? Or do we want a standard for this? I'm conflicted as this seems like a very specific use case? @ChristophWurst do you need this for Mail as well? :)

skjnldsv (changes requested):
See comments/questions :)

src/components/AppContentListItem/AppContentListItem.vue
ma12-co: Yeah, you're right. Should we add a toggle for this style in AppNavigationCounter?

src/components/AppContentListItem/AppContentListItem.vue
ma12-co: It is used in mail too, yep!

src/components/AppContentListItem/AppContentListItem.vue
skjnldsv: I don't know, can't we just change the style of the counter?

ma12-co: I thought that that counter without bubble was needed somewhere, but sure!

ma12-co:
Current state

https://user-images.githubusercontent.com/26852655/111650600-5a83e100-87fd-11eb-8c4f-6c852ade4ddc.mp4

src/components/AppContentListItem/AppContentListItem.vue
```diff
@@ -0,0 +1,418 @@
...
+-->
+<docs>
+
+# Usage
+```
+<AppContentListItem
+		:title="'Title of the element'"
+		:bold="false"
+		:details="'One hour ago'">
+		<template v-slot:icon>
```
korelstar, line 29:
```suggestion
		<template #icon>
```

src/components/AppContentListItem/AppContentListItem.vue
```diff
@@ -0,0 +1,418 @@
...
+			<avatar size="44" user="janedoe" display-name="Jane Doe" />
+		</template>
+		<template v-slot:subtitle>
```
korelstar, line 32:
```suggestion
		<template #subtitle>
```

src/components/AppContentListItem/AppContentListItem.vue
```diff
@@ -0,0 +1,418 @@
...
+			In this slot you can put both text and other components such as icons
+		</template>
+		<AppNavigationCounter
+			slot="counter">
```
korelstar, line 36:
```suggestion
		<AppNavigationCounter #counter>
```

src/components/AppContentListItem/AppContentListItem.vue
```diff
@@ -0,0 +1,418 @@
...
+			7
+		</AppNavigationCounter>
+		<template slot="actions">
```
korelstar, line 39:
```suggestion
		<template #actions>
```

korelstar (commented):
Just a small suggestion for harmonization.

korelstar:
Just realized that this component has no edit in place functionality like `AppNavigationListItem`. Since the plan for the notes app is to move the notes list from `AppNavigationListItem` to `AppContentListItem` I would need this functionality in order to be able to rename notes.

Would this be conform design-wise? Can this be added?

src/components/AppContentListItem/AppContentListItem.vue
```diff
@@ -0,0 +1,418 @@
...
+			}
+		},
+
+		conversationLinkAriaLabel() {
+			return t('spreed', 'Conversation "{conversationName}"', { conversationName: this.title })
+		},
+
+		conversationSettingsAriaLabel() {
+			return t('spreed', 'Settings for conversation "{conversationName}"', { conversationName: this.title })
+		},
```
korelstar, line 227: Shouldn't this be generic?

korelstar (commented):
Another question: How can I close the actions menu from outside (e.g. after the action is executed)? As far as I can see, there are data attributes `menuOpen` and `displayActions`, but there is no prop for this.

ma12-co:
> How can I close the actions menu from outside (e.g. after the action is executed)? As far as I can see, there are data attributes menuOpen and displayActions, but there is no prop for this. 

Several actions have  the `closeAfterClick` prop. What 's your use case here?

korelstar:
> Several actions have the `closeAfterClick` prop. What 's your use case here?

- Most of my actions are changing something on the server. Therefore, the action menu should close only after a successful server response (in the meanwhile the `loading` state is active).
- As far as I can see, `ActionInput` does not have this prop.

ma12-co:
> I would need this functionality in order to be able to rename notes.

Wouldn't it be better avoid this and do the renaming directly into the AppContent? Each note could have a more explicit title bar on the top and a click there could just trigger an editing state for the bar maybe?

src/components/AppContentListItem/AppContentListItem.vue
ma12-co: Pr here @skjnldsv https://github.com/nextcloud/nextcloud-vue/pull/1769

korelstar:
> > I would need this functionality in order to be able to rename notes.
> 
> Wouldn't it be better avoid this and do the renaming directly into the AppContent? Each note could have a more explicit title bar on the top and a click there could just trigger an editing state for the bar maybe?
>
> I just feel like making this component editable/open_input would complicate the design too much

The goal of the notes app is to provide distraction-free notes taking. Due to this design goal, we try to avoid all GUI elements that may disturb people from concentrating on writing. This was promoted especially by @jancborchardt . I'm not sure if we really should drop this goal, just because the implementation will be a bit more complicated.

ma12-co:
@korelstar In general I personally think we should avoid these popover based editing flows but that's for another discussion. I would focus for now on complying with the requirements of the apps that already use this component and get this in as asap. Let's discuss further features in a separate ticket, and when @jancborchardt is back too :)

korelstar:
@ma12-co Okay.

Btw.: The documentation is missing the general usage of the `AppContentList*` components. I.e., it would be helpful to provide a description of the intended hierarchy. E.g., I'm not sure if the following usage is correct:

```
<Content>
	<AppNavigation>
		<AppNavigationItem />
	</AppNavigation>
	<AppContent>
		<AppContentList>
			<AppContentListItem />
		</AppContentList>
		<AppContentDetails>Main App content</AppContentDetails>
	</AppContent>
</Content>
```

ma12-co:
> Btw.: The documentation is missing the general usage of the AppContentList* components. I.e., it would be helpful to provide a description of the intended hierarchy. E.g., I'm not sure if the following usage is correct

Thinking about this, in talk we really use this component in other places, like the participant selection for a new group conversation, or the participant display in the right sidebar. 
I would just keep it as a general purpose tall, avatar/text/actions/details component. 
Because of that, we could just call it listItem maybe? 
cc @skjnldsv

skjnldsv:
> I would just keep it as a general purpose tall

I would as well, yes.
Same as the [ListItemIcon](https://nextcloud-vue-components.netlify.app/#/Components/ListItemIcon)

src/components/index.js
```diff
@@ -79,7 +79,7 @@ export {
 	AppContent,
 	AppContentDetails,
 	AppContentList,
-	AppContentListItem,
+	ListItem,
```
skjnldsv, line 82: No alphabetical anymore? :stuck_out_tongue:

ma12-co:
current state, I've been testing with talk and I think this is ready for review

https://user-images.githubusercontent.com/26852655/112837908-604ab380-9094-11eb-83b4-b002258bb19d.mp4

src/components/ListItem/ListItem.vue
```diff
@@ -0,0 +1,457 @@
...
+				Button three
+			</ActionButton>
+		</template>
+	</listItem>
+```
+</docs>
+
+<template>
+	<nav-element
+		class="list-item_wrapper"
```
skjnldsv, line 58:
```suggestion
		class="list-item__wrapper"
```

src/components/ListItem/ListItem.vue
```diff
@@ -0,0 +1,457 @@
...
+		:counter-number="44"
+		:counter-highlighted="true">
+		<template #icon>
+			<avatar size="44" user="janedoe" display-name="Jane Doe" />
+		</template>
+		<template v-slot:subtitle>
+			In this slot you can put both text and other components such as icons
+		</template>
+		<AppNavigationCounter
+			#counter">
```
skjnldsv, line 38:
```suggestion
		<AppNavigationCounter #counter>
```

src/components/ListItem/ListItem.vue
```diff
@@ -0,0 +1,457 @@
...
+</docs>
+
+<template>
+	<nav-element
+		class="list-item_wrapper"
+		v-bind="navElement">
+		<a
+			:id="anchorId"
+			ref="list-item"
+			:class="{ 'active' : active }"
```
skjnldsv, line 63:
```suggestion
			:class="{ 'list-item__wrapper--active' : active }"
```

src/components/ListItem/ListItem.vue
```diff
@@ -0,0 +1,457 @@
...
+
+		handleMouseleave() {
+			if (!this.menuOpen) {
+				this.displayActions = false
+			}
+		},
+
+		handleTab(e) {
+			if (this.focused && this.hasActions) {
+				console.debug('preventdefault')
```
skjnldsv, line 319:
```suggestion
```

src/components/ListItem/ListItem.vue
```diff
@@ -0,0 +1,457 @@
...
+				e.preventDefault()
+				this.$refs.actions.$refs.menuButton.focus()
+				this.focused = false
+			} else {
+				this.displayActions = false
+				this.$refs.actions.$refs.menuButton.blur()
+			}
+		},
```
skjnldsv, line 327: Please add jsDocs to your methods, I have no idea why we're handling the tab here?

You're trying to only show the actions if there was a focus on the item before? Like requiring a double tab? If so I would prefer you to watch the focus event and not the tab. You will cover all the mouse/keyboard use-cases in one go :wink:

src/components/ListItem/ListItem.vue
```diff
@@ -0,0 +1,457 @@
...
+		 counterHighlighted: {
+			 type: Boolean,
+			 default: false,
+		 },
+
+	},
+
+	data() {
+		return {
+			hovered: false,
```
skjnldsv, line 231: `hovered` doesn't seems to be used anywhere else?

src/components/ListItem/ListItem.vue
```diff
@@ -0,0 +1,457 @@
...
+			return !!this.$slots.subtitle
+		},
+	},
+
+	watch: {
+
+		menuOpen(newValue) {
+			console.debug('menuopen', newValue)
+			// A click outside both the menu and the root element hides the actions again
+			if (!newValue && !this.hovered) {
```
skjnldsv, line 275: `this.hovered` is always false

src/components/ListItem/ListItem.vue
```diff
@@ -0,0 +1,457 @@
...
+		},
+
+		/**
+		 * If different from from 0 this component will display the
+		 * AppNavigationCounter component
+		 */
+		 counterNumber: {
+			 type: Number,
+			 default: 0,
+		 },
```
skjnldsv, line 217: Could be a string as well ?

src/components/ListItem/ListItem.vue
```diff
@@ -0,0 +1,457 @@
...
+			href="#"
+			class="list-item"
+			:aria-label="linkAriaLabel"
+			@mouseover="handleHover"
+			@focus="handleFocus"
+			@blur="handleBlur"
+			@mouseleave="handleMouseleave"
+			@keydown.tab="handleTab"
+			@click="onClick"
+			@keydown.esc="displayActions = false">
```
skjnldsv, line 73: Not fond of direct prop edit in here :)

src/components/ListItem/ListItem.vue
```diff
@@ -0,0 +1,457 @@
...
+			if (!newValue && !this.hovered) {
+				this.displayActions = false
+			}
+		},
+	},
+
+	mounted() {
+		console.debug('this.route', this.route, 'this.to', this.to)
+	},
+
```
skjnldsv, line 284:
```suggestion
```

src/components/ListItem/ListItem.vue
```diff
@@ -0,0 +1,457 @@
...
+		// Check if the subtitle slot is populated
+		hasSubtitle() {
+			return !!this.$slots.subtitle
+		},
+	},
+
+	watch: {
+
+		menuOpen(newValue) {
+			console.debug('menuopen', newValue)
```
skjnldsv, line 273:
```suggestion
```

src/components/ListItem/ListItem.vue
```diff
@@ -0,0 +1,457 @@
...
+						</span>
+						<span
+							v-if="hasDetails && !displayActions"
+							class="list-item-content__line-one__details">
+							{{ details }}
+						</span>
+					</div>
+					<div class="list-item-content__line-two"
+						:class="{'bold': bold}">
+						<span v-if="hasSubtitle" class="list-item-content__line-two__subtitle">
```
skjnldsv, line 92: Not really BEM compliant :thinking:

src/components/ListItem/ListItem.vue
```diff
@@ -0,0 +1,457 @@
...
+			@mouseover="handleHover"
+			@focus="handleFocus"
+			@blur="handleBlur"
+			@mouseleave="handleMouseleave"
+			@keydown.tab="handleTab"
+			@click="onClick"
+			@keydown.esc="displayActions = false">
+			<!-- @slot This slot is used for the avatar or icon -->
+			<slot name="icon" />
+			<div class="list-item-content">
```
skjnldsv, line 76:
```suggestion
			<slot name="icon" />

			<div class="list-item-content">
```

You can space things a lot more :)
Should help making the template more clear

As well as using  Comments to describe what those BEM wrappers are or does :wink:

skjnldsv (changes requested):

src/components/ListItem/ListItem.vue
```diff
@@ -0,0 +1,457 @@
...
+	.actions {
+		position: absolute;
+		top: 4px;
+		right: 12px;
+	}
+}
+
+// listItem
+.list-item {
+	position: relative;
```
skjnldsv, line 350:
```suggestion
	position: relative;
	list-style: none;
```
	
image

skjnldsv (changes requested):

src/components/ListItem/ListItem.vue
ma12-co: This is used in the `menuOpen` watcher

src/components/ListItem/ListItem.vue
ma12-co: So the  problem I encountered is I have to display the actions even after the root element is blurred. What I need is: 

* User tabs once
-> the root element is highlighted and the actions are displayed
* User tabs again
-> the main element is blurred and the actions are focused

If I'd only handle the focus event and toggle the actions on and off, the actions would disappear upon pressing tab and would  not be focusable.

src/components/ListItem/ListItem.vue
```diff
@@ -0,0 +1,454 @@
...
+						<span
+							v-if="hasDetails && !displayActions"
+							class="line-one__details">
+							{{ details }}
+						</span>
+					</div>
+
+					<!-- Second line, subtitle and counter -->
+					<div class="line-two"
+						:class="{'bold': bold}">
```
skjnldsv, line 98: BEM

src/components/ListItem/ListItem.vue
```diff
@@ -0,0 +1,454 @@
...
+
+			<!-- Main content -->
+			<div class="list-item-content">
+				<div class="list-item-content__main">
+
+					<!-- First line, title and details -->
+					<div class="line-one">
+						<span
+							class="line-one__title"
+							:class="{'bold': bold}">
```
skjnldsv, line 86: BEM

src/components/ListItem/ListItem.vue
```diff
@@ -0,0 +1,454 @@
...
+<style lang="scss" scoped>
+
+.list-item__wrapper{
+	position: relative;
+	width: 100%;
+	.actions {
+		position: absolute;
+		top: 4px;
+		right: 12px;
+	}
```
skjnldsv, line 362: Leftover

skjnldsv (approved):

skjnldsv:
Don't forget to rebase and squash though! :warning:

nickvergessen (approved):
🐘

PR #619, opened 2019-10-02
Make "more icon" of actions customizable

### LINKED ISSUE #618: Allow to provide custom icon to be shown if more than one action inside Actions
Currently, as soon as you have multiple `Action` components inside an `Actions` component, it shows the three-dot menu icon. There are scenarios ([like here in Jan's mockup](https://github.com/nextcloud/calendar/issues/1021#issue-393233730)) where you would like to replace that icon with another one.

@skjnldsv @ChristophWurst Are you fine with that change? 
In that case i will send a pull-request :)

georgehrke:
Default:
16C35659-0F6C-4319-AA8B-EDBB846C7807

Custom icon:
98A0FCCB-088B-45FD-B29E-1D19145DB3E2

fixes #618

### DIFF
--- a/src/components/Actions/Actions.vue
+++ b/src/components/Actions/Actions.vue
@@ -72,6 +72,7 @@ https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-action
 		<!-- If more than one action, create a popovermenu -->
 		<a v-click-outside="closeMenu"
 			class="icon action-item__menutoggle"
+			:class="defaultIcon"
 			href="#" aria-haspopup="true"
 			:aria-controls="randomId"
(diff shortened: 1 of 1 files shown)


raimund-schluessler:
Not adding the `action-item__menutoggle--default-icon` class somehow changes the lineheight if used in a table row. I guess it has something to do with `::before { content: "";}`. I will try to investigate.

skjnldsv:
Argh! Did not saw that! :)
Thanks for the awesome help @raimund-schluessler !

I love our work on the vue components! Thanks for being here! :heart:

raimund-schluessler:
PR to fix is in #625

Issue #615, opened 2019-10-01
Close Actions after Action was clicked


raimund-schluessler:
I have the following component creating a dropdown with popovermenu items:
```vue
<template>
	<div class="menu-icon">
		<div v-click-outside="close" class="header-icon icon-more" @click="toggle" />
		<div :class="{ 'open': open }" class="popovermenu">
			<popover-menu :menu="menu" />
		</div>
	</div>
</template>

<script>
import PopoverMenu from 'nextcloud-vue/dist/Components/PopoverMenu'
import ClickOutside from 'vue-click-outside'

export default {
	components: {
		PopoverMenu,
	},
	directives: {
		ClickOutside,
	},
	props: {
		menu: {
			type: Array,
			required: true,
		}
	},
	data: function() {
		return {
			open: false,
		}
	},
	methods: {
		toggle: function() {
			this.open = !this.open
		},
		close() {
			this.open = false
		},
	}
}
</script>
```

I want to close the dropdown when an item is clicked. A simple click handler on the popover-menu doesn't work, because the click is prevented in popover-menu. The alternative could be to implement a two-way sync for `this.open` similar to https://github.com/nextcloud/nextcloud-vue/pull/278, and closing the menu from the parent component after action is handled, but this gets messy quickly when you have multiple dropdowns in the parent component (you need to track which dropdown is opened).

I propose to emit a click event like `this.$emit('clickedAction)` in the popover-menu item which the `Dropdown` component can process if necessary using `@clickedAction="close".

skjnldsv:
Argh, we deprecated the popover menu. In favour of the actions component. Could you migrate to it?

raimund-schluessler:
> Argh, we deprecated the popover menu. In favour of the actions component. Could you migrate to it?

Oh, I wasn't aware of this. I will have a look.

skjnldsv:
@raimund-schluessler ping me if you need help! But our doc is really awesome now :)
https://nextcloud-vue-components.netlify.com/

raimund-schluessler:
@skjnldsv Well, the Actions component has the same problem. When I want to close the Actions menu after an Action was clicked, I need to do this from the parent component via the `open` property. As said above, this is cumbersome when you have multiple Actions in one component.

Having it self-contained inside the Actions component (at least optional) would be a huge improvement in my opinion.

skjnldsv:
> When I want to close the Actions menu after an Action was clicked

Could you give me a use case on why this is required?
[...]

PR #622, opened 2019-10-02
Allow to close the actions menu after click

raimund-schluessler:
Implements #615

Since it is apparently not possible to listen for an emitted event on a slot (https://github.com/vuejs/vue/issues/4332), I just call the `closeMenu` method of the  `Actions` component directly from the `ActionButton`. I guess this is ok, since an `ActionButton` will always have an `Actions` component as parent.

Also, it is working as expected 😉

### DIFF
--- a/src/mixins/actionText.js
+++ b/src/mixins/actionText.js
@@ -21,6 +21,7 @@
  *
  */
 import actionGlobal from './actionGlobal'
+import GetParent from 'Utils/GetParent'
 
 export default {
 	mixins: [actionGlobal],
@@ -68,7 +69,10 @@ export default {
 			this.$emit('click', event)
 
 			if (this.closeAfterClick) {
-				this.$parent.closeMenu()
+				const parent = GetParent(this, 'Actions')
+				if (parent) {
+					parent.closeMenu()
+				}
 			}
 		}
 	}
--- a/src/mixins/actionText.js
+++ b/src/mixins/actionText.js
@@ -39,6 +39,13 @@ export default {
 		title: {
 			type: String,
 			default: ''
+		},
+		/**
+		 * Whether we close the Actions menu after the click
+		 */
+		closeAfterClick: {
+			type: Boolean,
+			default: false
 		}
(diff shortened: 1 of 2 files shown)


src/mixins/actionText.js
```diff
@@ -59,6 +66,10 @@ export default {
 			 * @type {Event}
 			 */
 			this.$emit('click', event)
+
+			if (this.closeAfterClick) {
+				this.$parent.closeMenu()
```
skjnldsv: Will not work if the component is wrapped :thinking: 
I would use the internal Get parent function https://github.com/nextcloud/nextcloud-vue/blob/master/src/utils/GetParent.js

raimund-schluessler: What do you mean with wrapped? Could you give a short example so I understand and can check if getParent fixes the issue?

raimund-schluessler: I think I get it. Will adjust the PR.

skjnldsv: Yes! :)
If you create an api allowing external apps to register new actions into a menu (like in server in  the public sharing menu, or in the users management, in the user row end of line menu) external apps will provide their component: like ActionButton. But the app will want to have their own actions handlers!

So it will look like 
```vue
<CustomAction>
	<ActionButton @click="dosomething" />
</CustomAction>
```
And then 
```vue
<Actions>
	<component :is="CustomAction" />
</Actions>
```

raimund-schluessler: Fixed it.

src/mixins/actionText.js
```diff
@@ -59,6 +67,13 @@ export default {
 			 * @type {Event}
 			 */
 			this.$emit('click', event)
+
+			if (this.closeAfterClick) {
+				const parent = GetParent(this, 'Actions')
+				if (parent) {
```
skjnldsv:
```suggestion
				if (parent && parent.closeMenu) {
```

Never too careful ;)

PR #625, opened 2019-10-03
Fix different height of actions with custom icon

raimund-schluessler:
With #619 the actions menu causes a parent element to have a larger height than without a custom icon (custom icon on second Actions):
Screenshot_2019-10-03 Inventar - Nextcloud

This PR fixes this issue:
Screenshot_2019-10-03 Inventar - Nextcloud(1)

### DIFF
--- a/src/components/Actions/Actions.vue
+++ b/src/components/Actions/Actions.vue
@@ -476,6 +476,10 @@ $arrow-margin: ($clickable-area - 2 * $arrow-width)  / 2;
 
 		opacity: $opacity_normal;
 
+		&:before {
+			content: '';
+		}
+
 		&--default-icon {
 			font-size: $icon-size;
 			@include iconfont('more');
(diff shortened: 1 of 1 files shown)


skjnldsv:
We should probabyl add an example on the docs for this :)

skjnldsv:
> Specify the icon to be shown as placeholder when more than one action is inside the actions component

Also the prop comment is maybe not clear that the default is the three dot menu.
We should maybe not encourage people to change the icon, but in very specific cases only. Could you take care of this as well? :)
What do you think? cc @georgehrke

raimund-schluessler:
Added an example and hopefully improved the docs.

src/components/Actions/Actions.vue
```diff
@@ -42,6 +42,16 @@ https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-action
 	<ActionLink icon="icon-external" title="Link" href="https://nextcloud.com" />
 </Actions>
 ```
+
+### Multiple actions with custom icon
+
+```
+<Actions :default-icon="'icon-edit'">
```
skjnldsv:
```suggestion
<Actions default-icon="icon-edit">
```

PR #650, opened 2019-10-14
add SettingsInputText compontents

GretaD:


src/components/SettingsInputText/SettingsInputText.vue
```diff
@@ -0,0 +1,138 @@
...
+	$input-margin: 4px;
+
+	.input-wrapper {
+		display: flex;
+		flex-wrap: wrap;
+		width: 100%;
+		max-width: 400px;
+	}
+
+	&__label {
```
georgehrke: You can't use `&` in the root of the scss definition.
It always has to be nested inside another definition.

GretaD: of course :)

src/components/SettingsInputText/SettingsInputText.vue
```diff
@@ -0,0 +1,162 @@
...
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+
+<template>
+	<form @submit.prevent="">
+		<div class="input-wrapper">
+			<label :for="id">{{ label }}</label>
+			<input :id="id" type="text" :value="inputVal"
+				:disabled="disabled" @input="$emit('input', $event.target.value)">
```
skjnldsv, line 29: Please no inline $emit, it's cleaner in a dedicated method

src/components/SettingsInputText/SettingsInputText.vue
```diff
@@ -0,0 +1,162 @@
...
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+
+<template>
+	<form @submit.prevent="">
```
skjnldsv, line 25: What if I want to listen to the form? ;)
Maybe re-emit it?
Not sure people would use it, but it should be cleaner than an empty function here :thinking:

src/components/SettingsInputText/SettingsInputText.vue
```diff
@@ -0,0 +1,162 @@
...
+		<div class="input-wrapper">
+			<label :for="id">{{ label }}</label>
+			<input :id="id" type="text" :value="inputVal"
+				:disabled="disabled" @input="$emit('input', $event.target.value)">
+			<input type="submit" class="icon-confirm" value=""
+				@click="$emit('update', inputVal)">
+		</div>
+		<p v-if="hint !== ''" class="hint">
```
skjnldsv, line 33: so if hint is an Object, it will be displayed here ;)
What you want is just `v-if="hint"`

skjnldsv, line 33: so if hint is an Object, it will be displayed here ;)
What you want is just `v-if="hint"`

src/components/SettingsInputText/SettingsInputText.vue
```diff
@@ -0,0 +1,162 @@
...
+			{{ hint }}
```
skjnldsv, line 34: Btw, maybe it would be cleaner to have the hint as the slot?
```vue
<SettingsInputText
	label="Enter password"
	v-model="password">
		Please enter a password to execute those administrative tasks
</SettingsInputText>
```

src/components/SettingsInputText/SettingsInputText.vue
```diff
@@ -0,0 +1,162 @@
...
+  -
+  -->
+
+<template>
+	<form @submit.prevent="">
+		<div class="input-wrapper">
+			<label :for="id">{{ label }}</label>
+			<input :id="id" type="text" :value="inputVal"
+				:disabled="disabled" @input="$emit('input', $event.target.value)">
+			<input type="submit" class="icon-confirm" value=""
```
skjnldsv, line 30: Missing some aria data I think :thinking:

src/components/SettingsInputText/SettingsInputText.vue
```diff
@@ -0,0 +1,162 @@
...
+			type: Boolean,
+			default: false
+		},
+
+		/**
+		 * id attribute of the select group element
+		 */
+		id: {
+			type: String,
+			default: () => 'action-' + GenRandomId(),
```
skjnldsv, line 81: not an action

src/components/SettingsInputText/SettingsInputText.vue
```diff
@@ -0,0 +1,162 @@
...
+	},
+	data() {
+		return {
+			inputVal: this.value
+		}
+	},
+	computed: {
+		/**
+		* @returns {string}
+		* @constructor
```
skjnldsv, line 93:
```suggestion
```

src/components/SettingsInputText/SettingsInputText.vue
```diff
@@ -0,0 +1,162 @@
...
+		*/
+		GenRandomId() {
+			return 'settings-input-text-' + this.GenRandomId
+		}
+	},
+	watch: {
+		value(newVal) {
+			this.inputVal = this.value
```
skjnldsv, line 101: Why keeping a local data here?
The input value should be enough, I don't thing there is a need to keep track of this here with a watcher :thinking:

src/components/SettingsInputText/SettingsInputText.vue
```diff
@@ -0,0 +1,162 @@
...
+		}
+	},
+	beforeCreate: function() {
+		this.GenRandomId = this.GenRandomId.toString()
+		GenRandomId += 1
+	}
```
skjnldsv, line 107: Leftover?
```suggestion
```

src/components/SettingsInputText/index.js
```diff
@@ -0,0 +1,25 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+import SettingsInputText from './SettingsInputText'
+
+export default SettingsInputText
+export { SettingsInputText }
```
skjnldsv, line 25:
```suggestion
```
No needed anymore

skjnldsv (dismissed):

src/components/SettingsInputText/SettingsInputText.vue
```diff
@@ -0,0 +1,201 @@
...
+
+		// if disabled, change cursor
+		&:disabled {
+			cursor: default;
+		}
+
+			.hint {
+				color: var(--color-text-lighter);
+			}
+		}
```
juliushaertl, line 198: This is one `}` to much ;)

GretaD: thank you :), fixed. Can you please have a general look and merge it if all is good :D

ChristophWurst (approved):
Looks good to me :)

src/components/SettingsInputText/SettingsInputText.vue
```diff
@@ -0,0 +1,200 @@
...
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+
+<template>
+	<form ref="form"
+		:disabled="disabled"
+		@submit.prevent="onSubmit">
```
skjnldsv, line 26:
```suggestion
		@submit.prevent.stop="onSubmit">
```

src/components/SettingsInputText/SettingsInputText.vue
```diff
@@ -0,0 +1,200 @@
...
+			this.$emit('input', event)
+			/**
+			 * Emitted when the inputs value changes
+			 * @type {string}
+			 */
+			this.$emit('update:value', event.target.value)
+		},
+		onSubmit(event) {
+			event.preventDefault()
+			event.stopPropagation()
```
skjnldsv, line 109:
```suggestion
```
Not necessary if you remove the inline event prevent/stop on line 26 :)

src/components/SettingsInputText/SettingsInputText.vue
```diff
@@ -0,0 +1,200 @@
...
+			if (!this.disabled) {
+				/**
+				 * Emitted on submit of the input field
+				 * @type {Event}
+				 */
+				this.$emit('submit', event)
+			} else {
+				// ignore submit
+				return false
+			}
```
skjnldsv, line 119:
```suggestion
			}
```

You already prevent and stop, there shouldn't be the need for a return at this point :thinking:

src/components/SettingsInputText/SettingsInputText.vue
```diff
@@ -0,0 +1,200 @@
...
+			 */
+			this.$emit('change', event)
+		},
+	},
+}
+</script>
+
+<style lang="scss" scoped>
+
+	$input-margin: 4px;
```
skjnldsv, line 134: Nice!
Thanks for using variables like that! :rocket:

src/components/SettingsInputText/SettingsInputText.vue
```diff
@@ -0,0 +1,200 @@
...
+		@submit.prevent="onSubmit">
+		<div class="input-wrapper">
+			<label :for="id" class="action-input__label">{{ label }}</label>
+			<input :id="id"
+				type="text"
+				:value="value"
+				:disabled="disabled"
+				@input="onInput"
+				@change="onChange">
+			<input :id="idSubmit" type="submit" class="action-input__submit">
```
skjnldsv, line 35: If there is no value, the browser will add a fallback. We should propably control the text with our translation so that we use Nextcloud's user language and not the browsers :)

Something like ?
```suggestion
			<input :id="idSubmit" :value="t('Submit')" type="submit" class="action-input__submit">
```

skjnldsv (approved):
Some comments to fix :)
Looks good otherwise!

PR #651, opened 2019-10-14
add SettingsSection component

GretaD:


src/components/SettingsSection/SettingsSection.vue
```diff
@@ -0,0 +1,53 @@
...
+</template>
+
+<script>
+export default {
+	name: 'SettingsSection',
+	components: {
+
+	},
+	props: {
+		initial: {
```
juliushaertl: title and description are missing, initial is not needed

src/components/SettingsSection/SettingsSection.vue
```diff
@@ -0,0 +1,53 @@
...
+			type: Object,
+			required: true
+		}
+	}
+}
+
+</script>
+
+<style scoped>
+	p {
```
juliushaertl: Styles for .section should be taken over from settings.scss

src/components/SettingsSection/SettingsSection.vue
```diff
@@ -0,0 +1,53 @@
...
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+
+<template>
+	<div class="section">
+		<h2>{{ title }}</h2>
```
juliushaertl: We should also add the option to specify a documentation URL here that is then linked with an info icon like we use it at some places:

image

src/components/SettingsSection/SettingsSection.vue
```diff
@@ -0,0 +1,75 @@
+<!--
+  - @copyright Copyright (c) 2019 Julius Härtl <jus@bitgrid.net>
```
ChristophWurst: no? :)

src/components/SettingsSection/SettingsSection.vue
```diff
@@ -0,0 +1,75 @@
...
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+
+<template>
+	<div class="section">
+		<h2>{{ title }}</h2>
+		<p>{{ description }}</p>
+		<a href="docUrl" class="icon-info" :title="t('settings', 'Open Documentation')"/>
```
ChristophWurst: test data to remove?

src/components/SettingsSection/SettingsSection.vue
GretaD: I adapted the SettingsSection from richdocuments that Julius did. So we are co-authors :) need to add mine too

src/components/SettingsSection/SettingsSection.vue
```diff
@@ -0,0 +1,78 @@
...
+<script>
+export default {
+	name: 'SettingsSection',
+	components: {
+
+	},
+	props: {
+		title: {
+			type: String,
+			default: '',
```
ChristophWurst: not necessary if the prop is required

ChristophWurst:
Please rebase

src/components/SettingsSection/SettingsSection.vue
```diff
@@ -0,0 +1,77 @@
...
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+
+<template>
+	<div class="section">
+		<h2>{{ title }}</h2>
+		<p>{{ description }}</p>
+		<a href="docUrl" class="icon-info" :title="title" />
```
georgehrke: Shouldn't `docUrl` be a property?

skjnldsv:
Rebased and ready to go!

@GretaD see what I did, it might help you for the next pending pr: :wink: 
- Added documentation
- Imported icons so we don't rely on the server css
- Used different class names so we don't rely on the server css
- Used BEM class naming convention
- Made sure accessibility was ok (external doc title translation) 
- Used our existing variables so we don't have random px units laying around :grin: 

image

skjnldsv:
Looks perfect in the documentation too :)

PR #653, opened 2019-10-15
add SettingsSelectGroup component

GretaD:


src/components/SettingsSelectGroup/SettingsSelectGroup.vue
```diff
@@ -0,0 +1,126 @@
...
+			default: () => []
+		},
+		disabled: {
+			type: Boolean,
+			default: false
+		}
+	},
+	data() {
+		return {
+			inputValObjects: [],
```
juliushaertl: We can actually simplify this a bit, by removing the inputValObjects and using a computed property as :value and manually emitting the changed value with an input event.

src/components/SettingsSelectGroup/SettingsSelectGroup.vue
```diff
@@ -0,0 +1,121 @@
...
+		:disabled="disabled" @input="update"
+		@search-change="asyncFindGroup">
+		<span slot="noResult">{{ t('settings', 'No results') }}</span>
+	</Multiselect>
+</template>
+
+<script>
+import axios from 'nextcloud-axios'
+import Multiselect from 'nextcloud-vue/dist/Components/Multiselect'
+let uuid = 0
```
ChristophWurst: this looks suspicious. is there any way around this module-global variable? This will be shared across all instances of the component if I'm not mistaken.

skjnldsv: Yeah, we have an uuid generator in the vue components for this purpose exactly :)

GretaD: yes, this one right: https://github.com/nextcloud/nextcloud-vue/blob/master/src/utils/GenRandomId.js

will update asap

src/components/SettingsSelectGroup/SettingsSelectGroup.vue
```diff
@@ -0,0 +1,146 @@
...
+		},
+		id() {
+			return 'settings-select-group-' + this.uuid
+		},
+		groupsArray() {
+			return Object.values(this.groups)
+		}
+	},
+	created: function() {
+		this.uuid = uuid.toString()
```
ChristophWurst: where do you define `uuid`?

GretaD: oh, wait, i should have changed `uuid` into  `GenRandomId`

skjnldsv:
What is the status on this? :)

ChristophWurst:
That's this for, actually? A server feature/PR?

juliushaertl:
> That's this for, actually? A server feature/PR?

Not really, but we use a group selection in quite some settings, so it would be nice to have it as a shared component.

src/components/SettingsSelectGroup/SettingsSelectGroup.vue
```diff
@@ -0,0 +1,149 @@
...
+		GenRandomId() {
+			return 'settings-select-group-' + this.GenRandomId
+		},
+		groupsArray() {
+			return Object.values(this.groups)
+		}
+	},
+	created: function() {
+		this.GenRandomId = GenRandomId.toString()
+		GenRandomId += 1
```
ChristophWurst: :thinking: :thinking: :thinking: 

`GenRandomId` is a function. Why are you calling `.toString()`? I think you want to invoke it, right?

Also, adding +1 to a function on the next line :speak_no_evil:

src/components/SettingsSelectGroup/SettingsSelectGroup.vue
```diff
@@ -0,0 +1,139 @@
...
+	<Multiselect :value="inputValue"
+		:options="groupsArray" :options-limit="5"
+		:placeholder="label"
+		track-by="id"
+		label="displayname"
+		class="multiselect-vue" :multiple="true"
+		:close-on-select="false" :tag-width="60"
+		:disabled="disabled" @input="update"
+		@search-change="asyncFindGroup">
+		<span slot="noResult">{{ t('settings', 'No results') }}</span>
```
ChristophWurst: you have to run `npm run l10n:extract` to update the .pot file

ChristophWurst: sorry, it's actually just `t('No results')` here, no app, as of #854

ChristophWurst: Fix it and run `npm run l10n:extract` again

src/components/SettingsSelectGroup/SettingsSelectGroup.vue
```diff
@@ -0,0 +1,139 @@
...
+			return Object.values(this.groups)
+		}
+	},
+	methods: {
+		update() {
+			this.$emit('input', this.inputValue.map((element) => element.id))
+		},
+		asyncFindGroup(query) {
+			query = typeof query === 'string' ? encodeURI(query) : ''
+			return axios.get(OC.linkToOCS(`cloud/groups/details?search=${query}&limit=10`, 2))
```
skjnldsv:
```suggestion
			return axios.get(OC.linkToOCS(`cloud/groups/details?search=${query}&limit=10`, 2))
```

Please import @nextcloud/router and use the generateOcsUrl function instead :)

src/components/SettingsSelectGroup/SettingsSelectGroup.vue
```diff
@@ -0,0 +1,139 @@
...
+				.then((response) => {
```
skjnldsv: Please use a try...catch with async/await instead of .then/.catch syntax :rocket:

PR #656, opened 2019-10-16
Fix spacer order

skjnldsv:

### DIFF
--- a/src/components/AppNavigationSpacer/AppNavigationSpacer.vue
+++ b/src/components/AppNavigationSpacer/AppNavigationSpacer.vue
@@ -32,5 +32,6 @@ export default {
 <style scoped>
 	.app-navigation-spacer {
 		height: 22px;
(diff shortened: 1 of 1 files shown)


jancborchardt:
Actually the whitespace seems to only be shown when the navigation does not scroll. That is, for accounts with more folders or for smaller screens there is no whitespace separation. Additionally, it can happen that when you "Show more folders" it actually causes the whitespace to vanish and then show again on collapse.

The whitespace should always be shown as it is important for visual separation – even on small screens. ;)

skjnldsv:
What is exactly happening when it's scrolling?
Is it shrinking?

jancborchardt:
Just zoom a bit so that the navigation scrolls, no whitespace to be found:
whitespace lost

Correct is this with vertical whitespace between the unified inbox, and the accounts:
whitespace shown

PR #659, opened 2019-10-17
Move to browserslist global config

skjnldsv:
Because there was some weird stuff, let's make sure we support proper list


babel.config.js
```diff
@@ -0,0 +1,15 @@
+module.exports = {
+	plugins: [
+		'@babel/plugin-transform-runtime',
+		'@babel/plugin-transform-react-jsx'
```
ChristophWurst: @juliushaertl do we really need this? Added in https://github.com/nextcloud/nextcloud-vue/pull/439

PR #660, opened 2019-10-17
Use regular text color for selected tag to not de-emphasize

jancborchardt:
Can be tested in Mail with the to, cc and bcc fields @skjnldsv @ChristophWurst   Otherwise there was no difference between placeholder and tag, especially cause there we remove the border.

### DIFF
--- a/src/components/Multiselect/index.scss
+++ b/src/components/Multiselect/index.scss
@@ -6,7 +6,7 @@
 	padding: 0 !important;
 	display: inline-block;
 	/* override this rule with your width styling if you need */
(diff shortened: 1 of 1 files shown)


skjnldsv:
Please add screeenshot when it's about design! Our design lead is very picky! :D

skjnldsv:
Bump? :wink:

jancborchardt:
@skjnldsv could you review it so we can get it over with? ;) Sorry, appreciate it

jancborchardt:
:bowing_man:

PR #678, opened 2019-10-22
move dependencies to devDependencies

korelstar:
This is just an idea for optimization, please review it detailed. The approach is as follows: 

Due to the use of webpack, the `@nextcloud/vue` release package contains all used libraries (e.g. `v-tooltip`) directly inside the generated files in the `dist` folder. Hence, when using `@nextcloud/vue` in an app, these libraries are not needed separately at runtime and, hence, there should be no need to install those libraries locally. However, those libraries are listed in `package.json`' section `dependencies`. Therefore, when installing `@nextcloud/vue`, those libraries are nevertheless installed locally.

This can be improved by moving those libraries to the `devDependencies` section. Then, these libraries are only used inside the `nextcloud-vue` project for building `@nextcloud/vue` but apps are not required to install them.


ChristophWurst:
Are you sure about this change? As far as I can tell many of these dependencies are *runtime dependencies* (e.g. Axios), hence I don't think this should be moved out of `dependencies`.

korelstar:
No, I'm not ;-). I just wanted to start a discussion on this.

Regarding runtime dependencies: you're using webpack in your build and release process. Hence, all runtime dependencies are included in the release file that you publish on NPM. Therefore, it is not *required* to include them as `dependencies` (you need to have them in `devDependecies` though).

Please feel free to test this :-)

juliushaertl:
> However, those libraries are listed in package.json' section dependencies. Therefore, when installing @nextcloud/vue, those libraries are nevertheless installed locally.

I totally missed that when looking into making dependencies "deduplicatable". We should then just be able to mark our dependencies as external and have much smaller builds that bundle the nextcloud-vue dependencies at build-time of the apps that use it. Also that should make sure that dependencies are properly deduped. Let me do some testing if that works like I have it in mind.

skjnldsv:
Closing after #679

korelstar:
Perfect result, good that we talked about this :-)

PR #679, opened 2019-10-22
Mark dependencies as webpack externals

juliushaertl:
This PR adds the external dependencies we use to the webpack externals list, so they are not included in the bundle. Since they are part of the dependency section they will still be installed by npm when app developers install our package and webpack will take care of resolving them properly then.

Before
image

After
image

The size of the saving in the app bundles of course depends on which components are used, but when using multiple components that include the usage of v-tooltip, popper.js will only be included once with this PR.

A quick test run on the deck vue branch saved me ~300kb in duplicate dependencies.

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -8,6 +8,8 @@ const IconfontName = 'iconfont-vue'
 
 const { DefinePlugin } = require('webpack')
 
+const nodeExternals = require('webpack-node-externals')
+
 // scope variable
 const md5 = require('md5')
 const PACKAGE = require('./package.json')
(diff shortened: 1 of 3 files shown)


korelstar:
Nice! Why not marking `vue` as external dependency, too?

juliushaertl:
It is already :wink: https://github.com/nextcloud/nextcloud-vue/pull/679/files#diff-5c572bb8f24a00a84aa28f22d3a39ba5R44

skjnldsv:
> Since they are part of the dependency section they will still be installed by npm when app developers install our package and webpack will take care of resolving them properly then.

But then they will be considered as extraneous dependencies, no? 
https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md

juliushaertl:
> Forbid the import of external modules that are not declared in the package.json's dependencies

No, as they are still part of the dependencies in package.json. And for the app itself only exports of @nextcloud/vue should be imported so the linting rule should still be fine.

ChristophWurst:
I gave this a test with Mail. The bundle size is a sum of all .js files, without map files.

Built against nc-vue master: 2692B
Built against this branch: 2444B

This means the new bundle will be just 90.79% of the size. This is quite impressive considering that Mail uses many of the components of this lib.

webpack.common.js
```diff
@@ -46,7 +46,17 @@ module.exports = {
 			commonjs2: 'vue',
 			amd: 'vue',
 			root: 'Vue'
-		}
+		},
+		'@nextcloud/axios': '@nextcloud/axios',
```
skjnldsv: Can't we automate this somehow?
Instead of manually specifying them?

juliushaertl: I pushed another commit to exclude all from node_modules. See https://www.npmjs.com/package/webpack-node-externals#why-not-just-use-a-regex-in-the-webpack-config for why the additional package. :wink: 

Seems to still work fine :)

skjnldsv: Ah yes, I forgot about this https://github.com/skjnldsv/vueexample/blob/74d79f574abea20326369d6a7cedbc7afd7df353/package.json#L78

PR #688, opened 2019-10-25
fix action's long-text style

### LINKED ISSUE #575: ActionButton: mixing "long" text with "short" text looks odd
The developer can't decide whether the text of an `ActionButton` should be interpreted as "long" or as "short". The problem is, that both types are rendered differently. If you have some actions that are "short" and some that are "long", the result looks really odd, see screenshot:

Screenshot

korelstar:
fixes #575, see screenshot over there.

### DIFF
--- a/src/assets/action.scss
+++ b/src/assets/action.scss
@@ -90,6 +90,7 @@
 		p {
 			width: 150px;
 			padding: #{$icon-margin / 2} 0;
+			margin: auto;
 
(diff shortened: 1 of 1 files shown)


skjnldsv:
Screenshot? :)

korelstar:
as expected:

| Before | After |
|--|----|
| Screenshot | Screenshot

PR #698, opened 2019-10-28
remove menu button

GretaD:
ref #[2110](https://github.com/nextcloud/mail/issues/2110)
On narrow screens, the message view has both a menu button and a back arrow button. In this view only the back arrow button is necessary and it should be in place of the menu button.


skjnldsv:
Well, if you do this there is no way to open the appnavigation anymore :thinking:

GretaD:
> Well, if you do this there is no way to open the appnavigation anymore thinking

Yes i know :) i only pushed it so Christoph would understand what I was doing. Its not done yet. A discussion is happening on the main ticket.

GretaD:
new pull request on mail app ref #[2211](https://github.com/nextcloud/mail/pull/2211)

PR #707, opened 2019-11-05
🐛 FIX: exact-active-class support to not always highlight root element as active

violoncelloCH:
properly pass the exact prop to the `router-link`


skjnldsv:
AH no, not for the stable0-12

PR #710, opened 2019-11-07
Adjust header elements

nickvergessen:
cc @jancborchardt as discussed

### DIFF
--- a/src/components/AppSidebar/AppSidebar.vue
+++ b/src/components/AppSidebar/AppSidebar.vue
@@ -111,18 +111,18 @@
 						class="app-sidebar-header__star" @click.prevent="toggleStarred" />
 
 					<!-- main title -->
-					<h3 v-if="!titleEditable" class="app-sidebar-header__title">
+					<h2 v-if="!titleEditable" class="app-sidebar-header__title">
 						{{ title }}
-					</h3>
+					</h2>
 
 					<input v-if="titleEditable" v-focus class="app-sidebar-header__title-input"
 						type="text" :placeholder="titlePlaceholder" :value="title"
 						@input="onTitleInput">
 
 					<!-- secondary title -->
-					<h4 v-if="subtitle.trim() !== ''" class="app-sidebar-header__subtitle">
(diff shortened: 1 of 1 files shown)


src/components/AppSidebar/AppSidebar.vue
```diff
@@ -517,16 +517,15 @@ $desc-height: 46px;
 				}
 			}
 			// titles
-			h3, h4 {
+			h2, p.app-sidebar-header__subtitle {
```
jancborchardt:
```suggestion
			h2,
			.app-sidebar-header__subtitle {
```
Best only use the class, just in case the HTML element ever changes

src/components/AppSidebar/AppSidebar.vue
```diff
@@ -535,7 +534,7 @@ $desc-height: 46px;
 			}
 
 			// subtitle
-			h4 {
+			p.app-sidebar-header__subtitle {
```
jancborchardt:
```suggestion
			.app-sidebar-header__subtitle {
```
Similar as above :)

src/components/AppSidebar/AppSidebar.vue
```diff
@@ -517,16 +517,16 @@ $desc-height: 46px;
 				}
 			}
 			// titles
-			h3, h4 {
+			h2,
```
skjnldsv:
```suggestion
			.app-sidebar-header__title,
```

src/components/AppSidebar/AppSidebar.vue
```diff
@@ -517,16 +517,16 @@ $desc-height: 46px;
...
 				width: 100%;
 				white-space: nowrap;
 				text-overflow: ellipsis;
 				overflow: hidden;
 				margin: 0;
 			}
 			// main title
-			h3 {
-				font-size: 16px;
+			h2 {
```
skjnldsv:
```suggestion
			.app-sidebar-header__title {
```

jancborchardt:
Btw @georgehrke is this something which then needs to be changed for the component which Calendar uses, or is it just the same?

PR #712, opened 2019-11-07
🐛 FIX: allow proper linebreak in multiselect for narrow mobile screens

violoncelloCH:
e.g. in Mail app
image
vs
image

### DIFF
--- a/src/components/Multiselect/index.scss
+++ b/src/components/Multiselect/index.scss
@@ -63,7 +63,7 @@
 		cursor: pointer;
 		position: relative;
 		border-radius: 3px;
-		height: 34px;
+		min-height: 34px;
 		/* tag wrapper */
 		.multiselect__tags-wrap {
 			align-items: center;


jancborchardt:
Wondering if we should do this, or limit to 1 line and ellipsize. Cause for long names or email addresses or so, they might break to 3 rows or something and then it looks extra weird?

violoncelloCH:
That's what came through my mind as well while doing this but I wasn't sure what would be better. In the use case of the mail app I want to see which account I'm sending from, however if we ellipse too much this would only display my name (which is probably the same for all accounts) and cut off the end / relevant part of the mail (domain)...
What do you think @jancborchardt ?

jancborchardt:
Is it possible to limit what is shown inside the multiselect depending on screen width?

Meaning we could show both name and email in the open dropdown, but only the email when directly picked. (This only on mobile of course.)

Cause the "from" info is comparably the least important info on this screen, and thus should not take 2 lines.
Also, on mobile the vertical space is limited, especially when writing which means the keyboard is extended.

skjnldsv:
What is the status here?

jancborchardt:
@skjnldsv do you have an idea about the question in my comment? Or @violoncelloCH? :)

skjnldsv:
> @skjnldsv do you have an idea about the question in my comment? Or @violoncelloCH? :)

You could maybe implement this with your own multiselect template, but that would be tricky, there is no _steps_ methods to display a text or not. Only solution in my head would be to use media queries for it :thinking:

skjnldsv:
Rebased.

PR #746, opened 2019-11-25
Add proper hover/focus/active feedbacks to AppNavigationItem

ma12-co:

### DIFF
--- a/src/components/AppNavigationItem/AppNavigationItem.vue
+++ b/src/components/AppNavigationItem/AppNavigationItem.vue
@@ -114,7 +114,8 @@ Just set the `pinned` prop.
 			'app-navigation-entry--pinned': pinned,
 			'app-navigation-entry--editing' : editing,
 			'app-navigation-entry--deleted': undo,
-			'app-navigation-entry--collapsible': collapsible
+			'app-navigation-entry--collapsible': collapsible,
+			'active': isActive
 		}"
 		class="app-navigation-entry">
 		<!-- Icon and title -->
@@ -361,6 +362,9 @@ export default {
 			return {
 				is: 'li'
 			}
+		},
+		isActive() {
+			return this.to && this.$route === this.to
 		}
 	},
 	watch: {
@@ -427,14 +431,6 @@ export default {
 	width: 100%;
 	min-height: $clickable-area;
 
-	&.active,
-	a:hover,
-	a:focus,
-	a:active {
-		color: var(--color-main-text);
-		box-shadow: inset 4px 0 var(--color-primary);
-	}
-
 	/* hide deletion/collapse of subitems */
 	&.app-navigation-entry--deleted,
 	&.app-navigation-entry--collapsible:not(.app-navigation-entry--opened) {
@@ -455,11 +451,22 @@ export default {
 		padding: 0;
 		padding-right: $icon-margin;
 		white-space: nowrap;
-		color: var(--color-text-light);
+		color: var(--color-main-text);
 		background-repeat: no-repeat;
 		background-position: $icon-margin center;
 		background-size: $icon-size $icon-size;
 		line-height: $clickable-area;
+		&:hover,
+		&:hover ~ .app-navigation-entry__utils,
+		&:focus,
+		&:focus ~ .app-navigation-entry__utils {
+			background-color: var(--color-background-hover);
+		}
+		&.active,
+		&:active,
+		&:active ~ .app-navigation-entry__utils {
+			background-color: var(--color-primary-light);
+		}
 
 		.app-navigation-entry-icon {
 			display: flex;


src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -114,7 +32,8 @@ Just set the `pinned` prop.
 			'app-navigation-entry--pinned': pinned,
 			'app-navigation-entry--editing' : editing,
 			'app-navigation-entry--deleted': undo,
-			'app-navigation-entry--collapsible': collapsible
+			'app-navigation-entry--collapsible': collapsible,
+			'active': isActive
```
korelstar: Will `<AppNavigationItem class="active">` still work with this? This is needed if the prop `to` is not used.

ma12-co: Yes @korelstar, we're going to have both :)

juliushaertl:
I remember we had some mockups that specified that, does anyone have a link?

korelstar:
https://github.com/nextcloud/nextcloud-vue/issues/425#issuecomment-527914136

ma12-co:
> I remember we had some mockups that specified that, does anyone have a link?

@juliushaertl that's the orignal idea but since then things have changed: we're not going to have the left border anymore (both in focus and active), therefore focus and hover are going to be the same

ma12-co:
@jancborchardt I don't think we've agreed on the final colors!

jancborchardt:
@ma12-co nice! How do I best test this – npm link with the Photos app?

ma12-co:
> @ma12-co nice! How do I best test this – npm link with the Photos app?

I think the best way is to use the vueexample app by @skjnldsv:
https://github.com/skjnldsv/vueexample
but remember to checkout bugfix/noissue/update-to-use-the-new-npm-module to test it
And yes, npm link :)

ma12-co:
ping @jancborchardt :)

jancborchardt:
@ma12-co will review asap, sorry! :) Is there also a pull request to have this in Files, that is the old styles.scss?

ma12-co:
> Is there also a pull request to have this in Files, that is the old styles.scss?

not that I know of

skjnldsv:
Is it ready for reviews? Labels?

ma12-co:
@skjnldsv we were discussing the colors, but I guess everything else can be reviewed. @jancborchardt pick the colors :)

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -455,11 +452,22 @@ export default {
...
+		color: var(--color-main-text);
 		background-repeat: no-repeat;
 		background-position: $icon-margin center;
 		background-size: $icon-size $icon-size;
 		line-height: $clickable-area;
+		&:hover,
+		&:hover ~ .app-navigation-entry__utils,
+		&:focus,
+		&:focus ~ .app-navigation-entry__utils {
+			background-color: var(--color-background-hover);
```
jancborchardt: The --color-background-hover needs to be added to css-variables.scss first as well for this to work. :)

ma12-co: done :)

jancborchardt:
Should we get it in? :) And then also proceed using it for the navigation in Files, and for action menus and header menus?

ma12-co:
Build was failing here, I lint-fixed and rebased

korelstar:
Sorry, I hadn't the time to review this, before. But now, I'm wondering if the merged result is really intended. Unfortunately, this PR has no screenshot, although it is about design!

However, here is what these changes do:

| Before | After |
|---|---|
| grafik | grafik |

It's good that `hover` and `active` doesn't look equal, but now `active` and `normal` have the same design! I don't think that this was intended, maybe the CSS is broken (the `.active` CSS class was used to be applied to `.app-navigation-entry`, but the new CSS requires `.app-navigation-entry-link` to have `.active`, which is impossible to realize).

The second thing is the text-color: before, the text-color was `--color-text-light` and only `active` had `--color-main-text`. Now, every app-navigation-item has `--color-main-text`. The difference is subtile on above screenshots, but it's better visible if there are many items. Is this really intended?

@jancborchardt @ma12-co

ma12-co:
Thanks for the review @korelstar :) 

> but the new CSS requires .app-navigation-entry-link to have .active, which is impossible to realize).

I don't understand, can you explain this further? This is currently working in our talk app at least

> . Now, every app-navigation-item has --color-main-text.

This intended, yes. From now on there wil be no change in text color for hover/active/focus states.

korelstar:
Hmm. Do you use this component in Talk at all? I can't find it: https://github.com/nextcloud/spreed/search?utf8=%E2%9C%93&q=AppNavigationItem&type=Code , could you give me a hint, please?

What I tried to say was: `AppNavigationItem` has the following (abstract) structure:
```
<li class="app-navigation-entry">
  <a class="app-navigation-entry-link">...</a>
</li>
```

If an item is active, the `<li>` gets also the `active` class (see https://github.com/nextcloud/nextcloud-vue/blob/50ee5a20d3b8c126bfda33b6ea1eb422a4651df2/src/components/AppNavigationItem/AppNavigationItem.vue#L118). This results in:
```
<li class="app-navigation-entry active">
  <a class="app-navigation-entry-link">...</a>
</li>
```

However, there is no stylesheet in the `AppNavigationItem` component for `app-navigation-entry active`. Instead, there is a stylesheet for `app-navigation-entry-link active` (see https://github.com/nextcloud/nextcloud-vue/blob/50ee5a20d3b8c126bfda33b6ea1eb422a4651df2/src/components/AppNavigationItem/AppNavigationItem.vue#L465-L469).

This means, I would have to add the `active` class to an inner element of `AppNavigationItem` in order to achieve:
```
<li class="app-navigation-entry">
  <a class="app-navigation-entry-link active">...</a>
</li>
```

But this is not possible, since I can't add CSS classes to that element from outside the `AppNavigationItem` component (see https://github.com/nextcloud/nextcloud-vue/blob/50ee5a20d3b8c126bfda33b6ea1eb422a4651df2/src/components/AppNavigationItem/AppNavigationItem.vue#L122-L124).

However, this approach seems not to be correct to me. I think it should be sufficient to add the `active` CSS class to the root element of `AppNavigationItem` (like it was before of this PR). But this means we have to move [this CSS part](https://github.com/nextcloud/nextcloud-vue/blob/50ee5a20d3b8c126bfda33b6ea1eb422a4651df2/src/components/AppNavigationItem/AppNavigationItem.vue#L465-L469) to [here](https://github.com/nextcloud/nextcloud-vue/blob/50ee5a20d3b8c126bfda33b6ea1eb422a4651df2/src/components/AppNavigationItem/AppNavigationItem.vue#L433)

Is this now understandable? Maybe I have overseen something, but it definitely doesn't work for me, now.

raimund-schluessler:
I can confirm that this PR broke the hover and active feedback. There is no indication when hovering an item and the active item is missing the box-shadow at the left side.

ma12-co:
> Hmm. Do you use this component in Talk at all?

@korelstar we don't use it but I did kind of the same for another component that has a very similar structure and behave similarly. This was tested with vueexample and it all worked. I will look more into this today, thanks for the heads up :) 

> There is no indication when hovering an item and the active item is missing the box-shadow at the left side.

@raimund-schluessler, the box-shadow highlight is going to be dropped everywhere in Nextcloud.

skjnldsv:
> the box-shadow highlight is going to be dropped everywhere in Nextcloud.

How so? 
https://github.com/nextcloud/nextcloud-vue/issues/425#issuecomment-527914136

ma12-co:
>How so?

Further developments of that while discussing it with @jancborchardt in September's contributor week. The box shadow used to serve the function of differentiating between hover and focus, but now we won't have that distinction anymore. Plus, If it's properly spaced, when the element is not active, the white-space that's left there feels weird in components that have an avatar right next to it.

raimund-schluessler:
> @raimund-schluessler, the box-shadow highlight is going to be dropped everywhere in Nextcloud.

Is there an image somewhere how it *should* look like? The last one I saw was https://github.com/nextcloud/nextcloud-vue/issues/425#issuecomment-527914136 and I thought this was the agreed one.

Because at the moment there is no indication that an entry is active, there is no feedback when hovering and only while you keep the entry clicked the background-color changes. This looks very broken at the moment.

skjnldsv:
If I remember correctly the idea was to have the blue border for active only.
And the grey background for focus & hover, right?

raimund-schluessler:
> If I remember correctly the idea was to have the blue border for active only.
> And the grey background for focus & hover, right?

That's what I thought as well. And a primary-color 10% opacity background-color for active.

ma12-co:
> So we're not making it disappear from nextcloud, but just removing it from the hover/focus feedback

No no, I swear it's gone! This has been discussed in September and confirmed this January's contributor's week. Talk 8 already behaves like that. I'm sorry that we didn't post it publicly on github. Here's the new spec:

* no box shadows, neither in focus nor active states;
* hover and focus: color-background-hover;
* active: color-primary-light;

skjnldsv:
I'm surprised it's been approved by Jan considering there is no differentiation for colourblind people if we remove the vertical bar.
I see no differences between active and focus :thinking:

raimund-schluessler:
> Here's the new spec:
> 
>     * no box shadows, neither in focus nor active states;
> 
> 
> Only the background of the element will change:
> 
>     * hover and focus: color-background-hover;
> 
>     * active: color-primary-light;

I am also a bit surprised, but even this spec is not respected at the moment.

skjnldsv:
Let me also add that the contrast between the light background (e6f3fa) and the default background (ffffff) doesn't pass wcag, meaning we definitely need proper feedback. I don't really care what solution we go to, but the current one in talk is not acceptable in an accessibility point of view I'd say :)

ma12-co:
I am also a bit surprised, but even this spec is not respected at the moment.

I know @raimund-schluessler, there's a bug with the active state. I'll try and fix it today

> I'm surprised it's been approved by Jan

Well this was actually Jan's Idea!
 
> there is no differentiation for colourblind people if we remove the vertical bar.

I don't agree here as it entirely depends on the primary color. If a colorblind person sets the primary color to one that he/she can actually distinguish, then there should be enough differentiation on each primary element in nc.

> the contrast between the light background (e6f3fa) and the default background (ffffff) doesn't pass wcag

For this we can simply darken the varialbles, right?

skjnldsv:
> I don't agree here as it entirely depends on the primary color. If a colorblind person sets the primary color to one that he/she can actually distinguish, then there should be enough differentiation on each primary element in nc.

Not really. The background is light enough to not really matter much on which primary colour you picked. This is just a tint, in the end this is still a high luminance setting :)

> For this we can simply darken the varialbles, right?

Again, I can't be everywhere, nor do I want to spend time on all issues.
@jancborchardt and you decided to change the old behaviour, I'm 100% fine with changing. But we bragged about being wcag compliant one release ago, and now we aren't on our main navigation. The only thing I want is thinking design with accessibility in mind and not have regression :+1:

ma12-co:
Ok, I'll leave you and Jan discuss about this, just one more remark:

Looking at [this contrast calculator](https://contrastchecker.com/), right now the contrast ratio between the default primary light blue and white is 2.63. [From what I understand we just have to reach 3](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html) to be WCAG compliant so it doesn't look like we're that far off :)

skjnldsv:
The contrast between the background and the light one is 1.13 :)
image


Also it still doesn't fix the colourblind issue between grey-hover and blue-active :/

ma12-co:
You're right, I don't know how I got that number. Well, as I said, stepping back and leaving it to you and Jan :)

skjnldsv:
More like up to Jan :p

raimund-schluessler:
Would be good to get some feedback from @jancborchardt here, because at the moment every app using nextcloud/vue@1.3.0 has a broken sidebar active/hover feedback.

Also, e.g. Firefox also has a colored border at the top of the hovered/active tabs, so we wouldn't be alone using this way of highlighting 😉

skjnldsv:
> Would be good to get some feedback from @jancborchardt here

vacations, he'll be back in a week :)
(and i'll be gone for 2)

korelstar:
@jancborchardt @skjnldsv Did you find some time to discuss this issue?

skjnldsv:
We just got back, sorry, not yet!

raimund-schluessler:
I would like to release Tasks 0.12.0 in the next days and would appreciate a fix / consensus. Otherwise I need to hot-fix it in the Tasks code.

ma12-co:
@raimund-schluessler, we merged a fix yesterday for the active state, sorry that it took so long. Can you confirm that it works with tasks?

raimund-schluessler:
@ma12-co Ah, great. Sorry I didn't notice that. I will check in the evening.

raimund-schluessler:
@ma12-co The active feedback works as expected now.

skjnldsv:
Awesome! :)

jancborchardt:
Sorry for the late reply! :) Here’s answers to the 2 important issues from what I can see – let me know if I missed anything else:



> I'm surprised it's been approved by Jan considering there is no differentiation for colourblind people if we remove the vertical bar.
> I see no differences between active and focus :thinking:

The reason is that these 2 states don’t really need to be differentiated down to that. Only 1 item is ever active, and only 1 item is ever focused/hovered. And you know what you focus/hover because you are doing that.


> Let me also add that the contrast between the light background (e6f3fa) and the default background (ffffff) doesn't pass wcag, meaning we definitely need proper feedback. I don't really care what solution we go for, but the current one in talk is not acceptable in an accessibility point of view I'd say :)
> https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html

Keep in mind that these color contrast guidelines are for text on background, not for differences between backgrounds. :) A large area of different shade needs less contrast than thin text, where it’s also important that it’s actually readable. A large area only needs to be visible.

jancborchardt:
@skjnldsv can you clarify your :-1:?

Issue #748, opened 2019-11-26
Add ability to tweak the modal backdrop color 


ma12-co:
Right now the modal casts this very dark backdrop on the whole page. I think that is good for photo viewing but very aggressive for anything else, as it almost completely hides the UI underneath.

![Screenshot_20191126_104549](https://user-images.githubusercontent.com/26852655/69618374-171ac780-103a-11ea-8fa8-d126be367e2a.png)

I think it would make sense to have at least 2 options: this very dark one, and a more clear one, as per material design specs: https://material.io/components/dialogs/#anatomy

What do you think @nextcloud/designers?

skjnldsv:
do it in css? ^^

skjnldsv:
> at least 2 options: this very dark one, and a more clear one

ah right, that would be cleaner indeed :)
please do! :rocket: :heart:

ma12-co:
Actually IMO we should make the dark option completely black, as the UI behind the current one is so dark that it's kinda useless and it seems almost an error/unwanted.

jancborchardt:
Yes, I thought we agreed on only using this very dark one only for picture viewing? e.g. the First run wizard and all the normal modals should use a rather light backdrop.

This does not need to be configurable. Basically the default is a light backdrop, and Viewer simply uses a very dark backdrop.

skjnldsv:
@jancborchardt so a toggle between the viewer mode and the first run wizard?

jancborchardt:
> @jancborchardt so a toggle between the viewer mode and the first run wizard?

No, as said above:

> This does not need to be configurable. Basically the default is a light backdrop, and Viewer simply uses a very dark backdrop.

So everyone uses the default, which is a lighter backdrop. Viewer force overrides it. If we see that any other app needs it, we can make the Viewer backdrop style available to others as well.

jancborchardt:
Nevermind, https://github.com/nextcloud/nextcloud-vue/pull/750 seems good :)

PR #750, opened 2019-11-27
Fix modal backdrop

ma12-co:
Fix: #748

- [x] Make default backdrop light
- [x] Add prop to toggle black backdrop
- [x] Make modal box shadow softer

Here's how it looks: 

Screenshot from 2019-11-27 15-42-15

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -60,7 +60,10 @@ export default {
 </docs>
 <template>
 	<transition name="fade">
-		<div ref="mask" class="modal-mask" @click="handleMouseMove"
+		<div ref="mask"
+			class="modal-mask"
+			:class="{ 'modal-mask--dark' : dark }"
+			@click="handleMouseMove"
 			@mousemove="handleMouseMove" @touchmove="handleMouseMove">
 			<!-- Header -->
 			<transition name="fade-visibility">
@@ -256,6 +259,11 @@ export default {
 		canClose: {
 			type: Boolean,
 			default: true
+		},
+		/** Makes the modal backdrop black if true  */
+		dark: {
+			type: Boolean,
+			default: false
 		}
 	},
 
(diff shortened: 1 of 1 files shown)


src/components/Modal/Modal.vue
```diff
@@ -459,7 +467,11 @@ $header-size: 50px;
 	display: block;
 	width: 100%;
 	height: 100%;
-	background-color: rgba(0, 0, 0, .92);
+	background-color: rgba(0, 0, 0, .5);
+}
+
+.dark {
```
skjnldsv: bem `--dark`

src/components/Modal/Modal.vue
```diff
@@ -643,7 +655,7 @@ $header-size: 50px;
 		transition: transform 300ms ease;
 		border-radius: var(--border-radius-large);
 		background-color: var(--color-main-background);
-		box-shadow: 0 2px 8px rgba(0, 0, 0, .33);
+		box-shadow: 0 0 40px rgba(0, 0, 0, .2);
```
skjnldsv: why changing the box shadow?

ma12-co: to make it less harsh, with the opacity at 90% it was not a problem but after bringing that down It was too strong IMO

skjnldsv: Could you add screenshots of the comparison pls :)

ma12-co: Before: 
Screenshot from 2019-11-27 16-06-54

After:
Screenshot from 2019-11-27 16-07-36

src/components/Modal/Modal.vue
```diff
@@ -60,7 +60,10 @@ export default {
 </docs>
 <template>
 	<transition name="fade">
-		<div ref="mask" class="modal-mask" @click="handleMouseMove"
+		<div ref="mask"
+			class="modal-mask"
+			:class="{ 'modal-mask--dark' : dark }"
```
skjnldsv:
```suggestion
			:class="{ 'modal-mask--dark': dark }"
```

PR #751, opened 2019-11-28
fix performance issue in isMobile

### LINKED ISSUE #577: Performance issues in App-Navigation
With #486, there was introduced a huge performance impact. I did some tests with [this App.vue](https://gist.github.com/korelstar/8d6fc42b8d69c2c84a70be298f39b451) (please use it within `vueexample` app). That app creates 200 `AppNavigationItem`s, each with three `ActionButton`s. It provides debug logs in the JavaScript console (please activate timestamps in your browser). After loading the app in your Nextcloud, please click the button "Toggle navigation" in order to show the 200 items and have a look at the console with debug logs.

Test environment:
- Firefox ESR 60.8 (Debian)
- Intel N3160 processor
Result: ~5-6 seconds

@skjnldsv wrote something about performance audits in browser tools. Here is what Firefox gave me:

<details>
<summary>Click to show screenshots</summary>

Screenshot

Screenshot
</details>

It looks like `isMobile` is the origin of the problem. After removing all `isMobile` stuff from `AppNavigationItem`, Firefox show me this:

<details>
<summary>Click to show screenshots</summary>

Screenshot

Screenshot
</details>

This looks much better!

2 seconds are still much, but this is far way better than 6 seconds.

What we have to do:
- [ ] replace `isMobile`by another solution. Maybe this could be realized in this case with CSS only?
- [ ] find more bottlenecks and eliminate them

korelstar:
fixes #577
uses the idea of @paulschwoerer (https://github.com/nextcloud/nextcloud-vue/issues/577#issuecomment-559399080) Thanks!!! :heart:

### DIFF
--- a/src/utils/IsMobileState.js
+++ b/src/utils/IsMobileState.js
@@ -1,7 +1,8 @@
 /**
- * @copyright Copyright (c) 2019 John Molakvoæ <skjnldsv@protonmail.com>
+ * @copyright Copyright (c) 2019 Kristof Hamann, Paul Schwörer
  *
- * @author John Molakvoæ <skjnldsv@protonmail.com>
+ * @author Kristof Hamann
+ * @author Paul Schwörer
  *
  * @license GNU AGPL version 3 or any later version
  *
@@ -20,24 +21,29 @@
  *
  */
 
-import { IsMobileState } from 'Utils/IsMobileState'
+import Vue from 'vue'
 
-export default {
+export const IsMobileState = new Vue({
 	data() {
 		return {
 			isMobile: false
 		}
 	},
-	mounted() {
-		IsMobileState.$on('changed', this.onIsMobileChanged)
-		this.isMobile = IsMobileState.isMobile
+	watch: {
+		isMobile(val) {
+			this.$emit('changed', val)
+		}
+	},
+	created() {
+		window.addEventListener('resize', this.handleWindowResize)
+		this.handleWindowResize()
 	},
 	beforeDestroy() {
-		IsMobileState.$off('changed', this.onIsMobileChanged)
+		window.removeEventListener('resize', this.handleWindowResize)
 	},
 	methods: {
-		onIsMobileChanged(val) {
-			this.isMobile = val
+		handleWindowResize() {
+			this.isMobile = document.documentElement.clientWidth < 1024
 		}
 	}
-}
+})


src/mixins/isMobile/index.js
```diff
@@ -20,26 +20,24 @@
...
 		return {
-			isMobile: this._isMobile()
+			isMobile: false
 		}
 	},
-	beforeMount() {
-		window.addEventListener('resize', this._onResize)
+	mounted() {
+		IsMobileState.$on('changed', this.onIsMobileChanged)
+		this.isMobile = IsMobileState.isMobile
```
paulschwoerer: This line should not be needed, as the initial state is `false` for both the mixin and the resize listener and the 'changed' event will fire if the state is changed to `true`. Also this depends on the internal structure of the `IsMobileState` component, which should be avoided.

korelstar: This is needed if a component using this mixing is created after another component already used the mixin. In that case, `IsMobileState.isMobile` could be already set to `true` and no further event is fired. So we definitely need this!

paulschwoerer: True! Did not think of that. Do you think it would make sense to watch changes to the [$listeners](https://vuejs.org/v2/api/#vm-listeners).changed property, so we can fire the 'changed' event every time a new listener joins? This would remove the coupling between the mixin and the component. Don't know if that's overkill though or would even introduce other side effects.

src/utils/IsMobileState.js
```diff
@@ -0,0 +1,49 @@
+/**
+ * @copyright Copyright (c) 2019 Nextcloud community
```
skjnldsv: please use the original creator of the file :)

src/utils/IsMobileState.js
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @author Kristof Hamann
+ * @author Paul Schwörer
```
skjnldsv: feel free to enter your mail as well in copyrights :)

src/mixins/isMobile/index.js
korelstar: I'm not sure, I think it's fine as it is. What do you think, @skjnldsv ?

src/utils/IsMobileState.js
korelstar: done :-)

src/mixins/isMobile/index.js
skjnldsv: I think we should try as much as possible to mimic the current behaviour which works very well (despite the performance issue) :)

korelstar: Ok, this means LGTM?

skjnldsv: If it works, yes? :p

PR #764, opened 2019-12-10
Mobile fixes for Viewer header

jancborchardt:
Fix position of action elements by moving the sidebar toggle to the right, and fix title overlap on mobile screens.

## Before
Viewer header before

## After
Viewer header after

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -499,7 +499,7 @@ $header-size: 50px;
 		overflow-x: hidden;
 		box-sizing: border-box;
 		width: 100%;
-		padding: 0 #{$clickable-area * 3} 0 12px; // maximum actions is 2 but we violate it here
+		padding: 0 #{$clickable-area * 3} 0 12px; // maximum actions is 3
 		transition: padding ease 100ms;
 		white-space: nowrap;
 		text-overflow: ellipsis;
@@ -511,7 +511,7 @@ $header-size: 50px;
 	@media only screen and (min-width: $breakpoint-mobile/2) {
 		.modal-title {
 			text-align: center;
-			padding-left: #{$clickable-area * 3}; // maximum actions is 2 but we violate it here
+			padding-left: #{$clickable-area * 3}; // maximum actions is 3
 		}
 	}
 
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -74,12 +74,6 @@ export default {
 						{{ title }}
 					</div>
 					<div class="icons-menu">
-						<!-- Actions menu -->
-						<Actions class="header-actions">
-							<!-- @slot List of actions to show -->
-							<slot name="actions" />
-						</Actions>
-
 						<!-- Play-pause toggle -->
 						<button v-if="hasNext && enableSlideshow"
 							v-tooltip.auto="playPauseTitle"
@@ -102,6 +96,12 @@ export default {
 									r="15" cx="25" cy="25" />
 							</svg>
 						</button>
+
+						<!-- Actions menu -->
+						<Actions class="header-actions">
+							<!-- @slot List of actions to show -->
+							<slot name="actions" />
+						</Actions>
 
 						<!-- Close modal -->
 						<Actions v-if="canClose" class="header-close">
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -498,13 +498,21 @@ $header-size: 50px;
 	.modal-title {
 		overflow-x: hidden;
 		box-sizing: border-box;
-		max-width: 100%;
-		padding: 0 #{$clickable-area * 2}; // maximum actions is 2
+		width: 100%;
+		padding: 0 #{$clickable-area * 3} 0 12px; // maximum actions is 2 but we violate it here
 		transition: padding ease 100ms;
 		white-space: nowrap;
 		text-overflow: ellipsis;
 		color: #fff;
 		font-size: $icon-margin;
+	}
+
+	// On wider screens the title can be centered
+	@media only screen and (min-width: $breakpoint-mobile/2) {
+		.modal-title {
+			text-align: center;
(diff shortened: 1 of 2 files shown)


src/components/Modal/Modal.vue
```diff
@@ -498,15 +498,23 @@ $header-size: 50px;
 	.modal-title {
 		overflow-x: hidden;
 		box-sizing: border-box;
-		max-width: 100%;
-		padding: 0 #{$clickable-area * 2}; // maximum actions is 2
+		width: 100%;
+		padding: 0 #{$clickable-area * 3} 0 12px; // maximum actions is 2 but we violate it here
```
skjnldsv: what is this 12 px?

skjnldsv: "but we violate it here" also, what is this? :p

src/components/Modal/Modal.vue
```diff
@@ -498,15 +498,23 @@ $header-size: 50px;
...
 		transition: padding ease 100ms;
 		white-space: nowrap;
 		text-overflow: ellipsis;
 		color: #fff;
 		font-size: $icon-margin;
 	}
 
+	// On wider screens the title can be centered
+	@media only screen and (min-width: $breakpoint-mobile/2) {
```
skjnldsv: half the mobile breakpoint?

jakobroehrl:
How can I test this?
Do I have to pull the nextcloud-vue repo and checkout this branch?

src/components/Modal/Modal.vue
jancborchardt: > what is this 12 px?

Just a padding we often use, and based on a scale of 4, which was proposed by Jenni. Yes it needs standardization. Something for a separate pull request though as it requires more work, but this is an urgent fix.



> "but we violate it here" also, what is this? :p

Just to clarify that normally it should be 2 actions max. ;) And it literally had the *2 in there, but was showing 3 icons.
In this case actually 3 is fine though cause the close X is extra.

src/components/Modal/Modal.vue
jancborchardt: Yes, cause it looks bad to already do it at the mobile breakpoint. Test and you will see. :)

jancborchardt:
> How can I test this?
> Do I have to pull the nextcloud-vue repo and checkout this branch?

Yup! :) Just follow the instructions in the development setup docs at https://github.com/nextcloud/nextcloud-vue#development-setup – the app you need to do the `npm link @nextcloud/vue` in is `apps/viewer` in the server. :)

src/components/Modal/Modal.vue
skjnldsv: I understand, I wanted to make sure this was not a mistake :+1:

src/components/Modal/Modal.vue
skjnldsv: > Just a padding we often use, and based on a scale of 4, which was proposed by Jenni. Yes it needs standardization. Something for a separate pull request though as it requires more work, but this is an urgent fix.

Then please use $icon-margin as this si the one we use in viewer.
We never use random px values here :)

> And it literally had the *2 in there, but was showing 3 icons.

Yeah, because the close icon is not in the viewer.
Please put back the original comment as this one doesn't help :p

skjnldsv: > Yeah, because the close icon is not in the viewer.

actually, wrong answer, this is because the slideshow is not an `Actions`

jancborchardt:
> @jancborchardt shouldn't we have more padding to the left of the play button?

Was thinking about that as well, but tbh the overlap is not bad, and on mobile you need the space?

src/components/Modal/Modal.vue
jancborchardt: > Then please use $icon-margin as this is the one we use in viewer.

I was looking at icon-margin, but this has nothing to do with icons?

Also, regarding the random px values, see https://github.com/nextcloud/nextcloud-vue/pull/764/files#diff-a25f5501f9ebbd73f59b3335f9545c4bR530 ;)

jancborchardt: @skjnldsv please add a commit with the changes you want, or propose them as suggestions in commits. It’s difficult to keep track otherwise.

skjnldsv: > Also, regarding the random px values, see [nextcloud/nextcloud-vue/pull/764/files#diff-a25f5501f9ebbd73f59b3335f9545c4bR530](https://github.com/nextcloud/nextcloud-vue/pull/764/files#diff-a25f5501f9ebbd73f59b3335f9545c4bR530) ;)

AAArgh :grin: 
/me shivers

src/components/Modal/Modal.vue
```diff
@@ -498,15 +498,23 @@ $header-size: 50px;
...
+		.modal-title {
+			text-align: center;
+			padding-left: #{$clickable-area * 3}; // maximum actions is 2 but we violate it here
```
skjnldsv:
```suggestion
			padding-left: #{$clickable-area * 3}; // maximum actions is 3
```

src/components/Modal/Modal.vue
skjnldsv:
```suggestion
		padding: 0 #{$clickable-area * 3} 0 12px; // maximum actions is 3
```

skjnldsv: > I was looking at icon-margin, but this has nothing to do with icons?

it's indexed on the clickable area :)

jancborchardt:
Applied your suggestions @skjnldsv – ready to go now? :)

skjnldsv:
> ready to go now? :)

where are we going? :D

jancborchardt:
> > ready to go now? :)
> 
> where are we going? :D
> :framed_picture: :sun_with_face: :beach_umbrella:

→ https://en.wikivoyage.org/wiki/Yogyakarta come over Monsieur ;)
Not so much :beach_umbrella: but lots of :sun_with_face: and also :motor_scooter: 

Fixed the DCO sign-off issue too, should run through now.

PR #766, opened 2019-12-10
Fix Modal backdrop

ma12-co:
As discussed in the design group: bring opacity back to .92!

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -469,7 +469,7 @@ $header-size: 50px;
 	height: 100%;
 	background-color: rgba(0, 0, 0, .5);
 	&--dark {
-		background-color: black;
+		background-color: rgba(0, 0, 0, .92);
 	}
 }
 


jancborchardt:
Also heeeh – I could do it similarly to add a `main-background` prop, so the background is light by default for the Text app? (And icons dark) Seems pretty good, cause I was about to do that quite hackily …

ma12-co:
What do you mean @jancborchardt ? The background is already light by default: https://github.com/nextcloud/nextcloud-vue/blob/323929c2b08bb911419b78b68a96942cbc23bd76/src/components/Modal/Modal.vue#L266
I haven't touched the icons though

jancborchardt:
@ma12-co with "light" I mean "#fff" (that is `var(--color-main-background)`), not "50% transparent black". :)

skjnldsv:
@jancborchardt you should not use the variables here as it will be reverted in dark mode ;)

jancborchardt:
> @jancborchardt you should not use the variables here as it will be reverted in dark mode ;)

The Text app background will also be inverted in dark mode, that’s why I do want to use the variables. :)

skjnldsv:
@jancborchardt then I misunderstood, I thought you were refering to the modal dark bg

jancborchardt:
Just cc'ing @juliushaertl for info when he’s back from vacation cause this is the way we should probably do the Text app background fix so it’s white. :)

PR #772, opened 2019-12-16
Update vue2-datepicker to current version

dartcafe:
The vue2-datepicker ist outdated and so the documentation. Would be great, if we could update this, but some props  (especially the lang settings) have changed.


skjnldsv:
Yes https://github.com/nextcloud/nextcloud-vue/pull/762 :p

dartcafe:
Ehm. Oops.

skjnldsv:
I mean, there is not much work done on the others than here :p
We keep postponning as no one have time for it :(

PR #831, opened 2020-01-22
Use the popover component for the avatar menu

ma12-co:

fix https://github.com/nextcloud/server/issues/18634

### DIFF
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -55,9 +55,7 @@
 			v-if="hasMenu"
 			placement="auto"
 			:open="contactsMenuOpenState">
-			<template>
-				<PopoverMenu :menu="menu" />
-			</template>
+			<PopoverMenu :menu="menu" />
 			<template slot="trigger">
 				<div :class="contactsMenuLoading ? 'icon-loading' : 'icon-more'" :style="{'width': size + 'px', 'height': size + 'px'}" />
 			</template>
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -52,13 +52,14 @@
 			:srcset="avatarSrcSetLoaded"
 			alt="">
 		<Popover
+			v-if="hasMenu"
 			placement="auto"
 			:open="contactsMenuOpenState">
 			<template>
-				<PopoverMenu :is-open="contactsMenuOpenState" :menu="menu" />
+				<PopoverMenu :menu="menu" />
 			</template>
 			<template slot="trigger">
-				<div v-if="hasMenu" class="icon-more" :style="{'width': size + 'px'}" />
+				<div class="icon-more" :style="{'width': size + 'px'}" />
 			</template>
 		</Popover>
 
@@ -600,11 +601,11 @@ export default {
 			display: flex;
 			cursor: pointer;
 			opacity: 0;
-			cursor: pointer;
 			background: none;
 			font-size: 18px;
 			align-items: center;
 			justify-content: center;
+
 			@include iconfont('more');
 			&::before {
 				display: block;
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -87,7 +87,7 @@
 			:open="contactsMenuOpenState">
 			<template>
 				<ul>
-					<PopoverMenuItem v-for="(item, key) in contactsMenuActions" :key="key" :item="item" />
+					<PopoverMenuItem v-for="(item, key) in menu" :key="key" :item="item" />
 				</ul>
 				<p>Dummy content</p>
 			</template>
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -51,7 +51,16 @@
 			:src="avatarUrlLoaded"
 			:srcset="avatarSrcSetLoaded"
 			alt="">
-		<div v-if="hasMenu" class="icon-more" />
+		<Popover
+			placement="auto"
+			:open="contactsMenuOpenState">
+			<template>
+				<PopoverMenu :is-open="contactsMenuOpenState" :menu="menu" />
+			</template>
+			<template slot="trigger">
+				<div v-if="hasMenu" class="icon-more" :style="{'width': size + 'px'}" />
+			</template>
+		</Popover>
 
 		<!-- Avatar status -->
 		<div v-if="showUserStatusIconOnAvatar" class="avatardiv__user-status avatardiv__user-status--icon">
@@ -82,22 +91,13 @@
 		<div v-if="userDoesNotExist && !iconClass" class="unknown">
 			{{ initials }}
 		</div>
-		<Popover
-			placement="auto"
-			:open="contactsMenuOpenState">
-			<template>
-				<ul>
-					<PopoverMenuItem v-for="(item, key) in menu" :key="key" :item="item" />
-				</ul>
-				<p>Dummy content</p>
-			</template>
-		</Popover>
 	</div>
 </template>
 
 <script>
 import { getBuilder } from '@nextcloud/browser-storage'
 import { directive as ClickOutside } from 'v-click-outside'
+import PopoverMenu from '../PopoverMenu'
 import { getCurrentUser } from '@nextcloud/auth'
 import { subscribe, unsubscribe } from '@nextcloud/event-bus'
 import axios from '@nextcloud/axios'
@@ -106,7 +106,6 @@ import Tooltip from '../../directives/Tooltip'
 import usernameToColor from '../../functions/usernameToColor'
 import { userStatus } from '../../mixins'
 import Popover from '../Popover/Popover'
-import PopoverMenuItem from '../PopoverMenu/PopoverMenuItem'
 
 const browserStorage = getBuilder('nextcloud').persist().build()
 
@@ -132,7 +131,7 @@ export default {
 	},
 	components: {
 		Popover,
-		PopoverMenuItem,
+		PopoverMenu,
 	},
 	mixins: [userStatus],
 	props: {
@@ -592,20 +591,20 @@ export default {
 
 	&--with-menu {
 		cursor: pointer;
-		.icon-more {
+		::v-deep .trigger {
 			position: absolute;
 			top: 0;
 			left: 0;
+		}
+		.icon-more {
 			display: flex;
-			align-items: center;
-			justify-content: center;
-			width: inherit;
-			height: inherit;
 			cursor: pointer;
 			opacity: 0;
+			cursor: pointer;
 			background: none;
 			font-size: 18px;
-
+			align-items: center;
+			justify-content: center;
 			@include iconfont('more');
 			&::before {
 				display: block;
@@ -717,12 +716,6 @@ export default {
 	.popovermenu-wrapper {
 		position: relative;
 		display: inline-block;
-	}
-
-	.popovermenu {
-		display: block;
-		margin: 0;
-		font-size: var(--default-font-size);
 	}
 }
 
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -82,21 +82,22 @@
 		<div v-if="userDoesNotExist && !iconClass" class="unknown">
 			{{ initials }}
 		</div>
-
-		<!-- Menu container -->
-		<div v-if="hasMenu"
-			v-show="contactsMenuOpenState"
-			class="popovermenu"
-			:class="`menu-${menuPosition}`">
-			<PopoverMenu :is-open="contactsMenuOpenState" :menu="menu" />
-		</div>
+		<Popover
+			placement="auto"
+			:open="contactsMenuOpenState">
+			<template>
+				<ul>
+					<PopoverMenuItem v-for="(item, key) in contactsMenuActions" :key="key" :item="item" />
+				</ul>
+				<p>Dummy content</p>
+			</template>
+		</Popover>
 	</div>
 </template>
 
 <script>
 import { getBuilder } from '@nextcloud/browser-storage'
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -39,7 +39,7 @@
 		"md5": "^2.2.1",
 		"regenerator-runtime": "^0.13.3",
 		"v-click-outside": "^3.0.0",
-		"v-tooltip": "^2.0.0-rc.33",
+		"v-tooltip": "^2.0.3",
```
skjnldsv: unrelated?

src/components/Avatar/Avatar.vue
```diff
@@ -65,22 +65,27 @@
...
 
 <script>
 
 /* global OC oc_userconfig */
 import { directive as ClickOutside } from 'v-click-outside'
-import { PopoverMenu } from 'Components/PopoverMenu'
 import axios from '@nextcloud/axios'
 import Tooltip from 'Directives/Tooltip'
 import uidToColor from './uidToColor'
+import Popover from '../Popover/Popover'
```
skjnldsv:
```suggestion
import Popover from 'Components/Popover'
```

src/components/Avatar/Avatar.vue
```diff
@@ -65,22 +65,27 @@
...
+import PopoverMenuItem from '../PopoverMenu/PopoverMenuItem'
```
skjnldsv:
```suggestion
import PopoverMenuItem from 'Components/PopoverMenu/PopoverMenuItem'
```

package.json
ma12-co: yeah totally

src/components/Avatar/Avatar.vue
```diff
@@ -65,22 +65,27 @@
 		<div v-if="userDoesNotExist" class="unknown">
 			{{ initials }}
 		</div>
-		<div v-if="hasMenu" v-show="contactsMenuOpenState"
-			class="popovermenu"
-			:class="`menu-${menuPosition}`">
-			<PopoverMenu :is-open="contactsMenuOpenState" :menu="menu" />
-		</div>
+		<Popover v-if="hasMenu" :show="contactsMenuOpenState">
```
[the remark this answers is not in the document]
ma12-co: That was it, thanks :)

skjnldsv:
Any news here Marco?

ma12-co:
This needs #832 to be finished first

package.json
georgehrke: Still needs to be removed

src/components/Avatar/Avatar.vue
```diff
@@ -65,22 +65,31 @@
...
 			{{ initials }}
 		</div>
-		<div v-if="hasMenu" v-show="contactsMenuOpenState"
-			class="popovermenu"
-			:class="`menu-${menuPosition}`">
-			<PopoverMenu :is-open="contactsMenuOpenState" :menu="menu" />
-		</div>
+		<Popover
+			v-if="hasMenu"
+			placement="auto"
```
georgehrke, line 70: Should this be influenceable from the outside component?

skjnldsv: Don't think so

MorrisJobke:
@ma12-co 🏓

ma12-co:
Pong!

src/components/Avatar/Avatar.vue
```diff
@@ -77,24 +76,29 @@
...
-import PopoverMenu from '../PopoverMenu'
-import { getCurrentUser } from '@nextcloud/auth'
 import axios from '@nextcloud/axios'
+import { getCurrentUser } from '@nextcloud/auth'
 import { generateUrl } from '@nextcloud/router'
 import Tooltip from '../../directives/Tooltip'
 import usernameToColor from '../../functions/usernameToColor'
 import { userStatus } from '../../mixins'
+import Popover from '../Popover/Popover'
+import PopoverMenuItem from '../PopoverMenu/PopoverMenuItem'
```
skjnldsv, line 101: Not sure we want to keep this PopoverMenuItem component. What about Actions?

ma12-co: Ideally yes! Followup? 
This should actually fix many placement bugs for now

skjnldsv (dismissed):

skjnldsv:
@ma12-co rebase?

ma12-co:
Rebased locally but it seems that it doesn't work anymore

PVince81:
also hoping this could magically fix https://github.com/nextcloud/spreed/issues/4147 since it changes the DOM structure

skjnldsv:
Ping?

PVince81:
Indeed, broken after rebase:
image

Maybe because of the recent Popover changes ?

PVince81:
Yep, indeed.

Now the popover is appended into the body, which is good news as it doesn't overlap so would fix https://github.com/nextcloud/spreed/issues/4147

However, all the styles relative to "avatardiv" are not applied any more since it's not inside the avatar div any more, so will need to shift around many of the styles.

I'll push what I have now as I have a local fix for the menu items list.

PVince81:
Current state:
image

As you can see, no more overlap with avatar icon :tada: 

- [ ] missing display:flex on "ul" element / fix icon size
- [ ] remove dummy debug content
- [ ] check what styles to reuse/keep

Regarding styles, I think none of them is applyable from inside the Avatar component since they would need to target an element in the body. I'd expect such styles to already be defined on the Popover component itself and the other ones to be obsolete anyway.

PVince81:
I managed to make it work, current state:
image

image
(here automatic positioning seems to have decided to appear on top, probably because I was close to the bottom of the window)

Had to apply styles on the "trigger" from popover because for whatever reason it insisted on positioning itself a few rows lower.

So far it seems to work fine except for one detail: upon first rendering, the popover appears a few rows lower and then quickly moves to the correct position.

- [ ] debug popover initial position glithc

PVince81:
strangely this popover moving issues reminds me of https://github.com/nextcloud/nextcloud-vue/issues/1610
as if maybe the v-tooltip itself has some kind of animation where it would slide the popover to its destination somehow

PVince81:
a possible clue here:
image

a transform3d was applied on the element itself and is expecting an animation on the transform property.

I grepped for this in nextcloud-vue and v-tooltip, but nothing found.
So this could be in popper.js which is used by v-tooltip.

PVince81:
and here we go, it's here... https://github.com/popperjs/popper-core/blob/v1.16.1/packages/popper/src/modifiers/computeStyle.js#L88

I'll try to find a way to disable it.

Otherwise we'll need to hack it to hide the element visually until the animation delay is finished...

PVince81:
I managed to make it not use transform3d by setting `Popper.Defaults.modifiers.computeStyle.gpuAcceleration = false` in v-tooltip's Popper instance.

However now it's using "top" and "left" for positioning and I still see it appear first at the wrong offset and fixing itself immediately.
So this is not a transition problem. There must be some other code that recalculates the position.
I did add logging in v-tooltip's delayed methods but they never fire.

PVince81:
aha, it seems Popper's style routine runs twice with different coordinates:
image

this is only the first time a specific tooltip/popover is open.
the second time the position stays the same so there is no such log entry any more.

now to find out in what level this redundancy is happening...

PVince81:
Looking at the stack traces of both, the first one is just triggered by `_scheduleShow` of "v-tooltip"
but the second one is triggered by v-tooltip's directive's `_handleResize` from a ResizeObserver from vue-resize.

Okay, interesting. So if I disable the resize handler by setting `:handle-resize="false"` in the `VPopover` element, then it looks like this:
image

One thing I did notice with a breakpoint was that if I stopped on the resize handler, only the arrow (shifted) was visible, not the contents of the tooltip. So maybe whenever the tooltip contents changes, it will fire that event and then reposition the tooltip/popover.

While the latter is nice, it is using `window.requestAnimationFrame` before doing an update, so it makes the shifting visible.

Next up to find out what exactly is being resized and seeing if we can have it populated earlier so that no resize happens initially...

PVince81:
Okay, I found the culprit: when clicking, first the popover will open with an empty contents because at this time the contacts menu is not loaded yet. Only then once loaded, it will render into the contents which triggers a resize.

Next up: to find a way to not display the popover while loading.

PVince81:
It's worse: even if I numb "toggleMenu", it seems the click on the trigger will still trigger the v-tooltip, even though "VPopover" has "open=false". So it seems more work is needed here to get all this under control and avoid redundancy.

PVince81:
Okay, so I managed to fix the flying arrow issue from https://github.com/nextcloud/nextcloud-vue/issues/1610 by making the popover invisible until "open" is actually true.

While this works for the call view popover like in https://github.com/nextcloud/nextcloud-vue/issues/1610, it doesn't help with the avatar menu

The problem with the avatar menu it seems to be that whenever the menu is loaded and "open" is set to true, the v-tooltip will already render the contents, but in the wrong location. And this triggers the resize handler which will put the tooltip at the correct position.

So next would be to try to find a way to pre-render the tooltip offscreen while opening the first time, and then letter popper reposition it...

The interesting thing now is that opening the menu will now always show it shifted first. Even when reopening. Before these changes it would only do so on the first open.

Needs further research... :exploding_head:

ma12-co:
Thanks for taking over @PVince81 :)

PVince81:
I've tried to add a loading spinner "fake menu entry" to make sure we got some initial contents in the tooltip. This approach makes it possible for the tooltip to be directly positioned in the right place. Then when the single menu entry loads it doesn't move.
So far so good.
However, as soon as there are multiple menu entries, the popover will jump again after loading.
Furthermore, even with a single one, the quick switch between loading spinner is annoying (that menu loads quickly usually).

Yet another approach I found (without loading spinner) was to add an additional delay before actually showing the tooltip, to let the libraries time to recompute its position before it actually becomes visible. Basically a delay between setting `:open` to `true` and removing an extra `visibility: hidden` from the tooltip. I guess I'll go with the latter now as I feel we've lost enough time with this mess.

PVince81:
Ok, the fix is in with "visibility: hidden" with a delay.

I've also added a loading spinner on the trigger, which lets the user know that something is loading so that the UI doesn't feel sluggish while loading the menu.

Now while the avatar menus work fine now, there's still a problem with the popover from the call view: when you click the actions menu at the bottom, the popover does appear and causes a scrollbar glitch. It seems the "__hidden" class is not enough and needs to be applied on the popover element itself instead.

PVince81:
Okay, all works now...

Tested with Talk:
- [x] avatar menu in messages list: spinner appears, menu appears directly in correct position
- [x] avatar menu in participants list: spinner appears, menu appears directly in correct position
- [x] popover top menu in callview appears with no arrow animation glitch
- [x] popover bottom menu in callview appears with no arrow animation glitch, and no scrollbar glitch (scrollbar was appearing for a millisecond then disappearing)

@ma12-co can you test and review ?

PVince81:
I know, the hack is ugly but that's the only way I found to work around the fact that the library will display the tooltip first before repositioning, instead of showing it off screen, evaluating the size, and then repositioning. Maybe there are better ways that would require patching the libs. I saw there will be new versions of the libs (v-tooltip and popper.js), something to look into separately.

PVince81:
Now a question is still in the air: why use the Popover directly and not use Actions instead ? The latter already has keyboard navigation and everything needed to make it accessible :thinking:

PVince81:
> Now a question is still in the air: why use the Popover directly and not use Actions instead ? The latter already has keyboard navigation and everything needed to make it accessible thinking

I had a quick try with Actions and it comes with additional challenges, and the hack for preventing the repositioning glitch doesn't work there: https://github.com/nextcloud/nextcloud-vue/pull/1670

ma12-co (commented):
Tested and works, cannot approve as I'm the original author :sweat_smile:

src/components/Avatar/Avatar.vue
```diff
@@ -51,7 +51,17 @@
 			:src="avatarUrlLoaded"
 			:srcset="avatarSrcSetLoaded"
 			alt="">
-		<div v-if="hasMenu" class="icon-more" />
+		<Popover
+			v-if="hasMenu"
+			placement="auto"
+			:open="contactsMenuOpenState">
+			<template>
```
ChristophWurst, line 58: this template element seems redundant, the could should work without if I'm not mistaken

src/components/Popover/Popover.vue
```diff
@@ -74,6 +74,32 @@ export default {
...
+		}
+	},
+
+	watch: {
+		open(newValue) {
+			if (newValue === true) {
+				// show with delay to allow position calculation to happen for dynamic contents,
+				// otherwise the popover will appear in the wrong place first and be visible,
+				// and then in the next tick it repositions itself
+				window.setTimeout(() => {
```
ChristophWurst, line 95: https://vuejs.org/v2/api/#vm-nextTick?

ChristophWurst (approved):
Minor remarks, looks good otherwise :+1:

src/components/Popover/Popover.vue
PVince81: nope, sadly next tick happens too early...

PVince81: I'll add this in the comment so the next dev won't be wondering

PVince81:
Crap, I retested this and I still see the position shifting glitch from time to time... Might be a race against v-tooltip's timers...
Anyway, at least the latest hacks reduce the probability.

Eventually we'll need to look into upgrading v-tooltip/popper.js, maybe patching it to make it work properly...

src/components/Avatar/Avatar.vue
PVince81: removed now

PVince81 (approved):
Approving on behalf of @ma12-co

PVince81:
regression: https://github.com/nextcloud/nextcloud-vue/issues/1681

PR #832, opened 2020-01-24
Feature/641/use popover component for actions

### LINKED ISSUE #641: Use the Popover component for Actions
I think we should use the newly created Popover component to wrap the actions in. The new component automatically positions itself so it's visible and also reacts to window resizes.

ma12-co:
fix #641

### DIFF
--- a/l10n/messages.pot
+++ b/l10n/messages.pot
@@ -10,7 +10,7 @@ msgstr ""
 msgid "{tag} (restricted)"
 msgstr ""
 
-#: src/components/Actions/Actions.vue:249
+#: src/components/Actions/Actions.vue:246
 msgid "Actions"
 msgstr ""
 
--- a/src/components/Actions/Actions.vue
+++ b/src/components/Actions/Actions.vue
@@ -101,59 +101,64 @@ https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-action
 	<!-- more than one action -->
 	<div v-else
 		v-show="hasMultipleActions || forceMenu"
-		ref="fullmenu"
-		v-click-outside="closeMenu"
 		:class="{'action-item--open': opened}"
-		class="action-item"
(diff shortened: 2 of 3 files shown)


src/assets/action.scss
```diff
@@ -24,7 +24,7 @@
 @mixin action-active {
 	li {
 		&.active {
-			box-shadow: inset 4px 0 var(--color-primary);
```
skjnldsv: please only do the POpover change, keep things separate otherwise it's complcated to review. If you plan to ask for a design change, this should be done in another place :)

src/assets/action.scss
```diff
@@ -71,6 +70,8 @@
 		&:hover,
 		&:focus {
 			opacity: $opacity_full;
+			background-color: var(--color-background-hover);
```
skjnldsv: same here.

skjnldsv:
Any news here Marco?

georgehrke:
Needs a rebase

skjnldsv:
So, one issue is Actions inside a menu.
Since the menu is now out of the component, it is dedected by the click-outside directive and close everything on click.

Anyone have any idea?

georgehrke:
> So, one issue is Actions inside a menu.
Since the menu is now out of the component, it is dedected by the click-outside directive and close everything on click.

We can probably use https://github.com/nextcloud/nextcloud-vue/pull/983 and whitelist the popover class?

skjnldsv:
> We can probably use #983 and whitelist the popover class?

Would be great yes! :)

skjnldsv:
So, I'm close to it, but I'm stuck.
It focus the first entry on open, but never the first time it is opened :thinking: 
Anyone wanna help?

georgehrke (approved):
Minor issue:
- The popover doesn't automatically close when tabbing over the last item

PR #839, opened 2020-01-27
Use github actions for linting and building

skjnldsv:


skjnldsv:
The only issue I have with this is that we need to update the node versioning manually. It's no different to how it was with travis, But I realise it's not the best :)
Any thoughts?

PR #840, opened 2020-01-27
Fix Tab default values

skjnldsv:
After #833 

```
vue.runtime.esm.js:1888 TypeError: Cannot read property 'name' of undefined
    at a.default (AppSidebarTab.js:formatted:1)
    at vue.runtime.esm.js:1666
    at Pt (vue.runtime.esm.js:1619)
    at i (vue.runtime.esm.js:4665)
    at vue.runtime.esm.js:4698
    at gn (vue.runtime.esm.js:4639)
    at a.t._init (vue.runtime.esm.js:5006)
    at new a (vue.runtime.esm.js:5154)
    at vue.runtime.esm.js:3283
    at init (vue.runtime.esm.js:3114)
```

### DIFF
--- a/src/components/AppSidebarTab/AppSidebarTab.vue
+++ b/src/components/AppSidebarTab/AppSidebarTab.vue
@@ -41,22 +41,19 @@ export default {
 	props: {
 		id: {
 			type: String,
-			default: () => this.name.toLowerCase().replace(/ /g, '-')
+			required: true
 		},
 		name: {
 			type: String,
-			default: '',
 			required: true
 		},
 		icon: {
 			type: String,
-			default: '',
 			required: true
 		},
 		order: {
 			type: Number,
-			default: 0,
-			required: false
+			default: 0
 		}
 	},
 


juliushaertl:
Hm, I tough I could work around changing the API, i thought this has worked when testing back then.

georgehrke:
This would introduce a breaking change that `AppSidebarTab` requires id from now on.
What about having id as optional property and using a computed property to returns id or if its `null` use `this.name.toLowerCase().replace(/ /g, '-')`?

skjnldsv:
> This would introduce a breaking change that `AppSidebarTab` requires id from now on.
> What about having id as optional property and using a computed property to returns id or if its `null` use `this.name.toLowerCase().replace(/ /g, '-')`?

I would say no because using the name (which is translated in every nc app) is definitely not a good thing and will for sure break.
Let's force id and make sure devs knows they have to provide it (it's just a string, let's not over react, :grin: )

ChristophWurst:
I agree that this is a breaking change. You can't use this new code without modifying your app's code to add an `id`.

> Let's force id and make sure devs knows they have to provide it

Yes, with a major version bump. Otherwise they might assume their existing code is still compatible.

ChristophWurst:
Why was this breaking change released as a minor update? https://github.com/nextcloud/nextcloud-vue/releases/tag/v1.3.0

ChristophWurst:
As the majority of packages follow semver, I don't test minor bumps. Therefore I've set Dependabot to merge minor releases automatically. If I don't have tests that cover the usage of this changed component it will happily merge the PR and I have a broken app.

Please let's stick to semantical versions. I don't want to have to check the full change log for every version bump. If we have a breaking change so be it. Then app devs have to fix them. There is no real need to stick to the same major version for ages.

skjnldsv:
Yeah, we never did it that well on the components. I'll try to be more careful

PR #844, opened 2020-01-31
Only fire Multiselect input function once

raimund-schluessler:
This PR aims to fix the double call of the input callback for the Multiselect. I'd like to add a jest test for this as well.

Since jest is not configured yet for this repo, I wonder whether I should fix the Multiselect bug in a different issue and only set up jest here?

Locally, jest runs fine already, but it's not run by the Github actions yet.

### DIFF
--- a/.github/workflows/npm-test.yml
+++ b/.github/workflows/npm-test.yml
@@ -7,18 +7,18 @@ on:
       - stable*
 
 jobs:
-  build:
+  jest:
     runs-on: ubuntu-latest
     strategy:
       matrix:
         node-version: [11.x]
     steps:
       - uses: actions/checkout@v1
-      - name: Use Node.js ${{ matrix.node-version }}
+      - name: Setup node ${{ matrix.node-version }}
         uses: actions/setup-node@v1
         with:
           node-version: ${{ matrix.node-version }}
-      - name: npm install
+      - name: Install dependencies
         run: npm ci
       - name: Test
         run: npm run test
--- a/.github/workflows/npm-test.yml
+++ b/.github/workflows/npm-test.yml
@@ -1,0 +1,26 @@
+name: Test
+
+on:
+  pull_request:
+    branches:
+      - master
+      - stable*
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [11.x]
+    steps:
+      - uses: actions/checkout@v1
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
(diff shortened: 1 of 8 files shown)


raimund-schluessler:
I now added a test which reproduces the issue that the input function is called twice. Will proceed to fix the actual issue now.

raimund-schluessler:
@nextcloud/vuejs @skjnldsv I set up the jest test environment to be run with GitHub actions, added a test to check for the bug (multiselect input called twice) and fixed the bug. The added jest test would fail before this PR, but passes now. I also tested if the change fixes https://github.com/nextcloud/tasks/issues/778 and it works fine there.

Would be great to get a review, so we can get the fix in.

.github/workflows/npm-test.yml
```diff
@@ -0,0 +1,26 @@
...
+    strategy:
+      matrix:
+        node-version: [11.x]
+    steps:
+      - uses: actions/checkout@v1
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: npm install
```
skjnldsv: This is stupid, but can I ask if we can have the same case for the tests names? :see_no_evil: 
image

```suggestion
      - name: Install dependencies
```

.github/workflows/npm-test.yml
```diff
@@ -0,0 +1,26 @@
+name: Test
+
+on:
+  pull_request:
+    branches:
+      - master
+      - stable*
+
+jobs:
+  build:
```
skjnldsv:
```suggestion
  jest:
```

.github/workflows/npm-test.yml
```diff
@@ -0,0 +1,26 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [11.x]
+    steps:
+      - uses: actions/checkout@v1
+      - name: Use Node.js ${{ matrix.node-version }}
```
skjnldsv:
```suggestion
      - name: Setup node ${{ matrix.node-version }}
```

.github/workflows/npm-test.yml
raimund-schluessler: Fine with me, I also wondered how I should name the tests 😃

skjnldsv: I standardised it for server/contacts & photos so far :woman_shrugging: 
https://github.com/nextcloud/photos/tree/master/.github/workflows

raimund-schluessler: Then I guess we need another PR for nextcloud-vue 😉

PR #848, opened 2020-02-03
Adjust test syntax

skjnldsv:
So apparently CI is not really needed (not used anywhere on https://github.com/actions/setup-node)

I moved everything to a proper matrix so we get the same syntax over our tests

What do you think?


skjnldsv:
The merging broke the pr :p
Let's create a new one

PR #854, opened 2020-02-13
Add gettext-based translations

### LINKED ISSUE #97: Add translation plugin
Should we add translated string in this lib or make the devs add their own?
I like the idea of a standard coming directly from here.
It would probably be stuff like "No more results" or "An error has occurred" I guess. No big sentences, but still.

We should make sure developpers can dynamically import stuff from us so they don't load every language then :)

Any opinions?
@nextcloud/vue

ChristophWurst:
Fixes #97 

Todo

- [x] Test with a real app
- [x] Test placeholder replacement (not sure if `%s` or `{placeholder}` should be used
- [ ] Set up https://docs.transifex.com/transifex-github-integrations/github-tx-ui


build/extract-l10n.js
```diff
@@ -0,0 +1,25 @@
...
+            arguments: {
+                text: 0,
+                context: 1
+            }
+        }),
+        JsExtractors.callExpression('n', {
+            arguments: {
+                text: 1,
+                textPlural: 2,
+                context: 3
```
nickvergessen: Actually count js 3 and context is 4

ChristophWurst: Yep, count is definitely 3. I think we don't even use context. At least not for `ngettext` https://github.com/alexanderwallin/node-gettext#Gettext

ChristophWurst:
> Test placeholder replacement (not sure if `%s` or `{placeholder}` should be used

We'll eventually need this for every package that is translated, so I added this functionality to the l10n package: https://github.com/nextcloud/nextcloud-l10n/pull/88

nickvergessen:
> Test placeholder replacement (not sure if %s or {placeholder} should be used

JS uses `{}` + `%n`
PHP uses `%s`, `%1$s` + `%n`

Ref https://github.com/nextcloud/documentation/pull/1827

ChristophWurst:
> JS uses `{}` + `%n`

Done via https://github.com/nextcloud/nextcloud-l10n/pull/88. Please have a look :)

PR #896, opened 2020-03-02
Fix appnavigationitem's active state

ma12-co:
### How to test:
1) Open an app that uses the appnavigationitem component;
2) Click on a navigation item (the .active class should be added to the wrapping `<li>`);

### Before this pr: 
The item was not highlighted in light primary color.

### After this pr: 
The item is properly highlighted.


juliushaertl:
Please remove the committed composer.phar :wink:

ma12-co:
> Please remove the committed composer.phar wink

Uooups I bulk staged and it got me xD

PR #913, opened 2020-03-14
Add breadcrumb component

### LINKED ISSUE #800: Breadcrumb component
Would it be of interest to have a component for the breadcrumb bar? I currently use a breadcrumb bar in one of my apps and wonder whether other apps might also need it.

raimund-schluessler:
This PR adds a breadcrumb component as it is used in the files app, closes #800.
The component resizes with the viewport and hides folders in a dropdown if the width is to small. The first (root folder) and the last folder are always shown though. The component also emits an event when something is dropped on a folder. The dropdown with the hidden folders opens on hover, so one can drop on the hidden folders as well.

Some things don't work yet on netlify, but do work locally for me:
- [x] The initial resize on page load does not work. Resizing the browser window leads to a correct resizing
- [x] ~The `background-color` on hover is not working. Something seems to be off with this line https://github.com/nextcloud/nextcloud-vue/pull/913/files#diff-f1fe4599663441afb9babaefe49e0c7aR349~  
var(--color-primary-light) doesn't seem to be available on netlify, but since it is used for other components, I guess it is fine?
- [x] The breadcrumb arrow is not found

@skjnldsv Do you have an idea regarding the last two problems?

I didn't test the component in the Files app. Looking at https://github.com/nextcloud/server/blob/e4948f5d86a7df2a683e288e55f6f626d90d7cf0/apps/files/js/breadcrumb.js I guess there is more stuff to add on the component, but I would like to leave this to someone familiar with the Files app.

### DIFF
--- a/src/components/Breadcrumb/Breadcrumb.vue
+++ b/src/components/Breadcrumb/Breadcrumb.vue
@@ -1,0 +1,203 @@
+<!--
+ - @copyright Copyright (c) 2020 Raimund Schlüßler <raimund.schluessler@mailbox.org>
+ -
+ - @author Raimund Schlüßler <raimund.schluessler@mailbox.org>
+ -
+ - @license GNU AGPL version 3 or any later version
+ -
+ - This program is free software: you can redistribute it and/or modify
+ - it under the terms of the GNU Affero General Public License as
+ - published by the Free Software Foundation, either version 3 of the
+ - License, or (at your option) any later version.
+ -
+ - This program is distributed in the hope that it will be useful,
+ - but WITHOUT ANY WARRANTY; without even the implied warranty of
+ - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ - GNU Affero General Public License for more details.
+ -
+ - You should have received a copy of the GNU Affero General Public License
+ - along with this program. If not, see <http://www.gnu.org/licenses/>.
+ -
+ -->
+
+<docs>
+
+### General description
+
+This component is meant to be used inside a Breadcrumbs component.
+
+</docs>
+
+<template>
+	<div ref="crumb"
+		class="crumb"
+		:class="{'crumb--with-action': $slots.default, 'crumb--hovered': hovering}"
+		draggable="false"
+		@dragstart.prevent="() => {/** Prevent the breadcrumb from being draggable. */}"
+		@drop.prevent="dropped"
+		@dragover.prevent="() => {}"
+		@dragenter="dragEnter"
+		@dragleave="dragLeave">
+		<element :is="tag" :to="to" :href="href">
+			<span v-if="icon" :class="icon" class="icon" />
+			<span v-else>{{ title }}</span>
+		</element>
+		<Actions :force-menu="forceMenu">
+			<!-- @slot All action elements passed into the default slot will be used -->
+			<slot />
+		</Actions>
+	</div>
+</template>
+
+<script>
+import Actions from '../Actions'
+
+export default {
+	name: 'Breadcrumb',
+	components: {
+		Actions,
+	},
+	props: {
+		/**
+		 * The displayed title of the breadcrumb.
+		 */
+		title: {
+			type: String,
+			required: true,
+		},
+		/**
+		 * The router-link to prop [https://router.vuejs.org/api/#to](https://router.vuejs.org/api/#to)
+		 * If set, the breadcrumbs will be rendered by router-link.
+		 */
+		to: {
+			type: String,
+			default: undefined,
+		},
+		/**
+		 * Set this prop if your app doesn't use vue-router, breadcrumbs will show as normal links.
+		 */
+		href: {
+			type: String,
+			default: undefined,
+		},
+		/**
+		 * Set a css icon-class to show an icon instead of the title text.
+		 */
+		icon: {
+			type: String,
+			default: '',
+		},
+		/**
+		 * Disable dropping on this breadcrumb.
+		 */
+		disableDrop: {
+			type: Boolean,
+			default: false,
+		},
+		/**
+		 * Force the actions to display in a three dot menu
+		 */
+		forceMenu: {
+			type: Boolean,
+			default: false,
+		},
+	},
+	data() {
+		return {
+			/**
+			 * Variable to track if we hover over the breadcrumb
+			 */
+			hovering: false,
+		}
+	},
+	computed: {
+		/**
+		 * Determines which element tag to use
+		 *
+		 * @returns {String} the tag
+		 */
+		tag() {
+			return this.to ? 'router-link' : 'a'
+		},
+	},
+	methods: {
+		/**
+		 * Function to handle a drop on the breadcrumb.
+		 * $emit the event and the path, remove the hovering state.
+		 *
+		 * @param {Object} e The drop event
+		 * @returns {boolean}
+		 */
+		dropped(e) {
+			/**
+			 * Don't do anything if dropping is disabled.
+			 */
+			if (this.disableDrop) {
+				return false
+			}
+			/**
+			 * Event emitted when something is dropped on the breadcrumb.
+			 * @type {null}
+			 */
+			this.$emit('dropped', e, this.to || this.href)
+			this.$parent.$emit('dropped', e, this.to || this.href)
+			this.hovering = false
+			return false
+		},
+		/**
+		 * Add the hovering state on drag enter
+		 *
+		 * @param {Object} e The drag enter event
+		 */
+		dragEnter(e) {
+			/**
+			 * Don't do anything if dropping is disabled.
+			 */
+			if (this.disableDrop) {
+				return
+			}
+			this.hovering = true
+		},
+		/**
+		 * Remove the hovering state on drag leave
+		 *
+		 * @param {Object} e The drag leave event
+		 */
+		dragLeave(e) {
+			/**
+			 * Don't do anything if dropping is disabled.
+			 */
+			if (this.disableDrop) {
+				return
+			}
+			// Don't do anything if we
+			// - leave towards a child element.
+			// - or are still within the crumb
+			if (e.target.contains(e.relatedTarget)
+				|| this.$refs.crumb.contains(e.relatedTarget)) {
+				return
+			}
+			this.hovering = false
+		},
+	},
+}
+</script>
+
+<style lang="scss" scoped>
+.breadcrumb div.crumb {
+	&--with-action a {
+		padding-right: 2px;
+	}
+
+	a {
+		align-items: center;
+		display: inline-flex;
+
+		> span {
+			overflow: hidden;
+			text-overflow: ellipsis;
+			white-space: nowrap;
+		}
+	}
+}
+</style>
--- a/src/components/Breadcrumbs/Breadcrumbs.vue
+++ b/src/components/Breadcrumbs/Breadcrumbs.vue
@@ -33,12 +33,19 @@ is dropped on a creadcrumb.
 ```vue
 <template>
 	<div>
-		<div class="container container__wide">
-			<Breadcrumbs :breadcrumbs="breadcrumbs" @dropped="dropped" />
-		</div>
-		<br />
-		<div class="container container__narrow">
-			<Breadcrumbs :breadcrumbs="breadcrumbs" @dropped="dropped" />
+		<div class="container">
+			<Breadcrumbs @dropped="dropped">
+				<Breadcrumb title="Home" href="/" @dropped="droppedOnCrumb" />
+				<Breadcrumb title="Folder 1" href="/Folder 1" />
+				<Breadcrumb title="Folder 2" href="/Folder 1/Folder 2" :disable-drop="true" />
+				<Breadcrumb title="Folder 3" href="/Folder 1/Folder 2/Folder 3" />
+				<Breadcrumb title="Folder 4" href="/Folder 1/Folder 2/Folder 3/Folder 4" />
+				<Breadcrumb title="Folder 5" href="/Folder 1/Folder 2/Folder 3/Folder 4/Folder 5" :disable-drop="true">
+					<ActionButton icon="icon-share" @click="alert('Share')">
+						Share
+					</ActionButton>
+				</Breadcrumb>
+			</Breadcrumbs>
 		</div>
 		<br />
 		<div class="dragme" draggable="true">
@@ -49,21 +56,12 @@ is dropped on a creadcrumb.
 
 <script>
 export default {
-	computed: {
-		breadcrumbs() {
-			const path = 'Folder 1/Folder 2/Folder 3/Folder 4'
-			const crumbs = (path === '') ? [] : path.split('/')
-			return [{ name: 'Home', path: '/' }].concat(crumbs.map((crumb, i) => {
-				return {
-					name: crumb,
-					path: '/' + crumbs.slice(0, i + 1).join('/'),
-				}
-			}))
-		},
-	},
 	methods: {
-		dropped(path) {
-			alert('Dropped something on ' + path)
+		dropped(e, path) {
+			alert('Global drop on ' + path)
+		},
+		droppedOnCrumb(e, path) {
+			alert('Drop on crumb ' + path)
 		},
 	}
 }
@@ -71,14 +69,7 @@ export default {
 <style>
 .container {
 	display: inline-flex;
-}
-
-.container__wide {
-	width: 800px;
-}
-
-.container__narrow {
-	width: 250px;
+	width: 100%;
 }
 
 .dragme {
@@ -91,126 +82,55 @@ export default {
 ```
 </docs>
 
-<template>
-	<div ref="container" class="breadcrumb">
-		<div v-for="(crumb, index) in crumbs1"
-			:key="`f1${index}`"
-			:ref="`crumb_${index}`"
-			class="crumb svg folder"
-			:class="{'hidden': isHidden(index)}"
-			draggable="false"
-			@dragstart="dragstart"
-			@drop="dropped(index, $event)"
-			@dragover="dragOver($event)"
-			@dragenter="($event) => dragEnter(index, $event)"
-			@dragleave="dragLeave">
-			<a :href="'#' + crumb.path">
-				<span v-if="!index" :class="rootIcon" class="icon" />
-				<span v-else>{{ crumb.name }}</span>
-			</a>
-		</div>
-		<div v-if="hiddenCrumbs.length" class="crumb svg">
-			<Actions class="dropdown"
-				:force-menu="true"
-				:open.sync="actionsOpen"
-				draggable="false"
-				@dragstart.native="dragstart"
-				@dragover.native="dragOver($event)"
-				@drop.native="dropOnActions"
-				@dragenter.native="actionsOpen = true"
-				@dragleave.native="closeActions">
-				<ActionRouter
-					v-for="(crumb, index) in hiddenCrumbs"
-					:key="`dropdown${index}`"
-					:to="crumb.path"
-					icon="icon-folder"
-					class="crumb"
-					draggable="false"
-					@dragstart.native="dragstart"
-					@drop.native="dropped(hiddenIndices[index], $event)"
-					@dragover.native="dragOver($event)"
-					@dragenter.native="($event) => dragEnter(hiddenIndices[index], $event)"
-					@dragleave.native="dragLeave">
-					{{ crumb.name }}
-				</ActionRouter>
-			</Actions>
-		</div>
-		<div v-for="(crumb, index) in crumbs2"
-			:key="`f2${index}`"
-			:ref="`crumb_${index + crumbs1.length}`"
-			class="crumb svg folder"
-			:class="{'hidden': isHidden(index + crumbs1.length)}"
-			draggable="false"
-			@dragstart="dragstart"
-			@drop="dropped(index + crumbs1.length, $event)"
-			@dragover="dragOver($event)"
-			@dragenter="($event) => dragEnter(index + crumbs1.length, $event)"
-			@dragleave="dragLeave">
-			<a :href="'#' + crumb.path">
-				<span>{{ crumb.name }}</span>
-			</a>
-		</div>
-	</div>
-</template>
-
 <script>
+import Vue from 'vue'
 import debounce from 'debounce'
-import { Actions } from 'Components/Actions'
-import { ActionRouter } from 'Components/ActionRouter'
+import Actions from '../Actions'
+import ActionRouter from '../ActionRouter'
+import ActionLink from '../ActionLink'
+import ValidateSlot from '../../utils/ValidateSlot'
 
 export default {
+	name: 'Breadcrumbs',
 	components: {
 		Actions,
-		ActionRouter
+		ActionRouter,
+		ActionLink,
 	},
 	props: {
-		breadcrumbs: {
-			type: Array,
-			required: true,
-			default: () => []
-		},
+		/**
+		 * Set a css icon-class for the icon of the root breadcrumb to be used.
+		 */
 		rootIcon: {
 			type: String,
-			required: false,
-			default: 'icon-home'
-		}
+			default: 'icon-home',
+		},
 	},
-	data: function() {
+	data() {
 		return {
+			/**
+			 * The breadcrumbs which should be shown in a dropdown Actions menu.
+			 */
+			hiddenCrumbs: [],
+			/**
+			 * Array to track the hidden breadcrumbs by their index.
+			 * Comparing two crumbs somehow does not work, so we use the indices.
+			 */
 			hiddenIndices: [],
-			actionsOpen: false
 		}
 	},
-	computed: {
-		crumbs1() {
-			if (this.hiddenIndices.length) {
-				return this.breadcrumbs.slice(0, Math.round(this.breadcrumbs.length / 2))
-			}
-			return this.breadcrumbs
-		},
-
-		crumbs2() {
-			if (this.hiddenIndices.length) {
-				return this.breadcrumbs.slice(Math.round(this.breadcrumbs.length / 2))
-			}
-			return []
-		},
-
-		hiddenCrumbs() {
-			const crumbs = []
-			for (let jj = 0; jj < this.hiddenIndices.length; jj++) {
-				crumbs.push(this.breadcrumbs[this.hiddenIndices[jj]])
-			}
-			return crumbs
-		}
+	beforeMount() {
+		// Filter all invalid items, only Breadcrumb components are allowed
+		ValidateSlot(this.$slots.default, ['Breadcrumb'], this)
 	},
-	watch: {
-		breadcrumbs: function(val) {
-			this.actionsOpen = false
-			this.$nextTick(() => this.handleWindowResize())
-		}
+	beforeUpdate() {
+		// Also check before every update
+		ValidateSlot(this.$slots.default, ['Breadcrumb'], this)
 	},
 	created() {
+		/**
+		 * Add a listener so the component reacts on resize
+		 */
 		window.addEventListener('resize', debounce(() => {
 			this.handleWindowResize()
 		}, 100))
@@ -218,95 +138,180 @@ export default {
 	mounted() {
 		this.handleWindowResize()
 	},
+	updated() {
+		/**
+		 * Check the size on update
+		 */
+		this.$nextTick(() => {
+			this.handleWindowResize()
+		})
+	},
 	beforeDestroy() {
 		window.removeEventListener('resize', this.handleWindowResize)
 	},
 	methods: {
+		/**
+		 * Close the actions menu
+		 *
+		 * @param {Object} e The event
+		 */
 		closeActions(e) {
-			// Get the correct element, the events are bound to the <a>
-			if (e.target.closest) {
-				const target = e.target.closest('.crumb .dropdown')
-				// Don't do anything if we leave towards a child element.
-				if (target.contains(e.relatedTarget)) {
-					return
-				}
-				this.actionsOpen = false
+			// Don't do anything if we leave towards a child element.
+			if (this.$refs.actions.$el.contains(e.relatedTarget)) {
+				return
 			}
+			this.$refs.actions.opened = false
 		},
-
-		isHidden(index) {
-			return this.hiddenIndices.includes(index)
-		},
-
+		/**
+		 * Check the width of the breadcrumb and hide breadcrumbs
+		 * if we overflow otherwise.
+		 */
 		handleWindowResize() {
+			// All breadcrumb components passed into the default slot
+			const breadcrumbs = this.$slots.default || []
+			// If there is no container yet, we cannot determine its size
 			if (this.$refs.container) {
-				const nrCrumbs = this.breadcrumbs.length
+				const nrCrumbs = breadcrumbs.length
 				const hiddenIndices = []
 				const availableWidth = this.$refs.container.offsetWidth
-				const totalWidth = this.getTotalWidth()
+				const totalWidth = this.getTotalWidth(breadcrumbs)
 				let overflow = totalWidth - availableWidth
 				// If we overflow, we have to take the action-item width into account as well.
 				overflow += (overflow > 0) ? 51 : 0
 				let i = 0
+				// We start hiding the breadcrumb in the center
 				const startIndex = Math.floor(nrCrumbs / 2)
+				// Don't hide the first and last breadcrumb
 				while (overflow > 0 && i < nrCrumbs - 2) {
+					// We hide elements alternating to the left and right
 					const currentIndex = startIndex + ((i % 2) ? i + 1 : i) / 2 * Math.pow(-1, i + (nrCrumbs % 2))
-					overflow -= this.getWidth(this.$refs[`crumb_${currentIndex}`][0])
+					// Calculate the remaining overflow width after hiding this breadcrumb
+					overflow -= this.getWidth(breadcrumbs[currentIndex].elm)
 					hiddenIndices.push(currentIndex)
 					i++
 				}
-				this.hiddenIndices = hiddenIndices.sort()
+				// We only update the hidden crumbs if they have changed,
+				// otherwise we will run into an infinite update loop.
+				if (!this.arraysEqual(this.hiddenIndices, hiddenIndices.sort())) {
+					// Get all breadcrumbs based on the hidden indices
+					this.hiddenCrumbs = hiddenIndices.map((index) => { return breadcrumbs[index] })
+					this.hiddenIndices = hiddenIndices
+				}
 			}
 		},
+		/**
+		 * Checks if two arrays are equal.
+		 * Only works for primitive arrays, but that's enough here.
+		 *
+		 * @param {Array} a The first array
+		 * @param {Array} b The second array
+		 * @returns {boolean} Wether the arrays are equal
+		 */
+		arraysEqual(a, b) {
+			if (a.length !== b.length) return false
+			if (a === b) return true
+			if (a === null || b === null) return false
 
-		getTotalWidth() {
-			return this.breadcrumbs.reduce((width, crumb, index) => width + this.getWidth(this.$refs[`crumb_${index}`][0]), 0)
+			for (let i = 0; i < a.length; ++i) {
+				if (a[i] !== b[i]) {
+					return false
+				}
+			}
+			return true
 		},
-
+		/**
+		 * Calculates the total width of all breadcrumbs
+		 *
+		 * @param {Array} breadcrumbs All breadcrumbs
+		 * @returns {Integer} The total width
+		 */
+		getTotalWidth(breadcrumbs) {
+			return breadcrumbs.reduce((width, crumb, index) => width + this.getWidth(crumb.elm), 0)
+		},
+		/**
+		 * Calculates the width of the provided element
+		 *
+		 * @param {Object} el The element
+		 * @returns {Integer} The width
+		 */
 		getWidth(el) {
-			const hide = el.classList.contains('hidden')
-			el.classList.remove('hidden')
+			if (!el.classList) return 0
+			const hide = el.classList.contains('crumb--hidden')
+			el.classList.remove('crumb--hidden')
 			const w = el.offsetWidth
 			if (hide) {
-				el.classList.add('hidden')
+				el.classList.add('crumb--hidden')
 			}
 			return w
 		},
-		cancelEvent(e) {
-			e.stopPropagation()
-			e.preventDefault()
-			return false
-		},
-		dropOnActions(e) {
-			this.actionsOpen = false
-			return this.cancelEvent(e)
-		},
-		dragstart(e) {
-			return this.cancelEvent(e)
-		},
-		dropped(index, e) {
-			this.cancelEvent(e)
-			// If it is the last element in the path,
-			// don't do anything
-			if (index === (this.breadcrumbs.length - 1)) {
-				return
-			}
-			this.$emit('dropped', this.breadcrumbs[index].path)
-			this.actionsOpen = false
-			const crumbs = document.querySelectorAll('.crumb')
-			crumbs.forEach((f) => { f.classList.remove('over') })
-			return false
-		},
-		dragOver(e) {
+		/**
+		 * Prevents the default of a provided event
+		 *
+		 * @param {Object} e The event
+		 * @returns {boolean}
+		 */
+		preventDefault(e) {
 			if (e.preventDefault) {
 				e.preventDefault()
 			}
 			return false
 		},
-		dragEnter(index, e) {
-			// If it is the last element in the path,
-			// don't do anything
-			if (index === (this.breadcrumbs.length - 1)) {
+		/**
+		 * Handles the drag start.
+		 * Prevents a breadcrumb from being draggable.
+		 *
+		 * @param {Object} e The event
+		 * @returns {boolean}
+		 */
+		dragStart(e) {
+			return this.preventDefault(e)
+		},
+		/**
+		 * Handles when something is dropped on the breadcrumb.
+		 *
+		 * @param {Object} e The drop event
+		 * @param {String} path The path of the breadcrumb
+		 * @param {boolean} disabled Whether dropping is disabled for this breadcrumb
+		 * @returns {boolean}
+		 */
+		dropped(e, path, disabled) {
+			/**
+			 * Don't emit if dropping is disabled.
+			 */
+			if (!disabled) {
+				/**
+				 * Event emitted when something is dropped on the breadcrumb.
+				 * @type {null}
+				 */
+				this.$emit('dropped', e, path)
+			}
+			// Close the actions menu after the drop
+			this.$refs.actions.opened = false
+			// Remove all hovering classes
+			const crumbs = document.querySelectorAll('.crumb')
+			crumbs.forEach((f) => { f.classList.remove('crumb--hovered') })
+			return this.preventDefault(e)
+		},
+		/**
+		 * Handles the drag over event
+		 *
+		 * @param {Object} e The drag over event
+		 * @returns {boolean}
+		 */
+		dragOver(e) {
+			return this.preventDefault(e)
+		},
+		/**
+		 * Handles the drag enter event
+		 *
+		 * @param {Object} e The drag over event
+		 * @param {boolean} disabled Whether dropping is disabled for this breadcrumb
+		 */
+		dragEnter(e, disabled) {
+			/**
+			 * Don't do anything if dropping is disabled.
+			 */
+			if (disabled) {
 				return
 			}
 			// Get the correct element, in case we hover a child.
@@ -314,12 +319,24 @@ export default {
 				const target = e.target.closest('.crumb')
 				if (target.classList && target.classList.contains('crumb')) {
 					const crumbs = document.querySelectorAll('.crumb')
-					crumbs.forEach((f) => { f.classList.remove('over') })
-					target.classList.add('over')
+					crumbs.forEach((f) => { f.classList.remove('crumb--hovered') })
+					target.classList.add('crumb--hovered')
 				}
 			}
 		},
-		dragLeave(e) {
+		/**
+		 * Handles the drag leave event
+		 *
+		 * @param {Object} e The drag leave event
+		 * @param {boolean} disabled Whether dropping is disabled for this breadcrumb
+		 */
+		dragLeave(e, disabled) {
+			/**
+			 * Don't do anything if dropping is disabled.
+			 */
+			if (disabled) {
+				return
+			}
 			// Don't do anything if we leave towards a child element.
 			if (e.target.contains(e.relatedTarget)) {
 				return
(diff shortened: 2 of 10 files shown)


skjnldsv:
> * The breadcrumb arrow is not found

You need to add your icon into https://github.com/nextcloud/nextcloud-vue/tree/master/src/assets/iconfont
And rebuild the icon font cache :)

skjnldsv:
Regarding the syntax, I know the objects are rather simple, but we moved away from using objects-data components towards proper nested components.

I'm wondering if having something like:
```vue
<Breadcrumbs @dropped="dropped">
	<Breadcrumb title="Folder 1" />
	<Breadcrumb title="Folder 2" />
	<Breadcrumb title="Folder 3" />
	<Breadcrumb title="Folder 4">
		<ActionButton title="Share" icon="icon-sharing" />
	</Breadcrumb>
</Breadcrumbs>
```
cc @nextcloud/vuejs

georgehrke:
@skjnldsv Shouldn't the `dropped` event be on the individual `Breadcrumb` components instead of the `Breadcrumbs` container?

skjnldsv:
> @skjnldsv Shouldn't the `dropped` event be on the individual `Breadcrumb` components instead of the `Breadcrumbs` container?

yes sorry, bad copy/paste :)

skjnldsv:
> > @skjnldsv Shouldn't the `dropped` event be on the individual `Breadcrumb` components instead of the `Breadcrumbs` container?

Well, actually instead of having it on every Breadcrumb, we could also make them droppable and forward-emit to the parent? What is the most used case scenario? All droppable but one or two? Or only a few droppable? ? :thinking:

raimund-schluessler:
Fine with me. I will adjust the component.

raimund-schluessler:
> Well, actually instead of having it on every Breadcrumb, we could also make them droppable and forward-emit to the parent? What is the most used case scenario? All droppable but one or two? Or only a few droppable? ? :thinking: 

I would say you should be able to drop on every breadcrumb but the last one (since this is the folder which is currently open).

skjnldsv:
> I would say you should be able to drop on every breadcrumb but the last one (since this is the folder which is currently open).

I mean, those are probably gonna be used in a vFor, so it's not going to look that bad:

```vue
<Breadcrumbs>
	<Breadcrumb v-for="crumb in bread" :title="crumb.title" @dropped="dropped(crumb)" />
</Breadcrumbs>
```

raimund-schluessler:
I can't seem to figure out how to put the Actions menu in between the crumbs using the approach disucussed above.

```vue
<Breadcrumbs>
	<Breadcrumb title="Folder 1" @dropped="dropped" />
	<Breadcrumb title="Folder 2" />

	<!-- Actions have to go here -->

	<Breadcrumb title="Folder 3" />
	<Breadcrumb title="Folder 4">
		<ActionButton title="Share" icon="icon-sharing" />
	</Breadcrumb>
</Breadcrumbs>
``` 

How can I do that? @skjnldsv

skjnldsv:
> ```
> <Breadcrumb title="Folder 4">
> 		<ActionButton title="Share" icon="icon-sharing" />
> 	</Breadcrumb>
> ```

this part you mean?
You'll need to ad an <Actions> component inside your `Breadcrumb` component and then a `<slot />`.
Similar to what we do in the sidebar for example: https://github.com/nextcloud/nextcloud-vue/blob/3a17faff30f290628192f731809f18faefd6d177/src/components/AppSidebar/AppSidebar.vue#L143-L415

raimund-schluessler:
No, I mean the actions menu with the hidden crumbs that will show up in between the crumbs when the width is to narrow.

skjnldsv:
Ah, I guess it's a fancy calculation on the vue $children and what is possible to be displayed or not :)
Then it's a matter of display a v-for of ActionLinks/ActionButton/ActionRouter?

raimund-schluessler:
The calculation is fine. It is the same as it is now. What I can’t figure out is how I show the actions *inbetween* the provided slots.

skjnldsv:
You inject it from the Breadcrumbs component?
Not sure I get the issue, wanna jump into a call?

raimund-schluessler:
> Not sure I get the issue, wanna jump into a call?

Sorry, I can't call right now.

What I try to do is show the Actions let's say after the second (number dynamically calculated based on number of breadcrumbs) breadcrumb.

But using this the Actions are shown only after all the crumbs.
```vue
<template>
	<div class="breadcrumb">
		<slot /> <!-- All crumbs in here -->
		<Actions>
	</div>
</template>
```

I would kind of need two slots and dynamically asssign a crumb to the first or second slot.

```vue
<template>
	<div class="breadcrumb">
		<slot /> <!-- Only first half here -->
		<Actions>
		<slot /> <!-- Second half here -->
	</div>
</template>
```

I hope that makes it clear. If not we can have a call later maybe.

skjnldsv:
Ah right!

Two options in my mind:
1. we cheat with css order to slip in the Actions (some might say it can be messy; conflict with the tab order and therefore the accessibility).
2. we do a *fake* slot that is hidden and do a v-for amongst all $children/$slot array that we split in two (more complicated part)

raimund-schluessler:
>     2. we do a _fake_ slot that is hidden and do a v-for amongst all $children/$slot array that we split in two (more complicated part)

Ok, I will give this a try. The first approach doesn't sound right somehow. Thanks!

skjnldsv:
> Ok, I will give this a try. The first approach doesn't sound right somehow. Thanks!

We do similar things on the Actions componeent if you wanna look, but I'm afraid it's also a bit messy. I would love to do it a bit more cleanly. 
I think the solution in those cases is probably to not use a single file component and template here.

But to use a proper render function, so that we only do this mathematically.
https://stackoverflow.com/a/48897774/3885878

raimund-schluessler:
After looking a bit into the details, I think this will be much more complicated than it is at the moment. I will try to make it work and provide a PR.

skjnldsv:
I mean, to be fair the link I gave is probably the easiest thing to do.
Rough idea:

```js
Vue.component('BreadCrumbs',{
  render: function(createElement){
        const crumbs = []
		const slots = this.$slots.default
		// Let's exclude the ones we don't want here
		listStart = 
        slots.slice(/* whatever the only we want */).forEach((element,index) => {
			// we got a real element
            if (element.tag) {
                crumbs.push(createElement('BreadCrumb',{},[element]))
            }
			if (/* thi sis time to inject the actions */) {
				crumbs.push(/*Actions... etc*/)
			}
			
        });

        return createElement('div',{ class: 'breadcrumb' }, crumbs)
    }

})

```

raimund-schluessler:
I made some progress with the render function approach. Breadcrumbs are shown/hidden if necessary, the actions show up properly.

Left to do:
- [ ] Drag and drop is not working at all yet
- [ ] Breadcrumb separator is not found
- [ ] Adjust documentation

skjnldsv:
> * Breadcrumb separator is not found

https://github.com/nextcloud/nextcloud-vue/tree/master/src/assets/iconfont

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -0,0 +1,362 @@
...
+		const breadcrumbs = (this.$slots.default || []).filter(node => !!node && !!node.componentOptions)
+		const crumbs = []
+		// The first half of the breadcrumbs
+		const crumbs1 = this.hiddenIndices.length
+			? breadcrumbs.slice(0, Math.round(breadcrumbs.length / 2))
+			: breadcrumbs
+		crumbs1.forEach((crumb, i) => {
+			if (crumb.elm && crumb.elm.classList) {
+				if (this.isHidden(i)) {
+					crumb.elm.classList.add('hidden')
```
skjnldsv: Make sure we have the hidden styling in the library so we do not rely on the server's css :)

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -0,0 +1,362 @@
...
+				if (target.contains(e.relatedTarget)) {
+					return
+				}
+				if (target.classList && target.classList.contains('crumb')) {
+					target.classList.remove('over')
+				}
+			}
+		},
+	},
+	render: function(h) {
```
skjnldsv: Maybe not use `h` here but a nice name so it's easier to follow. Vuejs suggest usng `createElement` iirc

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -0,0 +1,362 @@
...
+	.crumb {
+		&.over {
+			background-color: var(--color-primary-light);
+		}
+
+		&.action-item {
+			ul {
+				overflow-y: auto;
+				-webkit-overflow-scrolling: touch;
+				min-height: calc(44px * 1.5);
```
skjnldsv:
```suggestion
				min-height: calc($clickable-area * 1.5);
```

https://github.com/nextcloud/nextcloud-vue/blob/3a17faff30f290628192f731809f18faefd6d177/src/assets/variables.scss#L26

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -0,0 +1,362 @@
...
+				max-height: calc(100vh - 50px * 2);
```
skjnldsv: What is this for?

src/components/index.js
```diff
@@ -44,6 +44,7 @@ import AppNavigationSpacer from './AppNavigationSpacer'
 import AppSidebar from './AppSidebar'
 import AppSidebarTab from './AppSidebarTab'
 import Avatar from './Avatar'
+import Breadcrumbs from './Breadcrumbs'
```
skjnldsv: Missing Breadcrumb too?

src/components/Breadcrumb/Breadcrumb.vue
```diff
@@ -0,0 +1,94 @@
...
+<docs>
+
+### General description
+
+This component is meant to be used inside a Breadcrumbs component.
+
+</docs>
+
+<template>
+	<div class="crumb svg folder"
```
skjnldsv: svg should not be needed here, this doesn't do anything

skjnldsv: Same for folder, what does it means?

src/components/Breadcrumb/Breadcrumb.vue
```diff
@@ -0,0 +1,94 @@
...
+		draggable="false">
+		<a :href="'#' + path">
```
skjnldsv: Make sure we're compatible with vue-router too :)

src/components/Breadcrumb/Breadcrumb.vue
```diff
@@ -0,0 +1,94 @@
...
+			default: '',
+		},
+	},
+}
+</script>
+
+<style lang="scss" scoped>
+
+.breadcrumb {
+	width: calc(100% - 88px);
```
skjnldsv: 88 px?

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -0,0 +1,362 @@
...
+	updated() {
+		this.$nextTick(() => {
+			this.handleWindowResize()
+		})
+	},
+	beforeDestroy() {
+		window.removeEventListener('resize', this.handleWindowResize)
+	},
+	methods: {
+		closeActions(e) {
```
skjnldsv: Unused now?

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -0,0 +1,362 @@
...
+	props: {
+		rootIcon: {
+			type: String,
+			required: false,
+			default: 'icon-home',
+		},
+	},
+	data: function() {
+		return {
+			hiddenIndices: [],
```
skjnldsv: Please add comments on what are those vars used for :)

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -0,0 +1,362 @@
...
+			hiddenCrumbs: [],
+			actionsOpen: false,
```
skjnldsv: Is this really needed?
You can also use the Actions as $ref and trigger open/close from there instead of keeping a duplicate v-model prop here :)

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -0,0 +1,362 @@
...
+		window.addEventListener('resize', debounce(() => {
+			this.handleWindowResize()
+		}, 100))
+	},
+	mounted() {
+		this.handleWindowResize()
+	},
+	updated() {
+		this.$nextTick(() => {
+			this.handleWindowResize()
```
skjnldsv: I guess the calculations can be done on the render directly, no?
So that it's instantly resized?

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: I added this so the menu would not overflow the screen and start to scroll. But this should rather go into the individual apps using the component since it depends on where the breadcrumbs are placed (or the Actions component should take care of it).

src/components/Breadcrumb/Breadcrumb.vue
raimund-schluessler: `folder` was a leftover from a previous version where I got the elements by searching for css selectors. Should be possible to remove it.

src/components/Breadcrumb/Breadcrumb.vue
raimund-schluessler: I will have a look.

src/components/Breadcrumb/Breadcrumb.vue
raimund-schluessler: Should be removed. This is a leftover from how I use the Breadcrumbs in my app.

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: Not sure yet. It is used for closing the actions after dragging over them, which is not workig yet and needs to be implemented.

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: I will check.

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: I am not sure if this works. The calculation of the width might be wrong when the vNode is not in the DOM yet. I will check.

src/components/Breadcrumbs/Breadcrumbs.vue
skjnldsv: Yes, the ACtions shoukd be taking care of this :)
especially after #832

src/components/Breadcrumb/Breadcrumb.vue
skjnldsv: Nice, let's clean things up :rocket:

src/components/Breadcrumbs/Breadcrumbs.vue
skjnldsv: Ah right! Make sense :)

raimund-schluessler:
I added the breadcrumb icon to the icon-font, but it is misaligned and I can't figure out how to place it correctly. The `::before` element occurs, well, before the content, although the icon should be shown at the right side after the text. Also, the vertical alignment is off. @skjnldsv Do I need to wrap it in a div or span and place this one?

skjnldsv:
> I added the breadcrumb icon to the icon-font, but it is misaligned and I can't figure out how to place it correctly.

what's great about this is a simple display flex, align/justify center and that's all perfect ;)

raimund-schluessler:
> 
> 
> > I added the breadcrumb icon to the icon-font, but it is misaligned and I can't figure out how to place it correctly.
> 
> what's great about this is a simple display flex, align/justify center and that's all perfect ;)

But this won't show the arrow to the right of the text instead of the left where it is currently, wouldn't it?

skjnldsv:
> But this won't show the arrow to the right of the text instead of the left where it is currently, wouldn't it?

you can play with the order if you want then.
It doesn't really matter as this is a unfocusable  element

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -274,6 +274,8 @@ export default {
 		}
 	},
 	render: function(createElement) {
+		const first = this.$slots.default[0]
+		first.componentOptions.propsData.icon = this.rootIcon
```
skjnldsv: You might have to use Vue.set to make sure Vue is aware of any key creation/change

raimund-schluessler: Works without Vue.set already, but let's add it.

raimund-schluessler:
We will have to emit the dropped event from the Breadcrumbs component, not the single crumbs. This is because in case we hide breadcrumbs, the hidden onces will be shown in an Actions menu which is injected by the Breadcrumbs component. And since we want to drop on the hidden breadcrumbs as well, the event will come from the parent. Breadcrumbs component.

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -274,8 +275,13 @@ export default {
 		}
 	},
 	render: function(createElement) {
+		// Add the root icon to the first breadcrumb
 		const first = this.$slots.default[0]
-		first.componentOptions.propsData.icon = this.rootIcon
+		if (first) {
```
skjnldsv: Does it make sense to render at all if there is no first?
Also be careful, the first item might not be the BreadCrumb (can be a space, some text, a comment...)

src/components/Breadcrumb/Breadcrumb.vue
raimund-schluessler: The `<a>` are rendered by `router-link` now as it is in the `ActionRouter`. So I think this should be fine.

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: I added the class.

raimund-schluessler:
- [ ] `router-link` doesn't render correctly on netlify, works locally.

raimund-schluessler:
We have to listen to the `$emit` of the `Breadcrumb` components in the `Breadcrumbs` component, but I don't know how to do it yet. Hints are welcome 😉

skjnldsv:
> Found a solution by emitting directly in parent scope from child component:
> [#913 (files)](https://github.com/nextcloud/nextcloud-vue/pull/913/files#diff-0a286c9725b45c614ea798be690a9187R82)
> 
> Don't know if this is a good solution, though.

I don't understand what was the issue? Shuldn't we listen to the `Breadcrumb`component for the `dropped` event? Do you want to have a global event on the `Breadcrumbs` one?

raimund-schluessler:
We *have* to have a global one on the breadcrumbs component, because it emits when we drop on a hidden crumb.

skjnldsv:
> because it emits when we drop on a hidden crumb

Can we drop on an Action ?
Could be nice to support?

Anyway, I don't fully grasp the different uses cases, so to answer your question: https://github.com/vuejs/vue/issues/4332#issuecomment-263444492
Parent seems to be fine

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -0,0 +1,385 @@
...
+	},
+	beforeDestroy() {
+		window.removeEventListener('resize', this.handleWindowResize)
+	},
+	methods: {
+		closeActions(e) {
+			// Get the correct element, the events are bound to the <a>
+			if (e.target.closest) {
+				const target = e.target.closest('.crumb .dropdown')
+				// Don't do anything if we leave towards a child element.
```
skjnldsv: I don't understand what is this part about ?

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -0,0 +1,385 @@
...
+	},
+	data: function() {
+		return {
+			hiddenIndices: [],
+			hiddenCrumbs: []
+		}
+	},
+	created() {
+		window.addEventListener('resize', debounce(() => {
+			this.handleWindowResize()
```
skjnldsv: Isn't it a duplicate with mounted & updated?
Have you checked how many times it's triggered on page init?

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -0,0 +1,385 @@
...
+		<div class="dragme" draggable="true">
+			<span>Drag me onto the breadcrumbs.</span>
+		</div>
+	</div>
+</template>
+
+<script>
+export default {
+	computed: {
+		breadcrumbs() {
```
skjnldsv: Do we want to show example that complicated?
I think I would be ok to not usea v-for for the docs, dev should understand how it works?

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -0,0 +1,385 @@
...
+		return createElement('div', { class: 'breadcrumb', ref: 'container' }, crumbs)
+	}
+}
+</script>
+
+<style lang="scss" scoped>
+@import '../../fonts/scss/iconfont-vue';
+
+.breadcrumb {
+	width: calc(100% - 88px);
```
skjnldsv: Still don't get this 88px? What does it represent? :)

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -0,0 +1,385 @@
...
+					},
+					ref: 'actions',
+					nativeOn: {
+						dragstart: this.dragstart,
+						dragenter: this.openActions,
+						dragleave: this.closeActions
+					}
+				}, this.hiddenCrumbs.map(crumb => {
+					const path = crumb.componentOptions.propsData.path
+					return createElement(ActionRouter, {
```
skjnldsv: I can also be `ActionLink` if there is no `:to`, right? :thinking:

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: I am not done cleaning up. I will remove it. It is because I had the snapper on the left and an other action on the right, so I needed it to be 88 px smaller than the page width.

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: Shouldn't there always be a `:to`?

skjnldsv: If it's a simple link, no :)
Not every apps uses vue-router ;)

src/components/Breadcrumbs/Breadcrumbs.vue
skjnldsv: Sorry, I'll leave this pr alone then xD
Ping be again when you need!

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: Fine with me.

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: Hm, how should we decide if it is an ActionLink or ActionRouter then? They both need a path/href.

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: No, it's fine. Your comments are very helpful. I just can't completely keep up with them 😅

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: This prevents that the dropdown is immediately closed when you hover from the Actions three-dot icon towards the single Actions. The drag-leave event is bound to the `<a>`, when you leave it, `closeActions` is called, we then need to check if we are still in the Actions menu of have left it completely. I will try to add a better comment for this.

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: Maybe have to props: `href` and `to`. `to` makes it a ActionRouter, `href` will lead to an `ActionLink`. This should fix the netlifiy issue not rendering `router-link` as well.

raimund-schluessler: I added two props now, `to` and `href`. When `to` is set, it uses `router-link` and `ActionRouter`, if not, it uses `a` and `ActionLink`.

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: I checked and we have to do it separately. The width of the parent is not available until we render. At least I didn't find a way, if someone has an idea, let me know.

raimund-schluessler:
@skjnldsv I think the code works well now. I will go through it once more and add more comments, but in general it should be ok already.

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: I don't think so, but I will check. In the created hook I only add the listener.

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -0,0 +1,380 @@
...
+						crumb.elm.classList.add('hidden')
+					} else {
+						crumb.elm.classList.remove('hidden')
+					}
+				}
+				crumbs.push(crumb)
+			})
+		},
+	},
+	render: function(createElement) {
```
skjnldsv: I would love a little bit more details on this function. If someone have to maintain it afterwards. Extra comments will not hurt the bundle size :wink:

raimund-schluessler:
> One thing missing though is the Action inside the Breadcrumb, like we have on files with the sharing icon, mind to add a slot for it? :)

Ah, you are right, I forgot about that. Will add it.

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: Will do that. I currently also check how often it triggers.

src/components/Breadcrumb/Breadcrumb.vue
```diff
@@ -0,0 +1,148 @@
...
+		},
+	},
+	methods: {
+		cancelEvent(e) {
+			e.stopPropagation()
+			e.preventDefault()
+			return false
+		},
+		dragstart(e) {
+			return this.cancelEvent(e)
```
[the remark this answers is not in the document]
raimund-schluessler: `e.preventDefault()` is necessary so it doesn't navigate to the crumb when you drop and to prevent dragging crumbs.

`e.stopPropagation()` doesnt' seem to be necessary. I removed it.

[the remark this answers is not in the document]
skjnldsv: Ah, I thought I posted it too, but you can avoid using prevent/stop with the events modifier directly
https://vuejs.org/v2/guide/events.html#Event-Modifiers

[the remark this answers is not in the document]
raimund-schluessler: I added the `.prevent` modifier to the `Breadcrumb` component template. For the `Breadcrumbs` render function modifiers are not available and one should handle it in the handler instead https://vuejs.org/v2/guide/render-function.html#The-Data-Object-In-Depth

src/components/Breadcrumb/Breadcrumb.vue
```diff
@@ -0,0 +1,148 @@
...
+		},
+		dropped(e) {
+			this.cancelEvent(e)
+			this.$parent.$emit('dropped', this.to || this.href)
```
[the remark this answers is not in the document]
raimund-schluessler: > 
> 
> Please also emit for this component maybe?
> So we can chose between watching a single one versus making all of the crumbs droppable?

Fine with me. But one has to be careful then. If you have handlers on both, you will also get two events fired.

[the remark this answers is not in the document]
raimund-schluessler: > Also, could be nice if we could disable the droppable function on a per-crumb basis?
> What do you think? Too much?

Makes sense. Dropping on the last crumb might be unwanted, since the item is already in this folder. Also helpful if moving something to this folder is not allowed (read-only?). I will add that.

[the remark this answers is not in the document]
raimund-schluessler: Dropping can be disabled now by `:disableDrop`.

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -0,0 +1,380 @@
...
+This component renders a breadcrumb bar. It adjusts to the available width
+by hiding breadcrumbs in a dropdown menu and emits an event when something
+is dropped on a creadcrumb.
+
+### Usage
+
+```vue
+<template>
+	<div>
+		<div class="container container__wide">
```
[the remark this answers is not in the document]
raimund-schluessler: I removed the second example.

skjnldsv:
@raimund-schluessler btw you can make sure to only allow specific vnode with this: https://github.com/nextcloud/nextcloud-vue/blob/3beb864232dd1473efd786d53f74c03128ba5443/src/utils/ValidateSlot.js#L31

raimund-schluessler:
> @raimund-schluessler btw you can make sure to only allow specific vnode with this:
> 
> https://github.com/nextcloud/nextcloud-vue/blob/3beb864232dd1473efd786d53f74c03128ba5443/src/utils/ValidateSlot.js#L31

Thanks for the hint. Seems to work quite well. But it has issues with text nodes like this one:
```
{
asyncFactory: undefined
asyncMeta: undefined
​children: undefined
​componentInstance: undefined
​componentOptions: undefined
​context: undefined
​data: undefined
​elm: #text " "
fnContext: undefined
​fnOptions: undefined
​fnScopeId: undefined
​isAsyncPlaceholder: false
​isCloned: false
​isComment: false
​isOnce: false
​isRootInsert: true
​isStatic: false
​key: undefined
​ns: undefined
​parent: undefined
​raw: false
​tag: undefined
​text: " "
​<prototype>: Object { … }
}
```

Should I patch ValidateSlot to filter these nodes as well?

skjnldsv:
> Should I patch ValidateSlot to filter these nodes as well?

What kind of nodes?
Never had any issues with the other components so far, but feel free to fix it sure :)

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: I removed the `mounted` hook, on initial loading only the `updated` hook fires now, which seems to be enough.

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: As stated above, I removed the `mounted` hook.

src/components/Breadcrumb/Breadcrumb.vue
```diff
@@ -100,6 +103,13 @@ export default {
 			return this.to ? 'router-link' : 'a'
 		},
 	},
+	beforeMount() {
+		// filter invalid items
+		ValidateSlot(this.$slots.default, ['Actions'], this)
```
skjnldsv: Ah, you should not bother here, just already put the Actions inside, wrapping the slot.
That way the Actions component will take care of it.
```vue
<BreadCrumb>
	<ActionButton />
</BreadCrumb>
```

Like we do in the Sidebar for example :)
https://github.com/nextcloud/nextcloud-vue/blob/ed35e7754341b1e0497141fa66743994b3b8b735/src/components/AppSidebar/AppSidebar.vue#L150

raimund-schluessler:
> > Should I patch ValidateSlot to filter these nodes as well?
> 
> What kind of nodes?
> Never had any issues with the other components so far, but feel free to fix it sure :)

Not exactly sure. Seems to be a text node with only white space, which occured when I added an 'Actions' node as a negative test. Couldn't really track it down where it comes from, but it looks like this:
Unbenannt

for this (obviously invalid) template:
```
<Breadcrumbs root-icon="icon-bw icon-items" @dropped="moveEntities">
	<Breadcrumb v-for="(crumb, index) in breadcrumbs"
		:key="crumb.path"
		:title="crumb.title"
		:to="crumb.path"
		:disable-drop="index === (breadcrumbs.length - 1)" />
	<Actions>
		<ActionButton icon="icon-sharing">
			Share
		</ActionButton>
	</Actions>
</Breadcrumbs>
```

raimund-schluessler:
@skjnldsv These text nodes currently appear on netlify as well and break the Breadcrumbs component. I will adjust ValidateSlot.

raimund-schluessler:
@skjnldsv ValidateSlots has an issue with `forEach` and `slice`. This combination will skip the next element after you sliced, thus not checking every slot. See here https://stackoverflow.com/questions/21811630/splicing-a-javascript-array-from-within-the-callback-passed-to-foreach

I will try to fix that.

raimund-schluessler:
I send a PR to fix the ValidateSlot function in #952. It is required for the breadcrumbs to work properly. I will rebase this PR here once the fix is merged.

raimund-schluessler:
Requires #954

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -0,0 +1,398 @@
...
+		background-image: none;
+		display: inline-flex;
+		height: $clickable-area;
+		padding: 0;
+
+		&::before {
+			display: flex;
+			align-items: center;
+			order: 1;
+			color: #D7D7D7;
```
skjnldsv: Not compliant to the dark theme :)

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -0,0 +1,398 @@
...
+	div.crumb {
+		@include iconfont('breadcrumb');
+		background-image: none;
+		display: inline-flex;
+		height: $clickable-area;
+		padding: 0;
+
+		&::before {
+			display: flex;
+			align-items: center;
```
skjnldsv: you should not need the display flex. It will have no effect here, it only contains text

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -0,0 +1,398 @@
...
+		if (this.hiddenCrumbs.length) {
+			crumbs.push(createElement('div', { class: 'crumb' },
+				[createElement('Actions', {
+					class: 'dropdown',
+					props: {
+						forceMenu: true,
+					},
+					ref: 'actions',
+					nativeOn: {
+						dragstart: this.dragStart,
```
skjnldsv: Cannot use event modifiers here?

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -0,0 +1,398 @@
...
+		ActionRouter,
+		ActionLink,
+	},
+	props: {
+		/**
+		 * Set a css icon-class for the icon of the root breadcrumb to be used.
+		 */
+		rootIcon: {
+			type: String,
+			required: false,
```
skjnldsv:
```suggestion
```

required false is the default

src/components/Breadcrumb/Breadcrumb.vue
```diff
@@ -0,0 +1,157 @@
...
+	},
+	methods: {
+		dropped(e) {
+			/**
+			 * Don't do anything if dropping is disabled.
+			 */
+			if (this.disableDrop) {
+				return false
+			}
+			this.$emit('dropped', this.to || this.href)
```
skjnldsv: I would probably just re-emit the original event ?
People might want to get the data differently than to/href?

skjnldsv:
Shall we also add this into its own header in the documentation? https://github.com/nextcloud/nextcloud-vue/blob/181bd616ec2983ca1891ba5c50276cb1d8f56850/styleguide.config.js#L53

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: Hm, I just took the color from the breadcrumb.svg which is the same for light and dark. Which color should we use? `var(--color-main-text)`?

raimund-schluessler: I don't see any color variable which resembles the color we have at the moment.

raimund-schluessler: If at all `var(--color-text-maxcontrast)` comes close, but it doesn't seem to make much sense.

Screenshot_2020-03-28 Inventar - Nextcloud
Screenshot_2020-03-28 Inventar - Nextcloud(1)

raimund-schluessler: `var(--color-border-dark)` might work as well, but I think it is a bit to light in light theme and to dark in dark theme:

Screenshot_2020-03-28 Inventar - Nextcloud(2)
Screenshot_2020-03-28 Inventar - Nextcloud

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: No, this is not possible. From the docs https://vuejs.org/v2/guide/render-function.html#The-Data-Object-In-Depth:
```
  // Event handlers are nested under `on`, though
  // modifiers such as in `v-on:keyup.enter` are not
  // supported. You'll have to manually check the
  // keyCode in the handler instead.
  on: {
    click: this.clickHandler
  },
  // For components only. Allows you to listen to
  // native events, rather than events emitted from
  // the component using `vm.$emit`.
  nativeOn: {
    click: this.nativeClickHandler
  },
```

raimund-schluessler:
> Shall we also add this into its own header in the documentation?

Makes sense. I was already wondering how to do that 😄

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: It is needed to align the breadcrumb icon correctly. Without this style it looks like this:
Screenshot_2020-03-28 Inventar - Nextcloud(1)

src/components/Breadcrumb/Breadcrumb.vue
raimund-schluessler: I think it is unnecessarily complicated to extract the path from the original event (if at all possible, I didn't see a nice way by quickly looking at the event). How about we emit both the event and the path like `this.$emit('dropped', e, this.to || this.href)` (and in the `Breadcrumbs` as well).

skjnldsv: You cannot do that, emit only take one parameter :)
Why not add to/href to the event object then?

src/components/Breadcrumbs/Breadcrumbs.vue
skjnldsv: interesting, I would have use line-height then as this is text. I did not thought flex would align it :thinking:

src/components/Breadcrumbs/Breadcrumbs.vue
skjnldsv: Border seems more appropriate as this is after all, a border :stuck_out_tongue_closed_eyes:

src/components/Breadcrumbs/Breadcrumbs.vue
skjnldsv: Looks good then! Let's leave it like that!

raimund-schluessler: `line-height: 44px` also works. I have no preference here. Shall we go for line-height then?

src/components/Breadcrumb/Breadcrumb.vue
raimund-schluessler: Interesting, when I tried it, it worked without problems. But let's just add it to the event then. Shall we do `e.path = this.to || this.href`?

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: > 
> 
> Looks good then! Let's leave it like that!

Alright.

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: Fine with me, it will just change the current look a bit.

src/components/Breadcrumb/Breadcrumb.vue
skjnldsv: Ah sorry, yes you can! I confused myself with vuex methods/actions :)

skjnldsv: So let's leave it like that! https://vuejs.org/v2/api/#vm-emit

src/components/Breadcrumbs/Breadcrumbs.vue
skjnldsv: Let me check with ie

src/components/Breadcrumb/Breadcrumb.vue
skjnldsv: This is still not done :)

raimund-schluessler: > 
> 
> This is still not done :)

What do you mean? I removed the `ValidateSlot` function from the component.

raimund-schluessler:
I also have to add documentation, but I wanted to do this after the code is fixed.

src/components/Breadcrumb/Breadcrumb.vue
```diff
@@ -0,0 +1,157 @@
...
+		@dragstart.prevent="() => {}"
+		@drop.prevent="dropped"
+		@dragover.prevent="() => {}"
+		@dragenter="dragEnter"
+		@dragleave="dragLeave">
+		<element :is="tag" :to="to" :href="href">
+			<span v-if="icon" :class="icon" class="icon" />
+			<span v-else>{{ title }}</span>
+		</element>
+		<slot />
```
skjnldsv: See https://github.com/nextcloud/nextcloud-vue/pull/913#discussion_r399481447

```suggestion
		<Actions>
			<slot />
		</Actions>
```

Of course spacing needs to be fixed a bit when we have one action:
Capture d’écran_2020-03-28_14-08-24

src/components/Breadcrumb/Breadcrumb.vue
raimund-schluessler: Ah, do you mean, I should wrap the slot within an `Actions`, so the user only has to provide the `ActionButton` etc.?

src/components/Breadcrumb/Breadcrumb.vue
skjnldsv: Also like we do in the Sidebar we should maybe forward the :force-menu="forceMenu" prop :)

src/components/Breadcrumbs/Breadcrumbs.vue
skjnldsv: All good in ie!
image

src/components/Breadcrumb/Breadcrumb.vue
raimund-schluessler: Will do that.

src/components/Breadcrumb/Breadcrumb.vue
raimund-schluessler: I fixed that and forwarded the force-menu prop.

src/components/Breadcrumb/Breadcrumb.vue
raimund-schluessler: Done.

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -39,13 +39,16 @@ is dropped on a creadcrumb.
...
-					<Actions>
-						<ActionButton icon="icon-share" @click="alert('Share')">
-							Share
-						</ActionButton>
-					</Actions>
+					<ActionButton icon="icon-share" @click="alert('Share')">
+						Share
+					</ActionButton>
+				</Breadcrumb>
+				<Breadcrumb title="Folder 6" href="/Folder 1/Folder 2/Folder 3/Folder 4/Folder 5/Folder 6" :forceMenu="true">
```
skjnldsv:
```suggestion
				<Breadcrumb title="Folder 6" href="/Folder 1/Folder 2/Folder 3/Folder 4/Folder 5/Folder 6">
```
Keep it simple I'd say :wink:

raimund-schluessler: Something is messed up anyway. I will check.

skjnldsv: Okay, I'll stop reviewing and let you clean things up! This is a phenomenal work!! :tada: 
Loved it a lot!!!

Ping me again when you need me! :)

raimund-schluessler:
@skjnldsv From my side everything is done. Waiting for your review 😃

src/components/Breadcrumb/Breadcrumb.vue
```diff
@@ -0,0 +1,197 @@
...
+		@dragstart.prevent="() => {/** Prevent the breadcrumb from being draggable. */}"
+		@drop.prevent="dropped"
+		@dragover.prevent="() => {}"
+		@dragenter="dragEnter"
+		@dragleave="dragLeave">
+		<element :is="tag" :to="to" :href="href">
+			<span v-if="icon" :class="icon" class="icon" />
+			<span v-else>{{ title }}</span>
+		</element>
+		<Actions :force-menu="forceMenu">
```
skjnldsv: ALignment is not ideal, I would reduce the right padding to 2 or 3px for the link.
Capture d’écran_2020-03-28_21-49-11
___


Peek 28-03-2020 21-47
Capture d’écran_2020-03-28_21-48-01

src/components/Breadcrumb/Breadcrumb.vue
```diff
@@ -0,0 +1,197 @@
...
+			<!-- @slot All action elements passed into the default slot will be used -->
```
skjnldsv: Nice!!

raimund-schluessler:
> 
> 
> Width calculation doesn't seems to work anymore (and a small design comment!) 🙈
> Capture d’écran_2020-03-28_21-51-33

Hm, the calculation is fine, but it doesn't trigger. Resizing the browser window will trigger it. I will check why adding new elements doesn't work.

src/components/Breadcrumb/Breadcrumb.vue
```diff
@@ -38,7 +38,11 @@ This component is meant to be used inside a Breadcrumbs component.
 		@dragover.prevent="() => {}"
 		@dragenter="dragEnter"
 		@dragleave="dragLeave">
-		<element :is="tag" :to="to" :href="href">
+		<element
+			:is="tag"
+			:to="to"
+			:href="href"
+			:style="[$slots.default ? {paddingRight: '2px'} : {}]">
```
skjnldsv: I prefer adding specific classes for this rather than inline styling :)

We usually use BEM class structure, so `.crumb--with-action` ?

raimund-schluessler: Alright, will change that.

raimund-schluessler:
Resizing seems to work now. I will check tomorrow which change made it work and clean up afterwards.

skjnldsv:
Can't thank you enough for onboarding on this journey :grin: 
This is a great looking component you built!!! :bowing_woman: :hugs:

raimund-schluessler:
The `mounted` seems to be necessary to make the component reacting on added breadcrumbs on netlify. I think it is not such a big deal performance wise, since it only runs once after the component was mounted.

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -0,0 +1,510 @@
...
+			align-items: center;
+			order: 1;
+			color: var(--color-border-dark);
+			font-size: 26px;
+		}
+		&.hidden {
+			display: none;
+		}
+
+		&--with-action::v-deep a {
```
skjnldsv: This should go into the Breadcrumb component, no?

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -0,0 +1,510 @@
...
+</script>
+
+<style lang="scss" scoped>
+@import '../../fonts/scss/iconfont-vue';
+
+.breadcrumb {
+	width: 100%;
+	flex-grow: 1;
+
+	div.crumb {
```
skjnldsv: This whole section should go into the Breadcrumb component, no?

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -0,0 +1,510 @@
...
+		&--with-action::v-deep a {
+			padding-right: 2px;
+		}
+	}
+	/*
+	 * We need the 'over' rule to also apply to ActionRouter
+	 * and ActionLink, so we need a deep selector here. See
+	 * https://vue-loader.vuejs.org/guide/scoped-css.html#deep-selectors
+	 */
+	&::v-deep .over {
```
skjnldsv: Same for this section?

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: Yeah, you are right, this can go into into the Breadcrumb component.

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: This is needed for the crumbs hidden in the Actions menu as well. But we can have this one line in both components.

skjnldsv: Make sense ! :)

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: Adjusted it now.

skjnldsv:
Capture d’écran_2020-03-29_13-41-36

Ah right, now it's missing this of course :)
Can't you use the BreacCrumb component for the menu too? :thinking:

raimund-schluessler:
> 
> 
> Capture d’écran_2020-03-29_13-41-36
> 
> Ah right, now it's missing this of course :)
> Can't you use the BreacCrumb component for the menu too? 🤔

Ah, damn it.

Using the Breadcrumb component there too would mean quite some adjustments to the drag and drop handlers, I think. One would need to open the Actions menu, which is then in the Breadcrumb component. And it should only open when it has Breadcrumbs inside, not when it has e.g. the Share Button.

To be honest, just adding the breadcrumb separator style seems far easier.

skjnldsv:
> To be honest, just adding the breadcrumb separator style seems far easier.

Let's go for this then! :)

skjnldsv:
> To be honest, just adding the breadcrumb separator style seems far easier.

Let's go for this then! :)
Feel free to put it back in the BreadCrumbs component with a comment.
Don't duplicate code though ;)

raimund-schluessler:
I think I got it back working.

raimund-schluessler:
Fuck this, the hidden breadcrumbs are not hidden anymore. I need a break.

raimund-schluessler:
So, I think everything is working now.

skjnldsv:
Congratulations!! Now please enjoy your Sunday and don't come back until Tuesday :heart:

jancborchardt:
Aaawesome stuff @raimund-schluessler! :)

We should also clarify in the documentation when it is to be used (when you have lots of deep nesting) or when not (when you only have 1 or 2 levels).

For example: The Deck app and Forms never needed breadcrumbs but used it, and e.g. the News app and Mail app do not need it and never used it as they do folder management in the navigation and it’s intentionally limited to some extent. @raimund-schluessler what kind of app is it you are working on, if you like to share? :)

skjnldsv:
> what kind of app is it you are working on, if you like to share? :)

I think it was deck. But the plan was to remove it as you commented in #800, so this one will most likely be for apps :)

raimund-schluessler:
> what kind of app is it you are working on, if you like to share? :)

I work on an inventory app, because I lost track of the tools I have (and other stuff). It has folders to sort the items, that's what I use the breadcrumbs for. It's not really publishable yet, but I will do so at some point.

Looks like this:
Screenshot_2020-03-31 Inventar - Nextcloud

skjnldsv:
Looks good! Nice app idea! :)

jancborchardt:
Oh wow, that looks nice! Great that you are working out components from it! :)

PR #924, opened 2020-03-20
Force single multiselect value to be shown when the input is not active

juliushaertl:
Fixes showing the multiselect value in the user management when the Multiselect is single value but taggable and searchable. The main issue is that the input always has a `position: absolute;` set, so even though the value label has a `z-index: 1;` it will never be shown on top.

Before:
image

After:
image

@skjnldsv Please have a closer look if that has any downsides, since the multiselect styling is quite complex :wink: I also cannot reproduce this in the nextcloud-vue docs so there might be some server styles interfering here, but I could not find the culprit. :see_no_evil: 

Found when testing https://github.com/nextcloud/server/pull/18818

### DIFF
--- a/src/components/Multiselect/index.scss
+++ b/src/components/Multiselect/index.scss
@@ -10,6 +10,12 @@
 	position: relative;
 	background-color: var(--color-main-background);
 
+	/* Force single multiselect value to be shown when not active */
+	&:not(.multiselect--active) .multiselect__single {
+		position: absolute;
+		width: 100%;
+	}
+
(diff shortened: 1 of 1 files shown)


skjnldsv:
> @skjnldsv Please have a closer look if that has any downsides, since the multiselect styling is quite complex  I also cannot reproduce this in the nextcloud-vue docs so there might be some server styles interfering here, but I could not find the culprit.

I don't remember tbh.
Test the various examples of the vue docs and if it works as you expected, it's good :)

PR #937, opened 2020-03-25
Add ability to set a custom aria-label value for the actions menu button

ma12-co:
(This is strictly what I need for talk)

### DIFF
--- a/src/components/ActionButton/ActionButton.vue
+++ b/src/components/ActionButton/ActionButton.vue
@@ -44,7 +44,9 @@ If you're using a long text you can specify a title
 
 <template>
 	<li :class="{ 'action--disabled': disabled }">
-		<button class="action-button"
+		<button
+			class="action-button"
+			:aria-label="ariaLabel"
 			:class="{ focusable: isFocusable }"
 			@click="onClick">
 			<!-- icon -->
--- a/src/components/Actions/Actions.vue
+++ b/src/components/Actions/Actions.vue
@@ -57,7 +57,6 @@ https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-action
 	<!-- if only one action, check if we need to bind to click or not -->
 	<element v-if="isValidSingleAction && !forceMenu"
 		v-tooltip.auto="firstAction.text"
-		:aria-label="firstAction.text"
 		v-bind="firstActionBinding"
 		:class="[firstAction.icon, firstActionClass]"
 		class="action-item action-item--single"
@@ -252,6 +251,7 @@ export default {
 						is: 'a',
 						href: this.firstAction.href,
 						target: this.firstAction.target,
+						ariaLabel: this.firstAction.ariaLabel,
 					}
 				}
 				if (tag === 'ActionRouter') {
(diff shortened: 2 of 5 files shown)


src/components/Actions/Actions.vue
```diff
@@ -57,7 +57,7 @@ https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-action
 	<!-- if only one action, check if we need to bind to click or not -->
 	<element v-if="isValidSingleAction && !forceMenu"
 		v-tooltip.auto="firstAction.text"
-		:aria-label="firstAction.text"
+		:aria-label=" !!ariaLabel ? ariaLabel : firstAction.text"
```
nickvergessen:
```suggestion
		:aria-label="ariaLabel || firstAction.text"
```

ma12-co: Are you sure??? Wouldn't that return true or false??

nickvergessen: Open your browsers console and do:
```
var a = ''
var b = 'bar'
a || b

> "bar"
```

src/components/ActionButton/ActionButton.vue
```diff
@@ -44,7 +44,9 @@ If you're using a long text you can specify a title
 
 <template>
 	<li :class="{ 'action--disabled': disabled }">
-		<button class="action-button"
+		<button
+			class="action-button"
+			:aria-label="ariaLabel"
```
skjnldsv: I guess this could even get to the ActionTextMixin ?
And be allowed in all ActionButton/Link/Router?

ma12-co: You mean the prop??

skjnldsv: prop yes :)
But also to the other templates then :thinking:

ma12-co: Only link and router?

src/components/Actions/Actions.vue
```diff
@@ -264,6 +263,7 @@ export default {
 				if (tag === 'ActionButton') {
 					return {
 						is: 'button',
+						ariaLabel: this.firstAction['ariaLabel'],
```
skjnldsv: this is only for ActionButton

skjnldsv: Also 
```suggestion
						ariaLabel: this.firstAction.ariaLabel,
```

src/components/ActionButton/ActionButton.vue
skjnldsv: Well, All the ones that are allowed as single action, so ActionLink, ActionRouter & ActionButton
https://github.com/nextcloud/nextcloud-vue/blob/0f1a4f317763efeb8beb0cf742245c554b731f6a/src/components/Actions/Actions.vue#L249

src/components/Actions/Actions.vue
skjnldsv: Did you pushed @ma12-co ? :p

skjnldsv: This is still happening here

ma12-co: uouups :)

PR #948, opened 2020-03-27
Fix actions aria labels

ma12-co:

### DIFF
--- a/src/components/Actions/Actions.vue
+++ b/src/components/Actions/Actions.vue
@@ -251,7 +251,7 @@ export default {
 						is: 'a',
 						href: this.firstAction.href,
 						target: this.firstAction.target,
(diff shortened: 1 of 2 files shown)


src/components/Actions/Actions.vue
```diff
@@ -184,6 +188,13 @@ export default {
 			type: String,
 			default: 'action-item__menutoggle--default-icon',
 		},
+		/**
+		 * Aria label for the actions menu
+		 */
+		ariaLabel: {
+			type: String,
+			default: t('Actions'),
```
skjnldsv: Not sure you can use this here. Mixins scope doesn't go to props I think

ChristophWurst: `import { t } from '../../l10n'`

PR #951, opened 2020-03-27
fix aria-controls

### LINKED ISSUE #950: Actions aria not compliant
Because [we now use v-if for the actions](https://github.com/nextcloud/nextcloud-vue/pull/569), we became non-compliant because the menu containing the id doesn't exist on the page.

image

image

@nextcloud/accessibility

korelstar:
fixes  #950

### DIFF
--- a/src/components/Actions/Actions.vue
+++ b/src/components/Actions/Actions.vue
@@ -90,7 +90,7 @@ https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-action
 			:aria-expanded="opened"
 			@click.prevent="toggleMenu"
 			@keydown.space.exact.prevent="toggleMenu" />
-		<div v-if="opened"
+		<div v-show="opened"
 			ref="menu"
 			v-click-outside="closeMenu"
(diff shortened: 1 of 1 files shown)


raimund-schluessler:
This PR breaks opening the actions menu.

raimund-schluessler:
Reverting https://github.com/nextcloud/nextcloud-vue/pull/951/files#diff-2bfd37a0e52db490178eb64c7675ecc9R93 fixes it. No idea why, though.

raimund-schluessler:
As per https://github.com/nextcloud/nextcloud-vue/pull/952#issuecomment-605412886

> Yeah, we cannot have a vif on slot 🙈
> `<template v-if="opened"><slot /></template>`

But also, I think the `v-click-outside="closeMenu"` immediately closes the menu again (which wasn't the case before, since the element was not in the DOM due to `v-if`.

skjnldsv:
add some debug logs and see?

raimund-schluessler:
> add some debug logs and see?

Yeah, I checked it. That’s the reason.

skjnldsv:
Awesome :)

PR #952, opened 2020-03-27
Fix the validate slots function

raimund-schluessler:
The `ValidateSlot` function had two issues:

- `forEach` in combination with `splice` does not work. `splice` mutates the original array, so the `forEach` loop skips the next element after a splice occurs. Hence, some elements are not checked.
- If an array entry was not a vue component, the check if it is a forbidden component will always lead to false, since `node.componentOptions.tag` fails with `undefined`. Hence, an element which is not a html element and not a vue component would not be filtered out. This happens to e.g. text nodes which the browser sometimes renders if there is a whitespace in the markup.

This PR is necessary for #913.

### DIFF
--- a/src/utils/ValidateSlot.js
+++ b/src/utils/ValidateSlot.js
@@ -40,10 +40,13 @@ const ValidateSlot = (slots, allowed, vm) => {
 		const isVueComponent = !!node.componentOptions && typeof node.componentOptions.tag === 'string'
 		const isForbiddenComponent = isVueComponent && allowed.indexOf(node.componentOptions.tag) === -1
 
-		// if not a vue component or component not in allowed tags
-		if (isHtmlElement || isForbiddenComponent) {
-			// warn
-			Vue.util.warn(`${isHtmlElement ? node.tag : node.componentOptions.tag} is not allowed inside the ${vm.$options.name} component`, vm)
+		// if html element or not a vue component or vue component not in allowed tags
+		if (isHtmlElement || !isVueComponent || isForbiddenComponent) {
+			// only warn when html elment or forbidden component
+			// sometimes text nodes are present which are hardly removeable by the developer and spam the warnings
+			if (isHtmlElement || isForbiddenComponent) {
+				Vue.util.warn(`${isHtmlElement ? node.tag : node.componentOptions.tag} is not allowed inside the ${vm.$options.name} component`, vm)
+			}
 
 			// cleanup
 			slots.splice(index, 1)
--- a/src/utils/ValidateSlot.js
+++ b/src/utils/ValidateSlot.js
@@ -32,7 +32,9 @@ const ValidateSlot = (slots, allowed, vm) => {
 	if (slots === undefined) {
 		return
 	}
-	slots.forEach((node, index) => {
+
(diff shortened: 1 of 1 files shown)


raimund-schluessler:
Damn it, it seems this broke the Actions component. I will have a look soon.

skjnldsv:
Ah, it was working for me :thinking:

src/utils/ValidateSlot.js
```diff
@@ -32,21 +32,26 @@ const ValidateSlot = (slots, allowed, vm) => {
...
 		const isHtmlElement = !node.componentOptions && node.tag && allowed.indexOf(node.tag) === -1
 		const isVueComponent = !!node.componentOptions && typeof node.componentOptions.tag === 'string'
 		const isForbiddenComponent = isVueComponent && allowed.indexOf(node.componentOptions.tag) === -1
 
-		// if not a vue component or component not in allowed tags
-		if (isHtmlElement || isForbiddenComponent) {
-			// warn
-			Vue.util.warn(`${isHtmlElement ? node.tag : node.componentOptions.tag} is not allowed inside the ${vm.$options.name} component`, vm)
+		// if html element or not a vue component or vue component not in allowed tags
+		if (isHtmlElement || !isVueComponent || isForbiddenComponent) {
```
skjnldsv:
```suggestion
if (isHtmlElement || isVueComponent) {
```

Maybe ?

skjnldsv: isForbiddenComponent already include isVueComponent :)

skjnldsv: @raimund-schluessler

raimund-schluessler: > 
> 
> isForbiddenComponent already include isVueComponent :)

Well, not really. If it is `isVueComponent === false` then `node.componentOptions.tag === undefined` and this will lead to `(allowed.indexOf(node.componentOptions.tag) === -1) === undefined` which makes `isForbiddenComponent === false` although it is not a vue component.

raimund-schluessler:
It's not this PR which broke it. It is #951 which was also merged recently.
If I revert this line https://github.com/nextcloud/nextcloud-vue/pull/951/files#diff-2bfd37a0e52db490178eb64c7675ecc9R93 it works again.

skjnldsv:
> It's not this PR which broke it. It is #951 which was also merged recently.
> If I revert this line https://github.com/nextcloud/nextcloud-vue/pull/951/files#diff-2bfd37a0e52db490178eb64c7675ecc9R93 it works again.

Yeah, we cannot have a vif on slot :see_no_evil: 
`<template v-if="opened"><slot /></template>`

skjnldsv:
Btw <slot /> was deprecated, we should move to `<template v-slot:default>` one day :see_no_evil:

raimund-schluessler:
> Yeah, we cannot have a vif on slot 🙈
> `<template v-if="opened"><slot /></template>`

Hm, this doesn't fix it either.

raimund-schluessler:
See https://github.com/nextcloud/nextcloud-vue/pull/951#issuecomment-605414116

PR #961, opened 2020-03-29
Better sections for app containers

skjnldsv:
The Content doesn't really display properly, because this is of course not meant to be used within a page.
Shall we leave this like that?

Capture d’écran_2020-03-29_15-21-59

### DIFF
--- a/styleguide.config.js
+++ b/styleguide.config.js
@@ -15,26 +15,26 @@ module.exports = {
 		module: {
 			rules: require('./webpack.dev.js').module.rules.filter(
 				rule => rule.use !== 'eslint-loader'
-			)
-		}
+			),
+		},
 	}),
 	exampleMode: 'expand',
 	usageMode: 'expand',
 	components: 'src/components/*/*.vue',
 	getComponentPathLine(componentPath) {
-		const name = path.basename(componentPath, '.vue');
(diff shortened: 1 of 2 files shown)


skjnldsv:
Ping?

styleguide.config.js
```diff
@@ -15,75 +15,94 @@ module.exports = {
...
+		},
 	}),
 	exampleMode: 'expand',
 	usageMode: 'expand',
 	components: 'src/components/*/*.vue',
 	getComponentPathLine(componentPath) {
-		const name = path.basename(componentPath, '.vue');
-		return `import { ${name} } from '@nextcloud/vue/dist/Components/${name}';`;
+		const name = path.basename(componentPath, '.vue')
+		return `import { ${name} } from '@nextcloud/vue/dist/Components/${name}';`
```
georgehrke, line 26: We should remove the semicolon here was well :)

skjnldsv:
```suggestion
		return `import { ${name} } from '@nextcloud/vue/dist/Components/${name}'`
```

skjnldsv:
> It would be nice to show `Content` between `App containers` and `AppNavigation`, but this change is also good with me :)

I think it's sorted alphabetically, did not saw an order parametter in the vuestyleguide documentation :(

PR #962, opened 2020-03-29
Use breadcrumb component for hidden actions

raimund-schluessler:
@skjnldsv I thought about your comment to use the Breadcrumb component for the hidden actions menu, it seems this was easier than I thought.

I will enjoy my Sunday now 😉

### DIFF
--- a/src/components/Breadcrumb/Breadcrumb.vue
+++ b/src/components/Breadcrumb/Breadcrumb.vue
@@ -46,7 +46,10 @@ This component is meant to be used inside a Breadcrumbs component.
 			<span v-if="icon" :class="icon" class="icon" />
 			<span v-else>{{ title }}</span>
 		</element>
-		<Actions :force-menu="forceMenu" :open="open" @update:open="onOpenChange">
+		<Actions ref="actions"
+			:force-menu="forceMenu"
+			:open="open"
+			@update:open="onOpenChange">
 			<!-- @slot All action elements passed into the default slot will be used -->
 			<slot />
 		</Actions>
@@ -105,6 +108,13 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		/**
+		 * Open state of the Actions menu
+		 */
+		open: {
+			type: Boolean,
+			default: false,
+		},
 	},
 	data() {
 		return {
@@ -112,10 +122,6 @@ export default {
 			 * Variable to track if we hover over the breadcrumb
 			 */
 			hovering: false,
-			/**
-			 * Variable for the open state of the Actions menu
-			 */
-			open: false,
 		}
 	},
 	computed: {
--- a/src/components/Breadcrumb/Breadcrumb.vue
+++ b/src/components/Breadcrumb/Breadcrumb.vue
@@ -38,11 +38,15 @@ This component is meant to be used inside a Breadcrumbs component.
 		@dragover.prevent="() => {}"
 		@dragenter="dragEnter"
 		@dragleave="dragLeave">
-		<element :is="tag" :to="to" :href="href">
+		<element
+			:is="tag"
+			v-if="title || icon"
+			:to="to"
+			:href="href">
 			<span v-if="icon" :class="icon" class="icon" />
 			<span v-else>{{ title }}</span>
 		</element>
-		<Actions :force-menu="forceMenu">
+		<Actions :force-menu="forceMenu" :open="open" @update:open="onOpenChange">
 			<!-- @slot All action elements passed into the default slot will be used -->
 			<slot />
 		</Actions>
@@ -108,6 +112,10 @@ export default {
 			 * Variable to track if we hover over the breadcrumb
 			 */
 			hovering: false,
+			/**
+			 * Variable for the open state of the Actions menu
+			 */
+			open: false,
 		}
 	},
 	computed: {
@@ -121,6 +129,19 @@ export default {
 		},
 	},
 	methods: {
+		/**
+		 * Function to handle changing the open state of the Actions menu
+		 * $emit the open state.
+		 *
+		 * @param {boolean} open The open state of the Actions menu
+		 */
+		onOpenChange(open) {
+			/**
+			 * Event emitted when the open state of the Actions menu changes
+			 * @type {null}
+			 */
+			this.$emit('update:open', open)
+		},
 		/**
 		 * Function to handle a drop on the breadcrumb.
 		 * $emit the event and the path, remove the hovering state.
@@ -184,7 +205,27 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-.breadcrumb div.crumb {
+@import '../../fonts/scss/iconfont-vue';
+
+.crumb {
+	@include iconfont('breadcrumb');
+	background-image: none;
+	display: inline-flex;
+	height: $clickable-area;
+	padding: 0;
+
+	&::before {
+		display: flex;
+		align-items: center;
+		order: 1;
+		color: var(--color-border-dark);
+		font-size: 26px;
+	}
+
+	&--hidden {
+		display: none;
+	}
+
 	&--with-action a {
 		padding-right: 2px;
 	}
--- a/src/components/Breadcrumbs/Breadcrumbs.vue
+++ b/src/components/Breadcrumbs/Breadcrumbs.vue
@@ -89,6 +89,7 @@ import Actions from '../Actions'
 import ActionRouter from '../ActionRouter'
 import ActionLink from '../ActionLink'
 import ValidateSlot from '../../utils/ValidateSlot'
+import Breadcrumb from '../Breadcrumb'
 
 export default {
 	name: 'Breadcrumbs',
@@ -96,6 +97,7 @@ export default {
 		Actions,
 		ActionRouter,
 		ActionLink,
+		Breadcrumb,
 	},
 	props: {
 		/**
@@ -160,7 +162,7 @@ export default {
 			if (this.$refs.actions.$el.contains(e.relatedTarget)) {
 				return
 			}
-			this.$refs.actions.opened = false
+			this.$refs.actions.open = false
 		},
 		/**
 		 * Check the width of the breadcrumb and hide breadcrumbs
@@ -286,7 +288,7 @@ export default {
 				this.$emit('dropped', e, path)
 			}
 			// Close the actions menu after the drop
-			this.$refs.actions.opened = false
+			this.$refs.actions.open = false
 			// Remove all hovering classes
 			const crumbs = document.querySelectorAll('.crumb')
 			crumbs.forEach((f) => { f.classList.remove('crumb--hovered') })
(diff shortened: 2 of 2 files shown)


raimund-schluessler:
Hiding is broken by this. Will fix it.

raimund-schluessler:
I give up here.

The problem is, that when we use the Breadcrumb component for the dropdown actions, it also shows up in the `this.$slots.default` array and I can't seem to be able to filter it out. Filtering by `this.$refs.actions` does not work, and even filtering by CSS class is not working. As a result, the Actions dropdown will be hidden and the resize calculation is off.

Weirdly, this problem only occurs on netlify. When I use it directly in a Nextcloud app, it works correctly and the Actions dropdown is not in the `$slots.default` array.

raimund-schluessler:
Turns out the problem was that I didn't specify `key` for the Actions dropdown breadcrumb, so Vue used in-place patching and patched a normal Breadcrumb to be the Actions dropdown breadcrumb. This mixed up which breadcrumb was hidden and which was not.

I think it works now. I will clean up, then you can have a look.

raimund-schluessler:
@skjnldsv I also just realized that the hidden breadcrumbs where sorted wrongly, since `hiddenIndices.sort()` sorts the indices as Strings by default, so `10` would come before `8`. I fixed this as well.

src/components/Breadcrumb/Breadcrumb.vue
```diff
@@ -101,13 +105,24 @@ export default {
...
 	data() {
 		return {
 			/**
 			 * Variable to track if we hover over the breadcrumb
 			 */
 			hovering: false,
+			/**
+			 * Variable to the open state of the Actions menu
+			 */
+			opened: this.open,
```
skjnldsv: This won't work if the open var changes on the parent.
You just init it on component creation, but you also need to watch.

I think just emitting again the `open:update` event should be enough :) so that you can directly do `<Breadcrumb :open.sync="open" />`

So I would just do
```
<Actions :force-menu="forceMenu" :open="open" @update:open="onOpenChange">
...

methods: {
	onOpenChange(open) {
		this.$emit('update:open, open)
	}
}
```

raimund-schluessler:
Ready for review now.

Thanks for sticking with me @skjnldsv.

src/components/Breadcrumb/Breadcrumb.vue
raimund-schluessler: With this change it doesn't open anymore when I do ` this.$refs.actions.opened = true` in the Breadcrumbs component. How would you open it then?

skjnldsv: It doesn't open at all?
Weird! `opened` is a autonomous variable inside Actions. So changing it should work out of the box :thinking: Can you debug if it toggle twice maybe?

raimund-schluessler: Well, `this.$refs.actions.opened = true` is a bit misleading. It rather is `this.$refs.actions` is the breadcrumb component, not the Actions themself. In the `Breadcrumbs` don't have a refs to the Actions directly, since they are rendered by the `Breadcrumb` component.

skjnldsv: Just tested, and it worked :man_shrugging:

skjnldsv: Peek 30-03-2020 10-37

skjnldsv: Though, we should force a max-height :see_no_evil: 
Shall we do that for the whole Actions component or here?

raimund-schluessler: Like this 201b3c305204a4b85b0c8d70de6773db0a83afdb then?

raimund-schluessler: > Though, we should force a max-height 🙈
> Shall we do that for the whole Actions component or here?

I had a max-height in the initial PR, but I think we agreed to put it in the Actions menu.

skjnldsv:
```suggestion
```

src/components/Breadcrumb/Breadcrumb.vue
```diff
@@ -38,11 +38,15 @@ This component is meant to be used inside a Breadcrumbs component.
...
+		<element
+			:is="tag"
+			v-if="title || icon"
+			:to="to"
+			:href="href">
 			<span v-if="icon" :class="icon" class="icon" />
 			<span v-else>{{ title }}</span>
 		</element>
-		<Actions :force-menu="forceMenu">
+		<Actions :force-menu="forceMenu" :open="opened" @update:open="onOpenChange">
```
skjnldsv:
```suggestion
		<Actions :force-menu="forceMenu" :open="open" @update:open="onOpenChange">
```

raimund-schluessler:
With these changes I get

```
[Vue warn]: Property or method "open" is not defined on the instance but referenced during render. Make sure that this property is reactive, either in the data option, or for class-based components, by initializing the property. See: https://vuejs.org/v2/guide/reactivity.html#Declaring-Reactive-Properties.

found in

---> <Breadcrumb> at src/components/Breadcrumb/Breadcrumb.vue
       <Breadcrumbs> at src/components/Breadcrumbs/Breadcrumbs.vue
         <TheItemsOverview> at src/components/TheItemsOverview.vue
           <AppContent> at src/components/AppContent/AppContent.vue
             <App> at src/App.vue
               <Root>
```
when trying to open the Actions menu from the Breadcrumb component. How do we open it then?

raimund-schluessler:
@skjnldsv Would you mind just pushing your changes to this branch?

raimund-schluessler:
I might have it now.

skjnldsv:
> I might have it now.

xD Don't ask me to push if you do it too :see_no_evil:

raimund-schluessler:
> 
> 
> > I might have it now.
> 
> xD Don't ask me to push if you do it too 🙈

Sorry, did we mess something up now?

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -424,7 +424,7 @@ export default {
 				// Add handlers so the Actions menu opens on hover
 				nativeOn: {
 					dragstart: this.dragStart,
-					dragenter: () => { this.$refs.actions.opened = true },
+					dragenter: () => { this.$refs.actions.open = true },
```
skjnldsv: I don't think you can change the prop from the outside, no?

raimund-schluessler: There is no prop anymore, just a data variable. Is this not what you meant? If not, I will keep my fingers off the code and let you do the last changes 🙈

skjnldsv:
> Sorry, did we mess something up now?

Nono, all good! I was working on a different approach, but it's a bit messier!
If your solution works, it's perfect!

raimund-schluessler:
> 
> 
> > Sorry, did we mess something up now?
> 
> Nono, all good! I was working on a different approach, but it's a bit messier!
> If your solution works, it's perfect!

I think it does. I will rebase, and then let's have a last look (hopefully) 😃

src/components/Breadcrumbs/Breadcrumbs.vue
skjnldsv: Nono, I meant that open is a prop passed to the BreadCrumb component, right?
You cannot mutate prop directly, prop are not a data variable, they're supposed to just be a reference to an existing variable. 

I'm expecting Vue to throw "Vue Error: Avoid Mutating a Prop Directly"

skjnldsv:
> I think it does. I will rebase, and then let's have a last look (hopefully)

Please do :)
I'll wait :tea:

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: Yes, if `open` was a prop, there would be the error you mentioned. But open is in the data function:

```
	data() {
		return {
			/**
			 * Variable to track if we hover over the breadcrumb
			 */
			hovering: false,
			/**
			 * Variable for the open state of the Actions menu
			 */
			open: false,
		}
	},
```

so Vue is fine with mutating it.

Anyway, I will rebase now. Feel free to push any changes which you find necessary.

skjnldsv: Ah right! Then it won't work.
There is no binding between the open prop of Breadcrumb to the Actions.
You cannot do `<Breadcrumb :open="true">` anymore.
My idea was to completely get rid of the intermediate open within the Breadcrumb. Just forward the prop to the Actions directly :)

raimund-schluessler:
Rebased.

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: > You cannot do `<Breadcrumb :open="true">` anymore.

That's true. Wasn't sure if we want this.

> My idea was to completely get rid of the intermediate open within the Breadcrumb. Just forward the prop to the Actions directly :)

Do you mind implementing this? I think you have a better idea how to achieve it already.

skjnldsv:
Pushed :)

raimund-schluessler:
Now it gives this warning, since the prop is mutated directly:

```
[Vue warn]: Avoid mutating a prop directly since the value will be overwritten whenever the parent component re-renders. Instead, use a data or computed property based on the prop's value. Prop being mutated: "open"

found in

---> <Actions> at src/components/Actions/Actions.vue
       <Breadcrumb> at src/components/Breadcrumb/Breadcrumb.vue
         <Breadcrumbs> at src/components/Breadcrumbs/Breadcrumbs.vue
           <TheItemsOverview> at src/components/TheItemsOverview.vue
             <AppContent> at src/components/AppContent/AppContent.vue
               <App> at src/App.vue
                 <Root>
```

skjnldsv:
> Now it gives this warning, since the prop is mutated directly:

Ah right, this should be `opened` in Actions, sorry!  Fix incoming

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -157,10 +159,10 @@ export default {
 		 */
 		closeActions(e) {
 			// Don't do anything if we leave towards a child element.
-			if (this.$refs.actions.$el.contains(e.relatedTarget)) {
+			if (this.$refs.actionsBreadcrumb.$el.contains(e.relatedTarget)) {
 				return
 			}
-			this.$refs.actions.opened = false
+			this.$refs.actionsBreadcrumb.$refs.actions.open = false
```
raimund-schluessler: Should be `this.$refs.actionsBreadcrumb.$refs.actions.opened = false`, right?
The other occurences as well.

skjnldsv: Fixed :)

raimund-schluessler:
> 
> 
> > Now it gives this warning, since the prop is mutated directly:
> 
> Ah right, this should be `opened` in Actions, sorry! Fix incoming

Great, works well now! 👍

skjnldsv:
It would probably be cleaner to have some dedicated methods in Actions though.

so we could do `$refs.actions.close()` $refs.actions.open()`
https://github.com/nextcloud/nextcloud-vue/blob/077f1cde8a83edd449617eb7f4c3b99e09519591/src/components/Actions/Actions.vue#L361

raimund-schluessler:
> 
> 
> It would probably be cleaner to have some dedicated methods in Actions though.
> 
> so we could do `$refs.actions.close()` $refs.actions.open()`
> 
> https://github.com/nextcloud/nextcloud-vue/blob/077f1cde8a83edd449617eb7f4c3b99e09519591/src/components/Actions/Actions.vue#L361

Yeah, might be nice. But something for another PR 😉

PR #974, opened 2020-03-30
We don't need an intermediate data & watcher for Actions open state

skjnldsv:
A bit cleaner!
Since we're only using the given prop and emitting a change we don't need this weird watch/sync mechanism.

The  Actions already contain its own separation & watcher so you can modify the state without triggering a `[Vue warn]: Avoid mutating a prop directly since the value will be overwritten whenever the parent component re-renders. Instead, use a data or computed property based on the prop's value. Prop being mutated: "open"`

### DIFF
--- a/src/components/AppNavigationItem/AppNavigationItem.vue
+++ b/src/components/AppNavigationItem/AppNavigationItem.vue
@@ -145,7 +145,7 @@ Just set the `pinned` prop.
 		<div v-if="hasUtils" class="app-navigation-entry__utils">
 			<slot name="counter" />
 			<Actions menu-align="right"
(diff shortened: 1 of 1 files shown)


ChristophWurst:
> Prop being mutated: "open"

:thinking: was this even the reason for the warning? The `open` prop in this component is never directly assigned. Is it possible the warning was triggered by an `open` prop in a parent component?

ChristophWurst:
In this PR you touch `menuOpen`. The warning says `open`. And both exist.

skjnldsv:
The warning is an example and does not reflect this exact pr

PR #975, opened 2020-03-30
Cleanup code and properly toggle Breadcrumb menu

skjnldsv:
```vue
<Actions ref="actions" @open="log" @close="log">
	<ActionButton>Test</ActionButton>
	<ActionButton>Test</ActionButton>
	<ActionButton>Test</ActionButton>
</Actions>
```

~~Allows us to do `this.$refs.actions.toggleMenu(true)`~~

Cleaner than having to always pass an open property or do `this.$refs.actions.opened = true` 
What do you think @nextcloud/vuejs ?

~~We cannot use close/open methods name as `this.open` is already used by the prop~~

Fixed differently by properly passing the open prop through the children

___
This also fixes a missing close event when clicking the menu icon to close

### DIFF
--- a/src/components/Actions/Actions.vue
+++ b/src/components/Actions/Actions.vue
@@ -52,6 +52,7 @@ https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-action
 	<ActionLink icon="icon-external" title="Link" href="https://nextcloud.com" />
 </Actions>
 ```
+
 </docs>
 <template>
 	<!-- if only one action, check if we need to bind to click or not -->
@@ -332,29 +333,39 @@ export default {
 
 	methods: {
 		// MENU STATE MANAGEMENT
-		toggleMenu(e) {
-			this.opened = !this.opened
+		toggleMenu(state) {
+			if (typeof state === 'boolean') {
+				this.opened = state
+			} else {
+				this.opened = !this.opened
+			}
+
 			// focus first on menu open after opening the menu
 			if (this.opened) {
 				this.$nextTick(() => {
 					this.onOpen()
 					this.focusFirstAction()
 				})
+
 				/**
 				 * Event emitted when the popover menu is opened
-				 * @type {null}
 				 */
-				this.$emit('open', e)
+				this.$emit('open')
 			} else {
 				this.offsetX = 0
 				this.offsetY = 0
 				this.offsetYArrow = 0
 				this.rotateArrow = false
+
+				/**
+				 * Event emitted when the popover menu is closed
+				 */
+				this.$emit('close')
 			}
 
 			/**
 			 * Event emitted when the popover menu open state is changed
-			 * @type {bool}
+			 * @type {boolean}
 			 */
 			this.$emit('update:open', this.opened)
 		},
@@ -367,23 +378,22 @@ export default {
 			if (this.opened) {
 				/**
 				 * Event emitted when the popover menu open state is changed
-				 * @type {bool}
+				 * @type {boolean}
 				 */
 				this.$emit('update:open', false)
+
 				/**
 				 * Event emitted when the popover menu is closed
-				 * @type {null}
 				 */
-				this.$emit('close', e)
+				this.$emit('close')
 
+				// close everything
+				this.opened = false
+				this.offsetX = 0
+				this.offsetY = 0
+				this.offsetYArrow = 0
+				this.rotateArrow = false
 			}
-
-			// close everything
-			this.opened = false
-			this.offsetX = 0
-			this.offsetY = 0
-			this.offsetYArrow = 0
-			this.rotateArrow = false
 		},
 
 		/**
--- a/src/components/Breadcrumbs/Breadcrumbs.vue
+++ b/src/components/Breadcrumbs/Breadcrumbs.vue
@@ -123,6 +123,20 @@ export default {
 			 * Comparing two crumbs somehow does not work, so we use the indices.
 			 */
 			hiddenIndices: [],
+
+			/**
+			 * This is the props of the middle Action menu
+			 * that show the ellipsised breadcrumbs
+			 */
+			menuBreadcrumbProps: {
+				// Don't show a title for this breadcrumb, only the Actions menu
+				title: '',
+				forceMenu: true,
+				// Don't allow dropping directly on the actions breadcrumb
+				disableDrop: true,
+				// Is the menu open or not
+				open: false,
+			},
 		}
 	},
 	beforeMount() {
@@ -168,7 +182,7 @@ export default {
 			if (this.$refs.actionsBreadcrumb.$el.contains(e.relatedTarget)) {
 				return
 			}
-			this.$refs.actionsBreadcrumb.$refs.actions.opened = false
+			this.menuBreadcrumbProps.open = false
 		},
 		/**
 		 * Call the resize function after a delay
@@ -302,7 +316,8 @@ export default {
 				this.$emit('dropped', e, path)
 			}
 			// Close the actions menu after the drop
-			this.$refs.actionsBreadcrumb.$refs.actions.opened = false
+			this.menuBreadcrumbProps.open = false
+
 			// Remove all hovering classes
 			const crumbs = document.querySelectorAll('.crumb')
 			crumbs.forEach((f) => { f.classList.remove('crumb--hovered') })
@@ -424,23 +439,24 @@ export default {
 			// Use a breadcrumb component for the hidden breadcrumbs
 			crumbs.push(createElement('Breadcrumb', {
 				class: 'dropdown',
-				// We want the Actions to always show as a dropdown menu
-				props: {
-					// Don't show a title for this breadcrumb, only the Actions menu
-					title: '',
-					forceMenu: true,
-					// Don't allow dropping directly on the actions breadcrumb
-					disableDrop: true,
-					hideable: false,
-				},
+
+				props: this.menuBreadcrumbProps,
+
 				// Add a ref to the Actions menu
 				ref: 'actionsBreadcrumb',
 				key: 'actions-breadcrumb-1',
 				// Add handlers so the Actions menu opens on hover
 				nativeOn: {
 					dragstart: this.dragStart,
-					dragenter: () => { this.$refs.actionsBreadcrumb.$refs.actions.opened = true },
+					dragenter: () => { this.menuBreadcrumbProps.open = true },
 					dragleave: this.closeActions,
+				},
+				on: {
+					// Make sure we keep the same open state
+					// as the Actions component
+					'update:open': (open) => {
+						this.menuBreadcrumbProps.open = open
+					},
 				},
 			// Add all hidden breadcrumbs as ActionRouter or ActionLink
 			}, this.hiddenCrumbs.map(crumb => {
--- a/l10n/messages.pot
+++ b/l10n/messages.pot
@@ -10,7 +10,7 @@ msgstr ""
 msgid "{tag} (restricted)"
 msgstr ""
 
-#: src/components/Actions/Actions.vue:196
+#: src/components/Actions/Actions.vue:197
 msgid "Actions"
 msgstr ""
 
--- a/src/components/AppContent/AppContent.vue
+++ b/src/components/AppContent/AppContent.vue
@@ -87,6 +87,7 @@ export default {
 	min-height: 100%;
 	// Overriding server styles TODO: cleanup!
 	margin: 0 !important;
+	min-width: 0;
 }
 
 </style>


raimund-schluessler:
Want to adjust https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/Breadcrumbs/Breadcrumbs.vue#L291 and following here already?

skjnldsv:
> Want to adjust [/src/components/Breadcrumbs/Breadcrumbs.vue@`master`#L291](https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/Breadcrumbs/Breadcrumbs.vue#L291) [](/nextcloud/nextcloud-vue/blob/HEAD@{2020-03-30T10:48:30Z}/src/components/Breadcrumbs/Breadcrumbs.vue#L291) and following here already?

yes, and I also created a bug somewhere... debugging right now :)

ChristophWurst:
> Cleaner than having to always pass an open property

It's tempting. But IMO what makes Vue great is that so much is based on declared variables rather than imperative invocations. Yes, you need to have your own `open` var in the component that uses the actions menu, but also it will allows Vue to re-render the template any time, derive the state/props from your data and it will result in what you expected. With the method approach, the state remains internal to the actions menu and thus could be removed by Vue on a re-render.

Also, during unit testing you can shallow mount, run a method and assert the new state. If you call methods on child components you'd have to inject stubs with watchers to assert that a state was changed.

So this is a :-1: for me. I favor the declarative approach over the imperative one.

skjnldsv:
> but also it will allows Vue to re-render the template any time, derive the state/props from your data and it will result in what you expected. With the method approach, the state remains internal to the actions menu and thus could be removed by Vue on a re-render.

This is where you lost me.
The state is already internal to the actions, this is why I created this pr.
It doesn't mean you need to do it with this method, but it means that you can.

I really don't see the benefit of having yes another open variable in your component that just serves as a buffer to forward this to the actions.

This is the exact reason $refs have been created https://vuejs.org/v2/guide/components-edge-cases.html#Accessing-Child-Component-Instances-amp-Child-Elements



> With the method approach, the state remains internal to the actions menu and thus could be removed by Vue on a re-render.

Again, the actions state is internal anyway, and it's mandatory otherwise blocking the open prop to a boolean will prevent the menu to self close.
For me components should be self managed, if we rely on outter state, we rely on the dev to do the right thing and implement its own methods. Which I don't want to do.

ChristophWurst:
What a Vue component renders depends on its state. The state is a mix of props, data and derived data (computed props, Vuex, globals, ...). Vue caches instances of components, but sometimes also discards them and creates a new one. In that case, all internal state is lost. The open state will be reset. With props you could still create a new instance of the component and get the same result as you had before.

Does that explain it better?

I know this is a bit of a theoretic concern.

> For me components should be self managed, if we rely on outter state, we rely on the dev to do the right thing and implement its own methods. Which I don't want to do.

Yes, they should be self-contained. Anything that is not a concern to the outside world should be only accessible from within the component. But here we deliberately let the consumer of the component interfere with the state. Thus it is part of the API and not an internal anymore. The question is just how we allow the interaction.

> it's mandatory otherwise blocking the open prop to a boolean will prevent the menu to self close

No. It doesn't have to be mandatory. You can still have the prop as initial value, keep `open` as a property of *data* and `$emit('update:open')` whenever the value changes. If the outer component does not specify a prop or doesn't use the `.sync`, the emit just won't have any listener. Am I missing something?

skjnldsv:
> No. It doesn't have to be mandatory. You can still have the prop as initial value, keep `open` as a property of _data_ and `$emit('update:open')` whenever the value changes. If the outer component does not specify a prop or doesn't use the `.sync`, the emit just won't have any listener. Am I missing something?

If you just forward the prop and doesn't have intermediate state like Actions open/opened, not using a sync modifier will prevent the component from changing its state. This is what I meant, but I guess this just go back to the concept of self-contained components.


So how would you say this differs from the Sidebar for example? 
`OCA.Files.Sidebar.open(path)`

> The open state will be reset. With props you could still create a new instance of the component and get the same result as you had before.

Ah, yes sure! 
It make sense for this then :)
For me this is just a code design discussion, I really have no strong preferences, but thought it was much more cleaner to be able to call them directly than having to pass a prop though multiple components.


See the Breadcrumb example (look at this code). Either we have a local prop in the Breadcrumbs and then another in the Breadcrumb to finally reach the Actions, or we just call the action directly. The later seems much cleaner to me :)

ChristophWurst:
> If you just forward the prop and doesn't have intermediate state like Actions open/opened, not using a sync modifier will prevent the component from changing its state. This is what I meant, but I guess this just go back to the concept of self-contained components.

Interesting. So if I use the component with a variable or constant without `.sync`, the actions can't open? I thought in that case the change would just flow one-way and not back, yet it could change internally :thinking:

skjnldsv:
> Interesting. So if I use the component with a variable or constant without `.sync`, the actions can't open? I thought in that case the change would just flow one-way and not back, yet it could change internally

No, you can because I made sur it was a self contained component :wink: 
I was speaking about why it was necessary to have a local data that was not the exact reflection of the parent component state.

ChristophWurst:
> In Actions, we have the prop `open`, and the data `opened`. If we only used `open`, we would not be able to change the original variable (from the parent) if the sync modifier was not here (or listening to the update event).

Got it. I remember having the same issue in Mail.

But I think this is fine. We expose a prop so the user of the component can influence the value. And if they don't we just keep the state internal.

skjnldsv:
> But I think this is fine. We expose a prop so the user of the component can influence the value. And if they don't we just keep the state internal.


> See the Breadcrumb example (look at this code). Either we have a local prop in the Breadcrumbs and then another in the Breadcrumb to finally reach the Actions, or we just call the action directly. The later seems much cleaner to me :)

What about this ^ ?
I don't see the benefit of forcing devs to pass a prop through 2 (or more) different parent components :)

ChristophWurst:
> I don't see the benefit of forcing devs to pass a prop through 2 (or more) different parent components :)

If I understand this correctly for the `Breadcrumb.vue` we have an optional `open`, but we can't change its state as the prop might not have a value behind it? So we just need to have an internal var and a prop like we have for actions and everything works smooth both for those who use the prop (with sync binding) and those who just ignore it?

Yes, the internal data property + prop + update event is some coding work, but it makes the components nice to use on the outside, no?

> forcing devs to pass a prop through 2 (or more) different parent components

I can't follow here. Do you have an example where that is necessary? Which specific prop is it?

> much cleaner

Sorry for this comparison, but would you also consider setting a class on a DOM element with jQuery more clean because you can do it with one simple instruction?

skjnldsv:
> I can't follow here. Do you have an example where that is necessary? Which specific prop is it?

`Breadcrumbs > Breadcrumb > Actions`.
When you drag and drop, we close the main ellipsised middle `Actions` menu, unless you're hovering this menu in that case we open it. So We need to control the `Actions` component from within `Breadcrumbs` :)



> Sorry for this comparison, but would you also consider setting a class on a DOM element with jQuery more clean because you can do it with one simple instruction?

This is very different, a component is not just a dom element.
This is both an element and a javascript object that have its own functions.

ChristophWurst:
> So We need to control the `Actions` component from within `Breadcrumbs`

I think I get the picture now. So in order to make this work, every developer has to specify the prop. I think this is just a bit poorly designed, no? This is an internal requirement. You could also just emit an event to all children to close them, right? Then the consumer of the components wouldn't have to deal with that and yet the components are loosely coupled.

skjnldsv:
> If I get it right `this.$refs.actionsBreadcrumb.$refs.actions.opened = true` is the important line

yes :)


> Then the consumer of the components wouldn't have to deal with that and yet the components are loosely coupled.

This is exactly what I'm also trying to avoid :+1: 


> You could also just emit an event to all children to close them, right?

I would be fine for that.
Is there any downside to this pattern? Here it's ok because no one will use multiple breadcrumbs, but what would be the cost of performances if we were to do this for every Actions component use for example?
Theoretically speaking of course, if we think the best way to send a data through multiple layers is to use events, I'm down for that! But I have no clue about any potential backfire here :thinking:

ChristophWurst:
For one I do not think emitting an event is much more work than the update flow of a prop or invoking a setter, but also we won't emit often here. This should be fine.

skjnldsv:
> For one I do not think emitting an event is much more work than the update flow of a prop or invoking a setter, but also we won't emit often here. This should be fine.

awesome! Thanks a lot!

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -122,6 +122,21 @@ export default {
 			 * Comparing two crumbs somehow does not work, so we use the indices.
 			 */
 			hiddenIndices: [],
+
+			/**
+			 * This is the props of the middle Action menu
+			 * that show the ellipsised breadcrumbs
+			 */
+			menuBreadcrumbProps: {
```
skjnldsv, line 130: @ChristophWurst, @korelstar, @raimund-schluessler  clean enough ?
We use the prop as object here, so it's passed as a reference through the Breadcrumb and therefore we can toggle it directly from `this.menuBreadcrumbProps.open = true/false` :)

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -412,22 +427,16 @@ export default {
...
 			crumbs.push(createElement('Breadcrumb', {
 				class: 'dropdown',
-				// We want the Actions to always show as a dropdown menu
-				props: {
-					// Don't show a title for this breadcrumb, only the Actions menu
-					title: '',
-					forceMenu: true,
-					// Don't allow dropping directly on the actions breadcrumb
-					disableDrop: true,
-					hideable: false,
```
skjnldsv, line 422: @raimund-schluessler this seemed to not be usd anywhere, so I deleted it, I think it was a leftover :)

raimund-schluessler: Yes, this was a leftover. Thanks for taking care.

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: Fine with me.

skjnldsv: A little review then? :pray:

raimund-schluessler: > A little review then? 🙏

Sorry, I don't have time at the moment, but I will try to have a more detailed look in the evening.

src/components/Actions/Actions.vue
```diff
@@ -332,29 +333,39 @@ export default {
...
-				this.$emit('open', e)
+				this.$emit('open')
 			} else {
 				this.offsetX = 0
 				this.offsetY = 0
 				this.offsetYArrow = 0
 				this.rotateArrow = false
+
+				/**
+				 * Event emitted when the popover menu is opened
```
ChristophWurst, line 361: closed?

src/components/Breadcrumbs/Breadcrumbs.vue
ChristophWurst: I'm having a bit of a hard time grasping this, but I think it's fine

src/components/Actions/Actions.vue
skjnldsv:
```suggestion
				 * Event emitted when the popover menu is closed
```

src/components/Breadcrumbs/Breadcrumbs.vue
ChristophWurst: yeah, that part I got :) I was just confused to suddenly see breadcrumbs here when it originally was about actions

skjnldsv:
Rebased and suqashed, waiting for @raimund-schluessler review as this is is touching the breadcrumbs :)

raimund-schluessler:
Hm, I found a slightly weird behaviour that needs some steps to reproduce:

1. Have a breadcrumbs bar with hidden breadcrumbs
2. Drag something on the Actions menu to open the menu
3. Drop it on the three-dot menu
4. The Actions menu stays open, close it with a click outside
5. Drag something on the Actions menu again. It only opens on the second try

Before, it would open immediately, so I suspect a problem in the `toggleMenu` function https://github.com/nextcloud/nextcloud-vue/pull/975/files#diff-2bfd37a0e52db490178eb64c7675ecc9R336

I will have a look.

raimund-schluessler:
> Hm, I found a slightly weird behaviour that needs some steps to reproduce:
> 
>     1. Have a breadcrumbs bar with hidden breadcrumbs
> 
>     2. Drag something on the Actions menu to open the menu
> 
>     3. Drop it on the three-dot menu
> 
>     4. The Actions menu stays open, close it with a click outside
> 
>     5. Drag something on the Actions menu again. It only opens on the second try
> 
> 
> Before, it would open immediately, so I suspect a problem in the `toggleMenu` function https://github.com/nextcloud/nextcloud-vue/pull/975/files#diff-2bfd37a0e52db490178eb64c7675ecc9R336
> 
> I will have a look.

I couldn't figure out the problem yet, but I found other issues #1020, #1022 🙈

skjnldsv:
> I couldn't figure out the problem yet, but I found other issues #1020, #1022

I actually cannot reproduce your issues :/
It opens every time for me :thinking: 

Are you sure this is this pull request and not happening on master?

raimund-schluessler:
> > I couldn't figure out the problem yet, but I found other issues #1020, #1022
> 
> I actually cannot reproduce your issues :/
> It opens every time for me 🤔
> 
> Are you sure this is this pull request and not happening on master?

Yes, this happens only with this PR. On master it always opens. It happens locally and on netlify, in both latest Firefox and Chrome. Can you let me know which program you use for the screencast? Then I would provide a short video how the problem occurs.

skjnldsv:
> Can you let me know which program you use for the screencast?

I use Peek! :)
I tried it on Chromium browser (brave)

raimund-schluessler:
> I use Peek! :)
> I tried it on Chromium browser (brave)

I used https://github.com/NickeManarin/ScreenToGif/ now, since I am on windows.

Screencast to show the problem below:
Actions

skjnldsv:
> 3. Drop it on the three-dot menu

Ah, I did not dropped on the three dot
I thought we forbid dropping?
I'll debug

raimund-schluessler:
> Ah, I did not dropped on the three dot
> I thought we forbid dropping?
> I'll debug

Yes, dropping there is forbidden, so nothing really happens, but you can still release the mouse button there.

skjnldsv:
Ah, so I have a new bug now that I rebased @raimund-schluessler 
Peek 16-04-2020 09-31

raimund-schluessler:
> 
> 
> Ah, so I have a new bug now that I rebased @raimund-schluessler
> Peek 16-04-2020 09-31

Weird, that looks like the bug I had before https://github.com/nextcloud/nextcloud-vue/pull/1020.

skjnldsv:
> Hm, I found a slightly weird behaviour that needs some steps to reproduce:
> 
> 1. Have a breadcrumbs bar with hidden breadcrumbs
> 2. Drag something on the Actions menu to open the menu
> 3. Drop it on the three-dot menu
> 4. The Actions menu stays open, close it with a click outside
> 5. Drag something on the Actions menu again. It only opens on the second try

I pushed a fix!

raimund-schluessler:
> Ah, so I have a new bug now that I rebased @raimund-schluessler
> Peek 16-04-2020 09-31

I know its trivial, but are you sure that everything is build with the latest commits and the browser loaded it?

raimund-schluessler:
> I pushed a fix!

Nice, the `on` handler fixed the issue with the actions not opening.

raimund-schluessler:
What about the other problem with the hidden crumbs appearing on Actions hover? Does this still happen to you?

skjnldsv:
> What about the other problem with the hidden crumbs appearing on Actions hover? Does this still happen to you?

Let's merge and see in a followup pr :thinking:

raimund-schluessler:
> Let's merge and see in a followup pr 🤔

Can you tell me, under which circumstances the problem occurs? Maybe I can debug it?!

raimund-schluessler:
> Can you tell me, under which circumstances the problem occurs? Maybe I can debug it?!

I can reproduce it now.

raimund-schluessler:
> > What about the other problem with the hidden crumbs appearing on Actions hover? Does this still happen to you?
> 
> Let's merge and see in a followup pr 🤔

PR is in https://github.com/nextcloud/nextcloud-vue/pull/1027.

PR #983, opened 2020-04-02
Allow to add exclude list for v-click-outside

### LINKED ISSUE #862: AppNavigationSettings automatically close on "click outside"
In Talk we use the app navigation setting to selec tthe attachment folder.
Unluckily clicking anything in the filepicker will close the settings menu.
This feels very weird and we need a way to prevent this.

> Peek 2020-02-19 13-50

georgehrke:
fixes #862 

Example: https://github.com/nextcloud/calendar/pull/2155

### DIFF
--- a/src/components/AppNavigationSettings/AppNavigationSettings.vue
+++ b/src/components/AppNavigationSettings/AppNavigationSettings.vue
@@ -22,7 +22,7 @@
 
 <template>
 	<div id="app-settings"
-		v-click-outside="closeMenu"
+		v-click-outside="clickOutsideConfig"
 		:class="{ open }">
 		<div id="app-settings-header">
 			<button class="settings-button"
@@ -41,11 +41,15 @@
 <script>
 import { directive as ClickOutside } from 'v-click-outside'
 import { t } from '../../l10n'
+import { excludeClickOutsideClasses } from '../../mixins'
 
 export default {
 	directives: {
 		ClickOutside,
 	},
+	mixins: [
+		excludeClickOutsideClasses,
+	],
 	props: {
 		title: {
 			type: String,
@@ -56,6 +60,10 @@ export default {
 	data() {
 		return {
 			open: false,
+			clickOutsideConfig: {
+				handler: this.closeMenu,
+				middleware: this.clickOutsideMiddleware,
+			},
 		}
 	},
 	methods: {
--- a/src/mixins/excludeClickOutsideClasses.js
+++ b/src/mixins/excludeClickOutsideClasses.js
@@ -1,0 +1,72 @@
+/**
+ * @copyright Copyright (c) 2020 Georg Ehrke <georg-nextcloud@ehrke.email>
+ *
+ * @author Georg Ehrke <georg-nextcloud@ehrke.email>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+export default {
+	props: {
+		/**
+		 * A class-name or an array of class-names
+		 * to be ignored when clicking outside
+		 * an element
+		 */
+		excludeClickOutsideClasses: {
+			type: String | Array,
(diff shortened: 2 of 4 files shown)


georgehrke:
> `npm ERR! enoent ENOENT: no such file or directory, open '/home/runner/work/nextcloud-vue/nextcloud-vue/node_modules/fsevents/node_modules/detect-libc/package.json'`

Fun ...

skjnldsv:
> Fun ...

yeah, there is a weird thing going on, just reload the tests

georgehrke:
@skjnldsv Is this what you meant in https://github.com/nextcloud/nextcloud-vue/issues/862#issuecomment-607790477? :)

src/components/AppNavigationSettings/AppNavigationSettings.vue
```diff
@@ -56,6 +60,12 @@ export default {
 	data() {
 		return {
 			open: false,
+			clickOutsideConfig: {
```
georgehrke, line 63: Maybe we can find a more elegant way for this part ...

skjnldsv:
> @skjnldsv Is this what you meant in [#862 (comment)](https://github.com/nextcloud/nextcloud-vue/issues/862#issuecomment-607790477)? :)

Yes, looks great!
Works fine? Should we allow to provide something else than a class? cc @ChristophWurst & @juliushaertl

ChristophWurst:
> Works fine? Should we allow to provide something else than a class? cc @ChristophWurst & @juliushaertl

I'd say this is typically static, so a list of strings is fine.

georgehrke:
also cc @nickvergessen to check if this covers your use case :)

nickvergessen:
Maybe @ma12-co can judge better? The settings menu of the navigation should stay open when navigating throught the file picker when selecting the talk upload folder.

ma12-co:
> Maybe @ma12-co can judge better? The settings menu of the navigation should stay open when navigating throught the file picker when selecting the talk upload folder.

It should probably stay open until the next click outside i think. I.e. not close upon clicking "choose".

georgehrke:
> It should probably stay open until the next click outside i think. I.e. not close upon clicking "choose".

Can you give an example of what you mean (for Talk?)

ma12-co:
Nevermind @georgehrke, we're going to move soon to a design similar to what mail has for our global settings

skjnldsv:
@georgehrke what is missing? :)

georgehrke:
I was mostly waiting for feedback from @nickvergessen and @ma12-co 

Let me rebase and fix the linting issues.

src/mixins/excludeClickOutsideClasses.js
```diff
@@ -0,0 +1,67 @@
...
+ *
+ */
+
+export default {
+	props: {
+		excludeClickOutsideClasses: {
+			type: String | Array,
+			default: [],
+		},
+	},
```
skjnldsv, line 29: Documentation :)

georgehrke:
> Nevermind @georgehrke, we're going to move soon to a design similar to what mail has for our talk global settings

Is there already an issue for this btw? 

We also want to use that approach in the calendar, so it would be really helpful to just implement it in nc/vue.

georgehrke:
> npm ERR! 404 Not Found - GET https://registry.npmjs.org/@nextcloud/axios/-/axios-1.3.3.tgz
npm ERR! 404 
npm ERR! 404  '@nextcloud/axios@1.3.3' is not in the npm registry.

🙃

georgehrke:
Now it's failing with another random package. I guess its best to just wait half an hour ...

skjnldsv:
https://status.npmjs.org/

```
Update - The ongoing issue may also affect installation of scoped packages. We continue investigating these issues.
Aug 6, 08:55 UTC
Update - We are continuing to investigate this issue.
Aug 6, 08:29 UTC
Investigating - We are investigating reports of issues when accessing the npmjs.com website.
Aug 6, 08:29 UTC

```

src/mixins/excludeClickOutsideClasses.js
```diff
@@ -0,0 +1,72 @@
...
+			return !this.hasNodeOrAnyParentClass(event.target, excludedClassList)
+		},
+		/**
+		 * Checks if given node or any of it's parents have a class of classArray
+		 *
+		 * @param {Element} node Node to test
+		 * @param {Array} classArray List of classes to check for
+		 * @returns {Boolean}
+		 */
+		hasNodeOrAnyParentClass(node, classArray) {
```
raimund-schluessler, line 62: @georgehrke Since I have to implement something similar for https://github.com/nextcloud/tasks/issues/1276, I wonder why you didn't just use `closest()` https://developer.mozilla.org/en-US/docs/Web/API/Element/closest. Is there any downside to it?

skjnldsv: No ie11 support, but I guess this will be polyfilled by babel :thinking: 
https://caniuse.com/element-closest

raimund-schluessler: Ok, then it shouldn't matter for Tasks, since it doesn't work with IE anyway 🙈 Thanks a lot!

Issue #791, opened 2020-01-01
Open and close AppNavigation


korelstar:
Currently, the AppNavigation is always visible by default and can't be hidden. There is one exception: in mobile mode, AppNavigation is hidden by default and can be opened by clicking on the AppNavigationToggle (hamburger icon).

For the notes app, some people (@ma12-co and @jancborchardt) proposed to close the AppNavigation by default even on non-mobile mode, see the discussion around https://github.com/nextcloud/notes/issues/413#issuecomment-569496547 and https://github.com/nextcloud/notes/issues/364#issuecomment-527123829 . This allows us to use the AppNavigation in a better way.

In order to realize this, we need to do two the following things:

- [x] allow to close the AppNavigation in non-mobile mode
- [ ] let the app control if the AppNavigation should be open or closed and provide actions for this
- [x] maybe change the design of AppNavigationToggle in order to make more clear what is behind this button (e.g. for the notes app it is the list of categories). Maybe the designers have of good idea on how to improve this?

What do you think about this approach?

/cc @jancborchardt @ma12-co @nextcloud/designers

jancborchardt:
To analyze it before we get ahead of ourselves – there’s basically 2 categories of apps:
- Apps with an additional content list: Mail, Contacts (so far) and possibly Notes (if we go with moving to a 3-column layout). Here it _could_ make sense to hide the AppNavigation by default.
- Apps with only the AppNavigation: Files, Photos, Talk, Notes, Activity, Calendar, Tasks, Deck, Social, etc. Here it does _not_ make sense to hide the AppNavigation by default.

So then thinking about the special cases:
- Mail: This is a high-traffic app and if you use it for email you likely have it open all day. It would be ok to hide the AppNavigation by default for simplicity, but if you use folders a lot, it would be annoying to need to open it always, so it should stay open and remember the state.
- Contacts: This is not often accessed as people tend to do contacts stuff on mobile or integrated into the OS, and even then it’s not a high-traffic app. It would be ok to hide the AppNavigation, but showing it is also fine.
- Notes: Is for now a simple app layout with 2 columns. Moving to 3-columns would complexify it. For sure, and hiding the AppNavigation then makes sense.

ma12-co:
> And that the state if it’s open or closed is remembered per app, per user.

Totally agree on this, but we'd need some back-end support for it, right?

[...]

PR #989, opened 2020-04-03
Allow to toggle sidebar and always show the toggle

ma12-co:
- [x] Move the `AppNavigationToggle` in the `AppNavigation`
- [x] Move from #app-navigation IDs to have independent styling from server
- [x] Move `AppNavigation` open state IN the `AppNavigation`
- [x] Create a global `toggle-navigation` to emit with th event bus for any app to open the navigation (was originally the start of this pair programming, we ended up fixing far too much ^^')
- [x] Allow to toggle the navigation even when in desktop mode
- [x] Changed the animation/css from translate and shifting the `AppContent` to the left to using proper flex layout and not weird absolute positioning anymore :tada:
- [x] In mobile, opening the navigation does NOT shift the content, it just hovers it now. It avoid changing the whole layout and improve the rendering performances
- [x] Document the new global event (cc @ChristophWurst)
- [x] Emit event at first load

### QA
- [x] Swiping
- [x] Toggling navigation
- [x] Scrolling navigation
- [x] Not being able to scroll left and see the hidden navigation
- [x] Settings

fix https://github.com/nextcloud/server/issues/15831
___
@ma12-co, @skjnldsv & @ChristophWurst on a nice beer programming session on a sweet Friday afternoon :beers:

### DIFF
--- a/src/components/AppContent/AppContent.vue
+++ b/src/components/AppContent/AppContent.vue
@@ -2,6 +2,8 @@
  - @copyright Copyright (c) 2019 Christoph Wurst <christoph@winzerhof-wurst.at>
  -
  - @author Christoph Wurst <christoph@winzerhof-wurst.at>
+ - @author Marco Ambrosini <marcoambrosini@pm.me>
+ - @author John Molakvoæ <skjnldsv@protonmail.com>
  -
  - @license GNU AGPL version 3 or any later version
  -
@@ -21,10 +23,7 @@
  -->
 
 <template>
-	<main id="app-content" class="no-snapper" :style="opened ? 'transform: translateX(300px)' : ''">
-		<AppNavigationToggle :aria-expanded="opened"
-			aria-controls="app-navigation"
-			@click="toggleNavigation" />
+	<main id="app-content" class="no-snapper">
 		<!-- @slot Provide content to the app content -->
 		<slot />
 	</main>
@@ -32,7 +31,7 @@
 
 <script>
 import Hammer from 'hammerjs'
-import AppNavigationToggle from '../AppNavigationToggle'
+import { emit } from '@nextcloud/event-bus'
 
 /**
  * App content container to be used for the main content of your app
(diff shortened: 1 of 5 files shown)


skjnldsv:
Capture d’écran_2020-04-03_20-00-17

skjnldsv:
image

korelstar:
Another thing: if I want the app-navigation to be closed by default (even in non-mobile mode), then I would have to emit an event using the event-bus on app initialization? Wouldn't it be easier to provide a prop for the default state?

skjnldsv:
> Another thing: if I want the app-navigation to be closed by default (even in non-mobile mode), then I would have to emit an event using the event-bus on app initialization? Wouldn't it be easier to provide a prop for the default state?

I would be up for a default state, but does that make sense in any app? We still want to show the Navigation by default in almost all of them, no?

korelstar:
Yes, there is a need. Please see https://github.com/nextcloud/nextcloud-vue/issues/791 (and the linked issues) for a discussion on this.

skjnldsv:
> Yes, there is a need. Please see #791 (and the linked issues) for a discussion on this.

Commenting on #791 as there seems to be a mix up with various issues since it was last discussed :p

jancborchardt:
Very nice!


>* `optional` remember state ?

What is meant by this? If we have remembering, it should be per-app, and it should _always_ work, not giving app developers a way to disable it.

skjnldsv:
> What is meant by this? If we have remembering, it should be per-app, and it should _always_ work, not giving app developers a way to disable it.

It means it can be in a separate pr to keep it clean.

georgehrke:
@ma12-co Thanks! This also fixes an issue we had in the calendar, where the toggle was shown, even though there is no app-navigation. ❤️

ma12-co:
you should thank @skjnldsv more than me :)

skjnldsv:
> We should make sure that the spec for "New message" and "New contact" buttons as discussed in [nextcloud/server#16934 (comment)](https://github.com/nextcloud/server/issues/16934#issuecomment-542201187) is implemented before merging this. Otherwise the issue of **"Mobile menu toggle for navigation overlaps with content (especially in Mail & Contacts)"** will not only exist on mobile but on desktop too.

I don't know actually.
I would rather keep it simple here, otherwise it's harder to review.
Also because technically they're not the same and require us to call and decide of the proper implementation.
This doesn't create regressions technically. Let's fix the remaining issues in OP and merge it :)

src/components/AppContent/AppContent.vue
```diff
@@ -30,13 +30,40 @@
...
 /**
  * App content container to be used for the main content of your app
  *
  */
 export default {
 	name: 'AppContent',
+
+	mounted() {
+		this.mc = new Hammer(this.$el, { cssProps: { userSelect: 'text' } })
+		this.mc.on('swipeleft swiperight', e => {
```
juliushaertl, line 45: Can we add a toggle for this, as for deck where we have horizontal scrolling we had quite some issues in the past when reordering cards with drag-and-drop caused the sidebar drag to trigger.

src/components/AppContent/AppContent.vue
```diff
@@ -21,78 +23,59 @@
...
-	data() {
-		return {
-			// closed by default on mobile mode
-			opened: false,
-		}
-	},
+
 	mounted() {
 		this.mc = new Hammer(this.$el, { cssProps: { userSelect: 'text' } })
 		this.mc.on('swipeleft swiperight', e => {
```
ChristophWurst, line 45:
```suggestion
		this.mc.on('swipeleft swiperight', this.handleSwipe)
```

and `this.mc.off(...) in unmounted

src/components/AppContent/AppContent.vue
ma12-co: Done!

juliushaertl: Awesome :rocket:

src/components/AppContent/AppContent.vue
ma12-co: Done!

ma12-co:
@skjnldsv I guess those conflicts are because of #1002, what should I do here?

skjnldsv:
> @skjnldsv I guess those conflicts are because of #1002, what should I do here?

Git rebase with strategy `ours` to keep the current master files and manually `npm i @nextcloud/event-bus` again

skjnldsv:
Docs fails https://app.netlify.com/sites/nextcloud-vue-components/deploys/5e8e1c5c0203164eaddaf323

skjnldsv:
WORKED

ma12-co:
> WORKED
Oh nice!

jancborchardt:
> > We should make sure that the spec for "New message" and "New contact" buttons as discussed in [nextcloud/server#16934 (comment)](https://github.com/nextcloud/server/issues/16934#issuecomment-542201187) is implemented before merging this. Otherwise the issue of **"Mobile menu toggle for navigation overlaps with content (especially in Mail & Contacts)"** will not only exist on mobile but on desktop too.
> 
> I don't know actually.
> I would rather keep it simple here, otherwise it's harder to review.
> Also because technically they're not the same and require us to call and decide of the proper implementation.
> This doesn't create regressions technically. Let's fix the remaining issues in OP and merge it :)

Fine with me as well, as then it will make it obvious that it will have to be fixed. :D

PR #995, opened 2020-04-06
Fix datepicker design regressions

skjnldsv:
We're back to how it was :rocket: 

Peek 06-04-2020 18-51

image
image
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,7 @@ CHANGELOG.md
 
 # Unit test / coverage reports
 coverage
+
+# Docs compiled build
+styleguide/build/
+styleguide/index.html
--- a/src/components/DatetimePicker/DatetimePicker.vue
+++ b/src/components/DatetimePicker/DatetimePicker.vue
@@ -30,13 +30,46 @@
 
 ### Example
 ```vue
-	<DatetimePicker
-		:value="new Date()"
-		type="datetime"
-		@update:value="alert('Value updated')"
-		@change="alert('Date changed')" />
+<template>
+	<span>
+		<DatetimePicker
+			v-model="time"
+			type="datetime" />
+		{{ time }}
+	</span>
+</template>
+<script>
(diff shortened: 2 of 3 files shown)


skjnldsv:
> The javascript part looks good
You can test it in the docs ;)
https://deploy-preview-995--nextcloud-vue-components.netlify.com/

raimund-schluessler:
The scrollarea for the minute selection is to long. The scrollbar is not necessary at all, but you can still scroll:

Unbenannt

skjnldsv:
> The scrollarea for the minute selection is to long. The scrollbar is not necessary at all, but you can still scroll:

No this is by design, because the selected one is put on the top so they added a bottomPadding of the height of the picker I guess :)

raimund-schluessler:
> > The scrollarea for the minute selection is to long. The scrollbar is not necessary at all, but you can still scroll:
> 
> No this is by design, because the selected one is put on the top so they added a bottomPadding of the height of the picker I guess :)

Ah, I see. Seems weird at first if you don't get the concept, but it makes sense.

raimund-schluessler:
One small issue, that the button is cut-off:

Untitled

How much it is cut-off depends on the month, I think it's worst for May and doesn't happen at all for Jan.

raimund-schluessler:
Besides https://github.com/nextcloud/nextcloud-vue/pull/995#issuecomment-609982955 it looks fine.

skjnldsv:
> Ah, I see. Seems weird at first if you don't get the concept, but it makes sense.

Yeah, kinda agree :thinking:

skjnldsv:
Fixed week panel too
Capture d’écran_2020-04-07_09-10-18

skjnldsv:
Thanks everyone for helping get this in quick :)
Let me release

raimund-schluessler:
The year button is still cut-off:
Unbenannt

PR #999, opened 2020-04-07
Fix message source size

GretaD:
fixes https://github.com/nextcloud/mail/issues/2852#issuecomment-610449076

@skjnldsv you proposed for the default to be full. The large version looks better as a pop up, rather than covering the whole page. What do you think? We can also maybe raise a bit the size of the large version.

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -264,7 +264,7 @@ export default {
 		},
 		size: {
 			type: String,
-			default: 'normal',
+			default: 'large',
 			validator: size => {
 				return ['normal', 'large', 'full'].indexOf(size) !== -1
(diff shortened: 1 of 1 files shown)


skjnldsv:
> @skjnldsv you proposed for the default to be full. The large version looks better as a pop up, rather than covering the whole page. What do you think? We can also maybe raise a bit the size of the large version.

I think this fix should go in mail xD
I didn't propose to change the default, but to properly set it there https://github.com/nextcloud/mail/blob/2cb4e8c8e8da956fa70de18ccde57296bca43115/src/components/Message.vue#L56 ;)

GretaD:
You closed my 999 ticket :broken_heart:

PR #1020, opened 2020-04-15
Check by index if hiding breadcrumbs is necessary

raimund-schluessler:
Due to vues inplace patching strategy comparing the crumbs was not reliable, so it could happen, that the hidden class was not applied to a crumb that should be hidden. Checking by the index is more reliable, as it only compares Integers.


raimund-schluessler:
> I did not experienced this bug but I trust your skills :)

I saw that while trying to debug #975 with a long breadcrumb bar. When dragging over the Actions menu, the hidden crumbs would appear. It didn't happen anymore with this PR.

PR #1022, opened 2020-04-15
Adjust breadcrumb size on app-navigation toggle

raimund-schluessler:
This PR allows the breadcrumbs component to react on toggling the app-navigation.

### DIFF
--- a/src/components/AppNavigation/AppNavigation.vue
+++ b/src/components/AppNavigation/AppNavigation.vue
@@ -57,7 +57,7 @@ emit('toggle-navigation', {
 	<div
 		class="app-navigation"
 		:class="{'app-navigation--close':!open }">
-		<AppNavigationToggle :open.sync="open" />
+		<AppNavigationToggle :open="open" @update:open="toggleNavigation" />
 		<slot />
 	</div>
 </template>
@@ -109,9 +109,14 @@ export default {
 		 */
 		toggleNavigation(state) {
 			this.open = state || !this.open
-			emit('navigation-toggled', {
-				open: this.open,
-			})
+			const bodyStyles = getComputedStyle(document.body)
+			const animationLength = parseInt(bodyStyles.getPropertyValue('--animation-quick')) | 100
+
+			setTimeout(() => {
+				emit('navigation-toggled', {
+					open: this.open,
+				})
+			}, animationLength)
 		},
 		toggleNavigationByEventBus({ open }) {
 			this.toggleNavigation(open)
--- a/src/components/AppNavigationToggle/AppNavigationToggle.vue
+++ b/src/components/AppNavigationToggle/AppNavigationToggle.vue
@@ -23,7 +23,6 @@
  -->
 <template>
 	<a class="app-navigation-toggle"
-		tabindex="0"
 		href="#"
 		:aria-expanded="open"
 		aria-controls="app-navigation"
@@ -32,7 +31,6 @@
 </template>
 
 <script>
-
 export default {
 
 	name: 'AppNavigationToggle',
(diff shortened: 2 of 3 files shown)


src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -139,6 +140,9 @@ export default {
 		window.addEventListener('resize', debounce(() => {
 			this.handleWindowResize()
 		}, 100))
+		subscribe('navigation-toggled', () => this.$nextTick(() => {
+			setTimeout(() => this.handleWindowResize(), 110)
```
raimund-schluessler, line 144: The timeout here is necessary since the app-navigation toggle is animated with 100 ms duration.

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -167,6 +170,14 @@ export default {
 			}
 			this.$refs.actionsBreadcrumb.$refs.actions.opened = false
 		},
+		/**
+		 * Call the resize function after a delay
+		 */
+		delayedResize() {
+			this.$nextTick(() => {
+				setTimeout(() => this.handleWindowResize(), 110)
```
raimund-schluessler, line 178: The timeout here is necessary since the app-navigation toggle is animated with 100 ms duration.

src/components/AppNavigationToggle/AppNavigationToggle.vue
```diff
@@ -47,6 +48,9 @@ export default {
 	methods: {
 		toggleNavigation() {
 			this.$emit('update:open', !this.open)
+			emit('navigation-toggled', {
```
skjnldsv, line 51: I would rather have this in the AppNavigation component
https://github.com/nextcloud/nextcloud-vue/blob/89fab6ea8e63592d65519479ac9650cb59343927/src/components/AppNavigation/AppNavigation.vue#L60

raimund-schluessler: Fine with me, but then we need a watcher for the open state.

src/components/Breadcrumbs/Breadcrumbs.vue
skjnldsv: AH, maybe we should make sure the event is trigerred after the animation then?

raimund-schluessler: If that is possible, it would be great. No idea how, though.

skjnldsv: Let me fix too :p

raimund-schluessler:
> Peek 16-04-2020 08-18

Since I ever wondered: Which tool do you use to record these screencasts? They are really helpful.

skjnldsv:
Peek! https://github.com/phw/peek

src/components/AppNavigation/AppNavigation.vue
```diff
@@ -109,9 +109,14 @@ export default {
 		 */
 		toggleNavigation(state) {
 			this.open = state || !this.open
-			emit('navigation-toggled', {
-				open: this.open,
-			})
+			const bodyStyles = getComputedStyle(document.body)
+			const animationLength = parseInt(bodyStyles.getPropertyValue('--animation-quick')) | 100
```
skjnldsv, line 113: Damn, it should be `||`

raimund-schluessler: Indeed. Weird that this doesn't give a build error.

skjnldsv: Because it's valid, it's a binary thing

PR #1027, opened 2020-04-16
Finish toggle animation before emitting event

raimund-schluessler:
The `navigation-toggled` event was emitted to soon, before the animation finished. Hence, the breadcrumb component would get the wrong width and hide breadcrumbs although not necessary. On hovering the actions menu, a resize would be triggered and the crumbs would be visible (correctly) now.

Giving a bit more time to finish the animation fixes the issue.

Follow-up to #975 @skjnldsv 

Before:
master

After:
fix

### DIFF
--- a/src/components/AppNavigation/AppNavigation.vue
+++ b/src/components/AppNavigation/AppNavigation.vue
@@ -116,7 +116,8 @@ export default {
 				emit('navigation-toggled', {
 					open: this.open,
 				})
-			}, animationLength)
+			// We wait for 1.5 times the animation length to give the animation time to really finish.
+			}, 1.5 * animationLength)
 		},
 		toggleNavigationByEventBus({ open }) {
 			this.toggleNavigation(open)


src/components/AppNavigation/AppNavigation.vue
```diff
@@ -110,7 +110,7 @@ export default {
 		toggleNavigation(state) {
 			this.open = state || !this.open
 			const bodyStyles = getComputedStyle(document.body)
-			const animationLength = parseInt(bodyStyles.getPropertyValue('--animation-quick')) | 100
+			const animationLength = 1.5 * (parseInt(bodyStyles.getPropertyValue('--animation-quick')) | 100)
```
skjnldsv, line 113: Maybe but the multiplier on line 119 instead?
And a nice comment explaining? :p

raimund-schluessler: Ok, will do that.

raimund-schluessler: Fixed.

PR #1028, opened 2020-04-16
Update large modal size

GretaD:
Increasing only the large size doesnt do anything, since the wrapper for preview and next button stops the modal to extend. So i fixed also the modal wrapper.

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -710,12 +710,13 @@ $header-size: 50px;
 	}
 	&--large {
 		.modal-container {
-			max-width: 70%;
+			max-width: 85%;
 			max-height: 90%;
 		}
 		.prev,
 		.next {
 			width: 10%;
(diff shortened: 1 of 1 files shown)


src/components/Modal/Modal.vue
```diff
@@ -637,7 +637,7 @@ $header-size: 50px;
 		align-items: center;
 		justify-content: center;
 		width: 15%;
-		min-width: 60px;
+		min-width: 35px;
```
jancborchardt, line 640: Min-width should be 44px, no? Otherwise looks good! :)

skjnldsv: Oh right, missed that! Indeed!
You can use `$clickable-area` @GretaD

skjnldsv: But I would rather not change this here please. Because this will affect all sizes.
Please ass dedicated `.prev, .next` children in the --large one

PR #1029, opened 2020-04-17
Pass Date object for Action Input Value

gary-kim:
For addressing nextcloud/server#20520

`vue2-datepicker` expects a v-model with a `Date` but the value type checks for string only.

### DIFF
--- a/src/components/ActionInput/ActionInput.vue
+++ b/src/components/ActionInput/ActionInput.vue
@@ -129,7 +129,7 @@ export default {
 		 * value attribute of the input field
 		 */
 		value: {
-			type: String,
+			type: [String, Date, Number],
 			default: '',
 		},
 		/**


src/components/DatetimePicker/DatetimePicker.vue
```diff
@@ -142,11 +144,17 @@ export default {
 		// eslint-disable-next-line
 		value: {
```
skjnldsv, line 145: So let's add typing then?
If we're enforcing a string?

georgehrke:
I'm not quite sure why this has to be done in nc/vue and not just in the files app where it's needed.
Many apps (including the Calendar) rely on supplying a Date object to vue2-datepicker.

src/components/DatetimePicker/DatetimePicker.vue
gary-kim: I was assuming that the typing was intentional. Okay then, I'll change the typing instead.

skjnldsv:
Closing then

gary-kim:
I think we should still merge this since other apps pass in `Date` objects?

src/components/ActionInput/ActionInput.vue
```diff
@@ -129,7 +129,7 @@ export default {
 		 * value attribute of the input field
 		 */
 		value: {
-			type: String,
+			type: [String, Date],
```
georgehrke, line 132: Since there is an input-type `Number`, should we also add `Number` as a type?

skjnldsv:
> I think we should still merge this since other apps pass in `Date` objects?

I disagree, this is up to the dev to properly pass the data required by the datepicker library
https://github.com/nextcloud/server/pull/20538#discussion_r410245366
https://github.com/mengxiong10/vue2-datepicker#value-type

We're not overwriting the library, the goal is to just have a design wrapper and a few shortcuts at most

skjnldsv:
Ah right, I confused myself! This is the other pr! :roll_eyes: 
Let's get this in of course!
We should comply to what the library accept then :)

georgehrke:
@skjnldsv check the changes again. All the changes from date to string are gone :)

gary-kim:
If I'm reading this correctly, it looks like `Date`, `Number`, and `String` are accepted. Let me push that in.

https://github.com/mengxiong10/vue2-datepicker/blob/fb8a79fecaaae2a21ca127f0ecf5299a7481d632/src/date-picker.vue#L336-L343

PR #1030, opened 2020-04-17
Fix deprecated OC globals

raimund-schluessler:
See https://github.com/nextcloud/documentation/pull/1950

### DIFF
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -85,7 +85,9 @@
 <script>
 import { directive as ClickOutside } from 'v-click-outside'
 import PopoverMenu from '../PopoverMenu'
+import { getCurrentUser } from '@nextcloud/auth'
 import axios from '@nextcloud/axios'
+import { generateUrl } from '@nextcloud/router'
 import Tooltip from '../../directives/Tooltip'
 import uidToColor from './uidToColor'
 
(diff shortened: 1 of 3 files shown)


src/components/Avatar/Avatar.vue
```diff
@@ -260,7 +262,7 @@ export default {
 			if (this.isMenuLoaded) {
 				return this.menu.length > 0
 			}
-			return !(this.user === OC.getCurrentUser().uid || this.userDoesNotExist || this.url)
+			return !(this.user === getCurrentUser().uid || this.userDoesNotExist || this.url)
```
ChristophWurst, line 265: the replacement works different as it may return null when the user is not known. so it should be `getCurrentUser()?.uid`

src/components/Avatar/Avatar.vue
```diff
@@ -359,15 +361,15 @@ export default {
...
+				let avatarUrl = generateUrl(
 					url,
 					{
 						user: user,
 						size: size,
 					})
 
 				// eslint-disable-next-line camelcase
-				if (user === OC.getCurrentUser().uid && typeof oc_userconfig !== 'undefined') {
+				if (user === getCurrentUser().uid && typeof oc_userconfig !== 'undefined') {
```
ChristophWurst, line 372: same

src/components/Avatar/Avatar.vue
raimund-schluessler: Wasn't aware of this. I will provide a PR later.

raimund-schluessler: Fix is in #1039.

PR #1032, opened 2020-04-18
Stop propagation on menu open

skjnldsv:
Before
Peek 18-04-2020 22-01-2

After
Peek 18-04-2020 22-01

### DIFF
--- a/src/components/Actions/Actions.vue
+++ b/src/components/Actions/Actions.vue
@@ -90,8 +90,8 @@ https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-action
 			aria-haspopup="true"
 			:aria-controls="randomId"
 			:aria-expanded="opened"
-			@click.prevent="toggleMenu"
-			@keydown.space.exact.prevent="toggleMenu" />
+			@click.prevent.stop="toggleMenu"
+			@keydown.space.exact.prevent.stop="toggleMenu" />
 		<div v-show="opened"
 			ref="menu"
 			:class="[`menu-${menuAlign}`, { 'open': opened }]"
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -46,7 +46,7 @@
 		}"
 		:style="avatarStyle"
 		class="avatardiv popovermenu-wrapper"
-		@click="toggleMenu">
+		@click.prevent.stop="toggleMenu">
 		<!-- avatar -->
 		<div v-if="iconClass" :class="iconClass" class="avatar-class-icon" />
 		<img v-else-if="isAvatarLoaded && !userDoesNotExist" :src="avatarUrlLoaded" :srcset="avatarSrcSetLoaded">


skjnldsv:
@ma12-co I also tried recently to add a click listener on talk so that clicking the message list focus the input, but then clicking the reply or the avatar menu was triggering the click too. This fixes it and will allow to implement this in talk if the talk team wants it

raimund-schluessler:
Hm, I am not sure about this one. Sometimes you want the click to propagate and do someting. In the Calendar app clicking an action e.g. closes the sharing menu, and I think that makes sense. In the Task app e.g. clicking the sort-order dropdown, closes the list renaming input.

I think it is better if the handler that you don't want to be called takes care of detecting where the click came from.

korelstar:
@raimund-schluessler

As far as I can see, this is not about clicking an action item, but clicking the actions three dot menu. I think it's good if that one is handled internally.

raimund-schluessler:
> As far as I can see, this is not about clicking an action item, but clicking the actions three dot menu. I think it's good if that one is handled internally.

Exactly, and I think clicking the three-dot menu should be propagated so other handlers can react to it.

I haven’t tried so far, but wouldn’t this change here also lead to multiple actions menus being open at the same time?

korelstar:
There are still events emitted on open/close, so you are able to react to them if you want:
https://github.com/nextcloud/nextcloud-vue/blob/e22fb09d95f0889869019e3332fd8a07303157d6/src/components/Actions/Actions.vue#L353

raimund-schluessler:
> There are still events emitted on open/close, so you are able to react to them if you want:
> 
> https://github.com/nextcloud/nextcloud-vue/blob/e22fb09d95f0889869019e3332fd8a07303157d6/src/components/Actions/Actions.vue#L353

Doesn’t this only works for parent components?

I will try this PR in an app later today. And let’s hear one other opinion here at least. @georgehrke  @juliushaertl

skjnldsv:
> In the Task app e.g. clicking the sort-order dropdown, closes the list renaming input.

AH, you mean it break the click outside?

raimund-schluessler:
> > In the Task app e.g. clicking the sort-order dropdown, closes the list renaming input.
> 
> AH, you mean it break the click outside?

Yes, it breaks all click-outside-handlers. And you have no way to listen for this click event, since `$emit` only works for parents. So in the Tasks app, the interface looks quite broken now. And it also means, that multiple menus are now open at the same time. This makes no sense in my opinion.

This is how it looks in the Tasks app now. Before, the sharing input, the renaming input and other actions would close when you open another actions dropdown.
stopPropagation

So, I am really against stopping propagation here. It is anyway a bad habit to stop propagation and should be used carefully, https://css-tricks.com/dangers-stopping-event-propagation/. Instead, one can check in the handler that should not be activated where the event came from.

E.g
```
if(!$event.target.closest('.action-item')) return
```
would return early, when the click came from an action. There might be other solutions which don't rely on the CSS class, but it works.

skjnldsv:
> Yes, it breaks all click-outside-handlers. And you have no way to listen for this click event, since `$emit` only works for parents. So in the Tasks app, the interface looks quite broken now. And it also means, that multiple menus are now open at the same time. This makes no sense in my opinion.

right, then it's not something we want :)
This goes against the whole purpose of our Actions click outside :+1:

skjnldsv:
Thanks @raimund-schluessler :)

PR #1033, opened 2020-04-18
Replace v-html directive for Multiselect highlighting

### LINKED ISSUE #1031: Get rid of v-html directive
I would like to get rid of the usage of `v-html`. First of all, I don't like the warning message in the lint action 😉 (indeed, it sometimes hides important errors / warnings, since you have to scroll).

Also, although we `escapeHtml`, I think it would be better to use a white-list instead of a black-list approach and only render an allowed tag around a matched string, instead of escaping every possible HTML tag in our text.

There are multiple directives out there which realize this using vues render function, e.g. https://github.com/Astray-git/vue-highlight-words or https://github.com/AlbertLucianto/vue-text-highlight. I haven't tested them so far, but they seem to work and are relatively light-weight. They would also allow a bit more sophisticated stuff, such as multiple search terms, which we don't use.

The alternative would be to write a very simple directive on our own really only wrapping a match in a `strong`, using a render function.

What do you think? I would be up for writing the directive if we don't want to use an existing one.

raimund-schluessler:
This PR adds a `Highlight` directive, which highlights a substring with `<strong>`. It allows us to get rid of the currently used `v-html` directive and fixes some bugs in the current implementation.

Closes #1031.

Problems fixed compared to the current solution:
- The casing of the searched term is now correctly maintained. Before, if you searched for `opt`, `Option 1` would've become `option 1`

| master | PR | 
| ------------- |:-------------:|
| master_casing     | pr_casing |

- In the ellipsised multiselect search terms are now correctly highlighted even if they span both parts

| master | PR | 
| ------------- |:-------------:|
| master_ellips | pr_ellips |

### DIFF
--- a/src/components/Highlight/Highlight.vue
+++ b/src/components/Highlight/Highlight.vue
@@ -1,0 +1,183 @@
+<!--
+  - @copyright Copyright (c) 2020 Raimund Schlüßler <raimund.schluessler@mailbox.org>
+  -
+  - @author Raimund Schlüßler <raimund.schluessler@mailbox.org>
+  -
+  - @license GNU AGPL version 3 or any later version
+  -
+  - This program is free software: you can redistribute it and/or modify
+  - it under the terms of the GNU Affero General Public License as
+  - published by the Free Software Foundation, either version 3 of the
+  - License, or (at your option) any later version.
+  -
+  - This program is distributed in the hope that it will be useful,
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+
+<docs>
+
+### General description
+
+Highlight a string with html &lt;strong&gt;. Accepts a substring to highlight or an array with ranges.
+
+### Usage
+
+```vue
+<template>
+	<div>
+		<Highlight text="Highlight me please!" search="me" />
+		<br />
+		<Highlight text="Highlight me please!" :highlight="[{ start: 4, end: 12 }]" />
+	</div>
+</template>
+```
+</docs>
+
+<script>
+import FindRanges from '../../utils/FindRanges'
+
+export default {
+	name: 'Highlight',
+	props: {
+		/**
+		 * The string to display
+		 */
+		text: {
+			type: String,
+			default: '',
+		},
+		/**
+		 * The string to match and highlight
+		 */
+		search: {
+			type: String,
+			default: '',
+		},
+		/**
+		 * The ranges to highlight, takes precedence over the search prop.
+		 */
+		highlight: {
+			type: Array,
+			default: () => [],
+		},
+	},
+	computed: {
+		/**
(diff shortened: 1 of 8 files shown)


src/components/Highlight/Highlight.vue
```diff
@@ -0,0 +1,184 @@
...
+		 * The indice ranges which should be highlighted.
+		 * If an array with ranges is provided, we use it. Otherwise
+		 * we calculate it based on the provided substring to highlight.
+		 *
+		 * @returns {Array} The array of ranges to highlight
+		 */
+		ranges() {
+			let ranges = []
+			// If the search term and the highlight array is empty, return early with empty array
+			if (!this.search && this.highlight.length === 0) return ranges
```
ChristophWurst, line 79: please always use a block `{}` for `if` :)

src/components/Highlight/Highlight.vue
```diff
@@ -0,0 +1,184 @@
...
+				}
+			}
+
+			/**
+			 * Validate the ranges array to be within the string length
+			 * and discard ranges which are completely out of bonds.
+			 */
+			return ranges.reduce((ranges, range) => {
+				if (range.start < this.text.length && range.end > 0) {
+					ranges.push({
```
ChristophWurst, line 102: I see two red flags here.

1) The array is modified as you iterate over it.
2) You modify the array with `.push` while you could just return a new array.

src/components/Highlight/Highlight.vue
```diff
@@ -0,0 +1,184 @@
...
+		},
+	},
+	/**
+	 * The render function to display the component
+	 *
+	 * @param {Function} createElement The function to create VNodes
+	 * @returns {VNodes} The created VNodes
+	 */
+	render: function(createElement) {
+		if (!this.ranges.length) return createElement('span', {}, this.text)
```
ChristophWurst, line 177: use a block please

src/components/Highlight/Highlight.vue
raimund-schluessler: The two `ranges` are different arrays. The `ranges` in the callback (`(ranges, range) => {}`) is the accumulator of the `reduce` function, which is initialized with `[]`, pushed to in the callback and then returned as the result of the reduce function. https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce

But I can change the names of the variables to make that more clear.

raimund-schluessler:
I resolved the remarks.

In the code is a slight duplication on extracting the valid ranges from the string. The code is used here https://github.com/nextcloud/nextcloud-vue/pull/1033/files#diff-bed02bc19bf35a2f1ce703f4b41ea881R89-R99 and the same code is necessary here https://github.com/nextcloud/nextcloud-vue/pull/1033/files#diff-6b9bd2669ec4844b5ec6d1954c00024aR103-R112.

Should I extract this code into a separate file and import it in both `Hightlight.vue` and `EllipsisedOption.vue`?

skjnldsv:
> Should I extract this code into a separate file and import it in both `Hightlight.vue` and `EllipsisedOption.vue`?

You can maybe find a place for it in the utils https://github.com/nextcloud/nextcloud-vue/tree/master/src/utils

raimund-schluessler:
> You can maybe find a place for it in the utils https://github.com/nextcloud/nextcloud-vue/tree/master/src/utils

Done.

From my side it is good to review and merge.

PR #1037, opened 2020-04-20
Use title property for showing a tooltip on single actions

juliushaertl:
`text` is not available and should use `title` instead as it is there in the Action components.

### DIFF
--- a/src/components/Actions/Actions.vue
+++ b/src/components/Actions/Actions.vue
@@ -57,7 +57,7 @@ https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-action
 <template>
 	<!-- if only one action, check if we need to bind to click or not -->
 	<element v-if="isValidSingleAction && !forceMenu"
-		v-tooltip.auto="firstAction.text"
+		v-tooltip.auto="firstAction.title"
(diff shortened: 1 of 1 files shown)


skjnldsv:
> `text` is not available and should use `title` instead as it is there in the Action components.

Text is available :)
https://github.com/nextcloud/nextcloud-vue/blob/c2c4b61c88de9f16b86ed12b8278a51175abd743/src/mixins/actionGlobal.js#L44

juliushaertl:
Ah, is there a reason it only checks for the text of the default slot and not uses the title attribute then?

https://github.com/nextcloud/nextcloud-vue/blob/c2c4b61c88de9f16b86ed12b8278a51175abd743/src/mixins/actionGlobal.js#L56

skjnldsv:
> Ah, is there a reason it only checks for the text of the default slot and not uses the title attribute then?

Because we don't use the title attribute in the Actions :)
It's only used in Action text for long text as far as I recall :thinking:

skjnldsv:
What is the status here?

juliushaertl:
Can be closed, since it works fine if not using the title prop but the default slot.

PR #1039, opened 2020-04-20
Don't throw error when getCurrentUser() returns null

raimund-schluessler:
See https://github.com/nextcloud/nextcloud-vue/pull/1030#discussion_r411329088

### DIFF
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -262,7 +262,7 @@ export default {
 			if (this.isMenuLoaded) {
 				return this.menu.length > 0
 			}
-			return !(this.user === getCurrentUser().uid || this.userDoesNotExist || this.url)
+			return !(this.user === getCurrentUser()?.uid || this.userDoesNotExist || this.url)
 		},
 		shouldShowPlaceholder() {
 			return this.allowPlaceholder && (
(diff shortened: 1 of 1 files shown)


skjnldsv:
Hum, there is a dedicated babel plugin for it https://babeljs.io/docs/en/next/babel-plugin-proposal-optional-chaining.html
Are we sure it's already in babel/preset-env?

raimund-schluessler:
I don’t know, I thought the compiler would complain if it doesn’t know it.

raimund-schluessler:
> 
> 
> Hum, there is a dedicated babel plugin for it https://babeljs.io/docs/en/next/babel-plugin-proposal-optional-chaining.html
> Are we sure it's already in babel/preset-env?

It is in babel/preset-env since 7.8.0: https://babeljs.io/blog/2020/01/11/7.8.0

> Babel 7.8.0 supports the new ECMAScript 2020 features by default: you don't need to enable individual plugins for nullish coalescing (??), optional chaining (?.) and dynamic import() anymore with preset-env.

raimund-schluessler:
Is there a reason, why we still require `"@babel/preset-env": "^7.3.1",` in https://github.com/nextcloud/nextcloud-vue/blob/master/package.json#L59?

PR #1041, opened 2020-04-20
Sync package.json and package-lock.json

raimund-schluessler:
Since I found it less than obvious which package versions we currently use (https://github.com/nextcloud/nextcloud-vue/pull/1039#issuecomment-616739240), I took the liberty to sync `package.json` with `package-lock.json` to show the versions which are actually installed.

We could also think about whether we want to pin the versions (go from e.g. `"@nextcloud/auth": "^1.2.3",` to `"@nextcloud/auth": "1.2.3",`, remove the `^`), so `package.json` and `package-lock.json` would always be in sync. And since we have dependabot, we anyway always get the newest version at some point.


ChristophWurst:
> We could also think about whether we want to pin the versions (go from e.g. `"@nextcloud/auth": "^1.2.3",` to `"@nextcloud/auth": "1.2.3",`, remove the `^`), so `package.json` and `package-lock.json` would always be in sync. And since we have dependabot, we anyway always get the newest version at some point.

There is a caveat: When you have more than one package that depends on `@nextcloud/auth` and here it's locked, you'll get the code twice as soon as the version numbers are out of sync. This bloats the code unnecessarily. It's better to have the version specified in ranges so npm can deduplicate based on semver compatibility.

ChristophWurst:
Also see https://github.com/nextcloud/nextcloud-dialogs/issues/89 cc @korelstar

raimund-schluessler:
> There is a caveat: When you have more than one package that depends on `@nextcloud/auth` and here it's locked, you'll get the code twice as soon as the version numbers are out of sync. This bloats the code unnecessarily. It's better to have the version specified in ranges so npm can deduplicate based on semver compatibility.

Interesting, I didn't know that. But how do you decide which version to require then? Should we then only increase the version in `package.json` when we need a feature or fix which is only present from a certain version onwards?

ChristophWurst:
> Interesting, I didn't know that. But how do you decide which version to require then? Should we then only increase the version in `package.json` when we need a feature or fix which is only present from a certain version onwards?

Yep, with https://semver.org/. If you expect just any vX then you can use "^X.0.0" but if vX.Y added a feature you use, then you have to use "^vX.Y". npm will resolve the dependency tree accordingly :)

PR #1046, opened 2020-04-22
White background for avatars with transparency, also in dark theme

jancborchardt:
### Before
Avatars with transparency have the generated color from the letter placeholder shine through. This of course should not happen and looks all kinds of weird with logos:
Mail app avatars before Mail app avatars dark theme before

### After
Now there is always white background. Also in the dark theme, as images with transparency usually expect white as background.
Maill app avatars after Mail avatars dark theme after

Please review @nextcloud/vuejs and cc @ChristophWurst @GretaD because this is very visible in Mail.

### DIFF
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -421,6 +421,12 @@ export default {
 		background-color: var(--color-text-maxcontrast);
 	}
 
+	&:not(&--unknown) {
+		// White background for avatars with transparency, also in dark theme
+		background-color: #fff !important;
+		box-shadow: 0 0 5px rgba(0, 0, 0, 0.05) inset;
+	}
+
 	&--with-menu {
 		cursor: pointer;
 		.icon-more {


src/components/Avatar/Avatar.vue
```diff
@@ -421,6 +421,12 @@ export default {
 		background-color: var(--color-text-maxcontrast);
 	}
 
+	&:not(&--unknown) {
+		// White background for avatars with transparency, also in dark theme
+		background-color: #fff !important;
```
skjnldsv, line 426: Why is the !important needed here?

src/components/Avatar/Avatar.vue
```diff
@@ -421,6 +421,12 @@ export default {
...
+		box-shadow: 0 0 5px rgba(0, 0, 0, 0.05) inset;
```
skjnldsv, line 427: For the antialiasing?

jancborchardt: @skjnldsv what do you mean by that question?

src/components/Avatar/Avatar.vue
jancborchardt: To overwrite the inline color that the avatar placeholder generates. Try removing it and it will look like the "before" images above.

src/components/Avatar/Avatar.vue
skjnldsv: I asked the reasons of this line? :)

skjnldsv: When you touch css, I would love to have more comments on why you're adding this or this, I recently reviewed some of your css and sometimes I don't understand exactly.

Please abuse css comments, it gets removed on compile anyway  ;)

jancborchardt: Check the screenshots, this is so that there is a circle around it. :) Could also be a border, but then it would show for all images which doesn’t look good. This will only show for images with transparency.

skjnldsv: I mean code comment, not github ;)

PR #1049, opened 2020-04-22
Remove loading spinner from avatar as it makes everything seem slow, fix #910

### LINKED ISSUE #910: Loading of message list appears slow due to spinner on avatars
When you open the Mail app and things load in, actually the last thing still appearing to load are the avatar previews with big spinners on all of them:
**Mail app:**
avatars loading just show letter placeholders

**Talk app:**
Talk avatar placeholder spinner 

This makes the performance appear unnecessarily slow, as even the first message is already displayed when that spinning still happens. It even seems as if you should wait until that is properly loaded.

Instead, we should just _not_ show the spinners as the avatars are non-essential. That’s exactly what we have the letter placeholders for. :) And once the avatars are loaded, we can quickly fade them in there, replacing the letter placeholder. And if an avatar doesn’t load – that’s totally fine. :)

jancborchardt:
Fix https://github.com/nextcloud/nextcloud-vue/issues/910

Please review @ahstro @ChristophWurst @GretaD @ma12-co as this is most present in Mail and Talk. :)

### DIFF
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -39,8 +39,6 @@
 	<div v-tooltip="tooltip"
 		v-click-outside="closeMenu"
 		:class="{
-			'icon-loading': !isAvatarLoaded && size > 16,
-			'icon-loading-small': !isAvatarLoaded && size <= 16,
 			'avatardiv--unknown': userDoesNotExist,
 			'avatardiv--with-menu': hasMenu
 		}"


skjnldsv:
What does it show when it's loading then now?

jancborchardt:
> What does it show when it's loading then now?

Only the generated color of the avatar placeholder. It would indeed be better to also show the letter though – how to do that?

juliushaertl:
> Only the generated color of the avatar placeholder. It would indeed be better to also show the letter though – how to do that?

Not sure if we should do that during loading phase or only if the avatar loading is failing for some reason. I doubt that we will be able to have the placeholder with the username letter pixel-perfect to match the avatar image file and therefore have some weird change when the actual image is loaded.

juliushaertl:
> Not sure what you mean by it needing to be pixel-perfect to match the avatar image file? The placeholder is 2 initial letters, the actual image is likely a photo?

Well, for users who did not upload an avatar the image will show the two letters. That is why i mention that having a HTML/CSS based fallback will probably not match the image exactly.

> Also, are you saying the pull request seems good to you as is with just the color? I would even say for avatars where we know that there’s an image there, we should directly show a grey background instead so we have no color flash for them at all.

Yes, I'm fine with that PR. :+1:

jancborchardt:
What’s the policy here? If all is green, we merge? :)

ma12-co:
Reviewers??

jancborchardt:
> Reviewers??

Yes – "all is green" meaning:
- checks ran through
- reviewers approved
- any other stuff which blocks it.

PR #1097, opened 2020-05-17
Fix click event on actions

jotoeri:
While the ActionRouter should provide the `@click`-event according to docs, it didn't emit on click. So i looked through the actions, comparing docs to actual behavoiur resp. the click-event.

- This PR includes now to emit the Click-event on `ActionRouter`, `ActionText` and `ActionTextEditable` as given on Docs.
- The `ActionSeparator` also should emit the event according to docs. In this case i would rather remove the event on docs, as i don't see this as use-case of the separator. I just didn't find the place to edit the doc.
- Concerning the `ActionTextEditable` i'm not sure, which way would be more reasonable - emit the event as i did here, or better remove the event from docs?

### DIFF
--- a/src/components/ActionRouter/ActionRouter.vue
+++ b/src/components/ActionRouter/ActionRouter.vue
@@ -27,7 +27,8 @@
 			:exact="exact"
 			class="action-router focusable"
 			:aria-label="ariaLabel"
-			rel="noreferrer noopener">
+			rel="noreferrer noopener"
+			@click.native="onClick">
 			<!-- icon -->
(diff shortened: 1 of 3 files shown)


jancborchardt:
Relevant issue in Forms: **Close navigation on mobile on new form #380**

jancborchardt:
What about the l10n test failure? Seems unrelated as there’s nothing translatable added here?

juliushaertl:
> What about the l10n test failure? Seems unrelated as there’s nothing translatable added here?

Yes that is unrelated and was fixed on master already.

PR #1098, opened 2020-05-17
Only show mouseover-title on text in navigationItem

jotoeri:
On forms, @jancborchardt recognized, that the actions-menu of navigationItems shows the form-title on mouseover.
This PR now shifts the title-attribute, into the Title-span, so it is only visible there. Actions-Menu as well as counter do not show the title anymore.

**Before** (imagine a cursor above the small label):
grafik
grafik

**After**:
grafik

grafik

### DIFF
--- a/src/components/AppNavigationItem/AppNavigationItem.vue
+++ b/src/components/AppNavigationItem/AppNavigationItem.vue
@@ -107,7 +107,6 @@ Just set the `pinned` prop.
 <template>
 	<!-- Navigation item, can be either an <li> or a <router-link> depending on the props -->
 	<nav-element v-bind="navElement"
-		:title="title"
 		:class="{
 			'app-navigation-entry--no-icon': !isIconShown,
 			'app-navigation-entry--opened': opened,
@@ -130,7 +129,9 @@ Just set the `pinned` prop.
 				class="app-navigation-entry-icon">
 				<slot v-if="!loading" v-show="isIconShown" name="icon" />
 			</div>
-			<span class="app-navigation-entry__title">{{ title }}</span>
+			<span class="app-navigation-entry__title" :title="title">
+				{{ title }}
+			</span>
 		</a>
 
 		<AppNavigationIconCollapsible v-if="collapsible" :open="opened" @click.prevent.stop="toggleCollapse" />


jancborchardt:
Works for me, but @skjnldsv I’m curious if this was added intentionally?

jotoeri:
Well also giving the question back to you, @jancborchardt. Would it be intentional design-wise? ;)

jancborchardt:
The idea for the title on the navigation item is that it shows the entire title if it’s ellipsized – which I would also keep (of course in the example it doesn’t really get across as it’s a very short name).

The point in the issue was about the same title being shown on the actions, which seems like duplication. There my question for @skjnldsv if this was because of some accessibility guidelines. :)

jotoeri:
Ok, i see. So only showing it on title & icon? (see fixup)

jotoeri:
Rebased & Squashed.

jancborchardt:
> Ok, i see. So only showing it on title & icon? (see fixup)

Yup! I would even say that only showing it on the title is fine. :)

jotoeri:
:D ok. So now it's only on title.^^

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -130,7 +129,9 @@ Just set the `pinned` prop.
 				class="app-navigation-entry-icon">
 				<slot v-if="!loading" v-show="isIconShown" name="icon" />
 			</div>
-			<span class="app-navigation-entry__title">{{ title }}</span>
+			<span class="app-navigation-entry__title" :title="title">
```
juliushaertl, line 132: We don't need a title here as there is some actual visible text content with the exact same information. 

https://a11yproject.com/posts/title-attributes/

> Short answer: Don’t use them, except in special circumstances.

jotoeri: From side of accessibility, yes, you're right.
Jan's argument to not fully remove it, was for long titles, that get cut and only show the three dots. So they are not visible anymore, and then to be able to see the full text?

Here on forms:
grafik

jancborchardt: Good point @juliushaertl, but as @jotoeri said the initial reason why we removed it is to offer a way for mouse-users to see the whole title when it’s longer than the width of the navigation.

So to just fix the issue at hand with it appearing _too often_ but not to cut off the entire functionality, I’d say we should merge this. :) We could then separately look into if we can e.g. only show the tooltip if the text overflows in the first place, if possible?

PR #1106, opened 2020-05-22
Avoid Navigation to overflow NC-Header

jotoeri:
Just saw, that the navigation overrides the Nextcloud-Header-Bar.

The header has a z-index of 2000, vue-body of 1000 and sidebar of 1500. Therefore i propose a z-index for navigation of 1800, to still override the sidebar in case, but appear below the NC-Header.

**Before:**
grafik

**After:**
grafik


jotoeri:
Fixes #1108

PR #1113, opened 2020-05-26
Adjust Tooltip font-size

jotoeri:
On forms, i want to use the tooltip, and wanted to slightly increase the font-size for readability to 13px.

@jancborchardt now said, that according to server-guidelines, no font should be smaller than 14px - So here it is - tooltip on 14px.


jancborchardt:
What does everyone think about the 14px sizing @nextcloud/designers?
It should be fine, but if that feels too big, 13px is also ok for now.

juliushaertl:
Any reason why we have a fixed value there? Just removing the explicit font size on the tooltip should make sure that it picks up the 0.875rem value from the body element.

ma12-co:
Ideally we should always use rem isn't it?

jotoeri:
Fine by me - Is there anything else to remove then? E.g. font-family? Or maybe even all font attributes?

jotoeri:
Ok, so i gave it a try, just removing step by step font- and text related attributes. As i committed now as fixup, i didn't see any difference on font.

Using full changes after fixup:
grafik

After first commit, only increasing font-size
grafik

Full width for reference after fixup
grafik

jancborchardt:
So for me it’s good if the font-size is the same as the base text size. @skjnldsv can you comment as to whether this whole font info should be inside this component?

juliushaertl:
It looked odd to me because I had my browser use a different default font size which then applies to the regular text due to the rem usage but not if you have a fixed value as in `14px` which the pr initially had. Since we don't use a fixed size in the server anymore but `100%` we should also not use fixed sizes anywhere else.

PR #1126, opened 2020-05-30
Fix actions navigation by keyboard

jotoeri:
Related and partially fixes the issue by @jancborchardt at forms https://github.com/nextcloud/forms/issues/416

- First commit changes to only prevent default key-events, if the action-menu is opened. Like this, navigating through content does not get killed. If closed, navigation works as usual, if opened, navigation is caught into action-menu.
- Second commit allows to focus the three-dot actionmenu-button, to be able to close the menu again on space by keyboard. Further, the keyboard navigation overflows, so people can navigate through the actionmenu in a round.
- Special change in second commit, is not to focus the first element, when opening the menu. First looks a bit unfamiliar, as we know the menu until now with always the blue bar. The reason behind this change is that in order to open the menu on keyboard by pressing space-bar and directly close it again, people (or at least me) would typically just press the space-bar a second time. If focusing the first action after opening menu, on the second space-press this action is executed. This is especially bad, if the action contains destructive operation.
- Third commit then avoids to set active-class on a foreign parent-element as e.g. App-Navigation, when focusing the three-dot-menu. Therefore i included a class `action` to mark and then focus the parent-li only on actions.

@jancborchardt To me, navigation is logical like this. When actionmenu is open, i tab through it in circles and can close it again and then continue on body. Can you check (e.g. on forms) if this is common thing or just in my head? ;)

### DIFF
--- a/l10n/messages.pot
+++ b/l10n/messages.pot
@@ -10,7 +10,7 @@ msgstr ""
 msgid "{tag} (restricted)"
 msgstr ""
 
-#: src/components/Actions/Actions.vue:247
+#: src/components/Actions/Actions.vue:249
 msgid "Actions"
 msgstr ""
 
--- a/src/components/ActionButton/ActionButton.vue
+++ b/src/components/ActionButton/ActionButton.vue
@@ -43,7 +43,7 @@ If you're using a long text you can specify a title
 </docs>
 
 <template>
-	<li :class="{ 'action--disabled': disabled }">
+	<li class="action" :class="{ 'action--disabled': disabled }">
 		<button
 			class="action-button"
 			:aria-label="ariaLabel"
--- a/src/components/ActionCheckbox/ActionCheckbox.vue
+++ b/src/components/ActionCheckbox/ActionCheckbox.vue
@@ -35,7 +35,7 @@ This component is made to be used inside of the [Actions](#Actions) component sl
 </docs>
 
 <template>
-	<li :class="{ 'action--disabled': disabled }">
+	<li class="action" :class="{ 'action--disabled': disabled }">
 		<span class="action-checkbox">
 			<input :id="id"
 				ref="checkbox"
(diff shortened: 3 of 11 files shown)


jancborchardt:
> When actionmenu is open, i tab through it in circles and can close it again and then continue on body.

Only one thing here (as discussed :) – after the last item of the action menu, the first answer should be focused and the action menu should close. (Instead of the menu looping.)

> Special change in second commit, is not to focus the first element, when opening the menu. First looks a bit unfamiliar, as we know the menu until now with always the blue bar. The reason behind this change is that in order to open the menu on keyboard by pressing space-bar and directly close it again, people (or at least me) would typically just press the space-bar a second time. If focusing the first action after opening menu, on the second space-press this action is executed. This is especially bad, if the action contains destructive operation.

So I totally agree with this. I just remember @skjnldsv said that this was intentional because of some guideline – could you clarify @skjnldsv?

jotoeri:
> Only one thing here (as discussed :) – after the last item of the action menu, the first answer should be focused and the action menu should close. (Instead of the menu looping.)

Included that. @jancborchardt can you double-check? :)

jancborchardt:
Great stuff, works almost perfectly now! :) Only 3 things I can find:
- [ ] If the question title is empty (but there are answers) it’s not possible to Tab into the answers, only into the question title field. Guess because the question is already editable, somehow the edit mode doesn’t get triggered for the whole set?
- [ ] Shift-Tabbing into answers from a previous question doesn’t seem to work?
- [ ] The navigation is dropped to the bottom? Also overlaps 1 px weirdly :D Not sure if related to this pull request or a Vue component issue:
   image

jotoeri:
Take care, @jancborchardt you're mixing up stuff here - This PR is only about tabbing through the Actions-Menu!
- Your first one is an interesting point, but a problem of my forms-PR https://github.com/nextcloud/forms/pull/427
- Your second point we talked about already, i have this in mind, but also is a problem of the forms-PR.
- Your third point will be solved by the pending-navigation PR on forms, as well as by #1123 . So also not part of this PR. 😉

jancborchardt:
@juliushaertl is this good to go from your view then? :)

PR #1148, opened 2020-06-08
Enh/active hover focus highlight

juliushaertl:
This aims to let the highlight on hover/focus/active state fill the whole line of the AppNavigationItem.

Before:
image

After:
Peek 2020-06-14 15-37

<details>
<summary>Solved</summary>
The only issue currently is that CSS hover state propagates to the parent nodes as well, so when hovering a nested child item, the whole parent will have a background set:

image

Any ideas how we could work around that with CSS?
</details>

### DIFF
--- a/src/components/AppNavigationItem/AppNavigationItem.vue
+++ b/src/components/AppNavigationItem/AppNavigationItem.vue
@@ -72,12 +72,12 @@ Wrap the children in a template with the `slot` property and use the prop `allow
 prevent the user from collapsing the items.
 
 ```
-<AppNavigationItem title="Item with children" :allowCollapse="true">
+<AppNavigationItem title="Item with children" :allowCollapse="true" :open="true">
 	<template>
 		<AppNavigationItem title="AppNavigationItemChild1" />
-		<AppNavigationItem title="AppNavigationItemChild2" />
-		<AppNavigationItem title="AppNavigationItemChild3"  />
-		<AppNavigationItem title="AppNavigationItemChild4"  />
+		<AppNavigationItem class="active" title="AppNavigationItemChild2" />
+		<AppNavigationItem title="AppNavigationItemChild3" />
+		<AppNavigationItem title="AppNavigationItemChild4" />
 	</template>
 </AppNavigationItem>
 ```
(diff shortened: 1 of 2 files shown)


juliushaertl:
Fixed the hover propagation, ready for review :)

jancborchardt:
@ma12-co @jakobroehrl are you up for second review? :)

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -440,13 +440,21 @@ export default {
 	box-sizing: border-box;
 	width: 100%;
 	min-height: $clickable-area;
+
 	// When .active class is applied, change color background of link and utils. The
 	// !important prevents the focus state to override the active state.
-	&.active,
-	&.active > a,
-	&.active > a ~ .app-navigation-entry__utils {
```
jotoeri, line 446: These are now included anyways into the `.active` without specification.

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -457,20 +465,6 @@ export default {
...
-		background-color: var(--color-main-background) !important;
-	}
-
-	&:hover,
-	&:hover > .app-navigation-entry__utils,
-	&:focus,
-	&:focus ~ .app-navigation-entry__utils {
-		background-color: var(--color-background-hover);
-	}
-
```
jotoeri: Simplified version above.

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -487,17 +481,6 @@ export default {
...
-		&:hover ~ .app-navigation-entry__utils,
-		&:focus,
-		&:focus ~ .app-navigation-entry__utils {
-			background-color: var(--color-background-hover);
-		}
-		&.active,
-		&:active,
-		&:active ~ .app-navigation-entry__utils {
-			background-color: var(--color-primary-light);
-		}
```
jotoeri, line 485: Now done on parent element for the whole line.

jotoeri:
Hm, focus does not bubble, so the parent-li does not show up focused, if the link inside gets focused. @jancborchardt do you quickly know a solution for this?

juliushaertl:
Oh that is indeed a lot simpler @jotoeri :+1: 

Amazing work.

jotoeri:
@juliushaertl Do you want to squash the commits? Or should i? This one should be good to go then, no?

PR #1150, opened 2020-06-10
Fix sidebar title height with subtitle

nickvergessen:
open the sidebar of a deck card:

Before | After
---|---
Bildschirmfoto von 2020-06-10 09-44-24 | Bildschirmfoto von 2020-06-10 09-44-27

not sure why the height was there in first place. But the padding from 
https://github.com/nextcloud/nextcloud-vue/blob/545b8ecb645d13cfafe1e25a2ad5f2c57fd4afe2/src/components/AppSidebar/AppSidebar.vue#L569

Seems to cause the problems with a fixed height.
So either we add more magic, or we don't use a fixed height anymore

### DIFF
--- a/src/components/AppSidebar/AppSidebar.vue
+++ b/src/components/AppSidebar/AppSidebar.vue
@@ -619,13 +619,8 @@ $top-buttons-spacing: 6px;
 			}
 			&--with-subtitle {
 				justify-content: space-between;
-				height: $desc-height;
-			}
-			&--editable {
-				height: $desc-height * .75;
 			}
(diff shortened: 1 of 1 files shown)


skjnldsv:
Hum, it doesn't seems to be an issue anywhere else, why only on deck?

nickvergessen:
Because deck is the only one which uses a subtitle?

juliushaertl:
Actually it was a deck issue with an old global css rule that caused this, see fix in https://github.com/nextcloud/deck/pull/2019

PR #1158, opened 2020-06-16
Export uidToColor as usernameToColor

nickvergessen:
Allows Talk (grid view) and Contacts (top header) to still calculate the color from the user name since `toRGB()` from Server is deprecated.


nickvergessen:
Follow up for talk: https://github.com/nextcloud/spreed/pull/3817

PR #1160, opened 2020-06-18
Revert "Fix Provide Navigation List" & provide footer-slot

jotoeri:
It's a pity, and i'm really sorry for that back & forth. But the fix to avoid the breaking-change makes the actions-menu on navigation hiding within the small ul, instead of flipping or showing above.
Easiest way to solve this, is now to revert the mentioned commit, unfortunately then making the already announced breaking change really a change, that breaks the manual usage of ul, but forces to use the slot.

For reference: 
- The navigation-list PR, released as breaking change: #1107
- The PR, that wanted to avoid the navigation-PR to be breaking, also within current release: #1123
- This PR: Now becoming itself a breaking change, as the first one looked like not braking anything (see vue-public talk)? 😟 

**Before**
grafik
**After**
grafik

### DIFF
--- a/src/components/AppNavigation/AppNavigation.vue
+++ b/src/components/AppNavigation/AppNavigation.vue
@@ -166,6 +166,7 @@ export default {
 	ul,
 	&__list {
 		position: relative;
+		height: 100%;
 		width: inherit;
 		overflow-x: hidden;
 		overflow-y: auto;
--- a/src/components/AppNavigation/AppNavigation.vue
+++ b/src/components/AppNavigation/AppNavigation.vue
@@ -65,6 +65,9 @@ emit('toggle-navigation', {
 		<ul class="app-navigation__list">
 			<slot name="list" />
 		</ul>
+
+		<!-- Footer for e.g. AppNavigationSettings -->
+		<slot name="footer" />
 	</div>
 </template>
 


nickvergessen:
That's awkward, as e.g. Talk defines mutliple ul's and that doesn't work anymore? 🤔

jotoeri:
Hm... The `height=100%` were included on vue-components V2.0 until now... So that just reverts the old setting.

And currently, the actions-menu on talk also does strange thing (due to that missing height)? 
grafik

I moreover also recognized on talk now, that the AppNavigationSettings go into the first slot, so maybe an additional footer-slot would be good. (see fixup)
- Looking into talk, you use on the navigation just one 'master'-ul. The multiple ul, that you mentioned are then within this list. So with this PR now (including fixup), if you change the current manual ul (the one with `class="left-sidebar__list"`) into the new `template #list` and put the AppNavigationSettings then within the `template #footer`, then everything should be like before...?

jotoeri:
@nickvergessen I saw you talked about on community-call, yesterday? So anything new here?

korelstar:
@skjnldsv Could you please publish a release with this fix?

skjnldsv:
ANyone can, please just create a tag with `npm version`, then generate a proper changelog, push the commit AND the tag, then create a release on github with the changelog ;)

korelstar:
Thanks, I tried it and it seems to work ... nice!

skjnldsv:
@korelstar you missed the part where you push the proper commit to master

> push the commit AND the tag

https://github.com/nextcloud/nextcloud-vue/blob/910d709ed2da52b05658bee450a6110ab6ff3b3b/package.json#L3

korelstar:
Ouch. I thought that `git push` will push the commit ;-)

However, I don't have the permissions to push directly to `master`. Therefore, I'm creating a PR now. But that makes this process more uncomfortable ...

skjnldsv:
> However, I don't have the permissions to push directly to `master`. Therefore, I'm creating a PR now. But that makes this process more uncomfortable ...

Ah, yes, this is unfortunate.
I guess it's fine to open a pr for this and push/create the tag after merge :shrug:

nickvergessen:
Or you do
git push origin master:anotherbranch
And then someone from us copies it over to master

marcelklehr:
mmmh, shouldn't this be v3 then following semver? :cry:

PR #1171, opened 2020-06-24
Fix/user bubble/slot size

skjnldsv:
image

- Cleaning up docs
- Added new slot to add whatever we want inside the bubble after the label
- Added docs for this new feature too
- Removed the use of the Popover component if the popover was empty, this avoid a weird unecessary complex html structure for such a simple user bubble
- Added a way to set size and avatar margin for the whole component

image

### DIFF
--- a/src/components/UserBubble/UserBubble.vue
+++ b/src/components/UserBubble/UserBubble.vue
@@ -36,8 +36,8 @@ This component has the following slot:
 
 ```vue
 <p>
-	Some text before <user-bubble user="admin" display-name="Admin Example" :url="'/test'">@admin@foreign-host.com</user-bubble> and after the bubble.
-	<user-bubble avatar-image="icon-group" display-name="test group xyz" :primary="true">Hey there!</user-bubble>
+	Some text before <UserBubble user="admin" display-name="Admin Example" :url="'/test'">@admin@foreign-host.com</UserBubble> and after the bubble.
+	<UserBubble avatar-image="icon-group" display-name="test group xyz" :primary="true">Hey there!</UserBubble>
 </p>
 ```
 
--- a/src/components/UserBubble/UserBubble.vue
+++ b/src/components/UserBubble/UserBubble.vue
@@ -35,39 +35,65 @@ This component has the following slot:
 ### Examples
 
 ```vue
-<template>
-	<p>
-		Some text before
-		<user-bubble :user="'admin'" :display-name="'Admin Example'" :url="'/test'">
-			@admin@foreign-host.com
-		</user-bubble>
-		and after the bubble. <user-bubble :avatar-image="'icon-group'" :display-name="'test group xy'" :primary="true">Hey there!</user-bubble>
-	</p>
-</template>
+<p>
+	Some text before <user-bubble user="admin" display-name="Admin Example" :url="'/test'">@admin@foreign-host.com</user-bubble> and after the bubble.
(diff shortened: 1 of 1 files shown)


skjnldsv:
>  I have no idea what I'm reviewing here, sorry

See the vue docs deploy preview :)
I also added the list of changes :rocket: 

Basically, better perfs, improved simpler design, added size props & title slot

violoncelloCH:
I can take a look this evening :)

src/components/UserBubble/UserBubble.vue
```diff
@@ -152,33 +240,37 @@ export default {
...
+	}
 
-.user-bubble-primary {
-	color: var(--color-primary-text);
-	background-color: var(--color-primary-element);
-}
+	&__content {
+		display: inline-flex;
+		max-width: 100%;
+		background-color: var(--color-background-dark);
```
violoncelloCH, line 256: Now that we have the title slot we should probably have `padding-right` on the content instead of margin-right on the title, as otherwise there is some issue at the right side of the bubble if there is something in the slot:
image

styleguide.config.js
```diff
@@ -22,7 +22,7 @@ module.exports = {
 	components: 'src/components/*/*.vue',
 	getComponentPathLine(componentPath) {
 		const name = path.basename(componentPath, '.vue')
-		return `import { ${name} } from '@nextcloud/vue/dist/Components/${name}'`
+		return `import ${name} from '@nextcloud/vue/dist/Components/${name}'`
```
violoncelloCH, line 25: is this related to the UserBubble component?

src/components/UserBubble/UserBubble.vue
```diff
@@ -83,47 +106,100 @@ export default {
...
+		/**
+		 * Override generated avatar, can be an url or an icon class
+		 */
 		avatarImage: {
 			type: String,
-			default: '',
+			default: undefined,
+			validator: str => {
+
+			}
```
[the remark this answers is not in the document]
skjnldsv: Nothing, leftover, sorry! Forgot to push :see_no_evil:

src/components/UserBubble/UserBubble.vue
```diff
@@ -83,47 +106,100 @@ export default {
...
-			return false
+		/**
+		 * If userbubble is empty, let's NOT
+		 * use the Popovercomponent
+		 * @returns {boolean}
+		 */
+		isPopover() {
+			return !this.popoverEmpty
+				? 'Popover'
+				: 'div'
```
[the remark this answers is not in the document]
skjnldsv: Definitely wrong! Thanks for the catch!

src/components/UserBubble/UserBubble.vue
skjnldsv: Good idea!!

styleguide.config.js
skjnldsv: Yes and no!
This is related to all components, I noticed the import text was wrong, we cannot import with the brackets if we use the direct component path.
I will move this to another pr for better clarity, let's not use this as a free pass!

skjnldsv:
@violoncelloCH everything should be addressed now :)
I specifically left out the title slot from the auto ellipsis. Because it's up to the dev to put whatever and handle it. If it's plain text, considering the title attribute is most likely the bets option here :wink:

PR #1172, opened 2020-06-26
Fix UserBubble documentation/click propagation & documentation config

skjnldsv:
After #1171 

image
image

### DIFF
--- a/styleguide.config.js
+++ b/styleguide.config.js
@@ -22,7 +22,7 @@ module.exports = {
 	components: 'src/components/*/*.vue',
 	getComponentPathLine(componentPath) {
 		const name = path.basename(componentPath, '.vue')
-		return `import { ${name} } from '@nextcloud/vue/dist/Components/${name}'`
+		return `import ${name} from '@nextcloud/vue/dist/Components/${name}'`
 	},
 	sections: [
 		{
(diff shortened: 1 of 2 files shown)


styleguide.config.js
```diff
@@ -22,7 +22,7 @@ module.exports = {
 	components: 'src/components/*/*.vue',
 	getComponentPathLine(componentPath) {
 		const name = path.basename(componentPath, '.vue')
-		return `import { ${name} } from '@nextcloud/vue/dist/Components/${name}'`
```
violoncelloCH, line 25: so are those parentheses not needed anymore or why are we removing them from the docs?

skjnldsv: No they're actually breaking.
We don't export submodules when getting the direct path.
See the readme: https://github.com/nextcloud/nextcloud-vue#usage

> To use a component, just import it:
> `import { AppNavigation } from '@nextcloud/vue'`
> 
> Depending on which components you use, you might want to only import individual (separately bundled) components:
> `import Avatar from '@nextcloud/vue/dist/Components/Avatar'`

violoncelloCH: thanks for the explanation @skjnldsv !

Issue #1140, opened 2020-06-05
AppNavigationItem sublevels


dehnhardt:
To display the folder structure in the mail app in a reasonable way, we have to display it in a multi-level structure. (https://github.com/nextcloud/mail/pull/3147)
We see the following problems there:
* from the 2nd level onwards, the folder icons are not hidden, as is the case with the first level, but are displayed further below the arrow icons. 
![image](https://user-images.githubusercontent.com/5556832/83886150-46c42e00-a747-11ea-8421-d5376eac99bd.png)
* In addition, the indentation in the layers is so wide that the width of the navigation is quickly insufficient.

I would like to request the following changes:
* The folder icons should be hidden in sublevels in the same way as on the first level
* The indentation of the sub-levels should be minimized (maybe not more than the with of the folder icon?):
Before
![image](https://user-images.githubusercontent.com/5556832/83888747-8e988480-a74a-11ea-8d47-12eb1d45815e.png)
After
![image](https://user-images.githubusercontent.com/5556832/83888784-99531980-a74a-11ea-8e1a-d3e354a06347.png)

jancborchardt:
Agree with your suggestion @dehnhardt! :) Just one detail which looks off in the proposal mockup: We should make sure that the icons of sub-entries are left-aligned with the text of the parent entry consistently.

dehnhardt:
You are right @jancborchardt, this will look better - even if it uses more of the space...

PR #1176, opened 2020-06-30
reduce indentation of the sub-levels

dehnhardt:
Signed-off-by: Holger Dehnhardt <holger@dehnhardt.org>

Fixes: https://github.com/nextcloud/nextcloud-vue/issues/1140

The indention of a sublevel now starts with the right edge of the parent.

From the second level on, :hover previously displayed the arrow icon and the navigation icon on top of each other.
image

The navigation icon is now also hidden for lower levels. 
Nav2

Unfortunately, :hover fades out all the navigation icons of the same branch. But I don't know what to do about that;-)
Nav1

### DIFF
--- a/src/components/AppNavigationItem/AppNavigationItem.vue
+++ b/src/components/AppNavigationItem/AppNavigationItem.vue
@@ -514,7 +514,7 @@ export default {
 		.app-navigation-entry {
 			display: inline-flex;
 			flex-wrap: wrap;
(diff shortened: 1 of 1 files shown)


skjnldsv:
cc @jancborchardt

dehnhardt:
I expected that only the icon over which the mouse pointer is current is hidden.
But if it is intended to hide all icons, there is no reason not to apply the changes this way.

PR #1184, opened 2020-07-04
Fix scrolling problem in the menu of boards

lp177:
Allow menu popup to overflow
https://github.com/nextcloud/deck/issues/2104


lp177:
We can optionnaly for backward compatibility of this choice add an scrollable parameter at true at line 91 and set this two css attributes if true at line 187

nickvergessen:
I think this was fixed with https://github.com/nextcloud/nextcloud-vue/pull/1160 already?

lp177:
Yes that seem to fix also, thanks

ma12-co:
So close this?

PR #1188, opened 2020-07-07
Add EmojiPicker Component

georgehrke:
| Light | Dark |
|:---------:|:------:|
|31685602-7C98-422A-8942-9EF23C074278|C456277E-7C72-464E-A702-BF07B5B33015|
|3BAEEB8C-973E-4011-AC72-E708CF0DDBEB|17E529A6-A297-4B81-B26F-71D890C3A829|


georgehrke:
cc @simonspa

georgehrke:
@simonspa @skjnldsv @nickvergessen @ma12-co I guess we want to put the EmojiPicker into a Popover component by default, right? (Similar to ColorPicker)

nickvergessen:
Yeah i guess popover is better.

Does this require things from -vue, or could it even be independent? I guess it will with the popover, right?

georgehrke:
The `emoji-mart-vue-fast` package itself is already a vue component.

simonspa:
What do you mean by  "popover"? Like the full color picker, opening its own window? I'd rather directly put it into a menu popping up directly from the button (like multiple actions) - not sure how you call this.

The popover option requires lots of mouse movement and the layout of the picker nicely fit in this popup menu already - maybe with a few less columns of emojis.

ma12-co:
> 'd rather directly put it into a menu popping up directly from the button (like multiple actions) 

@simonspa this is a popover :) 

Yes @georgehrke, go ahead with that. I cannot think of a different use-case that doesn't require the popover. If we ever find it, we can easily duplicate the template without wrapping it into the popover and provide a prop to disable it, right?

georgehrke:
Preview of documentation: https://5f0577f8c564a300078aa9a5--nextcloud-vue-components.netlify.app/#/Components/EmojiPicker

ChristophWurst:
> Preview of documentation

Two remarks

1) There is a bit of flickering going on when the popover is on the edge between displayed above the button or below, esp. when I scroll a bit as it's expanded
2) The triangle and the popover are separated by a little line
Bildschirmfoto von 2020-07-08 09-56-43

simonspa:
@ma12-co ah, thanks. :)

@georgehrke what I notice from the screenshots:

* somehow the popover is not centered over the triangle but slightly shifted to the left
* I would not display the "pick your emoji" message with large preview, I feel it's just a waste of space and looks odd (essentially preferring your first version)
* Somehow now we have a "frequently used" section plus a category in the header, I think one of them is enough
* how would it look like if we would reduce the number of emojis per line a bit?

simonspa:
Sorry, read the documentation now and see that the "preview" is optional. Also, "Frequently used" is just the scroll-through category displayed at the top... :)

src/components/EmojiPicker/EmojiPicker.vue
```diff
@@ -0,0 +1,236 @@
...
+		native() {
+			return this.activeSet === 'native'
+		},
+	},
+	methods: {
+		select(emojiObject) {
+			/**
+			 * Emits a string containing the emoji e.g. '👩🏿‍💻'
+			 */
+			this.$emit('select', emojiObject.native)
```
skjnldsv, line 181: Maybe also `update:select`? :)
To use with sync modifier?

src/components/EmojiPicker/EmojiPicker.vue
```diff
@@ -0,0 +1,236 @@
...
+				<button> Click Me </button>
+			</EmojiPicker>
+		</div>
+	</template>
+	<script>
+		export default {
+			methods: {
+				log(e) {
+					console.debug(e)
+				},
```
skjnldsv, line 45: Maybe the example could be easier than outputting to log?
Maybe adding a line with `selected emoji: 🦀` or something?

src/components/EmojiPicker/EmojiPicker.vue
```diff
@@ -0,0 +1,236 @@
...
```
skjnldsv, line 69: Same as other log entry

src/components/EmojiPicker/EmojiPicker.vue
georgehrke: I'm not entirely sure I understand how a sync modifier is helpful here, since there is no property which is being updated?

There is no selected emoji property you are providing to this component.

src/components/EmojiPicker/EmojiPicker.vue
```diff
@@ -0,0 +1,236 @@
...
+			type: Boolean,
+			default: true,
+		},
+	},
+	data() {
+		return {
+			emojiIndex: new EmojiIndex(data),
+			i18n: {
+				search: t('Search'),
+				notfound: t('No Emoji Found'),
```
nickvergessen, line 152:
```suggestion
				notfound: t('No emoji found'),
```

src/components/EmojiPicker/EmojiPicker.vue
```diff
@@ -0,0 +1,236 @@
...
+				categories: {
+					search: t('Search Results'),
```
nickvergessen, line 154:
```suggestion
					search: t('Search results'),
```

src/components/EmojiPicker/EmojiPicker.vue
```diff
@@ -0,0 +1,236 @@
...
+					recent: t('Frequently Used'),
```
nickvergessen, line 155:
```suggestion
					recent: t('Frequently used'),
```

src/components/EmojiPicker/EmojiPicker.vue
```diff
@@ -0,0 +1,236 @@
...
+					smileys: t('Smileys & Emotion'),
```
nickvergessen, line 156:
```suggestion
					smileys: t('Smileys & emotion'),
```

src/components/EmojiPicker/EmojiPicker.vue
```diff
@@ -0,0 +1,236 @@
...
+					people: t('People & Body'),
```
nickvergessen, line 157: and so on

src/components/EmojiPicker/EmojiPicker.vue
georgehrke: I agree on `Search emoji` and `Frequently used`, but `Smileys & Emotion`, `People & Body`, etc should be Title Case.

nickvergessen: Fine by me as well

src/components/EmojiPicker/EmojiPicker.vue
skjnldsv: Ah, sorry, I meant `update:emoji`
I misunderstood its use I guess :)

georgehrke:
> how would it look like if we would reduce the number of emojis per line a bit?

I checked how other emoji pickers are handling it and reduced it to 8, which is commonly used.

georgehrke:
I updated the screenshots on top.

Latest doc build: https://deploy-preview-1188--nextcloud-vue-components.netlify.app/#/Components/EmojiPicker

georgehrke:
@skjnldsv I'm not able to reproduce this:

> Changing tabs doesn't coincide with the actual scrolling

Kapture 2020-07-13 at 13 33 33

What browser / OS combination did you try?

georgehrke:
@skjnldsv That seems to be an upstream issue in https://serebrov.github.io/emoji-mart-vue/ :/ 
Would you be fine merging this for now and i will report it upstream?

skjnldsv:
> Would you be fine merging this for now and i will report it upstream?

Sure!

skjnldsv:
But please make sure @jancborchardt is ok with this before releasing!

georgehrke:
Upstream issue: https://github.com/serebrov/emoji-mart-vue/issues/77

skjnldsv:
@georgehrke will you do us the honour of pressing the merge button? :D

PR #1202, opened 2020-07-16
vertically align active checkmark to center

trev-dev:
I spotted an issue downstream that I can replicate with firefox, which shows an active checkbox being at the top of the color picker, rather than the vertical center: https://github.com/nextcloud/deck/issues/2102

I'm not sure if `align-content` was a typo or not, so I left it there.

Using `align-items: center;` will vertically align children in a flex container to the middle of said container when combined with `justify-content: center;`

### DIFF
--- a/src/components/ColorPicker/ColorPicker.vue
+++ b/src/components/ColorPicker/ColorPicker.vue
@@ -251,6 +251,7 @@ export default {
 		&-color-circle {
 			display: flex;
 			align-content: center;
+			align-items: center;
 			justify-content: center;
 			width: 34px;
 			height: 34px;
--- a/src/components/ColorPicker/ColorPicker.vue
+++ b/src/components/ColorPicker/ColorPicker.vue
@@ -250,7 +250,6 @@ export default {
 
 		&-color-circle {
 			display: flex;
-			align-content: center;
 			align-items: center;
 			justify-content: center;
 			width: 34px;


src/components/ColorPicker/ColorPicker.vue
```diff
@@ -251,6 +251,7 @@ export default {
 		&-color-circle {
 			display: flex;
 			align-content: center;
+			align-items: center;
```
juliushaertl, line 254: align-content doesn't have any effect here since it is a single line flex layout with just one element, so it would be good to remove that right away imo :+1:

juliushaertl:
Let's remove the unneeded align-content, works like a charm otherwise :+1: Thanks a lot

trev-dev:
> Let's remove the unneeded align-content, works like a charm otherwise +1 Thanks a lot

Done and dusted.

src/components/ColorPicker/ColorPicker.vue
trev-dev: > align-content doesn't have any effect here since it is a single line flex layout with just one element, so it would be good to remove that right away imo +1

https://github.com/nextcloud/nextcloud-vue/pull/1202/commits/1c51bab73d40e285e47e146d0d1eed591d578af8 Resolved :+1:

trev-dev:
I've submitted the requested changes, please let me know if there's anything more I can do to help.

PR #1250, opened 2020-08-02
add multiselect type to ActionInput

### LINKED ISSUE #774: ActionInput: add Multisect
It would be nice if we could provide a drop-down list for the `ActionInput` component (see https://github.com/nextcloud/notes/issues/413#issuecomment-565825366). Therefore, I propose to add the existing `Multiselect` component to `ActionInput`. It could be used instead of the default HTML input field if the new optional prop `options` is set. All other props would be passed, so further customization would be possible.

Any doubts on this? I would try to provide a PR otherwise.

korelstar:
Straight forward implementation: simply passes all attributes and events from/to multiselect component.

Closes #774

Preview of documentation: https://deploy-preview-1250--nextcloud-vue-components.netlify.app/#/Components/Actions?id=actioninput

### DIFF
--- a/src/components/ActionInput/ActionInput.vue
+++ b/src/components/ActionInput/ActionInput.vue
@@ -70,6 +70,7 @@ For the multiselect component, all events will be passed through. Please see the
 					:value="value"
 					:placeholder="text"
 					:disabled="disabled"
+					:class="{ focusable: isFocusable }"
 					class="action-input__multi"
 					v-bind="$attrs"
 					v-on="$listeners" />
--- a/src/components/ActionInput/ActionInput.vue
+++ b/src/components/ActionInput/ActionInput.vue
@@ -23,14 +23,15 @@
 
 <docs>
 This component is made to be used inside of the [Actions](#Actions) component slots.
-All undocumented attributes will be bound to the input or the datepicker. e.g. `maxlength`, `not-before`...
-
+All undocumented attributes will be bound to the input, the datepicker or the multiselect component, e.g. `maxlength`, `not-before`.
+For the multiselect component, all events will be passed through. Please see the multiselect component's documentation for more details and examples.
 ```
 <Actions>
 	<ActionInput icon="icon-edit" value="This is an input" />
 	<ActionInput icon="icon-edit">This is the placeholder</ActionInput>
 	<ActionInput icon="icon-close" :disabled="true" value="This is a disabled input" />
 	<ActionInput icon="icon-edit" type="date">Please pick a date</ActionInput>
+	<ActionInput icon="icon-edit" type="multiselect" :options="['Apple', 'Banana', 'Cherry']">Please pick a fruit</ActionInput>
 </Actions>
 ```
 </docs>
@@ -65,6 +66,14 @@ All undocumented attributes will be bound to the input or the datepicker. e.g. `
 					@input="onInput"
 					@change="onChange" />
 
+				<Multiselect v-else-if="isMultiselectType"
+					:value="value"
+					:placeholder="text"
+					:disabled="disabled"
+					class="action-input__multi"
+					v-bind="$attrs"
+					v-on="$listeners" />
+
 				<template v-else>
 					<input :id="id" type="submit" class="action-input__submit">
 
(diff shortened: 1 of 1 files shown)


skjnldsv:
Other than this! Great addition!! :+1:

skjnldsv:
> You need to implement the `focusable` class trick so that accessibility works :)

Ah, it doesn't seems that you can :/
https://github.com/shentao/vue-multiselect/blob/d8d29e7714449f4fca4bfe10d3bd21e72c46c9b7/src/Multiselect.vue#L47

korelstar:
Added the `focusable` class and focusing the `ActionInput` element using cursor keys works now.

src/components/ActionInput/ActionInput.vue
```diff
@@ -68,6 +68,7 @@ For the multiselect component, all events will be passed through. Please see the
 					:value="value"
 					:placeholder="text"
 					:disabled="disabled"
+					:class="{ focusable: isFocusable }"
```
skjnldsv, line 71: It won't work, it needs to be on the input so that the browser canfocus the proper element :/

korelstar: Did you test it? For me it works (Firefox). I think it works due to https://github.com/shentao/vue-multiselect/blob/d8d29e7714449f4fca4bfe10d3bd21e72c46c9b7/src/Multiselect.vue#L61

korelstar: sorry, I meant due to https://github.com/shentao/vue-multiselect/blob/d8d29e7714449f4fca4bfe10d3bd21e72c46c9b7/src/Multiselect.vue#L5

skjnldsv: You cannot select anything inside.
It properly opens it, but it's not fully working :)

But I don't think it's an issue with the component itself in the end, but more like a keydown event conflict. Once it's focused, we should prevent anything from opening.
The datepicker should not open on focus. Maybe just let the user enter a date or something. :shrug: 
And the multiselect is a tricky one too. Maybe have the search directly set to true so that keyboard users can directly find what they're looking for?
I don't know :shrug:

korelstar:
@skjnldsv what can we do in order to get this in? Honestly, I don't know how to fix the issue you have raised. Moreover, the issue existed already before for the datepicker. So, what should we do here?

skjnldsv:
> Honestly, I don't know how to fix the issue you have raised. Moreover, the issue existed already before for the datepicker. So, what should we do here?

https://github.com/nextcloud/nextcloud-vue/pull/1250#discussion_r464071712
Honestly, no idea how to proceed.
But I'd rather not implement something that doesn't work for keyboard/voice users.
I think it's best not to have it rather than a half-working feature :p

Jan mentioned adding a menu navigation feature. Where you click an ActionButton and it shift the whole menu on the left to make another set of entries open (instead of having a dropdown inside a dropdown).
Would this fix what you were trying to implement? :)

korelstar:
@skjnldsv I need to allow for input new values and comfortably select from existing values at the same time. If I understand Jan's idea correctly, that approach would allow for selecting existing options only. However, the first sub-action could be a plain input. This could work. But I'm unsure if this would be a good UX.

skjnldsv (approved):

jancborchardt (changes requested):
Looks good, but same as with the datepicker it’s strange that a hover triggers expanding the multiselect. I’d say only a focus event / clicking in the input should?

nickvergessen:
> Looks good, but same as with the datepicker it’s strange that a hover triggers expanding the multiselect. I’d say only a focus event / clicking in the input should?

Was about to post the same thing.

nickvergessen (approved):

skjnldsv:
> I’d say only a focus event / clicking in the input should?

It's due to the way we manage focus feedback.
When you hover an entry it focuses it to keep the currently focused index of the menu list. Otherwise you could have one entry focused with keyboard and another hovered with your mouse

skjnldsv:
Ping @jancborchardt :wink: 
https://github.com/nextcloud/nextcloud-vue/pull/1250#issuecomment-704721829

nickvergessen:
This is currently blocking the announcement center rewrite to Vue as one of the last things:
https://github.com/nextcloud/announcementcenter/pull/218

What's the status here? Can we merge it for now and fix the comment afterwards?

jancborchardt:
Hm, your call @skjnldsv on whether to do it in this PR or separately, but I still think that we should absolutely trigger neither datepicker nor multiselect on hover, we have to separate hover and focus there.

skjnldsv:
> we have to separate hover and focus there.

I'm fine with that, but then you'll lose the persistance between the two. Hovering should kill the focus then.

nickvergessen:
@korelstar can you adjust this?

korelstar:
@nickvergessen Sorry I'm a little bit lost, here. I think the criticised behavior is out of scope from this PR. This PR introduces MultiSelect while adopting the behavior of DatePicker. From my understanding, DatePicker's behavior has to be fixed first (in another PR) and then I will adopt it in this PR (if it is not automatically done).

PR #1259, opened 2020-08-05
Add user-status to Avatar component

georgehrke:
- [x] Icons need attribution (Material UI)
- [x] add white border around icons

| | |
|:---------:|:------:|
|61D15EEB-050C-4EB4-B565-CAA7DBC3B8D6|170CA5A4-4621-4ED2-A833-027CFF03A2F8|
|76349057-CF32-405F-9DFA-D66C3197A138||

### DIFF
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -54,7 +54,10 @@
 		<div v-if="hasMenu" class="icon-more" />
 
 		<!-- avatar status -->
-		<div v-if="status"
+		<div v-if="canDisplayUserStatus"
+			class="avatardiv__user-status"
+			:class="'avatardiv__user-status--' + userStatus.status" />
+		<div v-else-if="status"
 			class="avatardiv__status"
 			:class="'avatardiv__status--' + status"
 			:style="{ backgroundColor: `#${statusColor}` }">
@@ -91,6 +94,7 @@ import axios from '@nextcloud/axios'
 import { generateUrl } from '@nextcloud/router'
 import Tooltip from '../../directives/Tooltip'
 import usernameToColor from '../../functions/usernameToColor'
+import { userStatus } from '../../mixins'
 
 export default {
 	name: 'Avatar',
@@ -101,6 +105,7 @@ export default {
 	components: {
 		PopoverMenu,
 	},
+	mixins: [userStatus],
 	props: {
 		/**
 		 * Set a custom url to the avatar image
@@ -124,6 +129,13 @@ export default {
 		user: {
 			type: String,
 			default: undefined,
+		},
+		/**
+		 * Whether or not to display the user-status
+		 */
+		showUserStatus: {
+			type: Boolean,
+			default: true,
 		},
 		/**
 		 * Is the user a guest user (then we have to user a different endpoint)
@@ -191,6 +203,9 @@ export default {
 		},
 
 		/**
+		 * DEPRECATED!
+		 * This prop will be removed with nc/vue 3.0
+		 *
 		 * Declares a status indicator on the avatar
 		 * Available options are `positive`, `negative`, `neutral`
 		 */
@@ -238,6 +253,9 @@ export default {
 		}
 	},
 	computed: {
+		canDisplayUserStatus() {
+			return this.showUserStatus && this.hasStatus
+		},
 		getUserIdentifier() {
 			if (this.isDisplayNameDefined) {
 				return this.displayName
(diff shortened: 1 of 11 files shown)


src/components/Avatar/Avatar.vue
```diff
@@ -51,7 +51,10 @@
 		<div v-if="hasMenu" class="icon-more" />
 
 		<!-- avatar status -->
-		<div v-if="status"
+		<div v-if="userStatus.status"
```
georgehrke, line 54: canDisplayUserStatus

georgehrke:
There are two things which i would like to add later in a follow-up PR:
- [ ] Periodically refresh the status
- [ ] Have some kind of central store for statuses, so we don't fetch it dozens of times for the same user. This is especially important for Talk, where there are many Avatars of the same user.

src/components/Avatar/Avatar.vue
```diff
@@ -54,7 +54,10 @@
...
 
 		<!-- avatar status -->
-		<div v-if="status"
+		<div v-if="canDisplayUserStatus"
+			class="avatardiv__user-status"
+			:class="'avatardiv__user-status--' + userStatus.status" />
+		<div v-else-if="status"
 			class="avatardiv__status"
 			:class="'avatardiv__status--' + status"
 			:style="{ backgroundColor: `#${statusColor}` }">
```
skjnldsv, line 63: Is it still worth having the old status?
I'd rather have this as the new design you implemented :)

georgehrke: Right now they exist along, but the user-status takes priority.
We should mark it as deprecated and remove it with nc/vue 3.0 i would say.

Does that sound alright?

skjnldsv: Sure thing! :)

georgehrke: Done

ChristophWurst:
What is the expected handling of Nextcloud 19 and older? Send off the request anyway and catch the missing route error?

ChristophWurst:
> Periodically refresh the status

https://github.com/nextcloud/nextcloud-push if we have that one day

src/components/Avatar/Avatar.vue
```diff
@@ -346,6 +385,26 @@ export default {
...
 			this.isMenuLoaded = true
 		},
+		async fetchUserStatus() {
+			if (!this.user) {
+				return
+			}
+			const user = encodeURIComponent(this.user)
+
+			try {
+				const { data } = await axios.get(generateOcsUrl('apps/user_status/api/v1', 2) + `statuses/${user}`)
```
juliushaertl, line 395: I think this should be cached ideally in a way that multiple component instances can access the status. I assume this could perform very bad if an avatar is rendered multiple times e.g. in a talk conversation.

juliushaertl: Ah missed your todo list for later :D

georgehrke:
> What is the expected handling of Nextcloud 19 and older?
> Send off the request anyway and catch the missing route error?

We could check the ocs capabilities API, but that adds yet another request.

What is the best way to store the status among multiple instances of Avatar?
I would avoid adding vuex just for this.

georgehrke:
I moved the status fetching to a mixin, so we can also use it in the `AvatarSelectOption.vue ` component.
(see screenshot in original post)

georgehrke:
Squashed and rebased

nickvergessen:
Pushed a PR to allow to control the status displaying on the mention chip

src/components/UserBubble/UserBubble.vue
```diff
@@ -90,6 +90,7 @@ This component has the following slot:
 				:style="styles.avatar"
 				:disable-tooltip="true"
 				:disable-menu="true"
+				:show-user-status="showUserStatusOnAvatar"
```
skjnldsv, line 93: should maybe just `v-bind="$props"` so everything is forwarded to the Avatar?

nickvergessen: Fine by me, if we have no conflicting things?

src/mixins/userStatus.js
```diff
@@ -0,0 +1,60 @@
...
+					status,
+					message,
+					icon,
+				} = data.ocs.data
+				this.userStatus.status = status
+				this.userStatus.message = message || ''
+				this.userStatus.icon = icon || ''
+				this.hasStatus = true
+			} catch (e) {
+				console.error(e)
```
nickvergessen, line 56: Don't log 404 when it was "ok"
```
{"ocs":{"meta":{"status":"failure","statuscode":404,"message":"No status for the requested userId"},"data":[]}}
```

ChristophWurst:
> We could check the ocs capabilities API, but that adds yet another request.

`OC.getCapabilities` or similar?

georgehrke:
> > We could check the ocs capabilities API, but that adds yet another request.
> 
> `OC.getCapabilities` or similar?

Forgot we have that. Thx!
I added it to the PR.

juliushaertl:
There is also https://github.com/nextcloud/nextcloud-capabilities/

rullzer:
sooo merge?

skjnldsv:
> sooo merge?

https://github.com/nextcloud/nextcloud-vue/pull/1259#pullrequestreview-468530305

georgehrke:
@rullzer Fixed, please hit merge 😉

Issue #1090, opened 2020-05-11
Improvements to AppSidebar component


raimund-schluessler:
I currently move the right details view of the Tasks app to the AppSidebar component of nextcloud-vue, see https://github.com/nextcloud/tasks/pull/1030. I noticed a few things with `AppSidebar` which would limit the functionality of the Tasks app. This issue is to keep track of what needs to be improved on the `AppSidebar` component to make it usable for Tasks. I will likely add more items to the list, since I am far from done with https://github.com/nextcloud/tasks/pull/1030.

- ~Allow to specify a custom icon for the starred button
The Tasks app has multiple levels of priority for a task, low, medium and high. Each level is indicated by a different star color. Hence, the star icon needs to be adjustable (color would be enough, but I guess it is easier to just provide an icon).~ We will not show the star icon for the priority in the header anymore. The completed checkbox is colored in red, yellow or blue to show the priority in the header, priority can be set in the details.
![Screenshot_2020-05-11 Tasks - Nextcloud(3)](https://user-images.githubusercontent.com/2496460/81611906-333dd580-93dc-11ea-9a23-fd399daf8db2.png)
- ~Make the star icon bigger
This is a matter of taste, but I think it is a bit to small (see image above for a comparison of nextcloud-vue and current Tasks star size).~
- [x] The Tasks app needs the option to show a checkbox for toggling the completed state of a task. The checkbox could be shown at the place of the star icon. (The star icon for a task will not be shown anymore, since it was doubled anyway.) #1310 
- [x] Don't always show the input field for the editable sidebar title. #1288
The title should be rendered as `h2` and the input field should be visible after clicking the title. This can be optional, but it is a requirement for the next point. Also, I think it looks weird to permanently show the input field, since the title is then quite small and looks very different to the non-editable title.
Compare non-editable
![Screenshot_2020-05-11 Tasks - Nextcloud](https://user-images.githubusercontent.com/2496460/81611720-d80be300-93db-11ea-9667-0dd93c75392b.png)
and editable
![Screenshot_2020-05-11 Tasks - Nextcloud(1)](https://user-images.githubusercontent.com/2496460/81611738-de01c400-93db-11ea-806b-99785205a88a.png)
- [x] Linkify links in title string #1289
In Tasks we linkify links in the task summary, so you can navigate there without copy/paste. The same should be (optionally) done for the title in `AppSidebar`. This also means the editable title cannot be shown in an input field permanently.
[...]

PR #1261, opened 2020-08-06
Allow to show an empty sidebar

raimund-schluessler:
Allows to show a blank sidebar without header and tab navigation bar. Useful for showing a hint when loading data from the server (please ignore the unfinished sidebar implementation of the Tasks app in the GIF).

appsidebar

Start of implementing #1090.

### DIFF
--- a/src/components/AppSidebar/AppSidebar.vue
+++ b/src/components/AppSidebar/AppSidebar.vue
@@ -82,13 +82,37 @@
 		}
 	</script>
 	```
+	```
+
+	### Blank sidebar for e.g. empty content
+	```vue
+	<template>
+		<AppSidebar
+			:title="title"
+			blank="true">
+			<EmptyContent icon="icon-search">
+				Content not found.
+			</EmptyContent>
+		</AppSidebar>
+	</template>
+	<script>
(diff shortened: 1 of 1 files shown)


src/components/AppSidebar/AppSidebar.vue
```diff
@@ -82,13 +82,37 @@
...
+		</AppSidebar>
+	</template>
+	<script>
+		export default {
+			data() {
+				return {
+					title: 'cat-picture.jpg'
+				}
+			}
+		}
```
skjnldsv, line 105: We could just put this as an inline string for the example :)

raimund-schluessler: Fixed.

raimund-schluessler:
While working on #1262 I came to the conclusion that I will rework the AppSidebar component to use a render function (the `this.$children` array behaves inconsistent between browser and jest test environment, so it's probably better to not rely on it). Also, I realized that the close button and the navtabs (in case of more than one tab) need to be visible for an empty/blank sidebar as well. I hence close this PR and implement the blank option with the render function.

PR #1279, opened 2020-08-12
fix: toggle-navigation, open: false

azul:
Make sure this keeps the nav closed
if it already was closed.

Includes tests.

Signed-off-by: Azul <azul@riseup.net>

### DIFF
--- a/tests/unit/components/AppNavigation/AppNavigation.spec.js
+++ b/tests/unit/components/AppNavigation/AppNavigation.spec.js
@@ -1,0 +1,51 @@
+/**
+ * @copyright Copyright (c) 2020 Azul <azul@riseup.net>
+ *
+ * @author Azul <azul@riseup.net>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


tests/unit/components/AppNavigation/AppNavigation.spec.js
```diff
@@ -0,0 +1,57 @@
...
+			const wrapper = mount(AppNavigation)
+			emit('toggle-navigation', {open: undefined})
+			expect(wrapper.vm.$data.open).toBe(false)
+		})
+
+		xit('toggle without params', () => {
+			const wrapper = mount(AppNavigation)
+			emit('toggle-navigation')
+			expect(wrapper.vm.$data.open).toBe(false)
+		})
```
azul, line 47: Not sure if we want to support toggling the nav without any args.

Currently it is not working. I tried to implement it by adding a default to the bus handler but that did not help.

ma12-co: I think we don't want to do that. Do you have a use-case in mind in which that would be beneficial?

azul: I don't know... I just noticed that in the toggle function toggling as in
```
open = !open
```
was implemented and i kept that.

```
emit('toggle-navigation', {open: undefined})
```
seems like an odd syntax for that to me.

If we don't want to support it at all then i would remove the check for undefined and treat `open: undefined` just like
```
emit('toggle-navigation', {open: false})
```

azul: Okay... I dropped the test for the syntax without `open` and kept the old behavior for `open: undefined` because there may be apps that rely on it.
Docs say that `open: false` will close. That's what this pr achieves.

PR #1281, opened 2020-08-12
Allow empty sidebar, correctly check mixed content

### LINKED ISSUE #733: "Mixing tabs and non-tab-content is not possible" when providing slots
In Talk we provide tabs for the sidebar and also provide a template for the `secondary-actions` slot, but this throws a red warning in the console:

```
 [Vue warn]: Mixing tabs and non-tab-content is not possible. v-tooltip.esm.js:3980
    warn v-tooltip.esm.js:3980
    updateTabs AppSidebar.js:12
    mounted AppNavigationCounter.js:1115
    VueJS 13
    js talk.js:104855
    Webpack 3
```

Removing the slot works:
https://github.com/nextcloud/spreed/blob/f9881d670a28832d0f4da541ed91d26230453c07/src/components/Sidebar/Sidebar.vue#L29-L47

But I guess it should just not be that strict on checking and basically allow templates here that are for slots:
https://github.com/nextcloud/nextcloud-vue/blob/master/src/utils/ValidateChildren.js#L39

raimund-schluessler:
This PR allows to show an empty sidebar and assures to correctly check for mixing AppSidebarTab with other content in the default slot. The PR partially implements #1090 and closes #733.

Here is how it looks.
After:
after

Before:
before

I decided against converting the tabs part to a render function, because it seemed quite a lot of effort for just "doing it right". The template solution seems to work fine. The only thing which seems a bit dubious to me is watching `this.$children` for changes, but it was like that before and works, so 🤷‍♂️

### DIFF
--- a/tests/unit/components/AppSidebar/AppSidebarTabs.spec.js
+++ b/tests/unit/components/AppSidebar/AppSidebarTabs.spec.js
@@ -1,0 +1,111 @@
+/**
+ * @copyright Copyright (c) 2020 Raimund Schlüßler <raimund.schluessler@mailbox.org>
+ *
+ * @author Raimund Schlüßler <raimund.schluessler@mailbox.org>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+import { mount } from '@vue/test-utils'
+
+import Vue from 'vue'
+import AppSidebarTabs from '../../../../src/components/AppSidebarTabs/AppSidebarTabs.vue'
+import AppSidebarTab from '../../../../src/components/AppSidebarTab/AppSidebarTab.vue'
+import ActionButton from '../../../../src/components/ActionButton/ActionButton.vue'
+
+describe('AppSidebarTabs.vue', () => {
+	'use strict'
+
+	it('Issues no warning when using it without tabs.', () => {
+		const onWarning = jest.fn()
+		Vue.config.warnHandler = onWarning
+		const wrapper = mount(AppSidebarTabs, {
+			propsData: {
+				title: 'Sidebar title.'
+			},
+			slots: {
+				default: ['<div />'],
+			}
+
+		})
+		expect(onWarning).toHaveBeenCalledTimes(0)
+	})
+
+	it('Issues no warning when using secondary actions.', () => {
+		const onWarning = jest.fn()
+		Vue.config.warnHandler = onWarning
+		const wrapper = mount(AppSidebarTabs, {
+			propsData: {
+				title: 'Sidebar title.'
+			},
+			slots: {
+				default: '<div />',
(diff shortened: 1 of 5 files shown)


src/components/AppSidebarTabs/index.js
```diff
@@ -0,0 +1,23 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+import AppSidebarTabs from './AppSidebarTabs'
+
+export default AppSidebarTabs
```
skjnldsv, line 23: I'm wondering if this shouldn't just be next to the AppSidebar.vue as thi sis a non-exported component?

tests/unit/components/AppSidebar/AppSidebarTabs.spec.js
```diff
@@ -0,0 +1,111 @@
...
+ */
+
+import { mount } from '@vue/test-utils'
+
+import Vue from 'vue'
+import AppSidebarTabs from '../../../../src/components/AppSidebarTabs/AppSidebarTabs.vue'
+import AppSidebarTab from '../../../../src/components/AppSidebarTab/AppSidebarTab.vue'
+import ActionButton from '../../../../src/components/ActionButton/ActionButton.vue'
+
+describe('AppSidebarTabs.vue', () => {
```
skjnldsv, line 30: Oh wow, with tests?? :heart_eyes:

raimund-schluessler: Yeah, some tests I wrote, but mainly to check for #733. There could be much more tests 😉

src/components/AppSidebarTabs/index.js
raimund-schluessler: You mean to not export `AppSidebarTabs`, but directly import it in `AppSidebar`? Probably makes sense. I will adjust it in the evening.

In the future, we could in principle also export `AppSidebarTabs` (maybe export it as `Tabs`). The component should work without `AppSidebar` as well.

raimund-schluessler:
A small comment just to make you aware (in case you aren't). The structure of the `AppSidebar` component has slightly changed from before:

```
<aside class="app-sidebar">
    <header class="app-sidebar-header">
        Header
    </header>
    <nav class="app-sidebar-tabs__nav">
        Tabs Navbar
    <nav>
    <div class="app-sidebar-tabs__content>
        Content
    </div>
</aside>
```

to:
```
<aside class="app-sidebar">
    <header class="app-sidebar-header">
        Header
    </header>
    <div class="app-sidebar-tabs>
        <nav class="app-sidebar-tabs__nav">
            Tabs Navbar
        <nav>
        <div class="app-sidebar-tabs__content>
            Content
        </div>
    </div>
</aside>
```
Notice the additional `div` around the tabs stuff. This is due to ```AppSidebarTabs``` being a separate component now. As long as apps have no own CSS adjusting the AppSidebar style with to strict rules (something like `.app-sidebar > nav` or similar, everything should work. But since apps shouldn't do this anyway, I think it does not classify as a breaking change.

tests/unit/components/AppSidebar/AppSidebarTabs.spec.js
skjnldsv: Yes, unit tests are not my strength to be honest, I should really take the time to dive into this

skjnldsv:
> Notice the additional `div` around the tabs stuff. This is due to `AppSidebarTabs` being a separate component now. As long as apps have no own CSS adjusting the AppSidebar style with to strict rules (something like `.app-sidebar > nav` or similar, everything should work. But since apps shouldn't do this anyway, I think it does not classify as a breaking change.

I don't think that is much of an issue! :+1: 
Nice work!! :hugs:

src/components/AppSidebarTabs/index.js
raimund-schluessler: Done.

PR #1283, opened 2020-08-13
Actions: Introduce icon slots

### LINKED ISSUE #1270: ActionButton: Add icon slot so material ui icons can be used

marcelklehr:
fixes #1270

### DIFF
--- a/src/components/ActionButton/ActionButton.vue
+++ b/src/components/ActionButton/ActionButton.vue
@@ -49,10 +49,12 @@ If you're using a long text you can specify a title
 			:aria-label="ariaLabel"
 			:class="{ focusable: isFocusable }"
 			@click="onClick">
-			<!-- icon -->
-			<span :class="[isIconUrl ? 'action-button__icon--url' : icon]"
-				:style="{ backgroundImage: isIconUrl ? `url(${icon})` : null }"
-				class="action-button__icon" />
+			<!-- @slot Manually provide icon -->
+			<slot name="icon">
+				<span :class="[isIconUrl ? 'action-button__icon--url' : icon]"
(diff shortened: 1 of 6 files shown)


skjnldsv:
I don't think that is the approach we should take.
We should just make an Icon component that can take the aliases we talked about: `material:magnify`

And not allow an entire slot i'd say.

marcelklehr:
> And not allow an entire slot i'd say.

There's already a slot for AppNavigationItems, so I thought this would be consistent, until we figure out a clean way to go forward with icons/logos. Not really particular about this, though. I'm happy to wait until we have good solution.

juliushaertl:
> I don't think that is the approach we should take.
> We should just make an Icon component that can take the aliases we talked about: material:magnify

The issue i see with this is how we would then take care of not bundling all of the icons in the `@nextcloud/vue` library. A slot seems to be the cleanest solution to me where the tree shaking of the icons would be handled by the apps and reduce the implementation effort from the library perspective.

skjnldsv:
> The issue i see with this is how we would then take care of not bundling all of the icons in the `@nextcloud/vue` library. A slot seems to be the cleanest solution to me where the tree shaking of the icons would be handled by the apps and reduce the implementation effort from the library perspective.

Yeah, maybe here it would make sense then.
And only use the said Icon component for our php APIS. :thinking:

georgehrke:
@marcelklehr Is this ready for review? :)

georgehrke (approved):

skjnldsv (changes requested):
Please add documentation :)

skjnldsv (approved):

PR #1286, opened 2020-08-14
App Settings Dialog

ma12-co:
- [x] Proper styles
- [x] Write documentation

followup:
- [ ] Add slots for icons in the navigation
- [ ] Allow to open the setting with a specific anchor selected
- [ ] Compute whitespace at the end of the settings so that upon clicking on a navigation item, the first thing shown in the scroller is always the title of the clicked section  

Peek 2020-09-22 12-58

### DIFF
--- a/src/components/AppSettingsDialog/AppSettingsDialog.vue
+++ b/src/components/AppSettingsDialog/AppSettingsDialog.vue
@@ -1,0 +1,308 @@
+<!--
+ - @copyright Copyright (c) 2020 Marco Ambrosini <marcoambrosini@pm.me>
+ -
+ - @author Marco Ambrosini <marcoambrosini@pm.me>
+ -
+ - @license GNU AGPL version 3 or any later version
+ -
+ - This program is free software: you can redistribute it and/or modify
+ - it under the terms of the GNU Affero General Public License as
+ - published by the Free Software Foundation, either version 3 of the
+ - License, or (at your option) any later version.
+ -
+ - This program is distributed in the hope that it will be useful,
+ - but WITHOUT ANY WARRANTY; without even the implied warranty of
+ - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ - GNU Affero General Public License for more details.
+ -
+ - You should have received a copy of the GNU Affero General Public License
+ - along with this program. If not, see <http://www.gnu.org/licenses/>.
+ -
+ -->
+<docs>
+Just nest the `AppSettingSections` component into `AppSettingsDialog`,
+providing the section's title prop. You can put your settings within each
+`AppSettingsSection` component.
+
+```vue
+<template>
+	<div>
+		<button @click="settingsOpen = true">Show Settings</button>
+		<AppSettingsDialog :open.sync="settingsOpen" >
+			<AppSettingsSection title="Example title 1">
+				Some example content
+			</AppSettingsSection>
+			<AppSettingsSection title="Example title 2">
+				Some more content
+			</AppSettingsSection>
+		</AppSettingsDialog>
+	</div>
+</template>
+
+<script>
+export default {
+	data() {
+		return {
+			settingsOpen: false,
+		}
+	},
+}
+</script>
+```
+</docs>
+
+<script>
+import Modal from '../Modal'
+import isMobile from '../../mixins/isMobile'
+import debounce from 'debounce'
+
+export default {
+
+	name: 'AppSettingsDialog',
+
+	components: {
+		Modal,
+	},
+
+	mixins: [isMobile],
+
+	props: {
+		/**
+		 * Determines the open / closed state of the modal
+		 */
+		open: {
+			type: Boolean,
+			required: true,
+		},
+		/**
+		 * Hides the navigation on desktop if true
+		 */
+		hideNavigation: {
+			type: Boolean,
+			default: false,
+		},
+	},
+
+	data() {
+		return {
+			selectedSection: '',
+			linkClicked: false,
+			addedScrollListener: false,
+			scroller: null,
+		}
+	},
+
+	computed: {
+
+		hasNavigation() {
+			if (this.isMobile || this.hideNavigation) {
+				return false
+			} else {
+				return true
+			}
+		},
+
+		settingsNavigationAriaLabel() {
+			return t('Settings navigation')
+		},
+	},
+
+	mounted() {
+		// Select first settings section
+		this.selectedSection = this.$slots.default[0].componentOptions.propsData.title
+	},
+
+	updated() {
+		if (!this.scroller) {
+			// Get the scroller element
+			this.scroller = this.$el.querySelector('.app-settings__content')
+		}
+		if (this.scroller && !this.addedScrollListener) {
+			this.scroller.addEventListener('scroll', this.handleScroll)
+			this.addedScrollListener = true
+		}
+	},
+
+	beforeDestroy() {
+		this.scroller.removeEventListener('scroll', this.handleScroll)
+	},
+	methods: {
+
+		/**
+		 * Builds the settings navigation menu
+		 * @param {object} slots The default slots object passed from the render function.
+		 * @returns {array} the navigation items
+		 */
+		getSettingsNavigation(slots) {
+			// Array of navigationitems strings
+			const navigationItems = slots.filter(vNode => vNode.componentOptions).map(vNode => vNode.componentOptions.propsData?.title)
+			// Check for the uniqueness of section titles
+			navigationItems.forEach((element, index) => {
+				const newArray = [...navigationItems]
+				newArray.splice(index, 1)
+				if (newArray.indexOf(element) !== -1) {
+					throw new Error(`Duplicate section title found: ${element}. Settings navigation sections must have unique section titles.`)
+				}
+			})
+			return navigationItems
+		},
+
+		/**
+		 * Scrolls the content to the selected settings section.absolute
+		 * @param {string} item the name of the section
+		 */
+		handleSettingsNavigationClick(item) {
+			this.linkClicked = true
+			document.getElementById('settings-section_' + item.replace(/\s+/g, '')).scrollIntoView({
+				behavior: 'smooth',
+				inline: 'nearest',
+			})
+			this.selectedSection = item
+			setTimeout(() => {
+				this.linkClicked = false
+			}, 1000)
+		},
+
+		handleCloseModal() {
+			this.$emit('update:open', false)
+		},
+
+		// Remove selected section once the user starts scrolling
+		handleScroll() {
+			if (!this.linkClicked) {
+				debounce(() => {
+					this.selectedSection = ''
+					document.activeElement.blur()
+				}, 300)
+			}
+		},
+
+		handleLinkKeydown(keyDownEvent, item) {
+			if (keyDownEvent.code === 'Enter') {
+				this.handleSettingsNavigationClick(item)
+			}
+		},
+	},
+
+	render(createElement) {
+		/**
+		 * Build the navigation
+		 * @returns {object} the navigation
+		 */
+		const createAppSettingsNavigation = () => {
+			if (this.hasNavigation) {
+				return [createElement('div', {
+					attrs: {
+						class: 'app-settings__navigation',
+						role: 'tablist',
+						'aria-label': this.settingsNavigationAriaLabel,
+					},
+				}, [createElement('ul', {
+					attrs: {
+						class: 'navigation-list',
+						role: 'tablist',
+					},
+				}, this.getSettingsNavigation(this.$slots.default).map(item => {
+					return createListElemtent(item)
+				}))])]
+			} else {
+				return []
+			}
+		}
+
+		/**
+		 * Build each list element in the navigation
+		 * @param {object} item the navigation item
+		 * @returns {object} the list element
+		 */
+		const createListElemtent = (item) => createElement('li', {}, [createElement('a', {
+			class: {
+				'navigation-list__link': true,
+				'navigation-list__link--active': item === this.selectedSection,
+			},
+
+			attrs: {
+				role: 'tab',
+				'aria-selected': item === this.selectedSection,
+				tabindex: '0',
+			},
+
+			on: {
+				click: () => this.handleSettingsNavigationClick(item),
+				keydown: () => this.handleLinkKeydown(event, item),
+			},
+		}, item)])
+
+		// Return value of the render function
+		if (this.open) {
+			return createElement('Modal', {
+				on: {
+					close: () => { this.handleCloseModal() },
+				},
+			}, [
+				createElement('div', {
+					attrs: {
+						class: 'app-settings',
+					},
+				}, [...createAppSettingsNavigation(),
+					createElement('div', {
+						attrs: {
+							class: 'app-settings__content',
+						},
+					}, this.$slots.default)]),
+			])
+		} else {
+			return undefined
+		}
+	},
+}
+
+</script>
+
+<style lang="scss" scoped>
+
+.app-settings {
+	padding: 28px 16px;
+	display: flex;
+	height: 600px;
+	width: 100%;
+	&__navigation {
+		width: 200px;
+		margin-right: 20px;
+	}
+	&__content {
+		width: 350px;
+		max-width: 100vw;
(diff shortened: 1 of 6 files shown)


src/components/AppSettingsDialog/AppSettingsDialog.vue
```diff
@@ -0,0 +1,244 @@
...
+		toggleSettings(bool) {
+			this.open = bool
+			const bodyStyles = getComputedStyle(document.body)
+			const animationLength = parseInt(bodyStyles.getPropertyValue('--animation-quick')) || 100
+			setTimeout(() => {
+				emit('settings-toggled', {
+					open: this.open,
+				})
+				// We wait for 1.5 times the animation length to give the animation time to really finish.
+			}, 1.5 * animationLength)
```
skjnldsv, line 94: THis probably should be handled by the Modal component.
So you can listen like @doneOpen or something

ma12-co: It's the same as we did for the navigation. Currently the modal dosn't emit any event when the animation ends. What do you suggest?

ma12-co:
Review pleeeez :)

src/components/AppSettingsDialog/AppSettingsDialog.vue
```diff
@@ -0,0 +1,260 @@
...
+export default {
+	data() {
+		return {
+			settingsOpen: false,
+		}
+	},
+}
+</script>
+```
+ </docs>
```
skjnldsv, line 52:
```suggestion
<docs>
Just nest the `AppSettingSections` component into `AppSettingsDialog`,
providing the sectionTitle prop. You can put your settings within each
`AppSettingsSection` component.

```vue
<template>
	<div>
		<button @click="settingsOpen = true">Show Settings</button>
		<AppSettingsDialog :open.sync="settingsOpen" >
			<AppSettingsSection sectionTitle="Example title 1">
				Some example content
			</AppSettingsSection>
			<AppSettingsSection sectionTitle="Example title 2">
				Some more content
			</AppSettingsSection>
		</AppSettingsDialog>
	</div>
</template>

<script>
export default {
	data() {
		return {
			settingsOpen: false,
		}
	},
}
</script>
```
</docs>
```

Comment (skjnldsv) [inline src/components/AppSettingsDialog/AppSettingsDialog.vue:52]:
```suggestion
</docs>
```

src/components/AppSettingsDialog/AppSettingsDialog.vue
```diff
@@ -0,0 +1,260 @@
...
+		return {
+			selectedSection: '',
+		}
+	},
+
+	computed: {
+
+		content() {
+			return document.querySelector('.app-settings__content')
+		},
```
skjnldsv, line 90: Doesn't seems used?

src/components/AppSettingsDialog/AppSettingsDialog.vue
```diff
@@ -0,0 +1,260 @@
...
+
+		hasNavigation() {
+			if (this.isMobile || this.hideNavigation) {
+				return false
+			} else {
+				return true
+			}
+		},
```
skjnldsv, line 98: I like the hideNavigation thing!
Fancy!

skjnldsv (changes requested):
Great work!
Some missing mandatory parts:
- [x] No aria implementation
  https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html#rps_label
  You're probably gonna need `id`s as well for your `AppSettingsSection`
- [x] No keyboard navigation 
  https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html#kbd_label
- [x] The Navigation design is odd and we never used this anywhere else on Nextcloud.
  I think we should keep the same design we have on `/apps/settings`

Other than this! Great work :+1:

src/components/AppSettingsSection/AppSettingsSection.vue
```diff
@@ -0,0 +1,56 @@
...
+		<slot />
+	</div>
+</template>
+
+<script>
+export default {
+	name: 'AppSettingsSection',
+
+	props: {
+		sectionTitle: {
```
skjnldsv, line 38:
```suggestion
		title: {
```

title is enough, the component is called `AppSettingsSection`

src/components/AppSettingsSection/AppSettingsSection.vue
```diff
@@ -0,0 +1,56 @@
...
+ - GNU Affero General Public License for more details.
+ -
+ - You should have received a copy of the GNU Affero General Public License
+ - along with this program. If not, see <http://www.gnu.org/licenses/>.
+ -
+ -->`
+
+<template>
+	<div class="app-settings-section">
+		<h2 :id="sectionTitle"
```
skjnldsv, line 25: You need a specific id. And ideally a way to call the AppSettingsDialog directly with a provided id. I might want to have the second tab opened by default (shortcut to a setting for example)

src/components/AppSettingsDialog/AppSettingsDialog.vue
```diff
@@ -0,0 +1,260 @@
...
+		padding: 0 20px;
+		cursor: pointer;
+		white-space: nowrap;
+		text-overflow: ellipsis;
+		overflow: hidden;
+		background-color: transparent;
+		border: none;
+		&:hover {
+			background-color: var(--color-background-hover);
+		}
```
skjnldsv, line 253:
```suggestion
		&:hover,
		&:focus {
			background-color: var(--color-background-hover);
		}
```

src/components/AppSettingsDialog/AppSettingsDialog.vue
```diff
@@ -0,0 +1,260 @@
...
+		word-wrap: break-word;
+		padding: 0 20px;
+	}
+}
+
+.navigation-list {
+	&__link {
+		display: block;
+		font-size: 16px;
+		height: 44px;
```
skjnldsv, line 239: Please use the available variables

```suggestion
		height: $clickable-area;
```

src/components/AppSettingsDialog/AppSettingsDialog.vue
```diff
@@ -0,0 +1,260 @@
...
+	padding: 28px 16px;
+	display: flex;
+	height: 600px;
+	width: 100%;
+	&__navigation {
+		width: 200px;
+		margin-right: 20px;
+	}
+	&__content {
+		width: 400px;
```
skjnldsv, line 227: what if width is smaller than 400?

You also need a `max-width: 100vw;`

src/components/AppSettingsDialog/AppSettingsDialog.vue
```diff
@@ -0,0 +1,260 @@
...
+		overflow-y: auto;
+		overflow-x: hidden;
+		word-wrap: break-word;
```
skjnldsv, line 230:
```suggestion
```
Let's not enforce such drastic measures. Why was it needed?

src/components/AppSettingsDialog/AppSettingsDialog.vue
```diff
@@ -0,0 +1,260 @@
...
+}
+
+</script>
+
+<style lang="scss" scoped>
+.app-settings {
+	padding: 28px 16px;
+	display: flex;
+	height: 600px;
+	width: 100%;
```
skjnldsv, line 221: I'm confused about why the height is defined here but the width is defined in `&__content` ?

jancborchardt:
@ma12-co can you post screenshots of this? :) @skjnldsv mentioned something about tabs, and I’m a bit worried that we’re overengineering, since for now no app settings should really need tabs (as I think we discussed).

ma12-co:
> @ma12-co can you post screenshots of this? :) @skjnldsv mentioned something about tabs, and I’m a bit worried that we’re overengineering, since for now no app settings should really need tabs (as I think we discussed).

@jancborchardt I recall that we finally agreed on a gnome like left navigation which is can be hidden or shown on desktop depending on the app needs, and hidden always on mobile. So that's what I did :)

Screenshot from 2020-09-14 08-41-12

src/components/AppSettingsSection/AppSettingsSection.vue
ma12-co: follow up?

ma12-co:
> I think we should keep the same design we have on /apps/settings


@skjnldsv, that design is outdated and doesn't incorporate the new specs for hover - focus - active behaviors. The only new thing I introduced here is the rounded corners, which I think looks good, and it'll look even better once we'll have icons in those chips. 
This style is similar to the multiselection backgrounts in the talk new conversation flow, which you ported to the vue lib.

jancborchardt:
(As discussed @ma12-co let’s make sure the navigations are anchors instead of distinct pages.)

ma12-co:
Ok, here's the state of things. Keyboard navigation and continuous scroll added. I have to move on to other tasks that will keep me busy until the conf, could we get this in as is for now so that we can start using it in talk? 

and then for a followup:
- [ ] Add slots for icons in the navigation
- [ ] Allow to open the setting with a specific anchor selected
- [ ] Compute whitespace at the end of the settings so that upon clicking on a navigation item, the first thing shown in the scroller is always the title of the clicked section  


Peek 2020-09-22 12-58

src/components/AppSettingsSection/AppSettingsSection.vue
ma12-co: specific id added

nickvergessen:
Currently when you scroll the active state of the section disappears. I would have assumed it stays on the active item and as you scroll it also scrolls down?

ma12-co:
> I would have assumed it stays on the active item and as you scroll it also scrolls down

It's probably what we ultimately want, yes, but for now I think that this is OK. In Jan's reference (chrome settings) the nav element stays in an active state even if you scroll away for me. I thought that that was way worse of a behavior.

src/components/AppSettingsDialog/AppSettingsDialog.vue
```diff
@@ -0,0 +1,308 @@
...
+
+	mounted() {
+		// Select first settings section
+		this.selectedSection = this.$slots.default[0].componentOptions.propsData.title
+	},
+
+	updated() {
+		if (!this.scroller) {
+			// Get the scroller element
+			this.scroller = document.querySelector('.app-settings__content')
```
skjnldsv, line 118: Maybe use `this.$el` to avoid conflicts?

src/components/AppSettingsDialog/AppSettingsDialog.vue
```diff
@@ -0,0 +1,308 @@
...
+		}
+		if (this.scroller && !this.addedScrollListener) {
+			this.scroller.addEventListener('scroll', debounce(() => {
+				this.handleScroll()
+			}), 300)
+			this.addedScrollListener = true
+		}
```
skjnldsv, line 125:
```suggestion
		// Add the scroller scroll listener once after the first update
		if (this.scroller && !this.addedScrollListener) {
			this.scroller.addEventListener('scroll', debounce(() => {
				this.handleScroll()
			}), 300)
			this.addedScrollListener = true
		}
```

src/components/AppSettingsDialog/AppSettingsDialog.vue
```diff
@@ -0,0 +1,308 @@
...
+	methods: {
+
+		/**
+		 * Builds the settings navigation menu
+		 * @param {object} slots The default slots object passed from the render function.
+		 * @returns {array} the navigation items
+		 */
+		getSettingsNavigation(slots) {
+			// Array of navigationitems strings
+			const navigationItems = slots.filter(vNode => vNode.componentOptions).map(vNode => vNode.componentOptions.propsData.title)
```
skjnldsv, line 140: Please add optional chaining to be safe

src/components/AppSettingsDialog/AppSettingsDialog.vue
```diff
@@ -0,0 +1,308 @@
...
+		overflow-x: hidden;
+		padding: 0 20px;
+	}
+}
+
+.navigation-list {
+	&__link {
+		display: block;
+		font-size: 16px;
+		height: 44px;
```
skjnldsv, line 286:
```suggestion
		height: $clickable-area;
```

src/components/AppSettingsSection/AppSettingsSection.vue
```diff
@@ -0,0 +1,62 @@
...
+	props: {
+		title: {
+			type: String,
+			required: true,
+		},
+	},
+	computed: {
+		// generate an id for each settingssection based on the title without whitespaces
+		id() {
+			return 'settingsSection_' + this.title.replace(/\s+/g, '')
```
skjnldsv, line 46: Please use lowercase for id/classes

skjnldsv (changes requested):
A few comments!
Great work!!

src/components/AppSettingsDialog/AppSettingsDialog.vue
ma12-co: what do you mean here @skjnldsv ??

skjnldsv: https://duckduckgo.com/?q=optional+chaining&t=brave&ia=web

ma12-co: Can you show me exactly where you'd put it? I'm afraid I can't see the problem you're trying to solve

skjnldsv: Well you're filtering by chaining multiple objects properties.
Make sure everything is defined

src/components/AppSettingsDialog/AppSettingsDialog.vue
```diff
@@ -0,0 +1,308 @@
...
+		if (this.scroller && !this.addedScrollListener) {
+			this.scroller.addEventListener('scroll', debounce(() => {
+				this.handleScroll()
+			}), 300)
+			this.addedScrollListener = true
+		}
+	},
+
+	beforeDestroy() {
+		document.removeEventListener('scroll', this.handleScroll)
```
skjnldsv, line 129:
```suggestion
		this.scroller.removeEventListener('scroll', this.handleScroll)
```

Also, won't work as you're using another function on line 121. You're passing debounce and not `this.handleScroll`. The references are different

src/components/AppSettingsSection/AppSettingsSection.vue
```diff
@@ -0,0 +1,62 @@
...
+ - GNU Affero General Public License for more details.
+ -
+ - You should have received a copy of the GNU Affero General Public License
+ - along with this program. If not, see <http://www.gnu.org/licenses/>.
+ -
+ -->`
+
+<template>
+	<div :id="id" class="app-settings-section">
+		<h2 :id="title"
```
skjnldsv, line 25: h2 is most likely too high?
Should probably be h3 and the whole settings should have one h2 stating 'Settings for xxx' or something that define the Settings Modal as its own context.

https://www.w3.org/WAI/tutorials/page-structure/headings/

skjnldsv: @jancborchardt thoughts?

jancborchardt: Yeah, basically the dialog title e.g. "Talk settings" should be h2, meaning that any subcategory should be h3.

src/components/AppSettingsDialog/AppSettingsDialog.vue
ma12-co: Good catch, that was from before I added the debounce... So I passed in the whole function as it's defined above, good?

skjnldsv: No, you need to have the same reference, so put the debounce inside methods :)

skjnldsv: ALso, please rebase and squash everything into one commit! :rocket:

skjnldsv (approved):

ma12-co:
time to merge??????????????

skjnldsv:
> time to merge??????????????

missing 1 reviewer and l10n needs to be extracted again :)

ChristophWurst (approved):
:+1: :elephant:

ma12-co:
:tada::tada::tada::tada:

PR #1288, opened 2020-08-14
Allow to toggle editable title by click

raimund-schluessler:
This PR allows to show the title input field by clicking on the title. Also it makes sure that the title input is rendered in the same style as the title itself:

title

 Partially implements #1090.

### DIFF
--- a/src/components/AppSidebar/AppSidebar.vue
+++ b/src/components/AppSidebar/AppSidebar.vue
@@ -80,6 +80,37 @@
 	</script>
 	```
 
+	### Editable title after click
+	```vue
+	<template>
+		<AppSidebar
+			:title="title"
+			:title-editable.sync="titleEditable"
+			:title-placeholder="titlePlaceholder"
+			:subtitle="subtitle"
+			@update:title="titleUpdate">
+			<!-- Insert your slots and tabs here -->
+		</AppSidebar>
+	</template>
+	<script>
+		export default {
+			data() {
+				return {
+					title: 'cat-picture.jpg',
+					titlePlaceholder: 'Filename',
+					subtitle: 'last edited 3 weeks ago',
+					titleEditable: false
+				}
+			},
+			methods: {
+				titleUpdate(e) {
+					this.title = e
+				}
+			}
+		}
+	</script>
+	```
+
 	### Empty sidebar for e.g. empty content component.
 	```vue
 	<template>
--- a/src/components/AppSidebar/AppSidebar.vue
+++ b/src/components/AppSidebar/AppSidebar.vue
@@ -123,36 +123,55 @@
 				</div>
 
 				<!-- sidebar details -->
-				<div v-if="!empty" :class="{ 'app-sidebar-header__desc--with-star': canStar, 'app-sidebar-header__desc--with-subtitle': subtitle && !titleEditable, 'app-sidebar-header__desc--editable': titleEditable && !subtitle, 'app-sidebar-header__desc--with-subtitle--editable': titleEditable && subtitle}" class="app-sidebar-header__desc">
+				<div v-if="!empty"
+					:class="{
+						'app-sidebar-header__desc--with-star': canStar,
+						'app-sidebar-header__desc--with-subtitle': subtitle && !titleEditable,
+						'app-sidebar-header__desc--editable': titleEditable && !subtitle,
+						'app-sidebar-header__desc--with-subtitle--editable': titleEditable && subtitle
+					}"
+					class="app-sidebar-header__desc">
 					<!-- favourite icon -->
-					<a v-if="canStar"
-						:class="{ 'icon-starred': isStarred&& !starLoading, 'icon-star': !isStarred && !starLoading, 'icon-loading-small': starLoading }"
-						class="app-sidebar-header__star"
-						@click.prevent="toggleStarred" />
+					<div v-if="canStar" class="app-sidebar-header__star-action">
+						<a :class="{
+								'icon-starred': isStarred && !starLoading,
+								'icon-star': !isStarred && !starLoading,
+								'icon-loading-small': starLoading
+							}"
+							class="app-sidebar-header__star"
+							@click.prevent="toggleStarred" />
+					</div>
 
 					<!-- main title -->
-					<h2 v-if="!titleEditable" v-linkify="title" class="app-sidebar-header__title" />
-					<template v-if="titleEditable">
-						<form
-							class="rename-form"
-							@submit.prevent="onSubmitTitle">
-							<input
-								v-focus
-								class="app-sidebar-header__title-input"
-								type="text"
-								:placeholder="titlePlaceholder"
-								:value="title"
-								@keydown.esc="onDismissEditing"
-								@input="onTitleInput">
-							<button
-								class="icon-confirm"
-								type="submit" />
-						</form>
-					</template>
-					<!-- secondary title -->
-					<p v-if="subtitle.trim() !== ''" class="app-sidebar-header__subtitle">
-						{{ subtitle }}
-					</p>
+					<div class="app-sidebar-header__title">
+						<h2 v-show="!titleEditable"
+							v-linkify="title"
+							class="app-sidebar-header__maintitle"
+							@click.self="editTitle($event)" />
+						<template v-if="titleEditable">
+							<form
+								v-click-outside="() => onSubmitTitle()"
+								class="app-sidebar-header__maintitle-form"
+								@submit.prevent="onSubmitTitle">
+								<input
+									ref="titleInput"
+									v-focus
+									class="app-sidebar-header__maintitle-input"
+									type="text"
+									:placeholder="titlePlaceholder"
+									:value="title"
+									@keydown.esc="onDismissEditing"
+									@input="onTitleInput">
+								<button
+									class="icon-confirm"
+									type="submit" />
+							</form>
+						</template>
+						<!-- secondary title -->
+						<p v-if="subtitle.trim() !== ''" class="app-sidebar-header__subtitle">
+							{{ subtitle }}
+						</p>
+					</div>
 
 					<!-- header main menu -->
 					<Actions v-if="$slots['secondary-actions']" class="app-sidebar-header__menu" :force-menu="forceMenu">
@@ -177,6 +196,7 @@ import Focus from '../../directives/Focus'
 import Linkify from '../../directives/Linkify'
 import l10n from '../../mixins/l10n'
 import AppSidebarTabs from '../AppSidebarTabs/AppSidebarTabs'
+import { directive as ClickOutside } from 'v-click-outside'
 
 export default {
 	name: 'AppSidebar',
@@ -187,6 +207,7 @@ export default {
 	directives: {
 		focus: Focus,
 		linkify: Linkify,
+		ClickOutside,
 	},
 	mixins: [l10n],
 	props: {
@@ -199,6 +220,10 @@ export default {
 			default: '',
 			required: true,
 		},
+
+		/**
+		 * Allow to edit the sidebar title.
+		 */
 		titleEditable: {
 			type: Boolean,
 			default: false,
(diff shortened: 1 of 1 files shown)


src/components/AppSidebar/AppSidebar.vue
```diff
@@ -123,38 +123,42 @@
 				</div>
 
 				<!-- sidebar details -->
-				<div v-if="!empty" :class="{ 'app-sidebar-header__desc--with-star': canStar, 'app-sidebar-header__desc--with-subtitle': subtitle && !titleEditable, 'app-sidebar-header__desc--editable': titleEditable && !subtitle, 'app-sidebar-header__desc--with-subtitle--editable': titleEditable && subtitle}" class="app-sidebar-header__desc">
+				<div v-if="!empty" :class="{ 'app-sidebar-header__desc--with-star': canStar, 'app-sidebar-header__desc--with-subtitle': subtitle && !titleEditing, 'app-sidebar-header__desc--editable': titleEditing && !subtitle, 'app-sidebar-header__desc--with-subtitle--editable': titleEditing && subtitle}" class="app-sidebar-header__desc">
```
skjnldsv, line 126: Let's split classes to individual lines please :stuck_out_tongue:

src/components/AppSidebar/AppSidebar.vue
```diff
@@ -199,8 +205,13 @@ export default {
...
 			required: true,
 		},
+
+		/**
+		 * Allow to edit the sidebar title.
+		 * Accepts true, false, and 'toggle' for showing the input on click.
+		 */
 		titleEditable: {
-			type: Boolean,
+			type: [Boolean, String],
```
skjnldsv, line 214: Do we really want to have this as an option? I'm fine having the click to edit as default if `titleEditable` is true :)

src/components/AppSidebar/AppSidebar.vue
```diff
@@ -314,6 +329,19 @@ export default {
...
+			if (this.titleEditable === false) {
+				return
+			}
+			this.editing = true
+			// Focus the title input
+			this.$nextTick(
+				() => this.$refs.titleInput.focus()
+			)
+			this.$emit('start-editing')
+		},
```
skjnldsv, line 343: Please only emit the @update:titleEditable. If the dev don't implement the listener, then it will be disabled. Let's not have a managed inner state. It would be cleaner and allow the dev more flexibility :wink:

src/components/AppSidebar/AppSidebar.vue
```diff
@@ -445,53 +494,47 @@ $top-buttons-spacing: 6px;
...
 				width: $clickable-area;
 				height: $clickable-area;
 				padding: $icon-margin;
 			}
 			// main menu
 			.app-sidebar-header__menu {
-				position: absolute;
-				right: $clickable-area / 2;
+				height: 44px;
+				width: 44px;
```
skjnldsv, line 505:
```suggestion
				height: $clickable-area;
				width: $clickable-area;
```

src/components/AppSidebar/AppSidebar.vue
```diff
@@ -425,16 +472,18 @@ $top-buttons-spacing: 6px;
...
 			}
-			input.app-sidebar-header__title-input {
-				width: 100%;
+			input.app-sidebar-header__maintitle-input {
+				flex: 1 1 auto;
 				margin: 0;
 				padding: $desc-input-padding;
-				font-size: 16px;
+				font-size: 20px;
+				font-weight: bold;
```
skjnldsv, line 486: We rarely use bold on Nextcloud. @jancborchardt :)

src/components/AppSidebar/AppSidebar.vue
raimund-schluessler: I don't mind. I just didn't want to change the current behavior to much, because I wasn't sure if it is needed like that somewhere.

src/components/AppSidebar/AppSidebar.vue
raimund-schluessler: What do you mean with managed inner state? The `this.editing`?

src/components/AppSidebar/AppSidebar.vue
raimund-schluessler: It is to make the text appear the same way as the `<h2>` is rendered. Otherwise it will look different and jump when editing.

src/components/AppSidebar/AppSidebar.vue
skjnldsv: Yes, I mean with `this.editing`.
If you just forward the event to the parent with a sync modifier `update:editing` and let the dev the option to trigger the edit from the parent.
So you'd have to change from titleEditable to `titleEditing` or something and just fire the event on click. That way if the devs want to trigger the rename, they can just change the prop and if they want to forbid it, they can just not provide the prop or not provide a sync modifier (the editing prop will be immutable)

And instead of firing $emit('start-editing') the dev can just listen to the `update:editing` event. If it's true, then the editing is on.

raimund-schluessler: I think that only works if we only allow two states for titleEditable (`true` and `false`) but not the third `"toggle"` as it is now, could that be?

skjnldsv: You don't need toggle then.
Always allows it, if the dev doesn't want to change it, then it can disable the editable directly.

src/components/AppSidebar/AppSidebar.vue
raimund-schluessler: Done.

src/components/AppSidebar/AppSidebar.vue
raimund-schluessler: I removed the toggle option. When editing is allowed, the input shows after clicking.
@skjnldsv I don't know if I got your idea correctly, but I think a managed inner state with `this.titleEditing` is still necessary. Otherwise one cannot distinguish between if editing is forbidden/disabled or currently just not active.
But maybe I got it wrong. In that case feel free to adjust the code accordingly.

skjnldsv: No it should not be necessary.
If showing everything rely on the prop titleEditable and the dev doesn't allow it to change (so don't use the `.sync` modifier), it will not change, so clicking the title will not toggle the `titleEditable` and will have no outcome :)

If the dev want to allow the title to be editable, it should just add the sync modifier and clicking the title will then change the prop on the parent.

src/components/AppSidebar/AppSidebar.vue
```diff
@@ -123,36 +123,55 @@
...
+					}"
+					class="app-sidebar-header__desc">
 					<!-- favourite icon -->
-					<a v-if="canStar"
-						:class="{ 'icon-starred': isStarred&& !starLoading, 'icon-star': !isStarred && !starLoading, 'icon-loading-small': starLoading }"
-						class="app-sidebar-header__star"
-						@click.prevent="toggleStarred" />
+					<div v-if="canStar" class="app-sidebar-header__star-action">
+						<a :class="{
+								'icon-starred': isStarred&& !starLoading,
```
skjnldsv, line 137:
```suggestion
								'icon-starred': isStarred && !starLoading,
```

src/components/AppSidebar/AppSidebar.vue
raimund-schluessler: Ok, I think I got it now. Thanks for the feedback @skjnldsv!

raimund-schluessler:
Shall we also edit the editable title example to use `sync`?

skjnldsv:
> Shall we also edit the editable title example to use `sync`?

yes so we can have both please :)

raimund-schluessler:
I added an editable title after click example.

src/components/AppSidebar/AppSidebar.vue
```diff
@@ -314,6 +370,20 @@ export default {
 			this.$emit('update:starred', this.isStarred)
 		},
 
+		editTitle(event) {
+			// Don't edit the title if editing is disabled or if we clicked on a link
+			if (event.target.tagName === 'A') {
+				return
+			}
```
skjnldsv, line 377: Maybe you can use the `.exact` event specifier ?

skjnldsv:
I'm an idiiot, I thought this pr was already in, so this is why I edited my other pr #1303 ... :see_no_evil:

src/components/AppSidebar/AppSidebar.vue
raimund-schluessler: Good idea to use a specifier :+1. Works well.

raimund-schluessler:
> I'm an idiiot, I thought this pr was already in, so this is why I edited my other pr #1303 ... 🙈
> I'm tired :(

Don't worry, it's fine. But it explains #1303 🙂🙈

raimund-schluessler:
@skjnldsv So how should we proceed? Review and merge this one first and adjust #1303 afterwards, or the other way around?

skjnldsv:
> @skjnldsv So how should we proceed? Review and merge this one first and adjust #1303 afterwards, or the other way around?

yours first :)

raimund-schluessler:
@skjnldsv In case you are busy, I can also take care of #1303 in the evening (since I broke it in #1281).

src/components/AppSidebar/AppSidebar.vue
```diff
@@ -336,9 +402,13 @@ export default {
...
 		onSubmitTitle(event) {
+			// Disable editing
+			this.$emit('update:titleEditable', false)
 			this.$emit('submit-title', event)
 		},
 		onDismissEditing() {
+			// Disable editing
+			this.$emit('update:titleEditable', false)
 			this.$emit('dismiss-editing')
 		},
```
skjnldsv, line 413:
```suggestion
		onSubmitTitle(event) {
			// Disable editing
			this.$emit('update:titleEditable', false)
		},
		onDismissEditing() {
			// Disable editing
			this.$emit('update:titleEditable', false)
		},
```
I'm not sure this is something we want anymore?
We can already listen to the `update:titleEditable` event now.

skjnldsv: Same for the `this.$emit('input-title', event)`, doesn't make sense to have both, no? :)

skjnldsv (approved):
A few nitpicks

src/components/AppSidebar/AppSidebar.vue
raimund-schluessler: > I'm not sure this is something we want anymore?
> We can already listen to the `update:titleEditable` event now.

I think it would be good to emit `submit-title` and `dismiss-editing` separately, since you can then distinguish between the user pressing the `ESC` button and really submitting the title change by the button or clicking outside the input. When pressing ESC, as a user I wouldn't expect the title to be changed, but reset to the initial value. This wouldn't be possible if we only emit `update:titleEditable`.

raimund-schluessler: As for the `input-title` emission, I think we could remove it (although I didn't test this yet).

raimund-schluessler: > 
> 
> As for the `input-title` emission, I think we could remove it (although I didn't test this yet).

I will test this in the evening and remove it if possible.

raimund-schluessler: I think it's fine to remove the `input-title` event. Although it is a breaking change, I didn't find any repository in the nextcloud org which uses this. So I guess no one will notice the event is gone.

skjnldsv:
> @skjnldsv In case you are busy, I can also take care of #1303 in the evening (since I broke it in #1281).

Feel free to do so then :)

src/components/AppSidebar/AppSidebar.vue
```diff
@@ -123,36 +154,55 @@
...
-						{{ subtitle }}
-					</p>
+					<div class="app-sidebar-header__title">
+						<h2 v-show="!titleEditable"
+							v-linkify="title"
+							class="app-sidebar-header__maintitle"
+							@click.self="editTitle($event)" />
+						<template v-if="titleEditable">
+							<form
+								v-click-outside="() => onSubmitTitle()"
```
skjnldsv, line 184: @raimund-schluessler I just noticed this, we do not forward any event here. :see_no_evil: 
Any reason you did not just do that?

```suggestion
								v-click-outside="onSubmitTitle"
```

raimund-schluessler: No, I probably just overlooked it.

skjnldsv: All good, thanks :)

PR #1289, opened 2020-08-16
Linkify sidebar title

raimund-schluessler:
Enables linkifying the sidebar title. Partially implement #1090. URLs are now automatically converted to clickable links:

linkify

At the moment it is always enabled. Should we make that configurable with a prop?


skjnldsv:
I'm not sure about this to be honest. Does it make sense to have links there? :thinking:

raimund-schluessler:
> 
> 
> I'm not sure about this to be honest. Does it make sense to have links there? 🤔

For the Tasks app it absolutely makes sense. It was requested a few times before we implemented it as far as I remember, and I also happen to use it once in a while. It is quite convenient to just paste an URL in the task description which is then automatically rendered as a link, so it can be clicked easily. Otherwise you would need to copy the URL and paste it manually. This would be especially tedious when using #1288, since you would need to go to edit mode first.

But we can sure disable linkifying by default and enable it with a prop.

skjnldsv:
> But we can sure disable linkifying by default and enable it with a prop.

No, let's avoid too much options :p
Let get this in :rocket:

nickvergessen:
We kind of need a way to prevent this.

raimund-schluessler:
With https://github.com/nextcloud/nextcloud-vue/pull/1337 we disabled linkifying by default. You have to enable it with a prop now.

PR #1338, opened 2020-08-21
Fix sidebar layout in compact mode

raimund-schluessler:
Fixes the sidebar layout in compact mode, implements the second part of #1366:

Screenshot_2020-08-21 Tasks - Nextcloud

@skjnldsv Since there are many valid combinations of options (compact, with-star, with-header-figure, with-secondary-actions), it needs a good test to find any possibly remaining issues.

### DIFF
--- a/src/components/AppSidebar/AppSidebar.vue
+++ b/src/components/AppSidebar/AppSidebar.vue
@@ -126,7 +126,7 @@
 		<aside id="app-sidebar-vue" class="app-sidebar">
 			<header :class="{
 					'app-sidebar-header--with-figure': hasFigure,
-					'app-sidebar-header--compact': compact
+					'app-sidebar-header--compact': compact,
 				}"
 				class="app-sidebar-header">
 				<!-- close sidebar button -->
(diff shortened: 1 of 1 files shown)


src/components/AppSidebar/AppSidebar.vue
```diff
@@ -338,6 +342,9 @@ export default {
 		hasFigureClickListener() {
 			return this.$listeners['figure-click']
 		},
+		isCompact() {
+			return this.hasFigure && this.compact
```
skjnldsv, line 346: Okay, so, sorry about that, but I actually thought about this and I forgot one edge case.
One future addition is to go compact when scrolling down the sidebar. It will be cleaner for super long list like sharing can be.

Thus maybe leaving this to just the `compact` prop would be better?
What do you think?

raimund-schluessler: I don't mind, we can leave it to the prop.

skjnldsv (changes requested):
Css is much cleaner now :)

- [ ] Menu in expanded still misses the right alignment like `close`
  Capture d’écran_2020-08-24_12-42-10
- [ ] Thumbnail is not square anymore
  Capture d’écran_2020-08-24_12-43-31

raimund-schluessler:
@skjnldsv Given the huge number of valid combinations of props and slots for the component, I am quite afraid that I will break something else when fixing the issues you mentioned. Since #1339 seems to basically work now, I would like to take screenshots of the AppSidebar component for every combination we have before continuing here. We could then identify the combinations with layout issues and make sure that we don't introduce new problems when fixing the issues found. I will try to do so in the evening.

skjnldsv:
> @skjnldsv Given the huge number of valid combinations of props and slots for the component, I am quite afraid that I will break something else when fixing the issues you mentioned. Since #1339 seems to basically work now, I would like to take screenshots of the AppSidebar component for every combination we have before continuing here. We could then identify the combinations with layout issues and make sure that we don't introduce new problems when fixing the issues found. I will try to do so in the evening.

I'm fine doing this in a follow up yes :)

skjnldsv (approved):
Please move to just `compact` and rebase :)

raimund-schluessler:
> 
> 
> Please move to just `compact` and rebase :)

Done.

I cannot reproduce the issues you described in https://github.com/nextcloud/nextcloud-vue/pull/1338#pullrequestreview-473365867 by the way. To me it looks fine. Are you sure you tested the latest state?

skjnldsv:
> Are you sure you tested the latest state?

I think I pulled yes, let's get this in and see what comes next :)

skjnldsv (approved):

raimund-schluessler:
I will work on #1339 in the evening. Hopefully we can then check all different options properly.

PR #1339, opened 2020-08-23
Add visual regression testing with cypress

raimund-schluessler:
This PR tries to add visual regression testing with cypress. I think we really need this, since some components start to be that complex, that we cannot change them without automatically testing for visual regressions. E.g. the AppSidebar component has at least 64 valid combinations of props and slots only for the header, which all change the visual appearance of the component. Testing this manually is basically impossible, since every small change needs to be tested with every valid combination of options. So visual regression testing would help a lot to resolve #1338 and prevent similar problems in the future.

~However, the cypress setup does not work at the moment, I get the multiple similar errors like this:~ Fixed with a90e2b4d2b8bb52e57bc38bb9f7b56f47ba94768
```
Error: Webpack Compilation Error
./src/components/AppSidebar/AppSidebar.vue?vue&type=style&index=0&id=4a2a5f2f&lang=scss&scoped=true& (./node_modules/css-loader/dist/cjs.js!./node_modules/vue-loader/lib/loaders/stylePostLoader.js!./node_modules/resolve-url-loader!./node_modules/sass-loader/dist/cjs.js??ref--1-3!./node_modules/vue-loader/lib??vue-loader-options!./src/components/AppSidebar/AppSidebar.vue?vue&type=style&index=0&id=4a2a5f2f&lang=scss&scoped=true&)
Module build failed (from ./node_modules/sass-loader/dist/cjs.js):
Error: Node Sass does not yet support your current environment: Linux 64-bit with Unsupported runtime (80)
For more information on which environments are supported please see:
https://github.com/sass/node-sass/releases/tag/v4.14.1
 @ ./src/components/AppSidebar/AppSidebar.vue?vue&type=style&index=0&id=4a2a5f2f&lang=scss&scoped=true& (./node_modules/vue-style-loader!./node_modules/css-loader/dist/cjs.js!./node_modules/vue-loader/lib/loaders/stylePostLoader.js!./node_modules/resolve-url-loader!./node_modules/sass-loader/dist/cjs.js??ref--1-3!./node_modules/vue-loader/lib??vue-loader-options!./src/components/AppSidebar/AppSidebar.vue?vue&type=style&index=0&id=4a2a5f2f&lang=scss&scoped=true&) 4:14-382
 @ ./src/components/AppSidebar/AppSidebar.vue?vue&type=style&index=0&id=4a2a5f2f&lang=scss&scoped=true&
 @ ./src/components/AppSidebar/AppSidebar.vue
 @ ./tests/visual/components/AppSidebar/AppSidebar.spec.js
 @ multi ./tests/visual/components/AppSidebar/AppSidebar.spec.js
```

Now I get
```
Running:  AppSidebar/AppSidebar.spec.js                                                   (1 of 1)


  1) An uncaught error was detected outside of a test

  0 passing (862ms)
  1 failing

  1) An uncaught error was detected outside of a test:
     ReferenceError: The following error originated from your test code, not from Cypress.

  > require is not defined

When Cypress detects uncaught errors originating from your test code it will automatically fail the current test.

```
Will look into this tomorrow.

Any feedback is very welcome, since I am new to cypress. @skjnldsv I saw that you use cypress for the viewer app, maybe you have an idea what is wrong here.


cypress.json
```diff
@@ -0,0 +1,9 @@
+{
+  "projectId": "xysa6x",
```
skjnldsv, line 2: You need your own id https://dashboard.cypress.io/organizations/07c47405-2e3c-4482-bc47-727f1737f458/integrations/github
But it's a bit bugged :thinking: 
I cannot enable this repo right now.

cypress/start.sh
```diff
@@ -0,0 +1,17 @@
...
+else
+	echo "Launching docker server for the $appname app"
+	dockername="cypress_testing_$appname"
+
+	docker run -p 8000:80 --rm --name $dockername --volume $(pwd):/var/www/html/apps/$appname --env BRANCH=master --detach nextcloudci/server:server-16
+	npm i --no-save wait-on
+	npm run wait-on -- http://localhost:8000
+	docker exec --user www-data $dockername php occ app:enable $appname
+	docker exec --user www-data $dockername php occ config:system:set force_language --value en
+fi
```
skjnldsv, line 17: Not needed?

cypress/stop.sh
```diff
@@ -0,0 +1,12 @@
...
+appname=${PWD##*/}
+
+if [[ $appname == "cypress" ]]
+then
+	echo "Please run this app from your app root folder."
+else
+	echo "Killing server for the $appname app"
+	dockername="cypress_testing_$appname"
+	docker kill $dockername
+fi
```
skjnldsv, line 12: Not needed?

package.json
```diff
@@ -41,8 +43,12 @@
 		"@nextcloud/l10n": "^1.2.3",
 		"@nextcloud/router": "^1.0.2",
 		"core-js": "^3.6.5",
+		"cypress": "^5.0.0",
+		"cypress-image-snapshot": "^3.1.1",
+		"cypress-vue-unit-test": "^3.5.1",
 		"debounce": "1.2.0",
 		"emoji-mart-vue-fast": "^7.0.4",
+		"eslint-plugin-cypress": "^2.11.1",
```
skjnldsv, line 51: Should be dev dependencies

tests/visual/components/AppSidebar/AppSidebar.spec.js
```diff
@@ -0,0 +1,48 @@
+/**
```
skjnldsv, line 1: I' not fond of having cypress tests into the tests folder as this is not how cypress works and not how any other apps in Nextcloud does it. I'd prefer things global and the same across repos :wink: 
https://github.com/nextcloud/text/tree/master/cypress/integration
https://github.com/nextcloud/social/pull/243/files#diff-e011fa10528950fcd9a06a080c3aa384

skjnldsv:
>   > require is not defined
> Any feedback is very welcome, since I am new to cypress. @skjnldsv I saw that you use cypress for the viewer app, maybe you have an idea what is wrong here.

I have no clue. Never had this issue :(

tests/visual/components/AppSidebar/AppSidebar.spec.js
juliushaertl: I'd say for components that is a better approach. For E2E testing of apps the test structure is usually based on routes or individual views. Since we only test single components here having them near to the code makes sense imo.

juliushaertl:
> require is not defined

I'd try to switch the existing require calls with regular es5 imports. Require might fail if they are called within the browser context.

cypress.json
raimund-schluessler: Thanks.

tests/visual/components/AppSidebar/AppSidebar.spec.js
skjnldsv: @juliushaertl so where would you put them? :)
Alongside components?

cypress.json
skjnldsv: I also added the CYPRESS_RECORD_KEY to the secrets :)

cypress/start.sh
raimund-schluessler: Probably not, copied from the viewer repo. Although might be necessary to run in Docker to ensure reliable results on different platform. Let's see.

cypress/stop.sh
raimund-schluessler: Dito.

package.json
raimund-schluessler: Yes, of course.

tests/visual/components/AppSidebar/AppSidebar.spec.js
raimund-schluessler: I put them in the tests folder since I thought of them as tests as well (unit, integration and visual tests). But I am fine with whatever you dedice on.

raimund-schluessler:
It is this require that fails: https://github.com/nextcloud/nextcloud-vue/pull/1339/files#diff-4bcea4966f5e62ded90298c2b0907445R18-R20 Weird enough the previous one seems to run fine. Replacing it with `import` leads to this error:
```
 /var/www/nextcloud/apps/nextcloud-vue/cypress/plugins/index.js:22
import { onFilePreprocessor } from 'cypress-vue-unit-test/dist/preprocessor/webpack'
       ^

SyntaxError: Unexpected token {
    at Module._compile (internal/modules/cjs/loader.js:723:23)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at module.exports (/home/raimund/.cache/Cypress/5.0.0/Cypress/resources/app/packages/server/lib/plugins/child/run_plugins.js:192:15)
    at Object.<anonymous> (/home/raimund/.cache/Cypress/5.0.0/Cypress/resources/app/packages/server/lib/plugins/child/index.js:8:25)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:831:12)
    at startup (internal/bootstrap/node.js:283:19)
```

skjnldsv:
> import { onFilePreprocessor } from 'cypress-vue-unit-test/dist/preprocessor/webpack'

Try `import webpack from 'cypress-vue-unit-test/dist/preprocessor/webpack'`
and use `webpack.onFilePreprocessor` ?

raimund-schluessler:
> import webpack from 'cypress-vue-unit-test/dist/preprocessor/webpack'

Similar issue. I think `import` does not work in this context:

```
The plugins file is missing or invalid.

Your `pluginsFile` is set to `/var/www/nextcloud/apps/nextcloud-vue/cypress/plugins/index.js`, but either the file is missing, it contains a syntax error, or threw an error when required. The `pluginsFile` must be a `.js`, `.ts`, or `.coffee` file.

Or you might have renamed the extension of your `pluginsFile`. If that's the case, restart the test runner.

Please fix this, or set `pluginsFile` to `false` if a plugins file is not necessary for your project.

 /var/www/nextcloud/apps/nextcloud-vue/cypress/plugins/index.js:24
import webpack from 'cypress-vue-unit-test/dist/preprocessor/webpack'
       ^^^^^^^

SyntaxError: Unexpected identifier
    at Module._compile (internal/modules/cjs/loader.js:723:23)
```

raimund-schluessler:
I don't know why, but the `externals: [nodeExternals()],` config in webpack.common.js seems to be related to the issue. When I overwrite it in the cypress config https://github.com/nextcloud/nextcloud-vue/pull/1339/files#diff-4bcea4966f5e62ded90298c2b0907445R21 the error is fixed, cypress runs through and even creates a screenshot. So this is a step forward, I guess.

However, it seems that no styles are applied:

AppSidebar vue -- Renders snap

raimund-schluessler:
> However, it seems that no styles are applied:

Well, the component styles are there, but the global Nextcloud server styles, icons and fonts are missing. Any idea how to include them?

raimund-schluessler:
I think it basically works now. This is the screentshot cypress produces on my system for a simple AppSidebar:

AppSidebar vue -- Renders snap

Things to do:
- [ ] Fix jest to not run the visual tests
- [ ] Store screenshots within the visual tests folder?
- [ ] Take screenshots for all valid sidebar options (with and without figure, with and without star, with and without actions, etc. and all combinations)
- [ ] Cleanup?

skjnldsv:
Awesome work @raimund-schluessler :hugs:

raimund-schluessler:
I added a visual regression test for 24 combinations of figure, secondary-actions, starred and compact. Here an example image with all options:

AppSidebar.vue - starred_true - compact_true - header_image - secondary_button:
AppSidebar vue - starred_true - compact_true - header_image - secondary_button snap


Please have a look at them in `cypress\snapshots\AppSidebar\AppSidebar.spec.js`. So far I think they look good, but I haven't tested with long title and subtitle, without subtitle and in editable mode. So I guess there are more combinations to come 🙈

raimund-schluessler:
> 
> 
> I added a visual regression test for 24 combinations of figure, secondary-actions, starred and compact. Here an example image with all options:
> 
> AppSidebar.vue - starred_true - compact_true - header_image - secondary_button:
> AppSidebar vue - starred_true - compact_true - header_image - secondary_button snap
> 
> Please have a look at them in `cypress\snapshots\AppSidebar\AppSidebar.spec.js`. So far I think they look good, but I haven't tested with long title and subtitle, without subtitle and in editable mode. So I guess there are more combinations to come 🙈

skjnldsv (commented):
- [ ] Please add `cypress/videos` to gitignore :)
- [ ] Maybe also play with the background instead of the header?
- [ ] Shall we add a default SdidebarContent?
  - [ ] If not maybe we can only capture the header?

raimund-schluessler:
Looks like we have to run cypress in a Docker container as the images produced by the Github action differ from the ones on my machine. @skjnldsv Do you have an idea how to do that?

skjnldsv:
> Looks like we have to run cypress in a Docker container as the images produced by the Github action differ from the ones on my machine. @skjnldsv Do you have an idea how to do that?

Yes, check the action in viewer. There is a GitHub action for cypress :)

raimund-schluessler:
I don't seem to get this working. @skjnldsv I would appreciate if you could have a look at some point.

skjnldsv:
```
Cypress test command: npx cypress run --record --config-file cypress.json
/opt/hostedtoolcache/node/10.22.0/x64/bin/npx cypress run --record --config-file cypress.json
The plugins file is missing or invalid.

Your `pluginsFile` is set to `/home/runner/work/nextcloud-vue/nextcloud-vue/cypress/plugins/index.js`, but either the file is missing, it contains a syntax error, or threw an error when required. The `pluginsFile` must be a `.js`, `.ts`, or `.coffee` file.

Or you might have renamed the extension of your `pluginsFile`. If that's the case, restart the test runner.

Please fix this, or set `pluginsFile` to `false` if a plugins file is not necessary for your project.
```

skjnldsv:
> I don't seem to get this working. @skjnldsv I would appreciate if you could have a look at some point.

Let me have a look :+1:

raimund-schluessler:
Looks like `process.env.npm_package_version` in https://github.com/nextcloud/nextcloud-vue/blob/master/webpack.common.js#L16 is not defined when run via `npx`. When we directly run `npm run cypress` it runs fine, but the screenshots are different as stated above.

However, the screenshots are even slightly different when run locally depending on if I run via `npm run cypress` (cypress run) or with `npm run cypress:gui` (cypress open). Seems to be an aliasing issue.

skjnldsv:
`const appVersion = JSON.stringify(process.env.npm_package_version || 'test')` works

raimund-schluessler:
> weird, I'm using this in Viewer
> https://github.com/nextcloud/webpack-vue-config/blob/601dfd1c41b602883eafd21d334bbee6b8753cf7/webpack.js#L8

Is webpack called by cypress in the viewer repo? I had the impression cypress directly navigates to the url and doesn't invoke webpack there, since you don't test single components.

cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
+  "body": "Fixtures are a great way to mock data for responses to routes"
+}
```
skjnldsv, line 5: Not needed?

raimund-schluessler: Was auto-generated by `npm run cypress:gui`.

skjnldsv:
> Is webpack called by cypress in the viewer repo? I had the impression cypress directly navigates to the url and doesn't invoke webpack there, since you don't test single components.

The app is built by Cypress as far as I know yes :)
We can disable this iirc

skjnldsv:
https://github.com/cypress-io/github-action#custom-test-command

raimund-schluessler:
> 
> 
> `const appVersion = JSON.stringify(process.env.npm_package_version || 'test')` works
> 
> https://dashboard.cypress.io/projects/3paxvy/runs/1/specs

Should we then do
`const appVersion = JSON.stringify(process.env.npm_package_version || 'nextcloud-vue')`
or do you have an idea how to solve it properly?

raimund-schluessler:
> 
> 
> > Is webpack called by cypress in the viewer repo? I had the impression cypress directly navigates to the url and doesn't invoke webpack there, since you don't test single components.
> 
> The app is built by Cypress as far as I know yes :)
> We can disable this iirc
> 
> EDIT: https://github.com/cypress-io/github-action#custom-test-command

We might just run it with `npm run cypress` then.

skjnldsv:
Let me push :)

skjnldsv:
Btw https://github.com/palmerhq/cypress-image-snapshot/issues/138

skjnldsv:
AppSidebar vue ___ Renders AppSidebar vue        - starred_null        - compact_false        - header_none        - secondary_none       (1)
AppSidebar vue ___ Renders AppSidebar vue        - starred_null        - compact_false        - header_none        - secondary_none

Looks like it's related to the stupid compression and pixel scaling.
Might be the issue in https://github.com/cypress-io/cypress/issues/2102 then

raimund-schluessler:
> Looks like it's related to the stupid compression and pixel scaling.
> Might be the issue in [cypress-io/cypress#2102](https://github.com/cypress-io/cypress/issues/2102) then

Could also be because browser and OS are different.

raimund-schluessler:
> 
> 
> AppSidebar vue ___ Renders AppSidebar vue - starred_null - compact_false - header_none - secondary_none (1)
> AppSidebar vue ___ Renders AppSidebar vue - starred_null - compact_false - header_none - secondary_none
> 
> Looks like it's related to the stupid compression and pixel scaling.
> Might be the issue in [cypress-io/cypress#2102](https://github.com/cypress-io/cypress/issues/2102) then

Looks a little bit as if the font was different.

skjnldsv:
Seems to work with another plugin that seems supported
https://dashboard.cypress.io/projects/3paxvy/runs/4/test-results/825cd312-d8e3-4a68-864d-cef399271ce8

skjnldsv:
> Seems to work with another plugin that seems supported

Hum, not really. https://github.com/mjhea0/cypress-visual-regression/issues/50 :(

raimund-schluessler:
> > Seems to work with another plugin that seems supported
> 
> 
> 
> Hum, not really. https://github.com/mjhea0/cypress-visual-regression/issues/50 :(

Mind pushing your adjusted branch to the repo? Then I would play around with it a bit.

skjnldsv:
> Mind pushing your adjusted branch to the repo? Then I would play around with it a bit.

DOne, sorry! :)

raimund-schluessler:
> 
> 
> > Seems to work with another plugin that seems supported
> 
> Hum, not really. [mjhea0/cypress-visual-regression#50](https://github.com/mjhea0/cypress-visual-regression/issues/50) :(

Hm, no idea. I get the same error no matter what.

skjnldsv:
Fixed @raimund-schluessler 
The strings contained new lines :)

skjnldsv (approved):

raimund-schluessler:
Nice, it runs through on my machine now. But the screenshots taken are different from the previous ones. I will try to update them and push them here. Let's see.

skjnldsv:
> Nice, it runs through on my machine now. But the screenshots taken are different from the previous ones. I will try to update them and push them here. Let's see.

The screenshots seems to actually be the same :)
It currently fails because there is a tiny diff
```
The "AppSidebar.vue - starred_null - compact_false - header_none - secondary_button" image is different. Threshold limit exceeded! 
Expected: 0 
Actual: 0.13896919372599592
```

raimund-schluessler:
> 
> 
> > Nice, it runs through on my machine now. But the screenshots taken are different from the previous ones. I will try to update them and push them here. Let's see.
> 
> The screenshots seems to actually be the same :)
> It currently fails because there is a tiny diff
> 
> ```
> The "AppSidebar.vue - starred_null - compact_false - header_none - secondary_button" image is different. Threshold limit exceeded! 
> Expected: 0 
> Actual: 0.13896919372599592
> ```

That's what I meant with "different". The letters are rendered slightly different than before.

skjnldsv:
It seems it's not uploading the diff files though
Some example on how to upload the artifacts:

https://github.com/cypress-io/github-action/commit/22681e32e27bff3a01cb50a6bbdb511743193dee

skjnldsv:
We could also use https://github.com/meinaart/cypress-plugin-snapshots
Now that we have it working and we understand it better.
It seems up to date compared to https://github.com/palmerhq/cypress-image-snapshot

raimund-schluessler:
> We could also use https://github.com/meinaart/cypress-plugin-snapshots
> Now that we have it working and we understand it better.
> It seems up to date compared to https://github.com/palmerhq/cypress-image-snapshot

I don't mind. Both https://github.com/mjhea0/cypress-visual-regression and https://github.com/meinaart/cypress-plugin-snapshots seem to be fine.

The question is how to handle the small differences in the screenshots. Whether we can prevent the different rendering or increase the threshold.

raimund-schluessler:
@skjnldsv I think it uploaded correctly: https://dashboard.cypress.io/projects/3paxvy/runs/8/test-results/408ee0df-0b7a-41b3-b473-433ba35020af

skjnldsv:
> @skjnldsv I think it uploaded correctly: [dashboard.cypress.io/projects/3paxvy/runs/8/test-results/408ee0df-0b7a-41b3-b473-433ba35020af](https://dashboard.cypress.io/projects/3paxvy/runs/8/test-results/408ee0df-0b7a-41b3-b473-433ba35020af)

It's missing the superposée red diffs images :)

raimund-schluessler:
I moved the cypress runner into a docker container https://github.com/nextcloud/nextcloud-vue/pull/1339/commits/1e7d85e932d8d842c3ee50d62468ef01501ec633. This ensures that the local results and the results in CI are exactly the same, since they run in the samer docker container.

Now we only need to fix the uploading to cypress.io and cleanup a bit.

raimund-schluessler:
@skjnldsv Please have a look. I think it is in a good shape now.

I would add more conditions for AppSidebar (primary-actions, long title etc.) in a follow-up PR, so we can also test if it really finds regressions.

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      - name: Set up node ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Cypress run
+        env:
+          CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
+        run: npm run cypress:record
```
skjnldsv, line 35: I'm not fond, I think that complicates the setup a lot.
Why did we need to load this inside a docker container?

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,35 @@
...
+jobs:
+  jest:
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version: ['12.x']
+
+    name: Runner ${{ matrix.containers }}
```
skjnldsv, line 22: No parallelism anymore (Wasn't needed because we have 1 test, but how would we add more?)

.github/workflows/cypress.yml
raimund-schluessler: > Why did we need to load this inside a docker container?

Because the tests now pass, and failed before 😉 

The tiny differences we saw with the font rendering are caused by running the tests on different machines. Between them, maybe the browser, the display settings, the available fonts or whatever else differs slightly, leading to tiny differences in the rendering which make the tests fail.

Of course, we could increase the difference threshold to make the tests pass. But this is not really an option as it would lead to false positives.

The difference in this image is 0.1623 and is not a regression (it's the difference between running locally on my machine and inside the docker container):
AppSidebar vue - starred_null - compact_false - header_none - secondary_none-diff

The difference in this image is 4 times lower with 0.0406 and it clearly is a regression (I intentionally changed the padding of the close button):
AppSidebar vue - starred_null - compact_false - header_none - secondary_none-diff


So, running the visual tests in the very same environment ensured by a docker container seems like the only option to fix this. If you have another idea, I am open to check it out.

> I think that complicates the setup a lot.

It sure is complicated, but I got the impression, so is visual regression testing in general 🙈

.github/workflows/cypress.yml
raimund-schluessler: I don't think we gain much here. The whole `Cypress` action runs for 4:00 min. The cypress run took 35 s, of which running the actual AppSideber.visual.js test took 10 s. We would need to add a lot of tests to really save time.

.github/workflows/cypress.yml
skjnldsv: Cypress actually have dockerized github actions as well.
We could have use those?

I'm just not fond of having to also maintain a Docker setup there.

skjnldsv: https://github.com/cypress-io/github-action#docker-image

raimund-schluessler: I am fine with using this github action docker setup, but we have to as well locally run cypress within the container. How would you do that then?

PR #1348, opened 2020-08-25
Do not fetch user-status if user prop does not represent a user

### LINKED ISSUE #1342: Load online status only for users
When using the avatar component, it tries to load the online status from users and groups or external users as well. 
This results in a lot of 404s and slows down the avatar loading.

georgehrke:
fixes #1342

### DIFF
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -359,7 +359,7 @@ export default {
 	},
 	mounted() {
 		this.loadAvatarUrl()
-		if (this.user) {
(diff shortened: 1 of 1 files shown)


juliushaertl (approved):

skjnldsv (approved):

dartcafe (commented):
Im am not sure: shouldn't be isGuest checked aswell? In polls we use this for external users to prevent errors on generating the avatars.

dartcafe:
I withdraw the last comment. I changed using of the avatar.

Issue #1345, opened 2020-08-25
[2.6] Header Menu in AppSidebar not positioned correctly


georgehrke:
![34A7FED7-8F4A-405C-B860-7CDDEA044E17](https://user-images.githubusercontent.com/1250540/91163396-cc63d400-e6cd-11ea-8bbe-8f6854d1793a.png)

Tested on latest 2.6

georgehrke:
@nickvergessen This is what you also mentioned in the call

nickvergessen:
Okay, yeah in Talk it's the Actions menu in the participant list

skjnldsv:
Please check again with master.
https://github.com/nextcloud/nextcloud-vue/tree/master/cypress/snapshots/base/components/AppSidebar/AppSidebar.visual.js

raimund-schluessler:
I think this is fixed in master. For Tasks it looks like this now:

![Screenshot_2020-08-27 Tasks - Nextcloud(1)](https://user-images.githubusercontent.com/2496460/91486331-641f1900-e8ac-11ea-9d3e-9d10085f854d.png)

It slightly overlaps the scrollbar, but I think this is alright.

georgehrke:
![18D1D7E9-E11B-42A7-BD09-AE77FD12A23D](https://user-images.githubusercontent.com/1250540/92224537-39335700-eea2-11ea-9bcf-e9882d18dbc1.png)

Fixed in 2.6.4 👍

PR #1351, opened 2020-08-26
Allow to choose boundariesElement for actions' popover

ma12-co:

### DIFF
--- a/src/components/Actions/Actions.vue
+++ b/src/components/Actions/Actions.vue
@@ -117,7 +117,7 @@ https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-action
 			:handle-resize="true"
 			:open.sync="opened"
 			:placement="placement"
-			boundaries-element="body"
+			:boundaries-element="boundariesElement"
 			container="body"
 			@show="openMenu"
 			@apply-show="onOpen"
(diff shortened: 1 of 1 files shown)


nickvergessen (approved):

juliushaertl (approved):

ChristophWurst (approved):

georgehrke:
@ma12-co Does this fix https://github.com/nextcloud/nextcloud-vue/issues/1345 ? :)

nickvergessen:
> Does this fix #1345 ? :)

It allows you do fix it yeah, by specifying the element yourself.

georgehrke:
Well, not really, since you can't set custom props for the actions component for secondary actions:
https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/AppSidebar/AppSidebar.vue#L211

Let me see if adding `.app-sidebar` as a boundary works there.

ma12-co:
> Let me see if adding .app-sidebar as a boundary works there.

It does :)

ma12-co:
But it actually should be an object, not a string :/

ma12-co:
I'm testing now with the dom element body and it seems to work. Seems to be that we were passing in the string "body"

ma12-co:
Can confirm: It works by passing the html body element, without the need of specifying a boundarieselement. I have a problem with the prop validator though. Should I just remove it? 

Screenshot from 2020-08-26 11-03-00

georgehrke:
@ma12-co See https://github.com/Akryum/v-tooltip/blob/fa672f6d6d2004d9cb285ce4551f12c9ca1f6540/src/components/Popover.vue#L114L117

PR #1361, opened 2020-08-28
Fix ActionInput label height

tcitworld:
Noticed in Calendar. See https://nextcloud-vue-components.netlify.app/#/Components/Actions?id=actioninput

image

### DIFF
--- a/src/components/ActionInput/ActionInput.vue
+++ b/src/components/ActionInput/ActionInput.vue
@@ -309,6 +309,7 @@ $input-margin: 4px;
 
 		width: #{$clickable-area - $input-margin * 2};
 		height: #{$clickable-area - $input-margin * 2};
+		box-sizing: border-box;
 		margin: 0 0 0 -8px;
 		padding: 7px 6px;
 


src/components/ActionInput/ActionInput.vue
```diff
@@ -307,8 +307,8 @@ $input-margin: 4px;
 		align-items: center;
 		justify-content: center;
 
-		width: #{$clickable-area - $input-margin * 2};
-		height: #{$clickable-area - $input-margin * 2};
+		width: #{$clickable-area - $input-margin * 6};
+		height: #{$clickable-area - $input-margin * 6};
```
skjnldsv, line 311: But it doesn't make any sense to have 6 here.
The 2 were here to compensate from height/width.

skjnldsv (changes requested):

src/components/ActionInput/ActionInput.vue
raimund-schluessler: `box-sizing: border-box;` might be the better option here.

eneiluj: Works fine with `border-box` indeed.

raimund-schluessler (approved):

src/components/ActionInput/ActionInput.vue
skjnldsv: Iwonder if that wouldn't make sense to apply this to the whole actions component with the `*` selector. I thought I saw a comment about this somewhere, but can't find it.

If we want to be as independent from server's styling as possible, this might be better :thinking:

ma12-co (approved):

src/components/ActionInput/ActionInput.vue
skjnldsv: cc @raimund-schluessler

raimund-schluessler: > Iwonder if that wouldn't make sense to apply this to the whole actions component with the `*` selector. I thought I saw a comment about this somewhere, but can't find it.
> 
> If we want to be as independent from server's styling as possible, this might be better 🤔

I guess this should be ok, since we have this in the server anyway (only for content, but that effects most elements):
```
#content[class*='app-'] * {
 box-sizing:border-box
}
```

But if we do that, we could as well just do it globally for all vue components.

skjnldsv: > But if we do that, we could as well just do it globally for all vue components.

That is what I meant

raimund-schluessler: > 
> 
> > But if we do that, we could as well just do it globally for all vue components.
> 
> That is what I meant

Ah, ok. I thought you meant to only apply it to the Actions component. If we have it globally, we could also get rid of this https://github.com/nextcloud/nextcloud-vue/blob/master/tests/visual/components/AppSidebar/AppSidebar.visual.js#L40-L49 and this https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/Content/Content.vue#L73-L75, I think.

tcitworld:
Rebased and dropped my commit.

src/components/ActionInput/ActionInput.vue
skjnldsv: Sorry, my message wasn't very clear :p
what do you think @tcitworld ?

eneiluj:
Any objection to merge this? I think it's good to get that even if a more general rule is applied in the future.

skjnldsv (approved):

skjnldsv:
Please open a follup issue then :)

MorrisJobke:
Bildschirmfoto 2020-09-11 um 10 49 18


Also present in weather widget of 20.0.0 beta 4

PR #1362, opened 2020-08-28
Fix padding of sidebar without secondary actions

raimund-schluessler:
Fixes the right-padding of the sidebar title for sidebars (in compact mode || without figure) && without secondary actions 😉.

Before:
AppSidebar vue–subtitle_null–starred_false–compact_false–header_none–secondary_none–editable_false-base

After:
AppSidebar vue–subtitle_null–starred_false–compact_false–header_none–secondary_none–editable_false-actual


src/components/AppSidebar/AppSidebar.vue
```diff
@@ -522,6 +527,10 @@ $top-buttons-spacing: 6px;
 			// increase the padding to not overlap the menu
 			.app-sidebar-header__desc {
 				padding-right: #{$clickable-area * 2 + $top-buttons-spacing};
+
+				&.app-sidebar-header__desc--without-actions {
+					padding-right: #{$clickable-area + $top-buttons-spacing};
```
skjnldsv, line 532: Why does without actions have more padding than the default ?
The css doesn't make sense there :thinking: 

Without have 50px, default only have 6px ?

skjnldsv (changes requested):

src/components/AppSidebar/AppSidebar.vue
raimund-schluessler: Well, it's complicated. By default the menu action is part of the flow, so 6px is enough since the action menu itself will push the titles right border to the left. But as soon as you take the menu out of the flow (in case of no figure, or compact mode), the title needs a right-padding. And if there is no menu, the padding must be lower.

Feel free to change the CSS, the new base images in this PR will tell you if the CSS works. 😉

skjnldsv:
PIng? :)

raimund-schluessler:
> 
> 
> PIng? :)

To be honest, I don't really know how to further simplify / clarify the CSS. The visual appearance resulting is ok, as you see from the visual regression tests.

skjnldsv:
No, I'm just super cautious now. :sweat_smile:

raimund-schluessler:
> 
> 
> No, I'm just super cautious now. 😅

I understand that. So should we merge it as is?

skjnldsv (approved):

ma12-co (approved):

PR #1368, opened 2020-09-01
Fix active tab prop

skjnldsv:
Breaks server


raimund-schluessler:
Indeed, another regression by https://github.com/nextcloud/nextcloud-vue/pull/1281 🙈

raimund-schluessler (approved):

eneiluj (approved):
Works fine now. Tested in Cospend.

Issue #1386, opened 2020-09-04
Stop propagation of "click" event in Actions


julien-nc:
If Actions is inside a link (`<a>`), clicking on the menu trigger also triggers a click on the link.

This is the case in vue-dashboard's widget item component. I made a [dirty fix](https://github.com/nextcloud/nextcloud-vue-dashboard/blob/daaf9657678ca8ab4243f40793fa07ec22625bf3/src/DashboardWidgetItem.vue#L248) which prevents the click event on the link.

I quickly checked how this is managed in Actions and Popover components and didn't find a simple way to stop the click propagation. Any idea?

Btw, stopping the click propagation could be activated with an optional prop in Actions for example.

PR #1387, opened 2020-09-04
[WIP] In Action, stop propagation of click event coming from popover

julien-nc:
This is one way to solve #1386. 

@skjnldsv @raimund-schluessler @ma12-co Does it feel right?

This only works when there are multiple actions. Any clue to do it when there is just once?

I can't find a way to directly do that in the Popover component. Do we even want to do that?

### DIFF
--- a/src/components/Actions/Actions.vue
+++ b/src/components/Actions/Actions.vue
@@ -110,7 +110,8 @@ https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-action
 	<div v-else
 		v-show="hasMultipleActions || forceMenu"
 		:class="{'action-item--open': opened}"
-		class="action-item">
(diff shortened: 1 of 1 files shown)


ma12-co:
I honestly don't know what would be the best approach here... @skjnldsv ??

skjnldsv:
> If Actions is inside a link (<a>)

This is your issue, Actions cannot be inside a link. This is not compliant to html standards :100:

PVince81:
the Actions component got [rewritten recently](https://github.com/nextcloud/nextcloud-vue/pull/2911)

is this issue/fix still relevant ?

marcoambrosini (approved):

julien-nc:
Not relevant anymore.

PR #1405, opened 2020-09-15
Listen to user status events to update the status on avatars

nickvergessen:
### Steps
1. Go to any page that uses the Avatar component to show your own avatar, e.g. a Talk conversation participant tab
2. Change your status
3. Wait until your avatar shows the new status :skull_and_crossbones: 

Relevant event is emitted in server:
https://github.com/nextcloud/server/pull/22801

I kept the event neutral from `getCurrentUser` so it will automatically update statuses of other users when e.g. Talk emits their status updates in the future (since it fetches it with the 30s participant list update it makes sense)


skjnldsv (approved):

ma12-co (approved):
Works!

Issue #1406, opened 2020-09-16
Avatar initial don't support unicode


StCyr:
Reported here: https://github.com/nextcloud/mail/issues/3366

I believe it's because:

https://github.com/nextcloud/nextcloud-vue/blob/469b2aaf5b94bcbdba6690fb149a71c9bc87353e/src/components/Avatar/Avatar.vue#L318

Should be:

```js
return String.fromCharCode(this.getUserIdentifier.charCodeAt(0)).toUpperCase()
```

PR #1407, opened 2020-09-16
Adds supports for unicode-encoded names when creating an avatar from a name's initial

StCyr:
Fix to https://github.com/nextcloud/nextcloud-vue/issues/1406

not tested though :-)

Signed-off-by: Cyrille Bollu <cyrpub@bollu.be>

### DIFF
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -314,10 +314,20 @@ export default {
 			return this.displayName
 		},
 		initials() {
+			let initials
 			if (this.shouldShowPlaceholder) {
-				return this.getUserIdentifier.charAt(0).toUpperCase()
+				const user = this.getUserIdentifier
+				const idx = user.indexOf(' ')
+				if (user === '') {
+					initials = '?'
+				} else {
+					initials = String.fromCodePoint(user.codePointAt(0))
+					if (idx !== -1) {
+						initials = initials.concat(String.fromCodePoint(user.codePointAt(idx + 1)))
+					}
+				}
 			}
-			return '?'
+			return initials.toUpperCase()
 		},
 		menu() {
 			const actions = this.contactsMenuActions.map((item) => {


skjnldsv (approved):

ChristophWurst (approved):

ChristophWurst (changes requested):
Bildschirmfoto von 2020-10-13 11-12-38

Does not seem to work as expected?

ChristophWurst:
Use this 
Bildschirmfoto von 2020-10-13 11-15-43

Ref https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCharCode and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charCodeAt

StCyr:
> Use this
> Bildschirmfoto von 2020-10-13 11-15-43
> 
> Ref https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCharCode and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charCodeAt
> 
> And it also work with simple names
> 
> Bildschirmfoto von 2020-10-13 11-16-29

Thanks! I'll update my PR (sorry for the wrong solution I've proposed)

skjnldsv:
Btw shouldn't we use the double initials for the displayname?

StCyr:
> Btw shouldn't we use the double initials for the displayname?

I also wonder.

I think it was the case earlier, and then, at some point in time, NC has started to display only 1 initial...

skjnldsv:
> I think it was the case earlier, and then, at some point in time, NC has started to display only 1 initial...

It's the opposite, we showed only one, but now show two :wink: 
https://github.com/nextcloud/server/pull/18745

StCyr:
> > I think it was the case earlier, and then, at some point in time, NC has started to display only 1 initial...
> 
> It's the opposite, we showed only one, but now show two 😉
> [nextcloud/server#18745](https://github.com/nextcloud/server/pull/18745)
> https://github.com/nextcloud/server/blob/caff1023ea72bb2ea94130e18a2a6e2ccf819e5f/lib/private/Avatar/Avatar.php#L98-L100

Ha! nice 👍 

I can push this change here if you want, that would be great I think

skjnldsv:
> I can push this change here if you want, that would be great I think

Sure, it feels relevant enough! :+1:

src/components/Avatar/Avatar.vue
```diff
@@ -314,10 +314,20 @@ export default {
 			return this.displayName
 		},
 		initials() {
+			let initials
 			if (this.shouldShowPlaceholder) {
-				return this.getUserIdentifier.charAt(0).toUpperCase()
+				const userId = this.getUserIdentifier
```
skjnldsv, line 319: please use something else than `userId`, this is also the displayName :)

skjnldsv (approved):

ChristophWurst (approved):

PR #1422, opened 2020-09-30
Show icon instead of status if not in do not disturbe mode

juliushaertl:
image

How it looks in talk ([note that the duplicate icon needs a fixing there](https://github.com/nextcloud/spreed/pull/4253)):
image

I played around a bit with a circled white background but this might look odd depending on the emoji. 

@jancborchardt Feel free to play around with the styles if needed

### DIFF
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -54,7 +54,10 @@
 		<div v-if="hasMenu" class="icon-more" />
 
 		<!-- avatar status -->
-		<div v-if="canDisplayUserStatus"
+		<div v-if="userStatus.icon && userStatus.status !== 'dnd'" class="avatardiv__user-status avatardiv__user-status--icon">
+			{{ userStatus.icon }}
+		</div>
+		<div v-else-if="canDisplayUserStatus"
 			class="avatardiv__user-status"
 			:class="'avatardiv__user-status--' + userStatus.status" />
 		<div v-else-if="status"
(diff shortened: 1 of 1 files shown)


jancborchardt:
Nice work! :) 

@karlitschek what do you think? My only concern is that we now don’t show the online/away status for people with a custom status, but still for others – but that might just be fine.

nickvergessen:
Talk fix is in https://github.com/nextcloud/spreed/pull/4253

Ivansss:
I don't want to sound wrong but in my opinion this adds more confusion between "user status" and "status message".
I think this wording change https://github.com/nextcloud/server/pull/23113 was good to separate these two concepts.
But with this PR I think we are again mixing both concepts which could lead to a confusion.
I was not involved in the discussion so maybe I'm missing something. This is just a personal opinion :)

karlitschek:
The user set status is more important then 'online'. Again online doesn't tell me anything beside that someone used Nextcloud in the last 10min.

rullzer (approved):

karlitschek:
I don't get the special logic with 'do not disturb' here

jancborchardt:
> I don't get the special logic with 'do not disturb' here

The idea is that when people set "do not disturb" cause they are in a meeting or something else, that is more important to signify than the emoji of the general status.

We can also cut that at just overwrite "Do not disturb" as well – your call.

jancborchardt:
Guess needs a new release?

PR #1429, opened 2020-10-04
Add loading state on AppSidebar and remove unwanted slot bind

skjnldsv:
- Added loading prop to AppSidebar
- Removed non-working slot v-bind
- Moved active show/hide to the css instead of relying on v-show to avoid conflicts

### DIFF
--- a/src/components/AppSidebarTab/AppSidebarTab.vue
+++ b/src/components/AppSidebarTab/AppSidebarTab.vue
@@ -24,10 +24,12 @@
 	https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html -->
 
 <template>
-	<section v-show="isActive"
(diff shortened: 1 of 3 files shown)


src/components/AppSidebarTab/AppSidebarTab.vue
```diff
@@ -67,7 +69,8 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-section {
+.app-sidebar__tab {
+	display: none;
```
skjnldsv, line 73: By moving this to the css, we allow to use v-show on the AppSidebarTab as well, otherwise it was conflicting as we'd have two handlers, from the inside and from the outside.

ChristophWurst (approved):

juliushaertl (approved):

PR #1437, opened 2020-10-07
modal-container to overflow: auto

dartcafe:
fix https://github.com/nextcloud/polls/issues/1065

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -675,7 +675,7 @@ $header-size: 50px;
 	/* Content */
 	.modal-container {
 		display: block;
-		overflow: hidden;
+		overflow: auto;
 		padding: 0;
 		transition: transform 300ms ease;
 		border-radius: var(--border-radius-large);


skjnldsv (approved):

skjnldsv:
Still ok on Viewer! :+1:

juliushaertl:
The main issue i see with this is that the rounded corners will no longer be in place when scrollbar is shown:

Chromium:
image

Firefox:
image

Not sure how to solve that properly with the modal size adjusting to the content actually.

skjnldsv:
Ah, there it is. Now I remember why it were there ^^

ma12-co (commented):
I would add a div with `overflow: auto` within the modal itself and add a padding in the modal's div. It's what I've been doing in all talk modals.

PVince81:
not sure if related, for the app settings we had to use display: flex to solve scrolling issues, see https://github.com/nextcloud/nextcloud-vue/pull/1573/files#diff-b1113f0551fa6d858b014056d04440a7e6186aabd61fb30c592c11748ee6d463R273

@dartcafe ^

szaimen:
Superseded by https://github.com/nextcloud/nextcloud-vue/pull/2126

PR #1438, opened 2020-10-07
Fix suggestion for RichContenteditable component

eneiluj:
Bubble icon and autocomplete icons are always overridden by the background-image, even if there is no avatar URL (because it's not a user or there is no id).

This is an attempt to fix that :grin:.

refs #1433


skjnldsv (approved):
Nice catch!

PR #1456, opened 2020-10-12
Fix npm run build by removing obsolete command

PVince81:
Remove hide-modules from webpack production command

Without this fix, `npm run build` fails with an error as this argument was removed in webpack 4 and https://github.com/nextcloud/nextcloud-vue/pull/1452 just upgraded to that version


juliushaertl (approved):

juliushaertl:
Strange that CI did have the failure on the command but didn't fail: https://github.com/nextcloud/nextcloud-vue/runs/1239859214

ChristophWurst (approved):

PR #1457, opened 2020-10-12
Cache hasAvatar info in sessionStorage to avoid flicker

PVince81:
This is a crude POC that shows that when we save the "does this user has an avatar call", performance improves.

Possible solution for https://github.com/nextcloud/spreed/issues/4284

Maybe this should be stored in a store or something, but as I understand, Vue components in the library don't really have a store available.

Another approach would be to supply an optional property "hasAvatar" and the caching could be done from the outside in situations where the caller thinks it makes sense.

### DIFF
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -100,6 +100,18 @@ import Tooltip from '../../directives/Tooltip'
 import usernameToColor from '../../functions/usernameToColor'
 import { userStatus } from '../../mixins'
 
+function getUserHasAvatar(userId) {
+	const flag = window.sessionStorage.getItem('userHasAvatar-' + userId)
+	if (typeof flag === 'string') {
+		return Boolean(flag)
+	}
+	return null
+}
+
+function setUserHasAvatar(userId, flag) {
+	window.sessionStorage.setItem('userHasAvatar-' + userId, flag)
+}
+
 export default {
 	name: 'Avatar',
 	directives: {
@@ -466,6 +478,20 @@ export default {
 				urlGenerator(this.user, this.size * 4) + ' 4x',
 			].join(', ')
 
+			// skip loading
+			const userHasAvatar = getUserHasAvatar(this.user)
+			if (typeof userHasAvatar === 'boolean') {
+				this.isAvatarLoaded = true
+				this.avatarUrlLoaded = avatarUrl
+				if (!this.isUrlDefined) {
+					this.avatarSrcSetLoaded = srcset
+				}
+				if (userHasAvatar === false) {
+					this.userDoesNotExist = true
+				}
+				return
+			}
+
 			const img = new Image()
 			img.onload = () => {
 				this.avatarUrlLoaded = avatarUrl
@@ -473,10 +499,13 @@ export default {
 					this.avatarSrcSetLoaded = srcset
 				}
 				this.isAvatarLoaded = true
+				// re-get to avoid concurrent access
+				setUserHasAvatar(this.user, true)
 			}
 			img.onerror = () => {
 				this.userDoesNotExist = true
 				this.isAvatarLoaded = true
+				setUserHasAvatar(this.user, false)
 			}
 
 			if (!this.isUrlDefined) {
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -90,6 +90,7 @@
 </template>
 
 <script>
+import { getBuilder } from '@nextcloud/browser-storage'
 import { directive as ClickOutside } from 'v-click-outside'
 import PopoverMenu from '../PopoverMenu'
 import { getCurrentUser } from '@nextcloud/auth'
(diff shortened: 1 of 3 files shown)


PVince81:
Any thoughts on this approach ? @ma12-co @nickvergessen @skjnldsv 

If we are fine caching this information, where would be a better place ? Ideally that info should be available to other components as well like the VideoBackground in the Talk app which uses another separate query to get the avatar to put into the background in blur mode.

src/components/Avatar/Avatar.vue
```diff
@@ -466,17 +466,35 @@ export default {
 				urlGenerator(this.user, this.size * 4) + ' 4x',
 			].join(', ')
 
+			// skip loading
+			if (!window.userHasAvatar) {
```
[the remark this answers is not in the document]
skjnldsv: Yes, we are cleaning the window over the last nc releases, we should try our best not to pollute the window namespace anymore :)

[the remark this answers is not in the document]
PVince81: if we were to store information available for all apps, where would that be ?

in the past it used to be in the OC namespace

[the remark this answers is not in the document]
skjnldsv: browserstorage is good enough I think, or localstorage if you want :shrug:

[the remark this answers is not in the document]
PVince81: ok sounds good.

will likely also need some form of expiration, but I'm not sure if worth it as the use case would be that someone deletes their avatar and do not set a new one, which is rather rare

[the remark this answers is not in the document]
PVince81: I think I'll go with sessionStorage. It only lives within the current page and should be enough for the use case: preventing flickering within a call session.

Then it's easier to tell people to force refresh the page than clear cache whenever they noticed that an avatar of a collegue did not get updated.

PVince81:
Ready to review.

Note: I went for the format "userHasAvatar-" + userId instead of a single object to avoid repeated serialization/deserialization (also semi-concurrently)

PVince81:
Weird, I just checked the browser console and somehow nothing appears under sessionStorage.

and also somehow I didn't see flickering any more. I don't understand how it can be working.

investigating...

PVince81:
ok got it... `null` is returned by the sessionStorage instead of `undefined`, which the current logic treats as always `true`... so no flickering...

src/components/Avatar/Avatar.vue
```diff
@@ -100,6 +100,14 @@ import Tooltip from '../../directives/Tooltip'
 import usernameToColor from '../../functions/usernameToColor'
 import { userStatus } from '../../mixins'
 
+function getUserHasAvatar(userId) {
+	return window.sessionStorage.getItem('userHasAvatar-' + userId)
```
juliushaertl, line 104: We could probably apply a similar naming schema as we discusssed for the event names in https://github.com/nextcloud/nextcloud-event-bus/issues/131#issuecomment-683689107 That way we still can figure out where a value might be coming from when we fill out browser/session storage more over time.

@skjnldsv @ChristophWurst Didn't we have some planned prefix for the components library as well?

Basically someting like `vue-components:user-has-avatar.userID`

skjnldsv: Nah, components were generic, so `nextcloud:user-has-avatar.userID` iirc :thinking:

PVince81:
works nicely with Chrome, but in Firefox I still observe a flicker sometimes when hitting "stop following", despite cache hit.

let's still move forward with this as it at least helps resolve the annoying flicker on speaker change, which is already a big step forward

skjnldsv (changes requested):
See https://github.com/nextcloud/nextcloud-vue/pull/1457#discussion_r505552471

src/components/Avatar/Avatar.vue
ChristophWurst: https://github.com/nextcloud/nextcloud-browser-storage :thinking:

PVince81: ah, that's the global thing I was looking for, thanks :-)

I'll adjust

PVince81:
adjusted in https://github.com/nextcloud/nextcloud-vue/pull/1457/commits/7613d44ce026fe88502206279abf130e26e50143

I decided against introduce a service for just this once single occurrence, considering that there are no services in nextcloud-vue. If this becomes commonplace later on we could adjust.

Please also check https://github.com/nextcloud/spreed/pull/4379 (last commit) which uses the same storage location for the video background.

src/components/Avatar/Avatar.vue
```diff
@@ -100,6 +101,20 @@ import Tooltip from '../../directives/Tooltip'
 import usernameToColor from '../../functions/usernameToColor'
 import { userStatus } from '../../mixins'
 
+const browserStorage = getBuilder('nextcloud').persist().build()
```
ChristophWurst, line 104: but now it will use `localStorage` and not `sessionStorage` – intended?

PVince81: that's fine. will increase performance for an even longer time then.

PVince81:
good to merge ? :smile:

ChristophWurst (approved):

skjnldsv (approved):

PR #1462, opened 2020-10-13
l10n: Change the name of the label

Valdnet:


juliushaertl:
I'm not sure, but `Settings navigation` seems correct to me.

ma12-co (changes requested):
I also think it's correct, as it's the "navigation of the settings". To be 100% correct it should be "Settings' navigation" though, could you implement that? :)

PR #1468, opened 2020-10-14
Add button component

marcoambrosini:
Use https://github.com/nextcloud/spreed/pull/5114 to test

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29947,6 +29947,14 @@
 			"resolved": "https://registry.npmjs.org/vue-resize/-/vue-resize-0.4.5.tgz",
 			"integrity": "sha512-bhP7MlgJQ8TIkZJXAfDf78uJO+mEI3CaLABLjv0WNzr4CcGRGPIAItyWYnP6LsPA4Oq0WE+suidNs6dgpO4RHg=="
 		},
+		"vue-ripple-directive": {
+			"version": "2.0.1",
+			"resolved": "https://registry.npmjs.org/vue-ripple-directive/-/vue-ripple-directive-2.0.1.tgz",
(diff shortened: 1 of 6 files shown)


ma12-co:
current state of things 

https://user-images.githubusercontent.com/26852655/107407330-719a2980-6b01-11eb-82b9-0d74e27e3ac6.mp4

TODO
- [ ] add error success and warning states to contained button;
- [ ] Write documentation;

jancborchardt:
Very nice! :) Some considerations:
- Within this set, now the "Outlined button" which we originally thought about looks quite misplaced. Is this actually something that we need?
- The secondary contained button looks very dull and gray, an issue that we often have. Maybe we better use one with the 10% primary background color and full primary color as text color?

And it would be great to see some live examples of the component and how it would be used. E.g. how these buttons would look like in the Files app, Talk, Calendar etc.

ma12-co:
> The secondary contained button looks very dull and gray, an issue that we often have. Maybe we better use one with the 10% primary background color and full primary color as text color?

We need that button in the the talk call ui. In other places we can use an outlined and a contained primary? I've been mostly following material design guidelines and that's done quite often. Those types of buttons are meant to be / and should be mixed.

Screenshot from 2021-02-22 13-44-14

>And it would be great to see some live examples of the component and how it would be used. E.g. how these buttons would look like in the Files app, Talk, Calendar etc.

Should I stop here and do some more mock-ups then?

ma12-co:
Added ripple effect on touch and click, still have to tweak the colors a bit:

https://user-images.githubusercontent.com/26852655/108720611-f1b88a00-7518-11eb-9778-ccf0b8ec4fe0.mp4

ma12-co:
From @jancborchardt :

I'd say 3 main ones are enough:
1. Primary filled with white text (or dark if the primary is light)
2. Some sort of secondary, could be: Grey outline with primary color text, or primary 10% bg and 100% text.
3. Tertiary: text/icon only. In Material the text-only is primary, but for icons we definitely need the default to be grey (but could also be colored, or e.g. white like in talk)

PR #1474, opened 2020-10-15
Add emoji autocompletion

skjnldsv:
image

### DIFF
--- a/src/components/RichContenteditable/RichContenteditable.vue
+++ b/src/components/RichContenteditable/RichContenteditable.vue
@@ -151,7 +151,7 @@ import stringLength from 'string-length'
 import { t } from '../../l10n'
 import AutoCompleteResult from './AutoCompleteResult'
 import richEditor from '../../mixins/richEditor/index'
-import { allEmojis, emojiSearch } from '../../functions/emoji/index'
+import { emojiSearch, addRecent } from '../../functions/emoji'
 
 export default {
 	name: 'RichContenteditable',
@@ -242,13 +242,16 @@ export default {
 				// Where to inject the menu popup
 				menuContainer: this.menuContainer,
 				// Popup mention autocompletion templates
-				menuItemTemplate: item => `${item.original.value} :${item.original.short_name}`,
+				menuItemTemplate: item => `${item.original.native} :${item.original.short_name}`,
 				// Hide if no results
-				noMatchTemplate: () => t('No emojis found'),
+				noMatchTemplate: () => t('No emoji found'),
 				// Display raw emoji along with its name
-				selectTemplate: item => item.original.value,
+				selectTemplate: (item) => {
+					addRecent(item.original)
+					return item.original.native
+				},
 				// Pass the search results as values
-				values: (text, cb) => cb(emojiSearch(allEmojis, text)),
+				values: (text, cb) => cb(emojiSearch(text)),
 				// Class added to the menu container
 				containerClass: 'tribute-container-emoji',
 				// Class added to each list item
--- a/src/components/RichContenteditable/RichContenteditable.vue
+++ b/src/components/RichContenteditable/RichContenteditable.vue
@@ -151,6 +151,7 @@ import stringLength from 'string-length'
 import { t } from '../../l10n'
 import AutoCompleteResult from './AutoCompleteResult'
 import richEditor from '../../mixins/richEditor/index'
+import { allEmojis, emojiSearch } from '../../functions/emoji/index'
 
 export default {
 	name: 'RichContenteditable',
@@ -235,18 +236,23 @@ export default {
 			},
 			emojiOptions: {
 				trigger: ':',
-				// Search against id and label (display name)
-				lookup: result => result.short_names.join(' '),
+				// Don't use the tribute search function at all
+				// We pass search results as values (see below)
+				lookup: (result, query) => query,
 				// Where to inject the menu popup
 				menuContainer: this.menuContainer,
 				// Popup mention autocompletion templates
-				menuItemTemplate: item => item.original.value,
+				menuItemTemplate: item => `${item.original.value} :${item.original.short_name}`,
 				// Hide if no results
-				noMatchTemplate: () => '<span class="hidden"></span>',
-				// Inner display of mentions, display raw emoji
+				noMatchTemplate: () => t('No emojis found'),
+				// Display raw emoji along with its name
 				selectTemplate: item => item.original.value,
-				// Autocompletion results
-				values: EMOJIS,
+				// Pass the search results as values
+				values: (text, cb) => cb(emojiSearch(allEmojis, text)),
+				// Class added to the menu container
+				containerClass: 'tribute-container-emoji',
+				// Class added to each list item
+				itemClass: 'tribute-container-emoji__item',
 			},
 
 			// Represent the raw untrimmed text of the contenteditable
@@ -569,7 +575,7 @@ export default {
 </style>
 
 <style lang="scss">
-.tribute-container {
+.tribute-container, .tribute-container-emoji {
 	z-index: 9000;
 	overflow: auto;
 	min-width: 250px;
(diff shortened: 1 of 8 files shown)


PVince81:
good luck with insertion point, see https://github.com/nextcloud/spreed/issues/4265 :wink:

skjnldsv:
> good luck with insertion point, see [nextcloud/spreed#4265](https://github.com/nextcloud/spreed/issues/4265)

Looks like my appaorch actually fixes this directly.
But this is autocmpletion, different from the emoji picker. It's always inserted at the correct position :)

PVince81:
what's missing ?

skjnldsv:
> what's missing ?

image

skjnldsv:
Rebased

skjnldsv:
> what's missing ?

Joke aside, it's working, just need to be polished for the ui (not show the whole list, maybe show the shortcut? This is also not translated. Anyone can take over, this is mostly for talk, so I'll let you folks decide what needs to be done. cc @jancborchardt 

You can all test it in the deploy preview https://deploy-preview-1474--nextcloud-vue-components.netlify.app/

simonspa:
Concerning the UI, this is how Mattermost looks like, something I find very well working for emoji autocompletion:

image
image

A list of matching emojis, with a scrollbar if too many, and the emoji "name" / code listed behind the emoji.

skjnldsv:
@jancborchardt are you ok with this design (see comment above) ?

PVince81:
styles seems broken currently, might need a rebase first:
image

also, when typing just ":s" it's very slow/sluggish in my Firefox until I type more letters.
probably this is because it renders too many results until it gets narrowed down, maybe later need to add some virtual scrolling there

PVince81:
- [ ] BUG? doesn't autocomplete / replace ascii emojis like ":-)" (could be something for later or needs a DB extension)
- [ ] BUG: hitting enter key after typing ":-)" doesn't go to the next line, but after other words it does

skjnldsv:
Please take over @PVince81 :)

mejo-:
I put some time into this PR and think that it's in a state to be reviewed now :blush:

Instead of building another own emoji data set, the existing dataset by emoji-mart-fast is used. Same for emoji lookup/search functions. This brings unified dataset, defaults and recently used logic between EmojiPicker and emoji autocompletion.

The number of displayed results is limited to five (Github etc. do the same).

I also went with implementing the `emojiSearch` and `addRecent` in a seperate functions file, so it can be used again by others (e.g. the Text app, see https://github.com/nextcloud/text/pull/1961).

That's how it looks like at the moment:

recording

Looking forward to your thoughts @skjnldsv @PVince81 @jancborchardt @nimishavijay :blush:

mejo-:
Here's some more screenshots for the design/layout review:

2021-12-02T17:22:24,279207414+01:00
2021-12-02T17:22:43,980324362+01:00
2021-12-02T17:22:57,587558848+01:00
2021-12-02T17:23:26,187049977+01:00

nimishavijay:
Looks great actually! I only have a couple of points:
- How will this work with other languages? 
- Purely visual nitpicking, but a couple of pixels of space between the emoji and the text might be a good idea 
- I'm also thinking if it is a good idea to have a scrollbar instead of a hard limit of 5 results, because it might make it easier to find results that are buried down. 
For example, I call this ✅  a tick mark, so if I type `:white_tick` then ✅ shows up because it is the top result. But when I type `:tick` the emoji doesn't show up even though it is matched, because it is below the top 5 matched results. Having it scrollable would allow me to find more results.
Github limits the results to only 5, so I'm thinking if it is such a big deal after all 🤔 

What do you think? cc @jancborchardt also

mejo-:
>     * How will this work with other languages?

Unfortunately there's no support for translated emoji names in emoji-mart-vue yet. I'd suggest to discuss this in a separate follow-up issue/PR.

>     * Purely visual nitpicking, but a couple of pixels of space between the emoji and the text might be a good idea

Agreed. I now added `padding-right: 8px` to the emoji in the dropdown. See the updated screenshots: https://github.com/nextcloud/nextcloud-vue/pull/1474#issuecomment-984785282

>     * I'm also thinking if it is a good idea to have a scrollbar instead of a hard limit of 5 results, because it might make it easier to find results that are buried down.
>       For example, I call this white_check_mark  a tick mark, so if I type `:white_tick` then white_check_mark shows up because it is the top result. But when I type `:tick` the emoji doesn't show up even though it is matched, because it is below the top 5 matched results. Having it scrollable would allow me to find more results.
>       Github limits the results to only 5, so I'm thinking if it is such a big deal after all thinking

Not sure about this. We definitely need a limit here, otherwise we would list more than 3000 emojis without a query string. I like the short list of five results and that's what I discovered at Github and Gitlab. Element (the Matrix client) has a limit of 20 emojis (with a scrollbar), Rocketchat lists 10 emojis (with a scrollbar). Notions lists *a lot*, but just the emojis (without their name) in a table.

Here's how emoji autocompletion would look with a limit of 20 results:

recording

Let me know, which one you prefer :blush:

marcoambrosini (commented):
I'd also like to have the ability to scroll and In my opinion 10 results are enough. 

Also, we could round the results focus feedback a bit more (from 5 to 8px) and increase the padding of the container to 6 or 8px (screenshot is 8)
Screenshot 2021-12-28 at 10 35 30

marcoambrosini (approved):
But since this is already super awesome I'm just going to approve, these comments are just tiny details :)

mejo-:
> I'd also like to have the ability to scroll and In my opinion 10 results are enough.

Ok, I'll wait for one or two more opinions and raise the limit to 10 if nobody opposed :blush: 

> Also, we could round the results focus feedback a bit more (from 5 to 8px) and increase the padding of the container to 6 or 8px (screenshot is 8) Screenshot 2021-12-28 at 10 35 30

Agreed! I don't mind to quickly change this if others agree as well :blush:

jancborchardt:
Looks very nice! :)

- Agree with @marcoambrosini’s feedback on both limit and padding
- When the container is scrollable, the height should cut off at the _half_ of an entry rather than at the full entry cause that communicates better that there are more entries. :)
- Currently the width of the container changes based on the length of the text inside – we should make it a set width always, then it jumps around less.

mejo-:
>     * Currently the width of the container changes based on the length of the text inside – we should make it a set width always, then it jumps around less.

Good point :blush: What would a good width be? Currently it has `min-width: 180px` and `max-width: 300px`. Changing to `min-width: 300px` looks a bit too wide, especially on mobile and when searching for short names:

2021-12-28T14:51:20,252681508+01:00

Maybe `250px` is a good fixed width:

2021-12-28T14:53:06,647980626+01:00

mejo-:
Ok, did all the requested changes:

* [x] Limit to 10 results
* [x] `border-radius: 8px`
* [x] Fixed width of `300px` for container
* [x] `padding: 8px` for container
* [x] Show 5 entries and a half to show scrollbar

Screenshots (and gif) updated: https://github.com/nextcloud/nextcloud-vue/pull/1474#issuecomment-984785282

jancborchardt:
Very nice! From the gif the width seems a bit mich for most of the emojis, if it's 200px (and ellipsized text) then all good! 👍

mejo-:
> Very nice! From the gif the width seems a bit mich for most of the emojis, if it's 200px (and ellipsized text) then all good! +1

Ok, done: https://github.com/nextcloud/nextcloud-vue/pull/1474#issuecomment-984785282

jancborchardt (approved):
Nice! :)

src/components/RichContenteditable/RichContenteditable.vue
```diff
@@ -36,7 +36,7 @@ Note you need to register the [tooltip directive](https://nextcloud-vue-componen
 			:auto-complete="autoComplete"
 			:maxlength="100"
 			:user-data="userData"
-			placeholder="Try mentioning the user Test01"
+			placeholder="Try mentioning user @Test01 or inserting emoji :smile"
```
skjnldsv, line 39:
```suggestion
			placeholder="Try mentioning user @Test01 or inserting emoji :smile:"
```

Typo?

src/components/RichContenteditable/RichContenteditable.vue
```diff
@@ -46,7 +46,7 @@ Note you need to register the [tooltip directive](https://nextcloud-vue-componen
 			:maxlength="400"
 			:multiline="true"
 			:user-data="userData"
-			placeholder="Try mentioning the user Test01"
+			placeholder="Try mentioning user @Test01 or inserting emoji :smile"
```
skjnldsv, line 49:
```suggestion
			placeholder="Try mentioning user @Test01 or inserting emoji :smile:"
```
Typo?

skjnldsv (commented):
So, autocompletion is one thing, but we're not converting the data in emoji when rendering.
If I type the text `:smile:`, there will be no emoji displayed, right?
I guess that would be a nice second step for the RichEdior :)

src/components/RichContenteditable/RichContenteditable.vue
mejo-: I decided not to append a colon on purpose. The trailing colon is not needed for autocompletion and makes it harder to read in my opinion. It's more that the prepending colon is the trigger for autocompletion, and then you type your search query (e.g. "smile").

I agree that this might change when we implement automatic conversion of `:smile:` to an emoji, but I would say that's out of scope here :blush:

PR #1481, opened 2020-10-16
Fix autocomplete popup size and shadow

skjnldsv:
Per Jan's request
Capture d’écran_2020-10-16_10-26-28

Fix shadow and number of results showed


jancborchardt (approved):
Looks good

ma12-co (approved):

PR #1485, opened 2020-10-16
Make appSidebar scrollable inside

skjnldsv:
Joas told me recently they had to rely on hacks to make the sidebar inner scrollable.
The AppSidebar needs to be as simple as possible, so I made the inner tabs scrollable by default and integrated two new events:
- `scroll`, the tab is being scrolled
- `bottomReached`, the tab is scrolled to the bottom

I also fixed some variables on the css

### DIFF
--- a/src/components/AppSidebar/AppSidebar.vue
+++ b/src/components/AppSidebar/AppSidebar.vue
@@ -444,14 +444,17 @@ export default {
 }
 </script>
 <style lang="scss" scoped>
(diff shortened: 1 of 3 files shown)


raimund-schluessler:
I will only have time to look at it on Tuesday.

ma12-co (approved):
Tested on talk and doesn't break anything!

PVince81 (approved):
:+1:

PR #1491, opened 2020-10-20
Fix avatar caching for non-user rendering

skjnldsv:
- Do not cache if the user is not defined
- Added debug level
- Fallback to letters if the url is invalid

```vue
<Avatar
	:disable-menu="true"
	:disable-tooltip="true"
	display-name="Hello User"
	:is-no-user="true"
	:size="40"
	url="http://this.is.a.fakeUrl.jpg"
	class="app-content-list-item-icon" />

```

### DIFF
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -112,7 +112,9 @@ function getUserHasAvatar(userId) {
 }
 
 function setUserHasAvatar(userId, flag) {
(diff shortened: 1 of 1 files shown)


PVince81:
is that for guest users or another scenario ? (steps to reproduce would be nice for verification)

ChristophWurst (approved):

PVince81 (approved):
:+1: tested, makes sense

PR #1503, opened 2020-10-21
Fix setting navigation item blur

ma12-co:
Solves the first part of this comment: https://github.com/nextcloud/spreed/pull/4195#pullrequestreview-511633941

### DIFF
--- a/src/components/AppSettingsDialog/AppSettingsDialog.vue
+++ b/src/components/AppSettingsDialog/AppSettingsDialog.vue
@@ -115,7 +115,7 @@ export default {
 	updated() {
 		if (!this.scroller) {
 			// Get the scroller element
-			this.scroller = this.$el.querySelector('.app-settings__content')
+			this.scroller = this.$refs.settingsScroller
 		}
 		if (this.scroller && !this.addedScrollListener) {
 			this.scroller.addEventListener('scroll', this.handleScroll)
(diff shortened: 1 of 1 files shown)


src/components/AppSettingsDialog/AppSettingsDialog.vue
```diff
@@ -167,16 +167,18 @@ export default {
...
-					document.activeElement.blur()
-				}, 300)
+				this.blur()
 			}
 		},
 
+		// Remove selected section once the user starts scrolling
+		blur: debounce(function() {
+			this.selectedSection = ''
+			document.activeElement.blur()
```
nickvergessen, line 179: So when scrolling and clicking (or tabbing through items) this will unfocus the input after "random" 300ms
Sounds unexpected from a user perspective

PVince81 (approved):
Cleaner code :+1:

src/components/AppSettingsDialog/AppSettingsDialog.vue
ma12-co: :ok_hand:

skjnldsv (approved):

danxuliu (commented):
There seems to be an issue with the blur logic:

- Open the [AppSettingsDialog component documentation](https://nextcloud-vue-components.netlify.app/#/Components/AppSettingsDialog)
- Add tens of `<div>Some example content</div>` to the two `AppSettingsSection` (so the content is too large and a vertical scroll bar is shown once the AppSettingsDialog is opened)
- Show the settings
- Click on the second header and then quickly click on the first header

The dialog will scroll to the second header and then to the first header. However, the first header is blurred; if instead of quickly clicking the headers you wait a little then the first header will remain highlighted after the scrolling ends.

My quick guess (not verified) by looking at the code is that after the second click is done [the timer to restore `linkClicked` to `false`](https://github.com/nextcloud/nextcloud-vue/blob/fb2eeb4144111f01c6831ec5e579a4cbf3efdc98/src/components/AppSettingsDialog/AppSettingsDialog.vue#L161-L163) ends; as this happens while the scrolling is still on-going [the header is blurred](https://github.com/nextcloud/nextcloud-vue/blob/fb2eeb4144111f01c6831ec5e579a4cbf3efdc98/src/components/AppSettingsDialog/AppSettingsDialog.vue#L171-L173).

It would be nice too if, instead of blurring the headers when scrolling, the highlighted header was kept in sync with the first visible section. Although in that case maybe the headers should not be clickable if there is no scroll bar in the content... Anyway, I am just thinking out loud :-P

PR #1517, opened 2020-10-26
Adapt to new vue2-datepicker format props

eneiluj:
We are getting vue2-datepicker 3.7.0 and since 3.6.3, the `format` prop not a string/object anymore. It's now just a string and there is a new `formatter` object prop.

As there is no default for this prop in vue2-datepicker, I wonder if it's better to have `{}` or `null` as default in `@nc/vue`

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
 		"vue-color": "^2.7.1",
 		"vue-multiselect": "^2.1.6",
 		"vue-visible": "^1.0.2",
-		"vue2-datepicker": "^3.6.2"
+		"vue2-datepicker": "^3.6.3"
 	},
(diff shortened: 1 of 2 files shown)


skjnldsv (approved):
Nice catch!

GretaD (approved):

ChristophWurst:
> We are getting vue2-datepicker 3.7.0 and since 3.6.3, the `format` prop not a string/object anymore. It's now just a string and there is a new `formatter` object prop.

How could we release this as a minor change? This is breaking Calendar in https://github.com/nextcloud/calendar/pull/2700

Bildschirmfoto von 2021-01-11 11-44-07

PR #1518, opened 2020-10-26
Also set Modal height/width to allow children to define their height/width and not only max-height/width

skjnldsv:
For viewer, to automatically contain the image, we need a specific css pattern that rely on a combination of max and height/width.
If only one is defined the chaining is broken.

| Before | After |
|:---------:|:------:|
|dev skjnldsv com_apps_contacts_All%20contacts_adb58ef0-8850-4f41-8afa-6b6144ca894b_contacts (1)|dev skjnldsv com_apps_contacts_All%20contacts_adb58ef0-8850-4f41-8afa-6b6144ca894b_contacts|

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -685,13 +685,19 @@ $header-size: 50px;
 	&:not(&--large):not(&--full) .modal-container {
 		max-width: 900px;
 		max-height: 80%;
+		width: 900px;
+		height: 80%;
 	}
 
 	// Sizing
+	// Always set max and height/width simultaneously here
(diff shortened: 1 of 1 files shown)


eneiluj (commented):
Looks fine.

Corners are rounded before and not after. Is it expected?

skjnldsv:
> Corners are rounded before and not after. Is it expected?

Yes, haven't applied the rounded corners on the proper element. This is unrelated to this component, :)

GretaD (approved):

ChristophWurst (approved):
Excellent code

PR #1526, opened 2020-10-30
Use static emoji picker

PVince81:
Solves constant CPU usage and energy impact when used in Talk app.

Also updated to 7.0.7.

See https://github.com/nextcloud/spreed/issues/4501 for context about the performance issue and some research about it.

Note: not reproducible on the demo page with the emoji picker, so it might be due to some layout configuration.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11747,9 +11747,9 @@
 			"dev": true
 		},
 		"emoji-mart-vue-fast": {
-			"version": "7.0.6",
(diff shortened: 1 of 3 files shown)


eneiluj (approved):
Yay! I couldn't see any graphical difference between classic and static emoji picker. Totally worth using the static one!

This component is still quite heavy to load. I recently solved a big performance issue in Cospend by using `v-if` instead of `v-show` to avoid rendering it multiple times. Rendering 10 emoji pickers (even without triggering them) was taking more than one second. Maybe it's worth mentioning this in nc-vue doc.

Anyway, thanks @PVince81!

PR #1542, opened 2020-11-04
Add default font and line size

GretaD:
Signed-off-by: GretaD <gretadoci@gmail.com>

PR #1543, opened 2020-11-04
Fix menu font for apps

GretaD:
This fixes it for mail

before
accountmenub
after
accountmenu
before
threadmenub
after
threadmenu
before
mailboxmenub
After
mailboxmenu

### DIFF
--- a/src/assets/action.scss
+++ b/src/assets/action.scss
@@ -66,6 +66,7 @@
 		box-shadow: none;
 
 		font-weight: normal;
+		font-size: var(--default-font-size);
 		line-height: $clickable-area;
 
 		&:hover,
--- a/src/assets/action.scss
+++ b/src/assets/action.scss
@@ -42,7 +42,7 @@
 		}
 	}
 }
-	
+
 
 @mixin action-item($name) {
 	.action-#{$name} {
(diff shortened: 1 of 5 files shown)


GretaD:
this to be merged after this #1542

skjnldsv (approved):
It's been bothering me for months! ^^'
Yes pleeaaaase ! :rocket: :heart_eyes:

PVince81:
doesn't seem to fix https://github.com/nextcloud/spreed/issues/4330, so the problem for the latter might be elsewhere

PVince81:
not sure why the action buttons in the talk menu would behave differently, it seems that the font size of the "button" is stronger than the one of the "span" that's inside.

I had to move the font-size from the  `&__text` to the parent element to make it work:
```
diff --git a/src/assets/action.scss b/src/assets/action.scss
index 0f5d5e0..6a03eaf 100644
--- a/src/assets/action.scss
+++ b/src/assets/action.scss
@@ -65,6 +65,7 @@
                box-shadow: none;
 
+               font-size: var(--default-font-size);
                font-weight: normal;
                line-height: $clickable-area;
 
@@ -107,10 +108,6 @@
 
                &__title {
                        font-weight: bold;
-                       font-size: var(--default-font-size);
-               }
-               &__text {
-                       font-size: var(--default-font-size);
                }
        }
 }
```

skjnldsv:
> not sure why the action buttons in the talk menu would behave differently, it seems that the font size of the "button" is stronger than the one of the "span" that's inside.

Maybe this and inherit in the title/text elements?

PVince81:
with the above diff on top of the PR it works for me.

if it also works for mail then we could go with that ?

PVince81 (approved):
:+1: looks good now also for the Talk menu, thanks

PR #1550, opened 2020-11-05
Update richcontenteditable to allow use in the description component

ma12-co:
- Spans always the whole width;
- If not `contenteditable` it looks like a regular paragraph
- Added disable prop that makes it look like a disabled input field

Screenshot from 2020-11-11 11-52-06

### DIFF
--- a/src/components/RichContenteditable/RichContenteditable.vue
+++ b/src/components/RichContenteditable/RichContenteditable.vue
@@ -114,6 +114,7 @@ export default {
 			'rich-contenteditable__input--empty': isEmptyValue,
 			'rich-contenteditable__input--multiline': multiline,
 			'rich-contenteditable__input--overflow': isOverMaxlength,
+			'rich-contenteditable__input--disabled': disabled,
 		}"
 		:contenteditable="contenteditable"
 		:placeholder="placeholder"
@@ -177,6 +178,14 @@ export default {
 		contenteditable: {
 			type: Boolean,
 			default: true,
+		},
+
+		/**
+		 * Applies the 'disabled' styles
+		 */
+		disabled: {
+			type: Boolean,
+			default: false,
 		},
 
 		/**
@@ -281,6 +290,12 @@ export default {
 		// Update default value
 		const renderedContent = this.renderContent(this.value)
 		this.$refs.contenteditable.innerHTML = renderedContent
+
+		// Dirty hack to remove the contenteditable attribute at first load
+		// if the prop is set to false.
+		if (!this.contentEditable) {
+			this.$refs.contenteditable.contentEditable = false
+		}
 
 	},
 	beforeDestroy() {
@@ -441,20 +456,20 @@ export default {
 // Standalone styling, independent from server
 .rich-contenteditable__input {
 	overflow-y: auto;
-	width: auto;
+	width: 100%;
 	margin: 0;
-	padding: 6px;
+	padding: 8px;
 	cursor: text;
 	white-space: pre-wrap;
 	word-break: break-word;
 	color: var(--color-main-text);
 	border: 1px solid var(--color-border-dark);
-	border-radius: var(--border-radius);
+	border-radius: var(--border-radius-large);
 	outline: none;
 	background-color: var(--color-main-background);
 	font-family: var(--font-face);
 	font-size: inherit;
-	min-height: $clickable-area;
+	min-height: var(--default-line-height);
 	max-height: $clickable-area * 5.5;
 
 	// Cannot use :empty because of firefox bug https://bugzilla.mozilla.org/show_bug.cgi?id=1513303
@@ -465,16 +480,23 @@ export default {
 	}
 
 	&[contenteditable='false'] {
+		border: 0;
+		opacity: 1;
+		padding: 0;
+		border-radius: 0;
+	}
+
+	&--disabled {
+		border: 1px solid var(--color-border-dark);
+		padding: 8px;
 		cursor: default;
 		opacity: .5;
 		color: var(--color-text-lighter);
-		border: 1px solid var(--color-background-darker);
-		border-radius: var(--border-radius);
+		border-radius: var(--border-radius-large);
 		background-color: var(--color-background-dark);
 	}
 
 	&--multiline {
-		min-height: $clickable-area * 3;
 		// No max for mutiline
 		max-height: none;
 	}


src/components/RichContenteditable/RichContenteditable.vue
```diff
@@ -152,7 +153,7 @@ export default {
 		},
 		autoComplete: {
 			type: Function,
-			required: true,
+			default: () => {},
```
skjnldsv, line 156: @jancborchardt do we want autocompletion in all editors?
I tend to say yes

src/components/RichContenteditable/RichContenteditable.vue
```diff
@@ -176,7 +177,15 @@ export default {
 		 */
 		contenteditable: {
 			type: Boolean,
-			default: true,
+			default: false,
```
skjnldsv, line 180: That is a breaking change.
Default should be true as this is what this component is all about.

src/components/RichContenteditable/RichContenteditable.vue
```diff
@@ -441,20 +456,20 @@ export default {
...
 	margin: 0;
-	padding: 6px;
+	padding: 8px;
 	cursor: text;
 	white-space: pre-wrap;
 	word-break: break-word;
 	color: var(--color-main-text);
 	border: 1px solid var(--color-border-dark);
-	border-radius: var(--border-radius);
+	border-radius: var(--border-radius-large);
```
skjnldsv, line 467: Please add screenshots as this is a design change

skjnldsv (changes requested):

src/components/RichContenteditable/RichContenteditable.vue
ma12-co: Why would you enforce autocompletion in all contenteditables?

src/components/RichContenteditable/RichContenteditable.vue
ma12-co: Sure

Screenshot from 2020-11-11 11-52-06

src/components/RichContenteditable/RichContenteditable.vue
skjnldsv: Because that is an awesome feature and would make sense for it to be available in all areas

ma12-co: It sure is awesome but that dosn't mean it's always needed.
Requiring it all the time sounds like a good recipe for scaring developers off this component IMO. It's good to offer the ability.
In the case of talk's group description, what sort of auto-completion would be desirable?

skjnldsv: > what sort of auto-completion would be desirable?

Depends on what you wish to write, could be a group mention, a user mention... etc etc :shrug:

ma12-co: I see no need for those in that context, hence why I removed the "required" key

skjnldsv: Yes, I figured :)

jancborchardt: If it’s about autocompletion of @-mentions, I would say it would be nice to have indeed. It’s something which would also be amazing for the Notes app, Text and rich workspaces for example.
And in the case of Talk room descriptions, you could list e.g. contact people, or more info on the maintainers of an app in the team channel of an app. Makes it more interactive and friendly, and ties in nicely with the user profiles / hovercards.

So unless we have a set of specific examples where we say it would be _negative_ to allow autocompletion, then I’d say it’s good to have? I mean, it’s basically a helpful thing converting a plain text `@marco` into the proper bubble, just like a plain text `nextcloud.com` would be linkified → it’s convenient and looks nice.

ma12-co: We can have different opinions bout talk's description, to me that's just noise. But regardless of that this is a basic building block and by requiring devs to pass in an autocompletion function we're making it more tedious and less versatile and thus less likely to be adopted. 
... Anyway, reverting..

skjnldsv: @ma12-co I guess you already have the whole autocompletion thing already/almost available on talk.
Feel free to wrap it into a component in the talk repo for you to use (especially for the [`userData`](https://github.com/nextcloud/nextcloud-vue/blob/7372d9a62bffbc3f7aea3c27a4f62cbcf8c2d5b0/src/mixins/richEditor/index.js#L37) needed to render the initial mentions)

skjnldsv:
```
src/components/RichContenteditable/RichContenteditable.vue
 483:20  ✖  Expected single quotes   string-quotes

```

src/components/RichContenteditable/RichContenteditable.vue
jancborchardt: > to me that's just noise

@ma12-co that's fine, and in that case you wouldn't use it.
But other systems with @-mention functionality do usually have that everywhere, not only in their chat component, e.g. Dropbox in their "Rich workspace" equivalent, or Github in their wikis, Trello in their comments, etc etc.

src/components/RichContenteditable/RichContenteditable.vue
```diff
@@ -282,6 +291,12 @@ export default {
 		const renderedContent = this.renderContent(this.value)
 		this.$refs.contenteditable.innerHTML = renderedContent
 
+		// Dirty hack to remove the contenteditable attribute at first load
+		// if the prop is set to false.
+		if (!this.contentEditable) {
+			this.$refs.contenteditable.contentEditable = false
```
skjnldsv, line 297:
```suggestion
			this.$refs.contenteditable.contentEditable = this.contendeditable
```

skjnldsv: What about `contentEditable` vs `contenteditable` ?

src/components/RichContenteditable/RichContenteditable.vue
```diff
@@ -441,20 +456,20 @@ export default {
...
 	color: var(--color-main-text);
 	border: 1px solid var(--color-border-dark);
-	border-radius: var(--border-radius);
+	border-radius: var(--border-radius-large);
 	outline: none;
 	background-color: var(--color-main-background);
 	font-family: var(--font-face);
 	font-size: inherit;
-	min-height: $clickable-area;
+	min-height: var(--default-line-height);
```
skjnldsv, line 472: With this the default looks off and the height is 37px.
Capture d’écran_2020-11-13_09-21-42

src/components/RichContenteditable/RichContenteditable.vue
```diff
@@ -465,16 +480,23 @@ export default {
 	}
 
 	&[contenteditable='false'] {
+		border: 0;
+		opacity: 1;
+		padding: 0;
```
skjnldsv, line 485: Styling should be consistent. Removing padding makes it look entirely different.

src/components/RichContenteditable/RichContenteditable.vue
```diff
@@ -465,16 +480,23 @@ export default {
...
+		border-radius: 0;
+	}
+
+	&--disabled {
+		border: 1px solid var(--color-border-dark);
+		padding: 8px;
```
skjnldsv, line 491: Do you need to add padding again here?

src/components/RichContenteditable/RichContenteditable.vue
```diff
@@ -465,16 +480,23 @@ export default {
...
 		cursor: default;
 		opacity: .5;
 		color: var(--color-text-lighter);
-		border: 1px solid var(--color-background-darker);
-		border-radius: var(--border-radius);
+		border-radius: var(--border-radius-large);
```
skjnldsv, line 495: Same goes for radius, it's already defined before (line 467)

src/components/RichContenteditable/RichContenteditable.vue
```diff
@@ -465,16 +480,23 @@ export default {
...
 		background-color: var(--color-background-dark);
 	}
 
 	&--multiline {
-		min-height: $clickable-area * 3;
```
skjnldsv, line 477: No, multiline should look multiline.

ma12-co:
Closing in favor of custom component in talk as discussed with @jancborchardt

PR #1555, opened 2020-11-06
Add tooltips in app sidebar header

PVince81:
Added nicer looking tooltip for the close button.
Added optional tooltip for the app sidebar header.

Tested with Talk where I locally set the titleTolltip property to the same value as the title:
image

And the close button:
image

I didn't want to make it default to the title as it might not always look good depending on use case.
In the case of the Talk app, the title is a user defined room name.
For other apps it might be static text that doesn't ellipsize nor needs a tooltip.

### DIFF
--- a/src/components/AppSidebar/AppSidebar.vue
+++ b/src/components/AppSidebar/AppSidebar.vue
@@ -130,9 +130,10 @@
 				}"
 				class="app-sidebar-header">
 				<!-- close sidebar button -->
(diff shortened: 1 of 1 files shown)


PVince81:
here's the talk PR: https://github.com/nextcloud/spreed/pull/4551

skjnldsv (approved):
sure

ChristophWurst (approved):

PR #1558, opened 2020-11-09
Revert "Also set Modal height/width to allow children to define their height/width and not only max-height/width"

skjnldsv:
Reverts nextcloud/nextcloud-vue#1518

As @ma12-co mentionned, this was a wrong approach, as the modal should wrap the flow instead.
It make more sense to keep capping the height/width instead.
Sorry about the noise! :(

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -685,19 +685,13 @@ $header-size: 50px;
 	&:not(&--large):not(&--full) .modal-container {
 		max-width: 900px;
 		max-height: 80%;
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
Excellent revert

PVince81 (approved):
:+1:

PVince81:
@skjnldsv can you make DCO happy ?

skjnldsv:
> @skjnldsv can you make DCO happy ?

It's fine without for this tiny changes. I'll override it

PR #1566, opened 2020-11-11
Remove sidebar actions slot max height

ma12-co:
Needed to fit talk's description in the header slot.

### DIFF
--- a/src/components/AppSidebar/AppSidebar.vue
+++ b/src/components/AppSidebar/AppSidebar.vue
@@ -215,8 +215,8 @@
 						</Actions>
 					</div>
 				</div>
-				<div v-if="$slots['primary-actions'] && !empty" class="app-sidebar-header__action">
-					<slot name="primary-actions" />
+				<div v-if="$slots['description'] && !empty" class="app-sidebar-header__description">
+					<slot name="description" />
 				</div>
 			</header>
 
@@ -688,11 +688,10 @@ $top-buttons-spacing: 6px;
 			}
 		}
 
-		// sidebar action(s) slot
-		&__action {
+		// sidebar description slot
+		&__description {
 			display: flex;
 			align-items: center;
-			max-height: 50px;
 			margin: 0 10px;
 		}
 	}
@@ -736,8 +735,8 @@ $top-buttons-spacing: 6px;
 
 <style lang="scss">
 // ! slots specific designs, cannot be scoped
-// if any button inside the action slot, icrease visual padding
-.app-sidebar-header__action {
+// if any button inside the description slot, icrease visual padding
+.app-sidebar-header__description {
 	button, .button,
 	input[type='button'],
 	input[type='submit'],


skjnldsv (commented):
Maybe we should even rename the slot :p

ma12-co:
How would you like it named?? :)

skjnldsv:
> How would you like it named?? :)

No clue, could be description. I am not the best to find names :see_no_evil:

src/components/AppSidebar/AppSidebar.vue
```diff
@@ -215,8 +215,8 @@
 						</Actions>
 					</div>
 				</div>
-				<div v-if="$slots['primary-actions'] && !empty" class="app-sidebar-header__action">
-					<slot name="primary-actions" />
+				<div v-if="$slots['description'] && !empty" class="app-sidebar-header__action">
```
skjnldsv, line 218:
```suggestion
				<div v-if="$slots['description'] && !empty" class="app-sidebar-header__description">
```

src/components/AppSidebar/AppSidebar.vue
```diff
@@ -692,7 +692,6 @@ $top-buttons-spacing: 6px;
 		&__action {
```
skjnldsv, line 692:
```suggestion
		&__description {
```

src/components/AppSidebar/AppSidebar.vue
skjnldsv: Ah right, we already have app-sidebar-header__desc
Maybe a better name then?

`app-sidebar-header__action` => `app-sidebar-header__description`
`app-sidebar-header__desc` => `app-sidebar-header__details`

?

skjnldsv (commented):
In that case, since we're at it with the breaking changes, rename `secondary-actions` to  `actions` ? As well as `tertiary-actions` to `star` ? :thinking: :shrug: 

cc @raimund-schluessler

ma12-co:
I just wanted to remove a css rule though :sweat_smile:

raimund-schluessler:
> 
> 
> I just wanted to remove a css rule though 😅

Would this be a breaking change as well?

Renaming these slots requires every app using the appsidebar to adjust (I don't mind though, I don't use it yet, just thinking loud 🙈)

> As well as `tertiary-actions` to `star` ?

It's only a star by default, but since it's a slot, it can be anything. So `star` might be to specific as a slot name.

ma12-co:
I'm fine with anything here, just lmk what to do :)

skjnldsv:
> Would this be a breaking change as well?

There is another pr that touches the sidebar that is breaking change yes :)

skjnldsv:
> It's only a star by default, but since it's a slot, it can be anything. So `star` might be to specific as a slot name.

I would argue that 'tertiary-actions' isn't really self-explanatory either :stuck_out_tongue:

ma12-co:
So should I do this?? https://github.com/nextcloud/nextcloud-vue/pull/1566#pullrequestreview-528155537

ma12-co:
I got back to renaming only the slot and classes referring to that slot because of naming conflicts. Let's address the rest of the names in separate PRs if needed.

skjnldsv (approved):

Ivansss (approved):

PR #1573, opened 2020-11-16
Fix sizing of modal container and app settings dialog

PVince81:
The modal container's contents now correctly adjusts to its contents
height, but also the content adjusts to the modal's height when that one
is constrained by the window height.

This fixes scrollbar container issues in some screen resolutions with
the app settings dialog.

See https://github.com/nextcloud/spreed/pull/4576 and reduce the vertical window size to reproduce the issue.

### Before the fix:
image

- app-settings not fitting into the reserved space, so the overflow/scroll doesn't kick in at the right moment.

### After the fix:
image

image

- scrollbar appears and works
- clicking section headers scrolls correctly
- with less content the dialog also reduces its height to fit the contents (tested by manually deleting DOM elements)


src/components/Modal/Modal.vue
```diff
@@ -674,7 +674,7 @@ $header-size: 50px;
 
 	/* Content */
 	.modal-container {
-		display: block;
```
skjnldsv, line 677: No, please keep it as default html: block.
This is a breaking change.
If you want it flex, change it on your app.

Unless there is something I missed, but I don't think enforcing flex is a good idea. We should still stay close as the html basics I think :)

PVince81: I made it configurable, defaults to false.

This is needed for any instance of AppSettingsDialog for smaller screen resolutions.

src/components/Modal/Modal.vue
```diff
@@ -281,6 +281,13 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		/**
+		 * Makes the modal flex
+		 */
+		flex: {
```
skjnldsv, line 287: Oh please no. That's even worse :see_no_evil:

src/components/Modal/Modal.vue
skjnldsv: Then override it with v-deep.
That is why it's there for :see_no_evil:

src/components/Modal/Modal.vue
PVince81: Ok, we just clarified and using v-deep is ok for this use case. At first I thought v-deep was forbidden.
I'll adjust.

PVince81:
I've replaced all changes with a v-deep clause to set display:flex on modal-container only when using in the AppSettingsDialog.

@ma12-co can you confirm that the dialog now looks as originally intended with https://github.com/nextcloud/spreed/pull/4576 with smaller sizes ?

juliushaertl (approved):

ma12-co (approved):

skjnldsv (approved):

PR #1576, opened 2020-11-17
Fix action button text for long translations

GretaD:
the problem
before_longtezt

### DIFF
--- a/src/assets/action.scss
+++ b/src/assets/action.scss
@@ -90,7 +90,7 @@
 
 		// long text area
 		p {
-			width: 150px;
+			width: 185px;
 			padding: #{$icon-margin / 2} 0;
 
 			cursor: pointer;
@@ -107,6 +107,11 @@
 
 		&__title {
 			font-weight: bold;
+			text-overflow: ellipsis;
+			overflow: hidden;
+			white-space: nowrap;
+			max-width: 100%;
+			display: inline-block;
 		}
 	}
 }


GretaD:
@skjnldsv do you think this approach is good? This happens with different languages that have long translation for a certain text, and `white-space: pre-wrap` doesnt work.

skjnldsv:
I think it could make sense to enlarge a bit, but it might be just an issue with the title not having a width 100% and being ellipsised properly instead 🤔

GretaD:
> I think it could make sense to enlarge a bit, but it might be just an issue with the title not having a width 100% and being ellipsised properly instead thinking

I tried adding `width: 100%` that also needs a `display: block` on __title. But still doesnt work.

skjnldsv:
> I tried adding `width: 100%` that also needs a `display: block` on __title. But still doesnt work.

You need a bit more :)
```css
	text-overflow: ellipsis;
	overflow: hidden;
	white-space: nowrap;
	max-width: 100%;
	display: inline-block;
```

skjnldsv:
Capture d’écran_2020-11-17_19-36-36

src/assets/action.scss
```diff
@@ -90,7 +90,7 @@
 
 		// long text area
 		p {
-			width: 150px;
+			width: 185px;
```
skjnldsv, line 93: Make sure @jancborchardt is ok with this :)

skjnldsv (approved):

GretaD:
I also increased a bit the `p` width, to make sure to ellips only very long translations and not everything that is a bit longer than english

juliushaertl (approved):

src/assets/action.scss
juliushaertl: :ping_pong: @jancborchardt

PR #1581, opened 2020-11-18
Multiselect expose open-direction property

PVince81:
Expose the open-direction property to allow forcing an opening direction
for the dropdown.


PVince81:
required to fix server regression: https://github.com/nextcloud/server/issues/23617

PVince81:
seems I originally tested it wrong, the property is already exposed implicitly.

PR #1584, opened 2020-11-19
Fix contenteditable state ad first load

ma12-co:
Fixes a bug in which the `<div contentaditable>` would be `contenteditable = true` even if the `contenteditable` prop was set to false.

### DIFF
--- a/src/components/RichContenteditable/RichContenteditable.vue
+++ b/src/components/RichContenteditable/RichContenteditable.vue
@@ -280,6 +280,9 @@ export default {
 		// Update default value
 		this.updateContent(this.value)
 
+		// Removes the contenteditable attribute at first load if the prop is
+		// set to false.
+		this.$refs.contenteditable.contentEditable = this.contenteditable
 	},
 	beforeDestroy() {
 		if (this.tribute) {


src/components/RichContenteditable/RichContenteditable.vue
```diff
@@ -280,6 +280,9 @@ export default {
 		// Update default value
 		this.updateContent(this.value)
 
+		// Removes the contenteditable attribute at first load if the prop is
+		// set to false.
+		this.$refs.contenteditable.contentEditable = this.contenteditable
```
skjnldsv, line 285:
```suggestion
		this.$refs.contenteditable.contenteditable = this.contenteditable
```

ma12-co: Nope, that's the name of the actual key of a div.contenteditable html element. It's a weird convention given that in the template it's all lowercase 🤷

Screenshot from 2020-11-20 10-50-40

skjnldsv: That is awful ^^'
https://developer.mozilla.org/fr/docs/Web/API/HTMLElement/contentEditable

skjnldsv (approved):

Ivansss (approved):

PR #1585, opened 2020-11-19
Set navigation role and aria label on app navigation

PVince81:
Makes it possible for screen readers to say the text of aria label
followed by the term "navigation" when entering that region.

Also added an optional arial label for the toggle.

In the talk app I've set the aria label to "Conversation list" so when entering the left sidebar (or jumping to any element inside) the screen reader says "Conversation list navigation".

### DIFF
--- a/src/components/AppNavigation/AppNavigation.vue
+++ b/src/components/AppNavigation/AppNavigation.vue
@@ -57,6 +57,7 @@ emit('toggle-navigation', {
 	<div
 		id="app-navigation-vue"
 		class="app-navigation"
+		role="navigation"
 		:class="{'app-navigation--close':!open }">
 		<AppNavigationToggle :open="open" @update:open="toggleNavigation" />
 		<slot />


src/components/AppNavigation/AppNavigation.vue
```diff
@@ -85,6 +90,23 @@ export default {
...
+			type: String,
+			default: '',
+		},
+		/**
+		 * aria-label for the toggle button
+		 */
+		toggleAriaLabel: {
+			type: String,
+			default: '',
+		},
```
skjnldsv, line 107: Do we really need them?

```suggestion
		ariaLabel: {
			type: String,
			default: t('App navigation'),
		},
		/**
		 * aria-label for the toggle button
		 */
		toggleAriaLabel: {
			type: String,
			default: t('App navigation toggle'),
		},
```

src/components/AppNavigation/AppNavigation.vue
```diff
@@ -57,8 +57,13 @@ emit('toggle-navigation', {
 	<div
 		id="app-navigation-vue"
 		class="app-navigation"
+		role="navigation"
+		:aria-label="ariaLabel || t('App navigation')"
 		:class="{'app-navigation--close':!open }">
-		<AppNavigationToggle :open="open" @update:open="toggleNavigation" />
+		<AppNavigationToggle
+			:aria-label="toggleAriaLabel || t('App navigation toggle')"
```
skjnldsv, line 64:
```suggestion
			:aria-label="toggleAriaLabel"
```

src/components/AppNavigation/AppNavigation.vue
```diff
@@ -57,8 +57,13 @@ emit('toggle-navigation', {
 	<div
 		id="app-navigation-vue"
 		class="app-navigation"
+		role="navigation"
+		:aria-label="ariaLabel || t('App navigation')"
```
skjnldsv, line 61:
```suggestion
		:aria-label="ariaLabel"
```

skjnldsv (changes requested):
See comments

src/components/AppNavigation/AppNavigation.vue
PVince81: Seems it's possible to set the aria-label anyway from outside, so probably not needed as property.

With the toggle I have mixed feelings as it's currently not reachable with keyboard nav, but we might need to do it.

We could also agree to settle on "App navigation" and "App navigation toggle" for these names regardless of the app.

The only thing is that with the talk app it doesn't feel like an app navigation as it's the conversation list, hence the idea of having a custom text there.

PVince81:
Ok, I've decided to go only with the "role=navigation" which should be enough to make this a region to jump to.
For even more friendliness, app devs should supply an aria-label on their side (I did so in the talk app here https://github.com/nextcloud/spreed/pull/4652)

src/components/AppNavigation/AppNavigation.vue
skjnldsv: I'm fine with having this if you think it's easier, but default to an empty string and fallback in the template is not the way to go :)
props defaults are here for a reason ;)

skjnldsv (approved):

src/components/AppNavigation/AppNavigation.vue
skjnldsv: > Seems it's possible to set the aria-label anyway from outside, so probably not needed as property.

Only for the top container though, not for the toggle

ChristophWurst (approved):

src/components/AppNavigation/AppNavigation.vue
PVince81: ah ok.

anyway, I'll certainly return to this once I look into why we cannot tab to the toggle

PR #1586, opened 2020-11-19
Remote semantics of h6 in user bubble

PVince81:
This will prevent screen readers to say "Heading level 6" + the
contents instead of just reading the contents.

Tested with Orca and Firefox.

### DIFF
--- a/src/components/UserBubble/UserBubble.vue
+++ b/src/components/UserBubble/UserBubble.vue
@@ -94,9 +94,9 @@ This component has the following slot:
 				class="user-bubble__avatar" />
 
 			<!-- Title -->
-			<h6 class="user-bubble__title">
+			<span class="user-bubble__title">
 				{{ displayName || user }}
-			</h6>
+			</span>
 
 			<!-- @slot Optional slot just after the title -->
 			<span v-if="$slots.title" class="user-bubble__secondary">


juliushaertl (approved):

skjnldsv (approved):

src/components/UserBubble/UserBubble.vue
```diff
@@ -94,7 +94,7 @@ This component has the following slot:
 				class="user-bubble__avatar" />
 
 			<!-- Title -->
-			<h6 class="user-bubble__title">
+			<h6 class="user-bubble__title" role="presentation">
```
skjnldsv, line 97: Do we still need the h6 then? I am not sure that make sense if we strip of their default role

PVince81: we don't really, could use a div instead and need to check that the styling stays the same

skjnldsv: span then if it's just text

ma12-co (approved):

skjnldsv (changes requested):
Blocking https://github.com/nextcloud/nextcloud-vue/pull/1586/files#r528566017

PVince81:
I've replaced it now with a span and checked side by side with the old style: they look identical.

skjnldsv (approved):
:rocket:

PR #1587, opened 2020-11-19
Allow disabling the avatar menu

PVince81:
Hide clickability from screen reader by not registering the click event when not clickable.

### DIFF
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -44,7 +44,7 @@
 		}"
 		:style="avatarStyle"
 		class="avatardiv popovermenu-wrapper"
-		@click="toggleMenu">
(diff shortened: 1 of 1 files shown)


src/components/Avatar/Avatar.vue
```diff
@@ -44,7 +44,7 @@
 		}"
 		:style="avatarStyle"
 		class="avatardiv popovermenu-wrapper"
-		@click="toggleMenu">
+		v-on="disableMenu ? { click: toggleMenu } : {}">
```
skjnldsv, line 47: There is also this syntax since 2.6 `<div @[event]="handler" />` if you prefer

skjnldsv (approved):

ma12-co (approved):

PR #1592, opened 2020-11-20
Richcontenteditable: bind event listeners

ma12-co:
Useful to dismiss an edit from the parent.  

### DIFF
--- a/src/components/RichContenteditable/RichContenteditable.vue
+++ b/src/components/RichContenteditable/RichContenteditable.vue
@@ -121,6 +121,7 @@ export default {
 		class="rich-contenteditable__input"
 		role="textbox"
 		@input="onInput"
+		v-on="$listeners"
 		@keydown.delete="onDelete"
 		@keydown.enter.exact="onEnter"
 		@keydown.ctrl.enter.exact.stop.prevent="onCtrlEnter"


skjnldsv (commented):
You should be able to watch the @keydown.esc from the parent

ma12-co:
> You should be able to watch the @keydown.esc from the parent

Tried that first, didn't work for me

skjnldsv:
If so then let's forward all listeners

ma12-co:
I don't understand @skjnldsv, listeners are to be passed down to child components AFAIK. Here I need an event to come up to the parent in order to disable editing on `esc`

skjnldsv:
> I don't understand @skjnldsv, listeners are to be passed down to child components AFAIK. Here I need an event to come up to the parent in order to disable editing on `esc`

Not being able to bind to the component any events is an issue.
If any time a dev want to add a listener, they have to add a line in the vue components it doesn't scale. So let's send all listeners from the parent to the contenteditable and call it a day :)

src/components/RichContenteditable/RichContenteditable.vue
```diff
@@ -434,6 +435,13 @@ export default {
 			}
 			this.$emit('submit', event)
 		},
+		/**
+		 * Esc key pressed
+		 * @param {Event} event the keydown event
+		 */
+		onEsc(event) {
+			this.$emit('esc', event)
```
skjnldsv, line 443: Also, as a regular note, you should not change the event type. Here this is not an `esc` event, this is a  `keydown`.
Staying as close to html/js as possible is easier to understand for other devs

ma12-co: I initially called it `keydown.esc`, but it was not working. I believe It was conflicting with the vue listener on the parent.

skjnldsv: because the event is a `keydown`, not a `keydown.esc`

ma12-co:
> So let's send all listeners from the parent to the contenteditable and call it a day 

That's not what I need though. I need to trigger a method on the parent based on a keydown event that occurs in the child component

skjnldsv:
> > So let's send all listeners from the parent to the contenteditable and call it a day
> 
> That's not what I need though. I need to trigger a method on the parent based on a keydown event that occurs in the child component

Which is exactly what you're doing. `@xxxxx` in vue is called a listener https://vuejs.org/v2/guide/events.html#Listening-to-Events

ma12-co:
This is from the vue docs: 

Using the $listeners property, you can forward all event listeners on the component to a specific child element with v-on="$listeners".

That is not what I need! I'd need that if I was listening in the parent and triggering my method in the child. I need to listen in the child and trigger the method in the parent.

skjnldsv:
> That is not what I need! I'd need that if I was listening in the parent and triggering my method in the child. I need to listen in the child and trigger the method in the parent.

That is the only thing you can do in vue.
You cannot pass events from parent to children. Have you even tried what I suggested?

ma12-co:
> Have you even tried what I suggested?

Yes  

To accomplish what you want here, I think we should forward all keydown events to the parent, not the listeners to the child

skjnldsv (approved):

ma12-co:
You were right @skjnldsv, shame on me!

ChristophWurst (approved):

skjnldsv:
Unrelated error

PR #1601, opened 2020-11-24
Add required prop to DatetimePicker

jotoeri:
Adding the html-prop `required` to the datetimePicker input. -> Necessary for forms-app mandatory questions.

As the Datepicker does not directly support this prop, we need to use the `inputAttr`-prop of the Datepicker. The `inputAttr` requests an object that then is applied onto the html-input of the Datepicker.

### DIFF
--- a/src/components/DatetimePicker/DatetimePicker.vue
+++ b/src/components/DatetimePicker/DatetimePicker.vue
@@ -82,6 +82,7 @@ export default {
 		:type="type"
 		:value="value"
 		:append-to-body="false"
+		:input-attr="inputAttr"
 		v-bind="$attrs"
 		v-on="$listeners"
 		@select-year="handleSelectYear"
@@ -152,6 +153,20 @@ export default {
 			default() {
 				return new Date()
 			},
+		},
+
+		required: {
+			type: Boolean,
+			default: false,
+		},
+	},
+
+	computed: {
+		// An object of attributes, that are not explicitely provided as prop on vue2-datepicker. Applied there onto html input element.
+		inputAttr() {
+			return {
+				required: this.required,
+			}
 		},
 	},
 


src/components/DatetimePicker/DatetimePicker.vue
```diff
@@ -82,6 +82,7 @@ export default {
 		:type="type"
 		:value="value"
 		:append-to-body="false"
+		:input-attr="inputAttr"
 		v-bind="$attrs"
```
skjnldsv, line 86: We already bind all props here, I'm confused to why it doesn't work.

skjnldsv: Ah, because it needs an object.
Yeah, then let's use input-attr in forms directly then, this is why we added `v-bind="$attrs"`

jotoeri: As the vue2-datepicker does not take `required` as prop. So this is why the way through `inputAttr` is necessary. I could do that also directly in forms, i just thought it is a bit nicer to have the required-prop here directly accessible.

Also see https://github.com/mengxiong10/vue2-datepicker/blob/2eea21fe450df981ae45bd18e69394af4244619a/src/date-picker.js#L446

jotoeri: :man_shrugging:

skjnldsv: > i just thought it is a bit nicer to have the required-prop here directly accessible

Yeah it's a fair point. While we should stay as close as the original lib when wrapping components like that (datepicker, multiselect...) I'm also conflicted by staying as close as the html standards :see_no_evil: 

Both are valid points I think. @ChristophWurst @juliushaertl any thoughts?

skjnldsv: In any case, by doing this you remove the inputAttr from the available properties of the datetimepicker, which we should not do.
If we'd implement the required prop here, we should still keep both I think?

jotoeri: Any opinion, @ChristophWurst @juliushaertl ?
So should i reopen and fix this PR?

PR #1605, opened 2020-11-25
Add changelog file

PVince81:
Copied entries from Github Releases up to 2.6.0.

Because I heard that we're doing changelogs also in other repos like https://github.com/nextcloud/nextcloud-dialogs

For some reason CHANGELOG.md was in gitignore, maybe there were plans to automate this ?

Anyway, let me know if down to 2.6.0 is enough or whether it's useful to have even older entries.

I've adjusted the README to mention that whoever does a release needs to update the changelog as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,8 +17,6 @@ yarn-error.log*
 
 .marginalia
 
-CHANGELOG.md
-
 # Unit test / coverage reports
 coverage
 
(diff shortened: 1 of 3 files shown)


PVince81:
also note: I've reformatted the setcion titles to match the former format which seems to be based on https://keepachangelog.com/en/1.0.0/

juliushaertl (approved):

skjnldsv (approved):

PR #1614, opened 2020-11-27
Quickfix for action button and material design icon

PVince81:
When using a material design icon in the icon slot, this fix will
properly align it vertically.

We need it because the ActionButton consumer cannot override the styles with `::v-deep` since the popover itself might be in the document body

When testing the docs you might need to scroll up or down to see the dropdown because of https://github.com/nextcloud/server/issues/23872 which also occurs in the docs...

### DIFF
--- a/src/components/ActionButton/ActionButton.vue
+++ b/src/components/ActionButton/ActionButton.vue
@@ -49,8 +49,7 @@ You can also use a custom icon, for example from the vue-material-design-icons l
 		<ActionButton>
 			<HandLeft
 				slot="icon"
-				:size="24"
-				fill-color="#000000"
+				:size="16"
 				decorative
 				title="" />
 			Raise left hand
@@ -58,8 +57,7 @@ You can also use a custom icon, for example from the vue-material-design-icons l
 		<ActionButton>
 			<HandRight
 				slot="icon"
-				:size="24"
-				fill-color="#000000"
+				:size="16"
 				decorative
 				title="" />
 			Raise right hand
--- a/src/components/ActionButton/ActionButton.vue
+++ b/src/components/ActionButton/ActionButton.vue
@@ -40,6 +40,44 @@ If you're using a long text you can specify a title
 		<ActionButton icon="icon-delete" title="Long button" @click="alert('Delete')">This button is associated with a very long text.\nAnd with new lines too.</ActionButton>
 	</Actions>
 ```
+
+You can also use a custom icon, for example from the vue-material-design-icons library:
+
+```vue
+<template>
+	<Actions>
+		<ActionButton>
+			<HandLeft
+				slot="icon"
+				:size="24"
+				fill-color="#000000"
+				decorative
+				title="" />
+			Raise left hand
+		</ActionButton>
+		<ActionButton>
+			<HandRight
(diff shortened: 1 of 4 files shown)


PVince81:
Without the CSS fix:
image

With the CSS fix:
image

ChristophWurst (approved):

src/components/ActionButton/ActionButton.vue
```diff
@@ -40,6 +40,44 @@ If you're using a long text you can specify a title
...
+
+You can also use a custom icon, for example from the vue-material-design-icons library:
+
+```vue
+<template>
+	<Actions>
+		<ActionButton>
+			<HandLeft
+				slot="icon"
+				:size="24"
```
ma12-co, line 52:
```suggestion
				:size="16"
```

src/components/ActionButton/ActionButton.vue
```diff
@@ -40,6 +40,44 @@ If you're using a long text you can specify a title
...
+				fill-color="#000000"
+				decorative
+				title="" />
+			Raise left hand
+		</ActionButton>
+		<ActionButton>
+			<HandRight
+				slot="icon"
+				:size="24"
```
ma12-co, line 61:
```suggestion
				:size="16"
```

src/components/ActionButton/ActionButton.vue
```diff
@@ -40,6 +40,44 @@ If you're using a long text you can specify a title
...
+You can also use a custom icon, for example from the vue-material-design-icons library:
+
+```vue
+<template>
+	<Actions>
+		<ActionButton>
+			<HandLeft
+				slot="icon"
+				:size="24"
+				fill-color="#000000"
```
ma12-co, line 53:
```suggestion
```

src/components/ActionButton/ActionButton.vue
```diff
@@ -40,6 +40,44 @@ If you're using a long text you can specify a title
...
+				decorative
+				title="" />
+			Raise left hand
+		</ActionButton>
+		<ActionButton>
+			<HandRight
+				slot="icon"
+				:size="24"
+				fill-color="#000000"
```
ma12-co, line 62:
```suggestion
```

ma12-co (changes requested):
See comments

PVince81:
@ma12-co seems 16 is too small

here I used 16 for the material design icons:
image

hmmm, but 24 is maybe too big indeed:
image

here's 20:
image

let's go with 20 then ?

this is anyway just a code example... ideally in the future we should have a way to only provide the icon name...

ma12-co:
@PVince81 I agree that 18 or 20 would look better, but for now 16px is our default in Nextcloud. We can talk about bumping that up a bit but I wouldn't do it here.

PVince81:
> @PVince81 I agree that 18 or 20 would look better, but for now 16px is our default in Nextcloud. We can talk about bumping that up a bit but I wouldn't do it here.

ok, fine. adjusted. please review

PVince81:
rebased for cypress fix

ChristophWurst (approved):

juliushaertl (approved):

PR #1662, opened 2021-01-11
Allow ActionInput type=color

ChristophWurst:
Ref https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/color

Fixes 
Bildschirmfoto von 2021-01-11 12-15-59


skjnldsv (approved):

tcitworld:
`email` is also missing from this list. Should I create another PR?

PVince81 (approved):
:+1:

ChristophWurst:
> `email` is also missing from this list. Should I create another PR?

Yes please :)

PR #1669, opened 2021-01-15
Add link support to richcontenteditor and fix various issues

skjnldsv:
- Add link support on rendering
- Make sure we also decode &amp;
- Replaced badly used `value` variable
- Make sure regex are multilines and insensitive

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10814,6 +10814,29 @@
 				"sha.js": "^2.4.8"
 			}
 		},
+		"create-html-element": {
+			"version": "2.1.0",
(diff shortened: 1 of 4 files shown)


nickvergessen (approved):
Works and does the trick for https://github.com/nextcloud/spreed/issues/4950

PVince81 (approved):
:+1:

PVince81:
regression in Safari: https://github.com/nextcloud/spreed/issues/4995

seems this new linkify-url lib doesn't work with Safari
and now wondering why we didn't use linkifyjs instead

PVince81:
fix for the regression, make it use linkifyjs instead: https://github.com/nextcloud/nextcloud-vue/pull/1678

PR #1670, opened 2021-01-19
Feature/noid/avatar menu use actions component

PVince81:
This makes the Avatar component use the Actions component for its menu.

- [ ] requires: https://github.com/nextcloud/nextcloud-vue/pull/831 as base

image

Issues:

- [ ] need a way to set button size for Actions from attributes
- [ ] opening the menu has the repositioning/resize glitch from https://github.com/nextcloud/nextcloud-vue/pull/831, will need a different approach for differentiating "open state" vs "data loaded and safe to display" state. Maybe need to add a loading attribute to the actions
- [ ] keyboard nav still not working, can't enter the menu. Seems to be a bug in Actions component.
- [ ] make sure that all action types are supported, need to check all the formats returned by the endpoint

### DIFF
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -37,31 +37,33 @@
 </docs>
 <template>
 	<div v-tooltip="tooltip"
-		v-click-outside="closeMenu"
 		:class="{
 			'avatardiv--unknown': userDoesNotExist,
 			'avatardiv--with-menu': hasMenu
 		}"
 		:style="avatarStyle"
-		class="avatardiv popovermenu-wrapper"
-		v-on="!disableMenu ? { click: toggleMenu } : {}">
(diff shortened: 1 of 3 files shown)


nickvergessen:
I think this is the wrong way. Instead one of the next todos is to blow the menu up and show a bigger avatar similar to what github does:
Bildschirmfoto von 2021-01-20 12-14-03

PVince81:
Right, and we still need a place for the actions. Maybe a sub-menu inside that dialog, which itself will be the Actions menu.
The fact that actions are server-defined makes it more tricky.

Was there an existing discussion somewhere already about this new design ?

raimund-schluessler:
What is the plan here? The Avatar component is the last one to use the deprecated PopoverMenu component and I think it would be good to migrate away from it.

raimund-schluessler:
Superseded by #4017.

PR #1675, opened 2021-01-20
Fix longtext appearance for avatar popover menu

PVince81:
Use longtext with the avatar menu as it will mostly contain user names
and email addresses which can become long.

The longtext mode automatically triggers wrapping by default.
However, in the case of long email addresses we can't wrap on spaces, so
an ellipsis has been added as well.

(note, there is no white-space mode that can break in the middle of a string, so this is the best we got)

image

For https://github.com/nextcloud/spreed/pull/4976#issuecomment-763517914

### DIFF
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -376,7 +376,7 @@ export default {
 				return {
 					href: item.hyperlink,
 					icon: item.icon,
-					text: item.title,
+					longtext: item.title,
 				}
 			})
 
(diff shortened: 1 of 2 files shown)


nickvergessen (approved):

violoncelloCH:
> (note, there is no white-space mode that can break in the middle of a string, so this is the best we got)

maybe `word-break: break-all;` could do the trick for that one?

PVince81:
> > (note, there is no white-space mode that can break in the middle of a string, so this is the best we got)
> 
> maybe `word-break: break-all;` could do the trick for that one?

while it works for the email address, it breaks the other items in an unwanted way:
image

and at this level of the code we can't know what the contents of the menu items is about as it's returned by the server in a generic manner

ma12-co (approved):

PR #1677, opened 2021-01-22
Increase max width of action button

PVince81:
Allows for longer German words which cannot be wrapper when they don't
contain spaces.

In case a longer text if present that doesn't fit and cannot wrap, like
an email address, an ellipsis will now be shown.

Before:
Bildschirmfoto von 2021-01-21 15-51-16

After:
image

### DIFF
--- a/src/assets/action.scss
+++ b/src/assets/action.scss
@@ -100,13 +100,17 @@
 
 		// long text area
 		p {
-			width: 185px;
+			width: 220px;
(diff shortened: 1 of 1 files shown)


PVince81:
note: we already had to do something similar for the popover menu in the previous patch release: https://github.com/nextcloud/nextcloud-vue/pull/1675

ma12-co (approved):

ChristophWurst (approved):

Issue #1488, opened 2020-10-18
Move away from linkifyjs


skjnldsv:
Uses jQuery and is quite heavy for a simple string replacement
Let's move to https://bundlephobia.com/result?p=linkify-urls, highly supported and sindresorhus is an awesome dev with great libs! :)

dartcafe:
Thanks for this hint. I changed from linkifyjs to linkify-urls@2 for polls. Seems that version 3 does not work in safari as mentioned on the github page.

juliushaertl:
> Uses jQuery and is quite heavy for a simple string replacement

linkifyjs also works fine without jquery, but if linkify-urls is a sane replacement, of course saving those bytes makes sense :+1: 

https://bundlephobia.com/result?p=linkifyjs@2.1.9
https://bundlephobia.com/result?p=linkify-urls@3.1.1

raimund-schluessler:
We only use `linkifyjs/string` and webpack is smart enough to not bundle the whole `linkifyjs` package. However, with linkifyjs@3, `linkify-string` moved to its own package, so absolutely no need for `linkifyjs` anymore.

And `linkify-string` is even smaller than `linkify-urls`: https://bundlephobia.com/package/linkify-string@3.0.2

So I think we can close this (at least as soon as https://github.com/nextcloud/nextcloud-vue/pull/2275 is in).

PR #1678, opened 2021-01-22
Replace linkify-urls with linkifyjs

PVince81:
Linkify-urls is redundant as we already had a library.
Linkify-urls is not compatible with Safari as it's designed to be used
on Node JS and it uses non-polyfillable regexp features.

image

@nickvergessen please test this with Safari, also with Talk.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,6 @@
 		"emoji-mart-vue-fast": "^7.0.7",
 		"escape-html": "^1.0.3",
 		"hammerjs": "^2.0.8",
-		"linkify-urls": "^3.1.1",
 		"linkifyjs": "~2.1.9",
 		"md5": "^2.2.1",
 		"regenerator-runtime": "^0.13.5",
--- a/src/mixins/richEditor/index.js
+++ b/src/mixins/richEditor/index.js
@@ -21,7 +21,7 @@
  */
 
 import escapeHtml from 'escape-html'
-import linkifyUrls from 'linkify-urls'
+import linkifyStr from 'linkifyjs/string'
 import stripTags from 'striptags'
 import Vue from 'vue'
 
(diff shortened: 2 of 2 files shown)


nickvergessen (approved):
Works

dartcafe:
Is it a good idea to replace a 7kb dependency with a >500kb one?
Just mentioning:  #1488 
And linkify-urls@2 works great, also in Safari.

nickvergessen:
Linkify is already a dependency.
And at least the current linkify-urls causes a parse error on safari, which is why Talk and the vue lib docs are not working on recent safari

PVince81:
so far, as this is mostly fire fighting, I'd like to move forward with this fix and not use two different libraries.

separately we could look into finding a smaller library then

I think @dartcafe was referring to the old version of linkify-urls which doesn't seem to have the back-references in the regexp https://github.com/sindresorhus/linkify-urls/blob/v2.2.0 vs https://github.com/sindresorhus/linkify-urls/blob/2d04fd8121e6adc33d35f1303495966a39f0dfa8/index.js#L5.
But I'm not sure if we'll want to use an outdated / old major of a library. Maybe there's yet another library that would fit.
This will need further research.

ChristophWurst (approved):

PVince81:
I've raised https://github.com/nextcloud/nextcloud-vue/issues/1679 for researching a smaller dep that works everywhere

skjnldsv:
SO, what was the decision regarding https://github.com/nextcloud/nextcloud-vue/issues/1488 ?
Because linkifyjs uses jQuery and is quite heavy for a simple string replacement.

PVince81:
follow up in https://github.com/nextcloud/nextcloud-vue/issues/1488

there was no further discussion about it.

my PR was just an emergency thing to unblock a release due to Safari regressions

Issue #1681, opened 2021-01-25
[3.5.2] Popover - Trigger doesn't trigger anymore


dartcafe:
When clicking on the trigger, nothing happens. 
`doesShowMenu` stays `false`. When setting to `true` via devTools, the popup appears.
See also: https://nextcloud-vue-components.netlify.app/#/Components/Popover

PVince81:
I locally reverted the "popover-base-class" to "popover" and the popup now appears again.
But the weird thing is that despite displaying something, the "open" property isn't changing.
Hmm, I guess that property was added just to be able to open the popup externally (which is what I originally tested).

PVince81:
using the `@appy-show` event fixes the issue, but nullifies the workarounds from https://github.com/nextcloud/nextcloud-vue/pull/831 that exist to prevent the popover to shake around before stabilizing around the trigger.

one hope would be to use the [onFirstUpdate hook](https://popper.js.org/docs/v2/lifecycle/#hook-into-the-lifecycle) of popper to only display the popover after the first positioning has finished

PVince81:
This timeout thing feels pointless now... while I managed to make the show logic work for both cases, with a timeout with less than 1000 ms the repositioning glitch is still happening sometimes.

Next up, trying to find a way to hook into popper js while using v-tooltip's directive.

PVince81:
Now I've tried to revert everything to the previous state.

I noticed that in the Talk app when the avatar menu is opening the first time, a scrollbar appears! This might explain the weird shifting of the popover. This means we might not need any workaround at all.

PVince81:
Ah, forgot to mention that I managed to capture Popper's "onUpdate", but updating doShowMenu at that moment failed to make the popup appear. Maybe it happens too late, or the offscreen workflow is disturbing position calculation... Somehow it feels like this approach is a dead end.

Next I'll look into the scrollbar thing.

PVince81:
Ohhhh... it seems before the initial show the tooltip just gets appended without position: absolute, and this is causing the scrollbars:
<img width="591" alt="image" src="https://user-images.githubusercontent.com/277525/105837372-e8e2a000-5fce-11eb-8595-1b0adf3f4e59.png">

PVince81:
I tried adding "overflow: hidden" on the body. While it makes the scrollbars disappear, it doesn't fix the repositioning glitch.
So it seems there are two glitches then and the popper/tooltip position gets updated three times:

[...]

Issue #1683, opened 2021-01-26
Avatar popover glitch when opened the first time


PVince81:
### Steps

1. Open Nextcloud Talk
2. Create a conversation with a few participants
3. Open the conversation
4. Open the participants list
5. Click on the three dots on the avatars
6. Then do so again a few times

### Expected result
Popover always properly positioned.

### Actual result
Popover glitches from another position the first time it's rendered.
Then it moves to the correct position.
Subsequent clicks keeps it in the right position.

https://user-images.githubusercontent.com/277525/105861602-2fe08d80-5fef-11eb-9e6a-dec104f88f73.mp4

### Version
Talk master 525071cbe8581ca728233768b5f242c2835d69f9
Nextcloud-vue 3.5.2

This issue was already known and I had a workaround in https://github.com/nextcloud/nextcloud-vue/pull/831 but sadly wasn't accurate enough. Due to regressions I reverted the workaround as it wasn't useful enough in https://github.com/nextcloud/nextcloud-vue/issues/1681.

So this ticket here is to do further attempts at fixing this.
Might require digging deeper into v-tooltip->popper.js to find out why the position is initially incorrect.

PR #1685, opened 2021-01-26
Revert workaround for popper glitch, fix regression

PVince81:
Revert the workaround that existed to dodge popper.js glitch where the
popover appeared in the wrong position first and a few milliseconds
later in the right place.

This fixes a regression where the regular Popover wasn't working any
more.

Fixes https://github.com/nextcloud/nextcloud-vue/issues/1681

The glitch will be addressed in https://github.com/nextcloud/nextcloud-vue/issues/1683

### DIFF
--- a/src/components/Popover/Popover.vue
+++ b/src/components/Popover/Popover.vue
@@ -52,7 +52,7 @@ With a `<button>` as a trigger:
 <template>
 	<VPopover
 		v-bind="$attrs"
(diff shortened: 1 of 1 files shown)


PVince81:
Original code was introduced in https://github.com/nextcloud/nextcloud-vue/pull/831 where the Avatar got switched to use the Popover component

ma12-co (approved):

dartcafe (approved):
trigger triggers again. 👍

PR #1708, opened 2021-02-12
Improve settings dialog layout

### LINKED ISSUE #1629: Settings modal padding and sizing
Two issues which look a bit off in the settings modal, cc @ma12-co 
- [ ] `.app-settings` has padding which results in the content being cut off on the top and bottom as well as the scrollbar to float in the air
- [ ] The modal is just quite narrow, the width should be increased / adapted to look nice on desktop as well.

cc @GretaD for reference since you use it in the Mail app. :)

ma12-co:
fix #1629 

https://user-images.githubusercontent.com/26852655/107752618-e98d6d00-6d16-11eb-9670-47ac9c2e0b42.mp4

### DIFF
--- a/src/components/AppSettingsDialog/AppSettingsDialog.vue
+++ b/src/components/AppSettingsDialog/AppSettingsDialog.vue
@@ -275,7 +275,6 @@ export default {
 }
 
 .app-settings {
-	padding: 28px 16px;
 	display: flex;
 	width: 100%;
 	&__navigation {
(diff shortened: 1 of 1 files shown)


GretaD (approved):
tested with mail app and works good

jancborchardt (approved):
Looks better! :)

@ma12-co for Talk specifically I’d say it would look even nicer if audio and video settings would be left and right of each other (on wider views). Then the dropdowns are not so super wide, and you see more other settings too.

PR #1713, opened 2021-02-18
Fix settings navigation layout

ma12-co:
Don't ellipsize caption if there's available space.


ChristophWurst (approved):
:+1: :elephant: 

would have appreciated screenshots but I think this css change is small enough so even I understand it

skjnldsv (approved):

skjnldsv:
> so even I understand it

:scream:

PR #1718, opened 2021-02-19
Improve multiselect tag

GretaD:
When we select an item from the multiselect dropdown, the item it is not shown in the input field while the dropdown is active. This bring a lot of confusion to users as they are not sure if they selected smth or not. This PR fixes that.

### DIFF
--- a/src/components/Multiselect/index.scss
+++ b/src/components/Multiselect/index.scss
@@ -25,6 +25,7 @@
 			cursor: text !important;
 			// remove border radius on bottom opening
 			border-radius: var(--border-radius) var(--border-radius) 0 0;
+			display: block !important;
 		}
 
 		/* multiselect__limit hidden if active */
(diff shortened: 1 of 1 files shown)


GretaD:
@skjnldsv I did what you suggested(btw we do need the limit) and the final result is this one. On mail i added :limit="4", but it doesnt apply, see [here](https://github.com/nextcloud/mail/pull/4401/files#diff-77b33550ce11bdb5cd04cb76105da60c54610d3c204ab3c20f513e2336a66d51R32).

it still only shows one. Any idea whats blocking it? I tried width and such but didnt work.

skjnldsv (changes requested):
Capture d’écran_2021-03-01_16-16-12
This seems to conflict with the normal layout :thinking:

GretaD:
@skjnldsv my last fixup will do smth like this, without breaking(if i tested it right) the other multiselect options. The only thing i cannot fix is, moving the input below and not having it inline, like in this fixup. 

Kazam-screencast-00012

src/components/Multiselect/index.scss
```diff
@@ -100,9 +102,8 @@
 				/* require to override the default width
 				and force the tag to shring properly */
 				min-width: 0;
-				max-width: 50%;
 				max-width: fit-content;
-				max-width: -moz-fit-content;
```
skjnldsv, line 105: Why was this removed?

skjnldsv (approved):
Other than the comment, it looks good on the docs and code

ChristophWurst (approved):
changes look fine

Issue #1689, opened 2021-01-30
Clicking on an IFrame causes error in v-click-outside middleware


sam-19:
## Description

The v-click-outside middleware in `@nextcloud/vue/dist/Mixins/excludeClickOutsideClasses.js` throws an error when clicking an embedded IFrame. The error message in browser console is
```
Uncaught TypeError: Cannot read property 'contains' of undefined
    at a.hasNodeOrAnyParentClass (excludeClickOutsideClasses.js:22)
    at a.clickOutsideMiddleware (excludeClickOutsideClasses.js:22)
    at r (v-click-outside.umd.js:1)
    at v-click-outside.umd.js:1
```
and the culprit seems to be this line in the source
https://github.com/nextcloud/nextcloud-vue/blob/4b9218a53378aa287fa750adc8525b5ed87f23d7/src/mixins/excludeClickOutsideClasses/index.js#L64

## Suggestion

The middleware method should first check if `node` has the property `classList` before calling `node.classList.contains()`.

skjnldsv:
Damn, this is a high priority.
Let me fix asap

u451f:
FYI this also affects OnlyOffice using the Community Server. Some documents cannot be opened even if they were created on the instance. Thanks for fixing it! Looking forward to the update :)

skjnldsv:
cc @juliushaertl

muhyal:
Hi, 

Any workarounds?

Thanks.

PR #1723, opened 2021-02-24
Make sure we can check the classList on the node

skjnldsv:
Fix https://github.com/nextcloud/nextcloud-vue/issues/1689

See https://github.com/nextcloud/files_pdfviewer/issues/281


eneiluj (approved):
This also affects the whiteboard integration app. I didn't try it but it seems good.

ma12-co (approved):

PR #1733, opened 2021-03-02
Allow to inject preloaded user statuses

### LINKED ISSUE #1428: Make user_status on avatar scale
We ran into serious issues in our conference due to the number of requests from the user_status feature.
E.g. in Talk we already have the information available, but there is no way to inject the information into the avatar, so each of the 500 avatars in the participant list will cause a request to the server.

nickvergessen:
Fix #1428 

Improves talk loading (by skipping 40+ status requests) when the status is preloaded already

### DIFF
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -172,6 +172,13 @@ export default {
 			default: true,
 		},
 		/**
+		 * When the user status was preloaded via another source it can be handed in with this property to save the request.
+		 */
+		preloadedUserStatus: {
+			type: Object,
+			default: undefined,
+		},
+		/**
 		 * Is the user a guest user (then we have to user a different endpoint)
 		 */
 		isGuest: {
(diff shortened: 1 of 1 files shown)


src/components/Avatar/Avatar.vue
```diff
@@ -171,6 +171,13 @@ export default {
 			type: Boolean,
 			default: true,
 		},
+		/**
+		 * When the user status was preloaded via another source it can be handed in with this property to save the request.
+		 */
+		preloadedUserStatus: {
```
PVince81, line 177: Not sure if we even need to call it "preloaded". Technically, all properties are preloaded.

We could say in the doc that if this property is not set, its data will be fetched automatically from the server.

PVince81 (approved):
Brilliant idea!

src/components/Avatar/Avatar.vue
nickvergessen: userStatus is already a data attribute in the mixin and I didn't want to mess around with that.

nickvergessen:
Talk demo branch https://github.com/nextcloud/spreed/pull/5288

src/components/Avatar/Avatar.vue
nickvergessen: Added more docs

PVince81: ah, indeed.

PVince81:
another thing comes to mind: if the status is managed externally, do we want to still subscribe to the "user status update" channel or expect the preloaded property to be changed by the consumer ?

not sure what mechanism is usually sending those user status updates through that channel

nickvergessen:
No preload is just for initial loading, afterwards the status is updated via the events so it shows/updates everywhere at the same time.

skjnldsv (approved):

PR #1739, opened 2021-03-03
Create ActionCaption component

ma12-co:
Upstreaming talk's `actioncaption`, as per the original @jenniferpiperek design

Screenshot from 2021-03-03 15-52-30

### DIFF
--- a/src/components/ActionCaption/ActionCaption.vue
+++ b/src/components/ActionCaption/ActionCaption.vue
@@ -1,0 +1,65 @@
+<!--
+  - @copyright Copyright (c) 2021 Marco Ambrosini <marcoambrosini@pm.me>
+  -
+  - @author Marco Ambrosini <marcoambrosini@pm.me>
+  -
+  - @license GNU AGPL version 3 or any later version
+  -
+  - This program is free software: you can redistribute it and/or modify
+  - it under the terms of the GNU Affero General Public License as
+  - published by the Free Software Foundation, either version 3 of the
+  - License, or (at your option) any later version.
+  -
+  - This program is distributed in the hope that it will be useful,
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


src/components/ActionCaption/ActionCaption.vue
```diff
@@ -0,0 +1,65 @@
...
+			type: String,
+			required: true,
+		},
+	},
+}
+</script>
+
+<style lang="scss" scoped>
+	.app-navigation-caption {
+		color: var(--color-primary);
```
nickvergessen, line 54: White on white incoming... Needs to use color-primary-element?

nickvergessen (commented):
I think the color is too much inside an action component. Its okay in the navigation component but not in the action popover menu

ma12-co:
> I think the color is too much inside an action component. Its okay in the navigation component but not in the action popover menu

Imo it creates a nice separation between actions and caption, but I don't have a strong opinion here. @jancborchardt  your call

skjnldsv (changes requested):
Waiting fro @jancborchardt's return

jancborchardt:
> > I think the color is too much inside an action component. Its okay in the navigation component but not in the action popover menu
> 
> Imo it creates a nice separation between actions and caption, but I don't have a strong opinion here. @jancborchardt your call

I would agree the color _and_ boldness is too much for just a caption – since the options themselves are the important thing. I would actually set the caption in `var(--color-text-maxcontrast)` (and possibly not bold) and use the color (and maybe the boldness, but probably not) for the selected option instead.

ma12-co:
Allright, ready to review again

Screenshot from 2021-04-16 09-19-54

skjnldsv (approved):
Let's use variables :rocket:

skjnldsv (approved):

PVince81 (approved):
:+1:

PVince81:
ohhh, a conflict!

PR #1740, opened 2021-03-05
Add Caption component

ma12-co:
Screenshot from 2021-03-05 09-51-00

### DIFF
--- a/src/components/Caption/Caption.vue
+++ b/src/components/Caption/Caption.vue
@@ -1,0 +1,57 @@
+<!--
+  - @copyright Copyright (c) 2021 Marco Ambrosini <marcoambrosini@pm.me>
+  -
+  - @author Marco Ambrosini <marcoambrosini@pm.me>
+  -
+  - @license GNU AGPL version 3 or any later version
+  -
+  - This program is free software: you can redistribute it and/or modify
+  - it under the terms of the GNU Affero General Public License as
+  - published by the Free Software Foundation, either version 3 of the
+  - License, or (at your option) any later version.
+  -
+  - This program is distributed in the hope that it will be useful,
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  - GNU Affero General Public License for more details.
+  -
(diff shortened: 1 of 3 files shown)


src/components/Caption/Caption.vue
```diff
@@ -0,0 +1,57 @@
...
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+-->
+
+<template>
+	<div role="heading" class="app-navigation-caption">
```
skjnldsv, line 23: I'm a bit confused by this. We already have a Caption component for the navigation.
Is this component navigation only? It's also missing documentation.

I think the name should be more straightforward, and maybe the component should be in a dedicated section of the docs? Like 'Text', where we could also put the highlight component as well? :thinking:

ma12-co: > Is this component navigation only

It's used in both navigation and sidebar in Talk

> It's also missing documentation

Oopsss

> I think the name should be more straightforward, and maybe the component should be in a dedicated section of the docs? Like 'Text', where we could also put the highlight component as well? thinking

Not sure I get this, quick call?

PVince81:
@ma12-co what's the difference with your other similarly titled PR ? https://github.com/nextcloud/nextcloud-vue/pull/1739

ma12-co:
> @ma12-co what's the difference with your other similarly titled PR ? #1739

That component is for use within actions only.

src/components/Caption/Caption.vue
ChristophWurst: > Not sure I get this, quick call?

For transparency and posterity it would be cool to have the infos, decisions and so on here with the PR.

PVince81: @ma12-co what was the results of the call ? Can you summarize ?

ma12-co: Waiting for @jancborchardt 
Discussion here #1751

ma12-co:
closing in favor of #1863

PR #1759, opened 2021-03-12
Resizable columns

brueckner:
Hi! So this is the new counterpart to a Pull Request originally made in https://github.com/nextcloud/mail/pull/4123. 

It was suggested to move the implemention into the global component, so that other apps are able to use this as well. Here are some details about the implementation:

This is (still) using the [splitpanes](https://antoniandre.github.io/splitpanes/) package to do the heavy lifting. We've got two columns: `main` and `aside`. These two columns have `min`, `max` and default `size` widths defined [here](https://github.com/nextcloud/nextcloud-vue/compare/master...dotplex:enhancement/resizable-columns?expand=1#diff-937d3d5f5c1cfb86c454474540a49bc24a57cdabe0876f60e4e0aa018a5b9223R79-R90).

`splitpanes` fires an event called `resized` after a resize has occurred, which triggers a save to `localStorage` with the `size` values for `main` and `aside` in [`onPaneResize()`](https://github.com/nextcloud/nextcloud-vue/compare/master...dotplex:enhancement/resizable-columns?expand=1#diff-937d3d5f5c1cfb86c454474540a49bc24a57cdabe0876f60e4e0aa018a5b9223R127-R133). The pane sizes are **scoped to the parent component**, which means that each app implementing this will store their own sizes automatically. To do that, I am pulling the parent component's name (`this.$parent.$options.name`) to construct a [`paneIdentifier`](https://github.com/nextcloud/nextcloud-vue/compare/master...dotplex:enhancement/resizable-columns?expand=1#diff-937d3d5f5c1cfb86c454474540a49bc24a57cdabe0876f60e4e0aa018a5b9223R78) which is then used to store and read from localStorage. 

So for example, the Mail app implements this insde of the `MailboxThread` component. Thus the record in `localStorage` will look something like this: 

```
pane-sizes.MailboxThread: {"aside":"23.60","main":"76.40"}
```

As for the DOM structure: the `<slot>` inside `<main id="app-content-vue">` has now been wrapped inside `<div id="app-content-wrapper">`, which had to be moved up from within the component of the Mail app. This is something that needs to be checked I guess. Does any other app use that `#app-content-wrapper` div?

Apps that implement this component and just use the default slot should not be affected by this change, other than the new wrapping div. Maybe we could even move the default `<slot>` out of div#app-content-wrapper again... 🤔  That way there wouldn't be any change at all to a component that is not using the named slots. I'd love some feedback on that!

Oh and also: can someone please check why there are so many changes/additions to the `package-lock.json` file? Just adding the splitpanes-package shouldn't create that big of a difference I believe. 🤷‍♂️

### DIFF
--- a/src/components/AppContent/AppContent.vue
+++ b/src/components/AppContent/AppContent.vue
@@ -68,7 +68,17 @@ The list size must be between the min and the max width value.
 
 <template>
 	<main id="app-content-vue" class="app-content no-snapper">
-		<div class="app-content-wrapper">
+		<!-- Mobile view does not allow resizeable panes -->
+		<div v-if="isMobile"
+			:class="showDetails ? 'app-content-wrapper--show-details' : 'app-content-wrapper--show-list'"
+			class="app-content-wrapper app-content-wrapper--mobile">
+			<AppDetailsToggle v-if="showDetails" @click.native.stop.prevent="hideDetails" />
+
+			<slot name="list" />
+			<slot />
+		</div>
+
+		<div v-else class="app-content-wrapper">
 			<Splitpanes
 				v-if="hasList"
 				class="default-theme"
@@ -97,11 +107,15 @@ The list size must be between the min and the max width value.
 </template>
 
 <script>
-import Hammer from 'hammerjs'
+import 'splitpanes/dist/splitpanes.css'
+
 import { emit } from '@nextcloud/event-bus'
 import { getBuilder } from '@nextcloud/browser-storage'
 import { Splitpanes, Pane } from 'splitpanes'
-import 'splitpanes/dist/splitpanes.css'
+import Hammer from 'hammerjs'
+
+import AppDetailsToggle from './AppDetailsToggle'
+import isMobile from '../../mixins/isMobile'
 
 const browserStorage = getBuilder('nextcloud').persist().build()
 
@@ -113,9 +127,12 @@ export default {
 	name: 'AppContent',
 
 	components: {
-		Splitpanes,
+		AppDetailsToggle,
 		Pane,
+		Splitpanes,
 	},
+
+	mixins: [isMobile],
 
 	props: {
 		/**
@@ -159,6 +176,18 @@ export default {
 		paneConfigKey: {
 			type: String,
 			default: appName,
+		},
+
+		/**
+		 * When in mobile view, only the list or the details are shown
+		 * If you provide a list, you need to provide a variable
+		 * that will be set to true by the user when an element of
+		 * the list gets selected. The details will then show a back
+		 * arrow to return to the list that will update this prop to false.
+		 */
+		showDetails: {
+			type: Boolean,
+			default: true,
 		},
 	},
 
@@ -263,6 +292,13 @@ export default {
 				return listPaneSize
 			}
 		},
+
+		/**
+		 * The user clicked the back arrow from the details view
+		 */
+		hideDetails() {
+			this.$emit('update:showDetails', false)
+		},
 	},
 }
 </script>
@@ -277,6 +313,26 @@ export default {
 	// Overriding server styles TODO: cleanup!
 	margin: 0 !important;
 	background-color: var(--color-main-background);
+}
+
+// Mobile list/details handling
+.app-content-wrapper--mobile {
+	&.app-content-wrapper--show-list ::v-deep {
+		.app-content-list {
+			display: initial !important;
+		}
+		.app-content-details {
+			display: none !important;
+		}
+	}
+	&.app-content-wrapper--show-details ::v-deep {
+		.app-content-list {
+			display: none !important;
+		}
+		.app-content-details {
+			display: initial !important;
+		}
+	}
 }
 
 ::v-deep .splitpanes.default-theme {
--- a/src/components/AppContent/AppContent.vue
+++ b/src/components/AppContent/AppContent.vue
@@ -21,17 +21,89 @@
  - along with this program. If not, see <http://www.gnu.org/licenses/>.
  -
  -->
+<docs>
+### General description
+
+This components provides a wrapper around the main app's content.
+
+Single-column layouts can just use the default slot. A resizable column
+can be added by providing content to the named slot `list`.
+
+### Examples
+
+#### Usage: Single-column content
+```vue
+<template>
+	<AppContent>
+		<h2>Single-column main content</h2>
+	</AppContent>
+</template>
+```
+
+#### Usage: Two resizable columns
+```vue
+<template>
+	<AppContent>
+		<template slot="list">
+			<div>Resizable list content</div>
+		</template>
+
+		<div>Main content</div>
+	</AppContent>
+</template>
+```
+
+#### Overriding Defaults
+The default, min and max sizes (in percent) of the resizable list column can be overridden.
+The list size must be between the min and the max width value.
+
+```
+<AppContent
+	list-size="35"
+	list-min-width="20"
+	list-max-width="45"
+>...</AppContent>
+```
+</docs>
 
 <template>
 	<main id="app-content-vue" class="app-content no-snapper">
-		<!-- @slot Provide content to the app content -->
-		<slot />
+		<div class="app-content-wrapper">
+			<Splitpanes
+				v-if="hasList"
+				class="default-theme"
+				@resized="handlePaneResize">
+				<Pane class="splitpanes__pane-list"
+					:size="listPaneSize || paneDefaults.list.size"
+					:min-size="paneDefaults.list.min"
+					:max-size="paneDefaults.list.max">
+					<!-- @slot Provide a list to the app content -->
+					<slot name="list" />
+				</Pane>
+
+				<Pane class="splitpanes__pane-details"
+					:size="detailsPaneSize"
+					:min-size="paneDefaults.details.min"
+					:max-size="paneDefaults.details.max">
+					<!-- @slot Provide the main content to the app content -->
+					<slot />
+				</Pane>
+			</Splitpanes>
+
+			<!-- @slot Provide the main content to the app content -->
+			<slot v-else />
+		</div>
 	</main>
 </template>
 
 <script>
 import Hammer from 'hammerjs'
 import { emit } from '@nextcloud/event-bus'
+import { getBuilder } from '@nextcloud/browser-storage'
+import { Splitpanes, Pane } from 'splitpanes'
+import 'splitpanes/dist/splitpanes.css'
+
+const browserStorage = getBuilder('nextcloud').persist().build()
 
 /**
  * App content container to be used for the main content of your app
@@ -40,12 +112,99 @@ import { emit } from '@nextcloud/event-bus'
 export default {
 	name: 'AppContent',
 
+	components: {
+		Splitpanes,
+		Pane,
+	},
+
 	props: {
-		// Allows to disable the control by swipe of the app navigation open state
+		/**
+		 * Allows to disable the control by swipe of the app navigation open state
+		 */
 		allowSwipeNavigation: {
 			type: Boolean,
 			default: true,
 		},
+
+		/**
+		 * Allows you to set the default width of the resizable list in %
+		 * Must be between listMinWidth and listMaxWidth
+		 */
+		listSize: {
+			type: Number,
+			default: 20,
+		},
+
+		/**
+		 * Allows you to set the minimum width of the list column in %
+		 */
+		listMinWidth: {
+			type: Number,
+			default: 15,
+		},
+
+		/**
+		 * Allows you to set the maximum width of the list column in %
+		 */
+		listMaxWidth: {
+			type: Number,
+			default: 40,
+		},
+
+		/**
+		 * Specify the config key for the pane config sizes
+		 * Default is the global var appName if you use the webpack-vue-config
+		 * Otherwise it will be undefined and you will have to provide one
+		 */
+		paneConfigKey: {
+			type: String,
+			default: appName,
+		},
+	},
+
+	data() {
+		return {
+			contentHeight: 0,
+
+			hasList: false,
+			listPaneSize: this.restorePaneConfig(),
+		}
+	},
+
+	computed: {
+		paneConfigID() {
+			// Using the webpack-vue-config, appName is a global variable
+			return `pane-list-size-${this.paneConfigKey}`
+		},
+
+		detailsPaneSize() {
+			if (this.listPaneSize) {
+				return 100 - this.listPaneSize
+			}
+			return this.paneDefaults.details.size
+		},
+
+		paneDefaults() {
+			return {
+				list: {
+					size: this.listSize,
+					min: this.listMinWidth,
+					max: this.listMaxWidth,
+				},
+
+				// set the inverse values of the details column
+				// based on the provided (or default) values of the list column
+				details: {
+					size: 100 - this.listSize,
+					min: 100 - this.listMaxWidth,
+					max: 100 - this.listMinWidth,
+				},
+			}
+		},
+	},
+
+	updated() {
+		this.checkListSlot()
 	},
 
 	mounted() {
@@ -53,10 +212,15 @@ export default {
 			this.mc = new Hammer(this.$el, { cssProps: { userSelect: 'text' } })
 			this.mc.on('swipeleft swiperight', this.handleSwipe)
 		}
+
+		this.checkListSlot()
+		this.restorePaneConfig()
 	},
+
 	beforeDestroy() {
 		this.mc.off('swipeleft swiperight', this.handleSwipe)
 	},
+
 	methods: {
 		// handle the swipe event
 		handleSwipe(e) {
@@ -72,6 +236,31 @@ export default {
 				emit('toggle-navigation', {
 					open: false,
 				})
+			}
+		},
+
+		handlePaneResize(event) {
+			const listPaneSize = parseInt(event[0].size, 10)
+			browserStorage.setItem(this.paneConfigID, JSON.stringify(listPaneSize))
+			this.listPaneSize = listPaneSize
+			console.debug('AppContent pane config', listPaneSize)
+		},
+
+		// $slots is not reactive, we need to update this manually
+		checkListSlot() {
+			const hasListSlot = !!this.$slots.list
+			if (this.hasList !== hasListSlot) {
+				this.hasList = hasListSlot
+			}
+		},
+
+		// browserStorage is not reactive, we need to update this manually
+		restorePaneConfig() {
+			const listPaneSize = parseInt(browserStorage.getItem(this.paneConfigID), 10)
+			if (!isNaN(listPaneSize) && listPaneSize !== this.listPaneSize) {
+				console.debug('AppContent pane config', listPaneSize)
+				this.listPaneSize = listPaneSize
+				return listPaneSize
 			}
 		},
 	},
@@ -90,4 +279,44 @@ export default {
 	background-color: var(--color-main-background);
 }
 
+::v-deep .splitpanes.default-theme {
+	.app-content-list {
+		max-width: none;
+	}
+
+	.splitpanes__pane {
+		background-color: transparent;
+		transition: none;
+
+		&-list {
+			min-width: 200px;
+			position: sticky;
+			top: var(--header-height);
+
+			@media only screen and (max-width: $breakpoint-mobile) {
+				display: none;
+			}
+		}
+
+		&-details {
+			overflow-y: scroll;
+
+			@media only screen and (max-width: $breakpoint-mobile) {
+				min-width: 100%;
+			}
+		}
+	}
+
+	.splitpanes__splitter {
+		width: 9px;
+		margin-left: -5px;
+		background-color: transparent;
+		border-left: none;
+
+		&:before,
+		&:after {
+			display: none;
+		}
+	}
+}
 </style>
--- a/src/components/AppContent/AppContent.vue
+++ b/src/components/AppContent/AppContent.vue
@@ -72,7 +72,7 @@ The list size must be between the min and the max width value.
 		<div v-if="isMobile"
 			:class="showDetails ? 'app-content-wrapper--show-details' : 'app-content-wrapper--show-list'"
 			class="app-content-wrapper app-content-wrapper--mobile">
-			<AppDetailsToggle v-if="showDetails" @click.native.stop.prevent="hideDetails" />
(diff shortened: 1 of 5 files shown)


st3iny:
First of all, I'm really looking forward to getting this one merged :)

> Oh and also: can someone please check why there are so many changes/additions to the package-lock.json file? Just adding the splitpanes-package shouldn't create that big of a difference I believe. man_shrugging

Seems like you accidentally upgraded the lock file to version 2. There are 2 options for fixing:
1. Use npm <= v6 (lock file version 2 is used in >= v7), recreate and push the lock again.
2. We upgrade the lock file of the repository to v2 (similar to what we did in mail not long ago).

brueckner:
Ha - that was it, thanks man! :) My npm was on version `v7.5.4`. Switching to `6x`has solved it. 👌

ChristophWurst (approved):
Tested, works well and the changes make sense

src/components/AppContent/AppContent.vue
```diff
@@ -48,6 +73,31 @@ export default {
...
+					min: 60,
+					max: 85,
+				},
+			},
+		}
+	},
+
+	computed: {
+		paneSizes() {
+			const clientPaneSizes = localStorage.getItem(this.paneIdentifier)
```
skjnldsv, line 96: Hey, `localStorage` might not be available for every browser.
Could you use the [@nextcloud/browser-storage](https://www.npmjs.com/package/@nextcloud/browser-storage) package please? :)

src/components/AppContent/AppContent.vue
```diff
@@ -24,14 +24,34 @@
...
+			<!-- @slot Provide content to the app content -->
+			<slot />
+			<Splitpanes
+				class="default-theme"
+				@resized="onPaneResize">
+				<Pane class="aside"
+					:size="paneSizes.aside || paneDefaults.aside.size"
+					:min-size="paneDefaults.aside.min"
+					:max-size="paneDefaults.aside.max">
+					<slot name="aside" />
```
skjnldsv, line 37: Please use a different keyword than `aside`, we already have one and it's for the app-sidebar :)

Even if it's not technically an aside per se, I feel it's misleading! :wink:

src/components/AppContent/AppContent.vue
```diff
@@ -90,4 +147,47 @@ export default {
 	background-color: var(--color-main-background);
 }
 
+#app-content-wrapper {
+	overflow: hidden;
+	height: calc(100vh - var(--header-height));
+}
```
skjnldsv, line 153: Sorry, the `calc(100vh - var(--header-height));` trick is not acceptable. Some mobile browsers does not count the bottom touch bar in the `100vh` meaning it will still scroll vertically with the `100vh-headerheight`. You need to comply to the inherited height and the `display flex`.

skjnldsv (changes requested):
Thanks a lot for addressing this :)
I added some important comments :rocket:

Lerain:
> Thanks a lot for addressing this :)
> 
> I added some important comments :rocket: 

Thanks for the feedback John! I'll get on it tomorrow. :)

brueckner:
> Thanks a lot for addressing this :)
> I added some important comments 🚀

Thanks for the review & feedback John @skjnldsv, I'll get on it tomorrow! :)

src/components/AppContent/AppContent.vue
brueckner: Didn't know about that! I love that it has the exact same API, done! 👌

src/components/AppContent/AppContent.vue
brueckner: I changed it to `sidebar`, or do you have a less generic suggestion?

src/components/AppContent/AppContent.vue
brueckner: Good call. I've seen it used like that in `AppSidebar` and thought it was okay. It does need a specific height though, otherwise it'll mess up the individual scrolling of the sidebar and the main content. I went for a JS-calculated approach which mimics `calc(100vh - var(--header-height))` which looks like this: `window.innerHeight - document.getElementById('header').offsetHeight`. It is set on mount and updated an `window.resize` (new eventListener added). If there's a CSS-only solution I am missing here, please let me know!

brueckner:
A little later than expected, but today I was able to implement your suggestions @skjnldsv.     
Could you please have another look at it?

Cheers!

src/components/AppContent/AppContent.vue
```diff
@@ -24,14 +24,37 @@
...
+			<slot />
+			<Splitpanes
+				class="default-theme"
+				@resized="handlePaneResize">
+				<Pane class="sidebar"
+					:size="paneSizes.sidebar || paneDefaults.sidebar.size"
+					:min-size="paneDefaults.sidebar.min"
+					:max-size="paneDefaults.sidebar.max">
+					<slot name="sidebar" />
+				</Pane>
```
skjnldsv, line 38: Sidebar is what we have on the far right of the apps, please use something like `list` or something?

Also, what happens if you don't  have a list? Do we really need the Pane component to be used in that case?
Is it ignore? Maybe it would be better to have a v-if and v-else if there are no list slot to just output the main content?

src/components/AppContent/AppContent.vue
```diff
@@ -24,14 +24,37 @@
 
 <template>
 	<main id="app-content-vue" class="app-content no-snapper">
-		<!-- @slot Provide content to the app content -->
-		<slot />
+		<div id="app-content-wrapper" :style="`height: ${contentHeight}px;`">
+			<!-- @slot Provide content to the app content -->
```
skjnldsv, line 28: The documentation is misleading. It make it seems  like you are using this default slot to put the main content in here.

src/components/AppContent/AppContent.vue
```diff
@@ -24,14 +24,37 @@
...
+					:size="paneSizes.sidebar || paneDefaults.sidebar.size"
+					:min-size="paneDefaults.sidebar.min"
+					:max-size="paneDefaults.sidebar.max">
+					<slot name="sidebar" />
+				</Pane>
+				<Pane class="main"
+					:size="paneSizes.main || paneDefaults.main.size"
+					:min-size="paneDefaults.main.min"
+					:max-size="paneDefaults.main.max">
+					<slot name="main" />
```
skjnldsv, line 43: Please add documentation for this slot

src/components/AppContent/AppContent.vue
```diff
@@ -24,14 +24,37 @@
...
+			<!-- @slot Provide content to the app content -->
+			<slot />
+			<Splitpanes
+				class="default-theme"
+				@resized="handlePaneResize">
+				<Pane class="sidebar"
+					:size="paneSizes.sidebar || paneDefaults.sidebar.size"
+					:min-size="paneDefaults.sidebar.min"
+					:max-size="paneDefaults.sidebar.max">
+					<slot name="sidebar" />
```
skjnldsv, line 37: Please add documentation for this slot

src/components/AppContent/AppContent.vue
skjnldsv, line 38: Sidebar is what we have on the far right of the apps, please use something like list or something?

Also, what happens if you don't have a list? Do we really need the Pane component to be used in that case?
Is it ignore? Maybe it would be better to have a v-if and v-else if there are no list slot to just output the main content?

src/components/AppContent/AppContent.vue
skjnldsv, line 28: The documentation is misleading. It make it seems like you are using this default slot to put the main content in here.

src/components/AppContent/AppContent.vue
```diff
@@ -24,14 +24,37 @@
 
 <template>
```
skjnldsv, line 25: Can you add a `<doc />` section here to describe how we can  and should use this component? :)
You can find examples on others components like https://github.com/nextcloud/nextcloud-vue/blob/5a3270028d4b958fe0d31866d9fdb7a15d8a036c/src/components/SettingsSection/SettingsSection.vue#L24-L41

skjnldsv (changes requested):

skjnldsv:
Getting close @brueckner ! Thank you so much! It looks great! :hugs:

src/components/AppContent/AppContent.vue
brueckner: Hmmm. Well "list" is pretty specific to the usecase. It could be used for a preview of something for all we know. 

I like the idea with the v-if. What if we did that and verbosely call it `resizable-column`? 

So in case of a single column usage, we'd just have the default slot and if they want to add a resizable column, they just add that one named slot to it:

```
# Single main column
<AppContent>
   <div>Main content</div>
</AppContent>


# Two resizable columns
<AppContent>
   <div>Main content</div>
   <template slot="resizable-column">List of things</template>
</AppContent>
```

💡 ⚡ 

What do you think?

skjnldsv: Not really, we only have two allowed use-cases here: See my comment in https://github.com/nextcloud/mail/pull/4123#issuecomment-795618458

Either 
```vue
<AppContent>
	<AppContentList #list><!-- Mails listing --></AppContentList>
	<AppContentDetails #details><!-- Mail details --></AppContentDetails>
</AppContent>
```
Or
```vue
<AppContent>
	<!-- Main content -->
</AppContent>
```

skjnldsv: Also see https://docs.nextcloud.com/server/21/developer_manual/design/content.html

brueckner:
Alright - the new update changes a few things:

### Slots

There's a default slot and a `list` slot. If the `list` slot is not used, the component will behave just like before. If the `list` slot is used, the component will initialize the resizable columns feature. 

### Overriding defaults

I refactored the default settings used by the resizable columns. They can now be overriden by the parent component with 3 simple props:

```
<AppContent
    list-size="35"          # default: 25
    list-min-width="20"     # default: 15
    list-max-width="45"     # default: 40
>...</AppContent>
```

The inverse defaults for the `details` column are calculated automatically.

### Documentation

Has been added with two examples and a section about overriding defaults.

---

Thanks a ton @skjnldsv, this is now **much** nicer to use because of your input!

If there's anything else, just let me know!

src/components/AppContent/AppContent.vue
brueckner: Alright, I see. I changed it to list and details. ;)

src/components/AppContent/AppContent.vue
brueckner: Done!

src/components/AppContent/AppContent.vue
brueckner: Has been refactored and the (old) documentation has been removed in favor of the `<doc>` block.

src/components/AppContent/AppContent.vue
```diff
@@ -74,6 +190,30 @@ export default {
...
 		},
+		handlePaneResize(event) {
+			const paneSizes = {
+				list: event[0].size.toFixed(2),
+				details: event[1].size.toFixed(2),
+			}
+			browserStorage.setItem(this.paneIdentifier, JSON.stringify(paneSizes))
+		},
+		setContentHeight() {
+			this.contentHeight = window.innerHeight - document.getElementById('header').offsetHeight
```
raimund-schluessler, line 201: This currently breaks the docs on netlify:
```
TypeError: document.getElementById(...) is null
    setContentHeight webpack:///./src/components/AppContent/AppContent.vue?./node_modules/babel-loader/lib!./node_modules/vue-loader/lib??vue-loader-options:219
    mounted webpack:///./src/components/AppContent/AppContent.vue?./node_modules/babel-loader/lib!./node_modules/vue-loader/lib??vue-loader-options:186
    VueJS 14
    setCompiledPreview webpack:///./node_modules/vue-styleguidist/lib/client/rsg-components/Preview/PreviewAsync.js?:291
    componentDidMount webpack:///./node_modules/vue-styleguidist/lib/client/rsg-components/Preview/PreviewAsync.js?:138
    React 2
    unstable_runWithPriority webpack:///./node_modules/scheduler/cjs/scheduler.production.min.js?:19
    React 9
    render webpack:///./node_modules/vue-styleguidist/lib/client/index.js?:46
    <anonymous> webpack:///./node_modules/vue-styleguidist/lib/client/index.js?:54
    js https://deploy-preview-1759--nextcloud-vue-components.netlify.app/build/bundle.1310bebb.js:13447
    __webpack_require__ https://deploy-preview-1759--nextcloud-vue-components.netlify.app/build/bundle.1310bebb.js:64
    <anonymous> webpack:///multi_./styleguide/global.requires.js_./styleguide/assets/icons.css_./styleguide/assets/additional.scss_./node_modules/vue-styleguidist/lib/client/index?:4
    0 https://deploy-preview-1759--nextcloud-vue-components.netlify.app/build/bundle.1310bebb.js:19711
    __webpack_require__ https://deploy-preview-1759--nextcloud-vue-components.netlify.app/build/bundle.1310bebb.js:64
    <anonymous> https://deploy-preview-1759--nextcloud-vue-components.netlify.app/build/bundle.1310bebb.js:199
    <anonymous> https://deploy-preview-1759--nextcloud-vue-components.netlify.app/build/bundle.1310bebb.js:202
vue.esm.js:1743:13
```

raimund-schluessler (changes requested):
See https://github.com/nextcloud/nextcloud-vue/pull/1759#discussion_r608389022

raimund-schluessler:
Will this also allow to resize the AppSidebar? This would be very neat as well and even more apps would benefit from this. 🙂

skjnldsv:
> Will this also allow to resize the AppSidebar? This would be very neat as well and even more apps would benefit from this.

Nope. Might come later™ ?

skjnldsv:
Hey @brueckner! Did you have a bit of time to look at the other comments? :)

brueckner:
> Hey @brueckner! Did you have a bit of time to look at the other comments? :)

Oh, damn - yes and no. I have seen it, but have been distracted by other things and then forgot about it. I'll reply in the thread of @raimund-schluessler's comment. Thanks for the heads-up!

src/components/AppContent/AppContent.vue
brueckner: Hi and thanks for the feedback! How would I be able to reproduce this? 

Fixing this should be simple but I don't quite understand why this is happening in the first place, as this would only happen in case of a missing header (which seems unlikely). Any insight into why the build process is doing things differently than the regular build process?

brueckner: Well, like I said - I have no way of reproducing this so I am flying blind, but I made a change that makes the offending line a little more verbose but should get rid of the error you're seeing. Otherwise this PR is gonna be still open next christmas :P

brueckner:
I made a little update (see https://github.com/nextcloud/nextcloud-vue/pull/1759#discussion_r627966172) which should not break the docs on netlify anymore. I also made a reabase on master with no conflicts to solve. 👌

src/components/AppContent/AppContent.vue
```diff
@@ -40,22 +107,71 @@ import { emit } from '@nextcloud/event-bus'
...
+	},
+
 	props: {
 		// Allows to disable the control by swipe of the app navigation open state
 		allowSwipeNavigation: {
 			type: Boolean,
 			default: true,
 		},
+		// Allows you to set the default width of the resizable list in %
+		// Must be between listMinWidth and listMaxWidth
```
raimund-schluessler, line 122:
```suggestion
		/**
		 * Allows you to set the default width of the resizable list in %
		 * Must be between listMinWidth and listMaxWidth
		 */
```
This way it is parsed correctly by netlify and appears in the docs.

src/components/AppContent/AppContent.vue
```diff
@@ -40,22 +107,71 @@ import { emit } from '@nextcloud/event-bus'
...
+		listSize: {
+			type: Number,
+			required: false,
+			default: 25,
+		},
+		// Allows you to set the minimum width of the list column in %
```
raimund-schluessler, line 128:
```suggestion
		/**
		 * Allows you to set the minimum width of the list column in %
		 */
```

src/components/AppContent/AppContent.vue
```diff
@@ -40,22 +107,71 @@ import { emit } from '@nextcloud/event-bus'
...
+		listMinWidth: {
+			type: Number,
+			required: false,
+			default: 15,
+		},
+		// Allows you to set the maximum width of the list column in %
```
raimund-schluessler, line 134:
```suggestion
		/**
		 * Allows you to set the maximum width of the list column in %
		 */
```

src/components/AppContent/AppContent.vue
```diff
@@ -21,17 +21,84 @@
...
+
+#### Overriding Defaults
+The default, min and max sizes (in percent) of the resizable list column can be overridden.
+The list size must be between the min and the max width value.
+
+```
+<AppContent
+	list-size="35"          # default: 25
+	list-min-width="20"     # default: 15
+	list-max-width="45"     # default: 40
```
raimund-schluessler, line 64:
```suggestion
	list-size="35"
	list-min-width="20"
	list-max-width="45"
```
These comments lead to a `String contains invalid character error` on netlify and are not necessary, as the default can be seen from the props.

src/components/AppContent/AppContent.vue
```diff
@@ -40,22 +107,71 @@ import { emit } from '@nextcloud/event-bus'
 export default {
 	name: 'AppContent',
 
+	components: {
+		Splitpanes,
+		Pane,
+	},
+
 	props: {
 		// Allows to disable the control by swipe of the app navigation open state
```
raimund-schluessler, line 116:
```suggestion
		/**
		 * Allows to disable the control by swipe of the app navigation open state
		 */
```

raimund-schluessler (commented):
Code looks good so far. I didn't test how this works in a real app yet, but I will do so at the weekend. Especially the new `<div id="app-content-wrapper" />` wrapper around the main content needs good testing and the resize of the list might interfere with the `Breadcrumbs` component.

skjnldsv (approved):
Aside from other comments that needs fixing, good for me.

What shall we do with the other `AppContentDetails` and `AppContentList`? They are necessary now, right?

raimund-schluessler (dismissed):
See https://github.com/nextcloud/nextcloud-vue/pull/1759#issuecomment-836325106

brueckner:
First of all, thanks for taking the time to review this!

I like the idea of taking the default slot out of the app-content-wrapper. In fact, I have been thinking about that, too. I am currently trying to set up the tasks app, but I'm getting errors when creating tasks so that's holing me off at the moment. :/

raimund-schluessler:
> 
> 
> First of all, thanks for taking the time to review this!
> 
> I like the idea of taking the default slot out of the app-content-wrapper. In fact, I have been thinking about that, too. I am currently trying to set up the tasks app, but I'm getting errors when creating tasks so that's holing me off at the moment. :/

Let me know if I can help.

raimund-schluessler:
> 
> 
> > > First of all, thanks for taking the time to review this!
> > > I like the idea of taking the default slot out of the app-content-wrapper. In fact, I have been thinking about that, too. I am currently trying to set up the tasks app, but I'm getting errors when creating tasks so that's holing me off at the moment. :/
> > 
> > 
> > Let me know if I can help.
> > Edit: The Tasks app currently needs the latest nc/vue master to be build/run correctly, because I use some yet unreleased features for next version. So you need to run `npm link @nextcloud/vue` in the Tasks directory. Maybe that helps already.
> 
> I did link it, but that didn't help unfortunately.
> 
> Screenshot - Google Chrome Canary - 2021-05-10 at 16-21
> 
> So the `405` on that `remote.php/dav/` resource is wierd and it is triggered right away on load. It also happens in the calendar app, so I am not sure if it's actually related to the error that gets triggered when trying to add a task, but maybe you've seen that before somehow.

That seems to be a server issue. Somehow the DAV app does not responde correctly. But I don't really know why. It should be enabled and set up by default.

skjnldsv:
@raimund-schluessler what's the status here? We need to get this for 4.0.0 :)

raimund-schluessler:
> @raimund-schluessler what's the status here? We need to get this for 4.0.0 :)

I don't know the current status for sure. But I think there were no changes since I last had a look. Back then, it completely broke the layout of the Tasks app (and I guess every other app which uses scrolling on body). So this definitely needs to be fixed.

brueckner:
I'm gonna get back on it this week. I left off unable to fix the issue I was having with the Tasks app itself and thus unable to tackle the actual issue, which was pretty frustrating tbh. I'll report back once I've had a fresh look at it.

skjnldsv:
> I'm gonna get back on it this week

I'm on it, this has become quite urgent :)

skjnldsv:
Done! I fixed some of the vue logic and implemented a very important toggle between list and details for mobile view.

```vue
	<AppContent :show-details="showDetails" @update:showDetails="hideDetails">
		<!-- contacts list -->
		<template #list>
			<ContactsList
				:list="contactsList"
				:contacts="contacts"
				:search-query="searchQuery" />
		</template>

		<!-- main contacts details -->
		<ContactDetails :contact-key="selectedContact" />
	</AppContent>
```

Here `showDetails` is a computed that is true when an item of the list is selected
hideDetails just unselect the item of the list (e.g. a contact)

skjnldsv:
Peek 03-06-2021 12-14

skjnldsv:
Damn, I hate when github prevent me from pushing to a fork of a pr!

raimund-schluessler:
I can't test this until Monday.

skjnldsv:
> I can't test this until Monday.

If it works for contacts with a full-body scroll, I see no reason it should not for you :thinking:

skjnldsv:
> Damn, I hate when github prevent me from pushing to a fork of a pr!

Ah, it seems that you have not enabled us to push back to your branch @brueckner. Could you do that please?

brueckner:
Nice, good job John! :) You should have write access to the repo now.

skjnldsv:
> You should have write access to the repo now

You can just enable us to edit this pull request (bottom right checkbox of this PR)

skjnldsv (approved):

skjnldsv:
@raimund-schluessler tasks seems ok now
Peek 03-06-2021 18-41

raimund-schluessler:
@skjnldsv Seems to work well for Tasks now (I just checked that it doesn't break the app, I didn't check the functionality of the resizable columns, though).

raimund-schluessler (approved):

Issue #1760, opened 2021-03-12
Actions menu autofocus on elements doesn't work initially


PVince81:
### Steps

1. Tested in the Talk app
2. Focus on the new message form
3. Use shift+tab to tab up to the three dots menu.
4. Hit enter, then try to move up and down with the keyboard.

### Expected result
Able to move up/down

### Actual result
The item is not focussed.
Querying `document.activeElement` shows that the focus is still on the trigger.

Opening the menu a second time makes the issue disappear.

This is because the first render of the popover doesn't become visible directly and has an arbitrary delay defined by `requestAnimationFrame`, see my upstream report for details: https://github.com/Akryum/v-tooltip/issues/661

### Workaround

Depending whether upstream will accept a PR that emits a new event or not, there are other possible workarounds:

1. Extend our `Popover` component to monitor the DOM periodically in the `@show` handler, and only once the `.open` class appears, trigger another event like `afterShow`. Only do so if there's a handler set there.
2. Like 1 but do the periodic monitoring in the Actions class
3. Like 2 but simply retry focussing the first element if it failed

So far I prefer approach 1 as a temporary workaround until v-tooltip is fixed upstream.

PVince81:
no need to monitor the DOM, I managed to make it work by watching `this.$refs.popover.isOpen`

PR here: https://github.com/nextcloud/nextcloud-vue/pull/1761

PR #1761, opened 2021-03-12
Fix actions dropdown focus timing

PVince81:
Because "@after-show" is still happening too early due to v-tooltip's
use of "requestAnimationFrame" we can't rely on its events to tell us
when the popover is actually visible.

This workaround introduces an additional Popover event "after-show" and
"after-hide" that are based on monitoring the inner "isOpen" property
which is the one toggling the CSS visibility.

The Actions popover logic now uses "after-show" to reliably apply the
focus on the first action element.

Fixes https://github.com/nextcloud/nextcloud-vue/issues/1760
See https://github.com/Akryum/v-tooltip/issues/661 for the upstream issue.

### DIFF
--- a/src/components/Popover/Popover.vue
+++ b/src/components/Popover/Popover.vue
@@ -86,8 +86,18 @@ export default {
 			},
 			(val) => {
 				if (val) {
+					/**
+					 * Triggered after the tooltip was visually displayed.
+					 *
+					 * This is different from the 'show' and 'apply-show' which
+					 * run earlier than this where there is no guarantee that the
+					 * tooltip is already visible and in the DOM.
+					 */
 					this.$emit('after-show')
 				} else {
+					/**
+					 * Triggered after the tooltip was visually hidden.
+					 */
 					this.$emit('after-hide')
 				}
 			}
--- a/src/components/Popover/Popover.vue
+++ b/src/components/Popover/Popover.vue
@@ -51,6 +51,7 @@ With a `<button>` as a trigger:
 
 <template>
 	<VPopover
+		ref="popover"
 		v-bind="$attrs"
 		popover-base-class="popover"
 		popover-wrapper-class="popover__wrapper"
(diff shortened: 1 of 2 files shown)


src/components/Popover/Popover.vue
```diff
@@ -74,6 +75,24 @@ export default {
...
+	mounted() {
+		this.$watch(
+			() => {
+				// required because v-tooltip doesn't provide events
+				// and @show is too early
+				// see https://github.com/Akryum/v-tooltip/issues/661
+				return this.$refs.popover.isOpen
+			},
+		  (val) => {
+			  if (val) {
```
PVince81, line 88: weird indents... fixing...

PVince81:
tldr; fixes keyboard navigation for after opening the menu :smile:

src/components/Popover/Popover.vue
```diff
@@ -74,6 +75,24 @@ export default {
...
+		this.$watch(
+			() => {
+				// required because v-tooltip doesn't provide events
+				// and @show is too early
+				// see https://github.com/Akryum/v-tooltip/issues/661
+				return this.$refs.popover.isOpen
+			},
+			(val) => {
+				if (val) {
+					this.$emit('after-show')
```
skjnldsv, line 89: Please document those events

skjnldsv (changes requested):

ma12-co (approved):

PVince81:
@skjnldsv docs added

src/components/Popover/Popover.vue
skjnldsv: @PVince81 you can document them for our styleguide directly, no need for a readme :)
https://vue-styleguidist.github.io/docs/Documenting.html#events

skjnldsv (changes requested):

PVince81:
@skjnldsv nice, thanks for the hint. I've reverted the bits in the header and added docs close to `$emit`. Then squashed into the last commit.

Now it looks like this:
image

skjnldsv (approved):
giphy

PR #1767, opened 2021-03-18
Fix compact sidebar subtitle not visible

raimund-schluessler:
This fixes a CSS rule interference with a rule from server which lead to the sidebar subtitle not being visible in compact mode. This was originally reported here https://github.com/nextcloud/server/issues/24138

Explanation:

This rule
```css
#content[class*="app-"] * {
    box-sizing: border-box;
}
```
from server took precedence over the `box-sizing: content-box;` from the components, which, in combination with the fixed `height: $desc-height;`, lead to the height of the description being wrong.

I removed the fixed height (and the overwritten `border-box`), which should solve the issue. The alternative would be to set `box-sizing: content-box !important;` or to set a correct `height: $desc-height + 2*$desc-vertical-padding;`. But I think simply removing the height is cleaner.

It might be that we need to commit new base images for the visual regression tests, because the sidebar height changes by 2px because of this fix (this is because the subtitle height is given as 22px, but since it is not enforced by actually setting the height of the subtitle, the subtitle is 24px high in reality).

Please have a look @skjnldsv.

Edit:

Before:
Screenshot_2021-03-19 Dokumente - Dateien - Nextcloud

After:
Screenshot_2021-03-19 Dokumente - Dateien - Nextcloud(1)


raimund-schluessler:
@skjnldsv Six regression tests fail, with 0.013 difference. Could you have a look and upload new reference images if necessary? I think you did it the last time and the base images generated locally on my system differ from the ones the remote tests generate.

skjnldsv:
> Would we need to release a vue components 3.2.1 and 2.6.10

yes

skjnldsv:
https://github.com/nextcloud/nextcloud-vue/tree/stable2

skjnldsv:
We need to create the stable3 branch too

raimund-schluessler:
> 
> 
> https://github.com/nextcloud/nextcloud-vue/tree/stable2

Hm, the nextcloud-vue stable2 version is 2.9.0. Server stable 20 still lacks behind quite a bit. Is it realistic that server will get an update?

skjnldsv:
> Is it realistic that server will get an update?

depends on what the changelog between the two is?
but if there is no breaking changes, sure!

ma12-co (approved):

skjnldsv (approved):

skjnldsv:
> Hm, the nextcloud-vue stable2 version is 2.9.0. Server stable 20 still lacks behind quite a bit. Is it realistic that server will get an update?

https://github.com/nextcloud/nextcloud-vue/compare/v2.6.9...v2.9.0

skjnldsv:
Seems good enough!

skjnldsv:
@raimund-schluessler  care to backport? :see_no_evil:

raimund-schluessler:
> @raimund-schluessler  care to backport? :see_no_evil: 

@skjnldsv Thanks for taking care of the cypress screenshots! I will have a look at the backport.

raimund-schluessler:
@skjnldsv Backport is in https://github.com/nextcloud/nextcloud-vue/pull/1789.

By the way, updating the server CSS in https://github.com/nextcloud/nextcloud-vue/pull/1537 broke some other stuff from the sidebar too.

- the subtitle was cut at the bottom when the title was edited

| Before #1537 | After #1537 |
| ------------- |:-------------:|
| Screenshot_2021-03-19 Dokumente - Dateien - Nextcloud | Screenshot_2021-03-19 Dokumente - Dateien - Nextcloud |

- and the text is moving when enabling the edit mode at the moment, which was not the case before #1537.
editing

The cut subtitle was coincidentally fixed by this PR #1767 here 🙈 But the text is still moving when enabling the edit mode (which is a nitpick only). I will still create an issue for the moving text and have a look.

PR #1771, opened 2021-03-20
Allow last crumb to shrink if all crumbs collapsed

raimund-schluessler:
This PR allow the last breadcrumb in the breadcrumbs bar to shrink, in case all other breadcrumbs are already hidden. This prevents the breadcrumb from overflowing on narrow screens (e.g. iPhone 8).

Before:
Screenshot_2021-03-20 Inventar - Nextcloud

before

After:
Screenshot_2021-03-20 Inventar - Nextcloud(1)

after

### DIFF
--- a/src/components/Breadcrumbs/Breadcrumbs.vue
+++ b/src/components/Breadcrumbs/Breadcrumbs.vue
@@ -502,7 +502,7 @@ export default {
 			: []
 		this.addCrumbs(crumbs, crumbs2, crumbs1.length)
 
-		return createElement('div', { class: 'breadcrumb', ref: 'container' }, crumbs)
+		return createElement('div', { class: ['breadcrumb', { 'breadcrumb--collapsed': (this.hiddenCrumbs.length === breadcrumbs.length - 2) }], ref: 'container' }, crumbs)
 	},
 }
 </script>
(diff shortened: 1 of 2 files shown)


src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -512,6 +512,11 @@ export default {
 	width: 100%;
 	flex-grow: 1;
 
+	&.collapsed  .crumb:last-child {
+		min-width: 100px;
+		flex-shrink: 1;
+	}
```
skjnldsv, line 518: Please use BEM syntax
```suggestion
	&.crumb--collapsed  .crumb:last-child {
		min-width: 100px;
		flex-shrink: 1;
	}
```

skjnldsv (changes requested):
See comments :)

raimund-schluessler:
> 
> 
> See comments :)

Done.

skjnldsv (approved):

ChristophWurst (approved):
ah yes, css

src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -512,6 +512,11 @@ export default {
 	width: 100%;
 	flex-grow: 1;
 
+	&--collapsed  .crumb:last-child {
+		min-width: 100px;
```
skjnldsv, line 516: Hey @raimund-schluessler , old PR, but is this really intended?
image

raimund-schluessler: Well, I put it to prevent the last crumb to shrink to zero. But in your example, it's clearly not ideal.

skjnldsv: Ah, I'll see later if css magic allows that now that we are 4 years later :p

PR #1775, opened 2021-03-22
Minimize bundled translations

jotoeri:
#### Why?
On Password-Policy the usage of _SettingsSection_ instead of a simple _div_ blew up the package-size from `100 KiB` to `330 KiB` of which `140 KB `are the `l10n.js` of vue-Components. So i thought about how to reduce this.

#### Idea
The current implementation inserts the pot-file as json, and packs this within the bundle. However, there is a huge overhead on that kind of storing. So the idea was to reduce the overhead by storing only necessary data.
As model i initially used the json-format as translations are stored within apps.

#### Approach
- Unfortunately i didn't find a npm-package/a way to store in the desired format, the recommended package (po2json) seems to be unmaintained and breaks on plural-translations.
- Therefore this approach now manually transforms/compresses the format and decompresses it before inserting into gettextBuilder.
- I must admit on the one hand this feels a bit hacky, as it compresses to a custom format on webpack-build, just to store the data and decompresses then on l10n insertion manually.
=> `Webpack-build compresses -> Component stores compressed data -> l10n.js decompresses when loading gettext`
- The meta-information out of the header currently gets lost, but as far as i see, this works without. To be sure, on could think about restoring the `plural-forms`, `language` and `content-type` headers.

#### Result
This approach **reduces the size of the bundled l10n.js from** `142 KB` **to** `32 KB`**!**
Thus, the bundled password_policy settings reduce from `346 KiB` to `237 KiB`, while e.g. the forms-app bundle reduces from `2.75 MiB` to `2.21 MiB`.

#### More Information:

The way translations are stored currently (Only 1 plural and 2 singular translations):
```
{
  "locale": "de_DE",
  "json": {
    "charset": "utf-8",
    "headers": {
      "Last-Translator": "Mario Siegmann <mario_siegmann@web.de>, 2020",
      "Language-Team": "German (Germany) (https://www.transifex.com/nextcloud/teams/64236/de_DE/)",
      "Content-Type": "text/plain; charset=UTF-8",
      "Language": "de_DE",
      "Plural-Forms": "nplurals=2; plural=(n != 1);"
    },
    "translations": {
      "": {
        "": {
          "msgid": "",
          "comments": {
            "translator": "\nTranslators:\nPhilipp Fischbeck <pfischbeck@googlemail.com>, 2020\nProfDrJones <jones@fs.cs.hm.edu>, 2020\nMark Ziegler <mark.ziegler@rakekniven.de>, 2020\nMario Siegmann <mario_siegmann@web.de>, 2020\n"
          },
          "msgstr": [
            "Last-Translator: Mario Siegmann <mario_siegmann@web.de>, 2020\nLanguage-Team: German (Germany) (https://www.transifex.com/nextcloud/teams/64236/de_DE/)\nContent-Type: text/plain; charset=UTF-8\nLanguage: de_DE\nPlural-Forms: nplurals=2; plural=(n != 1);\n"
          ]
        },
        "%n Action": {
          "msgid": "%n Action",
          "msgid_plural": "%n Actions",
          "msgstr": [
            "%n Aktion",
            "%n Aktionen"
          ]
        },
        "Actions": {
          "msgid": "Actions",
          "comments": {
            "reference": "src/components/Actions/Actions.vue:254"
          },
          "msgstr": [
            "Aktionen"
          ]
        },
        "Activities": {
          "msgid": "Activities",
          "comments": {
            "reference": "src/components/EmojiPicker/EmojiPicker.vue:176"
          },
          "msgstr": [
            "Aktivitäten"
          ]
        },
      }
    }
  }
}
```
The way i would store it here (same translation Data!):
```
{
  "locale": "de_DE",
  "translations": {
    "%n Action": {                       // Plural translation needs some specific handling
      "pluralId": "%n Actions",
      "msgstr": [
        "%n Aktion",
        "%n Aktionen"
      ]
    },
    "Actions": "Aktionen",           // Singular Translation
    "Activities": "Aktivitäten",
  }
},
```

Colorful image of the **bundles before**:
grafik

Colorful image of the **bundles after**:
grafik

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -31,9 +31,28 @@ const translations = fs
 
 		const po = fs.readFileSync(path)
 		const json = gettextParser.po.parse(po)
+
+		// Compress translations Content
+		const translations = {}
+		for (key in json.translations['']) {
+			if (key !== '') {
+				// Plural
+				if ('msgid_plural'in json.translations[''][key]) {
+					translations[json.translations[''][key].msgid] = {
+						pluralId: json.translations[''][key].msgid_plural,
+						msgstr: json.translations[''][key].msgstr
+					}
+					continue
+				}
+
+				// Singular
+				translations[json.translations[''][key].msgid] = json.translations[''][key].msgstr[0]
+			}
+		}
+
 		return {
 			locale,
-			json,
+			translations,
 		}
 	})
 
--- a/src/l10n.js
+++ b/src/l10n.js
@@ -2,7 +2,38 @@ import { getGettextBuilder } from '@nextcloud/l10n/dist/gettext'
 
 const gtBuilder = getGettextBuilder()
 	.detectLocale()
-TRANSLATIONS.map(data => gtBuilder.addTranslation(data.locale, data.json))
+
+// Decompress Translations to gettext format and add to gtBuilder
+TRANSLATIONS.forEach((lang) => {
+	const translations = {}
+
+	for (const key in lang.translations) {
+		// Plural
+		if (lang.translations[key].pluralId) {
+			translations[key] = {
+				msgid: key,
+				msgid_plural: lang.translations[key].pluralId,
+				msgstr: lang.translations[key].msgstr,
+			}
+			continue
+		}
+
+		// Singular
+		translations[key] = {
+			msgid: key,
+			msgstr: [
+				lang.translations[key],
+			],
+		}
+	}
+
+	gtBuilder.addTranslation(lang.locale, {
+		translations: {
+			'': translations,
+		},
+	})
+})
+
 const gt = gtBuilder.build()
 
 const n = gt.ngettext.bind(gt)


raimund-schluessler:
> This approach reduces the size of the bundled l10n.js from 142 KB to 32 KB!
Thus, the bundled password_policy settings reduce from 346 KiB to 237 KiB, while e.g. the forms-app bundle forms reduces from 2.75 MiB to 2.21 MiB.

Why does the forms-app bundle size reduce more than the l10n.js file size? Is the same l10n.js file packed into the bundle multiple times?

jotoeri:
> Why does the forms-app bundle size reduce more than the l10n.js file size? Is the same l10n.js file packed into the bundle multiple times?

Yep, we currently import the components as single components (instead of nc/vue as a whole). And each component contains the l10n-file. You can see this also in the two bundle-analyzer-images.

raimund-schluessler:
> 
> 
> > Why does the forms-app bundle size reduce more than the l10n.js file size? Is the same l10n.js file packed into the bundle multiple times?
> 
> Yep, we currently import the components as single components (instead of nc/vue as a whole). And each component contains the l10n-file. You can see this also in the two bundle-analyzer-images.

Hm, I thought (or maybe more like hoped) that this would be optimized by webpack.

skjnldsv:
> Hm, I thought (or maybe more like hoped) that this would be optimized by webpack.

I guess we have some improvements to do in our webpack config :thinking:

webpack.common.js
```diff
@@ -18,8 +18,6 @@ const versionHash = md5(appVersion).substr(0, 7)
 const SCOPE_VERSION = JSON.stringify(versionHash)
 const ICONFONT_NAME = `iconfont-vue-${versionHash}`
 
-console.info('This build version hash is', versionHash, '\n')
```
skjnldsv, line 21: No, please let this here

skjnldsv (changes requested):
See comment

webpack.common.js
jotoeri: That is just for the bundle-analyzer. Regarding this PR, just have a look onto the last commit https://github.com/nextcloud/nextcloud-vue/commit/d82e0952528961d3b180cad4670d85cb57c4f3fc. Everything else would be removed otherwise.

jotoeri: Just pushed a stripped version. So it's easier to look into.

skjnldsv (approved):

raimund-schluessler:
> 
> 
> > Hm, I thought (or maybe more like hoped) that this would be optimized by webpack.
> 
> I guess we have some improvements to do in our webpack config 🤔

By the way, the font files also get duplicated in the final bundle. I had a look around, but couldn't find any solution for deduplicating this. If anyone has an idea, it would be great, since duplicating l10n.js and the fonts really has a huge impact on the bundle size.

skjnldsv:
Moving to webbpack5 might be a great thing to do and have better tree-shaking.

raimund-schluessler:
> Moving to webbpack5 might be a great thing to do and have better tree-shaking.
> 
> 

Not sure. In one of my apps I have webpack 5.27.1 and the code is still present multiple times in the final bundle. But I will check if using webpack5 in the vue-components as well has any influence.

jotoeri:
As far as I see, webpack sees the components from the outside as one single block. So i think having separate chunks for l10n and fonts should enable webpack to see it as separate requirement and therefore enable to import it only once... I also tried that, but didn't manage to import it on a quick-shot. Seems to need some specific handling on the apps to import it then.

jotoeri:
⬆️ Small fixup to optimize plural storing a bit.

jotoeri:
Opened a separate issue for deduplication, so we can keep track there. #1779

jotoeri:
Somebody else up for a second review here? 😊

ChristophWurst (approved):

PR #1776, opened 2021-03-22
Fix translation extraction

jotoeri:
Two small things:
- As plural-translations have a format of `n('%n Action', '%n Actions', n)` the extraction needs to use parameters 0 and 1. Didn't get recognized, as there are currently no plural-translations on this repo.
- Currently, translations do not get extracted if they are directly written within a html-property. Normally that should be fixed on extraction, however, the author seems to be on a bigger rewrite (since two years? https://github.com/lukasgeiter/gettext-extractor/issues/31 🤔), so this is a quick and simple workaround to put them on new lines.

### DIFF
--- a/src/components/AppSidebar/AppSidebar.vue
+++ b/src/components/AppSidebar/AppSidebar.vue
@@ -131,7 +131,7 @@
 				class="app-sidebar-header">
 				<!-- close sidebar button -->
 				<a
-					v-tooltip.auto="t('close')"
+					v-tooltip.auto="closeTranslated"
 					href="#"
 					class="app-sidebar__close icon-close"
 					@click.prevent="closeSidebar" />
@@ -237,9 +237,9 @@ import Actions from '../Actions'
 import Focus from '../../directives/Focus'
 import Linkify from '../../directives/Linkify'
 import Tooltip from '../../directives/Tooltip'
-import l10n from '../../mixins/l10n'
 import AppSidebarTabs from './AppSidebarTabs'
 import EmptyContent from '../EmptyContent/EmptyContent'
+import { t } from '../../l10n'
 import { directive as ClickOutside } from 'v-click-outside'
 
 export default {
@@ -257,8 +257,6 @@ export default {
 		ClickOutside,
 		Tooltip,
 	},
-
-	mixins: [l10n],
 
 	props: {
 		active: {
@@ -365,6 +363,7 @@ export default {
 
 	data() {
 		return {
+			closeTranslated: t('Close'),
 			isStarred: this.starred,
 		}
 	},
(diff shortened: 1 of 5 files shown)


src/components/AppSidebar/AppSidebar.vue
```diff
@@ -131,7 +131,8 @@
 				class="app-sidebar-header">
 				<!-- close sidebar button -->
 				<a
-					v-tooltip.auto="t('close')"
+					v-tooltip.auto="
+						t('Close')"
```
skjnldsv, line 135: Seems weird, and most likely invalid with our styleguide :thinking:

jotoeri: Yeah, however linter does not complain... 🤔 

Alternative approach would be this, is just a bit more of code...
```
data {
  return {
    closeTranslation: t('Close')
  }
}
```

skjnldsv: Why do you need the newline anyway?

jotoeri: `npm run l10n:extract` respectively `gettext-extractor` ([build/extract-l10n.js](https://github.com/nextcloud/nextcloud-vue/blob/9526649a798f259d2c637e4623bb5f9b609b410b/build/extract-l10n.js)) did not recognise it within an html-tag. So no extraction, no translations on transifex... 😢

skjnldsv: Can't we fix the `gettext-extractor` directly? No one is going to remember putting new lines :/

jotoeri: I didn't find a way... A PR on the repo seems not worth the work, as there are other PRs that just didn't get merged (see linked issue above and [this PR](https://github.com/lukasgeiter/gettext-extractor/pull/44)).
Only other way, that i could imagine would be to use a html-parser with specific html-tags t and n to put the translations in a second time, but thats even more hacky.
Or just some completely different parser-package... 🤔

jotoeri: Any preferences? As i said - i don't see us currently fixing the extractor.

- Using the given way or the proposed data/prop way? The latter is probably a bit cleaner, but 5 LoC more. The prop could also be done as object `l10n.close = t('Close')`
- Just came to my mind that i could additionally add a workflow spitting a warning if some pattern `="t(` is found. This is how i found those three occurences.

skjnldsv: > data {
>   return {
>     closeTranslation: t('Close')
>   }
> }

Let's use this syntax instead then, will be cleaner and easier to manage

jotoeri: Ok, so it's this version now.

The webpack-dev-build i looked into for moment, but it seems to be not fully straigth forward.

skjnldsv (approved):

raimund-schluessler (approved):

PR #1783, opened 2021-03-24
Bump @nextcloud/eslint-config and related packages

raimund-schluessler:
This PR updates nextcloud/eslint-config to 4.0.0 and bumps related packages otherwise creating problems resolving the dependency tree.


raimund-schluessler:
There is a linter error (and a warning) for the PopoverMenuItem component which is not trivially fixable (mutating of a prop). Since the PopoverMenu component is anyway deprecated, I wonder if it is worth the effort to fix the component, or if we should just remove it and make it a breaking change.

skjnldsv:
> or if we should just remove it and make it a breaking change

Add an eslint-ignore or remove it yes :)

raimund-schluessler:
> 
> 
> > or if we should just remove it and make it a breaking change
> 
> Add an eslint-ignore or remove it yes :)

Then let's remove it. We anyway need a new major version after https://github.com/nextcloud/nextcloud-vue/pull/1720 is merged.

raimund-schluessler:
Turns out we cannot remove the PopoverMenu component yet, because the Avatar component still uses it 🙈 I disabled the eslint warning about the mutated prop for now. This also makes this PR a non breaking change.

I created an issue for the Avatar/PopoverMenu migration: https://github.com/nextcloud/nextcloud-vue/issues/1784.

raimund-schluessler:
Not sure, what happened to the package-lock.json.

raimund-schluessler:
> Not sure, what happened to the package-lock.json.
> 
> Edit: I run npm version 7 which uses a new lockfile format. I will check if I can pin the lockfile version to 1.

Downgraded to npm@6 and fixed the package-lock.json. The PR should be good to go now.

ChristophWurst (approved):

skjnldsv (approved):

PR #1807, opened 2021-03-30
Ensure that crumbs to hide have correct class

raimund-schluessler:
In some cases, breadcrumbs which should be hidden would not have the correct class set, causing them to still be visible. This was because the breadcrumbs were not rendered yet, when the class was applied. Hence `crumb.elm` was `undefined` and the class could not be applied. In case a rerender was triggered (e.g. by opening the Dropdown menu), the correct class was then set and the crumbs were correctly hidden.

I changed the code to now double check after the render function was called and has finished (in the `updated` hook), that the correct class is set for the crumbs to hide.

This PR also fixes a small papercut introduced by #1771 where the `min-width` would cause the last crumb to be wider than necessary if the content was narrow. In some cases this caused the crumbs to be completely hidden, although one crumb could be visible. I fixed it by now temporarily removing `min-width` when getting the crumb width in javascript.

Before:
before

After:
after

### DIFF
--- a/src/components/Breadcrumbs/Breadcrumbs.vue
+++ b/src/components/Breadcrumbs/Breadcrumbs.vue
@@ -163,15 +163,26 @@ export default {
 		/**
 		 * Check the size on update
 		 */
-		this.$nextTick(() => {
-			this.handleWindowResize()
-		})
+		this.delayedResize()
+		/**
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
:elephant:

skjnldsv (approved):

PR #1808, opened 2021-03-31
Feature/noid/button2

ma12-co:


Pytal:
@marcoambrosini does this supersede #1468? Would be very useful for https://github.com/nextcloud/server/issues/28139 :)

image

marcoambrosini:
> @marcoambrosini does this supersede #1468? 

It does!

marcoambrosini:
state of affairs:


https://user-images.githubusercontent.com/26852655/129519036-dac53233-6255-481b-89cd-5091dcb1037a.mov


https://user-images.githubusercontent.com/26852655/129519087-0f1a218b-1deb-4bc8-a3cd-4ee8c8442c40.mov





still to do:

- [ ] Add fallbacks for primary colors that are not contrasty enough;
- [ ] Fine tune colors;
- [ ] Add ripple effect;
- [ ] Accessibility check;

src/components/Button/Button.vue
```diff
@@ -0,0 +1,294 @@
...
+	background-color: var(--color-primary);
+	color: var(--color-primary-text);
+	&:hover,
+	&:focus {
+		background-color: var(--color-primary-hover);
+	}
+}
+
+// Success
+.vue-success {
```
skjnldsv, line 265: BEM

src/components/Button/Button.vue
```diff
@@ -0,0 +1,294 @@
...
+	background-color: var(--color-success);
+	color: var(--color-primary-text);
+	&:hover,
+	&:focus {
+		background-color: var(--color-success-hover);
+	}
+}
+
+// Warning
+.vue-warning {
```
skjnldsv, line 275: BEM

src/components/Button/Button.vue
```diff
@@ -0,0 +1,294 @@
...
+	background-color: var(--color-primary-light);
+	color: var(--color-primary);
+	&:hover,
+	&:focus {
+		background-color: var(--color-primary-light-hover);
+	}
+}
+
+// Primary
+.vue-primary {
```
skjnldsv, line 255: BEM

src/components/Button/Button.vue
```diff
@@ -0,0 +1,294 @@
...
+	background-color: var(--color-warning);
+	color: var(--color-primary-text);
+	&:hover,
+	&:focus {
+		background-color: var(--color-warning-hover);
+	}
+}
+
+// Error
+.vue-error {
```
skjnldsv, line 285: BEM

src/components/Button/Button.vue
```diff
@@ -0,0 +1,294 @@
...
+	// We use around our buttons instead of an outline, so that the added "border"
+	// coincides with the border of the element.
+	&.tabbed {
+		box-shadow: 0px 0px 0px 2px var(--color-main-text);
+	}
+}
+
+// Button colors
+// Default
+.vue-default{
```
skjnldsv, line 245: Should most likely be in the default css then ?

src/components/Button/Button.vue
marcoambrosini: It's in the todo list :)

src/components/Button/Button.vue
```diff
@@ -0,0 +1,294 @@
...
+		padding: 0 16px 0 4px;
+	}
+
+	// Wide button spans the whole width of the container
+	&.wide {
+		width: 100%;
+	}
+
+	// We use around our buttons instead of an outline, so that the added "border"
+	// coincides with the border of the element.
```
skjnldsv, line 237: Is the outline not ok?

raimund-schluessler:
Wouldn't it be better to have nextcloud/server#25774 in here? Otherwise we will have a dependency on server, which I thought we wanted to avoid?! @skjnldsv

marcoambrosini:
> Wouldn't it be better to have nextcloud/server#25774 in here? Otherwise we will have a dependency on server, which I thought we wanted to avoid?! @skjnldsv

The problem is that the colours of the buttons depend on the theming and dark/light modes. I don't know if there's a straightforward way to do everything here.

src/components/Button/Button.vue
marcoambrosini: sure!

src/components/Button/Button.vue
marcoambrosini: No, because it's not possible to add a border radius to it, it's always a rectangle around the element and it doesn't look good with our buttons

marcoambrosini:
- [ ] Add support for tooltips

marcoambrosini:
- [ ] Write proper docs

marcoambrosini:
- [ ] Add disabled state

src/components/Button/Button.vue
skjnldsv: To be checked about the usual process on accessibility. It might be a bad thing to round the outline. I'm lacking knowledge here :)

skjnldsv:
> Wouldn't it be better to have [nextcloud/server#25774](https://github.com/nextcloud/server/pull/25774) in here?

Not sure, It's fine for component-only variables.

skjnldsv:
> state of affairs:
> 
>  Screen.Recording.2021-08-16.at.08.07.54.mov 
>  Screen.Recording.2021-08-16.at.08.12.31.mov 

Cursors needs a fix :)

marcoambrosini:
> Cursors needs a fix :)

Squeezed that in the aria commit :)

src/components/Button/Button.vue
```diff
@@ -0,0 +1,305 @@
...
+				'button-vue--tabbed': this.tabbed,
+			}
+		},
+	},
+
+	methods: {
+		// Forward the click event to the parent component
+		handleClick(event) {
+			this.$emit('click', event)
+		},
```
skjnldsv, line 166: Why is this needed?

skjnldsv: You can also bind all listeners on the button if that is your issue :wink:

src/components/Button/Button.vue
```diff
@@ -0,0 +1,305 @@
...
+
+		// When the tab key is lifted, the button has been "tabbed in"
+		handleTabUp() {
+			this.tabbed = true
+		},
```
skjnldsv, line 171: Why not use the focus event?

src/components/Button/Button.vue
marcoambrosini: It's not is it? It could be just instance of the <Button /> component?
like:
```
<Button @click="...">
```

marcoambrosini: Any drawbacks to that? ^

src/components/Button/Button.vue
```diff
@@ -0,0 +1,305 @@
...
+	font-weight: bold;
+	height: $clickable-area;
+	min-width: $clickable-area;
+	display: flex;
+	align-items: center;
+	justify-content: center;
+	cursor: pointer;
+	border-radius: var(--border-radius-pill);
+	transition: background-color 0.1s linear !important;
+	transition: border 0.1s linear;
```
skjnldsv, line 198:
```suggestion
	transition: background-color var(--animation-quick) linear !important;
	transition: border var(--animation-quick) linear;
```

src/components/Button/Button.vue
marcoambrosini: Because we don't want the thick border if focused with the mouse

skjnldsv: Make sense. Please add a comment about that :)

src/components/Button/Button.vue
```diff
@@ -0,0 +1,305 @@
...
+
+		/**
+		 * Always provide an aria-label to your button and try to make it more
+		 * specific than the button's title. E.g. if the title of the button is
+		 * "send" in the Mail app, the label could be "Send email"
+		 */
+		ariaLabel: {
+			type: String,
+			required: true,
+		},
```
skjnldsv, line 113: An aria-label is only required when the text inside the button is NOT self-explanatory.
e.g
```html
<button>Open menu</button>
<button aria-label="Close menu">X</button>
```

marcoambrosini: Yes, I was just thinking about this... What about requiring it anyway to encourage providing a better explanation, and then remove it in the component if `===` to the title?

The reasoning behind this is that many times the visual context gives meaning to the button, and we should try to provide this context as much as possible via aria labels.

marcoambrosini: So if no aria-label is provided it it would throw a console error that would make developers stop and think about this for a sec.

marcoambrosini: Otherwise I fear that this will be always blank

src/components/Button/Button.vue
marcoambrosini: It's in the data, where the `tabbed` variable is defined, should I move it??

src/components/Button/Button.vue
skjnldsv: > What about requiring it anyway to encourage providing a better explanation

The issue is that aria-bale is not supposed to be additional data, but the data iself iirc.
If you read this with a screen reader I assume it will read the aria label, but ignoring the button content. So why even provide two then :)

I'm fine having a throw if you don't provide a content nor aria-label then :+1:

marcoambrosini:
- [ ] Cursor pointer on all children

src/components/Button/Button.vue
```diff
@@ -120,6 +123,25 @@ export default {
...
+		/**
+		 * If you want more fine tuning over the tooltip behavior, you can pass
+		 * in a full "options' object instead of a tooltip string.
+		 * check https://github.com/Akryum/v-tooltip#other-options for more
+		 * information
+		 */
+		tooltipOptions: {
+			type: Object,
+			default: undefined,
+		},
```
skjnldsv, line 138: Why not just let the dev use v-tooltip on the component itself?

src/components/Button/Button.vue
skjnldsv: I guess more comments is always good ;)
I'm usually fond of comments on the implementations rather than the variable declaration, makes it easier to narrow :)

src/components/Button/Button.vue
skjnldsv: Well, theoritecally you should be able to listen to the click event without having the click listener within the component.
You can use `@click.native` if needed for example. But I'm wondering why `@click` would not just work :thinking:

src/components/Button/Button.vue
marcoambrosini: You've got a point sir, shall I remove this?

src/components/Button/Button.vue
marcoambrosini: Ok I'll try that but I remember that it I never accomplished to make a prop depend from another. In this case the required boolean would have to check another prop, the title. 
Otherwise I can perform the check in the mounted hook and throw there

src/components/Button/Button.vue
marcoambrosini: I'll test this but I think it will work. So I'll likely remove this too!

src/components/Button/Button.vue
skjnldsv: I think so yes :)

src/components/Button/Button.vue
skjnldsv: You can throw in the beforeMount if the check is not successful. But best is just letting the dev do maybe? Always a tough call between enforcing and documenting :)

marcoambrosini: I think that what you proposed, either title or aria label, is a good policy here :)

marcoambrosini:
- [ ] Enforce use of material design icon and maybe also the size

marcoambrosini:
Logic for fallback button colors: 
https://github.com/nextcloud/server/blob/da68952524116328a17e06c039ca98fc4d8c76af/apps/theming/css/theming.scss#L262-L302

marcoambrosini:
- [ ] Primary button has 3 levels (no matter if light or dark theme):
	- If the color is dark enough (e.g. Nextcloud blue) it will be that color, with white text on top
	- If it is a lighter color like yellow or orange, the text needs to turn black, just like the icons do for the header
	- If it is very very light like e.g. white, the background color needs to fall back to a light grey
- [ ] Secondary: to be defined
- [ ] The color-error and color-success buttons will _always_ have white text, no matter the theme nor primary color, as they are always on the same background.
- [ ] More padding to the right and left sides of text only buttons;

skjnldsv:
Seeing the amount of comments and changes, maybe it would be worth opening a nice fresh PR when ready :wink: 
Great work Marco! :rocket:

marcoambrosini:
@skjnldsv most of the comments are additions or changes to the server's color variables pr

src/components/Button/Button.vue
```diff
@@ -0,0 +1,354 @@
...
+
+		/**
+		 * Specifies the button color
+		 * Accepted values: primary, error, warning, success. If left empty,
+		 * the default button style will be used.
+		 */
+		color: {
+			type: String,
+			validator(value) {
+				return ['primary', 'error', 'warning', 'success', ''].indexOf(value) !== -1
```
Pytal, line 97:
```suggestion
				return ['primary', 'error', 'warning', 'success', ''].includes(value)
```

src/components/Button/Button.vue
marcoambrosini: Checked and it's acceptable and widely adopted :)

raimund-schluessler:
> 
> 
>    * [ ]  Enforce use of material design icon and maybe also the size

@marcoambrosini I think enforcing the size of the material design icon (i.e. a prop of a component in a slot) will be a real pain (at least as far as I know, maybe @skjnldsv knows more there). But do you think this is really necessary? We use material design icons in slots (e.g. for the `Actions` components) without enforcing anything there. 🤷

marcoambrosini:
@raimund-schluessler why do you think it would it be a pain? 
Material design icons default at 24px and we use 16px. If we don't do this I think it's much more likely that we'll have a lot of inconsistencies in how buttons look across Nextcloud.
Do you have a use-case in mind in which we would need a differently sized icon?

raimund-schluessler:
> 
> 
> @raimund-schluessler why do you think it would it be a pain?

Because I think there is no straight forward way of accessing the props of a child component which was passed into a slot. If I am not wrong, you would need to check the respective slot, check if the slot content really is a material-design-icon and set the prop, maybe similar to how it's done in the `Breadcrumbs` component:
https://github.com/nextcloud/nextcloud-vue/blob/dc15bff68a75854949fac62679887b8d4350b93c/src/components/Breadcrumbs/Breadcrumbs.vue#L153
https://github.com/nextcloud/nextcloud-vue/blob/dc15bff68a75854949fac62679887b8d4350b93c/src/components/Breadcrumbs/Breadcrumbs.vue#L435-L443

But that's quite some effort for just checking that the dev gave the correct icon size.

> Material design icons default at 24px and we use 16px. If we don't do this I think it's much more likely that we'll have a lot of inconsistencies in how buttons look across Nextcloud.

I think we (implicitly) agreed on using 24 px as the icon size for material design icons. We use this size throughout the vue-components and also in apps, e.g. here (@skjnldsv please correct me if I am wrong):
https://github.com/nextcloud/nextcloud-vue/blob/dc15bff68a75854949fac62679887b8d4350b93c/src/components/AppNavigationItem/InputConfirmCancel.vue#L45

> Do you have a use-case in mind in which we would need a differently sized icon?

No, I think a standard size is fine. I am just afraid that enforcing a common size is a lot of effort for relatively low benefit (especially since we don't enforce it anywhere else).

marcoambrosini:
> maybe similar to how it's done in the Breadcrumbs component

Thanks for the input, I'll check this :) 

> I think we (implicitly) agreed on using 24 px

Nope, there was some talk in the design call about increasing the size, but it was decided that for now we're staying at 16px.

skjnldsv:
> Nope, there was some talk in the design call about increasing the size, but it was decided that for now we're staying at 16px.

Careful, material design 16 is not 16px. There is a common margin, so we should use 20. In any case, I'm also against enforcing it, it's too much of a stretch. This is a button component, it needs to stay simple.

raimund-schluessler:
> 
> 
> > Nope, there was some talk in the design call about increasing the size, but it was decided that for now we're staying at 16px.
> 
> Careful, material design 16 is not 16px. There is a common margin, so we should use 20. In any case, I'm also against enforcing it, it's too much of a stretch. This is a button component, it needs to stay simple.

Would be great to document this somewhere (or is it already? 🙈). Since we use `:size="24"` in a lot of places already, this probably should be known and changed accordingly.

skjnldsv:
> Since we use `:size="24"` in a lot of places already, this probably should be known and changed accordingly.

:shrug: 24px seems to be the one that gives around 16px icons :)

marcoambrosini:
> Careful, material design 16 is not 16px

The svg contained in it is 16px by 16px, just like the previous icons we were using, isn't it? Anyway I'm ok with increasing the size.

Screenshot 2021-08-18 at 10 29 31

> I'm also against enforcing it, it's too much of a stretch. This is a button component, it needs to stay simple.

I don't agree here: we're seeing just in this conversation (3 nextcloud developers) that we're off about our idea of how big nextcloud icons are by 60% (16 to 24 px). That's a bit scary, and if we don't enforce a size, it will result in many inconsistencies, which defeats the purpose of making a component like this a bit.

As per the complexity, I think we're still far below what other UI blocks libraries do, e.g. https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/components/VBtn/VBtn.ts

skjnldsv:
> The svg contained in it is 16px by 16px, just like the previous icons we were using, isn't it? Anyway I'm ok with increasing the size.

The screenshot is a css icon, not a MaterialIcon :)
image
Capture d’écran_2021-08-18_10-49-06

marcoambrosini:
So this is a bit off topic since we'r shouldn't be discussing sizes here, but wether or not to enforce them. But since Jan and Nimisha will show up too, these are the differences between 16 and 24. 
I like the 24 better, but it feels a tad too big. I think a sweet spot might be 20, as @skjnldsv said before.

Screenshot 2021-08-18 at 10 44 59

Screenshot 2021-08-18 at 10 44 24

skjnldsv:
> I don't agree here: we're seeing just in this conversation (3 nextcloud developers) that we're off about our idea of how big nextcloud icons are by 60% (16 to 24 px). That's a bit scary, and if we don't enforce a size, it will result in many inconsistencies, which defeats the purpose of making a component like this a bit.

Yes, but you will increase the complexity of this component by 400% just for the sake of making sure devs uses 24 px. Maintainability is much more important here. Devs can do whatever they want, they can also not use this component. There is always a fine line between enforcing and strongly documenting requirements. :wink:

marcoambrosini:
If not overwriting props, I would at least throw errors. 

We need to make sure that devs: 
- Only slot in MDI icons into the button (currently they can slot in a flying dumbo gif);
- The tooltip of the icon is removed (MDI icons come with a tooltip with the icon name that shows up on hover);
- The size of the icon is the one we want;

I can perform these checks in a lifecycle hook just like with the aria label. And  with this the complexity stays low, each of these is 2 or 3 lines of code. 

Sounds good?

marcoambrosini:
It seems like throwing errors in a lifecycle hook stops the code execution, so the next check is not performed. I guess I'll have to go with just warnings 😓

skjnldsv:
> It seems like throwing errors in a lifecycle hook stops the code execution, so the next check is not performed. I guess I'll have to go with just warnings

Careful about performances for such checks as well

marcoambrosini:
- [ ] Check text vertical alignment (seems 2 px off towards bottom)

jancborchardt:
20px for icon size definitely looks best! 16 is too small, 24 is way too big.

>     * [ ]  Check text vertical alignment (seems 2 px off towards bottom)

(Yup, or 1px)

marcoambrosini:
Upon further discussion we'll reintroduce the no-background button, as it's already used in many other places.
Below are the _regular_, _hover_,  and _keyboard-focused_ states.

(Those colors are not legally binding) 😅 

Screenshot 2021-08-18 at 13 56 25

marcoambrosini:
Other than the fine tuning of the color variables [which is happening here](nextcloud/server#25774),. this is basically reviewable!

src/components/Button/Button.vue
```diff
@@ -0,0 +1,404 @@
...
+				'button-vue--text-only': this.textOnly,
+				'button-vue--icon-and-text': this.iconAndText,
+				[`button-vue--vue-${this.type}`]: this.type,
+				'button-vue--wide': this.wide,
+				'button-vue--tabbed': this.tabbed,
+			}
+		},
+	},
+
+	mounted() {
```
raimund-schluessler, line 168: These checks will only run once (when the component is mounted). If I programmatically change the content, no checks will happen.

src/components/Button/Button.vue
```diff
@@ -0,0 +1,404 @@
...
+		/**
+		 * Perform checks on the on what's passed into the icon slot.
+		 */
+		// Check if the slot is populated first
+		if (Object.keys(this.$slots).length !== 0) {
+			const slottedElement = this.$slots.default[0]
+			if (slottedElement) {
+			// Check if it's a Material Design Icon
+				if (slottedElement.data.staticClass.includes('material-design-icon')) {
+					console.warn("DESIGN WARNING: You need to use a material design icon for this button's icon. More info here: https://www.npmjs.com/package/vue-material-design-icons")
```
raimund-schluessler, line 185: Shouldn't this be called when it is not a material-design-icon?

src/components/Button/Button.vue
```diff
@@ -0,0 +1,404 @@
...
+ - You should have received a copy of the GNU Affero General Public License
+ - along with this program. If not, see <http://www.gnu.org/licenses/>.
+ -
+ -->
+
+<docs>
+
+### General description
+General purpose button component. See props for different options.
+[Use material design icons only for icons](https://www.npmjs.com/package/vue-material-design-icons) and remember to set their size to 16.
```
Pytal, line 27:
```suggestion
[Use material design icons only for icons](https://www.npmjs.com/package/vue-material-design-icons) and remember to set their size to 20.
```
As discussed https://github.com/nextcloud/nextcloud-vue/pull/1808#issuecomment-900998940 :)

src/components/Button/Button.vue
```diff
@@ -0,0 +1,404 @@
...
+	aria-label="Example aria label" />
+```
+
+</docs>
+
+<template>
+	<button
+		v-ripple
+		class="button-vue"
+		v-bind="$attrs"
```
Pytal, line 43:
```suggestion
		v-bind="$attrs"
		v-on="$listeners"
```
So we can use `@click` without `.native` :)

src/components/Button/Button.vue
```diff
@@ -0,0 +1,424 @@
...
+			 * tabbed to it. We use this to display a thick 'focus outline' only
+			 * when the user is navigating with the keyboard.
+			 */
+			tabbed: false,
+		}
+	},
+
+	computed: {
+		hasText() {
+			return this.text
```
skjnldsv, line 136: Not a boolean, confusing variable name :)

src/components/Button/Button.vue
```diff
@@ -0,0 +1,424 @@
...
+						console.warn('DESIGN WARNING: You need to set the size of the material design icon to 20px.')
+					}
+
+					// Check that the title is removed so that the tooltip on the icon is removed too.
+					if (slottedElement.data.ariaLabel !== '') {
+						console.warn('DESIGN WARNING: You need to add `title=""` as a prop on the material design icon that you passed into the default slot of the Button component.')
+					}
+				}
+			}
+		},
```
skjnldsv, line 249: I'm strongly against that. That sis too much.
If I want to use an inline image and not a MaterialIcon, so be it.
What if we suddenly want a bigger button for a specific area, etc etc.

The component needs to be simple, let's not over-charge this with unnecessary checks.

src/components/Button/Button.vue
```diff
@@ -0,0 +1,424 @@
...
+		this.checkIconSlot()
+		/**
+		 * Register observer that checks the icon slot if programmatically changed.
+		 */
+		const observer = new MutationObserver(this.checkIconSlot)
+		observer.observe(this.$el, {
+			childList: true,
+			subtree: true,
+		})
+		this.observer = observer
```
skjnldsv, line 188: How much is this being called per mount/update? I would kind of like a performance analysis on this. We will quickly have a lot of those buttons. If each of them adds even 50ms of runtime, this will adds up pretty quick.

(Irrelevant if we drop those checks as I suggest we do)

src/components/Button/Button.vue
skjnldsv: Btw, if you want to enforce the icon size, you can most likely do that with css

marcoambrosini: Ok, let's remove all this then!

src/components/Button/Button.vue
```diff
@@ -0,0 +1,381 @@
...
+	text="Example text"
+	color="primary"
+	aria-label="Example aria label" />
+```
+
+</docs>
+
+<template>
+	<button
+		v-ripple
```
raimund-schluessler, line 41: If this directive mandatory? I find it annoying to be honest.

src/components/Button/Button.vue
```diff
@@ -0,0 +1,381 @@
...
+		class="button-vue"
+		v-bind="$attrs"
+		:class="buttonClassObject"
+		:aria-label="ariaLabel"
+		:disabled="disabled"
+		v-on="$listeners"
+		@keydown.enter="makeActive"
+		@keyup.enter="makeInactive"
+		@click="tabbed=false"
```
raimund-schluessler, line 50: Can we have a separate function for this, please?

src/components/Button/Button.vue
```diff
@@ -0,0 +1,381 @@
...
+		@blur="handleBlur"
+		@keyup.tab.exact="handleTabUp"
+		@keyup.shift.tab="handleTabUp">
+		<span class="button-vue__wrapper">
+			<span v-if="hasIcon" class="button-vue__icon">
+				<slot />
```
raimund-schluessler, line 56: A description of this slot would be great.
```suggestion
				<!-- @slot Manually provide icon -->
				<slot />
```

src/components/Button/Button.vue
```diff
@@ -0,0 +1,381 @@
...
+		:disabled="disabled"
+		v-on="$listeners"
+		@keydown.enter="makeActive"
+		@keyup.enter="makeInactive"
+		@click="tabbed=false"
+		@blur="handleBlur"
+		@keyup.tab.exact="handleTabUp"
+		@keyup.shift.tab="handleTabUp">
+		<span class="button-vue__wrapper">
+			<span v-if="hasIcon" class="button-vue__icon">
```
raimund-schluessler, line 55: Is this wrapper necessary? Couldn't we just use the bare slot like in the ActionsXXX component and style it accordingly (we anyway only want material design icons in this slot):
https://github.com/nextcloud/nextcloud-vue/blob/16cf329d4c8d8b4ad99fcc4b811c624f121ce0e4/src/assets/action.scss#L92-L100

raimund-schluessler (changes requested):
The button styling in the docs looks broken:
docs

src/components/Button/Button.vue
marcoambrosini: I think it's safer this way cause at least if a non mdi icon is provided it won't break.

src/components/Button/Button.vue
marcoambrosini: Yes, for now all the buttons will have it, this has been approved by the design team. I think it provides a nice click visual feedback. What is it that annoys you about it? :)

raimund-schluessler: Well, maybe "annoying" is the wrong wording. To me it looks to playful and not professional and kind of distracts me with to much movement (especially that the background color expansion starts at the click position). That's the subjective part of my critics 😉 

As for the objective part: I haven't seen this feedback anywhere yet, so the visual metaphor of providing feedback is quite uncommon, I think. I furthermore suspect that it will work badly when using the keyboard only, because there is no click event when pressing enter. So, I guess the ripple-directive will do nothing and you end up with different feedback for the same action.

It would be nice if the netlify docs would work properly though, this would help test it.

raimund-schluessler: Addon: The last release of this directive as well as the last commit to its repo is four years old (2017) and we would get a dependency to vue@2.5.8 (and no vue@3 support for sure). I would consider this unmaintained, and look for an alternative. @skjnldsv @nextcloud/vuejs

raimund-schluessler: So maybe at least use this directive: https://github.com/justintaddei/v-wave (last release two months old, supports Vue3, seems to do the same as `v-ripple`)

marcoambrosini: Oh that could be a blocker yes. Thanks for the suggestion, I will look into this directive.
As per the keyboard navigation, I added a fallback behavior, there's a feedback but it's not as cool, it's just a "pressed" state with the change of colors.

I will try to update the docs soon. Do you know if there's any way to include MD icons in the docs??

raimund-schluessler: > I will try to update the docs soon. Do you know if there's any way to include MD icons in the docs??

Yes, it's quite straight-forward. We have it for the Actions component for example:
https://github.com/nextcloud/nextcloud-vue/blob/593b3bb411f4c9be8acf9eab8249f5ec351034ef/src/components/Actions/Actions.vue#L84-L123

marcoambrosini: Thanks @raimund-schluessler :)

package.json
```diff
@@ -63,7 +63,6 @@
 		"vue-color": "^2.7.1",
 		"vue-material-design-icons": "^4.11.0",
 		"vue-multiselect": "^2.1.6",
-		"vue-ripple-directive": "^2.0.1",
```
raimund-schluessler: So, no ripple directive at all anymore?

src/directives/index.js
```diff
@@ -23,3 +23,4 @@
 export { default as Focus } from './Focus'
 export { default as Linkify } from './Linkify'
 export { default as Tooltip } from './Tooltip'
+export { default as Ripple } from './Ripple'
```
raimund-schluessler, line 26: Needs to be removed if we really don't use the ripple directive anymore.

raimund-schluessler:
Would be nice to have this component in. I could use it well 😉 

@marcoambrosini For fixing the button styling in the docs, I guess you have to add the necessary variables from https://github.com/nextcloud/server/pull/25774 here:
https://github.com/nextcloud/nextcloud-vue/blob/master/styleguide/assets/variables.css

package.json
marcoambrosini: No, as it blocks upgrade to vue3

marcoambrosini:
Fixed docs, rebased and squashed everything 🎉 

Screenshot 2021-10-06 at 11 47 09

raimund-schluessler:
@marcoambrosini Seems to work really nice in general (besides the unexpected tab error). 👍 

However (and I really hope I am not to enervating 🙈), but I would propose to use the default slot instead of a `text` prop for the text. The `Button` component would be more in line with e.g. the `ActionButton` component then. I mean to have a markup like this:

```vue
<Button type="no-background">
	<template #icon>
		<Video
		:size="20" />
	</template>
	Example text
</Button>
```
instead of currently
```vue
<Button
	text="Example text"
	type="no-background">
	<Video
		:size="20" />
</Button>
```

Compare the `ActionButton` example in the docs:
```vue
<ActionButton>
	<template #icon>
		<HandBackLeft
			:size="20"
			decorative
			title="" />
	</template>
	Raise left hand
</ActionButton>
```

Sorry, that I didn't notice this earlier.

raimund-schluessler:
And cypress complains that the CSS changes have introduced an issue with the arrow button of the input fields in the AppSidebar (the border does not match the rest of the input field anymore):

AppSidebar vue-subtitle_null-starred_false-compact_false-header_image-secondary_button-editable_true-actual

marcoambrosini:
> use the default slot instead of a text prop

Ok will do

> And cypress complains that the CSS changes have introduced an issue with the arrow button

Isn't this only in the styleguide though?

raimund-schluessler:
> > use the default slot instead of a text prop
> 
> Ok will do

Great, thanks a lot. And sorry again for the extra work.

> > And cypress complains that the CSS changes have introduced an issue with the arrow button
> 
> Isn't this only in the styleguide though?

That's a tricky question. For the visual regression test we load the assets of the styleguide here:
https://github.com/nextcloud/nextcloud-vue/blob/2ce7608d76a8e19a037f7b626690571bd5644179/tests/visual/components/AppSidebar/AppSidebarMixin.js#L31-L33

We rely on the styleguide assets matching the CSS files of server to properly catch unintended changes (that's not really ideal to be honest, but was the easiest solution). If the visual tests report an issue, then either the styleguide assets are different from server and it would look fine in real world usage, or there really is an issue with the CSS. In both cases it needs fixing.

marcoambrosini:
> use the default slot instead of a text prop

I have found a problem with this, which is that slots are not reactive and depending on whether the text is present or not, different styles are applied to the button.

raimund-schluessler:
> 
> 
> > use the default slot instead of a text prop
> 
> I have found a problem with this, which is that slots are not reactive and depending on whether the text is present or not, different styles are applied to the button.

Hm, I have to admit I didn't think about that, just assumed that it works in the `ActionButton`. Would it be possible to make it reactive like this:
https://github.com/nextcloud/nextcloud-vue/blob/2ce7608d76a8e19a037f7b626690571bd5644179/src/mixins/actionGlobal.js#L36-L58

raimund-schluessler:
> 
> 
> > > use the default slot instead of a text prop
> > 
> > 
> > I have found a problem with this, which is that slots are not reactive and depending on whether the text is present or not, different styles are applied to the button.
> 
> Hm, I have to admit I didn't think about that, just assumed that it works in the `ActionButton`. Would it be possible to make it reactive like this:
> 
> https://github.com/nextcloud/nextcloud-vue/blob/2ce7608d76a8e19a037f7b626690571bd5644179/src/mixins/actionGlobal.js#L36-L58

raimund-schluessler:
Sorry, wrong button 🙈

raimund-schluessler:
@marcoambrosini I played around a bit with having the text using the default slot and I think it works just fine.

Since I didn't dare to push to your PR, I created another branch https://github.com/nextcloud/nextcloud-vue/pull/2305. Feel free to pull the commits if you are ok with them.

Also the reactivity seems to be fine (you can click on the primary Icon and text button to toggle between only icon and icon & text).

marcoambrosini:
sounds good, added your commit. Let's get this bad boy in!!

raimund-schluessler:
I tested it with the Tasks app (see https://github.com/nextcloud/tasks/pull/1820) and it seems to work well. The AppSidebar still looks fine and nothing broke as far as I can tell.

@marcoambrosini I took the freedom to rebase the branch and remove the `v-wave` dependency from `package.json` as it wasn't used at the moment. We should add it in a follow-up when adding the ripple effect, I think. I hope that is ok with you. If not, feel free to undo my changes, please.

Also, I noticed that naming the component `Button` prevents using the native HTML tag `button`. Honestly, I don't know if that's a good thing (it forces devs to use the Button component) or a problem (it prevents devs to use the native button tag 🙈). I just thought I should mention it, so everyone is aware.

One last thing for the designers is to confirm that the `tertiary` button should have a transparent background. It doesn't really look like a button to me to be honest, just like icon and text:
Screenshot 2021-10-07 at 21-58-41 Tasks - Nextcloud
But that's not up to me to decide so 🚀 from my side.

nimishavijay:
I think the tertiary button doesn't need a background. There are many places where a button is required for some action which isn't so important but a primary or secondary is still too eye catching. For those cases a tertiary button which is not styled very fancily will work :)

marcoambrosini:
> Yes, that's what I meant. I got a warning when I had both Button and a button in the same component. But I didn't further check on this problem to be honest.

Let's call it `nc-button`? cc @skjnldsv  @juliushaertl

> Great, it's fine for me. I was just wondering, sorry.

No worries, that's on me :)

marcoambrosini:
> Is there maybe the possibility to make the Button component self-contained (i.e. not dependent on a specific server version), by including all CSS properties and variables in nc-vue? That way, it could be used with any server version.

It heavily relies on the theming engine in the server. This is something we will encounter every time we'll have to add new Colors, I don't think that bringing everything to the component's scoped styles is a good solution.

PR #1809, opened 2021-03-31
Make sure to pass all props and attributes to single-action Actions

skjnldsv:
Only a very narrow set of attributes were passed to the element if only one action were displayed.

Example (the download is NOT applied to the action without this PR):

```vue
<ActionLink
	download="123456"
	icon="icon-download"
	:close-after-click="true"
	:href="currentFile.davPath">
	{{ t('viewer', 'Download') }}
</ActionLink>
```


raimund-schluessler (approved):

jotoeri (approved):
Didn't test, but looks good. 👍

PR #1818, opened 2021-04-02
3.9.0

skjnldsv:


raimund-schluessler:
I like the idea. We just need to be careful with breaking changes then, I guess.

skjnldsv:
> We just need to be careful with breaking changes then, I guess.

That's also why I'm finding the migration to release-by-pr a bit better. As we have more and more apps using this lib, I feel it's easier to catch potential mistakes in the releases.

I'm sure @ChristophWurst will be happy to catch a patch that should be a major :tada:

skjnldsv:
Sure, this will be a bit slower than release-on-master-push, but then we have a bit more insight on what's going on :thinking:

juliushaertl (approved):

ChristophWurst (approved):
:shrug:

raimund-schluessler:
> And we need to make sure that the release PR is always fully rebased to master, otherwise the changelog might miss some entries, no?

Haha, now already with the first release using the changelog-generator it happened what I suspected and the release notes are incomplete, because we didn't rebase the release PR to latest master before merging/releasing. 🙈 

E.g. https://github.com/nextcloud/nextcloud-vue/pull/1822 was merged after creating the release PR and is now missing from the changelog. Not a big deal for this particular PR, but there might be others missing.

PR #1833, opened 2021-04-09
Add id prop to ListItemIcon to prevent v-binding to the Avatar

skjnldsv:
Because we `v-bind="$attrs"` to the avatar component, declaring id pass it to the `$props` and make sure we are not having a duplicate ID on the page
image

### DIFF
--- a/src/components/ListItemIcon/ListItemIcon.vue
+++ b/src/components/ListItemIcon/ListItemIcon.vue
@@ -57,7 +57,7 @@ It might be used for list rendering or within the multiselect for example
 </docs>
 
 <template>
-	<span class="option" :style="cssVars">
+	<span :id="id" class="option" :style="cssVars">
 		<Avatar
 			v-bind="$attrs"
 			:disable-menu="true"
@@ -173,6 +173,14 @@ export default {
 		isNoUser: {
 			type: Boolean,
 			default: false,
+		},
+
+		/**
+		 * Unique list item ID
+		 */
+		id: {
+			type: String,
+			default: null,
 		},
 	},
 
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -125,6 +125,7 @@ function setUserHasAvatar(userId, flag) {
 
 export default {
 	name: 'Avatar',
+
 	directives: {
 		tooltip: Tooltip,
 		ClickOutside,


src/components/ListItemIcon/ListItemIcon.vue
```diff
@@ -57,7 +57,7 @@ It might be used for list rendering or within the multiselect for example
 </docs>
 
 <template>
-	<span class="option" :style="cssVars">
+	<span :id="id" class="option" :style="cssVars">
```
raimund-schluessler, line 60:
```suggestion
	<span id="" class="option" :style="cssVars">
```
Wouldn't this be enough then? Or do we need the possibility to bind an ID there?

skjnldsv: I like the idea of keeping the id possible. Especially on lists where you can scroll to the element with the ID hash

raimund-schluessler: Fine with me.

raimund-schluessler (approved):

raimund-schluessler:
On a side note: I find the naming of this component confusing. It should rather be `ListItemAvatar`, I believe. 😉🙈

src/components/ListItemIcon/ListItemIcon.vue
```diff
@@ -174,6 +174,14 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+
+		/**
+		 * Unique list item ID
+		 */
+		id: {
+			type: String,
+			default: '',
```
jotoeri, line 183: In case there is no id given to the component, does that work then?
Afaik on html, the id-prop is not allowed to be empty if present, no? So while it would probably work on most browsers, it won't produce valid html...? 🤔 Or does vue just throw it away, if it's an empty string?

raimund-schluessler: Good catch. There are now a lot of `id=""` in the HTML code.

skjnldsv: ah right, should be null!

skjnldsv: Done!

skjnldsv:
> On a side note: I find the naming of this component confusing. It should rather be `ListItemAvatar`, I believe.

Yeah, we had the same discussion with the incoming listItem componnt lol.
It have an icon or a menu at the end, that was my original naming intent, but the avatar is also relevant.
Feel free to open a pr for 4.0.0 :wink:

raimund-schluessler (approved):

ChristophWurst (approved):
I'll allow it

jotoeri (approved):

PR #1834, opened 2021-04-10
Fix Actions css

jotoeri:
Once again, just two small css-fixes:
- Longtext-actions had a fixed width, which was a bit unhandy in case the text length was just close above the switching-length
- Longtext-actions were a bit misaligned in height.

Left **before**, right **after:**
grafikgrafik

grafikgrafik

### DIFF
--- a/src/assets/action.scss
+++ b/src/assets/action.scss
@@ -100,13 +100,14 @@
 
 		// long text area
 		p {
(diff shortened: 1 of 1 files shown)


skjnldsv (approved):
image

raimund-schluessler (approved):

PR #1843, opened 2021-04-13
Fix accessibility of app navigation toggle

jancborchardt:
hover/focus feedback before:
menu button focus feedback before

hover/focus feedback after:
menu button after

### DIFF
--- a/src/components/AppNavigationToggle/AppNavigationToggle.vue
+++ b/src/components/AppNavigationToggle/AppNavigationToggle.vue
@@ -25,15 +25,24 @@
 	<a class="app-navigation-toggle"
 		href="#"
 		:aria-expanded="open ? 'true' : 'false'"
+		:aria-label="ariaLabel"
 		aria-controls="app-navigation-vue"
 		@click.prevent="toggleNavigation"
 		@keydown.space.exact.prevent="toggleNavigation" />
 </template>
 
 <script>
+import l10n from '../../mixins/l10n'
+
 export default {
 
 	name: 'AppNavigationToggle',
+
+	computed: {
+		ariaLabel() {
+			return this.open ? t('Close navigation') : t('Open navigation')
+		},
+	},
 
 	props: {
 		open: {
@@ -47,6 +56,8 @@ export default {
 			this.$emit('update:open', !this.open)
 		},
 	},
+
+	mixins: [l10n],
 }
 </script>
 
--- a/src/components/AppNavigationToggle/AppNavigationToggle.vue
+++ b/src/components/AppNavigationToggle/AppNavigationToggle.vue
@@ -38,11 +38,7 @@ export default {
 
 	name: 'AppNavigationToggle',
 
-	computed: {
-		ariaLabel() {
-			return this.open ? t('Close navigation') : t('Open navigation')
-		},
-	},
+	mixins: [l10n],
 
 	props: {
 		open: {
@@ -51,13 +47,17 @@ export default {
 		},
 	},
 
+	computed: {
+		ariaLabel() {
+			return this.open ? this.t('Close navigation') : this.t('Open navigation')
+		},
+	},
+
 	methods: {
 		toggleNavigation() {
 			this.$emit('update:open', !this.open)
 		},
 	},
-
-	mixins: [l10n],
 }
 </script>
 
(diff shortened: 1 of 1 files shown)


jancborchardt:
Also added the missing aria-label to the AppNavigationToggle, but couldn’t figure out how to make it translatable. (Tried via props and also via `:aria-label="open ? t('Close navigation') : t('Open navigation')"` but both just resulted in empty aria-label.)

raimund-schluessler:
> Also added the missing aria-label to the AppNavigationToggle, but couldn’t figure out how to make it translatable. (Tried via props and also via `:aria-label="open ? t('Close navigation') : t('Open navigation')"` but both just resulted in empty aria-label.)

@jancborchardt You need to import the `l10n` mixin like so (maybe adjust the path):
https://github.com/nextcloud/nextcloud-vue/blob/b1d53388b6aaae47a5609f2dc5fe476aca821c1f/src/components/SettingsSelectGroup/SettingsSelectGroup.vue#L48

and use it in the component like so:
https://github.com/nextcloud/nextcloud-vue/blob/b1d53388b6aaae47a5609f2dc5fe476aca821c1f/src/components/SettingsSelectGroup/SettingsSelectGroup.vue#L54

Then you can use `t('string')` in the template:
https://github.com/nextcloud/nextcloud-vue/blob/b1d53388b6aaae47a5609f2dc5fe476aca821c1f/src/components/SettingsSelectGroup/SettingsSelectGroup.vue#L38

jancborchardt:
Thanks a lot @raimund-schluessler (and @marcelklehr via chat too), works now and ready to review! :)

src/components/AppNavigationToggle/AppNavigationToggle.vue
```diff
@@ -25,12 +25,15 @@
 	<a class="app-navigation-toggle"
 		href="#"
 		:aria-expanded="open ? 'true' : 'false'"
+		:aria-label="open ? t('Close navigation') : t('Open navigation')"
```
raimund-schluessler, line 28: I just remembered that there seems to be an issue extracting the translation when it's within a html-property: https://github.com/nextcloud/nextcloud-vue/pull/1776

So it might be better to make it a computed property instead.

raimund-schluessler:
```suggestion
		:aria-label="ariaLabel"
```

src/components/AppNavigationToggle/AppNavigationToggle.vue
```diff
@@ -25,12 +25,15 @@
...
 		@click.prevent="toggleNavigation"
 		@keydown.space.exact.prevent="toggleNavigation" />
 </template>
 
 <script>
+import l10n from '../../mixins/l10n'
+
 export default {
 
 	name: 'AppNavigationToggle',
```
raimund-schluessler, line 39:
```suggestion
	name: 'AppNavigationToggle',
	computed: {
		ariaLabel() {
			return this.open ? t('Close navigation') : t('Open navigation')
		},
	},
```

src/components/AppNavigationToggle/AppNavigationToggle.vue
jancborchardt: Thanks @raimund-schluessler, added both changes! :)

raimund-schluessler:
@jancborchardt I took the liberty to fix the warning and import error on this branch. It works as it should now.

I am just not so sure about the background-color, as it differs from what we use on other Actions. And thinking about that, I figured that it would actually make sense to simply use an ActionButton for the AppNavigationToggle. So I created another PR which solves the same issue by using an ActionButton, see https://github.com/nextcloud/nextcloud-vue/pull/1844.

src/components/AppNavigationToggle/AppNavigationToggle.vue
```diff
@@ -72,6 +83,8 @@ export default {
 	&:hover,
 	&:focus {
 		opacity: $opacity_full;
+		background-color: var(--color-background-hover);
+		border-radius: var(--border-radius-pill);
```
skjnldsv, line 87: I would argue that the radius should most likely be half the size of the element, and that this value have nothing to do here (as it's 100px and referring to another usage), but it works :grin:

raimund-schluessler: @skjnldsv @jancborchardt How about #1844?

skjnldsv: yes

raimund-schluessler: I will close this PR in favor of #1844 then.

raimund-schluessler:
Closing in favor of #1844.

PR #1844, opened 2021-04-13
Use ActionButton for AppNavigationToggle

raimund-schluessler:
This is another approach to https://github.com/nextcloud/nextcloud-vue/pull/1843. We simply use `ActionButton` for the `AppNavigationToggle` now and let it do the heavy lifting.

hover/focus feedback after:
Unbenannt
Unbenannt

### DIFF
--- a/src/components/AppNavigationToggle/AppNavigationToggle.vue
+++ b/src/components/AppNavigationToggle/AppNavigationToggle.vue
@@ -25,7 +25,6 @@
 	<Actions class="app-navigation-toggle">
 		<ActionButton
 			:aria-expanded="open ? 'true' : 'false'"
-			:aria-label="label"
 			aria-controls="app-navigation-vue"
 			@click="toggleNavigation">
 			<Menu
--- a/src/components/AppNavigationToggle/AppNavigationToggle.vue
+++ b/src/components/AppNavigationToggle/AppNavigationToggle.vue
@@ -22,25 +22,39 @@
  -
  -->
 <template>
-	<a class="app-navigation-toggle"
-		href="#"
-		:aria-expanded="open ? 'true' : 'false'"
-		aria-controls="app-navigation-vue"
-		@click.prevent="toggleNavigation"
-		@keydown.space.exact.prevent="toggleNavigation">
-		<Menu :size="24" title="" decorative />
-	</a>
+	<Actions class="app-navigation-toggle">
+		<ActionButton
+			:aria-expanded="open ? 'true' : 'false'"
+			:aria-label="label"
+			aria-controls="app-navigation-vue"
+			@click="toggleNavigation">
+			<Menu
+				slot="icon"
+				:size="24"
+				title=""
+				decorative />
+			{{ label }}
+		</ActionButton>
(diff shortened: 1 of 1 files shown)


ma12-co (approved):

src/components/AppNavigationToggle/AppNavigationToggle.vue
```diff
@@ -22,60 +22,62 @@
...
-		@click.prevent="toggleNavigation"
-		@keydown.space.exact.prevent="toggleNavigation">
-		<Menu :size="24" title="" decorative />
-	</a>
+	<Actions class="app-navigation-toggle">
+		<ActionButton
+			:aria-expanded="open ? 'true' : 'false'"
+			:aria-label="label"
+			aria-controls="app-navigation-vue"
+			@click="$emit('update:open', !open)">
```
skjnldsv, line 30: really don't like actions inside templates, please put back into a method

skjnldsv, line 30: really don't like actions inside templates, please put back into a method

skjnldsv (changes requested):

src/components/AppNavigationToggle/AppNavigationToggle.vue
raimund-schluessler: Done.

jancborchardt (changes requested):
When testing with a screenreader, the aria label seems to be duplicated, it reads:
> navigation, open navigation collapsed pushbutton, open navigation

And when the navigation is open:
> navigation, close navigation expanded pushbutton, close navigation

So it seems we should actually _not_ have the aria-label because the "navigation" already comes from the `role="navigation"` on the parent. Then it would just be "navigation, expanded pushbutton".

raimund-schluessler:
> 
> 
> When testing with a screenreader, the aria label seems to be duplicated, it reads:
> 
> > navigation, open navigation collapsed pushbutton, open navigation
> 
> And when the navigation is open:
> 
> > navigation, close navigation expanded pushbutton, close navigation
> 
> So it seems we should actually _not_ have the aria-label because the "navigation" already comes from the `role="navigation"` on the parent. Then it would just be "navigation, expanded pushbutton".

@jancborchardt Would it be possible that you adjust that? I don't have a screenreader at hand.

jancborchardt:
> @jancborchardt Would it be possible that you adjust that? I don't have a screenreader at hand.

I could try – but which system are you on? If you use GNOME, you can simply go to "Universal Access" and activate the screenreader from there. :)

jancborchardt:
Fixed it :) @raimund-schluessler the issue was simply that the `aria-label` was duplicating the text which was already in the ActionButton, so removing the `aria-label` fixed it.

jancborchardt (approved):
Works great now! :)

jancborchardt:
@skjnldsv could you check and approve if your change request is addressed? :)

skjnldsv (approved):

jotoeri:
How about a backport here? 😉

Issue #613, opened 2019-09-30
Avatar component not focusable with keyboard


georgehrke:
Meaning that the contacts-menu is not also not accessible.

skjnldsv:
:fixit: :D

PVince81:
ideally we should move to the Actions button as it has everything needed for keyboard focus and moving through the items with arrows: https://github.com/nextcloud/nextcloud-vue/pull/1670

PVince81:
WIP PR here: https://github.com/nextcloud/nextcloud-vue/pull/1854

but not sure if it can be finishable as the interaction in the popover is missing

PR #1854, opened 2021-04-16
Make avatar menu focusable with keyboard

PVince81:
Fixes https://github.com/nextcloud/nextcloud-vue/issues/613

This makes the avatar component focussable with the keyboard and also hitting the enter key will open the menu.
However the menu is not navigable and I'm not sure if we want to reimplement arrow navigation there.

- [x] multiple menus can be open with keyboard as enter doesn't re-close the other menus => solved
- [x] impossible to navigate menu entries with keyboard => tabs

Also, I heard that there will be a revamp of that popover, so the interaction in that popover would be different

### DIFF
--- a/l10n/messages.pot
+++ b/l10n/messages.pot
@@ -17,6 +17,9 @@ msgstr ""
 msgid "Animals & Nature"
 msgstr ""
 
+msgid "Avatar of {displayName}"
+msgstr ""
+
 msgid "Cancel changes"
 msgstr ""
 
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -36,7 +36,9 @@
 
 </docs>
 <template>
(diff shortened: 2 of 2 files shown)


PVince81:
I managed to make this work by automatically setting the focus on the first element inside the menu.
Then it's possible to tab to other menu items. Sadly also possible to leave the popover by tabbing out but there isn't much we can do about this, it's a general, separate issue.

So you can hit enter to open the trigger, then hit enter again on a menu link.

And when the popover is open, if you hit the Escape key, the popover closes and you're back on the trigger. That's near.

Minimal, but neat.

PVince81:
Note: I've developed and tested this with the Talk app and with the participant list on the right of a conversation.

ma12-co (approved):
tested and works

GretaD (approved):

PR #1875, opened 2021-04-23
Add CheckboxRadio component

skjnldsv:
- Uses the Material design icons
- Supports simple, sets and single-set checkboxes + radio (see docs)
- Supports indeterminate state
- Very straightforward! :rocket: 

A few examples:
image
image
image

### DIFF
--- a/src/components/CheckboxRadio/CheckboxRadio.vue
+++ b/src/components/CheckboxRadio/CheckboxRadio.vue
@@ -1,0 +1,347 @@
+<!--
+  - @copyright Copyright (c) 2021 John Molakvoæ <skjnldsv@protonmail.com>
+  -
+  - @author John Molakvoæ <skjnldsv@protonmail.com>
+  -
+  - @license GNU AGPL version 3 or any later version
+  -
+  - This program is free software: you can redistribute it and/or modify
+  - it under the terms of the GNU Affero General Public License as
+  - published by the Free Software Foundation, either version 3 of the
+  - License, or (at your option) any later version.
+  -
+  - This program is distributed in the hope that it will be useful,
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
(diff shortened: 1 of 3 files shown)


jancborchardt:
Wooooooah very nice! :) Sooo good to have this standardized.

Some design feedback:
- Each of the rows should be 44px height to conform to the min clickable/tappable area. The resulting whitespace will also look nice as  separation.
- The radio set using checkmarks looks a bit off. Normally it’s a dot in a circle, like so https://material.io/develop/web/components/selection-controls/radio-buttons (and like we do currently already)
- What is the "Standard checkbox set" with a square inside the square? If that is the indeterminate setting – we should keep it how it looks like for us currently (with the dash/minus in there) since that is also kind of a standard: https://material.io/components/checkboxes#behavior

jancborchardt (approved):
Nice, works very well now, looks great and is accessible. :+1: design-wise!

src/components/CheckboxRadio/CheckboxRadio.vue
```diff
@@ -0,0 +1,347 @@
...
+		 * Unique id attribute of the input
+		 */
+		id: {
+			type: String,
+			default: () => 'checkbox-radio-' + GenRandomId(),
+			validator: id => id.trim() !== '',
+		},
+
+		/**
+		 * Input name. Especially required for radios set
```
ChristophWurst, line 156: "especially required" -> is it required or not?

src/components/CheckboxRadio/CheckboxRadio.vue
```diff
@@ -0,0 +1,347 @@
...
+		/**
+		 * Disabled state
+		 */
+		disabled: {
+			type: Boolean,
+			default: false,
+		},
+
+		/**
+		 * Disabled state
```
ChristophWurst, line 197: :thinking:

src/components/CheckboxRadio/CheckboxRadio.vue
skjnldsv: You can't do conditional required.
It is for radio, unless you want a single radio, which you can, buuuuut doesn't make any ses=nse xD

ChristophWurst: then say "required for radio, optional for checkbox"?

skjnldsv: :+1:

ChristophWurst (approved):

Issue #998, opened 2020-04-07
Datepicker in sidebar cut-off at certain zoom levels


raimund-schluessler:
Since https://github.com/nextcloud/nextcloud-vue/issues/994#issuecomment-609953726 was closed, here now the stand-alone issue.

The datepicker is cut-off at certain zoom levels. This happens e.g. for latest master of Tasks and Calendar, which both use nextcloud/vue@1.4.1. It also still happens after #995 was merged.

Tasks@nextcloud/vue-1.4.1 with 110% zoom:
<img width="1276" alt="Tasks" src="https://user-images.githubusercontent.com/2496460/78653976-2cdfa980-78c4-11ea-8118-0d142377bcae.png">

Tasks@nextcloud/vue-9f525decb32e9e89f0c4fdeefb266c356199c4d2 (after #995) with 110% zoom:
<img width="1278" alt="Tasks_latest" src="https://user-images.githubusercontent.com/2496460/78657116-99f53e00-78c8-11ea-86c0-3b92c89bba50.png">

Calendar@nextcloud/vue-1.4.1 with 110% zoom:
<img width="1277" alt="Calendar" src="https://user-images.githubusercontent.com/2496460/78656297-78478700-78c7-11ea-9a6b-0f2c4feaf96a.png">

The reason for the datepicker being cut-off is, that `#app-sidebar` has
```css
#app-sidebar[data-v-120227ac] {
    ...
    overflow-y: auto;
    overflow-x: hidden;
}
```
and the datepicker is a child of the sidebar. Is there a reason why we don't append the datepicker to the body: https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/DatetimePicker/DatetimePicker.vue#L83 ?

cc @georgehrke @skjnldsv

skjnldsv:
Yes, we had the same issue with navigation in #989. We ended up removing the id app-navigation and move to standalone classes so we're independent from the server.
And now that the content is properly designed, we don't need the overflow anymore I guess :)

raimund-schluessler:
This is still an issue in the latest version. The lines https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/AppSidebar/AppSidebar.vue#L522-L523 cut off everything overflowing the sidebar including the datepicker.

Can we remove the overflow, or should we adjust the datepicker to not be a descendent of the sidebar but e.g. body instead?

skjnldsv:
[...]

PR #1881, opened 2021-04-26
Don't scroll datepicker in AppSidebar with body

raimund-schluessler:
This is the IMHO better but also more interfering solution to #998 than https://github.com/nextcloud/nextcloud-vue/pull/1880 as it needs changes to AppSidebar and the datepicker library. In difference to #1880, the datepicker dropdown will not scroll with body.

In order to allow an absolutely positioned element (the datepicker) to overflow its ancestor although it has `overflow: hidden` set, the first positioned ancestor of the datepicker has to be an ancestor of the element with `overflow: hidden` as well (see https://css-tricks.com/popping-hidden-overflow/ for a bit of background and explanation).

This means we have to adjust `AppSidebar` to look like this:

```html
<aside id="app-sidebar-vue" class="app-sidebar"> <!-- first positioned parent, position: sticky; -->
	<div class="app-sidebar-wrapper"> <!-- element with overflow-x: hidden; overflow-y: auto; -->
		<datepicker> <!-- absolutely positioned datepicker dropdown-->
	</div>
</aside>
```

Then the datepicker is positioned relative to `aside` and is allowed to overflow `app-sidebar-wrapper`. Since the datepicker library currently assumes that the datepicker is `position: relative;` it needs a fix to account for another element to be the first positioned ancestor. I have done this in https://github.com/raimund-schluessler/vue2-datepicker/commit/df040dbd5929f8f83f19390f396c924099ac7a38 and would send a PR to the datepicker library when we decide that this is a viable solution.

The great thing is that the datepicker is not cut-off, all scrolling inside the AppSidebar still works and the datepicker is static when scrolling body (tested in Firefox and Chrome):

datepicker_better

The changes to AppSidebar look more drastic than they actually are. I just added another div wrapping the AppSidebar, but this changes the indentation of all elements.
However, since the AppSidebar structure would change, custom adjustments maybe done by some apps might break (but should be fixable easily).

### DIFF
--- a/src/components/AppSidebar/AppSidebar.vue
+++ b/src/components/AppSidebar/AppSidebar.vue
@@ -124,110 +124,111 @@
 <template>
 	<transition name="slide-right" appear>
 		<aside id="app-sidebar-vue" class="app-sidebar">
-			<header :class="{
-					'app-sidebar-header--with-figure': hasFigure,
-					'app-sidebar-header--compact': compact,
-				}"
-				class="app-sidebar-header">
-				<!-- close sidebar button -->
-				<a
-					v-tooltip.auto="closeTranslated"
-					href="#"
-					class="app-sidebar__close icon-close"
-					@click.prevent="closeSidebar" />
-
-				<!-- container for figure and description, allows easy switching to compact mode -->
-				<div class="app-sidebar-header__info">
-					<!-- sidebar header illustration/figure -->
-					<div v-if="hasFigure && !empty"
-						:class="{
-							'app-sidebar-header__figure--with-action': hasFigureClickListener
-						}"
-						class="app-sidebar-header__figure"
-						:style="{
-							backgroundImage: `url(${background})`
-						}"
-						@click="onFigureClick">
-						<slot class="app-sidebar-header__background" name="header" />
-					</div>
-
-					<!-- sidebar details -->
-					<div v-if="!empty"
-						:class="{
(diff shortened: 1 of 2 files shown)


nickvergessen:
Better reviewed as https://github.com/nextcloud/nextcloud-vue/pull/1881/files?w=1

But this is not my level of expertise

raimund-schluessler:
@nextcloud/vuejs @skjnldsv Ping :slightly_smiling_face: 
Could I have a decision if we want to pursue this further? If so, I would rebase the PR and push the necessary upstream changes further.

skjnldsv:
> Could I have a decision if we want to pursue this further?

Honestly, the amount of changes to this PR is humongous :(
I really would not touch the AppSidebar component as this is stable. I am afraid of yet another brekage

raimund-schluessler:
> > Could I have a decision if we want to pursue this further?
> 
> Honestly, the amount of changes to this PR is humongous :( I really would not touch the AppSidebar component as this is stable. I am afraid of yet another brekage

Well, humongous is a bit exaggerated. Neglacting whitespace changes, the PR "only" adds another `div` to the sidebar as a wrapper (and adjusts the CSS to make this work). But I get your point, breaking something is not excluded for sure.
I guess we can live with the Datepicker scrolling with body then.

skjnldsv:
> Well, humongous is a bit exaggerated

Sorry, that's a bit exaggerated yes :see_no_evil:

PR #1889, opened 2021-04-27
Add loading state to CheckboxRadio

skjnldsv:
Peek 27-04-2021 18-53


ma12-co (approved):
Looks good! 
off topic: I think that the hover color should be `var(--color-background-hover)`, we're using the primary light for active items and It would be nice to keep this consistent

skjnldsv:
> off topic: I think that the hover color should be `var(--color-background-hover)`, we're using the primary light for active items and It would be nice to keep this consistent

@jancborchardt can you comment here? :)
I'm fine with both

ChristophWurst (approved):
:heart: 

I had to manually do this in some apps and it always made the input jump when there is just a single pixel off between input and replacement image … guess that won't happen again :)

PR #1890, opened 2021-04-28
Implement switch mode to CheckboxRadioSwitch

### LINKED ISSUE #1710: Implement switch component
All accross Nextcloud we're using checkboxes as toggles. This is not an ideal UI/UX practice and we should use a switch in those cases.

More info: https://uxmovement.com/buttons/when-to-use-a-switch-or-checkbox/

For example, this should be a toggle: 

https://user-images.githubusercontent.com/26852655/107814937-97763700-6d6a-11eb-9eef-c05835739bee.mp4

skjnldsv:
Fix #1710 

image

### DIFF
--- a/src/components/CheckboxRadioSwitch/CheckboxRadioSwitch.vue
+++ b/src/components/CheckboxRadioSwitch/CheckboxRadioSwitch.vue
@@ -24,15 +24,16 @@
 
 ### General description
 
-This is a standard input checkbox/radio design
+This is a simple input checkbox, radio and switch design.
+Please have a look at proper usage and recommendations: https://material.io/components/checkboxes
 
 ### Standard checkbox
 ```vue
 <template>
 	<div>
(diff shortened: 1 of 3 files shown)


juliushaertl (approved):
Very nice

skjnldsv:
image

ma12-co (approved):
Very nice!
The only feedback I have is that I find it a bit small, could we increase the size of it to match the height of the checkbox and have the width scale proportionally?

skjnldsv:
> The only feedback I have is that I find the switch a bit small, could we increase the size of it to match the height of the checkbox and have the width scale proportionally?

Can be easily done here: https://github.com/nextcloud/nextcloud-vue/blob/866c3c030a4a7ebc8203f3e2597fcf12bfe81be7/src/components/CheckboxRadioSwitch/CheckboxRadioSwitch.vue#L268
Ask Jan about it :)

PR #1897, opened 2021-04-29
Fix appnavigationitem layout

### LINKED ISSUE #1868: AppNavigation counter have wrong right margin
image

ma12-co:
Fix #1868 and adds 4px padding on top and bottom too so that the actions hover circle doesn't intersect with the boundary of the element

Screenshot from 2021-04-29 12-31-39
Screenshot from 2021-04-29 12-31-58

### DIFF
--- a/src/components/AppNavigationItem/AppNavigationItem.vue
+++ b/src/components/AppNavigationItem/AppNavigationItem.vue
@@ -156,7 +156,10 @@ Just set the `pinned` prop.
 
 		<!-- Counter and Actions -->
 		<div v-if="hasUtils" class="app-navigation-entry__utils">
-			<slot name="counter" />
+			<div v-if="$slots.counter"
+				class="app-navigation-entry__counter-wrapper">
+				<slot name="counter" />
+			</div>
 			<Actions menu-align="right"
 				:placement="menuPlacement"
 				:open="menuOpen"
@@ -448,6 +451,7 @@ export default {
 	box-sizing: border-box;
 	width: 100%;
 	min-height: $clickable-area;
+	padding: 4px;
 
 	// When .active class is applied, change color background of link and utils. The
 	// !important prevents the focus state to override the active state.
@@ -582,11 +586,12 @@ export default {
 	display: flex;
 	align-items: center;
 	flex: 0 1 auto;
-	// visually balance the menu so it's not
-	// stuck to the scrollbar
-	.action-item {
-		margin-right: 2px;
-	}
+}
+
+/* counter */
+.app-navigation-entry__counter-wrapper {
+	// Add slightly more space to the right of the counter
+	margin-right: 2px;
 }
 
 // STATES


src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -156,7 +156,9 @@ Just set the `pinned` prop.
 
 		<!-- Counter and Actions -->
 		<div v-if="hasUtils" class="app-navigation-entry__utils">
-			<slot name="counter" />
+			<div class="app-navigation-entry__counter">
```
skjnldsv, line 159: v-if counter?

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -582,11 +585,12 @@ export default {
...
-	.action-item {
-		margin-right: 2px;
-	}
+}
+
+/* counter */
+.app-navigation-entry__counter {
+	// Add slightly more space to the right of the counter
+	margin-right: 2px;
 }
```
raimund-schluessler, line 594: I would remove this, so it is aligned right with the ActionItem. Then also the container is not necessary and the HTML structure doesn't have a double `app-navigation-entry__counter` element. I guess you wanted to center the counter above the ActionItem, but that anyway doesn't really work with a margin-right, since the counter can have different widths. See here (I disabled `margin-right` end it looks good imho):

AppNavigationItem

src/components/AppNavigationItem/AppNavigationItem.vue
raimund-schluessler: Having this container will lead to two element having the `app-navigation-entry__counter` class:
```
<div class="app-navigation-entry__counter">
   <div data-v-794921ed="" data-v-9941="" class="app-navigation-entry__counter">
		99+
	</div>
</div>
```
I think this doesn't make sense. And it isn't necessary anyway. See https://github.com/nextcloud/nextcloud-vue/pull/1897/files#r623671032

raimund-schluessler (changes requested):

src/components/AppNavigationItem/AppNavigationItem.vue
raimund-schluessler: However, this wouldn't give a margin between counter and action in case both are present. If we want to have this as well, I would propose to just add
```
&::v-deep .app-navigation-entry__counter {
   margin-right: 2px;
}
```
and leave the rest untouched.

ma12-co:
@raimund-schluessler I added the extra margin on the right of the counter for a visual balance, the element is much shorter than the actions component and it looked weird too close to the right border of the wrapper.

I can come up with a different classname though, what about `...__counter-wrapper`?

raimund-schluessler:
> 
> 
> @raimund-schluessler I added the extra margin on the right of the counter for a visual balance, the element is much shorter than the actions component and it looked weird too close to the right border of the wrapper.

The thing is that the counter can have different widths. So setting a fixed margin will only achieve a "balance" for a single width. In all other cases, it won't be balanced *and* not aligned at the right side. That's why I would prefer a right alignment.

raimund-schluessler:
> I can come up with a different classname though, what about `...__counter-wrapper`?

A wrapper won't be necessary if we just use `&::v-deep .app-navigation-entry__counter`.

ma12-co:
> A wrapper won't be necessary if we just use &::v-deep .app-navigation-entry__counter.

The thing is that appnavigationcounter will be deprecated and [developers will start using this instead](https://github.com/nextcloud/nextcloud-vue/pull/1895#issuecomment-829934215), so we'd have to v:deep two class names with the same rule

ma12-co:
> The thing is that the counter can have different widths. So setting a fixed margin will only achieve a "balance" for a single width. In all other cases, it won't be balanced and not aligned at the right side. That's why I would prefer a right alignment.

About this, the action's hover feedback is visible only when hovering, most of the times you only see the 3 dots icon, so the spacing of the element to the right is even bigger. 

This is what happens without that extra padding and that's what I'm trying to tackle here. It's sort of independent of the amount of characters in the counter, but the less of them, the more out of balance it looks. 

Screenshot from 2021-04-30 11-11-46

raimund-schluessler:
> 
> 
> > A wrapper won't be necessary if we just use &::v-deep .app-navigation-entry__counter.
> 
> The thing is that appnavigationcounter will be deprecated and [developers will start using this instead](https://github.com/nextcloud/nextcloud-vue/pull/1895#issuecomment-829934215), so we'd have to v:deep two class names with the same rule

Ah, yeah, fair enough. Then let's just use a different class name.

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -448,6 +450,7 @@ export default {
 	box-sizing: border-box;
 	width: 100%;
 	min-height: $clickable-area;
+	padding: 4px;
```
raimund-schluessler, line 453: @ma12-co This needs to be a `padding-right` only.

raimund-schluessler:
> This is what happens without that extra padding and that's what I'm trying to tackle here. It's sort of independent of the amount of characters in the counter, but the less of them, the more out of balance it looks.

This is how it looks with longer counters when right-aligned:

Screenshot_2021-04-30 Nextcloud Vue Style Guide


And the same elements with the changes from this PR:

Screenshot_2021-04-30 Nextcloud Vue Style Guide(1)

I agree the changes are subtle and probably a matter of taste. I just like alignment 😉 I leave it up to you.

src/components/AppNavigationItem/AppNavigationItem.vue
ma12-co: I introduced that to give a bit of air to the hover circle of the actions and again balance to the whole thing :) 

Before:
Screenshot from 2021-04-30 11-29-31

After:
Screenshot from 2021-04-30 11-28-34

raimund-schluessler: I often have the impression that we use quite a lot of white-space which sometimes wastes screen space. And I haven't had the feeling that more is necessary for AppNavigationItems (since e.g. the button background is only visible on hover). But then again, personal preference. If the others prefer it this way as well, fine with me.

raimund-schluessler: Maybe we could reduce it to two pixels on top and bottom, since with adjacent entries this sums up to twice the distance.

raimund-schluessler (approved):

skjnldsv (approved):

raimund-schluessler:
@ma12-co @skjnldsv This change (the wrapper around the counter) breaks the Calendar and the Tasks app, since both apps use the counter slot to show an avatar and the sharing action like so:
https://github.com/nextcloud/calendar/blob/master/src/components/AppNavigation/CalendarList/CalendarListItem.vue#L39-L45

Screenshot_2021-04-30 Tasks - Nextcloud

I am not really sure what is the proper solution here. The counter slot probably wasn't supposed to show an Avatar and an ActionItem in the first place. But there is no other slot to show something in the `app-navigation-entry__utils` element.
For now, adding
```css
	display: flex;
	align-items: center;
	flex: 0 1 auto;
}
```
mitigates the issue:

Screenshot_2021-04-30 Tasks - Nextcloud(1)

Also @ChristophWurst, because you will have the same problem in the Calendar app.

raimund-schluessler:
Also, increasing the height of the AppNavigationItem by 20% is to much, I think. Was this discussed with @nextcloud/designers and especially @jancborchardt? In my impression, we now waste a lot of screen space (especially height) in the AppNavigation, because every item is now 52px instead of 44px high.

Before we see 15+ entries:
Screenshot_2021-04-30 Tasks - Nextcloud(3)

After we only see 13 entries without a real benefit imho:

Screenshot_2021-04-30 Tasks - Nextcloud(2)

skjnldsv:
> The counter slot probably wasn't supposed to show an Avatar and an ActionItem in the first place.

Yep :disappointed: 

> Also, increasing the height of the AppNavigationItem by 20% is to much

Ah right @ma12-co, this is supposed to be a bugfix. This is too much, please let's keep the 44px, this is not a feature-pr :)

skjnldsv:
@ma12-co please only fix the margin :)

PR #1911, opened 2021-05-03
Remove AppNavigationItem flex order for AppNavigationCaption

skjnldsv:
This was a trick to differentiate the pinned to the non-pinned entries, but other elements does not have this `pinned` order feature (like the AppNavigationCaption), and therefore gets stuck all in the top.

image


raimund-schluessler:
Ah, this explains why I had to do this: 🙈 
https://github.com/nextcloud/tasks/blob/master/src/views/AppNavigation.vue#L449-L451

raimund-schluessler (approved):

ma12-co (approved):

PR #1912, opened 2021-05-03
Fix counter margin

### LINKED ISSUE #1868: AppNavigation counter have wrong right margin

ma12-co:
fix #1868

Bringing back only the counter fix.

### DIFF
--- a/src/components/AppNavigationItem/AppNavigationItem.vue
+++ b/src/components/AppNavigationItem/AppNavigationItem.vue
@@ -156,7 +156,10 @@ Just set the `pinned` prop.
 
 		<!-- Counter and Actions -->
 		<div v-if="hasUtils" class="app-navigation-entry__utils">
-			<slot name="counter" />
+			<div v-if="$slots.counter"
+				class="app-navigation-entry__counter-wrapper">
+				<slot name="counter" />
+			</div>
 			<Actions menu-align="right"
 				:placement="menuPlacement"
 				:open="menuOpen"
@@ -447,6 +450,7 @@ export default {
 	box-sizing: border-box;
 	width: 100%;
 	min-height: $clickable-area;
+	padding-right: 4px;
 
 	// When .active class is applied, change color background of link and utils. The
 	// !important prevents the focus state to override the active state.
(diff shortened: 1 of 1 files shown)


raimund-schluessler (changes requested):
Actions now miss the right margin:

Screenshot_2021-05-03 Nextcloud Vue Style Guide

ma12-co:
> Actions now miss the right margin

:facepalm: done :sweat_smile:

raimund-schluessler (approved):
Looks good now.

Screenshot_2021-05-03 Nextcloud Vue Style Guide(1)

raimund-schluessler:
The fix for the Calendar app will still be necessary, but I guess that's inevitable and can be fixed quickly in the Calendar app.

skjnldsv (approved):

skjnldsv (changes requested):
That's a breaker though, not really a fix

skjnldsv:
If we need to fix something for it to ork in another app, that's not ok. This needs to be backportable to stable3, meeaning there should be no other change needed on the app than bumping `@nextcloud/vue`'s version :)

raimund-schluessler:
> 
> 
> If we need to fix something for it to work in another app, that's not ok.
> This needs to be backportable to `stable3`, meaning there should be no other change needed on the app than bumping `@nextcloud/vue`'s version :)

Technically, yes. But using the counter slot like Tasks and Calendar do it, was never really supported by nextcloud/vue. So I am a bit undecided if we need to account for it.

skjnldsv:
The question is, does it works in any of 3.x.x of nextcloud/vue?
If so I assume following proper semver means we need to have this fixed properly?

raimund-schluessler:
> 
> 
> The question is, does it works in any of 3.x.x of nextcloud/vue?
> If so I assume following proper semver means we need to have this fixed properly?

Yes, it does work in 3.x.x. If we want to not break this usage, we could either try to not introduce this wrapper around the CounterBubble or maybe apply this rule to it: 
```css
    margin-right: 2px;
    display: flex;
    align-items: center;
    flex: 0 1 auto;
}
```
This should also keep it working, I think (didn't test it yet, can do it only in the evening).

skjnldsv:
Why do we need this wrapper at all btw?
I'm also ok adding the wrapper for 4.x.x but only the margin for 3.x.x @ma12-co :)

ma12-co:
> Why do we need this wrapper at all btw?
To apply the extra margin without using v-deep

ma12-co:
Applyed flex rules to solve calendar issue

Screenshot from 2021-05-06 10-45-04

skjnldsv (approved):

raimund-schluessler:
I can confirm it now works without adjustments, https://github.com/nextcloud/tasks/pull/1564.

PR #1946, opened 2021-05-10
Properly align appnavigationtoggle

marcoambrosini:
...Upstreaming more talk v-deeps

Before: 
Screenshot from 2021-05-10 15-25-53

After:
Screenshot from 2021-05-10 15-29-15

### DIFF
--- a/src/components/AppNavigationToggle/AppNavigationToggle.vue
+++ b/src/components/AppNavigationToggle/AppNavigationToggle.vue
@@ -76,8 +76,8 @@ export default {
 
 button.app-navigation-toggle {
 	position: absolute;
-	top: 0;
-	right: 0;
+	top: $topbar-margin;
+	right: - $topbar-margin;
 	margin-right: - $clickable-area;
 }
 
--- a/src/assets/variables.scss
+++ b/src/assets/variables.scss
@@ -55,3 +55,6 @@ $navigation-width: 300px;
 
 // mobile breakpoint
 $breakpoint-mobile: 1024px;
+
+// top-bar spacing
+$topbar-margin: 4px;
--- a/src/components/AppContent/AppContent.vue
+++ b/src/components/AppContent/AppContent.vue
@@ -29,6 +29,10 @@ This components provides a wrapper around the main app's content.
 Single-column layouts can just use the default slot. A resizable column
 can be added by providing content to the named slot `list`.
 
+### CSS variables
+In the css section some css variables are declared and will be available for
+all the children of the AppContent component
+
 ### Examples
 
 #### Usage: Single-column content
@@ -325,6 +329,10 @@ export default {
 	// Overriding server styles TODO: cleanup!
 	margin: 0 !important;
 	background-color: var(--color-main-background);
+
+	// Variables
+	// the whitespace between the topbar content and its edges
+	--topbar-margin: $topbar-margin: 4px;
 }
 
 // Mobile list/details handling
--- a/src/components/AppContent/AppContent.vue
+++ b/src/components/AppContent/AppContent.vue
@@ -332,7 +332,7 @@ export default {
 
 	// Variables
 	// the whitespace between the topbar content and its edges
-	--topbar-margin: $topbar-margin: 4px;
+	--topbar-margin: $topbar-margin;
 }
 
 // Mobile list/details handling


src/components/AppNavigationToggle/AppNavigationToggle.vue
```diff
@@ -79,8 +79,8 @@ export default {
 
 button.app-navigation-toggle {
 	position: absolute;
-	top: 0;
-	right: 0;
+	top: 8px;
```
skjnldsv, line 82: Somehow I feel those should be standardize or something :thinking:  :shrug:

ma12-co: What do you mean? Variables?

skjnldsv: yeah, like I see you're also aligning the input and menu of the appnavigation in talk, I assume you're also using 8px? I think that should really be standardized :)

raimund-schluessler:
Honestly, I do not think this is a good idea. The `AppNavigationToggle` has no own space in our layout, i.e. the `app-content` does not reserve space for it. This means we have a lot of problems with the AppNavigationToggle as it is positioned already, since it overlaps the content see https://github.com/nextcloud/server/issues/16934

If we start to move it around now, this will become even worse. See the Tasks app for example (sorry, it's always Tasks 🙈):

AppNavigation

So, in case you really decide to do this, it will be a breaking change at least.

ma12-co:
@raimund-schluessler I understand your concern but I think that the spacing of the toggle relative to other elements is off. We fixed that in talk and I'm trying to bring custom changes we have there upstream. Of course this would be a breaking change. @skjnldsv thoughts?

raimund-schluessler:
> @raimund-schluessler I understand your concern but I think that the spacing of the toggle relative to other elements is off. We fixed that in talk and I'm trying to bring custom changes we have there upstream. Of course this would be a breaking change. @skjnldsv thoughts?

The problem is that other apps will have to undo the custom changes of Talk because they won't work with their layout. So instead of having one custom CSS rule in Talk, we will have multiple (one in every other app instead).

raimund-schluessler:
> 
> 
> @raimund-schluessler I think that other apps could benefit from a standardized 8px spacing too.
> Looking at your screenshot, in tasks it seems that the toggle with the new margin is now almost aligned with the input field. With this change, If the input field had a height of 44px it could align perfectly with the toggle, similar to the avatar here:
> 
> Screenshot from 2021-05-11 09-59-44

I don't say that standardization is bad, or that other apps won't benefit from having it aligned. But unless we find a proper space for the `AppNavigationToggle` in the layout which accounts for its presence, every app will still have to do custom changes to it's position.

For example, in case we adapt the files app to use the vue-components as well at some point, having an 8 px top spacing will just not work with the breadcrumb bar (since it doesn't have a padding at the top).

And on a side note: In my impression, the AppNavigationToggle shouldn't be aligned with any element in the content area, since it doesn't logically belong there (e.g. having it aligned with the input field in Tasks, suggests a logical relation to it, which it hasn't). Having it top aligned with the AppNavigation pane is the proper solution imho, but that's just my taste.

raimund-schluessler:
>    * Tasks could be similar, as per your screenshot there’s very little to adjust @raimund-schluessler – what Android does is that the navigation toggle is actually _inside_ the search field on the left.

I don't have a problem adjusting the layout. I just understood it so that you want to have the same button position for all apps. And I am a bit skeptical that this will really work out well. Mainly because many apps don't position the AppNavigationToggle button well just as it is located now. And moving it further into the content will make this problem even worse (see e.g. Contacts or Calendar).

jancborchardt:
@raimund-schluessler your example looks great! :)

And yeah, the positioning probably doesn’t have to be exactly the same in all apps – but in no app should it stick to the very top left and be unaligned to any content, like it is now.

marcoambrosini:
Let's go with 4 then, I'll reduce talk's topbar height

marcoambrosini:
Done, please review!

skjnldsv:
> Either we have a dedicated variable that matches such spacing? ([#1946 (comment)](https://github.com/nextcloud/nextcloud-vue/pull/1946#discussion_r629481568)) Or we find something different?

Still doesn't answer this question?
Hard-coded px for such standard thing should not happen imho :)

marcoambrosini:
Ok, so `var(--top-bar-spacing)` @skjnldsv ??

skjnldsv:
> Ok, so `var(--top-bar-spacing)` @skjnldsv ??

:shrug:

skjnldsv (changes requested):
Hum, shouldn't those also be available as a css var within content?
We should maybe start doing this (add them to the Content so anything within can use it for example?)

marcoambrosini:
ping @skjnldsv

marcoambrosini:
What should we do here????

jancborchardt:
> Ok, so `var(--top-bar-spacing)` @skjnldsv ??

Isn’t that a good approach as you suggested?

jancborchardt:
By the way @raimund-schluessler, with the new breadcrumb component with bigger visual elements, maybe the 8px spacing also makes sense for Files now? The slightly more whitespace is definitely nice, and then it would fit for Talk, Files and probably others too.

marcoambrosini:
> I would like this new css variable to be documented tbh

@skjnldsv where should we add the docs? To this file?

skjnldsv:
> > I would like this new css variable to be documented tbh
> 
> @skjnldsv where should we add the docs? To this file?

Very good question :)
Maybe we need to dedicated design system documentation. 
There was the old one in the docs, but it's proably not very used nor up to date anymore.
We could add a dedicated component-related one here? :thinking: 

At the very least, I would add that to the AppContent documentation

raimund-schluessler:
How should we proceed here? We are heading for a new major release and it would be good to have this included, I think.

marcoambrosini:
Added a couple of lines of docs in the docs section and in the css section where the variable is declared, let's get this in?

raimund-schluessler (changes requested):

raimund-schluessler (approved):

PR #1948, opened 2021-05-11
Fix user bubble when username contains a space

artonge:
The user bubble is broken when there is a space in the username.

This was due to the parsed value being sanitized before applying regexp on it.

I simply changed the regexp to match `&quot;` instead of `"`.

This should fix: https://github.com/nextcloud/server/issues/26861

Before | After
-- | --
Screenshot from 2021-05-11 18-44-02 | Screenshot from 2021-05-11 18-45-17

### DIFF
--- a/src/components/RichContenteditable/RichContenteditable.vue
+++ b/src/components/RichContenteditable/RichContenteditable.vue
@@ -80,6 +80,19 @@ export default {
 						status: 'away',
 					},
 					subline: 'Visiting London',
+				},
+				'Test 03': {
+					icon: 'icon-user',
+					id: 'Test 03',
+					label: 'Test 03',
+					source: 'users',
+					status: {
+						clearAt: null,
+						icon: '🎡',
+						message: 'Having space in my name',
+						status: 'in space',
+					},
+					subline: 'Visiting London',
 				}
 			}
 		}
--- a/src/mixins/richEditor/index.js
+++ b/src/mixins/richEditor/index.js
@@ -32,7 +32,7 @@ const MENTION_START = '(?:^|\\s)'
 // Anything that is not text or end-of-line. Non-capturing group
 const MENTION_END = '(?:[^a-z]|$)'
 export const USERID_REGEX = new RegExp(`${MENTION_START}(@[a-zA-Z0-9_.@\\-']+)(${MENTION_END})`, 'gi')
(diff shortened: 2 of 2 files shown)


skjnldsv (approved):

ChristophWurst (approved):

ChristophWurst:
backport?

elhananjair:
Hello there is this issue(#26861 fixed yet?  because am still facing this issue in NC 21.0.4 Version.

artonge:
@elhananjair [NC 21.0.4 uses `nextcloud/vue@v3.2.0`](https://github.com/nextcloud/server/blob/v21.0.4/package.json#L42). This fix has only made it to `nextcloud/vue@v4.0.0` which will be use in NC 23

elhananjair:
> @elhananjair [NC 21.0.4 uses `nextcloud/vue@v3.2.0`](https://github.com/nextcloud/server/blob/v21.0.4/package.json#L42). This fix has only made it to `nextcloud/vue@v4.0.0` which will be use in NC 23

ow okay I will wait for it, thanks.

elhananjair:
hey @artonge am really sorry to bring this thing again, I was just curious that this issue is not appearing in some nextcloud applications, such as deck, So does this mean some apps in nextcloud uses different versions of vue or built with different mechanism? Thanks

artonge:
> hey @artonge am really sorry to bring this thing again, I was just curious that this issue is not appearing in some nextcloud applications, such as deck, So does this mean some apps in nextcloud uses different versions of vue or built with different mechanism? Thanks

Yes, not all applications use this Vue component.

elhananjair:
@artonge I don't think the issue has been fixed in NC 23. I am trying here on Talk and Files app and its the same as before

artonge:
> @artonge I don't think the issue has been fixed in NC 23. I am trying here on Talk and Files app and its the same as before

Can you clear your cache and check again? Can you open another issue if the problem still persist?

elhananjair:
> > @artonge I don't think the issue has been fixed in NC 23. I am trying here on Talk and Files app and its the same as before
> 
> Can you clear your cache and check again? Can you open another issue if the problem still persist?

I have installed clean NC 23 and didn't upgrade from previous versions. So I don't think clearing cache or something like that won't help I guess. As you said let me open this issue again.

PR #1967, opened 2021-05-17
Fix first and last action input padding

raimund-schluessler:
This fixes the padding of the ActionInput in case it is the first or last entry. The currently set margin lead to a gap between the popover arrow and the input element.

Before:
Screenshot_2021-05-17 Nextcloud Vue Style Guide(2)

After:
Screenshot_2021-05-17 Nextcloud Vue Style Guide(3)

In case one switches from ActionButton to ActionInput like it happens in Calendar, this will still lead to a "jumping" of the popover menu, since the ActionButton has a different height than the ActionInput with the additional padding (see https://github.com/nextcloud/calendar/pull/3016). But the gap is not there anymore.

actioninput

On a side note: I would say the additional padding is not really necessary (or at least to high). But since this is supposed to be a bug fix PR, I didn't remove it, but instead made it work properly 😉. @jancborchardt maybe to check if this is really what it is supposed to look like.

### DIFF
--- a/src/components/ActionInput/ActionInput.vue
+++ b/src/components/ActionInput/ActionInput.vue
@@ -409,12 +409,12 @@ $input-margin: 4px;
 // add the same bottomMargin as the right padding
 // for visual balance
 li:last-child > .action-input {
-	margin-bottom: $icon-margin - $input-margin;
+	padding-bottom: $icon-margin - $input-margin;
 }
 
 // same for first item
 li:first-child > .action-input {
-	margin-top: $icon-margin - $input-margin;
+	padding-top: $icon-margin - $input-margin;
 }
 
 </style>


ma12-co (approved):

juliushaertl (approved):
Fine by me for a bugfix

> On a side note: I would say the additional padding is not really necessary (or at least to high). But since this is supposed to be a bug fix PR, I didn't remove it, but instead made it work properly wink. @jancborchardt maybe to check if this is really what it is supposed to look like.

I agree, either that or we need to make sure the spacing applies to all elements. The only thing where this might be odd is with the hover/focus feedback of the ActionButton, which if we increase just the top padding would look rather unaligned with the background then.

raimund-schluessler:
> I agree, either that or we need to make sure the spacing applies to all elements. The only thing where this might be odd is with the hover/focus feedback of the ActionButton, which if we increase just the top padding would look rather unaligned with the background then.

Not sure if it's a good idea to apply it to all elements. This will look really unaligned.

PR #1992, opened 2021-05-31
AppNavigationItem: hide utils if edit mode is active

korelstar:
If there are other actions or a counter in an AppNavigationItem and the edit mode is activated, the close button was not accessible:

Screenshot


skjnldsv (approved):

PVince81 (approved):
:+1:

PR #2008, opened 2021-06-04
Fix documentation after #1759

skjnldsv:
It compiles well, but fails to render

https://deploy-preview-2008--nextcloud-vue-components.netlify.app/


kesselb (approved):

gary-kim (approved):
Seems fine. We could also make it such that it takes a string but that seems like unnecessary complexity.

PR #2013, opened 2021-06-07
Revert "Replace node-sass with sass"

skjnldsv:
Reverts nextcloud/nextcloud-vue#2005

It breaks the style import of the documentation :shrug: 
Let's revert for 4.0.0

https://deploy-preview-2013--nextcloud-vue-components.netlify.app/


raimund-schluessler (approved):
Styleguide looks good again.

ChristophWurst:
IMO https://github.com/nextcloud/nextcloud-vue/pull/1943 should be reconsidered after this so people with a default setup of Node can still contribute to the library. Right now you can only build if you have the old node installed.

ChristophWurst (approved):

PR #2015, opened 2021-06-07
Check slots on updated hook

ma12-co:

### DIFF
--- a/src/components/ListItem/ListItem.vue
+++ b/src/components/ListItem/ListItem.vue
@@ -289,6 +289,8 @@ export default {
 		return {
 			hovered: false,
 			focused: false,
+			hasActions: false,
+			hasSubtitle: false,
 			displayActions: false,
 			menuOpen: false,
 		}
@@ -298,10 +300,6 @@ export default {
 
 		hasDetails() {
 			return this.details !== ''
-		},
-
-		hasActions() {
-			return (!!this.$slots.actions)
 		},
 
 		// This is used to decide which outer element type to use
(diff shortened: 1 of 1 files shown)


src/components/ListItem/ListItem.vue
```diff
@@ -400,6 +402,12 @@ export default {
 		handleActionsUpdateOpen(e) {
 			this.menuOpen = e
 		},
+
+		// Check if subtitle and actions slots is populated
+		checkSlots() {
+			this.hasActions = !!this.$slots.actions
+			this.hasSubtitle = !!this.$slots.subtitle
```
skjnldsv, line 409: Only if they're different from `this.hasActions` or `this.hasSubtitle`.
or else, endless loop! :warning:

skjnldsv (changes requested):

src/components/ListItem/ListItem.vue
ma12-co: done!

skjnldsv (approved):

src/components/ListItem/ListItem.vue
```diff
@@ -400,6 +402,16 @@ export default {
 		handleActionsUpdateOpen(e) {
 			this.menuOpen = e
 		},
+
+		// Check if subtitle and actions slots is populated
```
ChristophWurst, line 406:
```suggestion
		// Check if subtitle and actions slots are populated
```

ChristophWurst (approved):

PR #2028, opened 2021-06-10
Safely check the global appName and handle its inexistence

ChristophWurst:
Variable is set -> default is used, developers *can* overwrite the prop value
Variable is not set -> prop become required

Right now you can't use the AppContent without the global variable, because the code that sets the default is always evaluated and will throw a ReferenceError before the developer can even specify their own value. Ref https://github.com/nextcloud/mail/pull/5165#issuecomment-858724634.

Then we can avoid hacks like https://github.com/nextcloud/spreed/commit/c6627446672379bf1913ffe898df460d1776b4af

The code is untested because neither @GretaD nor I can build this library at the moment.

### DIFF
--- a/src/components/AppContent/AppContent.vue
+++ b/src/components/AppContent/AppContent.vue
@@ -172,11 +172,10 @@ export default {
 		/**
 		 * Specify the config key for the pane config sizes
 		 * Default is the global var appName if you use the webpack-vue-config
-		 * Otherwise it will be undefined and you will have to provide one
 		 */
 		paneConfigKey: {
 			type: String,
-			default: appName,
+			default: '',
 		},
 
 		/**
@@ -203,8 +202,21 @@ export default {
 
 	computed: {
 		paneConfigID() {
-			// Using the webpack-vue-config, appName is a global variable
-			return `pane-list-size-${this.paneConfigKey}`
+			// If provided, let's use it
+			if (this.paneConfigKey !== '') {
+				return `pane-list-size-${this.paneConfigKey}`
+			}
+
+			try {
+				// Using the webpack-vue-config, appName is a global variable
+				// This will throw a ReferenceError when the global variable is missing
+				// In that case either you provide paneConfigKey or else it fallback
+				// to a global storage key
+				return `pane-list-size-${appName}`
+			} catch (e) {
+				console.info('[INFO] AppContent:', 'falling back to global nextcloud pane config')
+				return 'pane-list-size-nextcloud'
+			}
 		},
 
 		detailsPaneSize() {


src/components/AppContent/AppContent.vue
```diff
@@ -120,6 +120,16 @@ import isMobile from '../../mixins/isMobile'
 
 const browserStorage = getBuilder('nextcloud').persist().build()
 
+let hasGlobalAppName = true
+let globalAppName
+try {
+	// This will throw a ReferenceError when the global variable is missing
+	// In that case we have to make the pane config key required
+	globalAppName = appName
```
PVince81, line 128: usually doing a `if (window.appName)` should work for checking for global var existence

ChristophWurst: I thought that was only valid in a browser context, I'm not sure how webpack usually defines globals. Are they pushed to `window` @skjnldsv ?

PVince81:
I'm also surprised that eslint doesn't fail on CI ? Usually one would need to add something like `// global appName` at the top of the file to declare it

ChristophWurst:
> I'm also surprised that eslint doesn't fail on CI ? Usually one would need to add something like `// global appName` at the top of the file to declare it

https://github.com/nextcloud/nextcloud-vue/blob/23535c1f442e5a26beb8f4963a8b66160e9b05f9/.eslintrc.js#L6 …

src/components/AppContent/AppContent.vue
skjnldsv: No, we don't want them on window

src/components/AppContent/AppContent.vue
```diff
@@ -120,6 +120,16 @@ import isMobile from '../../mixins/isMobile'
...
+} catch (e) {
+	hasGlobalAppName = false
+}
```
skjnldsv, line 131: Honestly, this is even worse in terms of code clarity.

Either put that whole block as a service and import it or just remove the appName and make it required

skjnldsv (changes requested):

skjnldsv:
> Then we can avoid hacks like [nextcloud/spreed@c662744](https://github.com/nextcloud/spreed/commit/c6627446672379bf1913ffe898df460d1776b4af)

It's not a hack, that is just because none of you are using the webpakc vue config like you're supposed to.

raimund-schluessler:
Could we maybe use a factory function for the default of this prop in combination with a custom validator and check there if `appName` is defined.
So something like this (untested kind-of pseudo code):
```js
props: {
    paneConfigKey: {
      type: String,
      default: function() {
        // Don't fail hard if appName is not globally defined
        try {
            return appName
        } catch (e) {
            return undefined
        }
      },
      validator: propValue => {
        // Check that propValue is not undefined but set as prop
        // when appName is not globally defined
        return !!propValue
      }
    }
  }
```

src/components/AppContent/AppContent.vue
raimund-schluessler: Please don't make it required, this will break all apps.

ChristophWurst:
> Could we maybe use a factory function for the default of this prop in combination with a custom validator and check there if `appName` is defined.

Yep, that was an alternative idea of mine as well. I just couldn't find any Vue docs that said factories to be allowed for primitive values. Their docs only have them for objects and array.

Guess we'll just have to try and hope for the best :crossed_fingers:

skjnldsv:
@ChristophWurst @raimund-schluessler 
please check out [`fbf756f` (#2028)](https://github.com/nextcloud/nextcloud-vue/pull/2028/commits/fbf756fc6eb8f8298ee358c7b3810be95ec71037)

This should fix this pr. If you're ok Christoph, we can merge

ChristophWurst (commented):
+1

skjnldsv (approved):

marcoambrosini (approved):

PR #1781, opened 2021-03-23
Update webpack to 5.70.0

raimund-schluessler:
This PR updates webpack to a new major version 5.70.0.

PR #2037, opened 2021-06-14
Use the common webpack and stylelint configuration

raimund-schluessler:
Simplifies the configuration a bit.


webpack.js
```diff
@@ -0,0 +1,158 @@
...
+					},
+				},
+			},
+		],
+	},
+	{
+		test: /\.(js|vue)$/,
+		use: 'eslint-loader',
+		enforce: 'pre',
+	},
```
skjnldsv, line 140: not needed anymore

webpack.js
```diff
@@ -0,0 +1,158 @@
...
+		test: /\.js$/,
+		loader: 'babel-loader',
+		exclude: BabelLoaderExcludeNodeModulesExcept([
+			'tributejs',
+		]),
+	},
+	{
+		test: /\.(png|jpg|gif|svg|eot|ttf|woff|woff2)$/i,
+		loader: 'url-loader',
+	},
```
skjnldsv, line 155: already covered by the vue config

webpack.js
```diff
@@ -0,0 +1,158 @@
...
+webpackConfig.plugins.push(new DefinePlugin({
+	SCOPE_VERSION,
+	TRANSLATIONS: JSON.stringify(translations),
+}))
+// Alias the default server assets path to link to local one
+webpackConfig.resolve.alias = {
+	'/core/css/../img': path.join(__dirname, 'styleguide/assets/img'),
+	'/core/css/../fonts': path.join(__dirname, 'styleguide/assets/fonts'),
+}
+webpackConfig.module.rules = [
```
skjnldsv, line 107: just push the rules edit please (scss options)
https://github.com/nextcloud/webpack-vue-config#replaceedit-existing-rule

webpack.js
```diff
@@ -0,0 +1,158 @@
...
+				// Singular
+				translations[json.translations[''][key].msgid] = json.translations[''][key].msgstr[0]
+			}
+		}
+
+		return {
+			locale,
+			translations,
+		}
+	})
```
skjnldsv, line 55: Would be nice to put this into an external script file :)

skjnldsv:
Great step!
But I recently started thinking we should most likely step away from webpack and move to rollup for simpler configurations. Rollup is quite awesome for libs.
Then it would allow us to move the styleguide to its original configs and separate both processes (making sure the styleguide is importing the components the same way any other app would do) :)

webpack.js
raimund-schluessler: Removed it.

webpack.js
raimund-schluessler: Removed it.

webpack.js
raimund-schluessler: It's located in `translations.js` now.

raimund-schluessler:
@skjnldsv I fixed all issues mentioned.

> But I recently started thinking we should most likely step away from webpack and move to rollup for simpler configurations. Rollup is quite awesome for libs.
> Then it would allow us to move the styleguide to its original configs and separate both processes (making sure the styleguide is importing the components the same way any other app would do) :)

Fine with me, but I think that's something for another PR. 😉 And I have to admit I have zero experience with rollup, so I cannot really judge if it's a good idea or not.

juliushaertl:
> Edit: Will we have a common rollup configuration? I just saw nextcloud-vue-dashboard already uses rollup? @juliushaertl

Would be a good idea once we've sorted it out in nextclod-vue. I already looked into moving to rollup in the past which was mainly blocked by the iconfont generator which I couldn't get done via rollup, but I just noticed we don't do that anymore https://github.com/nextcloud/nextcloud-vue/commit/47fb0a17eb2e00c47d7d21eabba1ee326c3f5f4a so this might be worth another try, but maybe we can move that to a separate issue indeed. This would help a lot to reduce the size as the library internal modules could then also be treeshaken.

translations.js
```diff
@@ -0,0 +1,40 @@
+const fs = require('fs')
```
juliushaertl, line 1: I'd prefer to move this to `build/`

juliushaertl (commented):
Small nitpick, looks good otherwise :+1:

translations.js
raimund-schluessler: Makes sense, I moved the file.

marcoambrosini (approved):

skjnldsv (approved):

raimund-schluessler:
I don't have npm at hand right now, so if someone wants to rebase... 😉

skjnldsv:
> I don't have npm at hand right now, so if someone wants to rebase...

```
npm WARN @nextcloud/eslint-config@5.1.0 requires a peer of webpack@^5.4.0 but none is installed. You must install peer dependencies yourself.
npm WARN @nextcloud/webpack-vue-config@4.0.3 requires a peer of css-loader@^4.3.0 but none is installed. You must install peer dependencies yourself.
npm WARN @nextcloud/webpack-vue-config@4.0.3 requires a peer of sass@^1.32.8 but none is installed. You must install peer dependencies yourself.
npm WARN @nextcloud/webpack-vue-config@4.0.3 requires a peer of webpack@^5.4.0 but none is installed. You must install peer dependencies yourself.
npm WARN node-polyfill-webpack-plugin@1.1.3 requires a peer of webpack@>=5 but none is installed. You must install peer dependencies yourself.

```

skjnldsv:
we cannot switch to webpack 5 because of https://github.com/nextcloud/nextcloud-vue/pull/1781#issuecomment-805249307

skjnldsv (changes requested):
See comments

raimund-schluessler:
> 
> 
> we cannot switch to webpack 5 because of [#1781 (comment)](https://github.com/nextcloud/nextcloud-vue/pull/1781#issuecomment-805249307)

I didn't update webpack to 5, it is still at 4.42.1 in this PR, see https://github.com/nextcloud/nextcloud-vue/blob/fix/common-webpack-config/package.json#L128. Maybe something is off with your node or npm?

raimund-schluessler:
> 
> 
> > we cannot switch to webpack 5 because of [#1781 (comment)](https://github.com/nextcloud/nextcloud-vue/pull/1781#issuecomment-805249307)
> 
> I didn't update webpack to 5, it is still at 4.42.1 in this PR, see https://github.com/nextcloud/nextcloud-vue/blob/fix/common-webpack-config/package.json#L128. Maybe something is off with your node or npm?

Sorry, I might have gotten it wrong. Do `@nextcloud/eslint-config@5.1.0` and `@nextcloud/webpack-vue-config@4.0.3` require webpack@5? Should I downgrade these packages then?

raimund-schluessler:
> 
> 
> > > we cannot switch to webpack 5 because of [#1781 (comment)](https://github.com/nextcloud/nextcloud-vue/pull/1781#issuecomment-805249307)
> > 
> > 
> > I didn't update webpack to 5, it is still at 4.42.1 in this PR, see https://github.com/nextcloud/nextcloud-vue/blob/fix/common-webpack-config/package.json#L128. Maybe something is off with your node or npm?
> 
> Sorry, I might have gotten it wrong. Do `@nextcloud/eslint-config@5.1.0` and `@nextcloud/webpack-vue-config@4.0.3` require webpack@5? Should I downgrade these packages then?

I just checked the release notes, and I guess this is exactly the case. I will take care of this in the evening and downgrade the affected packages. https://github.com/nextcloud/nextcloud-vue/pull/1781#issuecomment-805249307 is unfortunately still not solved, I just checked that yesterday.

skjnldsv:
> Do `@nextcloud/eslint-config@5.1.0` and `@nextcloud/webpack-vue-config@4.0.3` require webpack@5

Yes :)

skjnldsv:
If you downgrade `@nextcloud/webpack-vue-config` you'll lose the /rule hacks. I would just try to move the styleguide config to a dedicated build process, instead of using our own webpack? If possible?

I think splitting the two woul dhelp. But I don't know if styleguidist can work by just importing an existing built lib?

This is too much of a blocker. https://github.com/vue-styleguidist/vue-styleguidist/tree/dev/examples

raimund-schluessler:
> If you downgrade `@nextcloud/webpack-vue-config` you'll lose the /rule hacks. I would just try to move the styleguide config to a dedicated build process, instead of using our own webpack? If possible?
> 
> 
> 
> I think splitting the two woul dhelp. But I don't know if styleguidist can work by just importing an existing built lib?
> 
> 
> 
> This is too much of a blocker. https://github.com/vue-styleguidist/vue-styleguidist/tree/dev/examples

This sounds complicated. Maybe we just wait until #1781 is resolved 🙈

raimund-schluessler:
@skjnldsv The weird thing is that I don't get these warning when doing npm install and the build and all checks also work without problems. Do you know why that's the case?

skjnldsv:
> The weird thing is that I don't get these warnings when running `npm install`

Npm 6 or 7?
Currently this requires npm 6.

skjnldsv:
image

raimund-schluessler:
> > The weird thing is that I don't get these warnings when running `npm install`
> 
> 
> 
> Npm 6 or 7?
> 
> Currently this requires npm 6.

I ran it with npm 6. Got no warnings regarding webpack 5 🤷‍♀️

skjnldsv:
Stylelint covered by https://github.com/nextcloud/nextcloud-vue/pull/2143

PR #2048, opened 2021-06-22
Fix AppContent mobile display

skjnldsv:
Initial somehow fallback to `inline`, For proper sizing and people using virtual scrollers, let's go back to traditional `block`
Also, because this is a high-level scoped selector, !important is not needed and will allow apps to override this if needed :shrug: 

Fixes contacts + circles

### DIFF
--- a/src/components/AppContent/AppContent.vue
+++ b/src/components/AppContent/AppContent.vue
@@ -332,18 +332,18 @@ export default {
 .app-content-wrapper--mobile {
 	&.app-content-wrapper--show-list ::v-deep {
 		.app-content-list {
-			display: initial !important;
+			display: block;
 		}
 		.app-content-details {
-			display: none !important;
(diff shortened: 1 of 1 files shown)


src/components/AppContent/AppContent.vue
```diff
@@ -332,18 +332,18 @@ export default {
...
 		}
 	}
 	&.app-content-wrapper--show-details ::v-deep {
 		.app-content-list {
-			display: none !important;
+			display: none;
 		}
 		.app-content-details {
-			display: initial !important;
+			display: block;
```
skjnldsv, line 346: Btw, we need this because stupid server have some conflicting rules on `.app-content-list + .app-content-details`

ChristophWurst (approved):

GretaD (approved):

PR #2090, opened 2021-07-07
Replace node-sass with sass, take II

ChristophWurst:
Like https://github.com/nextcloud/nextcloud-vue/pull/2005 but with latest deps. :popcorn:


package.json
```diff
@@ -54,12 +54,12 @@
 		"regenerator-runtime": "^0.13.5",
 		"splitpanes": "^2.3.6",
 		"string-length": "^5.0.0",
-		"striptags": "^3.1.1",
+		"striptags": "^3.2.0",
```
ChristophWurst, line 57: well, `npm update` was maybe a tad too much. but let's see if this gives us a working build

ChristophWurst:
stale

PR #2095, opened 2021-07-08
Add timezone picker

ChristophWurst:
Fixes https://github.com/nextcloud/nextcloud-vue/issues/2094

Bildschirmfoto von 2021-07-09 15-24-48


package.json
```diff
@@ -43,7 +43,9 @@
 		"@nextcloud/dialogs": "^3.0.0",
 		"@nextcloud/event-bus": "^2.0.0",
 		"@nextcloud/l10n": "^1.2.3",
+		"@nextcloud/logger": "^2.0.0",
 		"@nextcloud/router": "^2.0.0",
+		"calendar-js": "git+https://github.com/nextcloud/calendar-js.git",
```
ChristophWurst, line 48: this dependency is not ideal but there is only one module that is used in here so I *hope* it won't bloat our bundle sizes

ChristophWurst:
I would say we should also export this picker as component for apps. It's a sub component for the timezone aware date picker but in the calendar settings (and possibly one day in server) we need a generic timezone selection and then we will use this component without the wrapping date picker.

ChristophWurst:
This should be ready for review. Let me know if the structure is ok. The static timezones file is a resource, hence I moved it out of src as it was done in Calendar

marcoambrosini (approved):
Very cool, the only thing I would change is making it span the whole width of the container to avoid jumps in width and to let developers adapt it to their UI
EDIT: nevermind the comment, I'm just realizing that this would be a multiselect fix!

GretaD (approved):

src/components/TimezonePicker/TimezonePicker.vue
```diff
@@ -0,0 +1,124 @@
...
+			type: String,
+			default: 'floating',
+		},
+	},
+	computed: {
+		placeholder() {
+			return t('Type to search timezone')
+		},
+		selectedTimezone: {
+			get() {
```
skjnldsv, line 93: Not really needed :)

skjnldsv (approved):

Pytal (approved):

PR #2100, opened 2021-07-09
Optionally make the DatetimePicker timezone-aware

ChristophWurst:
- [x] Requires https://github.com/nextcloud/nextcloud-vue/pull/2095

Fixes https://github.com/nextcloud/nextcloud-vue/issues/2096

Bildschirmfoto von 2021-07-15 16-47-25

### DIFF
--- a/l10n/messages.pot
+++ b/l10n/messages.pot
@@ -74,6 +74,9 @@ msgstr ""
 msgid "Pick an emoji"
 msgstr ""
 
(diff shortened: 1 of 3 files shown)


Pytal (approved):

eneiluj (approved):
Nice! Works like a charm.

It would be nice to make the timezone value more visible, like show the popover on hover. Anyway it's still possible to display the timezone in the formatted date.

PR #2122, opened 2021-07-19
Set default slideshowdelay to 5000 ms

szaimen:
3000 ms is a bit short for most images, imho. Hence I propose increasing it to 5000 ms.


jancborchardt (approved):
Yep, seems good! :)

artonge (approved):

PR #2126, opened 2021-07-20
standardize modal sizes

szaimen:
Currently, there are many open points regarding modal sizes that each developer needs to decide and implement. This leads to unecessary code duplication, maintenance effort and inconsistent UI because each developer implements the sizing differently. Hence here is my proposal to standardize the modal sizes completely.

It also includes always using full screen modal on mobile which is the superior experience on mobile, imho.

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -295,7 +295,7 @@ export default {
 			type: String,
 			default: 'normal',
 			validator: size => {
-				return ['normal', 'large', 'full'].indexOf(size) !== -1
+				return ['small', 'medium', 'normal', 'large', 'full'].indexOf(size) !== -1
 			},
 		},
 		/**
@@ -673,9 +673,10 @@ $header-size: 50px;
 		display: flex !important;
 		align-items: center;
 		justify-content: center;
-		width: 15%;
-		min-width: 60px;
-		height: 100%;
+		width: 8%;
+		min-width: $clickable-area;
+		height: 35vw;
+		position: absolute;
 		transition: opacity 250ms,
 			visibility 250ms;
 
@@ -687,6 +688,12 @@ $header-size: 50px;
 		&.invisible[style*='display: none'] {
 			visibility: hidden;
 		}
+	}
+	.prev {
+		left: 0;
+	}
+	.next {
+		right: 0;
 	}
 
 	// buttons/icons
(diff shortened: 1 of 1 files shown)


src/components/Modal/Modal.vue
```diff
@@ -703,23 +703,29 @@ $header-size: 50px;
 	/* Content */
 	.modal-container {
 		display: block;
-		overflow: hidden;
+		overflow: auto; // avoids unecessary hacks if the content should be bigger than the modal
```
skjnldsv, line 706: Breaking change

src/components/Modal/Modal.vue
```diff
@@ -750,6 +758,17 @@ $header-size: 50px;
 			height: 35vw;
 		}
 	}
+
+	// Make modal always full screen on mobile independent from modal size
+	@media only screen and (max-width: $breakpoint-mobile/2) {
```
skjnldsv, line 763: Why breakpoint-mobile/2 ?
We never use that anywhere else on Nextcloud

src/components/Modal/Modal.vue
szaimen: yes, I know, hence I've added the breaking label. But it is necessary to make this work reliably.

src/components/Modal/Modal.vue
szaimen: Thats incorrect because there is actually already one place, in the same file: https://github.com/nextcloud/nextcloud-vue/blob/69ca1d83ef367ef985d039e3d73843c165bc132a/src/components/Modal/Modal.vue#L578

The problem is that breakpoint-mobile=1024 which is far too much but around 500px is the perfect breakpoint for making all modals full screen -> hence i've done it this way.

szaimen:
Thanks for the review @skjnldsv! 
I've just reworked some small things and added two more sizes which could potentially be useful.

src/components/Modal/Modal.vue
Pytal: If we're following SemVer then this would consitute a new major :)

szaimen: Jep. Would it be 4.1.0 or 5.0.0 ?

Pytal: 5.0.0 :)

szaimen: Added milestone 5.0.0 :)

src/components/Modal/Modal.vue
marcoambrosini: Should we just reduce the breakpoint then?

szaimen: Honestly fine by me but would be another breaking change ;)
@skjnldsv should decide if it is wort the trouble!
(Would also need to be changed in the ismobile mixin and other places where breakpoint-mobile is used)

szaimen:
I would like to get some more feedback on this from developers that already use modals in their apps.
Is this something we can do or it is just too much effort to remove all hacks that have been necesarry until this PR which will break implementation of Modal sizes?

jancborchardt:
Less decision sounds good. :) Can you add some screenshots of different examples of modals for reference?

szaimen:
Jep, Here are some examples of the different sizes :)

| Size | Desktop |
|---|---|
| Small | image |
| Medium | image |
| Normal | image |
| Large | image |
| Full | image |

src/components/Modal/Modal.vue
skjnldsv: let's leave like that, I should have commented my code, sorry for that :)
Nice catch @szaimen !

skjnldsv (approved):

Pytal (approved):

skjnldsv:
stable4 branch created https://github.com/nextcloud/nextcloud-vue/tree/stable4

jancborchardt:
@szaimen @skjnldsv @marcoambrosini what do you think about folding the "Medium" and "Normal" into one size? I mean, the name is basically the same, so quite confusing. And then we only have small, medium, large and full – more obvious. :) The size could be something inbetween Medium and Normal.

szaimen:
@jancborchardt there was already a change in the modal sizes here: https://github.com/nextcloud/nextcloud-vue/pull/2170#issuecomment-900687792
So there you can see the current catalog of sizes.

Apqrt from that, merging medium and normal into one size in between is fine by me.

jancborchardt:
@szaimen there also doesn’t really seem to be a difference between "prompt" and "small" except the height (which could simply adjust based on content, with a max-height)?

src/components/Modal/Modal.vue
```diff
@@ -703,51 +710,63 @@ $header-size: 50px;
...
 			max-height: 90%;
+			height: 700px;
+		}
+	}
+	&--full {
+		.modal-container {
+			width: 100%;
+			height: calc(100% - $header-size);
+			position: absolute;
+			top: $header-size;
```
skjnldsv, line 755: We could have gone for 
```suggestion
			height: auto;
			position: absolute;
			top: $header-size;
			bottom: 0;
```

skjnldsv: Which imho is better than calc with `$header-size` :)
@szaimen

szaimen: Would the result be the same?

PR #2181, opened 2021-08-20
Fix backwards navigation for listitem component

### LINKED ISSUE #2175: Add support for tabbing backwards on ListItem component
I don't know if it fits here, but I just noticed that tabbing through the messages only works forwards. Tabbing backwards (with shift+tab) does not work.

_Originally posted by @raimund-schluessler in https://github.com/nextcloud/mail/issues/5281#issuecomment-901090147_

marcoambrosini:
Fixes #2175


raimund-schluessler (approved):
Works (tested in the docs).

juliushaertl (approved):

PR #2197, opened 2021-08-31
Improve modal and appsettingsdialog docs

szaimen:
And fix some details of the initial implementation
Follow up to https://github.com/nextcloud/nextcloud-vue/pull/2126

### DIFF
--- a/src/components/AppSettingsDialog/AppSettingsDialog.vue
+++ b/src/components/AppSettingsDialog/AppSettingsDialog.vue
@@ -28,7 +28,7 @@ providing the section's title prop. You can put your settings within each
 <template>
 	<div>
 		<button @click="settingsOpen = true">Show Settings</button>
-		<AppSettingsDialog :open.sync="settingsOpen" >
+		<AppSettingsDialog :open.sync="settingsOpen" showNavigation=true>
 			<AppSettingsSection title="Example title 1">
 				Some example content
 			</AppSettingsSection>
@@ -298,11 +298,13 @@ export default {
 		overflow-y: auto;
 		overflow-x: hidden;
 		padding: 24px;
+		width: 100%;
 	}
 }
 
 .navigation-list {
 	height: 100%;
+	box-sizing: border-box;
 	overflow-y: auto;
 	padding: 12px;
 	&__link {
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -26,15 +26,14 @@
 <template>
 	<div>
 		<button @click="showModal">Show Modal</button>
-		<modal v-if="modal" @close="closeModal">
+		<modal v-if="modal" @close="closeModal" size="prompt">
 			<div class="modal__content">Hello world</div>
 		</modal>
 	</div>
 </template>
 <style scoped>
 	.modal__content {
-		width: 50vw;
-		margin: 10vw 0;
+		margin: 50px;
 		text-align: center;
 	}
 
@@ -291,6 +290,12 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		/**
+		 * Defines the modal size
+		 * Default is 'normal'
+		 * Available are 'prompt', 'small', 'medium', 'normal', 'large' and 'full'
+		 * All sizes except 'prompt' and 'small' change automatically to full-screen on mobile
+		 */
 		size: {
 			type: String,
 			default: 'normal',
@@ -524,7 +529,6 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-$header-size: 50px;
 
 .modal-mask {
 	position: fixed;
@@ -552,7 +556,7 @@ $header-size: 50px;
 	align-items: center;
 	justify-content: center;
 	width: 100%;
-	height: $header-size;
+	height: $header-height;
 	transition: opacity 250ms,
 		visibility 250ms;
 
@@ -594,14 +598,14 @@ $header-size: 50px;
 			align-items: center;
 			justify-content: center;
 			box-sizing: border-box;
-			margin: ($header-size - $clickable-area) / 2;
+			margin: ($header-height - $clickable-area) / 2;
 			padding: 0;
 		}
 
 		.play-pause-icons {
 			position: relative;
-			width: $header-size;
-			height: $header-size;
+			width: $header-height;
+			height: $header-height;
 			margin: 0;
 			padding: 0;
 			cursor: pointer;
@@ -621,14 +625,14 @@ $header-size: 50px;
 				box-sizing: border-box;
 				width: $clickable-area;
 				height: $clickable-area;
-				margin: ($header-size - $clickable-area) / 2;
+				margin: ($header-height - $clickable-area) / 2;
 				cursor: pointer;
 				opacity: $opacity_normal;
 			}
 		}
 
 		.header-actions {
-			margin: ($header-size - $clickable-area) / 2;
+			margin: ($header-height - $clickable-area) / 2;
 			color: white;
 		}
 
@@ -728,20 +732,22 @@ $header-size: 50px;
 	&--small {
 		.modal-container {
 			width: 350px;
+			max-height: 90%;
 			height: 450px;
 		}
 	}
 	&--medium {
 		.modal-container {
 			width: 500px;
+			max-height: 90%;
 			height: 600px;
 		}
 	}
 	&--normal {
 		.modal-container {
-			max-width: 70%;
+			max-width: 90%;
 			width: 700px;
-			max-height: 80%;
+			max-height: 90%;
 			height: 600px;
 		}
 	}
@@ -756,9 +762,9 @@ $header-size: 50px;
 	&--full {
 		.modal-container {
 			width: 100%;
-			height: calc(100% - $header-size);
+			height: calc(100% - var(--header-height));
 			position: absolute;
-			top: $header-size;
+			top: $header-height;
 			border-radius: 0;
 		}
 	}
@@ -769,9 +775,9 @@ $header-size: 50px;
 			max-width: initial;
 			width: 100%;
 			max-height: initial;
-			height: calc(100% - $header-size);
+			height: calc(100% - var(--header-height));
 			position: absolute;
-			top: $header-size;
+			top: $header-height;
 			border-radius: 0;
 		}
 	}


artonge:
Some information about the reasoning behind the numbers would help review :D

szaimen:
> Some information about the reasoning behind the numbers would help review :D

For which specifically or for all changes?

artonge:
> > Some information about the reasoning behind the numbers would help review :D
> 
> For which specifically or for all changes?

I would say most of them

src/components/Modal/Modal.vue
```diff
@@ -769,7 +776,7 @@ $header-size: 50px;
 			max-width: initial;
 			width: 100%;
 			max-height: initial;
-			height: calc(100% - $header-size);
+			height: calc(100% - 50px);
```
artonge, line 779: For example, we dropping `$header-size` ?

src/components/Modal/Modal.vue
```diff
@@ -728,20 +733,22 @@ $header-size: 50px;
...
 		.modal-container {
 			width: 500px;
+			max-height: 90%;
 			height: 600px;
 		}
 	}
 	&--normal {
 		.modal-container {
-			max-width: 70%;
+			max-width: 90%;
```
artonge, line 749: Why changing the size? Only aesthetics?

src/components/AppSettingsDialog/AppSettingsDialog.vue
```diff
@@ -298,11 +298,12 @@ export default {
 		overflow-y: auto;
 		overflow-x: hidden;
 		padding: 24px;
+		width: 100%;
```
szaimen, line 301: This makes sure that only one scrollbar is shown on the right of the modal (would otherwise show a second scrollbar )

src/components/AppSettingsDialog/AppSettingsDialog.vue
```diff
@@ -298,11 +298,12 @@ export default {
...
 	}
 }
 
 .navigation-list {
-	height: 100%;
+	height: calc(100% - 2 * 12px);
```
szaimen, line 306: This makes sure that having many entries doesn't show an additional scrollbar on the right (if it would be then bigger than the modal size)

src/components/Modal/Modal.vue
```diff
@@ -26,15 +26,14 @@
...
+		<modal v-if="modal" @close="closeModal" size="prompt">
 			<div class="modal__content">Hello world</div>
 		</modal>
 	</div>
 </template>
 <style scoped>
 	.modal__content {
-		width: 50vw;
-		margin: 10vw 0;
+		margin: 50px;
```
szaimen, line 36: defining a width is no longer necessary as the modal has a distinct size out of the box

src/components/Modal/Modal.vue
```diff
@@ -291,6 +290,12 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		/**
+		 * Defines the modal size
+		 * Default is 'normal'
+		 * Available are 'prompt', 'small', 'medium', 'normal', 'large' and 'full'
+		 * All sizes except 'prompt' and 'small' change automatically to full-screen on mobile
+		 */
```
szaimen, line 298: this propably doesn't need an explanation ;)

src/components/Modal/Modal.vue
```diff
@@ -26,15 +26,14 @@
 <template>
 	<div>
 		<button @click="showModal">Show Modal</button>
-		<modal v-if="modal" @close="closeModal">
+		<modal v-if="modal" @close="closeModal" size="prompt">
```
szaimen, line 29: just wanted to add a size to the example

src/components/AppSettingsDialog/AppSettingsDialog.vue
```diff
@@ -28,7 +28,7 @@ providing the section's title prop. You can put your settings within each
 <template>
 	<div>
 		<button @click="settingsOpen = true">Show Settings</button>
-		<AppSettingsDialog :open.sync="settingsOpen" >
+		<AppSettingsDialog :open.sync="settingsOpen" showNavigation=true>
```
szaimen, line 31: showing the navigation is more useful for showing an example

src/components/Modal/Modal.vue
```diff
@@ -728,20 +733,22 @@ $header-size: 50px;
 	&--small {
 		.modal-container {
 			width: 350px;
+			max-height: 90%;
```
szaimen, line 736: having a max-heigt is necessary e.g. if you rotate a smartphone that is only 350px wide into landscape mode (which would then be 350px high)

src/components/Modal/Modal.vue
```diff
@@ -728,20 +733,22 @@ $header-size: 50px;
...
 			height: 450px;
 		}
 	}
 	&--medium {
 		.modal-container {
 			width: 500px;
+			max-height: 90%;
```
szaimen, line 743: having a max-heigt is necessary e.g. if you rotate a smartphone that is only 350px wide into landscape mode (which would then be 350px high)

src/components/AppSettingsDialog/AppSettingsDialog.vue
artonge, line 306: Padding compensation? Could `box-sizing: border-box` be useful ?

https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing

src/components/Modal/Modal.vue
szaimen, line 749: This reduces the size difference between the width of the element when going from over the mobile breakpoint to below

src/components/Modal/Modal.vue
```diff
@@ -728,20 +733,22 @@ $header-size: 50px;
...
 			height: 600px;
 		}
 	}
 	&--normal {
 		.modal-container {
-			max-width: 70%;
+			max-width: 90%;
 			width: 700px;
-			max-height: 80%;
+			max-height: 90%;
```
szaimen, line 751: This reduces the size difference between the height of the element when going from over the mobile breakpoint to below

src/components/Modal/Modal.vue
```diff
@@ -756,7 +763,7 @@ $header-size: 50px;
 	&--full {
 		.modal-container {
 			width: 100%;
-			height: calc(100% - $header-size);
+			height: calc(100% - 50px);
```
szaimen, line 766: unfortunatley $header-size doesnt work here, don't know why.

src/components/Modal/Modal.vue
szaimen, line 779: unfortunatley $header-size doesnt work here, don't know why.

src/components/Modal/Modal.vue
szaimen: see https://github.com/nextcloud/nextcloud-vue/pull/2197#discussion_r699447477

src/components/AppSettingsDialog/AppSettingsDialog.vue
szaimen: see https://github.com/nextcloud/nextcloud-vue/pull/2197#discussion_r699448026

src/components/Modal/Modal.vue
szaimen: see https://github.com/nextcloud/nextcloud-vue/pull/2197#discussion_r699448026

src/components/AppSettingsDialog/AppSettingsDialog.vue
szaimen: jep, padding compensation. I'll try if your suggestion works.

szaimen: works, fixed

szaimen: done it differently now. See https://github.com/nextcloud/nextcloud-vue/pull/2197#discussion_r699447146

src/components/Modal/Modal.vue
```diff
@@ -26,15 +26,14 @@
 <template>
 	<div>
 		<button @click="showModal">Show Modal</button>
-		<modal v-if="modal" @close="closeModal">
+		<modal v-if="modal" @close="closeModal" size="prompt">
 			<div class="modal__content">Hello world</div>
 		</modal>
 	</div>
 </template>
 <style scoped>
```
artonge, line 34: So, I am really not sure about this one, but maybe setting the lang to `scss` can help to use variables. I am not sure where the variable is defined in the first place.

```suggestion
<style scoped lang="scss">
```

src/components/AppSettingsDialog/AppSettingsDialog.vue
artonge: Couldn't `max-width: 100%` do the trick ?

src/components/Modal/Modal.vue
szaimen: No, that doesn't seem to work unfortunatley. 
$header-size is set in the scss code of the modal here:
https://github.com/nextcloud/nextcloud-vue/blob/ab33f33ed82a71a72e53c9bbb83bcc27e7e9b6c2/src/components/Modal/Modal.vue#L532

src/components/AppSettingsDialog/AppSettingsDialog.vue
szaimen: good idea, will try that

szaimen:
> Couldn't max-width: 100% do the trick ?
That doesn't work as it doesn't expand to the width of the  element otherwise and would produce an additional scroller...

src/components/AppSettingsDialog/AppSettingsDialog.vue
szaimen: That doesn't work as it doesn't expand to the width of the element otherwise and would produce an additional scroller...

src/components/Modal/Modal.vue
artonge: Maybe @skjnldsv can give use some insight on css/scss good practices

szaimen: The problem is that
```
height: calc(100% - $header-size);
```
Doesnt work...

Pytal: Since we have https://github.com/nextcloud/nextcloud-vue/blob/c5f9c2e57ce158546dc49cf5ef92394cd0ffaa32/webpack.common.js#L113 which imports the SCSS variables into all files I think moving https://github.com/nextcloud/nextcloud-vue/blob/c5f9c2e57ce158546dc49cf5ef92394cd0ffaa32/src/components/Modal/Modal.vue#L527
into https://github.com/nextcloud/nextcloud-vue/blob/c5f9c2e57ce158546dc49cf5ef92394cd0ffaa32/src/assets/variables.scss would make it work :)

skjnldsv: https://github.com/nextcloud/server/blob/4f90766ba314171bbfc78d1e988307c50633e7f3/core/css/css-variables.scss#L59

szaimen: Done, thanks for the pointer! :)

szaimen: Shall I also change this everywhere where $header-size is currently used? :)

szaimen: @skjnldsv $header-size is only used in the modal component so it can be easily changed.

skjnldsv: Yes, change it here then :)

artonge (approved):

skjnldsv (approved):

src/components/Modal/Modal.vue
szaimen: Not done yet

szaimen: should be done now

szaimen:
should be ready now :)

PR #2210, opened 2021-09-06
Correctly align the settings label

### LINKED ISSUE #2138: Settings label not aligned to app-navigation-entry
Screenshot 2021-07-23 at 09-05-36 Aufgaben - Nextcloud

raimund-schluessler:
After:
Screenshot 2021-09-06 at 21-23-41 Tasks - Nextcloud

Before:
Screenshot 2021-09-06 at 21-24-43 Tasks - Nextcloud

Closes #2138

### DIFF
--- a/src/components/AppNavigationSettings/AppNavigationSettings.vue
+++ b/src/components/AppNavigationSettings/AppNavigationSettings.vue
@@ -79,6 +79,11 @@ export default {
 }
 </script>
 <style lang="scss" scoped>
+
+#app-settings-header .settings-button {
+	padding-left: 50px;
+}
+
 #app-settings-content {
 	display: block;
 	padding: 10px;


src/components/AppNavigationSettings/AppNavigationSettings.vue
```diff
@@ -79,6 +79,11 @@ export default {
 }
 </script>
 <style lang="scss" scoped>
+
+#app-settings-header .settings-button {
+	padding-left: 50px;
+}
+
```
skjnldsv, line 86: I would argue the opposite, should be fixed in files, no?
What is the AppNavigationItem using?

raimund-schluessler: This is not Files, this is Tasks and Calendar and literally every other app which uses AppNavigationItem. These 6 px here:
https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/AppNavigationItem/AppNavigationItem.vue#L520
add to the 44 px icon width, which make it 50 px in total.

raimund-schluessler: This was added in this commit https://github.com/nextcloud/nextcloud-vue/commit/5e7e3950d30e7188c55a3f414767fe16bdccd94d#

szaimen (commented):
Maybe add a `opacity: 1 !important;` so that the opacity is the same like for the trashbin?

Also, it currently looks like this on c.nc.c 
image
so not sure if this really needs fixing (or did you overwrite the stile already in the tasks app?

raimund-schluessler:
> 
> 
> Maybe add a `opacity: 1 !important;` so that the opacity is the same like for the trashbin?

This style is coming from server, I didn't want to mess with it, especially not in this tiny PR.

> Also, it currently looks like this on c.nc.c
> image
> so not sure if this really needs fixing (or did you overwrite the stile already in the tasks app?

I overwrote this style in the Tasks app (see here https://github.com/nextcloud/tasks/blob/005cd27fda7ce5f53d65c469bb254458fe3abe0d/src/views/AppNavigation.vue#L456-L458) because this misalignment annoyed me 🙈 This PR here upstreams this fix.

raimund-schluessler:
Superseded by #2211.

Issue #2199, opened 2021-09-01
Use material design icons everywhere


raimund-schluessler:
- [x] AppSidebar close icon
- [x] Settings icon

PR #2211, opened 2021-09-06
Use material design icon for settings icon

### LINKED ISSUE #2138: Settings label not aligned to app-navigation-entry

raimund-schluessler:
This PR replaces the current settings icon with the material design equivalent and also fixes the settings label alignment to the `AppNavigationItem` component. Supersedes #2210, closes #2138 and fixes the second part of #2199.

After:
Screenshot 2021-09-06 at 22-30-42 Tasks - Nextcloud

Before:
Screenshot 2021-09-06 at 21-24-43 Tasks - Nextcloud

### DIFF
--- a/src/components/AppNavigationSettings/AppNavigationSettings.vue
+++ b/src/components/AppNavigationSettings/AppNavigationSettings.vue
@@ -24,16 +24,17 @@
 	<div id="app-settings"
 		v-click-outside="clickOutsideConfig"
 		:class="{ open }">
-		<div id="app-settings-header">
+		<div id="app-settings__header">
 			<button
 				class="settings-button"
 				type="button"
 				@click="toggleMenu">
-				{{ title }}
+				<Cog class="settings-button__icon" :size="24" decorative />
+				<span class="settings-button__label">{{ title }}</span>
 			</button>
 		</div>
 		<transition name="slide-up">
-			<div v-show="open" id="app-settings-content">
+			<div v-show="open" id="app-settings__content">
 				<slot />
 			</div>
 		</transition>
@@ -45,9 +46,14 @@ import { directive as ClickOutside } from 'v-click-outside'
 import { t } from '../../l10n'
 import { excludeClickOutsideClasses } from '../../mixins'
 
+import Cog from 'vue-material-design-icons/Cog.vue'
+
 export default {
 	directives: {
 		ClickOutside,
+	},
+	components: {
+		Cog,
 	},
 	mixins: [
 		excludeClickOutsideClasses,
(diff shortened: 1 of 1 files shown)


szaimen (approved):
can you add `opacity: 1 !important;` so that the stile fits the trashbin?

raimund-schluessler:
> 
> 
> can you add `opacity: 1 !important;` so that the stile fits the trashbin?

This is a server style which I just copied. I guess the lower opacity is intendend.

szaimen:
> I guess the lower opacity is intendend.

jep, but only for the files app (because the other labels have the same opacity there).

raimund-schluessler:
> 
> 
> > I guess the lower opacity is intendend.
> 
> jep, but only for the files app (because the other labels have the same opacity there).

So it was wrong all the time in every other app, because the Files app messed with the opacity? 😄
I guess we can fix it then.

raimund-schluessler:
> 
> 
> can you add `opacity: 1 !important;` so that the stile fits the trashbin?

Fixed. I updated the screenshot in the first post.

src/components/AppNavigationSettings/AppNavigationSettings.vue
```diff
@@ -26,10 +26,11 @@
 		:class="{ open }">
 		<div id="app-settings-header">
 			<button
-				class="settings-button"
+				class="settings-button-vue"
```
skjnldsv, line 29: We could take that opportunity to drop the server styling and move to proper class BEM.

```suggestion
		<div class="app-settings">
			<button
				class="app-settings__button"
```

raimund-schluessler: Done.

skjnldsv: I meant the entire app-settings :)
But that could be another PR

skjnldsv (approved):

src/components/AppNavigationSettings/AppNavigationSettings.vue
raimund-schluessler: > 
> 
> I meant the entire app-settings :)
> But that could be another PR

I guess, I didn't understand what you mean. 🙈

szaimen:
> > can you add `opacity: 1 !important;` so that the stile fits the trashbin?
> 
> Fixed. I updated the screenshot in the first post.

How did you fix it? I don't see `opacity: 1 !important;` in the code...

raimund-schluessler:
> 
> 
> > > can you add `opacity: 1 !important;` so that the stile fits the trashbin?
> > 
> > 
> > Fixed. I updated the screenshot in the first post.
> 
> How did you fix it? I don't see `opacity: 1 !important;` in the code...

`opacity: 0.8` came from the server style, which does not apply anymore because the selectors are different now. So no need to overwrite it.

PR #2225, opened 2021-09-10
Do not duplicate character in Highlight directive, only highlight first match occurence

### LINKED ISSUE #2191: Highlight duplicates characters
When using the following example, the highlight component would actually duplicate the matched search string and printing `oooo` instead of `ooo`:

```
<Highlight text="ooo" search="oo" />
```

image

Seems like quite an edge case where the range would match both the first and last two occurences of "oo" in https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/Highlight/Highlight.vue#L122

raimund-schluessler:
This ensures that we do not duplicate characters in the Highlight component, even if the user supplies overlapping ranges. And that we only match the first occurence of overlapping matches when the user supplies a string to search for.

I also added some unit tests to check for the correct functioning.

After:
after

Before:
before

Closes #2191.


juliushaertl (approved):
Thanks for the fix and tests :heart:

GretaD (approved):

PR #2229, opened 2021-09-10
Don't rely on Vue.prototype.t and properly extract all strings

### LINKED ISSUE #586: Components rely on translate beeing bound to Vue.prototype.t
With https://github.com/nextcloud/tasks/pull/586 and https://github.com/nextcloud/tasks/pull/601 I realised, that the vue-components rely on the app to bind `t` to `Vue.prototype.t` (and probably `n` to be bound to `Vue.prototype.n`).

First, as stated https://vuejs.org/v2/cookbook/adding-instance-properties.html#The-Importance-of-Scoping-Instance-Properties, instance properties should be scoped (so start with a `$`).

Second, I didn't find this in the Readme of the components, so this should either be added, or the vue-components should bind `t` to the correct prototype themself (don't know, whether this is possible).

Changing the instance properties to be scoped would be a breaking change in case the components cannot bind `t` themself.

raimund-schluessler:
This removes the necessity to bind `window.t` to `Vue.prototype.t`. This makes using the components a bit less error prone on first usage and cleans up the Vue instance. The Netlify build still looks fine and e.g. the Calendar app works well.

We should furthermore not use `this.t(...)` (i.e. the `l10n` mixin) as it breaks the l10n extraction. The mixin is fine if we need `t` for the template though.

Closes #586.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -37,8 +37,6 @@ import Avatar from '@nextcloud/vue/dist/Components/Avatar'
 Some components require Nextcloud functionality to work, so it is currently recommended to extend Vue with the following:
 
 ```js static
-Vue.prototype.t = window.t
-Vue.prototype.n = window.n
 Vue.prototype.OC = window.OC
 Vue.prototype.OCA = window.OCA
 ```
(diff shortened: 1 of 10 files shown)


ChristophWurst (approved):

juliushaertl (approved):
:tada:

PR #2234, opened 2021-09-13
Fix UserBubble being sometimes empty

### LINKED ISSUE #2200: UserBubble is empty in v4.1.0
UserBubble is now rendered as an empty `div` (the wrapper):

```
<div data-v-5c0e21af="" data-v-ae66bf1a="" trigger="hover focus" class="user-bubble__wrapper"> </div>
```
which means the Popover `trigger` slot is not rendered anymore.

@raimund-schluessler This bug was introduced in #2160 (8228b6e7eadf2b707eb2fedac1c1e02882ff90ba). I tried to revert `src/components/UserBubble/UserBubble.vue` like it was before this commit and it works fine.

I didn't investigate further what's problematic in this commit. Maybe this syntax change is not strictly equivalent as before and some attributes are not visible inside the `<template>` tag.

eneiluj:
fixes #2200
This reverts changes from 8228b6e (#2160) on UserBubble.vue.

### DIFF
--- a/src/components/UserBubble/UserBubble.vue
+++ b/src/components/UserBubble/UserBubble.vue
@@ -75,35 +75,34 @@ This component has the following slot:
 		class="user-bubble__wrapper"
 		@update:open="onOpenChange">
 		<!-- Main userbubble structure -->
-		<template #trigger>
-			<div v-bind="isLinkComponent"
-				class="user-bubble__content"
-				:style="styles.content"
-				:class="primary ? 'user-bubble__content--primary' : ''"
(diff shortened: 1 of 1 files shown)


raimund-schluessler (approved):
Fixes it on netlify. 👍

raimund-schluessler:
I created a PR #2236 which now fixes the underlying problem instead of partially reverting the changes from #2160. I would propose to check and merge this one instead.

raimund-schluessler:
Superseded by #2236.

PR #2236, opened 2021-09-13
Fix empty user bubble

### LINKED ISSUE #2200: UserBubble is empty in v4.1.0

raimund-schluessler:
Fixes #2200. Supersedes #2234.

The problem was that we use a `div` in case the default slot is empty and then `<template #trigger>` is simply not rendered anymore.
I introduced a component which just renders a `div` as a wrapper for the #trigger slot in case the default slot is empty.

UserBubble without default slot on netlify:
Screenshot 2021-09-13 at 22-25-02 Nextcloud Vue Style Guide

### DIFF
--- a/src/components/UserBubble/UserBubble.vue
+++ b/src/components/UserBubble/UserBubble.vue
@@ -112,6 +112,7 @@ This component has the following slot:
 
 <script>
 import Popover from '../Popover'
+import UserBubbleDiv from './UserBubbleDiv'
 import Avatar from '../Avatar'
 
 export default {
@@ -119,6 +120,7 @@ export default {
 	components: {
 		Popover,
 		Avatar,
+		UserBubbleDiv,
 	},
 	props: {
 		/**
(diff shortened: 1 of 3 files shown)


nickvergessen (approved):
Fixes the issue

marcoambrosini (approved):

PR #2252, opened 2021-09-16
Stop filtering Actions in $slots

skjnldsv:
https://github.com/nextcloud/server/pull/28835


PVince81 (approved):
:+1:

would be good to add a comment + raise/refer a ticket to make it clear that this is a temporary workaround

eneiluj (approved):

PR #2273, opened 2021-09-21
add custom modal size option

szaimen:

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -291,16 +291,18 @@ export default {
 			default: false,
 		},
 		/**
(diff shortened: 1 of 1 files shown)


jancborchardt:
What is the use case for this? I would advise against this, also since mentioned in https://github.com/nextcloud/nextcloud-vue/pull/2126#issuecomment-922919283 there are already plenty (too many) modal sizes.

szaimen:
Fine by me then simply merging normal and medium into a size in between and small and prompt into a size without a height, correct?

szaimen:
done that in https://github.com/nextcloud/nextcloud-vue/pull/2274

PR #2208, opened 2021-09-06
Add outlined counter

marcoambrosini:
Required by talk in order to separate team mentions from direct mentions

Screenshot 2021-09-06 at 16 04 24

Signed-off-by: marco <marcoambrosini@pm.me>

PR #2286, opened 2021-09-28
Add backwards compatibility to counterbubble

marcoambrosini:


src/components/CounterBubble/CounterBubble.vue
```diff
@@ -62,10 +62,22 @@ export default {
 				return ['highlighted', 'outlined', ''].indexOf(value) !== -1
 			},
 		},
+
+		/**
+		 * DEPRECATED: use the prop "type" instead.
+		 * Toggles the highlighted state of the counter.
+		 */
+		counterHighlighted: {
```
raimund-schluessler, line 70: The prop was just called `highlighted` in the `CounterBubble` component before https://github.com/nextcloud/nextcloud-vue/pull/2208 (if I am not wrong)
```suggestion
		highlighted: {
```

src/components/CounterBubble/CounterBubble.vue
```diff
@@ -62,10 +62,22 @@ export default {
...
+			type: Boolean,
+			default: false,
+		},
 	},
 
 	computed: {
 		counterClassObject() {
+			if (this.counterHighlighted) {
```
raimund-schluessler, line 78:
```suggestion
			if (this.highlighted) {
```

raimund-schluessler (changes requested):
To really make it a non-breaking change, we also need the `counter-highlighted` prop back for the `ListItem` component.

And I would propose to backport #2208 plus this fix to stable4 only but not to master. Master can have breaking changes and become 5.0.0, no?

nickvergessen:
> And I would propose to backport #2208 plus this fix to stable4 only but not to master. Master can have breaking changes and become 5.0.0, no?

Sounds good as well

marcoambrosini:
Review addressed @raimund-schluessler , thank you :)

raimund-schluessler:
@marcoambrosini Looks good now. But could you add this commit (65ae85b2078502b51b0ae0545df7a0be3808feb2) directly to https://github.com/nextcloud/nextcloud-vue/pull/2287? Then we wouldn't need this PR here and master wouldn't have the deprecations.

marcoambrosini:
done :)

PR #2305, opened 2021-10-06
Feature/noid/button2 slots

raimund-schluessler:
Just for testing, not intended to supersede #1808.


marcoambrosini (commented):
the computed prop `hasText` would need to be updated too but other than that it looks good, I'll add this commit

PR #2323, opened 2021-10-18
Reduce size of status icons in Avatar component

CarlSchwan:
Now use 40% with of parent a max size of 18px. For talk participant bar
the size remains unchanged still 18px, but for the new avatar component
in the navbar the size is now 13px.

Before: 
image

After:
image

Fix https://github.com/nextcloud/server/issues/29353

Signed-off-by: Carl Schwan <carl@carlschwan.eu>

### DIFF
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -669,8 +669,10 @@ export default {
 		position: absolute;
 		right: -4px;
 		bottom: -4px;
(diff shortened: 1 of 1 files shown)


CarlSchwan:
> I think we could go even lower than that, like 30%, so that we have a proportion similar to what's in talk's participants bar

image

30% is to small in my opinion

PR #1939, opened 2021-05-07
Allow material design icons in AppSidebarTab

raimund-schluessler:
This PR implements material design icons in the AppSidebarTabs header. The icon is not gray anymore when focused, which I think makes sense, as the text is black as well.

Before
Screenshot_2021-05-08 Nextcloud Vue Style Guide(1)

After (sharing icon is provided by string):
Screenshot_2021-05-08 Nextcloud Vue Style Guide

Closes #1859.

PR #2374, opened 2021-11-29
Change background style target for tab icons

artonge:
The style was applied to the containing element instead of the icon element.

Don't know why the `background-position: center 8px;` was needed, so I removed it. It was also shifting icons toward the bottom when applied to the icon.

Tested in `styleguide`.

Fix: https://github.com/nextcloud/activity/issues/669
Problem introduced here: https://github.com/nextcloud/nextcloud-vue/pull/1939

### DIFF
--- a/src/components/AppSidebar/AppSidebarTabs.vue
+++ b/src/components/AppSidebar/AppSidebarTabs.vue
@@ -345,13 +345,12 @@ export default {
 		height: 25px;
 		transition: opacity var(--animation-quick);
 		opacity: $opacity_normal;
-		background-position: center 8px;
-		background-size: 16px;
 
 		& > span {
 			display: flex;
 			align-items: center;
 			justify-content: center;
+			background-size: 16px;
 		}
 	}
 


skjnldsv:
> Don't know why the `background-position: center 8px;` was needed,

To support the legacy icon methods iirc. But that changed when we added compatibility for the MaterialDesignIcons

raimund-schluessler:
I am a bit puzzled why this only happens for the activity icon. All other icon in the files sidebar of c.nc.c are fine, although they also use the legacy icons and not material design.

But your changes fix the problem, so 👍

raimund-schluessler:
> I am a bit puzzled why this only happens for the activity icon. All other icon in the files sidebar of c.nc.c are fine, although they also use the legacy icons and not material design.

I found out what causes this. The activity icon specifies a width, height and viewBox of 32 px in the svg:
```svg
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<svg xmlns="http://www.w3.org/2000/svg" height="32" width="32" version="1.0" viewBox="0 0 32 32">
 <path d="m16 1-10 18h11l-1 12 10-18h-11z" fill="#FFF"/>
</svg>
```
whereas all other icons only specify a size of 16px, e.g. spreed:
```svg
<svg width="16" height="16" version="1.1" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg"><path d="m7.9992 0.999a6.9993 6.9994 0 0 0-6.9992 6.9996 6.9993 6.9994 0 0 0 6.9992 6.9994 6.9993 6.9994 0 0 0 3.6308-1.024c0.86024 0.34184 2.7871 1.356 3.2457 0.91794 0.47922-0.45765-0.56261-2.6116-0.81238-3.412a6.9993 6.9994 0 0 0 0.935-3.4814 6.9993 6.9994 0 0 0-6.9991-6.9993zm8e-4 2.6611a4.34 4.3401 0 0 1 4.34 4.3401 4.34 4.3401 0 0 1-4.34 4.3398 4.34 4.3401 0 0 1-4.34-4.3398 4.34 4.3401 0 0 1 4.34-4.3401z" fill="#fff" stroke-width=".14"/></svg>
```

So for the larger activity icon we have to explicitly setting the size correctly.

raimund-schluessler (approved):

CarlSchwan (approved):

PR #2375, opened 2021-11-29
Add `buttonTitle` property to ActionButton and PopoverMenuItem

mejo-:
The HTML button element supports the `title` property for tooltips.
Let's allow to set it in the ActionButton and PopoverMenuItem
components.

The name is `buttonTitle` instead of `title` because ActionButton
already has a `title` property with a different meaning and in
PopoverMenuItem elements other than button are supported. So let's be
explicit that this property is *only* for the `title` property of
`button` elements.

### DIFF
--- a/src/components/ActionButton/ActionButton.vue
+++ b/src/components/ActionButton/ActionButton.vue
@@ -86,6 +86,7 @@ export default {
 			class="action-button"
 			:class="{ focusable: isFocusable }"
 			:aria-label="ariaLabel"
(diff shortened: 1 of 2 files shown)


mejo-:
This PR is required for https://github.com/nextcloud/text/pull/1963

mejo-:
Mh, you're right @marcoambrosini! I should have gone with `v-tooltip` in the first place. Thanks for the hint!

PR #2406, opened 2021-12-14
Decrease padding right of AppNavigationCaption and ListItem to 4 px

raimund-schluessler:
The right padding of `AppNavigationCaption` was to big, leading to a misalignment of the ActionMenus of `AppNavigationItem` and `AppNavigationCaption`. This PR adjusts the padding to the correct value, see
https://github.com/nextcloud/nextcloud-vue/blob/0173e9e9b14b12066415b0152eebfc67e32966ed/src/components/AppNavigationItem/AppNavigationItem.vue#L484

Before:
Screenshot 2021-12-14 at 21-53-09 Nextcloud Vue Style Guide

After:
Screenshot 2021-12-14 at 21-58-33 Nextcloud Vue Style Guide

### DIFF
--- a/src/components/AppNavigationCaption/AppNavigationCaption.vue
+++ b/src/components/AppNavigationCaption/AppNavigationCaption.vue
@@ -57,7 +57,7 @@ export default {
 .app-navigation-caption {
 	display: flex;
 	justify-content: space-between;
-	padding: 0 8px 0 $clickable-area/2;
+	padding: 0 4px 0 $clickable-area/2;
 
 	&__title {
 		font-weight: bold;
--- a/src/components/ListItem/ListItem.vue
+++ b/src/components/ListItem/ListItem.vue
@@ -432,6 +432,7 @@ export default {
 	flex: 0 0 auto;
 	justify-content: flex-start;
 	padding: 8px;
+	padding-right: 4px;
 	border-radius: 16px;
 	margin: 2px 0;
 	width: 100%;


nimishavijay (approved):
Much better now!

marcoambrosini (commented):
This will make it out of alignment with the ListItem's actions though, which is what this component was matching. I would increase the `appNavigationItem` right padding instead.
Screenshot 2021-12-15 at 10 20 09

raimund-schluessler:
> This will make it out of alignment with the `ListItem`'s actions though

Hm, indeed, this won't work well:
Screenshot 2021-12-15 at 10-42-17 Nextcloud Vue Style Guide

> which I'm sure you'd love @raimund-schluessler

@marcoambrosini I don't really have a preference here, to be honest. I just had the lengthy discussion in https://github.com/nextcloud/nextcloud-vue/pull/1912 in mind where the `right-padding: 4px;` was done and thought that these 4 px are the desired value. But since I would just like them all to be aligned, I am fine with increasing the right-padding of `AppNavigationItem` to 8 px (although I am almost certain, this will mess up something else we don't think about at the moment 🙈 ).

@nextcloud/designers Should I adjust the PR to increase the padding of `AppNavigationItem` instead?

raimund-schluessler:
> @nextcloud/designers Should I adjust the PR to increase the padding of `AppNavigationItem` instead?

For reference:

Everything with 8 px (increasing `AppNavigationItem` right padding):
8px

And with 4 px (decreasing `ListItem` right padding):
4px

raimund-schluessler:
I adjusted this PR to change everything to 4 px:
4px

And I also created another PR #2407 adjusting everything to 8 px as comparison. So take your choice, the PR receiving more approvals will get merged 😉🙈

raimund-schluessler:
I will close this according to the discussion on c.nc.c. We will go for #2407 for now and completely rework the `AppNavigationItem` later.

PR #2409, opened 2021-12-16
Reduce top whitespace in the sidebar of text app on mobile

luka-nextcloud:
Signed-off-by: Luka Trovic <luka@nextcloud.com>

Before the fix:
image

After the fix:
image

### DIFF
--- a/src/components/AppSidebar/AppSidebar.vue
+++ b/src/components/AppSidebar/AppSidebar.vue
@@ -681,7 +681,7 @@ $top-buttons-spacing: 6px;
 		// header background
 		&__figure {
 			width: 100%;
-			height: 250px;
+			height: 20vh;
 			max-height: 250px;
 			background-repeat: no-repeat;
 			background-position: center;


raimund-schluessler:
This won't work well in all cases. For example, here the preview image will be very small:
Screenshot 2021-12-20 at 10-48-33 05 - Files - Nextcloud

What you reduced is not whitespace. It is the area where a figure will be shown, e.g. for file previews (you also see that by the selector name `&__figure`). Also in the example you show, it is not really white space, but the file preview:
Screenshot 2021-12-20 at 10-52-29 Files - Nextcloud

So I would vote to not reduce this. If a preview is generally undesired, an option might be to show a button to toggle it on and off.
@nextcloud/designers

juliushaertl:
I think the issue is mostly that when using the sidebar within the viewer, showing the preview is rather unnecessary, so I'm wondering if we should just always hide it in that case?

jancborchardt:
> I think the issue is mostly that when using the sidebar within the viewer, showing the preview is rather unnecessary, so I'm wondering if we should just always hide it in that case? 

Yes, at least for text documents it always makes sense to hide it.

For pictures, on mobile it could still be nice to show them (or collapse them a bit into a smaller header with the title next to it)?

skjnldsv:
> I think the issue is mostly that when using the sidebar within the viewer, showing the preview is rather unnecessary, so I'm wondering if we should just always hide it in that case?

IIrc we used to hide it :thinking:

raimund-schluessler:
Since the underlying issue is not the component itself, but how it's used in the app, I close this PR. Please reopen if you think there are changes to the components necessary.

juliushaertl:
Was addressed with https://github.com/nextcloud/server/pull/31408

PR #2411, opened 2021-12-20
Adjust breadcrumb appearance

### LINKED ISSUE #2410: Improve the breadcrumbs component
As discussed with @jancborchardt and @marcoambrosini during the Vue components design review

breadcrumbs-grey-padding

- The right arrow that separates the folders is super thin and sort of disappears into the bg, so it can be replaced by `chevron-right` MDI icon
- We can have a better hover feedback, where the background colour is `#ededed` (`--color-background-dark` ?) and text is `--color-main-text`
- Also used `home` MDI icon here 

also cc @skjnldsv for Files

raimund-schluessler:
First step to fix #2410.
The breadcrumb now uses `ChevronRight` as a separator and has a background as hover feedback.
Screenshot 2021-12-20 at 22-47-26 Inventory - Nextcloud

Peek 21-12-2021 07-55

For a live example, please have a look at the docs:
https://deploy-preview-2411--nextcloud-vue-components.netlify.app/#/Components/Breadcrumbs?id=breadcrumbs-1

I had to rename the component class to `vue-crumb` since the server CSS kept interfering. If we want to use `crumb` instead, we will have to explicitly overwrite all interfering server rules.

Still todo see https://github.com/nextcloud/nextcloud-vue/issues/2410#issuecomment-998298730

### DIFF
--- a/src/components/Breadcrumbs/Breadcrumbs.vue
+++ b/src/components/Breadcrumbs/Breadcrumbs.vue
@@ -286,12 +286,12 @@ export default {
 		 */
 		getWidth(el) {
 			if (!el.classList) return 0
-			const hide = el.classList.contains('crumb--hidden')
+			const hide = el.classList.contains('vue-crumb--hidden')
 			el.style.minWidth = 'auto'
-			el.classList.remove('crumb--hidden')
+			el.classList.remove('vue-crumb--hidden')
 			const w = el.offsetWidth
 			if (hide) {
-				el.classList.add('crumb--hidden')
+				el.classList.add('vue-crumb--hidden')
 			}
 			el.style.minWidth = ''
 			return w
@@ -341,8 +341,8 @@ export default {
 			this.menuBreadcrumbProps.open = false
 
 			// Remove all hovering classes
-			const crumbs = document.querySelectorAll('.crumb')
-			crumbs.forEach((f) => { f.classList.remove('crumb--hovered') })
+			const crumbs = document.querySelectorAll('.vue-crumb')
+			crumbs.forEach((f) => { f.classList.remove('vue-crumb--hovered') })
 			return this.preventDefault(e)
 		},
 		/**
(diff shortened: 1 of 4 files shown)


src/components/Breadcrumbs/Breadcrumbs.vue
```diff
@@ -286,12 +286,12 @@ export default {
 		 */
 		getWidth(el) {
 			if (!el.classList) return 0
-			const hide = el.classList.contains('crumb--hidden')
+			const hide = el.classList.contains('vue-crumb--hidden')
```
skjnldsv, line 289: Maybe we should use a constant to store the `vue-crumb` class prefix :wink:

skjnldsv (approved):
Looks great!! :rocket:

src/components/Breadcrumbs/Breadcrumbs.vue
raimund-schluessler: Would be good. I put it on the list.

nimishavijay:
Looks so much better! Only 2 tiny suggestions regarding colour:
- in the normal state instead of `--color-main-text` with 0.7 opacity for the text we could use `--color-text-maxcontrast` with opacity 1 
- in hovered state we could use `--color-main-text` for the text with opacity 1 so that there is more difference between the normal and hovered state

What do you think?

raimund-schluessler:
I don't mind. I just tried to resemble the text color currently used in the breadcrumbs, where we have three states: normal, hovered and focused.

> - in the normal state instead of `--color-main-text` with 0.7 opacity for the text we could use `--color-text-maxcontrast` with opacity 1 

The normal (not hovered, not focuses) uses opacity 0.5 with --color-main-text at the moment. Opacity 0.7 is for hover.

> - in hovered state we could use `--color-main-text` for the text with opacity 1 so that there is more difference between the normal and hovered state

Hovered is opacity 0.7. Focused uses opacity 1.

So if we use opacity 1 for hovering, we need another way to indicate focus, or make focus and hover look the same. What do you prefer?

marcoambrosini (approved):

raimund-schluessler:
@nimishavijay I hope you don't mind that I merge this now, but it makes it a lot easier to implement the other improvements. I would propose that you do a follow-up PR with your design adjustments. I am fine with whatever @nextcloud/designers agree on.

jancborchardt:
> So if we use opacity 1 for hovering, we need another way to indicate focus, or make focus and hover look the same. What do you prefer?

Yup, would be best if focus and hover look the same. :) And then as @nimishavijay described, not using opacity as modifier for the text color but `--color-text-maxcontrast` and `--color-main-text`.

And hey, looks very nice, thank you so much! :)

marcoambrosini:
Could we use the button component in the mid run? If we'd do that we'd get the outline focus from it

raimund-schluessler:
> Could we use the button component in the mid run? If we'd do that we'd get the outline focus from it only when navigating with the keyboard and no changes with the mouse.

I would strongly advise against that.
- it would add a lot of complexity, because we would need to include the `Button` in the `Breadcrumb` component and have an `Actions` menu inside. I don't even know if that would work.
- it is syntactically the wrong element. The breadcrumbs are for navigation, so they need to be links/`a`s. Wrapping the button in a link would work, but doesn't make sense, in my opinion.
- the button component would need adjustment, since the icon is on the wrong side (left for the button, needs to be right here)

PR #2414, opened 2021-12-21
Allow custom action menu icon for breadcrumb

raimund-schluessler:
This PR implements to customize the menu icon for the `Actions` menu of each breadcrumb. We can now e.g. use `MenuDown` for the actions menu of the last breadcrumb. I adjusted the docs to demonstrate this:
Screenshot 2021-12-21 at 20-55-39 Nextcloud Vue Style Guide

### DIFF
--- a/src/components/Breadcrumb/Breadcrumb.vue
+++ b/src/components/Breadcrumb/Breadcrumb.vue
@@ -32,6 +32,7 @@ This component is meant to be used inside a Breadcrumbs component.
 	<div ref="crumb"
 		class="vue-crumb"
 		:class="{'vue-crumb--with-action': $slots.default, 'vue-crumb--hovered': hovering}"
+		:[crumbId]="''"
 		draggable="false"
 		@dragstart.prevent="() => {/** Prevent the breadcrumb from being draggable. */}"
 		@drop.prevent="dropped"
@@ -52,8 +53,12 @@ This component is meant to be used inside a Breadcrumbs component.
 		<Actions ref="actions"
 			:force-menu="forceMenu"
 			:open="open"
-			@update:open="onOpenChange"
-			container=".vue-crumb--with-action.dropdown">
+			:container="`.vue-crumb--with-action[${crumbId}]`"
+			@update:open="onOpenChange">
+			<template #icon>
+				<!-- @slot Slot for the custom menu icon -->
+				<slot name="menu-icon" />
+			</template>
 			<!-- @slot All action elements passed into the default slot will be used -->
 			<slot />
 		</Actions>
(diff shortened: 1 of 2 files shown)


skjnldsv (approved):

jancborchardt:
@raimund-schluessler nice! :) 2 things:
- The arrow-down action icon is a bit far away from the folder name. Could it be moved closer to it? The whole "Folder 5" entry should be clickable in this case, basically be a button with an icon on the _right_.
- Possibly for a different pull request: The current folder (the last one) would be better in bold so it’s clear which one has focus.

raimund-schluessler:
@jancborchardt Your requested changes are in https://github.com/nextcloud/nextcloud-vue/pull/2416. Would be great if we could get this one here in.

quentinguidee (approved):
Fine for me! We just have to keep in mind that with the implementation discussed in #2416 (I don't think it needs to be explicit in the docs?), another symbol than MenuDown or MenuUp (if the bar is at the bottom) will not be explicit and requires a separate button elsewhere on the interface (ex: if the action icon is "share", it is better to have a share button elsewhere on the interface)

raimund-schluessler:
> Fine for me! We just have to keep in mind that with the implementation discussed in #2416 (I don't think it needs to be explicit in the docs?), another symbol than MenuDown or MenuUp (if the bar is at the bottom) will not be explicit and requires a separate button elsewhere on the interface (ex: if the action icon is "share", it is better to have a share button elsewhere on the interface)

I am not sure I understand what you mean. Do you mean if there is only a single action which is not in a dropdown #2416 will not work well? Anyway, let's discuss this there.

quentinguidee:
> Do you mean if there is only a single action which is not in a dropdown #2416 will not work well?

Well, I think it depends on what it will do. Imagine a share icon.

If the share icon is part of the text and only here to say "the folder is shared", it's ok :

image

But if it's an action that says "share this folder", I don't think it will be intuitive. Passing my mouse over the folder 5, Imagine an user that don't really know that this "icon with three circles and two lines" is a share icon, he'll not understand what it does before clicking (whereas a button with an arrow is way more easy to understand without clicking).

In that case I think it would be better to have a real button next to the breadcrumbs bar (or elsewhere) with the text "Share" and this icon next to it.

So in my opinion anyone who uses a custom icon here will have to think carefully before doing so.

raimund-schluessler:
@quentinguidee I see what you mean and I agree that it might be problematic. But since this problem will be introduced by #2416 it is better to discuss this there. In this PR here the dropdown or single action is still clearly visible as a button/action since it's not part of the hover background yet. So I will merge this one here, rebase #2416 and we can further discuss the problem there.

PR #2419, opened 2021-12-27
Add option to always show actions

### LINKED ISSUE #2340: New list item component, action menu mobile use and accessbility
Mail uses the new app content list item component from nc/vue. This looks and works great on desktop, but on mobile there is a little accessibility issue IMO. The issues is that by default we always show the timestamp of the message, on hover (desktop) the actions menu handle shows. There is no hover on mobile, so that menu is a "hidden" feature. You can still click and get a menu, but a user wouldn't know about this.

@nextcloud/designers possibly something to discuss

marcoambrosini:
fixes #2340

Added new prop that allows developers to always show the actions to the right of details and counter. See second item in the list.

https://user-images.githubusercontent.com/26852655/147488721-bdec8b86-c327-41a7-87a2-0dc5369ea88f.mov

### DIFF
--- a/src/components/ListItem/ListItem.vue
+++ b/src/components/ListItem/ListItem.vue
@@ -52,6 +52,34 @@
 	</listItem>
 	<listItem
 		:title="'Title of the element'"
+		:bold="false"
+		:force-display-actions="true"
+		:details="'One hour ago'"
+		:counter-number="44"
+		counterType="highlighted">
+		<template #icon>
+			<avatar :size="44" user="janedoe" display-name="Jane Doe" />
+		</template>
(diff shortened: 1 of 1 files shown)


skjnldsv (approved):

GretaD (approved):
tested with mail, and works really nice

jancborchardt:
@marcoambrosini just for info – is this now a per-list-item setting? Cause it should always be applied to all ListItems in the AppContentList, right?

marcoambrosini:
It is on a per component basis, but in general you use only one component and loop over some data, so developers will likely have to add the prop only once.

PR #2431, opened 2022-01-03
Set sidebar tab content height to be 100 percent

### LINKED ISSUE #2430: App sidebar tab content height does not let it's child to fill the content
## Problem Description
Currently the app sidebar tab content has height defined as:
```css
.app-sidebar__tab {
	...
	min-height: 100%; // fill available height
	max-height: 100%; // scroll inside
        ...
}
```
https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/AppSidebarTab/AppSidebarTab.vue#L89-L90

The CSS definition likewise contains nice comments, but, in reality, this limits the inner child to fill the tab's content _(section)_ height.

## Actual Behaviour
problem

If we have to show some empty messages like this or have to center elements within the tab section, we've to add a hard-coded margin or padding to the element which is really not cool in terms of responsiveness.

Even if we try with display flex for the `section.app-sidebar__tab` child, the content is not fit with the `section` height.

## Expected Behaviour
If we just set the height of the sidebar__tab to `100%` everything just works fine and now we can fill the content inside the child elements.
Peek 2022-01-03 12-18

Even the scrolling works just as fine:

scrolling

kiranparajuli589:
Fixes #2430

### DIFF
--- a/src/components/AppSidebarTab/AppSidebarTab.vue
+++ b/src/components/AppSidebarTab/AppSidebarTab.vue
@@ -88,6 +88,7 @@ export default {
 	padding: 10px;
 	min-height: 100%; // fill available height
 	max-height: 100%; // scroll inside
+	height: 100%;
 	overflow: auto;
 
 	&:focus {


src/components/AppSidebarTab/AppSidebarTab.vue
```diff
@@ -86,8 +86,7 @@ export default {
 .app-sidebar__tab {
 	display: none;
 	padding: 10px;
-	min-height: 100%; // fill available height
-	max-height: 100%; // scroll inside
+	height: 100%;
```
raimund-schluessler, line 89: Please keep the min-height and max-height values, so it's less likely we break stuff. Adding `height: 100%;` should be ok.

raimund-schluessler:
I briefly checked with
- Files
- Talk
- Photos
- Calendar
- Tasks
- Deck

Adding `height: 100%` seems to be fine.

src/components/AppSidebarTab/AppSidebarTab.vue
kiranparajuli589: that's totally fine. I've reverted the deleted lines. :+1:

raimund-schluessler (approved):

skjnldsv (approved):
:elephant:

PR #2435, opened 2022-01-04
Fix md icon position for Actions with title

raimund-schluessler:
For Actions with menu-title showing a material design icon, the icon position was off.
Before:
Screenshot 2022-01-04 at 13-55-57 Nextcloud Vue Style Guide

After:
Screenshot 2022-01-04 at 14-01-27 Nextcloud Vue Style Guide

### DIFF
--- a/src/components/Actions/Actions.vue
+++ b/src/components/Actions/Actions.vue
@@ -724,8 +724,8 @@ export default {
 				height: 24px;
 				line-height: $icon-size;
 				position: absolute;
-				top: ($clickable-area - 24px) / 2;
-				left: ($clickable-area - 24px) / 2;
+				top: 0;
+				left: 0;
 			}
 		}
 


GretaD (approved):
code look good, didnt test it

artonge:
Does this still work with non material design icons? not sure if they are still used or not.

raimund-schluessler:
> Does this still work with non material design icons? not sure if they are still used or not.

Checkout the netlify docs: https://deploy-preview-2435--nextcloud-vue-components.netlify.app/#/Components/Actions?id=actions-1

It should work with non material design icons (provided via `default-icon` prop) and with material design icons provided by the icon slot.

The problem fixed by this PR was not visible in the docs, because we don't have an example with `menu-title` and md icon there.

artonge (approved):

PR #2436, opened 2022-01-04
Show menu title for single Action

raimund-schluessler:
This changes the behavior of the `Actions` component so that if a `menu-title` is provided, it can be shown for single Actions as well by setting the `forceTitle` prop to `true`. This is necessary for the changes to the Breadcrumbs component requested in #2416 so that the breadcrumb title of the last crumb is also shown if only a single Action is given. In case of multiple Actions, i.e. with a dropdown, or when `menu-title` or `force-title` is not set, nothing changes.

Before:
Screenshot 2022-01-04 at 15-03-58 Nextcloud Vue Style Guide

After:
Screenshot 2022-01-05 at 09-19-20 Nextcloud Vue Style Guide

~@skjnldsv @nextcloud/vuejs In the current form this is a breaking change, since `menu-title` will be shown if it is set, which it was not before for single actions. Do you prefer that I keep the previous behavior and make showing the title for single actions optional by a new prop?~
Edit: I decided to make it a non-breaking change by adding a prop `force-title` to show the title for single actions if a title is given. Having an additional prop seems ok, instead of introducing a breaking change.

### DIFF
--- a/src/components/Actions/Actions.vue
+++ b/src/components/Actions/Actions.vue
@@ -138,7 +138,8 @@ export default {
 		v-bind="firstActionBinding"
 		:class="{
 			[firstAction.icon]: firstAction.icon,
-			[firstActionClass]: firstActionClass }"
+			[firstActionClass]: firstActionClass,
+			'action-item--single--with-title': singleActionTitle }"
 		class="action-item action-item--single"
 		rel="nofollow noreferrer noopener"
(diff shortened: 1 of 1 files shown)


skjnldsv (approved):

jancborchardt (approved):
Looks good design-wise

PR #2444, opened 2022-01-06
Implement showing global crumb actions

raimund-schluessler:
In order to show an `Actions` or `Button`component (or actually anything else you put into the slot) directly adjacent to the right of the last crumb, the element needs to be a child of the breadcrumbs container. So I added a slot and take the elements width into account when hiding the overflowing crumbs.

This is how it looks now with a button:
crumb
And an Actions component:
crumb1

This is the last step (maybe 🙂) in order to make it look like @nimishavijay mockup:
mockup

### DIFF
--- a/src/components/Breadcrumbs/Breadcrumbs.vue
+++ b/src/components/Breadcrumbs/Breadcrumbs.vue
@@ -61,6 +61,14 @@ is dropped on a creadcrumb.
 						Download
 					</ActionButton>
 				</Breadcrumb>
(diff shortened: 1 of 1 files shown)


skjnldsv (approved):

nimishavijay (approved):
🚀

marcoambrosini (approved):

jancborchardt (commented):
Awesome work @raimund-schluessler! :)

PR #2483, opened 2022-02-09
AppNavigationItem.vue: updating of this.$slots

JuliaKirschenheuter:
fixes https://github.com/nextcloud/mail/issues/5987

this.$slots will be updated in created() and beforeUpdate() lifecycle-hooks


ChristophWurst (approved):
`this.$slots` is not reactive, hence we need to use the lifecycle hooks

:+1:

st3iny (approved):
Excellent :shipit:

PR #2492, opened 2022-02-11
Fix shrinking of last breadcrumb for narrow width

raimund-schluessler:
After the major changes to the breadcrumb bar in #2411 and #2444 the last breadcrumb does not shrink anymore on small screens. This PR fixes it.

| | Before | After |
| ------------- |:-------------:| -----:|
| Without Actions      | before_without  | after_without |
| With Actions      | before_with | after_with |

### DIFF
--- a/src/components/Breadcrumb/Breadcrumb.vue
+++ b/src/components/Breadcrumb/Breadcrumb.vue
@@ -296,8 +296,22 @@ export default {
 	// Adjust action item appearance for crumbs with actions
 	// to match other crumbs
 	&::v-deep .action-item {
+		// Adjustments necessary to correctly shrink on small screens
+		max-width: 100%;
+		.trigger {
+			max-width: 100%;
+		}
+
 		&__menutoggle--with-title,
 		&--single--with-title {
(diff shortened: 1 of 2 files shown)


jancborchardt (approved):
Nice, looks good! :)
Btw, at a certain width, should the "Home" icon maybe also be ellipsized into the 3-dot action menu?

GretaD (approved):
code looks good, didnt test it

raimund-schluessler:
> Btw, at a certain width, should the "Home" icon maybe also be ellipsized into the 3-dot action menu?

I don't have a strong opinion on this. It would save anouther 64 px in width, which would enable showing a bit more of the last folders name. But I am not sure if it's necessary.

This is how it looks on iPhone8.
Currently:
Screenshot 2022-02-14 at 13-42-02 Inventar - Nextcloud
Without the Home icon:
Screenshot 2022-02-14 at 13-44-47 Inventar - Nextcloud

jancborchardt:
@raimund-schluessler yeah, just a thought, we can see if that ever comes up, then we cpuld do that. :)

PR #2496, opened 2022-02-14
increase the small modal width

szaimen:

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -725,7 +725,8 @@ export default {
 	// Sizing
 	&--small {
 		.modal-container {
-			width: 330px;
+			width: 390px;
+			max-width: 100%;
 			max-height: 90%;
 		}
 	}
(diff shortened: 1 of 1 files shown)


jancborchardt:
Looks good, but max-width is 100% while max-height is 90% – those should probably be the same, right? Could do 90% (or 95% if needed) on both so it's still a modal-type.

Also @marcoambrosini this is the fix we talked about for the Talk modals. :)

szaimen:
> Looks good, but max-width is 100% while max-height is 90% – those should probably be the same, right? Could do 90% (or 95% if needed)

Not sure about this. I had mobile devices in mind where having as much width as possible is probably desired? But I guess 95% should be fine. What do you think?

jancborchardt:
We do full width/height for the other ones, but for the small size max 95% to have a little bit of modal-ness would be good. :)

szaimen:
> We do full width/height for the other ones, but for the small size max 95% to have a little bit of modal-ness would be good. :)

fine by me then

jancborchardt (approved):

szaimen:
@marcoambrosini could you give your 👍 if it looks fine to you?

marcoambrosini (approved):
Looks goood :)

PR #2516, opened 2022-02-22
Add button native type prop on <Button>

tcitworld:
Allows to have `<Button native-type="submit">Send</Button>` in a form and listen on the `@submit` event instead of the `@click` one.


ChristophWurst:
@JuliaKirschenheuter didn't you also make this suggestion because it would handle some accessibility aspects better?

marcoambrosini (approved):

JuliaKirschenheuter (approved):

PR #2526, opened 2022-02-28
Bump version to v5.0.1

max-nextcloud:
Signed-off-by: Max <max@nextcloud.com>


raimund-schluessler:
@max-nextcloud I would prefer if this was 5.1.0 since we have "Enhancements" and e.g. having a new prop in https://github.com/nextcloud/nextcloud-vue/pull/2516 is not a Fix only. Sorry, I saw this to late in the chat.

PR #2533, opened 2022-03-03
[MultiSelect] pass the #noResult slot to the underlying component

eneiluj:
Pretty much all the slots are passed down to Vue-Multiselect except `noResult` which is hardcoded. This is a naive way to pass it down and keep the default value if not defined. Maybe there's a more elegant way.

### DIFF
--- a/src/components/Multiselect/Multiselect.vue
+++ b/src/components/Multiselect/Multiselect.vue
@@ -215,7 +215,9 @@ export default {
 		</template>
 
 		<template #noResult>
-			<span>{{ t('No results') }}</span>
+			<slot name="noResult">
+				<span>{{ t('No results') }}</span>
+			</slot>
 		</template>
 	</VueMultiselect>
 </template>


skjnldsv:
It was done on purpose. Do we really need to make that slot available? What other message do you want to put? :)

kiranparajuli589:
it would be really helpful if we can change the **#noResult** message to `we're fetching for you` if it's still searching, and `sorry, no results` if no results are found after the search request has been completed.

a static message would not be too helpful in my opinion.

skjnldsv:
> if it's still searching

That should then be fixed and covered by the library itself?

eneiluj:
> It was done on purpose

@skjnldsv I feel like it's important to allow those using our MultiSelect to override this and set their own slot content. Alternative messages could be "Nothing found" or whatever text matching the app context (more than "no results").

By making this slot available we also allow conditional content in the slot.

> if it's still searching

@kiranparajuli589 Just to make sure we're on the same page: If we can override the #noResult slot, then you can define it like
``` vue
<template #noResult>
    <span v-if="stillSearching">we're fetching for you</span>
    <span v-else>sorry, no results</span>
</template>
```

kiranparajuli589:
> > if it's still searching
> 
> @kiranparajuli589 Just to make sure we're on the same page: If we can override the #noResult slot, then you can define it like
> 
> ```
> <template #noResult>
>     <span v-if="stillSearching">we're fetching for you</span>
>     <span v-else>sorry, no results</span>
> </template>
> ```
@eneiluj absolutely,

skjnldsv:
I would argue against it, as I would say it's best to be consistent in the feedback messages, thing that usually devs are not great at. :)
But I'm not the only one to decide here, other than this the code looks good, I'm ok to merge :shrug:

eneiluj:
@skjnldsv I agree devs are not always putting consistent messages :grin:. I still feel there is the need to be able to set the `#noResult` one in contexts where "No results" makes no sense or is not specific enough.

juliushaertl:
I think having at least the option to overwrite the message makes sense 👍

juliushaertl (approved):

juliushaertl:
Moving milestone to 6.0, if needed, please consider a backport to stable5 after merge

marcoambrosini (approved):

PR #2548, opened 2022-03-14
Add aria-description to AppNavigationItem

JuliaKirschenheuter:
required for https://github.com/nextcloud/calendar/issues/3955
used for https://github.com/nextcloud/calendar/pull/4045

Documentation for "aria-description": https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-description

### DIFF
--- a/src/components/AppNavigationItem/AppNavigationItem.vue
+++ b/src/components/AppNavigationItem/AppNavigationItem.vue
@@ -126,6 +126,7 @@ Just set the `pinned` prop.
 		<!-- Icon and title -->
 		<a v-if="!undo"
 			class="app-navigation-entry-link"
+			:aria-description="ariaDescription"
 			href="#"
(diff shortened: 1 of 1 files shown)


skjnldsv (approved):

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -354,6 +355,13 @@ export default {
 			type: String,
 			default: 'bottom',
 		},
+		/**
+		 * Entry aria details
+		 */
+		ariaDescription: {
+			type: String,
+			default: '',
```
skjnldsv, line 363: Should be null by default :)
image

skjnldsv (changes requested):

ChristophWurst (approved):

PR #2552, opened 2022-03-16
Make emojis selectable by tab

skjnldsv:
image

- on open : Search is in focus 

- if search is in focus, emojis are selectable by arrow keys  

- Categories and emojis are selectable by tab

Is this the behavior we want to see?

### DIFF
--- a/src/components/EmojiPicker/EmojiPicker.vue
+++ b/src/components/EmojiPicker/EmojiPicker.vue
@@ -92,11 +92,18 @@
 		popover-class="emoji-popover"
 		popover-inner-class="popover-emoji-picker-inner"
 		v-bind="$attrs"
-		v-on="$listeners">
+		v-on="$listeners"
+		@after-show="afterShow">
 		<template #trigger>
 			<slot />
 		</template>
+<<<<<<< HEAD
 		<Picker :auto-focus="true"
+=======
+		<Picker
+			ref="picker"
+			:auto-focus="true"
+>>>>>>> 4da67de7... make selectable by tab
 			color="var(--color-primary)"
 			:data="emojiIndex"
 			:emoji="previewFallbackEmoji"
@@ -217,6 +224,15 @@ export default {
 				this.open = false
 			}
 		},
+		afterShow() {
+			const picker = this.$refs.picker
+			const input = picker.$refs.search.$el.querySelector('input')
+
+			if (input) {
+				input.focus()
+			}
+
+		},
 	},
 }
 </script>
@@ -254,8 +270,15 @@ export default {
 		font-size: inherit;
 		height: 36px;
 		width: auto;
+
 		* {
 			cursor: pointer !important;
+		}
+
+		&:focus-visible {
+			background-color: var(--color-background-hover);
+			border: 2px solid var(--color-primary-element) !important;
+			border-radius: 50%;
 		}
 	}
 
@@ -310,8 +333,10 @@ export default {
 			flex-basis: calc(100% / 8);
 			text-align: center;
 
-			&:hover::before {
-				background-color: var(--color-background-hover);
+			&:hover::before,
+			&.emoji-mart-emoji-selected::before{
+				background-color: var(--color-background-hover) !important;
+				outline: 2px solid var(--color-primary-element);
 			}
 		}
 	}
--- a/src/components/EmojiPicker/EmojiPicker.vue
+++ b/src/components/EmojiPicker/EmojiPicker.vue
@@ -93,17 +93,13 @@
 		popover-inner-class="popover-emoji-picker-inner"
 		v-bind="$attrs"
 		v-on="$listeners"
-		@after-show="afterShow">
+		@after-show="afterShow"
+		@after-hide="afterHide">
 		<template #trigger>
 			<slot />
 		</template>
-<<<<<<< HEAD
-		<Picker :auto-focus="true"
-=======
-		<Picker
-			ref="picker"
+		<Picker ref="picker"
 			:auto-focus="true"
->>>>>>> 4da67de7... make selectable by tab
 			color="var(--color-primary)"
 			:data="emojiIndex"
 			:emoji="previewFallbackEmoji"
@@ -225,13 +221,43 @@ export default {
 			}
 		},
 		afterShow() {
+			// add focus trap in modal
 			const picker = this.$refs.picker
+			picker.$el.addEventListener('keydown', this.checkKeyEvent)
+			// set focus on input search field
 			const input = picker.$refs.search.$el.querySelector('input')
-
 			if (input) {
 				input.focus()
 			}
-
+		},
+		afterHide() {
+			// remove keydown listner if popover is hidden
+			const picker = this.$refs.picker
+			picker.$el.removeEventListener('keydown', this.checkKeyEvent)
+		},
+		checkKeyEvent(event) {
+			if (event.key !== 'Tab') {
+				return
+			}
+			const picker = this.$refs.picker
+			const focusableList = picker.$el.querySelectorAll(
+				'button, input'
+			)
+			const last = focusableList.length - 1
+			// escape early if only 1 or no elements to focus
+			if (focusableList.length <= 1) {
+				event.preventDefault()
+				return
+			}
+			if (event.shiftKey === false && event.target === focusableList[last]) {
+				// Jump to first item when pressing tab on the latest item
+				event.preventDefault()
+				focusableList[0].focus()
+			} else if (event.shiftKey === true && event.target === focusableList[0]) {
+				// Jump to the last item if pressing shift+tab on the first item
+				event.preventDefault()
+				focusableList[last].focus()
+			}
 		},
 	},
 }
@@ -274,12 +300,6 @@ export default {
 		* {
 			cursor: pointer !important;
 		}
-
-		&:focus-visible {
-			background-color: var(--color-background-hover);
-			border: 2px solid var(--color-primary-element) !important;
-			border-radius: 50%;
-		}
 	}
 
 	.emoji-mart-bar,
@@ -295,6 +315,11 @@ export default {
 		color: inherit !important;
 	}
 
+	.emoji-mart-search input:focus-visible {
+		box-shadow: inset 0 0 0 2px var(--color-primary);
+		outline: none;
+	}
+
 	.emoji-mart-bar {
 		&:first-child {
 			border-top-left-radius: var(--border-radius) !important;
@@ -307,6 +332,10 @@ export default {
 			border-radius: 0;
 			padding: 12px 4px;
 			height: auto;
+			&:focus-visible {
+				/* box-shadow: inset 0 0 0 2px var(--color-primary); */
+				outline: 2px solid var(--color-primary-element);
+			}
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


src/components/EmojiPicker/EmojiPicker.vue
```diff
@@ -307,8 +323,9 @@ export default {
 			flex-basis: calc(100% / 8);
 			text-align: center;
 
-			&:hover::before {
-				background-color: var(--color-background-hover);
```
skjnldsv, line 311: Should we lose th default hover? I think we should keep it, no?

src/components/EmojiPicker/EmojiPicker.vue
```diff
@@ -307,8 +323,9 @@ export default {
...
+			&:hover,
+			&.emoji-mart-emoji-selected{
+				border: 1px solid var(--color-primary);
```
skjnldsv, line 328: Border takes space, not the outline :)

skjnldsv: But I just tested and it doesn't seems to make the layout jump :thinking: 
I guess that's alright, even if I would prefer being consistent with the categories styling

src/components/EmojiPicker/EmojiPicker.vue
skjnldsv: Please not that the default styling is still here. Meaning there is still a rounded background, but it's not compatible with dark theme anymore. We need to put back the `background-color: var(--color-background-hover);`

skjnldsv: image

src/components/EmojiPicker/EmojiPicker.vue
skjnldsv, line 328:
```suggestion
			&:hover::before,
			&.emoji-mart-emoji-selected::before {
				background-color: var(--color-background-hover);
```

nimishavijay:
> on open : Search is in focus
if search is in focus, emojis are selectable by arrow keys

Looks good! :)

> Categories and emojis are selectable by tab

Not sure about this, I think arrow keys can be used to select emojis and categories also. Tab can used to change focus between categories, search, and emojis. Within that you navigate using arrow keys.

Also a couple more points from me:
- Esc should close the picker
- If you have selected the last emoji in the list, pressing the right arrow should cycle back to the first one again. Currently the right arrow does nothing if you are on the last emoji. Similarly for left arrow and first emoji as well. 
- I also think that if the emoji picker is open, tab should not make focus leave the picker, it should only cycle through categories, search and emojis. Currently if you are focused on the last emoji and you press tab, the focus jumps to the browser tab which is a bit jarring. What do you think?

cc @jancborchardt if there is anything else to add

src/components/EmojiPicker/EmojiPicker.vue
vanpertsch: yes, i see. 
If we leave it like this, we won't have an outline when we use the arrow keys. That was what I was trying to achieve.

vanpertsch: might be a chrome thing that i do not see the hover effect. I ca see it in firefox ...

skjnldsv: I can see it in chrome :)

skjnldsv:
> * If you have selected the last emoji in the list, pressing the right arrow should cycle back to the first one again. Currently the right arrow does nothing if you are on the last emoji. Similarly for left arrow and first emoji as well.
> * I also think that if the emoji picker is open, tab should not make focus leave the picker, it should only cycle through categories, search and emojis. Currently if you are focused on the last emoji and you press tab, the focus jumps to the browser tab which is a bit jarring. What do you think?

Not sure those two are doable

jancborchardt:
Only a detail from my side: Would look nicer to have the focus box be a circle instead of a square with rounded corners? :)

src/components/EmojiPicker/EmojiPicker.vue
vanpertsch: ... oh no. it is my monitor! never detected this before. I can't see the slightest change in any browser.  (thought i was going crazy for a second).
ok. take back my changes. sry for the confusion.

vanpertsch:
@skjnldsv

i removed  `background-color: var(--color-background-hover)` in line 332 and only gave it an outline. 

- `background-color: var(--color-background-hover)` is #f5f5f5 
- default emoji-picker on hover/selected is #f4f4f4

for me the overwritten default brought some confusion. 

**Before**
Screenshot 2022-03-23 at 10 00 40
with `background-color: var(--color-background-hover)` you can hardly see hover in dark mode
Screenshot 2022-03-23 at 10 00 52
here the default `background-color: #f4f4f4` is on selected

**After**
Screenshot 2022-03-23 at 10 02 08

Screenshot 2022-03-23 at 10 37 57


now` #f4f4f4` is background-color on hover, focus and selected in dark and light mode. is this an option or against to much against taking the customstyle `var(--color-background-hover)`?

src/components/EmojiPicker/EmojiPicker.vue
```diff
@@ -309,8 +327,9 @@ export default {
 			flex-basis: calc(100% / 8);
 			text-align: center;
 
-			&:hover::before {
-				background-color: var(--color-background-hover);
+			&:hover::before,
+			&.emoji-mart-emoji-selected::before{
+				outline: 2px solid var(--color-primary);
```
nickvergessen, line 332:
```suggestion
				outline: 2px solid var(--color-primary-element);
```

src/components/EmojiPicker/EmojiPicker.vue
```diff
@@ -253,9 +264,16 @@ export default {
...
 		height: 36px;
 		width: auto;
+
 		* {
 			cursor: pointer !important;
 		}
+
+		&:focus {
+			background-color: #f4f4f4;
+			border: 2px solid var(--color-primary) !important;
```
nickvergessen, line 274:
```suggestion
			border: 2px solid var(--color-primary-element) !important;
```

vanpertsch:
@nimishavijay & @jancborchardt can you check again if the picker looks good and works as expected like it is now?

> - If you have selected the last emoji in the list, pressing the right arrow should cycle back to the first one again. Currently the right arrow does nothing if you are on the last emoji. Similarly for left arrow and first emoji as well.

- [ ]  **is not possible (atm)**

> - I also think that if the emoji picker is open, tab should not make focus leave the picker, it should only cycle through categories, search and emojis. Currently if you are focused on the last emoji and you press tab, the focus jumps to the browser tab which is a bit jarring. 

- [x]  **focus stays inside** 

**Search is in focus on open**
Screenshot 2022-04-20 at 15 24 36|Screenshot 2022-04-20 at 15 46 25


**Categories selectable by tab** 
Screenshot 2022-04-20 at 15 24 25|Screenshot 2022-04-20 at 15 48 37

**emojis selectable by tab and arrow-keys**
Screenshot 2022-04-20 at 15 24 47|Screenshot 2022-04-20 at 15 47 04

skjnldsv:
Approval from my side, you can self-approve on my behalf @vanpertsch 
:)

vanpertsch (approved):

src/components/EmojiPicker/EmojiPicker.vue
```diff
@@ -217,6 +219,46 @@ export default {
 				this.open = false
 			}
 		},
+		afterShow() {
```
artonge, line 222: Isn't it necessary to remove the event listener on `after-hide`?

src/components/EmojiPicker/EmojiPicker.vue
```diff
@@ -217,6 +219,46 @@ export default {
...
+			const picker = this.$refs.picker
+			const input = picker.$refs.search.$el.querySelector('input')
+
+			// add focus trap in modal
+			picker.$el.addEventListener('keydown', this.checkKeyEvent)
+
+			if (input) {
+				input.focus()
+			}
+
```
artonge, line 232: Extra white line

src/components/EmojiPicker/EmojiPicker.vue
```diff
@@ -217,6 +219,46 @@ export default {
...
+		},
+		checkKeyEvent(event) {
+			const picker = this.$refs.picker
```
artonge, line 235: Shouldn't we escape early instead of wrapping all the code of this function in conditions?

```js
if (event.key !== 'Tab') {
    return
}
```

artonge: Also, `event.preventDefault()` seems to be called in any cases. Might be better to have only one line for that.

src/components/EmojiPicker/EmojiPicker.vue
```diff
@@ -217,6 +219,46 @@ export default {
...
+				event.preventDefault()
+				return
+			}
+			const last = focusableList.length - 1
+			if (
+				event.key === 'Tab'
+				&& event.shiftKey === false
+				&& event.target === focusableList[last]
+			) {
+				event.preventDefault()
```
artonge, line 251: Comment so it can't be misunderstood.

`// Jump to first item when pressing tab on the latest item`

src/components/EmojiPicker/EmojiPicker.vue
vanpertsch: ✅ i added it. (My thinking was, that it is not necessary, because the element is removed from the DOM. But i am not sure. :))

src/components/EmojiPicker/EmojiPicker.vue
vanpertsch: ✅

src/components/EmojiPicker/EmojiPicker.vue
```diff
@@ -217,6 +219,46 @@ export default {
...
+
+		},
+		checkKeyEvent(event) {
+			const picker = this.$refs.picker
+
+			const focusableList = picker.$el.querySelectorAll(
+				'button, input'
+			)
+			// escape early if only 1 or no elements to focus
+			if (focusableList.length < 2 && event.key === 'Tab') {
```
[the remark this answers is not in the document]
vanpertsch: ✅

src/components/EmojiPicker/EmojiPicker.vue
vanpertsch: ✅ i changed the process of the method. You think it works better now?

src/components/EmojiPicker/EmojiPicker.vue
vanpertsch: ✅

src/components/EmojiPicker/EmojiPicker.vue
```diff
@@ -217,6 +219,46 @@ export default {
...
+			const last = focusableList.length - 1
+			if (
+				event.key === 'Tab'
+				&& event.shiftKey === false
+				&& event.target === focusableList[last]
+			) {
+				event.preventDefault()
+				focusableList[0].focus()
+			} else if (
+				event.key === 'Tab'
```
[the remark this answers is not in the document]
vanpertsch: ✅

nimishavijay (approved):
🚀  :)

src/components/EmojiPicker/EmojiPicker.vue
artonge: Beside the extra console.log, it looks good :)

artonge: Beside the extra console.log, it looks good :)

vanpertsch: :) removed

artonge (approved):
:ok_hand:

PR #2555, opened 2022-03-17
Add action menu with trigger per link

### LINKED ISSUE #2543: AppNavigationItem `add xxx` design changes
Small adjustment after call with @nimishavijay:
- Whole "Add calendar" row is clickable
- Menu has the 3 entries, and is centered below the click area
- "+" is an icon on the left
image

Does that work @ChristophWurst? This was also the originally intended design for the "New calendar" section in the first place, but then for some reason (component limitation?) the action menu on the right was used.

_Originally posted by @jancborchardt in https://github.com/nextcloud/calendar/issues/3965#issuecomment-1048681112_

JuliaKirschenheuter:
resolves #2543 
resolves https://github.com/nextcloud/calendar/issues/3965

Add slot "trigger" for Actions.vue to place menu inside of popover.
Add trigger="click" to Popover for correct listeners.
Add AppNavigationItemTitle.vue with link.
Add style changes for link menu.
Add documentation for menu with link for AppNavigationItem.vue.

### DIFF
--- a/src/components/Actions/Actions.vue
+++ b/src/components/Actions/Actions.vue
@@ -161,7 +161,8 @@ export default {
 	<!-- more than one action -->
 	<div v-else
 		v-show="hasMultipleActions || forceMenu"
-		:class="{'action-item--open': opened}"
+		:class="{'action-item--open': opened,
+			'action-item--title--with-trigger': useTitleAsActionTrigger,}"
 		class="action-item">
 		<!-- If more than one action, create a popovermenu -->
 		<Popover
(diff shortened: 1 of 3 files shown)


JuliaKirschenheuter:
New link menu is working. 
image
Interactive documentation is not working well. @marcoambrosini @skjnldsv please let me know what is missing. Thank you!

@jancborchardt @nimishavijay please give a feedback ;)

marcoambrosini:
@jancborchardt why don't we use a button here?

marcoambrosini (commented):
@jancborchardt why don't we use a button here? Code-wise i think it's ok but design-wise this feels like a little bit of a hack. It is a button,  but text and icon are not horizontally centered and the menu is anchored in a weird spot to the right of the text.

raimund-schluessler:
> @jancborchardt why don't we use a button here? Code-wise i think it's ok but design-wise this feels like a little bit of a hack. It is a button, but text and icon are not horizontally centered and the menu is anchored in a weird spot to the right of the text.

It's meant to be used as an AppNavigationItem (here below the list of calendars in the Calendar app). I think it makes sense here to have the icon on the very left, just aligned with the other entries of the AppNavigation list:
Screenshot 2022-03-23 at 10-50-07 Aufgaben - Nextcloud

marcoambrosini:
Also, some time ago we talked about how we were going to steer all these "create new item" for the app navigation into using the caption component. Just like it's done for appointments right below it.
 
Screenshot 2022-03-23 at 10 49 28

jancborchardt:
@marcoambrosini the original intention was to just fix the accessibility here, and since one doesn't often add a new calendar (unlike adding an event) a button is too much there.

But you are right, if we want to do it futureproof and like in Contacts and for appointments, the "Calendars" caption with the + icon next to it is a good way to do it.

ChristophWurst:
> But you are right, if we want to do it futureproof and like in Contacts and for appointments, the "Calendars" caption with the + icon next to it is a good way to do it.

Does that mean we discard the changes in here and take that approach?

PR #2560, opened 2022-03-20
Debug iOS breadcrumbs issue

raimund-schluessler:
Do not merge this. :wink: 

This is just a debug PR to figure out why the breadcrumb on the right of the hidden breadcrumbs dropdown moves to the very right when opening the dropdown for the first time on iOS. Since I don't own a mac, I cannot debug this nicely with the webinspector and have to mess around like this :see_no_evil:

This is how it looks on iOS when opened:

If anyone can figure this out faster, please help :sweat_smile:


raimund-schluessler:
No idea what's wrong here. Probably needs a Mac to debug this.

PR #2562, opened 2022-03-21
Fix secondary button with old style as temporary solution

nickvergessen:

### DIFF
--- a/src/components/Button/Button.vue
+++ b/src/components/Button/Button.vue
@@ -256,12 +256,9 @@ export default {
 		type: {
 			type: String,
 			validator(value) {
-				// Disallowing secondary until we solve theming and
-				// accessibility issues.
-				// https://github.com/nextcloud/nextcloud-vue/issues/2538
-				return ['primary', 'tertiary', 'error', 'warning', 'success'].indexOf(value) !== -1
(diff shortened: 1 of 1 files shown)


src/components/Button/Button.vue
```diff
@@ -256,12 +260,9 @@ export default {
...
 			type: String,
 			validator(value) {
-				// Disallowing secondary until we solve theming and
-				// accessibility issues.
-				// https://github.com/nextcloud/nextcloud-vue/issues/2538
-				return ['primary', 'tertiary', 'error', 'warning', 'success'].indexOf(value) !== -1
+				return ['primary', 'secondary', 'tertiary', 'error', 'warning', 'success'].indexOf(value) !== -1
 			},
-			default: '',
+			required: true,
```
marcoambrosini, line 265: Why not accepting the empty type? The default button shouldn't require specifying a type IMO

marcoambrosini:
Screenshot

Screenshot 2022-03-21 at 09 58 00

src/components/Button/Button.vue
nickvergessen: Because then different classes are applied as discussed last week:
```js
[`button-vue--vue-${this.type}`]: this.type,
```

Or would `default: 'secondary'` work already?

marcoambrosini (approved):

GretaD (approved):

PR #2566, opened 2022-03-22
Use large corner radius for the Popover component

quentinguidee:
This PR *tries* to takes a further step towards the standardization of border radii. Currently, border radii are not uniform. Among other things, buttons have a very high border-radius, while Popovers are only allowed a 3px border-radius, and lists have none.

In the current state of this PR, here are some examples:

| Before | After |
| --- | --- |
| check-1 | check-2 |
| menu-1 | menu-2 |
| large-1 | large-2 |

Another example (done with devtools) of what it looks like in-app:

Capture d’écran 2022-03-22 à 15 49 50

In my opinion, it gives a more modern look to NextCloud.

To improve this PR, we *could* add some vertical padding, but this is open to discussion.

### DIFF
--- a/src/components/Popover/Popover.vue
+++ b/src/components/Popover/Popover.vue
@@ -120,7 +120,8 @@ $arrow-width: 10px;
 	&__inner {
 		padding: 0;
 		color: var(--color-main-text);
-		border-radius: var(--border-radius);
+		border-radius: var(--border-radius-large);
+		overflow: hidden;
 		background: var(--color-main-background);
 	}
 


jancborchardt (commented):
Looks much nicer, and I think @marcoambrosini will like this too. :)

@quentinguidee 2 things:
- The whitespace chars/tabs seems off in your edit
- What is the overflow:hidden for?

quentinguidee:
Without the `overflow: hidden`, when we hover an item, it overflow on the corners (you can even check that behaviour in NC for the moment, the 3px border-radius "comes back to 0" when hovering the first or the last item).

For the whitespace chars/tabs, I'm not sure to understand

marcoambrosini (approved):

JuliaKirschenheuter (approved):

ChristophWurst:
@jancborchardt could you please specify what major versions of Nextcloud should have this styling? E.g Calendar support Nextcloud 24 down to Nextcloud 21. If we upgrade to this change then installations of Nextcloud 21 will also see rounded corners for Calendar, but not for Talk. I'll let you judge if this is a breaking design change.

ChristophWurst:
@marcoambrosini the Talk app only targets one major version of Nextcloud at a time, right? So you wouldn't update `@nextcloud/vue` for the old major version that targets Nextcloud 21, right?

Compare https://github.com/nextcloud/spreed/blob/master/appinfo/info.xml#L55 to https://github.com/nextcloud/calendar/blob/main/appinfo/info.xml#L36.

jancborchardt:
@ChristophWurst seems fine by me to have this in Calendar too, also for lower. Nextcloud versions.

PR #2567, opened 2022-03-22
v5.1.2

marcoambrosini:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "@nextcloud/vue",
-	"version": "5.1.1",
+	"version": "5.1.2",
 	"lockfileVersion": 2,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "@nextcloud/vue",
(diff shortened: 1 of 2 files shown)


artonge (approved):

ChristophWurst:
I see https://github.com/nextcloud/nextcloud-vue/pull/2553 was release in the same fashion and the changelog wasn't complete either. @skjnldsv to the rescue. What should be released off master and what from stable branches? https://github.com/nextcloud/nextcloud-vue#releasing-a-new-version doesn't tell.

nickvergessen:
As far as I understood basically once the first breaking change comes into master we create a stable5 and then it's split

nickvergessen (approved):

ChristophWurst:
Then this should be a minor, shouldn't it? There are changes other than fixes.

marcoambrosini:
Ok, I'll do a minor and include https://github.com/nextcloud/nextcloud-vue/pull/2566#pullrequestreview-918362444

PR #2571, opened 2022-03-23
Use large border-radius in AppNavigation

### LINKED ISSUE #2460: Round background of AppNavigationItem
Time to do it? @jancborchardt @nimishavijay 

https://user-images.githubusercontent.com/26852655/150336491-6018fb53-97c0-43f8-b36f-1161f664861e.mov

quentinguidee:
*Border-radius!* Now for the AppNavigation (closes #2460).

Capture d’écran 2022-03-23 à 16 59 01

## Change(s) to do before merging this PR

- [ ] Rework the AppNavigationItem with children: they have to be outside of the main element to get border-radius working correctly. See why here: https://imgur.com/a/6viebTM


quentinguidee:
And with #2572:

https://user-images.githubusercontent.com/12123721/159807485-b56fedef-c7b6-428a-ac4e-12fb990f50a0.mov

GretaD:
Hey @quentinguidee, thank you for initiating this. Do you need any help to finish it?

quentinguidee:
Hi @GretaD, I don't have much time right now, but from what I had deduced, this PR first requires #2572 to work. But after that feel free to improve this PR!

Also, these two PR are made only using the styleguide. Don't hesitate to check if they work with Nextcloud without breaking anything ^^

GretaD (approved):

GretaD:
What is left now is the Action Buttons, should i add those here or open a new pr?

raimund-schluessler:
> What is left now is the Action Buttons, should i add those here or open a new pr?

I would vote for a separate PR. Makes reviewing and testing easier.

GretaD:
@raimund-schluessler can you please be the second reviewer?

raimund-schluessler (changes requested):
I think this is to much padding. It looks a bit weird to me:
Screenshot 2022-05-30 at 20-37-00 Tasks - Nextcloud

@jancborchardt @nimishavijay 

Besides the that, it works as expected.

src/components/AppNavigation/AppNavigation.vue
```diff
@@ -146,6 +146,7 @@ export default {
 	position: -webkit-sticky;
 	top: $header-height;
 	left: 0;
+	padding: 8px 12px;
```
raimund-schluessler, line 149: I would remove this padding altogether, so it's in line with e.g. the `ListItem` and how it's used in Mail.
```suggestion
```
And especially in horizontal direction, the padding uses a lot of space (nearly 10% of the available `AppNavigation`s horizontal space).

jancborchardt:
Agree the padding on right and left is a tad much, but it also should not be none.

How much is it in Talk, 4px @marcoambrosini? We should do the same, and the padding on the top of the list should be same as on sides.

raimund-schluessler:
> Agree the padding on right and left is a tad much, but it also should not be none.
> 
> How much is it in Talk, 4px @marcoambrosini? We should do the same, and the padding on the top of the list should be same as on sides.

4px sounds good. We use the same in the Actions dropdown now.

marcoambrosini:
Yes that sounds good. Mail is not a good benchmark here because the hover and active feedbacks of the listitem actually touches the edge, and that should never happen. Basically we should avoid visual tangents at all costs, as they create tension and discomfort. 

Screenshot 2022-05-31 at 16 30 49

raimund-schluessler:
I rebased the branch and adjusted the padding. This is how it looks now (with the new Action menu as reference):

Screenshot 2022-05-31 at 23-06-30 Tasks - Nextcloud

raimund-schluessler (approved):

marcoambrosini (approved):
very nice!

jancborchardt:
Is there a related pull request in Files for this? @PVince81 @skjnldsv or does it use the Vue components already?

PR #2572, opened 2022-03-23
Put the children of AppNavigationItem outside of the main item

quentinguidee:
This change is necessary for #2571 (see why in action: https://github.com/nextcloud/nextcloud-vue/pull/2571#issuecomment-1076889150). It puts AppNavigationItem's children out of the parent item, and displays them separately. This is the simplest implementation I found without recoding the entire AppNavigationItem. I'm open to ideas if you find something cleaner.

https://user-images.githubusercontent.com/12123721/159804992-fed21b82-7581-4f07-9ead-05b3c2ec4067.mov

https://user-images.githubusercontent.com/12123721/159805040-823d0330-ad3a-48dc-a3f9-0dd61bcd79fb.mov

## Important notes/questions to reviewers

* In the current implementation, a side effect is that it disables the ability to pass (like in the doc) `class="active"` directly to the AppNavigationItem. It seems like the only purpose of passing the `class="active"` directly is only for the example? The `isActive` property works fine and continues to push the blue background.

* As another side effect of this PR, when hovering a child, it doesn't display the hover background of the parent (as demonstrated in the screenshots). I can go back to the first one easily if needed. Just unsure if it was a bug or not.

### DIFF
--- a/src/components/AppNavigationItem/AppNavigationItem.vue
+++ b/src/components/AppNavigationItem/AppNavigationItem.vue
@@ -81,7 +81,7 @@ prevent the user from collapsing the items.
 <AppNavigationItem title="Item with children" :allowCollapse="true" :open="true">
 	<template>
 		<AppNavigationItem title="AppNavigationItemChild1" />
-		<AppNavigationItem class="active" title="AppNavigationItemChild2" />
+		<AppNavigationItem title="AppNavigationItemChild2" />
 		<AppNavigationItem title="AppNavigationItemChild3" />
 		<AppNavigationItem title="AppNavigationItemChild4" />
 	</template>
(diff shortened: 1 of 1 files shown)


raimund-schluessler (dismissed):
In principle, I like this idea (the grey hover background of the parent on the left of the child always bothered me), but the changes currently create invalid HTML.

See here:
AppNavigationItem

The `li` tag has no `ul` parent (which it has to have) and one of it's siblings is an `ul`, which is not allowed, see https://html.spec.whatwg.org/#the-ul-element.

So please change it so that `li`s always have an `ul` parent and only `li` siblings.

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -111,97 +111,101 @@ Just set the `pinned` prop.
...
-			@click="onClick">
-
-			<!-- icon if not collapsible -->
-			<!-- never show the icon over the collapsible if mobile -->
-			<div :class="{ 'icon-loading-small': loading, [icon]: icon && isIconShown }"
-				class="app-navigation-entry-icon">
-				<slot v-show="!loading && isIconShown" name="icon" />
+		}">
+		<!-- Navigation item, can be either an <li> or a <router-link> depending on the props -->
+		<nav-element v-bind="navElement"
```
raimund-schluessler, line 120: This needs an `ul` parent, since it's always a `li`

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -111,97 +111,101 @@ Just set the `pinned` prop.
...
+					<slot name="actions" />
+				</Actions>
+			</div>
+
+			<!-- Anything (virtual) that should be mounted in the component, like a related modal -->
+			<slot name="extra" />
+		</nav-element>
 
 		<!-- Children elements -->
 		<ul v-if="canHaveChildren && hasChildren" class="app-navigation-entry__children">
```
raimund-schluessler, line 205: This needs to be wrapped inside a `li`.

src/components/AppNavigationItem/AppNavigationItem.vue
GretaD: https://github.com/nextcloud/nextcloud-vue/blob/d01fcc7402af9f3f773dd575b61caa6e108be6ab/src/components/AppNavigation/AppNavigation.vue#L65-L68

that navigation item always appears on this slot, so there is an ul around it, right?
Sorry, im not sure, what exactly are you proposing to change..

raimund-schluessler: Look at the screenshot of the rendered HTML markup above. The `AppNavigationItem` component does not render a `li` anymore, but now gives `element` as its root element. This is not a valid child of `ul` rendered by `AppNavigation`. So the `AppNavigationItem` has to return a `li` element as its root element.

So, I basically just request valid markup (`li`s always need `ul` parents, and `ul` only have `li`s as children), which the current implementation does not give.

GretaD:
i have only rebased it so far, till i understand a bit better Raimunds perspective :)

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -117,7 +117,7 @@ Just set the `pinned` prop.
 			'app-navigation-entry--collapsible': collapsible,
 		}">
 		<!-- Navigation item, can be either an <li> or a <router-link> depending on the props -->
-		<nav-element v-bind="navElement"
+		<div v-bind="navElement"
```
raimund-schluessler, line 120: I would leave it as `nav-element` to keep it obvious that it's for navigation. You can make it render as a div via the `navElement` object.

GretaD: and have it both 'div'?

raimund-schluessler: I would think so, yes.

GretaD:
i finally managed to link my nc/vue with an app and tested this on mail and it doesnt work :(
I am trying to fix it but if anybody already have ideas, please write me here

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -400,18 +402,17 @@ export default {
 			}
 		},
 		// This is used to decide which outer element type to use
-		// li or router-link
 		navElement() {
 			if (this.to) {
 				return {
-					is: 'router-link',
-					tag: 'li',
+					is: 'div',
```
raimund-schluessler, line 408: This needs to stay router-link.
```suggestion
					is: 'router-link',
```

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -111,109 +111,111 @@ Just set the `pinned` prop.
...
+		</nav-element>
 		<!-- Children elements -->
 		<ul v-if="canHaveChildren && hasChildren" class="app-navigation-entry__children">
 			<slot />
 		</ul>
-
-		<!-- Anything (virtual) that should be mounted in the component, like a related modal -->
-		<slot name="extra" />
-	</nav-element>
+				</li>
```
raimund-schluessler, line 203: Indentation.
```suggestion
	</li>
```

GretaD:
yes, it works, many thanks @raimund-schluessler

GretaD (approved):

GretaD:
@raimund-schluessler can you pleas sign off your commits? DCO is not passing

raimund-schluessler:
> @raimund-schluessler can you please sign off your commits? DCO is not passing

Just squash them into yours. My changes were negligible (and I only have the webinterface at hand right now).

GretaD:
> > @raimund-schluessler can you please sign off your commits? DCO is not passing
> 
> Just squash them into yours. My changes were negligible (and I only have the webinterface at hand right now).

yes, right, i can do that

raimund-schluessler:
There is some more (CSS) stuff that needs fixing/adjustments. I will have a look and push some commits, if that is ok with you @GretaD.

raimund-schluessler:
There seems to be an issue with the netlify docs when the branch is from a different repository. I have no idea why, but opening #2704 triggered the docs build and now the docs preview links to the build for #2704.

I am hence closing this PR #2572 in favor of #2704. I hope that's fine with everyone.

PR #2576, opened 2022-03-24
Increase padding of ColorPicker

quentinguidee:
Necessary since Popover component got border-radius. Otherwise it feels a bit tight

| Before | After |
| --- | --- |
| cp-before- | cp-after- |
| cp-before | cp-after |


raimund-schluessler (approved):

nimishavijay (approved):
Looks good 👍🏽 unrelated but if possible we could also reduce the space above the 3 dot menu by a few pixels

marcoambrosini (approved):

PR #2583, opened 2022-03-28
Fix tertiary buttons hover feedback

marcoambrosini:
Add proper hover feedback to tertiary buttons whose icon aren't affected by the color css rule. e.g. emojis

https://user-images.githubusercontent.com/26852655/160430239-a1436f2d-9059-4b5d-821c-371f4d9bea41.mov


GretaD (approved):
code looks good, didnt test

PR #2587, opened 2022-03-31
Extend AppNavigationCaption with slot for custom actions icon

JuliaKirschenheuter:
Required for https://github.com/nextcloud/calendar/issues/3965

### DIFF
--- a/src/components/AppNavigationCaption/AppNavigationCaption.vue
+++ b/src/components/AppNavigationCaption/AppNavigationCaption.vue
@@ -7,6 +7,34 @@
 	</template>
 </AppNavigationCaption>
 ```
+
+### Element with a slot for custom actions icon
+```
+<template>
+	<AppNavigationCaption
+		title="Your caption goes here">
+		<template #actionsTriggerIcon>
+			<PlusIcon slot="icon" :size="20" />
+		</template>
+		<template #actions>
+			<ActionButton icon="icon-edit">Rename</ActionButton>
+			<ActionButton icon="icon-delete">Delete</ActionButton>
+			<ActionButton icon="icon-confirm">Validate</ActionButton>
+			<ActionButton icon="icon-download">Download</ActionButton>
+		</template>
+	</AppNavigationCaption>
+</template>
+<script>
+	import PlusIcon from 'vue-material-design-icons/Plus'
+
+	export default {
+		components: {
+			PlusIcon
+		}
+	}
+</script>
+```
+
 </docs>
 
 <template>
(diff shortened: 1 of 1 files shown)


marcoambrosini (commented):
It doesn't seem to work in the docs, or maybe I'm missing something?

There shouldn't be the need to modify the component to pass in a mdi icon into the slot. it should work with

```
<template #actions>
    <Plus #icon /> // This icon will be in the trigger of the action menu instead of the 3 dots
    <ActionButton />
    .... all other action elements
<template>
```

See https://deploy-preview-2587--nextcloud-vue-components.netlify.app/#/Components/Actions

JuliaKirschenheuter:
It should be right now. Please review ;)

src/components/AppNavigationCaption/AppNavigationCaption.vue
```diff
@@ -7,6 +7,23 @@
 	</template>
 </AppNavigationCaption>
 ```
+
+### Element with a slot for custom actions icon
+```
+<AppNavigationCaption
+	title="Your caption goes here">
+	<template #actionsTriggerIcon>
+		<PlusIcon slot="icon" :size="20" />
```
raimund-schluessler, line 16: The `PlusIcon` component is missing and not showing. You need to import it. Have a look at e.g. https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/Actions/Actions.vue#L88-L122 for how to.

JuliaKirschenheuter: ok, something went wrong =(
actually i thought, that it is already imported.
Thank you!

raimund-schluessler:
Works as it should now.

But I have to admit, I am a bit puzzled why the proposal of @marcoambrosini does not (I checked it in the docs now, and the icon does not render):
> ```
> <template #actions>
>     <Plus #icon /> // This icon will be in the trigger of the action menu instead of the 3 dots
>     <ActionButton />
>     .... all other action elements
> <template>
> ```

JuliaKirschenheuter:
> Works as it should now.
> 
> But I have to admit, I am a bit puzzled why the proposal of @marcoambrosini does not (I checked it in the docs now, and the icon does not render):
> 
> > ```
> > <template #actions>
> >     <Plus #icon /> // This icon will be in the trigger of the action menu instead of the 3 dots
> >     <ActionButton />
> >     .... all other action elements
> > <template>
> > ```

Because i need to put somehow custom icon slot to Actions via ```<template #icon>```.  There were no possibility at the AppNavigationCaption before.

PR #2588, opened 2022-03-31
v5.3.0

marcoambrosini:


GretaD (approved):
because yolo

nickvergessen (changes requested):
The ter… buttons look wrong. There is a too big difference with and without the background
https://nextcloud-vue-components.netlify.app/#/Button

PR #2600, opened 2022-04-04
Migrate to floating-vue

raimund-schluessler:
This migrates the popover library from `v-tooltip` to its successor `floating-vue`. There are no visual changes intended, so I don't provide any before/after screenshots.

The functions directly affected in the library are
- `Actions` component
- `EmojiPicker` component
- `ColorPicker` component
- `DatetimePicker` component
- `UserBubble` component
- `Popover` component
- `Avatar` component
- `Tooltip` directive

I checked that they still work as I think they should, but please check as well.

With the migration there come some breaking changes listed here: https://floating-vue.starpad.dev/migration/migration-from-v2.html#directive The most important one is the deprecation/renaming of some props, e.g.:
- `trigger` to `triggers`
- `open` to `shown`
- `boundariesElement` to `boundary`
- `classes` to `popperClass`

This is important to everyone using the `Popover` component directly, e.g. done in Talk @nickvergessen


marcoambrosini:
you rock @raimund-schluessler :)

PVince81:
to retest: avatar menu in Talk conversation which uses Popover with heavy workarounds to make sure it gets positioned properly despite the scroll and tooltip delay

marcoambrosini (approved):
Tested in the docs and works. I'd say let's merge and release v6.0.0-beta.0 so that we can upgrade and fix the breaking changes in the apps

juliushaertl (approved):

src/directives/Tooltip/index.js
```diff
@@ -20,15 +20,12 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  *
  */
+import { VTooltip, options } from 'floating-vue'
```
nickvergessen, line 23: So in talk we were setting the default container:
```js
Tooltip.options.defaultContainer = '#content-vue'
```

But with floating-vue the options are not on the exported object anymore.
So can we also export the options Object?

PVince81: background for defaultContainer: to make sure the tooltips stay visible even in fullscreen mode

PR #2604, opened 2022-04-06
closes#2569, changed fixed 4px to .38rem height.

Zackaria-Slimane:
The issue seems to be only in some browsers ( safari has it, firefox was fine) ,changed from fixed pixels to rem unit , displays the same accros the different browsers

### DIFF
--- a/src/components/ProgressBar/ProgressBar.vue
+++ b/src/components/ProgressBar/ProgressBar.vue
@@ -122,7 +122,8 @@ export default {
 	}
 	&::-webkit-progress-value {
 		background: linear-gradient(40deg, var(--color-primary-element) 0%, var(--color-primary-element-light) 100%);
-		border-radius: var(--border-radius);
+		border-radius: 2px:
+		height:.38rem
 	}
 	&::-moz-progress-bar {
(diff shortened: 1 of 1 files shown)


raimund-schluessler:
This does not fix it in Chrome for me.
This does:
```css
.progress-bar[value]::-webkit-progress-value {
    border-radius: 2px;
    height: 4px;
}
```

raimund-schluessler:
> This does not fix it in Chrome for me. This does:
> 
> ```css
> .progress-bar[value]::-webkit-progress-value {
>     border-radius: 2px;
>     height: 4px;
> }
> ```

This means the CSS here needs to be adjusted depending on the prop:
https://github.com/nextcloud/nextcloud-vue/blob/593bc9d7f0be0b87eae2906844ea67b9d9c7b0ac/src/components/ProgressBar/ProgressBar.vue#L120-L126

raimund-schluessler:
@Zackaria-Slimane I had an idea how to fix this nicely with a lot less code and implemented it in #2618. I hope you don't mind that I close this one here in favor of #2618.

PR #2610, opened 2022-04-06
don't give normal and large modals a fixed height

szaimen:
Fix https://github.com/nextcloud/server/issues/31854

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -735,7 +735,6 @@ export default {
 			max-width: 90%;
 			width: 600px;
 			max-height: 90%;
-			height: 600px;
 		}
 	}
 	&--large {
@@ -743,7 +742,6 @@ export default {
 			max-width: 90%;
 			width: 900px;
 			max-height: 90%;
-			height: 700px;
 		}
 	}
 	&--full {


src/components/Modal/Modal.vue
```diff
@@ -735,15 +735,13 @@ export default {
 			max-width: 90%;
 			width: 600px;
 			max-height: 90%;
-			height: 600px;
```
szaimen, line 738: maybe we should give it a min-heigt of e.g. 300px now that we don't enforce a specific size anymore?

szaimen:
and while we are at it, what do you think about changing the width and max-width here:
https://github.com/nextcloud/nextcloud-vue/blob/dc03f8d5db6a24309145c16c72360dd653a551c6/src/components/Modal/Modal.vue#L728-L730
to 
```
width: 400px;
max-width: 90%; 
```

marcoambrosini (approved):

szaimen:
@marcoambrosini removing the fixed height should just fall-back to the height of the elements inside (aka `fit-content`) so not sure if this is really a breaking change. But either way fine by me.

marcoambrosini:
🤔 couldn't it be that someone is relying on that height for scrollable content?

szaimen:
> 🤔 couldn't it be that someone is relying on that height for scrollable content?

I don't think so...
Also it is the wrong way to do that, imo... you should rather let the modal handle this...

dartcafe:
> 🤔 couldn't it be that someone is relying on that height for scrollable content?

I woud say the opposite is the case. I was reying on the fit-content-height and had to fix the forced height in every modal to get the behavior before this change was introduced.

src/components/Modal/Modal.vue
dartcafe: What about a `width: clamp(300px, 600px, 90vw);`?

szaimen:
So merge then or what are we waiting for?

(the initial problem should be fixed with this change)

src/components/Modal/Modal.vue
dartcafe: Yeah right, off topic. I just stumbled upon this definition, because I prefer `clamb`, `min `and `max `over `min-width`, `max-width`, ... .

skjnldsv (changes requested):
I'd like to wait for a minor, we need to release a patch first :pray:

skjnldsv (approved):
Actually it's a patch as it's a bugfix, sorry for misunderstanding

src/components/Modal/Modal.vue
szaimen: okay, feel free to open a PR for discussion. For me using min, max, etc. is fine :)

dartcafe: Ah, no. That's not valuable, because the result is the same. Just consumes time. For me this are just small  'by the way' improvements of the css code. I am more satified that I can remove my height hack after the release.

szaimen: okay 👍

PR #2616, opened 2022-04-07
Remove unused dependency vue-visible

raimund-schluessler:
This removes `vue-visible` which was not used.


ChristophWurst (approved):
so it's *invisble*? :drum:

juliushaertl (approved):

PR #2626, opened 2022-04-08
Allow put custom label to ActionRadio.vue

JuliaKirschenheuter:
resolves  https://github.com/nextcloud/mail/issues/1626

This is needed for extended labels as https://github.com/nextcloud/mail/issues/1626#issuecomment-1087526103

### DIFF
--- a/src/components/ActionRadio/ActionRadio.vue
+++ b/src/components/ActionRadio/ActionRadio.vue
@@ -50,22 +50,19 @@ So that only one of each name set can be selected at the same time.
 				class="radio action-radio__radio"
 				@keydown.enter.exact.prevent="toggleInput"
 				@change="onChange">
-			<label ref="label" :for="id" class="action-radio__label">{{ text }}</label>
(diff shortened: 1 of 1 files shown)


miaulalala (approved):

ChristophWurst:
I understand that this is a higher impact change but I don't see why it's bad per se.

Right now you can only use pure *text* for the label and no more specific markup.

The proposed alternative is to use a combination of action radio and action text. The difference I see is that with the alternative the actions are not *grouped* so that clicking the subline would also automatically select the radio. Only clicking the radio label selects the option.

I'm well aware that using a slot gives developers possible ways to create markup that isn't valid anymore.

Is there any other downside?

skjnldsv:
A radio label can only contains text, anything else is not valid for html.
This has a much higher impact. We use the double-action in a various other places and it works fine.
This is not a bad idea per se, but needs a much broader discussion regarding how we view the Actions in general and how do we make sure the component stays compliant to html/accessibility standards as well as NC design standards :)

PR #2628, opened 2022-04-09
Fix styleguide serve

raimund-schluessler:
Due to a problem of vue-styleguidist with webpack 5, we have to manually work around if for now. Fix is from https://github.com/vue-styleguidist/vue-styleguidist/pull/1308 and can be removed once the upstream fix is merged and released.


skjnldsv (approved):

skjnldsv:
Didn't even think of this! :grin:

PR #2629, opened 2022-04-09
small cleanup of small modal size

szaimen:

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -725,8 +725,8 @@ export default {
 	// Sizing
 	&--small {
 		.modal-container {
-			width: 390px;
-			max-width: 95%;
+			width: 400px;
+			max-width: 90%;
 			max-height: 90%;
 		}
 	}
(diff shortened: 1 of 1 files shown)


skjnldsv:
Could you explain the reasoning behind this?

src/components/Modal/Modal.vue
```diff
@@ -725,8 +725,8 @@ export default {
 	// Sizing
 	&--small {
 		.modal-container {
-			width: 390px;
-			max-width: 95%;
+			width: 400px;
+			max-width: 90%;
```
szaimen, line 729: Changing 390px to 400px is just for the sake of using a x00 number here instead of a random feeling number. Changing 95% to 90% is in order to align with the other used max values of mostly 90%.

src/components/Modal/Modal.vue
```diff
@@ -756,7 +756,7 @@ export default {
 
 	// Make modal full screen on mobile
 	@media only screen and (max-width: $breakpoint-mobile / 2) {
-		&:not(&--small) .modal-container {
+		.modal-container {
```
szaimen, line 759: Changing this media query to also addressing the small modal has the reason that I designed the small modal initially to be used e.g. for prompts with a limited height. In reality it gets used in many places mostly with mostly a pretty big height. Adjusting this makes sure that the mobile view is aligned to the other sizes and also looks better, imo.

szaimen:
> Could you explain the reasoning behind this?

done

skjnldsv (approved):

artonge (approved):

PR #2630, opened 2022-04-09
show the action buttons on top of the modal always by default

szaimen:
reason behind this: as noticed in the accessibility workshop, the action buttons of the modal should always visible. Changing the default to -1 will allow to opt-in into the hiding of the action buttons which is in my opinion the better way to do this.

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -107,17 +107,10 @@
 		<div ref="mask"
 			class="modal-mask"
 			:class="{ 'modal-mask--dark': dark }"
-			:style="cssVariables"
-			@click="handleMouseMove"
-			@mousemove="handleMouseMove"
-			@touchmove="handleMouseMove">
+			:style="cssVariables">
 			<!-- Header -->
 			<transition name="fade-visibility">
-				<div v-show="!clearView"
-					:class="{
-						invisible: clearView
-					}"
-					class="modal-header">
+				<div class="modal-header">
 					<div v-if="title.trim() !== ''" class="modal-title">
 						{{ title }}
 					</div>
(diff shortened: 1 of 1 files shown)


skjnldsv:
This is a breaking change @jancborchardt

jancborchardt:
@szaimen could you add before/after screenshot / gif? I don’t really understand what it does.

@skjnldsv why is this a breaking change?

szaimen:
> @szaimen could you add before/after screenshot / gif? I don’t really understand what it does.

Maybe @skjnldsv can? :)

szaimen:
@jancborchardt these navigation items:
image
hide by default after 3s if you don't move your mouse which is bad for mobile and accessibility wise.

After this change they are always visible by default if the developer did not specify a timeout which will hide these items again after a specific timeout...

Hope that is clear now? :)

jancborchardt:
Which of our modals would still use auto-hide here then, if any?

Cause as it came up in the accessibility workshop I would say we should remove the auto-hiding altogether. The close button should always be visible, and if a name is needed that one should also be shown (or not set in the first place), but not hide.

This should not be configurable by individual developers.

szaimen:
> This should not be configurable by individual developers.

So then shall I remove this option?

cc @skjnldsv

szaimen:
ping @skjnldsv 
for me it is fine to only change the default for now. WDYT? :)

skjnldsv:
I'm following Jan's decision :)
If the workshop suggested we should remove, let's set the default to disabled or remove the feature

ChristophWurst (approved):
:+1: :elephant: 

please squash

JuliaKirschenheuter (approved):

PR #2631, opened 2022-04-09
Migrate from node-sass to sass

raimund-schluessler:
This PR migrates from `node-sass` (which is deprecated) to `sass` and fixes the deprecation warnings.


skjnldsv (approved):
You're amazing! :cupid:

marcoambrosini (approved):

PR #2634, opened 2022-04-11
Add unit test for popover component

raimund-schluessler:
In preparation of the migration from `v-tooltip` to `floating-vue` in #2600, I wanted to write some unit tests for the affected components, such as `Popover`.

But no matter what I tried, I cannot get this popover to open, neither by default, clicking, or setting the prop after mounting. This PR contains some tests that should pass, but instead fail, because the popover does not open.

@nextcloud/vuejs If someone has an idea, what I am doing wrong here, I would be very grateful!

### DIFF
--- a/tests/unit/components/Popover/Popover.spec.js
+++ b/tests/unit/components/Popover/Popover.spec.js
@@ -1,0 +1,86 @@
+/**
+ * @copyright Copyright (c) 2022 Raimund Schlüßler <raimund.schluessler@mailbox.org>
+ *
+ * @author Raimund Schlüßler <raimund.schluessler@mailbox.org>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


skjnldsv:
> But no matter what I tried, I cannot get this popover to open, neither by default, clicking, or setting the prop after mounting

Maybe because its mountpoint is in body ?

tests/unit/components/Popover/Popover.spec.js
```diff
@@ -42,7 +42,11 @@ describe('Popover.vue', () => {
 			expect(wrapper.find('.content').isVisible()).toBe(false)
 		})
 		it('is open when prop is set', async () => {
+			const div = document.createElement('div')
+			div.id = 'root'
+			document.body.appendChild(div)
 			const wrapper = mount(Popover, {
+				attachTo: '#root',
```
marcoambrosini, line 49: shouldn't this be done via the `container` prop?

raimund-schluessler: `attachTo` is an option from @vue\test-utils -> mount function, see https://v1.test-utils.vuejs.org/api/options.html#attachto It specifies to which HTMLElement the component will be mounted to.
`container` is an option of the `Popover` component itself, specifying where the popover menu will be attached to. The popover should open no matter the `container` prop (but it also doesn't open if I specify it).
I am a bit lost to be honest.

marcoambrosini: 🤔

PR #2594, opened 2022-04-01
Upgrade @nextcloud/eslint-config to 7.0.2

raimund-schluessler:
This updates @nextcloud/eslint-config to the latest version and fixes most lint issues introduced by it.

PR #2640, opened 2022-04-17
Run eslint with webpack for vue files

raimund-schluessler:
This brings back running eslint on vue files when webpack runs (dev, watch, build) which got lost with #2594.

Before:
See that eslint only reports lint problems in js files when e.g. `npm run watch' runs.

After:
See that eslint now also reports lint problems in vue files when e.g. `npm run watch' runs.

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -149,7 +149,9 @@ module.exports = {
 		new StyleLintPlugin({
 			files: ['src/**/*.vue', 'src/**/*.scss', 'src/**/*.css'],
 		}),
-		new ESLintPlugin(),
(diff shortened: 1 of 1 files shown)


marcoambrosini (approved):

raimund-schluessler:
> So, we actually are moving away from this :) To have those as a separate step [nextcloud/webpack-vue-config#318](https://github.com/nextcloud/webpack-vue-config/pull/318)

Hm, ok. I just found it very helpful with fixing many linting errors in #2597 to immediately only see the ones I didn't fix yet. But I guess that's an edge case. Closing then.

PR #2642, opened 2022-04-19
Update the css for the datepicker

newhinton:
Also see nextcloud/calendar/pull/4105 for the discussion.

In this pr i fixed the missing margin on the buttons. I also widened the popup to fit all the controls aswell as made the text in those buttons centered.
Also i added color to the confirm button to match the 'primary' style of nextcloud. 

Here is an example of the fixes:
grafik

### DIFF
--- a/src/components/DatetimePicker/DatetimePicker.vue
+++ b/src/components/DatetimePicker/DatetimePicker.vue
@@ -49,6 +49,28 @@ export default {
 </script>
 ```
 
+### Example with confirm button
+```vue
+<template>
+	<span>
+		<DatetimePicker
+			v-model="time"
+			type="datetime"
+			confirm />
+		{{ time }}
+	</span>
+</template>
+<script>
+	export default {
+		data() {
+			return {
+				time: null,
+			}
+		},
+	}
+</script>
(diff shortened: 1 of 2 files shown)


newhinton:
I would also like to add the 'confirm' attribute to the documentation, where would i need to do that?

szaimen (approved):
thanks, lgtm! :)

szaimen:
> I would also like to add the 'confirm' attribute to the documentation, where would i need to do that?

You may simply add the confirm attribute here?
https://github.com/nextcloud/nextcloud-vue/blob/d806d43e6b70364e740d8a00d71df46b547b4a7a/src/components/DatetimePicker/DatetimePicker.vue#L37

nimishavijay (approved):
Looks great design wise! :)

skjnldsv (approved):

newhinton:
I have simply added a new section with the confirm button to the documentation now. I changed the tag back to review for that, albeit i think its fine

JuliaKirschenheuter (approved):

raimund-schluessler (approved):

PR #2654, opened 2022-04-25
add handle keyboard focus for accessibility

vanpertsch:
add focus-trap (https://github.com/focus-trap/focus-trap) for accessibility

https://user-images.githubusercontent.com/38529028/165922863-7cb781d4-2b3c-4a93-9c5f-06d6b26ba6c6.mp4

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -22,40 +22,84 @@
 
 <docs>
 
-```vue
-<template>
-	<div>
-		<button @click="showModal">Show Modal</button>
-		<modal v-if="modal" @close="closeModal" size="small">
-			<div class="modal__content">Hello world</div>
-		</modal>
-	</div>
-</template>
-<style scoped>
+	```vue
+	<template>
+		<div>
+			<button @click="showModal">Show Modal</button>
+			<modal v-if="modal" @close="closeModal" size="small">
+				<div class="modal__content">Hello world</div>
+			</modal>
+		</div>
+	</template>
+	<style scoped>
 	.modal__content {
 		margin: 50px;
 		text-align: center;
 	}
 
-</style>
-<script>
-export default {
-	data() {
-		return {
-			modal: false
-		}
-	},
-	methods: {
-		showModal() {
-			this.modal = true
+	</style>
+	<script>
+	export default {
+		data() {
+			return {
+				modal: false
+			}
 		},
-		closeModal() {
-			this.modal = false
+		methods: {
+			showModal() {
+				this.modal = true
+			},
+			closeModal() {
+				this.modal = false
+			}
 		}
 	}
-}
-</script>
-```
+	</script>
+	```
+
+	### Usage of popover in modal
+
+	* Set container property to .modal-mask to inject popover context of the modal:
+
+	```vue
+	<template>
+		<div>
+			<button @click="showModal">Show Modal</button>
+			<modal v-if="modal" @close="closeModal" size="small">
+				<EmojiPicker container=".modal-mask" @select="select">
+					<button>Select emoji {{ emoji }}</button>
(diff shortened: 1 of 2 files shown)


src/components/Modal/Modal.vue
```diff
@@ -525,6 +527,59 @@ export default {
...
+					firstFocusable.focus()
+				}
+			})
+		},
+		/**
+		 * set focus trap
+		 *
+		 * @param e
+		 */
+		checkForTabKeyEvent(e) {
```
PVince81, line 558: did you know about the "inert" attribute ? https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert

it's not supported yet (or experimentally) by browsers but there's a polyfill for it

vanpertsch: @PVince81 thanks. I thought maybe [https://github.com/focus-trap/focus-trap](https://github.com/focus-trap/focus-trap) is a good solution.

src/components/Modal/Modal.vue
```diff
@@ -525,6 +527,32 @@ export default {
 				this.slideshowTimeout.clear()
 			}
 		},
+		/**
+		 * Get List of focusableelmets from content container
+		 */
+		getFocusableList() {
+			const contentContainer = this.$refs.mask.querySelector('.modal-container')
+			const focusableList = contentContainer.querySelectorAll(
```
PVince81, line 535: I think now you don't need to query all, just the first element ?
basically rename the function to "getFirstFocusableElement" ? (assuming the focus trap doesn't do this already automatically)

vanpertsch: ...i do not longer need  getFocusableList at all. forgot to remove it.

skjnldsv (approved):

src/components/Modal/Modal.vue
```diff
@@ -525,6 +571,24 @@ export default {
 				this.slideshowTimeout.clear()
 			}
 		},
+		/**
+		 * Add focus trap for accessibility.
+		 */
+		useFocusTrap() {
+			const contentContainer = this.$refs.mask
+			this.$nextTick(function() {
```
artonge, line 579: Do you know why it wasn't working without it? Could be nice to add a comment :)

src/components/Modal/Modal.vue
```diff
@@ -525,6 +571,24 @@ export default {
...
+				this.focusTrap = createFocusTrap(contentContainer, {
+					onActivate: () => contentContainer.classList.add('is-active'),
+					onDeactivate: () => contentContainer.classList.remove('is-active'),
```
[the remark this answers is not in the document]
vanpertsch: You are right, we don not need it.

src/components/Modal/Modal.vue
vanpertsch: added

artonge (approved):
:+1:

szaimen:
Does it also work with multiple modals on top of each other e.g. like the text app uses?

skjnldsv:
> Does it also work with multiple modals on top of each other e.g. like the text app uses?

Tested, works fine
```vue
<template>
	<div>
		<button @click="modal = true">Show Modal</button>
		<modal v-if="modal" @close="modal = false" size="small">
			<div class="modal__content">Hello world</div>

			<button @click="modal2 = true">Show Modal 2</button>
			<modal v-if="modal2" @close="modal2 = false" size="small">
				<div class="modal__content">Hello world 2</div>
			</modal>

		</modal>
	</div>
</template>
<style scoped>    
	.modal__content {
		margin: 50px;
		text-align: center;
	}

</style>
<script>
export default {
	data() {
		return {
			modal: false,
			modal2: false
		}
	},
}
</script>
</script>
```

PR #2689, opened 2022-05-13
📝 add Material Design Icons Example

vinicius73:
Just a small example to be used as reference.
Ref: https://github.com/nextcloud/server/issues/32060


skjnldsv:
But we do already? :thinking: 
image

vinicius73:
> But we do already?

My bad, I didn't find it 😱

PR #2695, opened 2022-05-18
Change Prev/next links to buttons

vanpertsch:

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -155,14 +155,14 @@
 						</Actions>
 
 						<!-- Close modal -->
-						<Actions v-if="canClose && !closeButtonContained" class="header-close">
-							<ActionButton @click="close">
-								<template #icon>
-									<Close :size="iconSize" />
-								</template>
-								{{ t('Close') }}
-							</ActionButton>
-						</Actions>
+						<ButtonVue v-if="canClose && !closeButtonContained"
+								type="tertiary"
+								class="header-close"
+								:aria-label="t('Close')">
+							<template #icon>
+								<Close :size="iconSize" />
+							</template>
+						</ButtonVue>
 					</div>
 				</div>
 			</transition>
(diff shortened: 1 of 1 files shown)


src/components/Modal/Modal.vue
```diff
@@ -189,20 +189,19 @@
...
 							}"
 							@click.prevent.stop="previous">
-							<span class="icon-previous">
-								<ChevronLeft :size="40" title="" decorative />
-								<span class="hidden-visually">
-									{{ t('Previous') }}
-								</span>
-							</span>
-						</a>
+							<ActionButton aria-label="Previous item">
```
raimund-schluessler, line 198: I would propose to use an icon-only `Button` instead of `Actions` + `ActionButton`, see https://nextcloud-vue-components.netlify.app/#/Components/Button.

nickvergessen: Fixed

src/components/Modal/Modal.vue
```diff
@@ -155,14 +155,14 @@
...
 						<!-- Close modal -->
-						<Actions v-if="canClose && !closeButtonContained" class="header-close">
-							<ActionButton @click="close">
-								<template #icon>
-									<Close :size="iconSize" />
-								</template>
-								{{ t('Close') }}
-							</ActionButton>
-						</Actions>
+						<ButtonVue v-if="canClose && !closeButtonContained"
```
Pytal, line 158:
```suggestion
						<Button v-if="canClose && !closeButtonContained"
```
**suggestion (non-blocking):** Rename ButtonVue to Button where it's used

Pytal (approved):

src/components/Modal/Modal.vue
nickvergessen: It was used like this before. Don't know why.

marcoambrosini: When there are regular `button` elements the names conflict, so we import the cue button as ButtonVue

marcoambrosini (approved):

src/components/Modal/Modal.vue
```diff
@@ -155,14 +155,14 @@
 						</Actions>
 
 						<!-- Close modal -->
-						<Actions v-if="canClose && !closeButtonContained" class="header-close">
-							<ActionButton @click="close">
```
skjnldsv, line 159: Regression, fix in https://github.com/nextcloud/nextcloud-vue/pull/3066

PR #2696, opened 2022-05-18
add role & aria-labels for accessibility

vanpertsch:

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -107,11 +107,16 @@
 		<div ref="mask"
 			class="modal-mask"
 			:class="{ 'modal-mask--dark': dark }"
-			:style="cssVariables">
+			:style="cssVariables"
+			role="dialog"
+			aria-labelledby="modal-title"
+			aria-describedby="modal-description">
 			<!-- Header -->
 			<transition name="fade-visibility">
 				<div class="modal-header">
-					<div v-if="title.trim() !== ''" class="modal-title">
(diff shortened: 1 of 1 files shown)


PVince81:
@artonge @Pytal please review / take over to get this merged (needs rebase)

nickvergessen:
Ready to review

Pytal (approved):

src/components/Modal/Modal.vue
```diff
@@ -195,7 +200,7 @@
 					</transition>
 
 					<!-- Content -->
-					<div class="modal-container">
+					<div id="modal-description" class="modal-container">
```
vinicius73, line 203: Can we use a dynamic id here?

vinicius73 (approved):

nickvergessen:
There is still no stable5 branch, but i guess we should create it from 5.3.x

juliushaertl:
There is now, see #2907

nickvergessen:
I'll backport it manually

raimund-schluessler:
The ids might not be unique if there are multiple modals on the page. I think this needs to be enhanced with a randomly generated string,

nickvergessen:
The modals are not mounted until displayed. So should be fairly save for now?

raimund-schluessler:
> The modals are not mounted until displayed. So should be fairly save for now?

Depends. I think I saw a case in an app where two modals appeared on top of each other. Don't remember where, though.

PR #2704, opened 2022-05-20
Put the children of AppNavigationItem outside of the main item

raimund-schluessler:
Originally created by @quentinguidee in #2572:

This change is necessary for #2571 (see why in action: https://github.com/nextcloud/nextcloud-vue/pull/2571#issuecomment-1076889150). It puts AppNavigationItem's children out of the parent item, and displays them separately. This is the simplest implementation I found without recoding the entire AppNavigationItem. I'm open to ideas if you find something cleaner.

https://user-images.githubusercontent.com/12123721/159804992-fed21b82-7581-4f07-9ead-05b3c2ec4067.mov

https://user-images.githubusercontent.com/12123721/159805040-823d0330-ad3a-48dc-a3f9-0dd61bcd79fb.mov

## Important notes/questions to reviewers

* In the current implementation, a side effect is that it disables the ability to pass (like in the doc) `class="active"` directly to the AppNavigationItem. It seems like the only purpose of passing the `class="active"` directly is only for the example? The `isActive` property works fine and continues to push the blue background.

* As another side effect of this PR, when hovering a child, it doesn't display the hover background of the parent (as demonstrated in the screenshots). I can go back to the first one easily if needed. Just unsure if it was a bug or not.

### DIFF
--- a/src/components/AppNavigationItem/AppNavigationItem.vue
+++ b/src/components/AppNavigationItem/AppNavigationItem.vue
@@ -81,7 +81,7 @@ prevent the user from collapsing the items.
 <AppNavigationItem title="Item with children" :allowCollapse="true" :open="true">
 	<template>
 		<AppNavigationItem title="AppNavigationItemChild1" />
-		<AppNavigationItem class="active" title="AppNavigationItemChild2" />
+		<AppNavigationItem title="AppNavigationItemChild2" />
 		<AppNavigationItem title="AppNavigationItemChild3" />
 		<AppNavigationItem title="AppNavigationItemChild4" />
 	</template>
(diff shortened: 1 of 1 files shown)


raimund-schluessler:
@GretaD I fixed some more issues (collapsing didn't work, children were not styled correctly, pinning was broken, see my commits). I would suggest you test it again with Mail, and ideally with Calendar as well, probably.
Some Apps will need adjustments of their CSS, see for example the necessary changes for Tasks in https://github.com/nextcloud/tasks/pull/2007
So this will be a breaking change if we interpret it strictly.

GretaD:
tested with mail and looks good, but i still have this on my code

navelement

raimund-schluessler:
> tested with mail and looks good, but i still have this on my code
> 
> navelement

Hm, no idea. Maybe this depends on the browser or how the app is compiled :thinking: When I use it in Tasks, it just creates a `div` as one would expect.

GretaD (approved):
let move on then, the component work fine on mail(i tested it) and on tasks(Raimund tested it)

ChristophWurst (approved):
Best reviewed as https://github.com/nextcloud/nextcloud-vue/pull/2704/files?w=1

PR #2713, opened 2022-05-30
Use large border-radius in Action Button

GretaD:
ref: https://github.com/nextcloud/mail/issues/6200#issuecomment-1108416443

### DIFF
--- a/src/components/Actions/Actions.vue
+++ b/src/components/Actions/Actions.vue
@@ -800,6 +800,6 @@ export default {
 // We overwrote the popover base class, so we can style
 // the popover__inner for actions only.
 .popover.action-item__popover .popover__inner {
-	border-radius: 14px;
+	border-radius: var(--border-radius-large);
 }
 </style>
--- a/src/assets/action.scss
+++ b/src/assets/action.scss
@@ -25,7 +25,7 @@
 	li {
 		&.active {
 			background-color: var(--color-background-hover);
-			border-radius: var(--border-radius-large);
+			border-radius: 6px;
 			padding: 0;
 		}
 	}


ChristophWurst (approved):
:+1: :elephant:

ChristophWurst:
Ready for review or not?

raimund-schluessler:
@GretaD I had a bit of time and worked on your PR. I hope, you don't mind. If you don't like my changes, just discard them.

The result now looks like this:
Screenshot 2022-05-30 at 23-43-31 Tasks - Nextcloud

The values of the border radius might need some tweaking, though. Given that we should respect OuterRadius = InnerRadius + Padding, the OuterRadius is 14 px at the moment, which might be a bit large. If we want it smaller, we would need to reduce either padding and/or InnerRadius.

@nimishavijay @jancborchardt

src/components/Actions/Actions.vue
```diff
@@ -170,6 +170,7 @@ export default {
 			:placement="placement"
 			:boundaries-element="boundariesElement"
 			:container="container"
+			popover-base-class="action-item__popover"
```
raimund-schluessler, line 173: Necessary, so we can only style popovers from `Actions`.

src/components/Popover/Popover.vue
```diff
@@ -77,6 +77,13 @@ export default {
 		VPopover,
 	},
 
+	props: {
+		popoverBaseClass: {
```
raimund-schluessler, line 81: Needs some documentation.

raimund-schluessler:
> The values of the border radius might need some tweaking, though. Given that we should respect OuterRadius = InnerRadius + Padding, the OuterRadius is 14 px at the moment, which might be a bit large. If we want it smaller, we would need to reduce either padding and/or InnerRadius.

I adjusted the border radii. This looks like this now:
Screenshot 2022-05-31 at 07-26-35 Tasks - Nextcloud

This is closer to what @nimishavijay had in mind in https://github.com/nextcloud/mail/issues/6200#issuecomment-1108416443, I guess.

GretaD:
> Ready for review or not?

No, sorry, my first approach was not the perfect one and raimund had a good point on fixing everything here on vue instead of getting sever involved

GretaD:
> @GretaD I had a bit of time and worked on your PR. I hope, you don't mind. If you don't like my changes, just discard them.
> 
No problem, i was working on this yesterday and i did smth similar but the row jumped a bit on hover, thats why i didnt push. 
Is this ready for review? I checked the code and looks good but didnt test with mail or calendar

GretaD:
works good with mail too

st3iny (approved):
I tested it on calendar and it worked well. It looks very good too.

~I couldn't test on mail because the app won't load and I get the following error in the console:~
```
TypeError: __webpack_modules__[moduleId] is not a function
    __webpack_require__ isMobile.js:107
```
~But I don't think that's related.~

**EDIT:** Nevermind, I did `npm ci` and `npm run dev` again and it worked.

ChristophWurst:
Shall we tag this as breaking? I don't know how high impact design changes are handled here

PR #2715, opened 2022-06-01
Augment keyboard accessibility of AppSidebar

Pytal:
### To Do

- [x] Close button
- [x] Figure
- [x] Star
- [x] Edit title


Pytal:
@raimund-schluessler any idea on resolving the failing tests?

raimund-schluessler:
The visual regression tests fail, because the appearance of the AppSidebar changed. You need to download the snapshot artifacts from the cypress run, replace the snapshots in `cypress/snapshots/base` with the `actual` ones from the cypress run, check that you like the snapshots, i.e. your changes lead to the desired result, commit the new snapshots and see the tests pass.

However, there seems to be an issue with the tertiary_actions now, see the "Editable title after click with custom tertiary action" example in the docs, where the checkbox is squished now, because the `tertiary__actions` style is not applied anymore:
Screenshot 2022-06-02 at 23-18-14 Nextcloud Vue Style Guide

src/components/AppSidebar/AppSidebar.vue
```diff
@@ -725,14 +733,6 @@ $top-buttons-spacing: 6px;
 				margin-top: -2px;
 			}
 
-			.app-sidebar-header__tertiary-actions {
```
raimund-schluessler, line 728: This needs to stay, see my comment https://github.com/nextcloud/nextcloud-vue/pull/2715#issuecomment-1145352166

src/components/AppSidebar/AppSidebar.vue
```diff
@@ -181,19 +175,17 @@ include a standard-header like it's used by the files app.
...
-											'star--star': !isStarred,
-										}"
-										class="star"
-										:size="20"
-										decorative />
-								</a>
+									<template #icon>
+										<span v-if="starLoading" class="icon-loading-small" />
+										<Star v-else
+											:fill-color="isStarred ? '#FC0' : 'currentColor'"
```
raimund-schluessler, line 184: The string `'currentColor'` is no valid color, no?

Pytal: It is passed directly to the svg fill attribute https://github.com/robcresswell/vue-material-design-icons/blob/1bc89fe0a7a01914c247263947a976b854d7c14c/build.ts#L20 and is the same as the default https://github.com/robcresswell/vue-material-design-icons/blob/1bc89fe0a7a01914c247263947a976b854d7c14c/build.ts#L41

raimund-schluessler (changes requested):
@Pytal could you please remove the PNG files that where not there before (the one with the numbers in the file name)?

src/components/AppSidebar/AppSidebar.vue
JuliaKirschenheuter, line 184: would be great, if you could use scss variables instead of color code (#FC0) - this is better to maintain

JuliaKirschenheuter: could you please use :title="t('xx', 'xx')" for the description of the icon?

Pytal: > would be great, if you could use scss variables instead of color code (#FC0) - this is better to maintain

From a ripgrep it looks like this is the only component that uses this so I wouldn't want to inflate the variables file with a variable that is only used in one place until it is used elsewhere

> could you please use :title="t('xx', 'xx')" for the description of the icon?

Aria labels have been added to the buttons instead

artonge: Looks like it already exist: https://github.com/nextcloud/server/blob/master/core/css/variables.scss#L70= but not sure if it is easy to use in your case.

src/components/AppSidebar/AppSidebar.vue
```diff
@@ -398,7 +414,9 @@ export default {
 
 	data() {
 		return {
+			changeTitleTranslated: t('Change title'),
 			closeTranslated: t('Close'),
+			favoriteTranslated: t('Favorite'),
```
artonge, line 419: Couldn't this be declared where it's used?

src/components/AppSidebar/AppSidebar.vue
```diff
@@ -146,14 +146,6 @@ include a standard-header like it's used by the files app.
...
 				}"
 				class="app-sidebar-header">
-				<!-- close sidebar button -->
-				<a v-tooltip.auto="closeTranslated"
-					href="#"
-					class="app-sidebar__close"
-					@click.prevent="closeSidebar">
-					<Close class="app-sidebar__close-icon" :size="20" decorative />
-				</a>
-
```
artonge, line 156: No more close button?

src/components/AppSidebar/AppSidebar.vue
artonge: Same for `changeTitleTranslated`

src/components/AppSidebar/AppSidebar.vue
```diff
@@ -804,38 +814,6 @@ $top-buttons-spacing: 6px;
...
-						&:hover {
-							color: #FC0;
-						}
-					}
-					&--starred {
-						color: #FC0;
-						&:hover {
-							color: #000;
-							opacity: .5;
-						}
```
artonge, line 835: Is the hover behavior moved somewhere else?

artonge (changes requested):

src/components/AppSidebar/AppSidebar.vue
CarlSchwan: It was moved at the end of this file

src/components/AppSidebar/AppSidebar.vue
CarlSchwan: It's now using a Button, so maybe not needed anymore?

src/components/AppSidebar/AppSidebar.vue
artonge: Indeed

src/components/AppSidebar/AppSidebar.vue
Pytal: Yes not needed as it should behave as Button component

src/components/AppSidebar/AppSidebar.vue
Pytal: The extractor library does not support this https://github.com/lukasgeiter/gettext-extractor/issues/40 which is why it must be used this way in this library https://github.com/nextcloud/nextcloud-vue/blob/2ac55083188ff5ebb4a3defee052f2508b4b165e/src/components/EmojiPicker/EmojiPicker.vue#L180-L194

src/components/AppSidebar/AppSidebar.vue
Pytal: Available in server scss variables but not https://github.com/nextcloud/nextcloud-vue/blob/2ac55083188ff5ebb4a3defee052f2508b4b165e/src/assets/variables.scss so I think we should keep it as local as possible and not add it when it's only used once in the vue component library

artonge (approved):

raimund-schluessler:
@Pytal The color or opacity of the star is off, which looks weird if it is overlayed with the folder icon, see e.g. 

AppSidebar vue-subtitle_true-starred_true-compact_true-header_image-secondary_none-editable_true-base

and https://github.com/nextcloud/nextcloud-vue/pull/2715/files#diff-b888e1950c55db271730b2350fbe7aa975167f52b84930ee515b0f86387d81ee for the before/after comparison

Also, the height of the header now changes when you start the edit mode, making all content jump:
SidebarAfter

This was not the case before:
SidebarBefore

raimund-schluessler (changes requested):
See https://github.com/nextcloud/nextcloud-vue/pull/2715#issuecomment-1149609352

Pytal:
> @Pytal The color or opacity of the star is off, which looks weird if it is overlayed with the folder icon, see e.g.

This is because of the `tertiary-no-background` Button type for hover feedback purposes, see https://github.com/nextcloud/nextcloud-vue/commit/870d274a06eaf7a3f02c0751e16d6b9ea5c6f7d8

I have overridden the star button styles for this specific case

> Also, the height of the header now changes when you start the edit mode, making all content jump:

Fixed

raimund-schluessler:
@Pytal @jancborchardt @nimishavijay
The hover feedback of the star changed, see here:
Before:
Before

After:
After

Also, the distance between the title and the subtitle changed (due to not increasing the header height and keeping a certain padding between title and subtitle while editing):
Before:
Before

After:
After

I have no preference here regarding any of these two changes, just wanted to point them out so everyone is aware.

raimund-schluessler (approved):
Code looks good. Everything still works. There are some visual changes, that might need design approval though @nextcloud/designers, see https://github.com/nextcloud/nextcloud-vue/pull/2715#issuecomment-1150834638 (since the changes affect a fundamental component and we had issues with adjustments in the past).

Pytal:
> * Regarding title and subtitle positioning and spacing: That should be fixed as it’s a regression – both are too far down and too far apart, ideally should be same as before as that was ideal in relation to the filetype icon.

The title may either be editable or not so to prevent layout shifts due to the 44px Button component when editing the minimum height for the title must be 44px, I'll see about tweaking the positioning

> * On the favoriting: This would probably be for a follow up pull request, but the actual accessibility fix would be what we did in the Mail app:

Will be done here for now and the new component may be added separately

Pytal:
The star and star outline icons without color look like

<table>
<tr>
	<th>Before</th>
	<th>After</th>
</tr>
<tr>
	<td>
		
	</td>
	<td>
		
	</td>
</tr>
<tr>
	<td>
		
	</td>
	<td>
		
	</td>
</tr>
</table>

Without the color the star would blend in with the image it overlays if they are similarly colored, @jancborchardt @nimishavijay design suggestions?

Pytal:
@jancborchardt how should we keep the star visible with the new grayscale filled and outlined star icons seen in the screenshots above?

PR #2718, opened 2022-06-01
Add actions slot to dashboard widget item

### LINKED ISSUE #2685: Allow to have different actions for dashboard items
Background: https://github.com/nextcloud/spreed/pull/5923#issuecomment-921612795

In Talk we have the idea to add a "mark as read" action to conversations. However when there are no mentions, we list all conversations and some of them might be read, others are not. And it's really confusing to have the "Mark read" action for a conversation after you just pressed this.

Would be nice if the API can be extended, so we either can specify the actions per item, or we can somehow control if an action should be shown for an item.

raimund-schluessler:
This PR adds an `action` slot to the `DashboardWidgetItem` so that it's possible to provide different actions for each dashboard widget item. This also allows to provide material design icons for each action, which was not possible before. Closes #2685.

See https://github.com/nextcloud/tasks/pull/2017 for an example how this works.

In Tasks, we can now prevent showing actions for read-only tasks:

Before:
Screenshot 2022-06-01 at 22-46-13 Dashboard - Nextcloud

After:
Screenshot 2022-06-01 at 22-54-37 Dashboard - Nextcloud

### DIFF
--- a/src/components/DashboardWidgetItem/DashboardWidgetItem.vue
+++ b/src/components/DashboardWidgetItem/DashboardWidgetItem.vue
@@ -57,13 +57,16 @@ This component is meant to be used inside a DashboardWidget component.
 				</p>
 			</div>
 			<Actions v-if="gotMenu" :force-menu="forceMenu" menu-align="right">
-				<ActionButton v-for="(m, menuItemId) in itemMenu"
-					:key="menuItemId"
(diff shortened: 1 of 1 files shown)


jakobroehrl (approved):
looks good

GretaD (approved):
:+1:

PR #2736, opened 2022-06-07
Removed after show trigger call, this call was directing the focus to…

mstolf:
Removed after hide trigger call, this call was directing the focus to the div of the DotsHorizontal component, making that when clicking outside the component the "..." appear again

Fix [#7459](https://github.com/nextcloud/spreed/issues/7459)

### DIFF
--- a/src/components/Avatar/Avatar.vue
+++ b/src/components/Avatar/Avatar.vue
@@ -62,8 +62,7 @@
 			placement="auto"
 			:container="menuContainer"
 			:open="contactsMenuOpenState"
-			@after-show="handlePopoverAfterShow"
-			@after-hide="handlePopoverAfterHide">
+			@after-show="handlePopoverAfterShow">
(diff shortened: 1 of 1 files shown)


skjnldsv:
Can you rebase @mstolf  :)

src/components/Avatar/Avatar.vue
```diff
@@ -455,10 +454,6 @@ export default {
 				links[0].focus()
 			}
 		},
-		handlePopoverAfterHide() {
-			// bring focus back to the trigger
```
juliushaertl, line 459: This is still a case that should be preserved when closing the menu with the keyboard I guess. This was introduced with https://github.com/nextcloud/nextcloud-vue/pull/1854 for accessibility reasons.

juliushaertl (changes requested):

susnux:
Is this still relevant?

susnux:
Closing as there is no activity for two years.
If this is still relevant please reopen the PR.

PR #2738, opened 2022-06-07
Move close button within modal boundaries

### LINKED ISSUE #2737: Move the close button into the modals
The close buttons should be visible within the modal windows. 
cc @jancborchardt & @nimishavijay 

Screenshot 2022-06-07 at 15 31 02

proposed alternative (note that the close button is hovered there):
Screenshot 2022-06-07 at 15 36 47

marcoambrosini:
fix #2737

https://user-images.githubusercontent.com/26852655/172405851-89a1910d-9daa-4455-ba62-5f134145e07d.mov

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -381,12 +381,12 @@ export default {
 		},
 
 		/**
-		 * Pass in true if you want the modal 'close' button to be displayed
-		 * within the modal boundaries.
+		 * Pass in false if you want the modal 'close' button to be displayed
+		 * outside the modal boundaries, in the top right corner of the window
 		 */
 		closeButtonContained: {
 			type: Boolean,
-			default: false,
+			default: true,
 		},
 	},
 
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -166,7 +166,7 @@
 						</Actions>
 
 						<!-- Close modal -->
-						<Actions v-if="canClose" class="header-close">
+						<Actions v-if="canClose && !closeButtonContained" class="header-close">
 							<ActionButton @click="close">
 								<template #icon>
 									<Close :size="iconSize" title="" decorative />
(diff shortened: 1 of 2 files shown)


jancborchardt (approved):
Looks good design-wise

raimund-schluessler:
I didn't test it, but the code looks good. I would just vote for having the default as `true` and make it a breaking change for merging into master, since we anyway already have some breaking changes in master already that will need a version bump to 6.

src/components/Modal/Modal.vue
```diff
@@ -367,6 +379,15 @@ export default {
...
 			default: 'body',
 		},
+
+		/**
+		 * Pass in true if you want the modal 'close' button to be displayed
+		 * within the modal boundaries.
+		 */
+		closeButtonContained: {
+			type: Boolean,
+			default: false,
```
raimund-schluessler, line 389:
```suggestion
			default: true,
```
and mark the PR as breaking.

src/components/Modal/Modal.vue
```diff
@@ -207,6 +207,16 @@
 
 					<!-- Content -->
 					<div class="modal-container">
+						<!-- Close modal -->
+						<ButtonVue v-if="canClose && closeButtonContained"
+							type="tertiary"
+							class="modal-container__close"
+							:aria-label="t('spreed', 'Close modal')"
```
raimund-schluessler, line 214: The translation needs to come from a computed property, otherwise the extraction script won't find it unfortunately. Also, `spreed` is wrong 😉

raimund-schluessler (changes requested):
Looks good and works as it should, only https://github.com/nextcloud/nextcloud-vue/pull/2738/files#r892030091 needs to be fixed.

nimishavijay (approved):
Looks great 🚀

src/components/Modal/Modal.vue
marcoambrosini: 🙈

raimund-schluessler (approved):

PR #2757, opened 2022-06-10
Alternative style for CheckboxRadioSwitch (radio type)

eneiluj:
This was inspired by https://github.com/nextcloud/server/issues/26691

A new boolean prop `buttonVariant` toggles between the classic style and this one. It can be used with all types.

If the CheckboxRadioSwitch elements are displayed next to each other vertically, their borders are "merged".

https://user-images.githubusercontent.com/11291457/174578908-5616d945-6ad4-4c9c-89fc-5cda32732f2f.mp4

### DIFF
--- a/src/components/CheckboxRadioSwitch/CheckboxRadioSwitch.vue
+++ b/src/components/CheckboxRadioSwitch/CheckboxRadioSwitch.vue
@@ -81,12 +81,12 @@ export default {
 </script>
 ```
 
-### Standard radio set with alternative style
+### Standard radio set with alternative button style
 ```vue
 <template>
 	<div>
-		<CheckboxRadioSwitch :checked.sync="sharingPermission" value="r" name="sharing_permission_radio" type="radio" :alternative-style="true">Default permission read</CheckboxRadioSwitch>
-		<CheckboxRadioSwitch :checked.sync="sharingPermission" value="rw" name="sharing_permission_radio" type="radio" :alternative-style="true">Default permission read+write</CheckboxRadioSwitch>
+		<CheckboxRadioSwitch :checked.sync="sharingPermission" value="r" name="sharing_permission_radio" type="radio" :button-variant="true">Default permission read</CheckboxRadioSwitch>
+		<CheckboxRadioSwitch :checked.sync="sharingPermission" value="rw" name="sharing_permission_radio" type="radio" :button-variant="true">Default permission read+write</CheckboxRadioSwitch>
 		<br>
 		sharingPermission: {{ sharingPermission }}
 	</div>
@@ -154,7 +154,7 @@ export default {
 			'checkbox-radio-switch--checked': isChecked,
 			'checkbox-radio-switch--disabled': disabled,
 			'checkbox-radio-switch--indeterminate': indeterminate,
-			'checkbox-radio-switch--alternate-radio': isAlternativeRadio,
+			'checkbox-radio-switch--button-variant': buttonVariant,
 		}"
 		:style="cssVars"
 		class="checkbox-radio-switch">
@@ -171,7 +171,7 @@ export default {
 		<label :for="id" class="checkbox-radio-switch__label">
 			<div v-if="loading" class="icon-loading-small checkbox-radio-switch__icon" />
 			<icon :is="checkboxRadioIconElement"
-				v-else-if="!isAlternativeRadio"
+				v-else-if="!buttonVariant"
 				:size="size"
 				class="checkbox-radio-switch__icon"
 				title=""
@@ -233,9 +233,9 @@ export default {
 		},
 
 		/**
-		 * Toggle the alternative style (for radio type only)
+		 * Toggle the alternative button style
 		 */
-		alternativeStyle: {
+		buttonVariant: {
 			type: Boolean,
 			default: false,
 		},
@@ -323,10 +323,6 @@ export default {
 				return TYPE_RADIO
 			}
 			return TYPE_CHECKBOX
-		},
-
-		isAlternativeRadio() {
-			return this.type === TYPE_RADIO && this.alternativeStyle
 		},
 
 		/**
@@ -498,7 +494,7 @@ $spacing: 4px;
 		color: var(--color-primary-element-light);
 	}
 
-	&.checkbox-radio-switch--alternate-radio.checkbox-radio-switch {
+	&--button-variant.checkbox-radio-switch {
 		border: 2px solid var(--color-border-dark);
 		border-radius: var(--border-radius);
 
@@ -524,12 +520,12 @@ $spacing: 4px;
 				background-color: var(--color-background-dark);
 			}
 		}
+	}
 
-		.checkbox-radio-switch__label {
-			border-radius: 0 !important;
-			width: 100% !important;
-			margin: 0 !important;
-		}
+	&--button-variant &__label {
+		border-radius: 0 !important;
+		width: 100% !important;
+		margin: 0 !important;
 	}
 }
 </style>
--- a/src/components/CheckboxRadioSwitch/CheckboxRadioSwitch.vue
+++ b/src/components/CheckboxRadioSwitch/CheckboxRadioSwitch.vue
@@ -71,6 +71,27 @@ export default {
 	</div>
 </template>
 <script>
+	export default {
+		data() {
+			return {
+				sharingPermission: 'r',
+			}
+		}
+	}
+</script>
+```
+
+### Standard radio set with alternative style
+```vue
+<template>
+	<div>
+		<CheckboxRadioSwitch :checked.sync="sharingPermission" value="r" name="sharing_permission_radio" type="radio" :alternative-style="true">Default permission read</CheckboxRadioSwitch>
+		<CheckboxRadioSwitch :checked.sync="sharingPermission" value="rw" name="sharing_permission_radio" type="radio" :alternative-style="true">Default permission read+write</CheckboxRadioSwitch>
+		<br>
+		sharingPermission: {{ sharingPermission }}
+	</div>
+</template>
+<script>
 export default {
 	data() {
 		return {
@@ -133,6 +154,7 @@ export default {
 			'checkbox-radio-switch--checked': isChecked,
 			'checkbox-radio-switch--disabled': disabled,
 			'checkbox-radio-switch--indeterminate': indeterminate,
+			'checkbox-radio-switch--alternate-radio': isAlternativeRadio,
 		}"
 		:style="cssVars"
 		class="checkbox-radio-switch">
@@ -149,7 +171,7 @@ export default {
 		<label :for="id" class="checkbox-radio-switch__label">
 			<div v-if="loading" class="icon-loading-small checkbox-radio-switch__icon" />
 			<icon :is="checkboxRadioIconElement"
-				v-else
+				v-else-if="!isAlternativeRadio"
 				:size="size"
 				class="checkbox-radio-switch__icon"
 				title=""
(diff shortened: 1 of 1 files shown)


tcitworld:
> A new boolean prop alternativeStyle (any better idea?)

Might have different behaviour depending on the type of the input. Maybe something like `button-variant` instead ?

eneiluj:
`button-variant` is better, thanks.

This prop now has an effect on all types.

> Might have different behaviour depending on the type of the input.

Any suggestion/example?

tcitworld:
`alternativeStyle` could also be understood as an option to chose between the regular checkbox type and the switch type.

jancborchardt (changes requested):
Nice! :) Some design feedback:
- [ ] The last entry without icon should probably still be indented so the text is left-aligned with the others.
- [x] Currently each individual item has a border-radius on all corners – but actually it is a group which should only have border-radius on the 4 outside corners. So only on the top left and top right of the first entry and the bottom left and bottom right of the last entry.
- [x] Since we are increasing border-radius across the board, it’s probably best to go with `var(--border-radius-large)` for this

src/components/CheckboxRadioSwitch/CheckboxRadioSwitch.vue
```diff
@@ -210,6 +235,38 @@ export default {
...
+			default: false,
+		},
+
+		/**
+		 * Toggle the bottom left and right rounded corners for the alternative button style
+		 */
+		buttonVariantLast: {
+			type: Boolean,
+			default: false,
+		},
```
skjnldsv, line 268: This should be automatically calculated.
Grouped-by-name radio can easily be targeted with first and last :)

skjnldsv (changes requested):

src/components/CheckboxRadioSwitch/CheckboxRadioSwitch.vue
eneiluj: Alright, now using `:first-of-type` and `:last-of-type` which will only work if they are siblings.

We still need to get the information about whether they are direct successors or not (grouped or not). Still using a prop for that.

I also added the ability to group vertically (video is up-to-date in the first comment).

raimund-schluessler: > We still need to get the information about whether they are direct successors or not (grouped or not). Still using a prop for that.

I don't think that's necessary (or even helpful). Grouped checkboxes need to be direct siblings with no checkboxes as siblings that are not grouped. So they need to be wrapped in a common container. Otherwise the markup will fail, without or with the grouped prop.

See this for example:
```vue
<template>
<div>
	<div>
	First toggle with three options
		<CheckboxRadioSwitch :checked.sync="sharingPermission" value="r" name="sharing_permission_radio" type="radio" :button-variant="true" :button-variant-v-grouped="true">Default permission read</CheckboxRadioSwitch>
		<CheckboxRadioSwitch :checked.sync="sharingPermission" value="rw" name="sharing_permission_radio" type="radio" :button-variant="true" :button-variant-v-grouped="true">Default permission read+write</CheckboxRadioSwitch>
		<CheckboxRadioSwitch :checked.sync="sharingPermission" value="rw" name="sharing_permission_radio" type="radio" :button-variant="true" :button-variant-v-grouped="true">Default permission read+write</CheckboxRadioSwitch>
	</div>
	<div>
	Second toggle with two options
		<CheckboxRadioSwitch :checked.sync="sharingPermission" value="r" name="sharing_permission_radio" type="radio" :button-variant="true" :button-variant-v-grouped="true">Default permission read</CheckboxRadioSwitch>
		<CheckboxRadioSwitch :checked.sync="sharingPermission" value="rw" name="sharing_permission_radio" type="radio" :button-variant="true" :button-variant-v-grouped="true">Default permission read+write</CheckboxRadioSwitch>
		Other toggle not grouped
		<CheckboxRadioSwitch :checked.sync="sharingPermission" value="rw" name="sharing_permission_radio" type="radio" :button-variant="true">Default permission read+write</CheckboxRadioSwitch>
		<br>
		sharingPermission: {{ sharingPermission }}
	</div>
</div>
</template>
<script>
export default {
	data() {
		return {
			sharingPermission: 'r',
		}
	}
}
</script>
```

Which gives:
Screenshot 2022-06-20 at 12-18-46 Nextcloud Vue Style Guide

Notice that the corners of the second last checkbox are missing, since there is another one not belonging to the group afterwards. For the three grouped checkboxes wrapped in a container, it works fine.

So imho we only need a prop to indicate horizontal/vertical grouping (would be nice to figure that out by CSS as well, but I don't know how) and leave the rest to the proper markup (give a common container to the grouped boxes).

raimund-schluessler (changes requested):

jancborchardt (approved):
Looks great design-wise now! :)

@eneiluj regarding the border-radius, it could also be done by adding a parent/container and using first- and last-child on that.

src/components/CheckboxRadioSwitch/CheckboxRadioSwitch.vue
```diff
@@ -210,6 +234,30 @@ export default {
...
+
+		/**
+		 * If the elements are all direct successors and grouped vertically
+		 */
+		buttonVariantVGrouped: {
+			type: Boolean,
+			default: false,
+		},
```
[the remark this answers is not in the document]
eneiluj: :+1: Only one prop now.

src/components/CheckboxRadioSwitch/CheckboxRadioSwitch.vue
```diff
@@ -70,6 +70,27 @@ export default {
...
+	}
+</script>
+```
+
+### Standard radio set with alternative button style
+```vue
+<template>
+	<div>
+		<CheckboxRadioSwitch :checked.sync="sharingPermission" value="r" name="sharing_permission_radio" type="radio" :button-variant="true" :button-variant-grouped="vertical">Default permission read</CheckboxRadioSwitch>
+		<CheckboxRadioSwitch :checked.sync="sharingPermission" value="rw" name="sharing_permission_radio" type="radio" :button-variant="true" :button-variant-grouped="vertical">Default permission read+write</CheckboxRadioSwitch>
```
raimund-schluessler, line 89:
```suggestion
		<CheckboxRadioSwitch :checked.sync="sharingPermission" value="r" name="sharing_permission_radio" type="radio" :button-variant="true" button-variant-grouped="vertical">Default permission read</CheckboxRadioSwitch>
		<CheckboxRadioSwitch :checked.sync="sharingPermission" value="rw" name="sharing_permission_radio" type="radio" :button-variant="true" button-variant-grouped="vertical">Default permission read+write</CheckboxRadioSwitch>
```

src/components/CheckboxRadioSwitch/CheckboxRadioSwitch.vue
```diff
@@ -133,6 +154,9 @@ export default {
 			'checkbox-radio-switch--checked': isChecked,
 			'checkbox-radio-switch--disabled': disabled,
 			'checkbox-radio-switch--indeterminate': indeterminate,
+			'checkbox-radio-switch--button-variant': buttonVariant,
+			'checkbox-radio-switch--button-variant-v-grouped': buttonVariantGrouped === 'vertical',
+			'checkbox-radio-switch--button-variant-h-grouped': buttonVariantGrouped === 'horizontal',
```
raimund-schluessler, line 159: Minor nitpick:
```suggestion
			'checkbox-radio-switch--button-variant-v-grouped': buttonVariant && buttonVariantGrouped === 'vertical',
			'checkbox-radio-switch--button-variant-h-grouped': buttonVariant && buttonVariantGrouped === 'horizontal',
```
because at the moment you can have `buttonVariantGrouped="horizontal"` while `:buttonVariant="false"` which gives `margin-left: -2px` to an active checkbox.

eneiluj:
> So they need to be wrapped in a common container. Otherwise the markup will fail, without or with the grouped prop.

@raimund-schluessler What I meant is that from within the CheckboxRadioSwitch component, we can't know if all of the related CheckboxRadioSwitch are direct siblings (or if there are elements in between) so we need a prop for that. Of course they need to have a common direct parent but we need to know the intention of the developer as we can't guess it with CSS.

@jancborchardt If you mean adding one container around all the CheckboxRadioSwitch of a group, this is not possible as this component is only one element.
If you mean a container around each CheckboxRadioSwitch I don't see what difference it makes :grin:.
Also, last-child is not enough as you can have other type of children in the same container so it would not select what we want for sure.

Thank you both for the feedback :blue_heart: .

So we can now choose if we display them grouped with the buttonVariantGrouped prop. This has an effect on the first/last (selected with `:first-of-type` and `:last-of-type`) element's border radius and it merges the borders between elements.
When not grouped, all elements have rounded borders.

dartcafe:
Why not adding a new component like https://github.com/nextcloud/polls/blob/master/src/js/components/Base/RadioGroupDiv.vue?

marcoambrosini (commented):
Hi, I think that the style of this component departs a bit from our design language, most notably due to the use of borders, which is something we've been trying to move away for some time. cc @jancborchardt 
I know it might be too late for change but I'll leave an Idea of a similarly behaving component that we're using in our UI and whose style could probably fit here.


Screenshot 2022-06-21 at 13 00 26

This also happens to be somewhat consistent with how they do radios in material design

Screenshot 2022-06-20 at 18 22 18

src/components/CheckboxRadioSwitch/CheckboxRadioSwitch.vue
eneiluj: :+1:

src/components/CheckboxRadioSwitch/CheckboxRadioSwitch.vue
eneiluj: :+1:

raimund-schluessler (approved):
The code looks good and works as intended.

Regarding the design, I think @marcoambrosini proposal looks nice as well, but that's up to @nextcloud/designers to decide and can be improved later on.

Also, I would be in favor of the grouped checkboxes container component @dartcafe proposed, because
1. grouped checkboxes obligatorily need a parent element containing only the checkboxes belonging to the group (otherwise the style fails) and a provided container component would make this more clear
2. then the container would only need a single prop indicating whether the boxes are grouped vertically or horizontally

So something like this could be the result:
```
<CheckboxRadioSwitchGroup direction="vertical/horizontal">
	<CheckboxRadioSwitch :checked.sync="sharingPermission" value="r" name="sharing_permission_radio" type="radio"  :button-variant="true>Default permission read</CheckboxRadioSwitch>
	<CheckboxRadioSwitch :checked.sync="sharingPermission" value="rw" name="sharing_permission_radio" type="radio"  :button-variant="true>Default permission read+write</CheckboxRadioSwitch>
	<CheckboxRadioSwitch :checked.sync="sharingPermission" value="rwx" name="sharing_permission_radio" type="radio"  :button-variant="true>Default permission read+write+execute</CheckboxRadioSwitch>
</CheckboxRadioSwitchGroup>
```

eneiluj:
@dartcafe I made a similar component like your RadioGroupDiv for the specific needs of a project: https://github.com/eneiluj/integration_visavid/blob/master/src/components/RadioElementSet.vue
But I guess passing the options as an array of objects makes prop validation harder and gives less control than the suggestion of @raimund-schluessler to insert the group content in the default slot.

@marcoambrosini So we have multiple ways to go. Which one makes more sense?
* Leave the CheckboxRadioSwitch component untouched and create 2 new components: CheckboxRadioSwitchButton and CheckboxRadioSwitchButtonGroup. But CheckboxRadioSwitchButton would repeat most of CheckboxRadioSwitch's code...
* Merge this PR and implement CheckboxRadioSwitchGroup
* Any other idea?

skjnldsv:
> * Leave the CheckboxRadioSwitch component untouched and create 2 new components: CheckboxRadioSwitchButton and CheckboxRadioSwitchButtonGroup. But CheckboxRadioSwitchButton would repeat most of CheckboxRadioSwitch's code...

Yeah, I'm not in favour.
This component ships fancy utils and methods about grouping, naming, events etc etc.
That would just duplicate the code.

I would be ok creating a `CheckboxRadioSwitchGroup`, that would be easy to check the parent from the `CheckboxRadioSwitch` component just to see which direction is the prop set.

Let's keep it simple people! :rocket:

PR #2845, opened 2022-07-15
Add headline to sidebar sections to improve landmarks

nickvergessen:
> Section lacks heading. Consider using h2-h6 elements to add identifying
> headings to all sections, or else use a div element instead for any
> cases where no heading is needed.


Pytal (approved):

JuliaKirschenheuter (approved):
Ok, since AppSidebar would have ```<h2>``` heading, we could use here  ```<h3>``` as direct child. In this case it is also not possible to use ``` role="heading" + aria-level ``` -> and it is ok to use ``` <h3>``` with class="hidden-visually"

PR #2846, opened 2022-07-15
Fix HTML validation of AppNavigationToggle

nickvergessen:
> Error: Attribute icon not allowed on element button at this point.

### DIFF
--- a/src/components/AppNavigationToggle/AppNavigationToggle.vue
+++ b/src/components/AppNavigationToggle/AppNavigationToggle.vue
@@ -22,21 +22,20 @@
  -
  -->
 <template>
-	<Actions class="app-navigation-toggle">
-		<ActionButton :aria-expanded="open ? 'true' : 'false'"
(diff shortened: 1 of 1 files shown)


JuliaKirschenheuter (changes requested):
Hope i've understtod the problem right. Question: why not to fix this in https://github.com/nextcloud/nextcloud-vue/blob/de5b332374428795cac78b2a534af02ead8b0af0/src/components/ActionButton/ActionButton.vue#L79 and use src/components/Button/Button.vue instead? 

My suggestin is that we still need an ActionButton inside of AppNavigationToggle. But i also could be wrong.

nickvergessen:
The Actions menu with 1 entry is a "hack" from before we had the buttons component. We should use the button component directly to ease the code.
Yeah ultimately also the Actions should use the Button component directly, but due to the huge list of things to do I would like to fix one thing at a time. Redoing Actions to use Button takes a lot more time then fixing this simple usage.

PR #2847, opened 2022-07-15
Fix misnamed attribute menu-align

nickvergessen:


nickvergessen:
The whole UI looks weird now that the seetting is respected again.
Should we maybe not forward it and instead ask apps to use placement if they want it instead, so they can reconsider while touching it?

raimund-schluessler:
I would rather remove the attribute, than make it work. Placement right looks quire weird to me as well, e.g. for `ListItem`:
Screenshot 2022-08-10 at 22-13-22 Nextcloud Vue Style Guide

I think the standard auto works just fine.

marcoambrosini (approved):

marcoambrosini:
so should we just remove this?

raimund-schluessler:
Superseded by #3749.

PR #2856, opened 2022-07-18
Fix ListItem css

jotoeri:
This fixes two things:
- The ListItem was not really usable without subtitle but with counter. Different line heights and a completely messed up design were the result. This now brings that case onto one line, when only title and counter are available and fixes the ListItem height, so all ListItems are the same height. 

| Before | After |
|--|--|
[Full odd looking gif](https://user-images.githubusercontent.com/47433654/164996305-9cd5c448-f1bd-45de-8328-f6dd66b8ab91.gif)||
|old|non_compact|
 
- In connection to the fixed height, it introduces a compact-mode for the list item. Once used without subtitle or in a mixed case, the default sizes are quite large. Esp. if there are multiple elements without subtitle, but a few with subtitle, there is just too much space inbetween those items without. Thus the compact mode reduces the size of the ListItem a bit, so the mixed case looks good.

| Default | Compact Mode |
|--|--|
|non_compact|compact_mode|

### DIFF
--- a/src/components/ListItem/ListItem.vue
+++ b/src/components/ListItem/ListItem.vue
@@ -145,7 +145,8 @@
 
 				<!-- Main content -->
 				<div class="list-item-content">
-					<div class="list-item-content__main">
+					<div class="list-item-content__main"
+						:class="{ 'list-item-content__main--oneline': oneLine }">
 
 						<!-- First line, title and details -->
 						<div class="line-one"
@@ -360,6 +361,10 @@ export default {
 			}
 		},
 
+		oneLine() {
+			return !this.hasSubtitle && !this.showDetails
+		},
+
 		showCounter() {
 			return !this.displayActionsOnHoverFocus || this.forceDisplayActions
 		},
@@ -500,6 +505,7 @@ export default {
 	&-content__wrapper {
 		display: flex;
 		align-items: center;
+		height: 48px;
 	}
 
 	&-content {
@@ -510,16 +516,19 @@ export default {
 
 		&__main {
 			flex: 1 1 auto;
-			flex-direction: column;
 			width: 0;
 			margin: auto 0;
+
+			&--oneline {
+				display: flex;
+			}
 		}
 
 		&__actions {
 			flex: 0 0 auto;
 			align-self: center;
 			justify-content: center;
-
+			margin-left: 4px;
 		}
 	}
 
@@ -533,7 +542,8 @@ export default {
 	align-items: center;
 	justify-content: space-between;
 	white-space: nowrap;
-	margin: 0 auto;
+	margin: 0 auto 0 0;
+	overflow: hidden;
 	&--bold {
 		font-weight: bold;
 	}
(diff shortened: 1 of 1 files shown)


raimund-schluessler:
@jotoeri Could you please show before and after pictures and describe a bit more what was/is the problem you are fixing here? If we change the padding of `ListItem` and of the `Actions` inside, we might break the alignment with other components in the library, so there should b good reasoning, as we likely need to change it for other components as well then.

jotoeri:
> @jotoeri Could you please show before and after pictures and describe a bit more what was/is the problem you are fixing here? If we change the padding of `ListItem` and of the `Actions` inside, we might break the alignment with other components in the library, so there should b good reasoning, as we likely need to change it for other components as well then.

Well, that's why it has the developing label and i had to check if this is, what i wanted. :wink: 
- You're right, the padding of the list-item is not necessary anymore. Compared to my initial base on the last release, the padding of the appnavigation already increased, so the padding here would be doubled.
- For the actions margin-left i have a specific before/after here. It just increases the space a little bit, if the text overflow gets hidden:

| Before | After |
|---|---|
| actions_old | actions_new2 |
- For the whole PR i'll update the initial comment later :slightly_smiling_face:

jotoeri:
As i still need to double-check these changes with other apps, @raimund-schluessler do you know some apps that use the listItem?

Chartman123:
@jotoeri A quick search within the Nextcloud organisation here on GitHub only brought up the mail app:

https://github.com/search?q=org%3Anextcloud+%22import+ListItem%22&type=code

But there might be of course other apps that are not connected to the organisation here.

jotoeri:
Just checked it on Talk and didn't see any problems. So should be good to go now. :slightly_smiling_face:

src/components/ListItem/ListItem.vue
```diff
@@ -510,16 +534,19 @@ export default {
...
+				display: flex;
+			}
 		}
 
 		&__actions {
 			flex: 0 0 auto;
 			align-self: center;
 			justify-content: center;
-
+			margin-left: 4px;
```
jotoeri, line 549: | Before | After |
|---|---|
| actions_old | actions_new2 |

src/components/ListItem/ListItem.vue
```diff
@@ -565,15 +593,14 @@ export default {
 	&__subtitle {
 		overflow: hidden;
 		flex-grow: 1;
-		padding-right: 4px;
```
jotoeri, line 568: Shifted to counter margin, so it also applies if no subtitle used.

marcoambrosini (approved):

Chartman123 (approved):

nimishavijay (approved):

marcoambrosini (changes requested):
@jotoeri could you add a compact example in the docs too?

jotoeri:
:+1: @marcoambrosini Can i easily get that to half width? Just add a style to the surrounding `ul`? :thinking:
preview

marcoambrosini:
@jotoeri sure thingy :)

jotoeri:
preview

marcoambrosini (approved):

PR #2864, opened 2022-07-20
Turn the button into a link when href is provided

marcoambrosini:
Sometimes is useful to have links that are styled as buttons. This
allows to pass in an href prop to the button component and turns the
native root html element into an `a` when that's provided.

### DIFF
--- a/src/components/Button/Button.vue
+++ b/src/components/Button/Button.vue
@@ -198,8 +198,8 @@ button {
 </docs>
 
 <template>
-	<button class="button-vue"
-		v-bind="$attrs"
+	<root-element class="button-vue"
+		v-bind="rootElement"
 		:class="buttonClassObject"
 		:aria-label="ariaLabel"
 		:type="nativeType"
@@ -220,7 +220,7 @@ button {
 				<slot />
 			</span>
 		</span>
-	</button>
+	</root-element>
 </template>
 <script>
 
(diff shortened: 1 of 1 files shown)


CarlSchwan (approved):
:heart:

src/components/Button/Button.vue
```diff
@@ -300,6 +309,21 @@ export default {
...
+				return {
+					is: 'a',
+					href: this.href,
+					...this.$attrs,
+				}
+			}
+			return {
+				is: 'button',
+				...this.$attrs,
+			}
```
raimund-schluessler, line 324: @marcoambrosini I am not 100% sure, but I think `href` should be included in `$attrs`, so this should work as well:
```suggestion
			return {
				is: this.href ? 'a' : 'button',
				...this.$attrs,
			}
```

raimund-schluessler (approved):
Works

raimund-schluessler:
We could also check if there is the `to` prop set, and alternatively make it a `router-link` if present.

src/components/Button/Button.vue
marcoambrosini: I tested without explicitly adding href and it didn't work 🤷‍♂️

marcoambrosini:
> We could also check if there is the to prop set, and alternatively make it a router-link if present.

Follow up!

src/components/Button/Button.vue
marcoambrosini: maybe cause it's a prop it's captured by $props and taken out of $attrs?

Issue #2500, opened 2022-02-17
[BITV/A11Y] EmptyContent accessibility


PVince81:
* ~Fix empty h2 in Dashboard widgets with EmptyContent view (an opportunity to test the warning below)~ h2 elements now render conditionally only when a title is passed
* ~Image should not have `role="img"`, but ideally be an img with `alt=""`~ Only accessible icons from https://github.com/robcresswell/vue-material-design-icons are accepted

Pytal:
### Requires

- [ ] https://github.com/nextcloud/nextcloud-vue/pull/2867

Pytal:
Since a title is required for the EmptyContent component, how would the first checkbox be fixed @juliushaertl

Pytal:
Since a title is required for the EmptyContent component but is not added in dashboard widgets as designed probably, how would the first checkbox be fixed @juliushaertl

juliushaertl:
I'd say we can just not render the h2 in the EmptyContent if the slot is empty.

PR #2867, opened 2022-07-21
Accessible EmptyContent component

Pytal:
Close https://github.com/nextcloud/nextcloud-vue/issues/2500

### To Do

- [x] Require accessible https://github.com/robcresswell/vue-material-design-icons exclusively with the icon slot and drop class based icons
	- Icons use inline `svg` markup https://github.com/robcresswell/vue-material-design-icons/blob/5.1.2/build.ts#L20-L28
	- Icons have `role="img"` https://github.com/robcresswell/vue-material-design-icons/blob/5.1.2/build.ts#L18
	- Icons are [decorative](https://www.w3.org/WAI/tutorials/images/decorative/) so they have the `aria-hidden` attribute https://github.com/robcresswell/vue-material-design-icons/blob/5.1.2/build.ts#L15
- [x] Remove mandatory title requirement
	- Allow more flexibility in component usage without a title e.g https://github.com/nextcloud/nextcloud-vue/blob/498364d197c5257d758a48853abf87f71ba1d8d2/src/components/DashboardWidget/DashboardWidget.vue#L143-L149

### DIFF
--- a/src/components/EmptyContent/EmptyContent.vue
+++ b/src/components/EmptyContent/EmptyContent.vue
@@ -24,12 +24,14 @@
 ### Basic use
 
 Use this component to display a message about an empty content.
-An icon and a title are mandatory.
-Providing an additional description is strongly advised.
+Providing an icon, title, and a description is strongly advised.
 
 ```
-<EmptyContent icon="icon-comment">
+<EmptyContent>
 	No comments
+	<template #icon>
+		<Comment />
+	</template>
 	<template #desc>No comments in here</template>
 </EmptyContent>
 ```
@@ -37,17 +39,21 @@ Providing an additional description is strongly advised.
 <template>
 	<EmptyContent>
 		Network error
-		<template #icon><Airplane /></template>
+		<template #icon>
+			<Airplane />
+		</template>
 		<template #desc>Unable to load the list</template>
 	</EmptyContent>
 </template>
 
 <script>
 import Airplane from 'vue-material-design-icons/Airplane'
+import Comment from 'vue-material-design-icons/Comment'
 
 export default {
 	components: {
-		Airplane
+		Airplane,
+		Comment,
 	}
 }
 </script>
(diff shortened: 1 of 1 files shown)


Pytal:
Would it make sense to recommend adding the accessible title prop to EmptyContent icons @nimishavijay?

It would show a tooltip like below whenever you hover over the icon

image

src/components/EmptyContent/EmptyContent.vue
```diff
@@ -57,8 +62,8 @@ export default {
 
 <template>
 	<div class="empty-content" role="note">
-		<div class="empty-content__icon" :class="icon" role="img">
-			<!-- @slot Optional icon slot -->
+		<div v-if="hasIcon" class="empty-content__icon">
+			<!-- @slot Mandatory material design icon slot -->
 			<slot name="icon" />
 		</div>
 		<h2 class="empty-content__title">
```
juliushaertl, line 69:
```suggestion
		<h2 v-if="hasTitle" class="empty-content__title">
```

juliushaertl: For https://github.com/nextcloud/nextcloud-vue/issues/2500

I'd probably not make it mandatory then for cases where the empty content is used below any other heading than h1

nimishavijay:
> Would it make sense to recommend adding the accessible title prop to EmptyContent icons?
It would show a tooltip like below whenever you hover over the icon

I would say this image is [decorative](https://www.w3.org/WAI/tutorials/images/decorative/), as all the information required by a person on this page is given through the text. 

If the purpose of a `title` prop here is to make the page screen-reader friendly, I would day we don't need to include it here because there should be a some supporting text under the icon, and that works for a screenreader also :)

cc @jancborchardt

jancborchardt:
Yep, as so often the icon is decorative and has its text there as h2, and thus should have empty alt text `alt=""` to designate that it's decorative and no title to prevent duplication.

juliushaertl (approved):

marcoambrosini (approved):

PR #2868, opened 2022-07-21
Create TextField component

### LINKED ISSUE #2866: Generic reusable and accessible text field
@marcoambrosini 

ref https://github.com/nextcloud/spreed/pull/7491/commits/c2fbf29392e9e981bdc6dd55263ddbb4f8d75e12#diff-a01cbef4930a8836e52c21626ec035892c8348cd361d3bb0c0abac177eaff798R32

- label
- placeholder support
- accessible name
- include optional tick mark (see profile page how fields are verifiable)

marcoambrosini:
fixes #2866

- [x] Optional clear button
- [x] Placeholder
- [x] Optional icon in front? (Do we still do this, or is it only the registration app?)
- [x] Enforce a11y with a label outside of the view port
- [ ] Announce valid input when the checkmark is shown 

https://user-images.githubusercontent.com/26852655/181216991-471388a9-d379-4aa9-ab2f-41802ac694dd.mov


nickvergessen:
Added some todos to the description

nickvergessen:
Icon sample SCSS

https://github.com/nextcloud/registration/pull/439/files#diff-a78ddfa3b59c4afc72dd3bbdfc8c954330bb1362b0972e1af3d7d2c634a7adfeR209-R215

CarlSchwan:
It looks really good, some points:

- I guess a height of 44px would be better: this would pass the AAA WGAC checks and also allow to put of a <Button> and a <TextField> in the same row and make then aligned
- It should be possible to change the type attribute of the field to support number, tel, email, search ...
- Expose some props from input: disabled, max, min, required, minlength, maxlength,readonly, autocomplete, autofocus, ...
- It would be good to have a PasswordField equivalent as a replacement for https://github.com/nextcloud/strengthify but probably best to add it in another separate PR

marcoambrosini:
> I guess a height of 44px would be better: this would pass the AAA WGAC checks and also allow to put of a and a in the same row and make then aligned 


I think that that would be too big for most purposes. I already made it slightly bigger than what we currently have. Also we're aiming at WCAG level 2 for most of our UI elements. 

> It should be possible to change the type attribute of the field to support number, tel, email, search ...


I would do this in separate components in order to have smaller and easier to maintain components. Once we have them we can create a "meta" `inputVue` component that receives the `type` prop, bind the attributes and forward the listeners so that it effectively becomes transparent.

> Make it possible to have a label inside the viewport and visible


Sounds good :) 

> Expose some props from input: disabled, max, min, required, minlength, maxlength,readonly, autocomplete, autofocus, ...


These are already bound because all the attrs are inherited by the inner html `<input />` element. I will add this to the docs though.

> It would be good to have a PasswordField equivalent as a replacement for https://github.com/nextcloud/strengthify but probably best to add it in another separate PR


Sounds good :)

marcoambrosini:
Update: visible label option and added docs about attribute inheritance

Screenshot 2022-07-27 at 16 14 07

webpack.dev.js
```diff
@@ -1,12 +1,12 @@
...
 module.exports = merge(common, {
 	mode: 'development',
 	devServer: {
 		historyApiFallback: true,
 		noInfo: true,
-		overlay: true
+		overlay: true,
 	},
-	devtool: 'source-map',
+	// devtool: 'source-map',
```
marcoambrosini, line 11: - [ ] remember to undo this

CarlSchwan:
> Update: visible label option and added docs about attribute inheritance
> 
> Screenshot 2022-07-27 at 16 14 07

hmm actually I would prefer if the API user could do something like this:

```
<div class="row">
   <label for="$ref.myField.id">My label</label>
   <TextField ref="myField" :label-outside="true"></TextField>
</div>
```

For example, this could be helpful in the personal settings page since we want to have our own label with a button next to it.

marcoambrosini:
What about adding this label icon from within the textfield component? 
I think it's a good idea to foresee the main patterns from within the component and stick to those. @jancborchardt and @nimishavijay do you have a clear picture of how many variants of this we have?

CarlSchwan:
> What about adding this label icon from within the textfield component, as a slot for example? I think it's a good idea to foresee the main patterns in the component design and stick to those. @jancborchardt and @nimishavijay do you have a clear picture of how many variants of this we have?

That would make sense and I would be happy with that too:

What I see as patterns are:

## personal page

image

- A label outside on top with one or two actions
- an additional aria-describeby description at the bottom
- an optional action inside

## theming section

image

- label and text field are side by side

## inline textfield

- placeholder text instead of a label
- Should be avoided when possible and in particular when there is multiple text fields next to each others

image

nickvergessen:
Yeah I also think that for now it would be better if you can pass in a `aria-describedby` only and we keep the label seperate from the input component when it shall be displayed.
Yet another style is used on `index.php/settings/admin` in the Email settings.

marcoambrosini:
- [ ] Receive aria-labelledby as a prop and if that's populated remove inner label 
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby

CarlSchwan:
aria-labelledby is not really a solution as I fear it will be too easy to forget to add the clicking behavior back in Javascript when using the component :(

> Note that while using aria-labelledby is similar in this situation to using an HTML [<label>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label) element with the for attribute, there are some very important differences. The aria-labelledby attribute only defines the accessible name. It doesn't provide any of <label>'s other functionality, such as making clicking on the labeling element activate the input it is associated with. That has to be added back in with JavaScript.

marcoambrosini:
Ok, let's just do what you suggested at the beginning, label outside.

src/components/TextField/TextField.vue
```diff
@@ -0,0 +1,373 @@
...
+and `minlength`.
+
+```
+<template>
+	<div class="wrapper">
+		<TextField :value.sync="text1"
+			label="Type something here"
+			:can-clear="text1 !== ''"
+			@clear="clearText">
+			<Magnify :size="16 " />
```
nickvergessen, line 37:
```suggestion
			<Magnify :size="16" />
```

src/components/TextField/TextField.vue
```diff
@@ -0,0 +1,373 @@
...
+			<div v-if="success" class="text-field__icon text-field__icon--trailing">
+				<Check :size="18" />
+			</div>
+			<!-- clear text button -->
+			<Button v-if="canClear"
+				type="tertiary-no-background"
+				class="text-field__clear-button"
+				@click="clearText">
+				<Close slot="icon"
+					:size="20" />
```
nickvergessen, line 126: I think `<template #icon>` was more future proof?

marcoambrosini:
- [ ] additional style needed in a follow-up https://github.com/nextcloud/forms/pull/1270

src/components/TextField/TextField.vue
```diff
@@ -0,0 +1,405 @@
...
+			label="Type something here"
+			:label-visible="true"
+			@clear="clearText">
+			<Lock :size="16" />
+		</TextField>
+		<div class="external-label">
+			<label for="$refs.textField.id">External label</label>
+			<TextField :value.sync="text4"
+				ref="textField"
+				:label-visible="true"
```
CarlSchwan, line 54:
```suggestion
```

Not needed as far i can tell

CarlSchwan (approved):
Looks good to me 👍

src/components/TextField/TextField.vue
```diff
@@ -0,0 +1,405 @@
...
+				:labelOutside= "true"
```
nickvergessen, line 55:
```suggestion
				:label-outside= "true"
```

src/components/TextField/TextField.vue
nickvergessen: default is false, so its needed?

src/components/TextField/TextField.vue
```diff
@@ -0,0 +1,405 @@
...
+			return this.success
+		},
+
+		hasPlaceholder() {
+			return this.placeholder !== '' && this.placeholder !== undefined
+		},
+
+		computedPlaceholder() {
+			if (this.labelVisible) {
+				return this.placeholder ? this.placeholder : ''
```
nickvergessen, line 255:
```suggestion
				return this.hasPlaceholder ? this.placeholder : ''
```
To be consistent with below?

src/components/TextField/TextField.vue
```diff
@@ -0,0 +1,405 @@
...
+				<slot />
+			</div>
+
+			<!-- Success icon -->
+			<div v-if="success" class="text-field__icon text-field__icon--trailing">
+				<Check :size="18" />
+			</div>
+
+			<!-- clear text button -->
+			<Button v-if="canClear"
```
nickvergessen, line 139:
```suggestion
			<Button v-else-if="canClear"
```

Since they can't be used together make it obvious with if-else?

nickvergessen (commented):
In case of passwords the clear button should be replaced with an Eye-con to show the password instead of ***** ?
Or not part of v1 of the component?

src/components/TextField/TextField.vue
```diff
@@ -0,0 +1,253 @@
...
+			trailing-button-icon="close"
+			:show-trailing-button="text4 !== ''"
+			@trailing-button-click="clearText">
+			<Lock :size="16" />
+		</TextField>
+		<div class="external-label">
+			<label for="$refs.textField.id">External label</label>
+			<TextField :value.sync="text5"
+				ref="textField"
+				:label-outside= "true"
```
nickvergessen, line 63:
```suggestion
				:label-outside="true"
```
**Suggestion  (non-blocking)**

src/components/index.js
```diff
@@ -71,3 +71,5 @@ export { default as RichContenteditable } from './RichContenteditable/index.js'
 export { default as SettingsSection } from './SettingsSection/index.js'
 export { default as UserBubble } from './UserBubble/index.js'
 export { default as Button } from './Button/index.js'
+export { default as InputField } from './InputField/index.js'
```
nickvergessen, line 74: I thought you didn't want to expose this?

src/components/TextField/TextField.vue
```diff
@@ -0,0 +1,253 @@
...
+				'email',
+				'tel',
+				'url',
+			].includes(value),
+		},
+
+		/**
+		 * The hidden input label for accessibility purposes. This will also
+		 * be used as a placeholder unless the placeholder prop is populated
+		 * with a different string.
```
nickvergessen, line 170: **Strongly suggesting** Should state that it's required when labelOutside is not set.

nickvergessen (approved):

src/components/index.js
marcoambrosini: you're right

PR #2875, opened 2022-07-22
Adjust loading icon style, replace `fillColor` prop

raimund-schluessler:
This PR adjusts the style of the vue loading icon to the old style of the loading icon:
Screenshot 2022-07-22 at 22-13-17 Nextcloud Vue Style Guide

It also replaces the `fillColor` with an `appearance` prop. This prevents the usage of arbitrary colors, and allows to select whether the icon adjusts its color to the Nextcloud color scheme, or always shows as `light` or `dark`, independent of the color scheme.

At the moment, the colors for the `light` and `dark` mode are hard-coded, since I didn't find any color variables that don't change with the color scheme. Any hints here are welcome.

Follow-up to #2727.

### DIFF
--- a/src/components/LoadingIcon/LoadingIcon.vue
+++ b/src/components/LoadingIcon/LoadingIcon.vue
@@ -43,8 +43,8 @@
 		<svg :width="size"
 			:height="size"
 			viewBox="0 0 24 24">
-			<path :fill="colorCircle" d="M12,4V2A10,10 0 1,0 22,12H20A8,8 0 1,1 12,4Z" />
-			<path :fill="colorSpinner" d="M12,4V2A10,10 0 0,1 22,12H20A8,8 0 0,0 12,4Z">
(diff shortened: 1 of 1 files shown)


marcoambrosini (approved):

marcoambrosini (commented):
🐘

marcoambrosini:
just a heads up

juliushaertl (approved):

jancborchardt:
Thanks a lot @raimund-schluessler!

Good on using the existing style – not only for consistency reasons with apps which didnt adopt it yet, but also cause that loading icon without full circle is not core Material Design, but a community icon from materialdesignicons.com vs being from https://material.io/icons

How you now fixed it (and how it looked like) looks very nice, and is also used by lots of other platforms like e.g. Twitter or Twitch. :)

jancborchardt (approved):

PR #2877, opened 2022-07-23
Bring back AppNavigationToggle tooltip

### LINKED ISSUE #2876: `AppNavigationToggle` lacks tooltip
After https://github.com/nextcloud/nextcloud-vue/pull/2846 the `AppNavigationToggle` lacks the tooltip (because it now uses a `Button` instead of an `ActionButton`). @nickvergessen

raimund-schluessler:
This brings back the tooltip on `AppNavigationToggle` which got lost with #2846. Closes #2876.

### DIFF
--- a/src/components/AppNavigationToggle/AppNavigationToggle.vue
+++ b/src/components/AppNavigationToggle/AppNavigationToggle.vue
@@ -22,7 +22,8 @@
  -
  -->
 <template>
-	<Button class="app-navigation-toggle"
+	<Button v-tooltip.auto="label"
+		class="app-navigation-toggle"
 		type="tertiary"
 		:aria-expanded="open ? 'true' : 'false'"
 		:aria-label="label"
@@ -36,12 +37,17 @@
 
 <script>
 import Button from '../Button/index.js'
+import Tooltip from '../../directives/Tooltip/index.js'
 import { t } from '../../l10n.js'
 
 import Menu from 'vue-material-design-icons/Menu'
 
 export default {
 	name: 'AppNavigationToggle',
+
+	directives: {
+		tooltip: Tooltip,
+	},
 
 	components: {
 		Button,


src/components/AppNavigationToggle/AppNavigationToggle.vue
```diff
@@ -23,6 +23,7 @@
  -->
 <template>
 	<Button class="app-navigation-toggle"
+		v-tooltip.auto="label"
```
nickvergessen, line 26: GitHub Actions / eslint 

> Attribute "v-tooltip.auto" should go before "class"

src/components/AppNavigationToggle/AppNavigationToggle.vue
```diff
@@ -36,13 +37,18 @@
...
 import { t } from '../../l10n.js'
 
 import Menu from 'vue-material-design-icons/Menu'
 
 export default {
 	name: 'AppNavigationToggle',
 
+	directives: {
+		tooltip: Tooltip,
+	},
```
nickvergessen, line 50: Is this needed? It's not being done in https://github.com/nextcloud/nextcloud-vue/blob/d806d43e6b70364e740d8a00d71df46b547b4a7a/src/components/AppContent/AppDetailsToggle.vue#L24

raimund-schluessler: Good question.

https://github.com/nextcloud/nextcloud-vue/blob/498364d197c5257d758a48853abf87f71ba1d8d2/src/components/Actions/Actions.vue#L248 does it like this. But I will check.

nickvergessen: Maybe the other component only works by accident as every importer does it by accident as they import the Tooltip component:
https://github.com/nextcloud/spreed/blob/master/src/main.js#L50

raimund-schluessler: I tried it with Tasks where we don't separately import the directive with `Vue.use()` and I get an error `[Vue warn]: Failed to resolve directive: tooltip` when I don't import it explicitly here. So, it seems to be needed.

And I guess we should actually also import it explicitly in https://github.com/nextcloud/nextcloud-vue/blob/d806d43e6b70364e740d8a00d71df46b547b4a7a/src/components/AppContent/AppDetailsToggle.vue#L24

raimund-schluessler: I fixed the `AppDetailsToggle` in https://github.com/nextcloud/nextcloud-vue/pull/2887

nickvergessen (approved):

marcoambrosini (approved):

PR #2880, opened 2022-07-24
Make actions use button-vue as single element

raimund-schluessler:
Signed-off-by: Raimund Schlüßler <raimund.schluessler@mailbox.org>


raimund-schluessler:
Superseded by #2911.

Issue #2884, opened 2022-07-25
`MultiselectTags` only displays 5 tags


R0Wi:
### Usecase

Currently i develop a new feature in my app (https://github.com/R0Wi/workflow_ocr/issues/88) where i'd like to use the `MultiselectTags` component to show **all** available tags to the user and let him/her choose **0-n** tags.

### Expected

`MultiselectTags` component should contain all tags from the current system.

### Actual
Because of the line
https://github.com/nextcloud/nextcloud-vue/blob/498364d197c5257d758a48853abf87f71ba1d8d2/src/components/MultiselectTags/MultiselectTags.vue#L69

only the first 5 tags are presented to the user which is quite confusing i think. According to the [docs](https://vue-multiselect.js.org/#sub-getting-started) the prop `options-limit` truncates the available options inside the multiselect dropdown.

raimund-schluessler:
Implemented with #2973.

PR #2885, opened 2022-07-25
Show all tags inside `MultiselectTags`

R0Wi:
Closing #2884

### DIFF
--- a/src/components/MultiselectTags/MultiselectTags.vue
+++ b/src/components/MultiselectTags/MultiselectTags.vue
@@ -66,7 +66,6 @@ export default {
 <template>
 	<Multiselect :value="inputValue"
 		:options="tags"
-		:options-limit="5"
 		:placeholder="label"
 		track-by="id"
 		:custom-label="tagLabel"
--- a/tests/unit/components/MultiselectTags/MultiselectTags.spec.js
+++ b/tests/unit/components/MultiselectTags/MultiselectTags.spec.js
@@ -1,0 +1,62 @@
+/**
+ * @copyright Copyright (c) 2022 Robin Windey <ro.windey@gmail.com>
+ *
+ * @author Robin Windey <ro.windey@gmail.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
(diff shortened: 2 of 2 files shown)


src/components/MultiselectTags/MultiselectTags.vue
```diff
@@ -66,7 +66,6 @@ export default {
 <template>
 	<Multiselect :value="inputValue"
 		:options="tags"
-		:options-limit="5"
```
raimund-schluessler, line 69: I would prefer to keep the ability to limit the tags count and expose it as a prop. If we set the default value of this prop to `undefined` (though I am honestly not sure, this is possible, but let's test), the `slice` method that vue-multiselect uses will just return all tags.

R0Wi: Thank's for your quick feedback. To be honest i think that a prop `limit` is not very useful for showing a tag list. The server response would be cut at an arbitrary point, resulting in inconsistent data presented to the user.

If we need a possibility to filter the results received by the server, i'd go with an optional `tagsFilter` prop of type `function` which receives the full tag list and returns a filtered one. A "limt" function could then be easily implementer, too. What do you think?

raimund-schluessler: > If we need a possibility to filter the results received by the server, i'd go with an optional `tagsFilter` prop of type `function` which receives the full tag list and returns a filtered one. A "limt" function could then be easily implementer, too. What do you think?

I am not so keen on making this much more complicated than it is now. I just think that the limit was there for a reason and removing it completely might break some uses cases. So keeping the possibility to limit the tag count while allowing to show all seems like a good compromise to me (it will still be a breaking change if we remove the limit by default).

R0Wi: I see your point but i really wonder why one should limit the displayed options since there's automatic scrolling happening if there are more options than we can display. 

But ok for compatibility reasons: let's introduce a `limit` prop with default value `5` and let's check if it shows all available options when setting this prop to `null`?

R0Wi: > But ok for compatibility reasons: let's introduce a limit prop with default value 5 and let's check if it shows all available options when setting this prop to null?

@raimund-schluessler @skjnldsv would really love to get some feedback on this. I think with that solution we would not break anything and i could realize my usecase for my app so this would be quite important to me. If that's a desirable solution i can change the code accordingly 👍

raimund-schluessler: I am not the person in charge of deciding that 😉
But from my side that sounds like a sane approach. I just guess you need to set the prop to `undefined` instead of `null` to get all tags (judging from what I think how `Array.slice` works)

raimund-schluessler:
@R0Wi It turns out your idea of implementing a filter function is indeed necessary, see #2973 and might be the best way to solve your issue.

R0Wi:
Can be closed because the filter functionality was implemented in https://github.com/nextcloud/nextcloud-vue/pull/2973

PR #2886, opened 2022-07-25
Add HeaderMenu component

CarlSchwan:
This component is used for the notification and full-text search icon in
the navbar.

TODOS:

- [ ] Add doc
- [ ] Currently app can't use it and register their own menu item without patching server first, decide if this is important


src/components/HeaderMenu/HeaderMenu.vue
```diff
@@ -0,0 +1,224 @@
...
+	directives: {
+		ClickOutside,
+	},
+
+	mixins: [
+		excludeClickOutsideClasses,
+	],
+
+	props: {
+		id: {
```
raimund-schluessler, line 64: This could use some documentation, so it is shown nicely in the docs.

src/components/HeaderMenu/HeaderMenu.vue
```diff
@@ -0,0 +1,224 @@
...
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+<template>
+	<div :id="id"
+		v-click-outside="clickOutsideConfig"
+		:class="{ 'header-menu--opened': opened }"
+		class="header-menu">
+		<a class="header-menu__trigger"
```
raimund-schluessler, line 27: I don't really know how the component is supposed to work/behave, but this looks like it could be a vue button, no?

marcoambrosini (commented):
The docs fail to build. Could you fix that or add screenshots?

PR #2890, opened 2022-07-26
Emit click event from checkboxradioswitch component

marcoambrosini:
For some reason a click event is not emitted when clicking on the
checkboxradioswitch component. This commit brings this functionality to
the component.


skjnldsv (changes requested):
Please explain the reasoning behind needing to watch for a click?

PR #2894, opened 2022-07-27
Add more explicit title to the sidebar close button

juliushaertl:
To make sure that a screen reader user has some context on what the button actually does


raimund-schluessler (changes requested):
Needs l10n extract.

marcoambrosini (approved):

artonge (approved):

Issue #2891, opened 2022-07-26
Multiselect produces invalid HTML


nickvergessen:
Ref https://github.com/nextcloud/spreed/issues/7651

### Error: Bad value for attribute name on element input: Must not be empty.

This was brought in upstream:
https://github.com/shentao/vue-multiselect/blob/2.1.6/src/Multiselect.vue#L157-L165

But is not named or piped through by the Vue lib.

Fixable with a patch in Talk just adding something like:
```html
name="default_group_notification"
```

- [ ] Maybe we should add this as required prop or fall back to the id value?

---

### Error: Bad value nope for attribute autocomplete on element input: The string nope is not a valid autofill field name.

https://github.com/shentao/vue-multiselect/blob/2.1.6/src/Multiselect.vue#L51

Upstream issue again, but changed to `off` in master:
https://github.com/shentao/vue-multiselect/pull/1064

But there was no release since: https://github.com/shentao/vue-multiselect/issues/1518

---

### Error: Element div not allowed as child of element span in this context. (Suppressing further errors from this subtree.)

Comes from
https://github.com/nextcloud/nextcloud-vue/blob/a08bfb147a970a34abd0881ecd305f5f78b699cc/src/components/Multiselect/EllipsisedOption.vue#L24

So would need someone to look into it and transform it to a span or something.

- [x] #2895

nickvergessen:
@PVince81 any opinion on first and second Error?

nickvergessen:
Seems to be fixed now, at least talk validation passes

PR #2895, opened 2022-07-27
Fix issues with Multiselect component

nickvergessen:
Fix step 3 of https://github.com/nextcloud/nextcloud-vue/issues/2891

1. ❌ Error: Element div not allowed as child of element span in this context. (Suppressing further errors from this subtree.)
2. Loading icon background color issue

Before | After
---|---
Peek 2022-07-27 10-39 | Peek 2022-07-27 10-38

### DIFF
--- a/src/components/Multiselect/EllipsisedOption.vue
+++ b/src/components/Multiselect/EllipsisedOption.vue
@@ -21,7 +21,7 @@
   -->
 
 <template>
-	<div class="name-parts" :title="name">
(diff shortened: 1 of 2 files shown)


raimund-schluessler (approved):
Fine with me. But where does the wrong background of the loading icon come from?

nickvergessen:
okay.. thanks for resisting one more time.
Now I found the correct spot

raimund-schluessler (approved):

Issue #2755, opened 2022-06-10
Expandable buttons should have aria-expanded attribute


JuliaKirschenheuter:
Accessibility issue.

All expandable buttons (mostly in a sidebar) should have `aria-expanded` attribute.

Screen reader users should be told of this function if the area is open or not.

![image](https://user-images.githubusercontent.com/6078378/173071258-f31e01c3-56e1-42b5-a759-3026aff59c6b.png)

![image](https://user-images.githubusercontent.com/6078378/173071307-d6c1d2ad-e8ac-4a14-ad0c-b6da4974dbfa.png)

PR #2898, opened 2022-07-28
Add aria-expended attribute

JuliaKirschenheuter:
Add aria-expended attribute to link
Correct aria-expended attribute of button in Actions.vue

Fixes https://github.com/nextcloud/nextcloud-vue/issues/2755

### DIFF
--- a/src/components/Actions/Actions.vue
+++ b/src/components/Actions/Actions.vue
@@ -190,7 +190,7 @@ export default {
 					aria-haspopup="menu"
 					:aria-label="ariaLabel"
 					:aria-controls="randomId"
-					:aria-expanded="opened ? 'true' : 'false'"
+					:aria-expanded="opened.toString()"
 					type="button"
 					@focus="onFocus"
 					@blur="onBlur">
--- a/src/components/AppNavigationItem/AppNavigationItem.vue
+++ b/src/components/AppNavigationItem/AppNavigationItem.vue
@@ -130,6 +130,7 @@ Just set the `pinned` prop.
 				class="app-navigation-entry-link"
 				:aria-description="ariaDescription"
 				href="#"
+				:aria-expanded="opened.toString()"
 				@click="onClick">
 
 				<!-- icon if not collapsible -->


marcoambrosini (approved):

src/components/Actions/Actions.vue
```diff
@@ -190,7 +190,7 @@ export default {
 					aria-haspopup="menu"
 					:aria-label="ariaLabel"
 					:aria-controls="randomId"
-					:aria-expanded="opened ? 'true' : 'false'"
+					:aria-expended="opened.toString()"
```
marcoambrosini, line 193:
```suggestion
					:aria-expanded="opened.toString()"
```

src/components/AppNavigationItem/AppNavigationItem.vue
```diff
@@ -130,6 +130,7 @@ Just set the `pinned` prop.
 				class="app-navigation-entry-link"
 				:aria-description="ariaDescription"
 				href="#"
+				:aria-expended="opened.toString()"
```
marcoambrosini, line 133:
```suggestion
				:aria-expanded="opened.toString()"
```

marcoambrosini (changes requested):

marcoambrosini:
Almost missed that :)

src/components/AppNavigationItem/AppNavigationItem.vue
JuliaKirschenheuter: Thank you Marco) sorry, typo

marcoambrosini (approved):

raimund-schluessler (approved):
Just a question out of curiosity: does the value really need to be a string `true` or can it also be a Boolean `true`? That's not clear to me from the W3 docs.

JuliaKirschenheuter:
> Just a question out of curiosity: does the value really need to be a string `true` or can it also be a Boolean `true`? That's not clear to me from the W3 docs.

HTML attributes have the following features:

- Their name is case-insensitive (id is same as ID).
- Their values are always strings.

https://javascript.info/dom-attributes-and-properties

Issue #2824, opened 2022-07-12
Accessibility: make sure, that the focus of all interactive elements is consistent


JuliaKirschenheuter:
This is an accessibility issue.

Some elements on the page have not the same focus styles. There are some examples: 

![image](https://user-images.githubusercontent.com/6078378/178430441-f6a82214-5818-4366-8969-c62f6f3c0885.png)
![image](https://user-images.githubusercontent.com/6078378/178430501-2de950b2-d86b-4752-b1d6-d6d972c055d6.png)
![image](https://user-images.githubusercontent.com/6078378/178430620-7065d9a6-67b0-4180-add8-8bec45bacb72.png)
![image](https://user-images.githubusercontent.com/6078378/178430736-25b2b6c9-c74d-479b-aa98-dab3bac4d9a5.png)
![image](https://user-images.githubusercontent.com/6078378/178430886-54917dab-b1fa-43c2-adb5-7f37e4c0d501.png)
![image](https://user-images.githubusercontent.com/6078378/178431279-8cbe27cb-d5f8-45a3-a647-9dea7647bca0.png)
![image](https://user-images.githubusercontent.com/6078378/178431427-4410bd6e-722c-43de-9c49-44f4725139c7.png)

Try to make sure, that all interactive elements have the same focus on a NC Vue level.

Fenn-CS:
@nextcloud/designers Please, can you suggest a color for the focus? So that the focus color can be defined as a variable in nextcloud core css?

I have seen more use of
```css
outline: 2px solid var(--color-main-text) !important;
```

Example : https://github.com/nextcloud/nextcloud-vue/pull/2899

`--color-main-text` is darkish and I wonder if it matches the theme design. Even with the growing use, there are still places where different colors and border sizes are used for focus hence the need for us to pick a color and introduce two new scss variables for example `--color-action-focus`, `--border-size-focus`

jancborchardt:
@Fenn-CS so Airbnb for example uses black+white so it works on any background, any theme, even images:
![image](https://github.com/nextcloud/nextcloud-vue/assets/925062/10737238-4887-4ed5-8aa8-def9a2075bcb)
This seems best, if possible?

jancborchardt:
@marcoambrosini at least this is what I also remember you suggested? cc @nextcloud/designers too.

nimishavijay:
> Airbnb for example uses black+white so it works on any background, any theme, even images:

[...]

PR #2899, opened 2022-07-28
Button accessibility improvements

marcoambrosini:
Bunch of simplification and accessibility fixes for the Button component.

- Removed unnecessary JS logic for keyboard navigation (apparently adding `!important` always activates the `focus-visible` when navigating with the keyboard)
- Removed the lightening of backgrounds as `focus` feedback. There's already a huge outline, so best avoid reducing contrast
- Removed opacity change for `tertyary-no-background` buttons. These are buttons whose use we should limit for very specific use-case such as within an input field. The only hover feedback will be the `cursor` change to `pointer`, while for keyboard navigation we have the beefy outline here too.

https://user-images.githubusercontent.com/26852655/181483901-2d1a3155-d6a0-415a-b0db-2bb3b92dc0e2.mov


src/components/Button/Button.vue
```diff
@@ -622,22 +548,18 @@ export default {
...
 			background-color: transparent;
-			opacity: 1;
 		}
 	}
 
 	// Tertiary on primary color (like the header)
 	&--vue-tertiary-on-primary {
 		color: var(--color-primary-text);
 		background-color: transparent;
-		opacity: .7;
```
nickvergessen, line 636: This is then not inline with how default app icons in the topbar look.
Can we skip this from here until the decision about app icons was done?

marcoambrosini: I think it was stated multiple times in in various design calls that we're going to abandon this, cc @jancborchardt & @nimishavijay

nickvergessen: The app list in the topbar currently does not use this button, so changing this will make it inconsistent, unless the server is adjusted in parallel. So from my POV we should wait for the outcome and then do it in a follow up

marcoambrosini: This shouldn't be related to that special case. See comment there

src/components/Button/Button.vue
```diff
@@ -548,36 +497,13 @@ export default {
...
-			background-color: var(--color-primary-hover);
-		}
-		&.button-vue--vue-secondary {
-			box-shadow: 0 0 0 2px var(--color-main-text);
-		}
-		&.button-vue--vue-tertiary-no-background {
-			opacity: 1;
-		}
+	&:focus-visible {
+		outline: 2px solid var(--color-main-text) !important;
```
juliushaertl, line 501: On server we are using a primary color outline for a lot of elements, so wondering if using main-text is intended here?

juliushaertl: Ref: https://github.com/nextcloud/server/pull/31584

marcoambrosini: Yes this has been the case for the button component since the beginning

src/components/Button/Button.vue
marcoambrosini: Added special case for tertiary no bg

JuliaKirschenheuter (approved):
tested on mail - works! Thank you!

nickvergessen:
> Error: The "AppSidebar.vue-subtitle_null-starred_false-compact_true-header_none-secondary_none-editable_true" image is different. Threshold limit exceeded! 

Lots of issues

src/components/Button/Button.vue
juliushaertl: All right, just seen this is tracked separately with https://github.com/nextcloud/nextcloud-vue/issues/2824

nickvergessen:
> Removed unnecessary JS logic for keyboard navigation (apparently adding !important always activates the focus-visible when navigating with the keyboard)

Should be fixed now that the server doesn't simply touch everything anymore:
https://github.com/nextcloud/server/pull/33497

marcoambrosini:
> Should be fixed now that the server doesn't simply touch everything anymore:

Still, all that is not needed anymore.

marcoambrosini:
Let's get this in?

webpack.dev.js
```diff
@@ -0,0 +1,12 @@
+const { merge } = require('webpack-merge');
```
nickvergessen, line 1: This file intended?

marcoambrosini: no, it's been removed from master, let me delete

skjnldsv (approved):
Isn't the black border a tad too strong? :)
Otherwise fine by me

skjnldsv:
Rebasing...

PR #2910, opened 2022-08-02
Add NoteCard component

CarlSchwan:
image

Already used in the login pages and in a few settings


jancborchardt (changes requested):
Nice! :) Details:
- [ ] I’d probably cut this border, then it looks cleaner? Also would then look more like the callout boxes in Text
- [ ] Similar with the box-shadow – we don’t really use shadow a lot, and it doesn’t make a lot of sense for this container (we don’t use it for the Text callout boxes either)
- [ ] Is the text color accessible / readable on all the possible backgrounds? 0.2 opacity is used, whereas for color-primary-light we use 10%, so we should probably adjust the values here to 0.1 as well so it’s a bit less intense?

jancborchardt:
Oh and also, this seems like it can be aligned with the notification component? Basically same thing: Container and bar on the left, just one is placed in the content, one on the top right with an x button (and possibly other actions like "undo").

CarlSchwan:
> Nice! :) Details:
> 
>     * [ ]  I’d probably cut this border, then it looks cleaner? Also would then look more like the callout boxes in Text
> 
>     * [ ]  Similar with the box-shadow – we don’t really use shadow a lot, and it doesn’t make a lot of sense for this container (we don’t use it for the Text callout boxes either)
> 
>     * [ ]  Is the text color accessible / readable on all the possible backgrounds? 0.2 opacity is used, whereas for color-primary-light we use 10%, so we should probably adjust the values here to 0.1 as well so it’s a bit less intense?

The text color is accessible with 0.20 and I got a contrast ratio of >11. But I experimented a bit with the various transparency level and I kinda like 15% more:

15%

image

10%

nickvergessen:
Can you also make dark mode screenshots? Would like to see how that turns out.

jancborchardt:
Looks good @CarlSchwan! :) I’d go for 10% so it’s in line with our usage of the primary-light (as will the new design for tags and avatars), and also because it looks much nicer than the slightly darker background colors.

vinicius73:
I guess we can make it closer with `text app callouts` in some way.

callouts

CarlSchwan:
I updated the screenshot

src/components/NcNoteCard/NcNoteCard.vue
```diff
@@ -0,0 +1,156 @@
...
+		},
+	},
+}
+</script>
+
+<style lang="scss" scoped>
+.notecard {
+	color: var(--color-text-light) !important;
+	background-color: var(--note-background) !important;
+	border-left: 4px solid var(--note-theme);
```
tcitworld, line 127:
```suggestion
	border-inline-start: 4px solid var(--note-theme);
```
If you want to support rtl languages.

jancborchardt (approved):
Looks great! :) And now this we can also use for the login screen for the "Wrong account name or password" notice

src/components/NcNoteCard/NcNoteCard.vue
```diff
@@ -0,0 +1,156 @@
...
+export default {
+	name: 'NcNoteCard',
+
+	props: {
+		/**
+		 * Type of the attribute
+		 */
+		type: {
+			type: String,
+			default: () => 'warning',
```
raimund-schluessler, line 79: Doesn't need to be a function
```suggestion
			default: 'warning',
```

src/components/NcNoteCard/NcNoteCard.vue
```diff
@@ -0,0 +1,156 @@
...
+			validator: type => type === 'success' || type === 'warning' || type === 'error',
+		},
+		showAlert: {
+			type: Boolean,
+			default: () => false,
```
raimund-schluessler, line 84:
```suggestion
			default: false,
```

src/components/NcNoteCard/NcNoteCard.vue
```diff
@@ -0,0 +1,156 @@
...
+		},
+		heading: {
+			type: String,
+			required: false,
```
raimund-schluessler, line 88:
```suggestion
			default: '',
```

src/components/NcNoteCard/NcNoteCard.vue
```diff
@@ -0,0 +1,156 @@
...
+	name: 'NcNoteCard',
+
+	props: {
+		/**
+		 * Type of the attribute
+		 */
+		type: {
+			type: String,
+			default: () => 'warning',
+			validator: type => type === 'success' || type === 'warning' || type === 'error',
```
raimund-schluessler, line 80:
```suggestion
			validator: type => ['success', 'warning', 'error'].includes(type),
```

raimund-schluessler (approved):

Issue #2902, opened 2022-07-28
[6.0.0alpha] Actions infinite loop regression


PVince81:
See https://github.com/nextcloud/server/pull/33387#issuecomment-1197137774

@Pytal can you investigate ?

PVince81:
I suggest to first revert all changes in AppSidebar to see if the problem persists, in case it's only a side effect.

Pytal:
Findings as follows

The infinite loop is not triggered in AppSidebar but rather in Actions

The recent changes which touch related code are these fixes https://github.com/nextcloud/nextcloud-vue/pull/2878 and https://github.com/nextcloud/nextcloud-vue/pull/2532

With the `isValidSingleAction` fix the markup https://github.com/nextcloud/nextcloud-vue/blob/79f3c499f14c8d188c93da2258bb690977bfa8f3/src/components/Actions/Actions.vue#L135-L159 which depends on `firstAction` renders as expected but because children change at some point during the render lifecycle the watcher https://github.com/nextcloud/nextcloud-vue/blob/79f3c499f14c8d188c93da2258bb690977bfa8f3/src/components/Actions/Actions.vue#L477-L488 mutates `firstAction` while rendering leading to the "infinite update loop in a component render function" error, commenting out the assignment removes the error

Wrapping it in `$nextTick` doesn't seem to work so any suggestions here are welcome @PVince81

PVince81:
@vinicius73 @raimund-schluessler @marcoambrosini please help solve the infinite loop in the `Actions` component. I picked you as you are connected to the matching PRs.

raimund-schluessler:
> Findings as follows
> 
> 
> 
> The infinite loop is not triggered in AppSidebar but rather in Actions
> 
> 
> 
> The recent changes which touch related code are these fixes https://github.com/nextcloud/nextcloud-vue/pull/2878 and https://github.com/nextcloud/nextcloud-vue/pull/2532
> 
> 
> 
[...]

PR #2911, opened 2022-08-02
Use render function in Actions component

raimund-schluessler:
This PR aims at fixing #2902. It changes the `Actions` component to use a render function instead of a `template` and introduces the `Button` vue component as popover trigger and single action button. #2902 is fixed for me when checking e.g. on the Tasks dashboard. I hope that I have correctly re-implemented all features of the Actions component, but this needs extensive testing, especially since this component is so widely used throughout Nextcloud.

Since `Actions` now uses `Button`, there are some visual changes to the `Actions` trigger, which align the appearance to the `Button` component (and make the Cypress test fail for now). However, I don't know if this is wanted for the `Actions` component. I guess, that's a question for @jancborchardt and @nimishavijay. I don't have any before/after screenshots, because I think it's best viewed in the docs:
Before: https://nextcloud-vue-components.netlify.app/#/Components/Actions?id=actions-1
After: https://deploy-preview-2911--nextcloud-vue-components.netlify.app/#/Components/Actions?id=actions-1

~The only issue known to me (yet) is, that the `close-after-click` prop of e.g. the `ActionButton` introduced in #622 does not work in the production build, because `this.$parent` of the `ActionButton` is `null` (for reasons unknown to me). With a dev build, like in the docs, it works just fine. Any hint here would be very welcome.~
I just noticed that this is broken on master too when using `npm link`, `close-after-click` has no effect anymore. It works with `6.0.0-alpha.0` installed via `npm` though. So I guess this is either an issue with `npm link` or this is a "feature"/bug of vue 2.7.8 which we bumped in #2929. I will check and create a separate issue for this if necessary.


src/components/Actions/Actions.vue
```diff
@@ -662,6 +598,72 @@ export default {
...
+
+		/**
+		 * If we have a single action only and didn't force a menu,
+		 * we render the action as a standalone button
+		 */
+		if (this.isValidSingleAction && !this.forceMenu) {
+			console.debug('render single action')
+			return h('Button',
+				{
+					class: 'action-item action-item--single',
```
juliushaertl, line 624:
```suggestion
					class: 'action-item action-item--single button-vue',
```

Haven't tested fully but seems this is needed to avoid leaking server styles

src/components/Actions/Actions.vue
```diff
@@ -662,6 +598,72 @@ export default {
...
+				[this.singleActionTitle],
+			)
+		/**
+		 * Otherwise, we render the actions in a popover
+		 */
+		} else {
+			console.debug('render multiple actions')
+			return h('div',
+				{
+					class: 'action-item',
```
juliushaertl, line 635:
```suggestion
					class: 'action-item button-vue',
```

Same as above

src/components/Actions/Actions.vue
raimund-schluessler: Yes, there is still a lot of stuff missing and styles are not correct yet. I will fix these "details" once it's clear it works in principle and fixes the problems targeted. I clarified this in the first post.

raimund-schluessler:
This PR is ready for a first round of reviews.

skjnldsv (approved):
Thank you so much for this!!!
This is sooo much cleaner and faster!

Tested reactivity and the various supported configs, couldn't find an issue! :rocket:

marcoambrosini (approved):
Tested in the docs and work, code looks good to me too. I can't test in Talk because everything is still messed up by the floating vue stuff.
@raimund-schluessler can we create a separate pr with the cypress fix and get that in asap?

raimund-schluessler:
> @raimund-schluessler can we create a separate pr with the cypress fix and get that in asap?

What cypress fix do you mean?

marcoambrosini:
Oh nvm, I thought you fixed the snapshots that were failing on master here

raimund-schluessler:
> Oh nvm, I thought you fixed the snapshots that were failing on master here

No, I just had to update the snapshots, because we use `Actions` in the sidebar and it's appearance changed, because we use the `Button` component now for the popover trigger and single action button.

The snapshots look fine on master, though.

nimishavijay (approved):
>  However, I don't know if this is wanted for the Actions component. 

Consistency for similar functionality is a good thing! It looks great now (especially the keyboard a11y 💯)

raimund-schluessler:
@marcoambrosini Can we do a `6.0.0-alpha.1` release after this is merged? Since it's a big change, I wouldn't want to go for a final release immediately.

marcoambrosini:
> Can we do a 6.0.0-alpha.1 release after this is merged? Since it's a big change, I wouldn't want to go for a final release immediately.

@raimund-schluessler sounds good but don't we want this in v5.5.0 too?

raimund-schluessler:
> @raimund-schluessler sounds good but don't we want this in v5.5.0 too?

I wouldn't backport this. It's a quite big change, and it relies / uses changes that are only present in master / 6.0.0 e.g. the migration to floating-vue (which is breaking).
Also, I didn't implement icon-classes for the actions trigger and single action, since `Button` does not really support it (you could put it in the slot, though) and I thought we anyway discourage the usage of icon-classes with 6.0.0.

PR #2924, opened 2022-08-03
[stable5] add role & aria-labels for accessibility

nickvergessen:
Backport of #2696

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -108,6 +108,9 @@
 			class="modal-mask"
 			:class="{ 'modal-mask--dark': dark }"
 			:style="cssVariables"
+			role="dialog"
+			aria-labelledby="modal-title"
+			aria-describedby="modal-description"
 			@click="handleMouseMove"
 			@mousemove="handleMouseMove"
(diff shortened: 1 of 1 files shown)


src/components/Modal/Modal.vue
```diff
@@ -118,7 +121,9 @@
 						invisible: clearView
 					}"
 					class="modal-header">
-					<div v-if="title.trim() !== ''" class="modal-title">
+					<div v-if="title.trim() !== ''"
+						id="modal-title"
```
raimund-schluessler, line 125: This id might not be unique if we have multiple modals on the page.

src/components/Modal/Modal.vue
```diff
@@ -202,7 +207,7 @@
 					</transition>
 
 					<!-- Content -->
-					<div class="modal-container">
+					<div id="modal-description" class="modal-container">
```
raimund-schluessler, line 210: Might also not be unique, no?

raimund-schluessler (approved):
The backport itself is fine, but I think the original issue needs a follow-up. I commented there.

PR #2927, opened 2022-08-03
Use material design icons in actions docs

raimund-schluessler:
In an effort to encourage the usage of material design icons, I replaced the old icon classes in the Actions component docs with material design icons.


nickvergessen (approved):

GretaD (approved):
code looks good

PR #2929, opened 2022-08-03
Bump vue from 2.6.14 to 2.7.8

raimund-schluessler:
This bumps vue to 2.7.8.

This fixes an issue when you use `npm link` with an app that already uses `vue@2.7.8` (e.g. Tasks). Linking then fails with an obscure "vm._scope is undefined" error. Bumping vue here as well makes it work again. Installing nextcloud/vue via npm works in any case.

PR #2935, opened 2022-08-04
Update cypress snapshots

marcoambrosini:


raimund-schluessler:
I don't think that this is necessary. The cypress check ran just fine on GitHub (and now fails with the updated snapshots).

@marcoambrosini The problem you experience is probably that the tests fail locally for you (as well as for me). This is due to a different environment on GitHub and our local systems which makes the snapshots look subtly different. I didn't really figure out a way to solve this yet (besides running it in a Docker container both remote and locally, but this was quite a bit more complex than what we have currently).

So for now, out ground truth for the snapshots has to be the remote GitHub check. If you want them updated, you need to run the GitHub cypress check, download the new snapshots and commit them.

PVince81:
tests passed fine on https://github.com/nextcloud/nextcloud-vue/pull/2929 a few minutes ago, and this was rebased on master, so the snapshots on master must be fine

PR #2939, opened 2022-08-04
Add NcPasswordField component

CarlSchwan:
image


src/components/InputField/InputField.vue
```diff
@@ -262,10 +281,16 @@ export default {
...
+			&:focus {
+				box-shadow: 0 0 5px rgba(81, 203, 238, 1);
+			}
 		}
 
 		&--error {
 			border-color: var(--color-error) !important; //Override hover border color
+			&:focus {
+				box-shadow: 0 0 5px rgba(81, 203, 238, 1);
+			}
```
marcoambrosini, line 293: About this, could we actually not allow to have focus and success and error borders at the same time? 
I think that success and error borders should be toggled on only when the component is being actively used. So there's no need for the focus state on top of it.

marcoambrosini:
Screenshot 2022-08-05 at 09 31 41

Also button and icon should not be displayed at the same time there

src/components/InputField/InputField.vue
```diff
@@ -52,13 +54,15 @@
...
+			<div v-if="success || error"
+				class="input-field__icon input-field__icon--trailing"
+				:class="{'input-field__icon--trailing--with-button': success || error }">
 				<Check v-if="success" :size="18" />
 				<AlertCircle v-else-if="error" :size="18" />
 			</div>
 
 			<!-- trailing button -->
-			<Button v-else-if="showTrailingButton"
+			<Button v-if="showTrailingButton"
```
marcoambrosini, line 65:
```suggestion
			<Button v-else-if="showTrailingButton"
```

See comment above

CarlSchwan:
> Screenshot 2022-08-05 at 09 31 41
> 
> Also button and icon should not be displayed at the same time there. That was the purpose of the `v-else-if`

We kinda need to have both at the same time or at least have the button displayed instead of the icon. The issue is that when there is an error, it can be helpful to look at the clear text password and see the error so that the user can correct it.

marcoambrosini:
> at least have the button displayed instead of the icon

Ok let's give priority to the button, we can keep the v-else-if but swap the elements order

CarlSchwan:
New screnshot

image

nimishavijay:
Looks great! If the `!` icon is to indicate that the password is weak, then it can go on the left of the error message so that it's like "❗Password is insecure."

2 questions:
- Would it be possible to also specify why the password is insecure? Eg. Must contain numbers etc 
- If the password is secure could there be some indication, like the border color changing to `--color-success` or a message that says "✔️ Password is secure" or something like that

CarlSchwan:
New screenshot

image

The text is hardcoded in the example but in real life, if the `checkPasswordStrength` strength is enabled (by default disabled), it will use the text coming from the password_policy app

src/components/PasswordField/PasswordField.vue
```diff
@@ -0,0 +1,320 @@
...
+			default: 'current-password',
+		},
+
+		/**
+		 * Check if the user entered a valid password using the password_policy
+		 * app if available.
+		 *
+		 * Warning: this doesn't replace server side checking and will do nothing
+		 * if the password_policy app is disabled.
+		 */
```
marcoambrosini, line 259: Can we throw an error if this is `true` and the password_policy app is disabled?

src/components/PasswordField/PasswordField.vue
```diff
@@ -0,0 +1,320 @@
...
+	data() {
+		return {
+			isPasswordHidden: true,
+			internalErrorMessage: '',
+			passwordPolicy: null,
+		}
+	},
+
+	computed: {
+		computerErrorMessage() {
```
marcoambrosini, line 283: computed

src/components/PasswordField/index.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * @copyright Copyright (c) 2022 Marco Ambrosini <marcoambrosini@pm.me>
+ *
+ * @author Marco Ambrosini <marcoambrosini@pm.me>
```
marcoambrosini, line 4: Carl :)

webpack.dev.js
```diff
@@ -8,5 +8,4 @@ module.exports = merge(common, {
 		noInfo: true,
 		overlay: true,
 	},
-	devtool: 'source-map',
 })
```
marcoambrosini, line 11: remember to take this out

marcoambrosini (changes requested):

jancborchardt:
Looks good component-wise now. Can we also make the "New password" not have a line break there in the docs so it looks better? :)

src/components/InputField/InputField.vue
```diff
@@ -262,10 +285,16 @@ export default {
...
+			&:focus-visible {
+				box-shadow: rgb(248, 250, 252) 0px 0px 0px 2px, var(--color-primary-element) 0px 0px 0px 4px, rgba(0, 0, 0, 0.05) 0px 1px 2px 0px
+			}
 		}
 
 		&--error {
 			border-color: var(--color-error) !important; //Override hover border color
+			&:focus-visible {
+				box-shadow: rgb(248, 250, 252) 0px 0px 0px 2px, var(--color-primary-element) 0px 0px 0px 4px, rgba(0, 0, 0, 0.05) 0px 1px 2px 0px
+			}
```
marcoambrosini, line 297: So I think that we need to standardise how we do focus feedbacks across the board. For the button we chose to use `--color-main-text` because it provides the most contrast with whatever `--colora-main-background` is. Also we considered 2px of background color between the element's border and the focus feedback border but we thought that only "1 border" was sufficient. Whatever we decide, we should decide once and for all and apply it to all or most elements.

nimishavijay (approved):

src/components/PasswordField/PasswordField.vue
```diff
@@ -0,0 +1,318 @@
...
+		</template>
+	</InputField>
+</template>
+
+<script>
+
+import Eye from 'vue-material-design-icons/Eye'
+import EyeOff from 'vue-material-design-icons/EyeOff'
+import InputField from '../InputField/InputField.vue'
+import _ from 'lodash'
```
vinicius73, line 107: We should avoid to import full `lodash` package.
Instead of it, we can import only what we need.

```js
import { debounce } from 'lodash'
```

Besides that, it is highly recommended use [`lodash-es`](https://www.npmjs.com/package/lodash-es) instead of `lodash` .
It will make possible [`tree shaking`](https://webpack.js.org/guides/tree-shaking/) and improve the bundle size.

src/components/PasswordField/PasswordField.vue
```diff
@@ -0,0 +1,318 @@
...
+	},
+
+	methods: {
+		handleInput(event) {
+			this.$emit('update:value', event.target.value)
+		},
+		togglePasswordVisibility() {
+			this.isPasswordHidden = !this.isPasswordHidden
+		},
+		checkPassword: _.debounce(function(password) {
```
vinicius73, line 306: We already have the `debounce` package available.

https://github.com/nextcloud/nextcloud-vue/blob/26eafd2078f7d3ec624441c60eca7cae5c0ce782/package.json#L49

If is the only function what we are using from `lodash` we do not need a new package.

src/components/PasswordField/PasswordField.vue
```diff
@@ -0,0 +1,318 @@
...
+			axios.post(generateOcsUrl('apps/password_policy/api/v1/validate'), { password })
+				.then(res => res.data.ocs.data)
+				.then(data => {
+					if (data.passed) {
+						this.internalErrorMessage = t('Password is secure')
```
vinicius73, line 311: Using an error state to represent a success doesn't seem right.

src/components/PasswordField/PasswordField.vue
```diff
@@ -0,0 +1,318 @@
...
+			}
+			return null
+		},
+	},
+
+	watch: {
+		value(newValue) {
+			if (this.checkPasswordStrength) {
+				if (this.passwordPolicy === null) {
+					this.passwordPolicy = loadState('core', 'capabilities').password_policy
```
vinicius73, line 287: We can "preload" this value and avoid "recalculate" it every time.
Just move it to outside of component.

```js
const defaultPasswordPolicy = loadState('core', 'capabilities').password_policy

// -- 

this.passwordPolicy = defaultPasswordPolicy
```

src/components/PasswordField/PasswordField.vue
```diff
@@ -0,0 +1,318 @@
...
+
+	computed: {
+		computedHelperText() {
+			if (this.helperText.length > 0) {
+				return this.helperText
+			}
+			return this.internalErrorMessage
+		},
+		computedMinlength() {
+			if (this.$props.minlength) {
```
vinicius73, line 273: We should avoid use `this.$props.`
You can use `this.minlength` directly.

In cases like this, what we have some "logic" in to props values, you can join more computed in only one.

```js
{
  computed: {
    rules () {
      const { minlength, passwordPolicy } = this

      return {
         minlength: minlength ?? passwordPolicy?.minLength,
      }
    }
  }
}
```

Now we are able to use `rules.minlength` safety.

vinicius73 (changes requested):

src/components/PasswordField/PasswordField.vue
```diff
@@ -0,0 +1,318 @@
...
+	<InputField v-bind="$props"
+		:type="isPasswordHidden ? 'password' : 'text'"
+		:show-trailing-button="true"
+		:helper-text="computedHelperText"
+		v-on="$listeners"
+		@trailing-button-click="togglePasswordVisibility"
+		@input="handleInput">
+		<!-- Default slot for the leading icon -->
+		<slot />
+		<template slot="trailing-button-icon">
```
raimund-schluessler, line 95:
```suggestion
		<template #trailing-button-icon>
```

src/components/PasswordField/PasswordField.vue
```diff
@@ -0,0 +1,318 @@
...
+
+	methods: {
+		handleInput(event) {
+			this.$emit('update:value', event.target.value)
+		},
+		togglePasswordVisibility() {
+			this.isPasswordHidden = !this.isPasswordHidden
+		},
+		checkPassword: _.debounce(function(password) {
+			axios.post(generateOcsUrl('apps/password_policy/api/v1/validate'), { password })
```
raimund-schluessler, line 307: Can you please use aync/await? It makes it a lot more readable.

src/components/TextField/TextField.vue
```diff
@@ -57,9 +57,9 @@ and `minlength`.
 			<Lock :size="16" />
 		</TextField>
 		<div class="external-label">
-			<label for="$refs.textField.id">External label</label>
+			<label for="textField">External label</label>
 			<TextField :value.sync="text5"
-				ref="textField"
+				id="textField"
```
raimund-schluessler, line 62: This needs to be unique by adding a random id. Otherwise we will have the same id multiple times on the page (in the likely case of multiple textfields), which is invalid.

vinicius73: This in the docs/example part.
I had the same issue 😄

raimund-schluessler: Ah, right. Thanks for the hint.

vinicius73:
LGTM 👏🏼 

**Question:** Why did you use `:value.sync` instead of `v-model` ?

```vue
<PasswordField :value.sync="text1" label="Password" />
```

```vue
<PasswordField v-model="text1" label="Password" />
```

https://v2.vuejs.org/v2/guide/forms.html#v-model-with-Components

src/components/PasswordField/PasswordField.vue
```diff
@@ -0,0 +1,322 @@
...
+		},
+		togglePasswordVisibility() {
+			this.isPasswordHidden = !this.isPasswordHidden
+		},
+		checkPassword: debounce(async function(password) {
+			const { data } = await axios.post(generateOcsUrl('apps/password_policy/api/v1/validate'), { password })
+			if (data.ocs.data.passed) {
+				this.internalHelpMessage = t('Password is secure')
+			}
+			this.internalHelpMessage = data.reason
```
vinicius73, line 318: What is the expected behavior when this request fail?

vinicius73:
For a `PasswordField` component, a way to know if the password is "valid" or not to make forms able to flow looks relevant.


### Two events

```vue
<PasswordField @valid="onValid" @invalid="onInvalid" />
```

### Only one event

This event can dispatch a payload like `{ valid: Boolean }` 

```vue
<PasswordField @state-change="onStateChange" />
```

vinicius73:
While the request is running, we do not have any feedback about it.
Is it possible to have a loading icon or something similar?

src/components/PasswordField/PasswordField.vue
CarlSchwan: The expected behavior is that it doesn't fail. I added a try-catch and a logger to log just in case.

CarlSchwan:
> For a `PasswordField` component, a way to know if the password is "valid" or not to make forms able to flow looks relevant.
> ### Two events
> 
> ```
> <PasswordField @valid="onValid" @invalid="onInvalid" />
> ```
> 
> ### Only one event
> 
> This event can dispatch a payload like `{ valid: Boolean }`
> 
> ```
> <PasswordField @state-change="onStateChange" />
> ```

Added

CarlSchwan:
> While the request is running, we do not have any feedback about it.
> Is it possible to have a loading icon or something similar?

Generally, we only offer this feedback when the user submits a form and we want to make sure that they know that the data is being processed.

Here with the validity check done in the background, if we add a loading indicator, I fear the user will think that we already submitted the data by error. That the check is done in a network request and not in the browser is an implementation detail.

marcoambrosini:
I think we should add the disabled prop and use that while requests are being made. This already works via forwarding, but by capturing the attr as a prop we can add some docs about how to handle this

CarlSchwan:
> I think we should add the disabled prop and use that while requests are being made. This already works via forwarding, but by capturing the attr as a prop we can add some docs about how to handle this

Not sure I understand. We don't want to disable the text field while the password is checked, this is a background operation that is done while typing.

marcoambrosini:
I mean when you hit a submit button, we want to be able to disable the whole password field until the request has been made. I'm wasn't referring to validation

CarlSchwan:
Ready to merge?

marcoambrosini (approved):
👌

PR #2946, opened 2022-08-05
Make sure modal ids are unique

### LINKED ISSUE #2930: Modal ids are not unique in case there are multiple rendered on the same page
The ids might not be unique if there are multiple modals on the page. I think this needs to be enhanced with a randomly generated string,

_Originally posted by @raimund-schluessler in https://github.com/nextcloud/nextcloud-vue/issues/2696#issuecomment-1203783466_

nickvergessen:
Fix #2930

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -109,13 +109,13 @@
 			:class="{ 'modal-mask--dark': dark }"
 			:style="cssVariables"
 			role="dialog"
-			aria-labelledby="modal-title"
-			aria-describedby="modal-description">
+			:aria-labelledby="'modal-title-' + randId"
+			:aria-describedby="'modal-description-' + randId">
 			<!-- Header -->
 			<transition name="fade-visibility">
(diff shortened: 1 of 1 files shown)


src/components/Modal/Modal.vue
```diff
@@ -109,13 +109,13 @@
 			:class="{ 'modal-mask--dark': dark }"
 			:style="cssVariables"
 			role="dialog"
-			aria-labelledby="modal-title"
-			aria-describedby="modal-description">
+			:aria-labelledby="'modal-title-' + randId"
```
raimund-schluessler, line 112: Just a personal preference:
```suggestion
			:aria-labelledby="`modal-title-${randId}`"
```
and maybe make it a computed prop?

raimund-schluessler (approved):
Works as it is. Just a proposal in https://github.com/nextcloud/nextcloud-vue/pull/2946/files#r938743849.

marcoambrosini (approved):

PR #2952, opened 2022-08-05
Webpack config: Remove source-map

marcelklehr:
Fix local styleguide server run

### DIFF
--- a/webpack.dev.js
+++ b/webpack.dev.js
@@ -8,5 +8,4 @@ module.exports = merge(common, {
 		noInfo: true,
 		overlay: true,
 	},
-	devtool: 'source-map',
 })


skjnldsv (approved):

marcoambrosini (approved):

PVince81:
so expect devs to manually add the source map back for debugging in browser ?

juliushaertl:
When I checked this it seemed to be mainly caused by outdated dependencies of vue-styleguidedist, namely source-map. Just did a quick test locally and manually installing the latest webpack-sources / source-map package seems to solve the issue:

```
npm install --save-dev source-map
```

Might be nicer to still have the source maps.

marcelklehr:
the npm overrides feature is shipped with npm v8.3+ but also doesn't fix this for me, when pinning all instances in the tree of webpack-sources and source-map to the latest versions. :/ @juliushaertl Can you open a PR with the solution that worked for you, so we can test, perhaps?

juliushaertl:
Now it also seems broken again for me :/ sorry for the false hopes

marcoambrosini:
Aight, I think it's much better to have the docs working than the source maps, at least for me. If no one has time to investigate this further, I'd merge this

raimund-schluessler (approved):

Issue #2643, opened 2022-04-19
Rename Vue components to multi word components


dartcafe:
According to https://eslint.vuejs.org/rules/multi-word-component-names.html the component names should not be named with a single word to avoid conflicts with current or future html tags.

An example is the Button component, which could cause conflicts with the html tag button. 

Affected components: 
* Actions
* Avatar
* Breadcrump
* Breadcrumps
* Button
* Content
* Highlight
* Modal
* Popover

Maybe the vue3 migration could be a good point to optimize this.

vinicius73:
I believe we can use a prefix in all components.

- `NCActions`
- `NCAvatar`
- `NCButton`
- `NCContent`
- ...

It will prevent possible [collisions](https://v2.vuejs.org/v2/style-guide/?redirect=true#Multi-word-component-names-essential) and make clear whom components are from `@nextcloud/vue`

vinicius73:
I guess we can try something about this in v6 or #2959

korelstar:
This is getting critical due to the changes in `@nextcloud/eslint-config` v8.1.2. When updating an app's dependency, you cannot use single-worded components anymore, e.g. `Content` or `Button`. Hence, this issue has to be fixed in order to be able to use `@nextcloud/eslint-config` v8.1.2 in apps.

PR #2953, opened 2022-08-05
Add extra example for modal

PVince81:
To be able to see the previous/next buttons in the style guide

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -26,7 +26,51 @@
 	<template>
 		<div>
 			<button @click="showModal">Show Modal</button>
-			<modal v-if="modal" @close="closeModal" size="small">
+			<modal v-if="modal" @close="closeModal" size="small" title="Title">
+				<div class="modal__content">Hello world</div>
+			</modal>
+		</div>
+	</template>
+	<style scoped>
+	.modal__content {
+		margin: 50px;
+		text-align: center;
+	}
+
+	</style>
+	<script>
+	export default {
+		data() {
+			return {
+				modal: false
+			}
+		},
+		methods: {
+			showModal() {
+				this.modal = true
+			},
+			closeModal() {
+				this.modal = false
+			}
+		}
+	}
+	</script>
+	```
+
+	### Modal with more properties
+
+	```vue
+	<template>
+		<div>
+			<button @click="showModal">Show Modal with more properties</button>
+			<modal
+				v-if="modal"
+				@close="closeModal"
+				size="large"
+				:outTransition="true"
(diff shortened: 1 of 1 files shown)


PVince81:
ref, this was for testing and related to ticket https://github.com/nextcloud/nextcloud-vue/issues/2954

PVince81:
I've extended the example to include some fields, which makes it possible to test focus in the dialog and also notice that the focus returns to the original element (and closes https://github.com/nextcloud/nextcloud-vue/issues/2816)

Pytal (approved):

src/components/Modal/Modal.vue
```diff
@@ -57,6 +64,69 @@
...
+	}
+
+	</style>
+	<script>
+
+	import Button from '../Button/index.js'
+	import TextField from '../TextField/index.js'
+	export default {
+		components: {
+			TextField,
```
raimund-schluessler, line 107:
```suggestion
			TextField,
			Button,
```

src/components/Modal/Modal.vue
```diff
@@ -57,6 +64,69 @@
 	</script>
 	```
 
+	### Modal with more properties
+
+	```vue
+	<template>
+		<div>
+			<button @click="showModal">Show Modal with fields</button>
```
raimund-schluessler, line 72: Make this a vue button as well?

src/components/Modal/Modal.vue
```diff
@@ -57,6 +64,69 @@
...
+	}
+
+	.input-field {
+		margin: 12px 0px;
+	}
+
+	</style>
+	<script>
+
+	import Button from '../Button/index.js'
```
raimund-schluessler, line 103:
```suggestion
	import ButtonVue from '../Button/index.js'
```
and adjust the use accordingly.

marcoambrosini (approved):

src/components/Modal/Modal.vue
marcoambrosini: Should we actually rename the `Button` component into something else for 6.0.0? Importing like this is a bit painful.

raimund-schluessler: > Should we actually rename the `Button` component into something else for 6.0.0? Importing like this is a bit painful.

Fine with me.

raimund-schluessler: But this should be done in the context of https://github.com/nextcloud/nextcloud-vue/issues/2643 and shouldn't block this PR here, I think.

raimund-schluessler:
Follow-up to import `Button` correctly is in #2984.

PR #2959, opened 2022-08-05
[WIP] Migrate build to Vite

vinicius73:
A proof of concept to use Vite instead Webpack.

> Vite is a build tool that aims to provide a faster and leaner development experience for modern web projects.

https://vitejs.dev/guide/why.html

- [x] Add Vite build process
- [ ] Remove webpack from build
  - [ ] Migrate from styleguidist to another tool

This change works as expected without any specific errors (regards problems from `v6-alfa` version whom this PR is based)

https://github.com/nextcloud/text/pull/2779

---

# ESM Build

[ECMAScript Modules](https://nodejs.org/docs/latest-v16.x/api/esm.html) and Tree Shaking[^1] are largely used nowadays.

We do not need to target our packages as CommonJS modules only.

Webpack, Vite, Rollup, Parcel and Skypack are able to detect what kind of module we need to use.

- https://webpack.js.org/guides/package-exports/
- https://rollupjs.org/guide/en/#publishing-es-modules

Also, we have a good compatibility layer.

- https://www.npmjs.com/package/@rollup/plugin-commonjs
- https://www.npmjs.com/package/@originjs/vite-plugin-commonjs
- https://webpack.js.org/concepts/modules/

### ESM Native support

The main browsers are already compatible with ECMAScript Modules[^2]. Also, [Node.js](https://nodejs.org/docs/latest-v16.x/api/esm.html) is [natively compatible](https://caniuse.com/es6-module).

- Chrome >=87
- Firefox >=78
- Safari >=13
- Edge >=88

Bundlers like Vite target ESM as default for browsers and node environments.

This change improves the performance and bundle's size of our packages, since a lot of transpiration can be removed from the final bundle.

# Break Changes

No changes in any component have made. But, the import of components and functions had changes.

```diff
- import Actions from '@nextcloud/vue/dist/Components/Actions'
- import ActionButton from '@nextcloud/vue/dist/Components/ActionButton'
+ import { Actions, ActionButton } from '@nextcloud/vue' 
```

All components and functions are exposed directly in the main file. We do not need any more import each component or function file.

It not mean what an unused component will be bundled in the final application.
The Tree Shaking[^1] process take care of this and remove any dead code.

### Emoji functions

As all functions are directly expose, the function `addRecent` from `src/functions/emoji` have chanced to make clear what kind of function it is.

Now `addRecent` is called `emojiAddRecent`

### CSS

Vite generate CSS file separated from JavaScript files. We need to import it in or applications manually.

```js
import '@nextcloud/vue/style.css'
```

[^1]: [webpack](https://webpack.js.org/guides/tree-shaking/), [Roolup](https://rollupjs.org/guide/en/#tree-shaking) and [parcel](https://parceljs.org/features/production/#tree-shaking)
[^2]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules


vite.config.js
```diff
@@ -0,0 +1,70 @@
+import { fileURLToPath, URL } from 'node:url'
```
CarlSchwan, line 1: copyright missing

src/l10n.js
```diff
@@ -3,36 +3,39 @@ import { getGettextBuilder } from '@nextcloud/l10n/dist/gettext.js'
 const gtBuilder = getGettextBuilder()
 	.detectLocale()
 
-// Decompress Translations to gettext format and add to gtBuilder
-TRANSLATIONS.forEach((lang) => {
-	const translations = {}
+// eslint-disable-next-line no-lone-blocks
+{
```
max-nextcloud, line 7: I'm curious why you added this lone block. Does it have any effect? I don't see any scoped variables in there. They are all in the `TRANSLATIONS.forEach` loop anyway.

Is it related to the last example with the source type module here: https://eslint.org/docs/latest/rules/no-lone-blocks ?

max-nextcloud (commented):
Just a brief code review for now. Have not tried building / using this myself.

I also wonder about the timeline... will we ship this with a new major release that we then start using for NC 26?

vinicius73:
> I also wonder about the timeline... will we ship this with a new major release that we then start using for NC 26?

I don't know, the main block is styleguidist.
The build its self is working perfectly, but to make send migrate to Vite, keep webpack just for styleguidist do not look reasonable to me.

I will try https://histoire.dev/, but it is a large refactoring.

src/l10n.js
vinicius73: Just to avoid some inconsistencies when replace this entry with an array.

My idea is remove part of this code and avoid some unnecessary computations.

vinicius73:
image

I've been playing a bit with https://histoire.dev/
Works well, but some features are missing.

- It is not working with Vite v3 yet https://github.com/histoire-dev/histoire/issues/189 
- Auto doc is not available https://github.com/histoire-dev/histoire/issues/126

skjnldsv:
> Works well, but some features are missing.

Seems like very important features, so let's not move away from it yet I'd say

skjnldsv:
> Vite generate CSS file separated from JavaScript files. We need to import it in or applications manually.

Does that means we now have one big css instead of tiny ones?

susnux:
@vinicius73 about styleguideist, long story short: I overlooked your work here and also created a vite branch including API documentation using vue-doc-gen and vitepress.

Besides switching from webpack to vite I also switched from jest to vitest and from styleguidist to doc-gen + vitepress.
The only feature missing is the live preview of the components, because it is really a pain to have vue2 and vue3 installed in parallel (vitepress using vue3 and `@nextcloud/vue` components working with vue2).
Locally I could get it work, but only using experimental node functions (import maps).

Maybe you can get some inspiration from my branch: https://github.com/nextcloud/nextcloud-vue/compare/enh/vite?expand=1

skjnldsv:
@susnux can you open a PR from it?
So we can have a look?

susnux:
> can you open a PR from it? So we can have a look?

@skjnldsv of cause, https://github.com/nextcloud/nextcloud-vue/pull/3529

raimund-schluessler:
Closing in favor of https://github.com/nextcloud/nextcloud-vue/pull/3565

Issue #2906, opened 2022-07-31
Button too big


jotoeri:
As the Button-Component uses the box-shadow to create the border it is effectively 48px large. The `border-box` seems not to respect the shadow. Comparing the AppNavigationNew to the old version, the new Button feels like a monster.

As i didn't fully understand, why the shadow is used, this as an issue (instead of PR):

How can we reduce the size of the button component, so it really is 44px in height? Changing from shadow to border? Or reducing the button height to 40px, so it is effectively 44px with shadow? @marcoambrosini @jancborchardt ?

jancborchardt:
Sounds sensible – the element itself (incl border) should be 44px high, not 48. I also assume this is not intentional since you were always worried about the size of buttons @marcoambrosini? :)

raimund-schluessler:
The button really looks bigger than it actually is, due to
https://github.com/nextcloud/nextcloud-vue/blob/6e788124bafa8dd0dbb08cb1fc9ec29baba9109b/src/components/Button/Button.vue#L603
for example. In this image, the button is directly adjacent to the hovered dashboard entry, and does not actually overlap, but the `box-shadow` makes it look like it does:
![Screenshot 2022-08-01 at 21-48-24 Dashboard - Nextcloud](https://user-images.githubusercontent.com/2496460/182233480-92168b63-5c41-452b-87fb-9a07ba5ebf8c.png)

jotoeri:
Yes, so it seems to be mainly a problem of the secondary button. That one uses shadow-border not only for focus feedback, but also for its regular outline, which then results in the 48px.
I think the main border of the button should be moved to the border-property, so it is within the 44px height. Then just the question remains, how to handle that focus-border? Just changing the new inner borders color as currently done? Or keeping an additional shadow, effectively producing two borders? :thinking:

marcoambrosini:
Ok, I misunderstood. I thought you were referring to the focus feedback

PR #2962, opened 2022-08-06
Fix secondary Button height

jotoeri:
Fixing #2906

This shifts the secondary-button border from shadow to the border-property, thus making it visually of height 44px. The hover now changes this border to primary color, while the focus-border still is the black-shadow, as for all other buttons.

Parallel hover and focus looks slightly strange now, as the border and shadow are both shown, one in primary color, the other one black. This is however a quite specifc case, that both happens together?
@jancborchardt @marcoambrosini Could you please check, if this is visually ok? Or should we probably change the hover->focussed border to black, instead of the primary-color?

Left button with hover and focus in parallel:
button


raimund-schluessler:
I would be in favor of not using `box-shadow` for the buttons at all anymore. It makes the buttons look larger than they are and might overlap other content around them. Without `box-shadow` for the focus, the hover & focused issue would also be solved.

src/components/Button/Button.vue
```diff
@@ -509,13 +509,10 @@ export default {
 	}
 
 	&__icon {
-		height: $clickable-area;
```
raimund-schluessler, line 512: Why do you need to adjust the icon?

jotoeri: As the icon had already the 44px height, the button couldn't shrink to `40px height + 2*2px border`, but stayed at `44px height + 2*2px border`. Thus i removed the fixed icon-area and moved the icon-spacing to the margin.

marcoambrosini:
> @jancborchardt @marcoambrosini Could you please check, if this is visually ok? Or should we probably change the hover-focussed border to black, instead of the primary-color?

So this is actually not the style we want for the secondary button. The original one looks like this, but we couldn't do it originally because the theming engine we had.
Screenshot 2022-08-08 at 09 13 14

Would it be possible to go back to this theming-wise now @nickvergessen @skjnldsv?
If it's now, I would honestly get rid of the border altogether and play with the background for the hover feedback.

nickvergessen:
Hover focus should be black for now and especially since this PR is only supposed to fix the height let's not high-jack it.

marcoambrosini:
The hight doesn't need any fix without this border, which was never supposed to be there anyway. 

So as mentioned above, if the original secondary style is not possible at this stage, I would just remove the border and rely on background darkening. It's good enough for pointing devices because the cursor also changes, and for keyboard navigation we have the big outline around it. 

https://user-images.githubusercontent.com/26852655/183367605-1a993589-1e57-49dd-a8ba-313015c3c8a7.mov

raimund-schluessler:
Would also be nice to have for 6.0.0, especially since `Actions` use `Button` with #2911.

This can lead to glitches when the box-shadow behavior of `Button` is not respected, see the `Breadcrumbs` example with the visually overlapping buttons https://deploy-preview-2911--nextcloud-vue-components.netlify.app/#/Components/Breadcrumbs:
Screenshot 2022-08-08 at 12-38-55 Nextcloud Vue Style Guide

(Of course, they don't really overlap, and the issue here could be fixed with a horizontal margin around the buttons).

jotoeri:
Ok, so tbh. without border, the secondary button looks too light imo. See AppNavigationNew below. And if i imagine the Button to have that light blue, it feels to me like too many colors or shades of blue in the whole screen. So i think the current secondary button is not too bad. But that's different topic now.

So - second ping @jancborchardt for a design-decision, otherwise we go on here as @nickvergessen stated with the height reduction and secondary design will be clarified later by designers?

buttonborder

Changed this PR now to how @nickvergessen suggested:
Blue hover, Black focus border, Black focus->hover. See below, left one hover and focus in parallel, right one only hover. Only focus has the thin black shadow as before.

Screenshot from 2022-08-09 19-19-20

jotoeri:
So - Nothing new here.
Let's for now go on with this solution for the button height?

marcoambrosini:
>  the secondary button looks too light imo


We could make it a bit darker or use primary


> too many colors or shades of blue in the whole screen


It was using one of the existing color variables, not adding a new shade. The mockup is a bit misleading as it's not the right color. 



I still maintain that this border was never intended and we should just loose it and leave the height of the secondary button as is

jancborchardt:
Agree with @marcoambrosini. :) If we change the border stuff, we need to remove that stuff later anyway. Better to change it to what it was intended to be like now, and that is with the `color-primary-light` background, no border (except the focus ring of course).

jotoeri:
:+1: @marcoambrosini Can you take over then?

src/components/NcButton/NcButton.vue
```diff
@@ -509,13 +509,10 @@ export default {
...
 
 	&__icon {
-		height: $clickable-area;
-		width: $clickable-area;
-		min-height: $clickable-area;
-		min-width: $clickable-area;
 		display: flex;
 		justify-content: center;
 		align-items: center;
+		margin: auto 8px auto 12px;
```
skjnldsv, line 515: Where are those random values coming from?

jotoeri:
So - just gave it a try.
@jancborchardt @marcoambrosini Is this, what it was intended? See deploy-preview:
https://deploy-preview-2962--nextcloud-vue-components.netlify.app/#/Components/NcButton

---
-- Tbh. i personally am not happy, how the AppNavigationNew now looks like with this. I think another grey button would be good. (SecondAndAHalfAry?) :thinking: 

Screenshot from 2022-08-19 10-11-56

src/components/NcButton/NcButton.vue
raimund-schluessler, line 515: This has to be reverted. It breaks the icon alignment.

raimund-schluessler: > This has to be reverted. It breaks the icon alignment.

Most obvious for the `Actions` component:
Screenshot 2022-08-19 at 10-20-04 Nextcloud Vue Style Guide

jotoeri: Random to how it looks good :see_no_evil:. Left one is a bit higher so it looks properly with the rounded pill.

marcoambrosini: These brakes the icon-only buttons. It also seems that they accomplish the same as sizing the icon container. What's the reasoning here?

jotoeri: :disappointed: :+1:

raimund-schluessler: Won't work like this and is very subjective. I would leave it as it is, since it's not related to the intention of this PR of fixing the button height.

marcoambrosini:
> SecondAndAHalfAry

lol :) 

Yes that's how it's supposed to look. Actually one of the reasons why the buttons were redesigned was that we wanted to move away from gray buttons all-together. This needs to be tested with theming though. I'm not sure those color variables are going to work with our dark theme and other primary colors though

src/components/NcButton/NcButton.vue
jotoeri: Reason was, that the height was fixed twice, so it was necessary with the border thing and imo this would have been a bit cleaner. Anyways. Reverted it as of Raimunds comment. :wink:

marcoambrosini: Wow we spammed this poor comment xD

skjnldsv: Yes, wth guys :grin: 

In any case, no random values, base the sizing on `clickable-area` pls :)

skjnldsv (approved):

jotoeri:
> Actually one of the reasons why the buttons were redesigned was that we wanted to move away from gray buttons all-together

Hmm, so i think grey buttons are not that bad in terms of a sober and simple design. Having a look on the screenshot above, it's imo just too many colored elements now. But :man_shrugging:, not my business. :wink:

raimund-schluessler (approved):
Cypress base screenshots need an update.

jotoeri:
Yep, i'm currently searching for where to download the new screenshots. Can you tell me, where? :see_no_evil:

skjnldsv:
> Yep, i'm currently searching for where to download the new screenshots. Can you tell me, where? see_no_evil

https://github.com/nextcloud/nextcloud-vue/actions/runs/2888119942
Bottom

raimund-schluessler:
> Yep, i'm currently searching for where to download the new screenshots. Can you tell me, where? 🙈

From here https://github.com/nextcloud/nextcloud-vue/actions/runs/2888119942#artifacts

You need to take the actual screenshots and replace the base screenshots in cypress/snapshots/base

jotoeri:
Any easy way to rename these from `-actual` to `-base` ...? :thinking:

raimund-schluessler:
> Any easy way to rename these from `-actual` to `-base` ...? 🤔

Batch rename. But if you want, I can quickly do it.

jotoeri:
Ahh, wait, let me try. :wink:

raimund-schluessler:
> Ahh, wait, let me try. 😉

Haha, now we both pushed, sorry. But it should be fine now with your commits, besides the fixup stuff.

jotoeri:
👍 Just want to make sure it works, before squashing.

raimund-schluessler (approved):

PR #2970, opened 2022-08-08
Fix InputField focus state

marcoambrosini:
overrides a box shadow rule that effectively makes the border of
the input field 4px thick instead of 2.


nickvergessen:
Should be fixed by https://github.com/nextcloud/server/pull/33497 ?

skjnldsv (approved):

raimund-schluessler:
> Should be fixed by [nextcloud/server#33497](https://github.com/nextcloud/server/pull/33497) ?

This still happens for server@latest-master. I would say we merge this here.

raimund-schluessler (approved):

PR #2973, opened 2022-08-08
Allow to filter the available tags

nickvergessen:
Required to migrate behaviour like this
https://github.com/nextcloud/files_retention/blob/dcb68844ab563c261d3f8f8be5c90d2e3c9c8407/js/admin.js#L140-L143
to the Vue component.

### DIFF
--- a/src/components/MultiselectTags/MultiselectTags.vue
+++ b/src/components/MultiselectTags/MultiselectTags.vue
@@ -65,13 +65,13 @@ export default {
 
 <template>
 	<Multiselect :value="inputValue"
-		:options="tags"
+		:options="availableOptions"
 		:options-limit="5"
 		:placeholder="label"
 		track-by="id"
 		:custom-label="tagLabel"
 		:multiple="multiple"
-		:close-on-select="multiple"
+		:close-on-select="!multiple"
 		:tag-width="60"
 		:disabled="disabled"
 		@input="update">
@@ -113,6 +113,10 @@ export default {
 			type: Boolean,
 			default: true,
 		},
+		filter: {
+			type: Function,
+			default: () => true,
+		},
 	},
 	data() {
 		return {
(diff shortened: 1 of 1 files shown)


raimund-schluessler:
Hm, interesting. We have a related PR here https://github.com/nextcloud/nextcloud-vue/pull/2885 which aims at removing the limit of shown tags of currently 5.

How about we solve #2884 here as well, by removing `:options-limit="5"` and setting the default filter to slice the array to 5 elements to make it non-breaking? @nickvergessen @R0Wi

src/components/MultiselectTags/MultiselectTags.vue
```diff
@@ -65,13 +65,13 @@ export default {
 
 <template>
 	<Multiselect :value="inputValue"
-		:options="tags"
+		:options="availableOptions"
 		:options-limit="5"
```
raimund-schluessler, line 69:
```suggestion
```

src/components/MultiselectTags/MultiselectTags.vue
```diff
@@ -113,6 +113,10 @@ export default {
 			type: Boolean,
 			default: true,
 		},
+		filter: {
+			type: Function,
+			default: () => true,
```
raimund-schluessler, line 118:
```suggestion
			default: (element, index) => index < 5,
```

nickvergessen:
Would be fine by me

R0Wi:
Nice solution for me, too, thanks! I think the only things missing are the docs and maybe some tests? If you're interested i could provide them in a separate PR

raimund-schluessler (approved):

raimund-schluessler:
> Nice solution for me, too, thanks! I think the only things missing are the docs and maybe some tests? If you're interested i could provide them in a separate PR

That would be very welcome, I think.

nickvergessen:
@R0Wi very much welcome as I'm very bad in JS/docs and more a consumer only

nickvergessen:
Since this is non breaking I would then also backport it.

So you can also add the docs and tests here if you got the time

skjnldsv (approved):

PR #2981, opened 2022-08-09
Add aria-label to single action button

raimund-schluessler:
This adds a default `aria-label` with the value of the tooltip to the single action button, fixing the `You need to fill either the text or the ariaLabel props in the button component.` warning that currently occurs in the docs for single actions if the `ActionButton` has no `aria-label` set. Follow-up of #2911.


marcoambrosini (approved):

skjnldsv (approved):

skjnldsv:
backport?

PR #2984, opened 2022-08-09
Fix button import in modal example

raimund-schluessler:
This fixes the broken styleguide after #2953. `Button` was renamed to `ButtonVue` in #2980 because eslint complained. So importing it as `Button` does not work anymore and breaks the styleguide.

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -79,10 +79,12 @@
 					<h2>Please enter your name</h2>
 					<TextField label="First Name" :value.sync="firstName" />
 					<TextField label="Last Name" :value.sync="lastName" />
-					<Button
+					<ButtonVue
 						:disabled="!this.firstName || !this.lastName"
 						@click="closeModal"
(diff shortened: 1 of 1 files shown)


src/components/Modal/Modal.vue
```diff
@@ -79,10 +79,12 @@
 					<h2>Please enter your name</h2>
 					<TextField label="First Name" :value.sync="firstName" />
 					<TextField label="Last Name" :value.sync="lastName" />
-					<Button
+					<ButtonVue
 						:disabled="!this.firstName || !this.lastName"
 						@click="closeModal"
-						type="primary">Submit</Button>
+						type="primary">
+						Submit
```
GretaD, line 86: does this need translation?

raimund-schluessler: No, it's just in the docs, not in the component.

ChristophWurst (approved):

vinicius73 (approved):

PR #2985, opened 2022-08-09
 createElement to h for future vue3

raimund-schluessler:
This PR renames the `createElement` function used in the vue render function to `h` to simplify a future vue3 migration (in vue3, this function is imported from `vue` as `h`, see https://v3-migration.vuejs.org/breaking-changes/render-function-api.html#_2-x-syntax)


raimund-schluessler:
Netlify build fails because of https://github.com/nextcloud/nextcloud-vue/pull/2984

skjnldsv (approved):

vinicius73 (approved):

PR #2997, opened 2022-08-10
Fix regressions from floating-vue migration

raimund-schluessler:
This PR fixes two regressions from the migration to floating-vue in #2600. The `RichContenteditable` tooltip open variable name was not adjusted and the tooltip top arrow style was broken.

Before (tooltip only shown on hover):
Bildschirmfoto vom 2022-08-10 17-08-27

After (tooltip automatically shown):
Bildschirmfoto vom 2022-08-10 17-10-32


Pytal (approved):

PVince81 (approved):
:+1:

PR #2998, opened 2022-08-10
Import tooltip directive only locally

raimund-schluessler:
This changes the `RichContenteditable` component to locally import the `tooltip` directive. Hence, a global import by the app developer is not necessary anymore.


vinicius73 (approved):

PVince81 (approved):
:+1:

PR #2999, opened 2022-08-10
Also export tooltip options

### LINKED ISSUE #2993: Tooltip default options are not changable anymore in v6.0
Regression from #2600

In Talk we use:
https://github.com/nextcloud/spreed/blob/47e98d86d617925253f69f4a5a7ad67d63fd4d79/src/main.js#L84
```js
Tooltip.options.defaultContainer = '#content-vue'
```

which is needed so that e.g. avatar menus and other things are still shown while talk is in fullscreen mode.

Similar options are also available in the new library:
https://floating-vue.starpad.dev/migration/migration-from-v2.html#global-configuration

The problem is that the directive does not export? the options as pointed out in
https://github.com/nextcloud/nextcloud-vue/pull/2600/files#r933007379

raimund-schluessler:
This adds an export of the tooltip `options` object, so one can customize the tooltips if necessary.

Works like this:
```js
import { options as TooltipOptions } from '@nextcloud/vue/dist/Directives/Tooltip'
TooltipOptions.container = '#content-vue'
```

Closes #2993.

### DIFF
--- a/src/directives/Tooltip/index.js
+++ b/src/directives/Tooltip/index.js
@@ -29,3 +29,4 @@ options.themes.tooltip.distance = 10
 options.themes.tooltip['arrow-padding'] = 3
 
 export default VTooltip
+export { options }


nickvergessen (approved):

nickvergessen:
Thank you very much

PVince81 (approved):
:+1:

PR #3008, opened 2022-08-11
V6.0.0 beta.2

PVince81:
contains https://github.com/nextcloud/nextcloud-vue/pull/3006 which moves https://github.com/nextcloud/nextcloud-vue/pull/2911 from "breaking" to "enhancement".

- [x] defaultIcon for Actions regression: https://github.com/nextcloud/nextcloud-vue/pull/3006#issuecomment-1212462507
- [x] #3017 
 
note: the regressions were just a misunderstanding about whether we wanted to keep or break icon attr: https://github.com/nextcloud/nextcloud-vue/pull/3006#issuecomment-1212422191

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,15 +2,22 @@
 
 All notable changes to this project will be documented in this file.
 
+## v6.0.0-beta.2 (2022-08-11)
+
+#### :bug: Bug Fix
+
+* [#3006](https://github.com/nextcloud/nextcloud-vue/pull/3006) Fix single button action icon attribute ([@PVince81](https://github.com/PVince81))
+
 ## v6.0.0-beta.1 (2022-08-11)
 
 #### :boom: Breaking Change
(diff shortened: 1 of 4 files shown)


raimund-schluessler:
See my comment in https://github.com/nextcloud/nextcloud-vue/pull/3006#issuecomment-1212440101. 
Multiple actions with custom icon class is not working.

PVince81:
alright, thanks for the comment.
I'll need to dig deeper there.

raimund-schluessler:
> alright, thanks for the comment. I'll need to dig deeper there.

It should work pretty much the same as for the standalone trigger button, with the difference that the default-icon class should only be set if the icon slot of Actions is empty, and that in that case the icon slot of the button must not be populated.

PVince81:
see https://github.com/nextcloud/nextcloud-vue/pull/3006#issuecomment-1212451992 as it seems to be working already

PVince81:
as discussed, need to add back default-icon support as well for Actions

PVince81:
updated changelog, please review!

raimund-schluessler (approved):

skjnldsv (approved):

PR #3017, opened 2022-08-12
Fix icon-class, re-add default-icon support for Actions

### LINKED ISSUE #3016: `icon-class` support in `Actions` component broken
After the change in #3006, the support for icon-classes is now broken for `Actions`.
A markup like 
```
<template>
	<Actions menuTitle="test" forceTitle>
		<ActionButton icon="icon-delete">
			Delete
		</ActionButton>
	</Actions>
</template>
```
will lead to this rendering:
Screenshot 2022-08-12 at 09-56-41 Nextcloud Vue Style Guide

This is because the `ButtonVue` component used for showing the `Actions` component trigger has no support for icon-classes and simply adding the icon-class as in #3006 will lead to trouble.

So in order to fix that, one needs to create a span element, apply the icon-class to it, and populate the icon slot of button with it. Not really nice, but apparently needed for server, as a migration to material design icons only is to much effort for now.

@PVince81 @nickvergessen for reference.

raimund-schluessler:
This brings back the support for a `default-icon` and fixes the support of `icon` for the `Actions` component.
Follow-up to #3006, fixes #3016.

Please test this well, I only checked in in the docs.

### DIFF
--- a/src/components/Actions/Actions.vue
+++ b/src/components/Actions/Actions.vue
@@ -449,6 +449,16 @@ export default {
 		},
 
 		/**
+		 * Icon to show for the toggle menu button
+		 * when more than one action is inside the actions component.
+		 * Only replace the default three-dot icon if really necessary.
+		 */
+		defaultIcon: {
+			type: String,
+			default: '',
+		},
+
+		/**
 		 * Aria label for the actions menu
 		 */
 		ariaLabel: {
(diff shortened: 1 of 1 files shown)


CarlSchwan (approved):
Probably make sense to at some point have a slot for material icon since we are migrating to that slowly in other places.

But definitively not blocking or urgent, let's first get back `default-icon` so we can port our apps to Nextcloud Vue 6 :)

Thanks!

raimund-schluessler:
> Probably make sense to at some point have a slot for material icon since we are migrating to that slowly in other places.
> 
> But definitively not blocking or urgent, let's first get back `default-icon` so we can port our apps to Nextcloud Vue 6 :)
> 
> Thanks!

The component has a slot already, so you can start migrating immediately 😉
That was actually the reason why I didn't bother to re-implement the icon class support, because one can do everything with material design icons just as well.

PVince81 (approved):
:+1:

PVince81:
I did some migrations already back then on Talk and remember that it wasn't always straightforward because you had to find a matching icon and sometimes styles started to go crazy. This is why I wouldn't want to rush removing the icon attribute support.

nickvergessen:
> I did some migrations already back then on Talk and remember that it wasn't always straightforward because you had to find a matching MD icon and sometimes styles started to go crazy. This is why I wouldn't want to rush removing the icon attribute support.

Last week I replaced a lot of icons and buttons in Talk, it was mostly all pretty straight forward

PR #3018, opened 2022-08-12
Add jquery as depenendency for tributejs

juliushaertl:
Adds jquery to the vue library as a dependency, to avoid issues during npm link with other repos, which shouldn't matter in terms of building as those are marked as external anyways.

Required for https://github.com/nextcloud/server/pull/33517

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,6 +25,7 @@
 				"floating-vue": "^1.0.0-beta.18",
 				"focus-trap": "^6.9.4",
 				"hammerjs": "^2.0.8",
+				"jquery": "^3.6.0",
 				"linkify-string": "^3.0.4",
 				"md5": "^2.3.0",
 				"splitpanes": "^2.4.1",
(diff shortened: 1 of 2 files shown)


ChristophWurst (approved):

PVince81 (approved):
:+1:

Pytal (approved):

raimund-schluessler:
@juliushaertl could you please elaborate why this is necessary / helps? Tributejs has literally no dependencies.

juliushaertl:
@raimund-schluessler You're right, actually the issue is not triggered by tribute.js or nextcloud-vue but the web pack provide plugin that the server uses.

https://github.com/nextcloud/server/blob/master/webpack.common.js#L144

This causes the following tribute.js code to fail then as it tires to resolve it from the linked path:

```
~/repos/nextcloud/@nextcloud/nextcloud-vue bugfix/noid/popover-focus-trap✦ ➜ rg jQuery node_modules/tributejs/dist/tribute.esm.js
1497:    // Check if it is a jQuery collection
1498:    if (typeof jQuery !== "undefined" && el instanceof jQuery) {
1788:    // Check if it is a jQuery collection
1789:    if (typeof jQuery !== "undefined" && el instanceof jQuery) {
```

I'll see if there is a way to fix this in the server webpack config.

juliushaertl:
Found a way https://github.com/nextcloud/server/pull/33517/commits/d78b82b5c673a110d062b8101873285d2209ca62

PR #3020, opened 2022-08-12
Fix issues with focus trap on modal

juliushaertl:
This PR addresses two issues that are not directly visible in the components but with the way how our apps use them.

- Allow click outside even if the focus trap is set
- Allow apps (like viewer) to pass additional focus trap elements (like the sidebar) which will then be added to the trap so it can be accessed

Corresponding viewer commit: https://github.com/nextcloud/viewer/pull/1319/commits/a2fb11097d8a3a81c4f666edfc038b7b5c6aafef


PVince81 (approved):
:+1: nice catch

skjnldsv (approved):

PR #3021, opened 2022-08-12
Disable focus-trap for Popover by default

raimund-schluessler:
This PR disabled the focus trap for Popovers by default for the moment. Reason is, that focusing other elements from within an Actions menu does not work anymore. E.g. here: https://github.com/nextcloud/tasks/blob/master/src/components/TaskBody.vue#L583-L588 the call to `focus()` has no effect.

Also see this example (copy into docs):
```vue
<template>
    <div>
        <input ref="input" />
        <Actions>
            <ActionButton @click="focusInput" :close-after-click="true">
                <template #icon>
                    <Delete :size="20" />
                </template>
                Focus input
            </ActionButton>
            <ActionButton @click="actionDelete">
                <template #icon>
                    <Delete :size="20" />
                </template>
                Delete
            </ActionButton>
        </Actions>
    </div>
</template>
<script>
import Delete from 'vue-material-design-icons/Delete'

export default {
	components: {
		Delete,
	},
	methods: {
		actionDelete() {
			alert('Delete')
		},
		async focusInput() {
			await this.$nextTick()
            this.$refs.input.focus()
		},
	},
}
</script>
```
The input should be focused when clicking on "Focus input" in the menu.

### DIFF
--- a/src/components/Popover/Popover.vue
+++ b/src/components/Popover/Popover.vue
@@ -114,7 +114,7 @@ export default {
 		},
 		focusTrap: {
 			type: Boolean,
-			default: true,
+			default: false,
 		},
 	},
 


raimund-schluessler:
I guess the problem here is that focus-trap by default returns the focus to whatever element had focus upon activation of the focus trap. Would it be ok to set `returnFocusOnDeactivate` to `false` for the focus-trap, or is this a feature one really wants?
Question would be how to then hand over the focus to another element after closing the menu. @vinicius73 @PVince81 @juliushaertl Any ideas / opinions?

vinicius73:
> I guess the problem here is that focus-trap by default returns the focus to whatever element had focus upon activation of the focus trap. Would it be ok to set `returnFocusOnDeactivate` to `false` for the focus-trap, or is this a feature one really wants? Question would be how to then hand over the focus to another element after closing the menu. @vinicius73 @PVince81 @juliushaertl Any ideas / opinions?

This what the behavior who we want, but it should be optional.
We can provide a prop for this.

vinicius73:
We will merge it for now, and keep trying a solution for it.
It is important for accessibility issues.

vinicius73 (approved):

PVince81 (approved):
:+1:

PR #3023, opened 2022-08-12
Dismis focus trap when component gets destroyed

juliushaertl:
Otherwise the trap is kept active while the component is being destroyed e.g. on unshare of a link

### DIFF
--- a/src/components/Popover/Popover.vue
+++ b/src/components/Popover/Popover.vue
@@ -141,6 +141,10 @@ export default {
 		)
 	},
 
+	beforeDestroy() {
+		this.afterHide()
+	},
+
 	methods: {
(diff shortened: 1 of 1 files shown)


raimund-schluessler:
Let's see if this also solves https://github.com/nextcloud/nextcloud-vue/pull/3021.

vinicius73 (approved):

raimund-schluessler:
> Let's see if this also solves #3021.

Unfortunately not, see the example I added to #3021 where the input is still not focused if the focus-trap is dismissed afterDestroy (whereas it gets focused with focus-trap disabled completely in #3021).

raimund-schluessler (approved):

PR #2719, opened 2022-06-02
Add title section to settingsdialog

marcoambrosini:
..Also gets rid of the annoying title without top margin when scrolling into view  

https://user-images.githubusercontent.com/26852655/171647817-72448a07-33bc-40ba-b0a1-9592e9ee1044.mov

Signed-off-by: Marco Ambrosini <marcoambrosini@pm.me>

PR #3025, opened 2022-08-12
Fix app settings dialog

nickvergessen:
Before | After
---|---
Peek 2022-08-12 13-31 | Peek 2022-08-12 14-36

### DIFF
--- a/src/components/AppSettingsDialog/AppSettingsDialog.vue
+++ b/src/components/AppSettingsDialog/AppSettingsDialog.vue
@@ -359,8 +359,7 @@ export default {
 <style lang="scss" scoped>
 
 :deep(.modal-wrapper .modal-container) {
-	display: flex;
-	overflow: hidden;
+	overflow: initial;
 }
 
 .app-settings {
@@ -375,7 +374,6 @@ export default {
 	&__wrapper {
 		display: flex;
 		width: 100%;
-		overflow: hidden;
 		height: 100%;
 		position: relative;
 	}
--- a/src/components/AppSettingsDialog/AppSettingsDialog.vue
+++ b/src/components/AppSettingsDialog/AppSettingsDialog.vue
@@ -211,7 +211,6 @@ export default {
 		 * @param {string} item the ID of the section
 		 */
 		handleSettingsNavigationClick(item) {
-			console.error(item)
 			this.linkClicked = true
 			document.getElementById('settings-section_' + item).scrollIntoView({
 				behavior: 'smooth',


raimund-schluessler:
Just for reference, the respective lines were added in https://github.com/nextcloud/nextcloud-vue/pull/2719.

raimund-schluessler (changes requested):
With these changes, scrolling the components content in the docs does not work anymore. Maybe one rather needs to adjust the usage in Talk?

nickvergessen:
The usage in Talk is AppSettingsDialog with tones of AppSettingsSection inside.

Not sure if should be necessary for an app to overwrite overflows of a stand-alone modal component?
https://github.com/nextcloud/spreed/blob/d01450f87a68454a7acf6fa8e51ad8f0bec40ce6/src/components/ConversationSettings/ConversationSettingsDialog.vue

raimund-schluessler:
> The usage in Talk is AppSettingsDialog with tones of AppSettingsSection inside.

Quite the same happens in the docs example, no?

> Not sure if should be necessary for an app to overwrite overflows of a stand-alone modal component? https://github.com/nextcloud/spreed/blob/d01450f87a68454a7acf6fa8e51ad8f0bec40ce6/src/components/ConversationSettings/ConversationSettingsDialog.vue

I guess not. Does it maybe work if you remove the adjustments.

nickvergessen:
The only way I got it working is with these adjustments...

raimund-schluessler:
Really weird. There must be a reason, it works in the docs, but breaks with your changes.

raimund-schluessler:
I cloned and build the Talk app and I couldn't reproduce the scrolling issue in the settings dialog with the Talk master branch. It works just fine (besides #3029).

But I have the assumption that it could depend on how the bundler packed everything together, since this rule
https://github.com/nextcloud/nextcloud-vue/blob/0324a18e33e390e27549cf2c405ed70df2d41161/src/components/AppSettingsDialog/AppSettingsDialog.vue#L362-L365
conflicts with this rule
https://github.com/nextcloud/nextcloud-vue/blob/0324a18e33e390e27549cf2c405ed70df2d41161/src/components/Modal/Modal.vue#L868-L869
and the order of them in the bundle (which might vary) determines which one applies.

So the rule overwriting the style for AppSettingsDialog needs to be more specific. I will create a PR over the weekend.

nickvergessen:
Thanks for diving in as so often recently @raimund-schluessler

raimund-schluessler:
Alternative solution is in https://github.com/nextcloud/nextcloud-vue/pull/3035.

raimund-schluessler:
Closed with https://github.com/nextcloud/nextcloud-vue/pull/3035 and https://github.com/nextcloud/spreed/pull/7745.

PR #3028, opened 2022-08-12
🩹 enhance actions focus trap

vinicius73:
Related to #3021 

- add scoped default slot 
- expose `clearFocusTrap`
- add example in `Actions`


raimund-schluessler:
Not sure why, but the docs don't build/work.

src/components/Actions/Actions.vue
```diff
@@ -373,6 +419,26 @@ import DotsHorizontal from 'vue-material-design-icons/DotsHorizontal.vue'
 
 const focusableSelector = '.focusable'
 
+const getActions = (ctx) => {
```
raimund-schluessler, line 422: What's the reason you put this function outside the component? It's only called once, and I think it should work just fine in the render function as well, no?

src/components/Actions/Actions.vue
```diff
@@ -360,6 +360,52 @@ export default {
...
 
+### Scoped slot
+
+```vue
+<template>
+    <div>
+        <input ref="input" />
+        <Actions>
+            <template v-slot:default="{ clearFocusTrap }">
+				<ActionButton @click="focusInput(clearFocusTrap)" :close-after-click="true">
```
raimund-schluessler, line 371: Somehow the indentation is off.

src/components/Actions/Actions.vue
```diff
@@ -373,6 +419,26 @@ import DotsHorizontal from 'vue-material-design-icons/DotsHorizontal.vue'
...
+	 */
+	if (ctx.$slots.default !== undefined) {
+		return ctx.$slots.default.filter(action => action?.componentOptions?.tag)
+	}
+
+	/**
+	 * Expose some internal methods to parent template
+	 */
+	if (ctx.$scopedSlots.default) {
+		return ctx.$scopedSlots.default({ clearFocusTrap: ctx.clearFocusTrap })
```
raimund-schluessler, line 436: Don't we need to filter the Actions form the scoped slot as well?

src/components/Actions/Actions.vue
```diff
@@ -373,6 +419,26 @@ import DotsHorizontal from 'vue-material-design-icons/DotsHorizontal.vue'
 
 const focusableSelector = '.focusable'
 
+const getActions = (ctx) => {
+	/**
+	 * Filter the Actions, so that we only get allowed components.
+	 * This also ensure that we don't get 'text' elements, which would
+	 * become problematic later on.
+	 */
+	if (ctx.$slots.default !== undefined) {
```
raimund-schluessler, line 428:
```suggestion
	if (ctx.$slots.default) {
```

src/components/Actions/Actions.vue
```diff
@@ -497,6 +563,13 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		/**
+		 * Enable popover focus trap
+		 */
+		focusTrap: {
```
raimund-schluessler, line 569: Do we need this prop? If it works, is there a reason to disable the focus-trap?

src/components/Actions/Actions.vue
```diff
@@ -360,6 +360,52 @@ export default {
...
 ```
 
+### Scoped slot
+
+```vue
+<template>
+    <div>
+        <input ref="input" />
+        <Actions>
+            <template v-slot:default="{ clearFocusTrap }">
```
raimund-schluessler, line 370:
```suggestion
            <template #default="{ clearFocusTrap }">
```

src/components/Actions/Actions.vue
```diff
@@ -360,6 +360,52 @@ export default {
...
+				<ActionButton @click="focusInput(clearFocusTrap)" :close-after-click="true">
+					<template #icon>
+						<Delete :size="20" />
```
raimund-schluessler, line 373: If we add this example to the docs, I would chose a better icon :slightly_smiling_face:

raimund-schluessler: Maybe `Plus`?

src/components/Actions/Actions.vue
```diff
@@ -720,14 +796,7 @@ export default {
...
 	render(h) {
-		/**
-		 * Filter the Actions, so that we only get allowed components.
-		 * This also ensure that we don't get 'text' elements, which would
-		 * become problematic later on.
-		 */
-		const actions = (this.$slots.default || []).filter(
-			action => action?.componentOptions?.tag
-		)
+		const actions = getActions(this)
```
raimund-schluessler, line 799: This should be equivalent to the above function, no? And is easier to understand, I think.
```suggestion
		const actions = (this.$slots.default || this.$scopedSlots.default({ clearFocusTrap: this.clearFocusTrap }) || []).filter(
			action => action?.componentOptions?.tag
		)
```

src/components/Actions/Actions.vue
raimund-schluessler: I would keep it in the render function, to be honest. Looks cleaner to me.

raimund-schluessler:
> Not sure why, but the docs don't build/work.

Locally, the styleguide looks good. Maybe just a hiccup?

raimund-schluessler:
@vinicius73 I hope you don't mind, I pushed a commit with my suggestions. If you disagree with them, just discard my commit. The docs still don't show up, though.

src/components/Actions/Actions.vue
vinicius73: It is not true @raimund-schluessler 
Any content inside render will be called many times.

Any small changes in state, will run render again and again.
Because of that, is a problem call functions inside template functions.

Moving some parts of code to outside render is a performance improvement.

The Vue template compile does it in some parts of the template automatically.

src/components/Actions/Actions.vue
vinicius73: Not for my point of view.
I do not know why we are doing this filter

src/components/Actions/Actions.vue
vinicius73: yes, to make sure we can handle any edge case

src/components/Actions/Actions.vue
vinicius73: I am preparing a PR with these improvements.
To be true, `Action` is already refactored in my local branch, but I've intent to sent it in other moment.

vinicius73:
> @vinicius73 I hope you don't mind, I pushed a commit with my suggestions. If you disagree with them, just discard my commit. The docs still don't show up, though.


I disagree just the with `getActions`, but as I said, I am preparing other PR.

src/components/Actions/Actions.vue
raimund-schluessler: The function is anyway called in the render function every time something changes. I don't see a difference in putting it outside.

And the slot has to be re-evaluated on every render, otherwise changed slot content won't be noticed.

And it's not true that it's called on every minor change. E.g. opening the menu causes no re-render. Just put a console log inside and observe.

raimund-schluessler:
> > @vinicius73 I hope you don't mind, I pushed a commit with my suggestions. If you disagree with them, just discard my commit. The docs still don't show up, though.
> 
> 
> 
> 
> 
> I disagree just about `getActions`, but as I said, I am preparing other PR. :)
> 
> 

Actions was just rewritten, I would like to know what you think needs to change. Let's see your PR😉

src/components/Actions/Actions.vue
```diff
@@ -360,6 +360,54 @@ export default {
...
+	},
+	methods: {
+		actionDelete() {
+			alert('Delete')
+		},
+		async focusInput(clearFocusTrap) {
+			await clearFocusTrap({ returnFocus: false })
+			await this.$nextTick()
+			this.$refs.input.focus()
+		},
```
raimund-schluessler, line 405:
```suggestion
		focusInput(clearFocusTrap) {
			clearFocusTrap({ returnFocus: false })
			this.$nextTick(() => this.$refs.input.focus())
		},
```
For some reasong the netlify docs fails on aysnc methods, so this is a workaroung.

src/components/Popover/Popover.vue
```diff
@@ -179,17 +173,36 @@ export default {
 				this.$focusTrap.activate()
 			})
 		},
+		/**
+		 * Remove focus trap
+		 *
+		 * @param options
+		 */
+		async clearFocusTrap(options = {}) {
```
raimund-schluessler, line 181:
```suggestion
		clearFocusTrap(options = {}) {
```

src/components/Actions/Actions.vue
vinicius73: Realy?! 😱 

```suggestion
		focusInput(clearFocusTrap) {
			clearFocusTrap({ returnFocus: false })
			return this.$nextTick(() => this.$refs.input.focus())
		},
```

We must keep the promise chain

raimund-schluessler: Fine with me. Although it's just an example and has no effect for the library.

raimund-schluessler:
Underlying issue was solved with #3030.

PR #3037, opened 2022-08-15
```aria-expanded``` was fixed for Multiselect.vue

### PR DESCRIPTION
**Pull request by JuliaKirschenheuter**
Signed-off-by: julia.kirschenheuter <julia.kirschenheuter@nextcloud.com>

### DIFF
--- a/src/components/Multiselect/Multiselect.vue
+++ b/src/components/Multiselect/Multiselect.vue
@@ -168,11 +168,14 @@ export default {
 		:options="options"
 		:limit="maxOptions"
 		:loading="loading"
+		:aria-expanded="ariaExpanded.toString()"
 		:close-on-select="willCloseOnSelect"
 		:multiple="multiple"
 		:label="label"
 		:track-by="trackBy"
 		tag-placeholder="create"
+		@close="ariaExpanded = false"
+		@open="ariaExpanded = true"
 		v-on="$listeners">
 		<!-- This is the scope to format the list of available options in the dropdown
 			Two templates to avoid registering the slot unnecessary -->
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/components/Multiselect/Multiselect.vue [hunk 1]
```diff
@@ -353,6 +356,7 @@ export default {
 	data() {
 		return {
 			elWidth: 0,
+			ariaExpanded: 'false',
```
raimund-schluessler, line 359: Should be a boolean. Above as well.

src/components/Multiselect/Multiselect.vue
```diff
@@ -168,11 +168,14 @@ export default {
 		:options="options"
 		:limit="maxOptions"
 		:loading="loading"
+		:aria-expanded="ariaExpanded"
```
raimund-schluessler, line 171:
```suggestion
		:aria-expanded="ariaExpanded.toString()"
```

src/components/Multiselect/Multiselect.vue
JuliaKirschenheuter: in this case no, i could create ariaExpanded as Boolean, but then have to convert it to String within a template (ariaExpanded.toString())
please have a look https://v3-migration.vuejs.org/breaking-changes/attribute-coercion.html#overview

raimund-schluessler: Exactly, see my other comment https://github.com/nextcloud/nextcloud-vue/pull/3037#pullrequestreview-1072402046
I think that's more clear. Otherwise people might wonder why `false` is a string here.
Or is there anything wrong with having `toString()` in the template. I couldn't see anything like this in the link you posted.

raimund-schluessler (approved):

skjnldsv (approved):
Also maybe offer the fix upstream to the lib? :)

PR #3040, opened 2022-08-15
Allow ListItem components to link to external links

### LINKED ISSUE #3039: Allow listitem components to have external links
It can already be a `router-link` and there's already a `<a>` tag with `href="#"` being used.

We can just allow a new `href` prop and use it for the existing link.

tcitworld:
Closes #3039

I didn't do anything UI-wise to distinguish external links (which use `target="_blank"`, but we could change that).

### DIFF
--- a/src/components/ListItem/ListItem.vue
+++ b/src/components/ListItem/ListItem.vue
@@ -125,7 +125,7 @@
 <ul style="width: 350px;">
 	<listItem
 		:title="'Title of the element'"
-		:counter-number=1
+		:counter-number="1"
 		:compact="true" >
 		<template #icon>
 			<div class="icon-edit" />
@@ -151,7 +151,7 @@
 	</listItem>
 	<listItem
 		:title="'Title of the element'"
-		:counter-number=3
+		:counter-number="3"
 		:compact="true" >
 		<template #icon>
 			<div class="icon-edit" />
@@ -170,10 +170,14 @@
 	</listItem>
 	<listItem
 		:title="'Title of the element'"
-		:counter-number=4
-		:compact="true" >
+		:compact="true"
+		:counter-number="4"
+		href="https://nextcloud.com">
 		<template #icon>
 			<div class="icon-edit" />
+		</template>
+		<template #subtitle>
+			This one is with an external link
 		</template>
 	</listItem>
 </ul>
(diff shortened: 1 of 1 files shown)


src/components/ListItem/ListItem.vue
```diff
@@ -171,10 +171,14 @@
 	<listItem
 		:title="'Title of the element'"
 		:counter-number=4
+		href="https://nextcloud.com"
 		:compact="true" >
```
skjnldsv, line 175:
```suggestion
		:compact="true" >
		:counter-number="4"
		href="https://nextcloud.com"
```

src/components/ListItem/ListItem.vue
```diff
@@ -170,11 +170,15 @@
 	</listItem>
 	<listItem
 		:title="'Title of the element'"
+		:compact="true"
 		:counter-number=4
```
skjnldsv, line 174: Still invalid without quotes

skjnldsv: Actually, all the `counter-number` on this docs if possible :see_no_evil:

skjnldsv (approved):

src/components/ListItem/ListItem.vue
tcitworld: Ha, hadn't noticed that. Done

skjnldsv: Sorry, I should have been more clear ahah, thanks!! :hugs:

artonge (approved):

src/components/ListItem/ListItem.vue
```diff
@@ -187,7 +191,9 @@
 		<a :id="anchorId"
 			ref="list-item"
 			:class="{ 'list-item--active' : active }"
-			href="#"
+			:href="href"
+			target="_blank"
+			rel="noopener noreferrer"
```
jotoeri, line 196: These two now always apply! This breaks the component.

tcitworld: Shouldn't it be harmless on a `<a href="#">` tag? How does it break it?

skjnldsv: Yes, we use @click for it, no?

jotoeri: Yes, but the target blank seems to kick in before click is emitted. Thus the appnavigation on forms now opens a new tab with the link instead of triggering the router.

tcitworld: https://github.com/nextcloud/nextcloud-vue/pull/3069 should do it. Click event would have a different behaviour if `e.preventDefault()` wouldn't be used.

PR #3050, opened 2022-08-16
Fix Content component name to be eslint compliant

nickvergessen:


vinicius73:
Why not `NCContent`? #2643

nickvergessen:
Because the Button was renamed to ButtonVue, wo I just followed the pattern

raimund-schluessler:
> Why not `NCContent`? #2643

I would also find it sensible to fix it all together with the other component names. An `NC` prefix seems to make sense, but if we apply this to *all*  components it creates a lot of friction when updating to v6 (and changing all component names here). So I am not all set with this solution. `ContentVue` would be in line with e.g. `ButtonVue`, but that anyway was a not really discussed change, so as long as it is consistent, I am good with whatever we chose.

raimund-schluessler:
Input and consensus from @nextcloud/vuejs is required here, I think.

skjnldsv:
If to choose, I would go for NcContent too :)
Note the lowercase `c` because it's `Nextcloud` and not `NextCloud` :see_no_evil:

raimund-schluessler:
Let's do a little poll :laughing: 

- :heart: = `NcContent`
- :rocket: = `ContentVue`

Like and edit my comment if necessary.

raimund-schluessler:
Superseded by #3058.

PR #3057, opened 2022-08-17
🥛 Make app navigation semi-transparent

### LINKED ISSUE #3047: 🥛 Make app navigation semi-transparent
As per the mockups

Bildschirmfoto vom 2022-08-16 12-58-35

the app navigation component should become semi-transparent and let the background image show through with a blur.

## Specification

- 80% opaque white with 10px blur, like in Dashboard. Code: `background-color: var(--color-background-translucent); -webkit-backdrop-filter: var(--background-blur); backdrop-filter: var(--background-blur);`
- [AppNavigationNew](https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/AppNavigationNew/AppNavigationNew.vue) gets our Nextcloud fade, looks nice and clicky
- WCAG AA: Contrast of secondary text in the navigation works after darkening it to `5c5c5c`
- WCAG AAA: Just like for Dashboard, the high contrast theme should make the navigation background fully opaque and the secondary text color also color-main-text.

GretaD:
- [x] requires https://github.com/nextcloud/server/pull/33610

fixes #3047 

- [x] the image is just for testing, it will be added globally on server. I will remove it after review

appnaaav

### DIFF
--- a/src/components/NcAppNavigation/NcAppNavigation.vue
+++ b/src/components/NcAppNavigation/NcAppNavigation.vue
@@ -151,17 +151,17 @@ export default {
 	z-index: 1800;
 	height: calc(100vh - #{$header-height});
 	box-sizing: border-box;
-	background-color: var(--color-main-background);
 	-webkit-user-select: none;
 	-moz-user-select: none;
 	-ms-user-select: none;
 	user-select: none;
-	border-right: 1px solid var(--color-border);
 	display: flex;
 	flex-direction: column;
 	flex-grow: 0;
 	flex-shrink: 0;
-
+	background-color:  var(--color-main-background-blur, var(--color-main-background));
+	-webkit-backdrop-filter: var(--filter-background-blur, none);
+	backdrop-filter: var(--filter-background-blur, none);
 	&--close {
 		margin-left: - $navigation-width;
 		transform: translateX(-100%);
--- a/styleguide/assets/variables.css
+++ b/styleguide/assets/variables.css
@@ -37,6 +37,8 @@
 	--color-box-shadow: rgba(77, 77, 77, 0.5);
 	--color-border: #ededed;
 	--color-border-dark: #dbdbdb;
+	--color-main-background-blur: rgba(255,255,255,0.8);
+	--background-blur: blur(25px);
 	--border-radius: 3px;
 	--border-radius-large: 10px;
 	--border-radius-pill: 100px;
--- a/src/components/NcAppNavigationSettings/NcAppNavigationSettings.vue
+++ b/src/components/NcAppNavigationSettings/NcAppNavigationSettings.vue
@@ -90,7 +90,6 @@ export default {
 
 	&__header {
 		box-sizing: border-box;
-		background-color: var(--color-main-background);
 
 		.settings-button {
 			display: flex;


src/components/Content/Content.vue
```diff
@@ -85,6 +85,8 @@ export default {
 	display: flex;
 	padding-top: 50px;
 	min-height: 100%;
+	background: url("/src/assets/kamil-porembinski-clouds.jpg");
```
ChristophWurst, line 88: to be removed once server puts the image behind every app content page

ChristophWurst (commented):
Looks good so far

styleguide/assets/variables.css
```diff
@@ -37,6 +37,8 @@
 	--color-box-shadow: rgba(77, 77, 77, 0.5);
 	--color-border: #ededed;
 	--color-border-dark: #dbdbdb;
+	--color-main-background-semitransparent: rgba(255,255,255,0.8);
+	--background-blur: blur(10px);
```
jancborchardt, line 41: Let’s go with slightly higher, as per @marcoambrosini feedback :)
```suggestion
	--background-blur: blur(25px);
```

GretaD:
the variables names will be changes because i didnt see julius pr: https://github.com/nextcloud/server/pull/33610/files

but doesnt change the result

styleguide/assets/variables.css
```diff
@@ -37,6 +37,8 @@
 	--color-box-shadow: rgba(77, 77, 77, 0.5);
 	--color-border: #ededed;
 	--color-border-dark: #dbdbdb;
+	--color-main-background-semitransparent: rgba(255,255,255,0.8);
```
jancborchardt, line 40: We already have a variable for this actually, and it’s used in Dashboard: `var(--color-background-translucent)`

jancborchardt: This will also ensure it works with dark mode.

styleguide/assets/variables.css
GretaD: yes, Julius added new variables https://github.com/nextcloud/server/pull/33610/files and i will use his :)

jancborchardt (dismissed):
See comments :)

styleguide/assets/variables.css
GretaD: this is just for documentation, we get the variables from theming now

jancborchardt:
Ah ok got it, then just use @juliushaertl’s variables, my comments are moot. :)

src/components/Content/Content.vue
juliushaertl: Let me push that to my server pr as well then

juliushaertl: Available now as --image-main-background which will also adapt to theming https://github.com/nextcloud/server/pull/33610

ChristophWurst:
> requires [nextcloud/server#33610](https://github.com/nextcloud/server/pull/33610)

What are we going to do in terms of server compatibility? Is this a hard requirement? What happens in the absence of the new variables and/or the background image?

1) Declare the next major of this lib to only work with 25+
2) Make the component work with and without the vars/image in server

marcoambrosini (commented):
this border is not needed anymore https://github.com/nextcloud/nextcloud-vue/blob/0d73e367f29fdad1461e4d2bea6e216342119ddc/src/components/AppNavigation/AppNavigation.vue#L158

src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -151,17 +151,17 @@ export default {
...
 	user-select: none;
-	border-right: 1px solid var(--color-border);
 	display: flex;
 	flex-direction: column;
 	flex-grow: 0;
 	flex-shrink: 0;
-
+	background-color:  var(--color-main-background-blur);
+	-webkit-backdrop-filter: var(--filter-background-blur);
+	backdrop-filter: var(--filter-background-blur);
```
juliushaertl, line 164:
```suggestion
	background-color:  var(--color-main-background-blur, var(--color-main-background));
	-webkit-backdrop-filter: var(--filter-background-blur, none);
	backdrop-filter: var(--filter-background-blur, none);
```

To make it a non-breaking change and still usable on older Nextcloud versions.

src/components/NcContent/NcContent.vue
```diff
@@ -85,6 +85,8 @@ export default {
 	display: flex;
 	padding-top: 50px;
 	min-height: 100%;
+	background: url("/src/assets/kamil-porembinski-clouds.jpg");
+	background-size: contain;
```
juliushaertl, line 89: Let's remove this as it will be added with #3091

ChristophWurst (approved):
:+1: code looks good

juliushaertl (approved):

raimund-schluessler:
@juliushaertl So this will work with NC 24 and below as well, right?

juliushaertl:
Yes, as we have the fallback to the previous values in place.

PR #3060, opened 2022-08-18
Allow to render x items outside an Actions menu

skjnldsv:
Capture d’écran_2022-08-18_11-10-56

Currently, it's hardcoded, meaning we are not hecking the available space and changing the inline accordingly.
Might come later, but this is a good enough starting point

### DIFF
--- a/src/components/Actions/Actions.vue
+++ b/src/components/Actions/Actions.vue
@@ -554,6 +554,7 @@ export default {
 
 		/**
 		 * Display x items inline out of the dropdown menu
+		 * Will be ignored if `forceMenu` is set
 		 */
 		inline: {
 			type: Number,
--- a/src/components/Actions/Actions.vue
+++ b/src/components/Actions/Actions.vue
@@ -422,6 +422,7 @@ import Tooltip from '../../directives/Tooltip/index.js'
 import GenRandomId from '../../utils/GenRandomId.js'
 import { t } from '../../l10n.js'
 
+import Vue from 'vue'
 import DotsHorizontal from 'vue-material-design-icons/DotsHorizontal.vue'
 
 const focusableSelector = '.focusable'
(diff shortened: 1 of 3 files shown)


src/components/Actions/Actions.vue
```diff
@@ -497,6 +550,14 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+
+		/**
+		 * Display x items inline out of the dropdown menu
+		 */
+		inline: {
+			type: Number,
+			default: 0,
```
raimund-schluessler, line 559:
```suggestion
			default: 1,
```
Otherwise it's a breaking change.
But I am not completely sold, as it conflicts a bit with the `forceMenu` prop.

raimund-schluessler: See the failing `Acitons` test.

skjnldsv: No, that seems counter intuitive. Let me fix the usage instead :)

> But I am not completely sold, as it conflicts a bit with the forceMenu prop.

Hum, Why though? You can force the menu if you want and if there is enough items to render there?

skjnldsv:
Adding more tests right now

src/components/Actions/Actions.vue
raimund-schluessler: I would say forceMenu should force all actions into the menu, also the inlined ones. But that's a matter of interpretation, I guess. And since it will keep it's behaviour unless you add inline=xx I am fine with both.

PS: I accidentally edited your above comment, instead of quoting it, and then made it worse by trying to restore it. Sorry for this 🙈

skjnldsv: That's a good question!
I can make it override any inline, sure! :)
I can even show a warning! let's do this!

> PS: I accidentally edited your above comment, instead of quoting it, and then made it worse by trying to restore it. Sorry for this :see_no_evil:

Ahah, no worries! :D

skjnldsv: Done, with testing added

CarlSchwan (approved):
Nice stuff

raimund-schluessler:
I think it works as I should, will do a review later.

But without automatically adjusting to the available size, the usefulness is a bit limited, to be honest. You could achieve the same already by simply putting your inline actions as Buttons next to the `Actions` if you anyway have to specify an inline button count.

skjnldsv:
> But without automatically adjusting to the available size, the usefulness is a bit limited, to be honest. You could achieve the same already by simply putting your inline actions as Buttons next to the `Actions` if you anyway have to specify an inline button count.

Baby steps :)
It gives much more flexibility and is far easier to adjust that copying/moving ButtonVue to Actions back and forth.
Devs are also free to automatically implement the reactivity on the area they're usint the Actions component and therefore adjusting the `inline` prop accordingly.

marcoambrosini (approved):
Code looks good, tested in the docs and works. Awesome stuff :)

Issue #2954, opened 2022-08-05
[BITV] Modal next/previous buttons contrast possibly too low


PVince81:
Steps:
1. Needs https://github.com/nextcloud/nextcloud-vue/pull/2953 (or click https://deploy-preview-2953--nextcloud-vue-components.netlify.app/#/Components/Modal)
2. Open styleguide
3. Scroll down and click on "Show modal with more properties"
4. See the arrow buttons on the left and right of the modal

<img width="700" alt="image" src="https://user-images.githubusercontent.com/277525/183092802-2cfcc1a9-9aaf-4300-88f1-dfdd4369b570.png">

Now this is with a white background from the styleguide + darkened overlay color.

Whatever the solution is, I'm not sure if it can work with all backgrounds ?

The focus effect looks fine though when tabbing:
<img width="154" alt="image" src="https://user-images.githubusercontent.com/277525/183093317-17d19da4-c445-4f34-907e-3c44f84b6bd5.png">

PVince81:
@jancborchardt @nimishavijay

jancborchardt:
Since the background is always made dark with that overlay, the icons should be white – then it works contrast-wise as well. :)

PVince81:
@jancborchardt I assume white only when not hovered

PVince81:
@Pytal can you adjust ? this will is the last item in https://github.com/nextcloud/nextcloud-vue/issues/2501 so we can then close that one too

Pytal:
Fixed in https://github.com/nextcloud/nextcloud-vue/pull/3080

PR #2099, opened 2021-07-09
Fix videoplayer controls in viewer

szaimen:
In some aspect ratios the videoplayer controls are hidden behind the prev and next divs. This PR makes sure that the videoplayer controls are always usable.

## Before
https://user-images.githubusercontent.com/42591237/125089325-973a5780-e0ce-11eb-8671-de043990f7bc.mp4

## After
https://user-images.githubusercontent.com/42591237/125089347-9acdde80-e0ce-11eb-82a6-ca2f5228e5c9.mp4

Signed-off-by: szaimen <szaimen@e.mail.de>

PR #3080, opened 2022-08-20
Fix contrast of next/prev modal buttons

Pytal:
Fix https://github.com/nextcloud/nextcloud-vue/issues/2954

Before | After
--- | ---
image | image
image | image

### DIFF
--- a/src/components/NcModal/NcModal.vue
+++ b/src/components/NcModal/NcModal.vue
@@ -852,14 +852,17 @@ export default {
 		z-index: 10000;
 		// ignore display: none
 		display: flex !important;
-		align-items: center;
-		justify-content: center;
-		width: 8%;
-		min-width: $clickable-area;
 		height: 35vw;
 		position: absolute;
 		transition: opacity 250ms,
-			visibility 250ms;
+		visibility 250ms;
+		color: var(--color-primary-text);
+
+		&:focus-visible {
+			// Override NcButton focus styles
+			box-shadow: 0 0 0 2px var(--color-primary-text);
+			background-color: var(--color-box-shadow);
+		}
 
 		// we want to keep the elements on page
 		// even if hidden to avoid having a unbalanced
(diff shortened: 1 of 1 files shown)


PVince81 (approved):
:+1: nice

artonge (approved):

artonge (changes requested):

Pytal:
> Is it normal that the button is more than half the height of the screen ? Looks weird no ?

Adjusted

artonge (approved):

szaimen (changes requested):
Did someone tested these adjustments with the viewer app on mobile? Will ths next/prev button overlap the video controls again?

See https://github.com/nextcloud/nextcloud-vue/pull/2099

skjnldsv:
Considering the desig nchange and click area, I'd like an approval by @jancborchardt before we merge.

If the click area is so tiny, it will reduce the Viewer usability tremendously.
I'd be up to use proper margins to make sure the visual looks good, but the clickable prev/next area is still taking a fair amount of the page height

Pytal:
> Did someone tested these adjustments with the viewer app on mobile? Will ths next/prev button overlap the video controls again?
> 
> See #2099

Doesn't seem to overlap
image

Pytal:
We can restore the height to `35vw` if that's good with you @jancborchardt @artonge?

jancborchardt:
I'm fine with either height. @skjnldsv good point with the taller tappable area, but could we achieve that without it looking so tall visually as well?

Pytal:
Restored height to `35vw` and updated screenshots

> I'm fine with either height. @skjnldsv good point with the taller tappable area, but could we achieve that without it looking so tall visually as well?

Making part of the tap area visible and the rest invisible does not seem viable so we can just make it all visible as before

skjnldsv (approved):

PR #3081, opened 2022-08-20
Add related resources panel component

Pytal:
Add new component for https://github.com/nextcloud/server/issues/28320

### Screenshot (Shared items tab in Talk)

image

@ArtificialOwl since this component will be used in other apps how do we fetch the related resources of a Talk conversation, Deck board, Calendar event, et cetera?

### DIFF
--- a/src/components/NcRelatedResourcesPanel/NcRelatedResourcesPanel.vue
+++ b/src/components/NcRelatedResourcesPanel/NcRelatedResourcesPanel.vue
@@ -1,0 +1,199 @@
+<!--
+  - @copyright 2022 Christopher Ng <chrng8@gmail.com>
+  -
+  - @author Christopher Ng <chrng8@gmail.com>
+  -
+  - @license AGPL-3.0-or-later
+  -
+  - This program is free software: you can redistribute it and/or modify
+  - it under the terms of the GNU Affero General Public License as
+  - published by the Free Software Foundation, either version 3 of the
+  - License, or (at your option) any later version.
+  -
+  - This program is distributed in the hope that it will be useful,
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+-->
+
+<docs>
+### Usage
+
+Use this component to display the related resources of a given item.
+
+```
+<template>
+	<NcRelatedResourcesPanel provider-id="talk"
+		:item-id="conversationId" />
+</template>
+
+<script>
+export default {
+	data() {
+		return {
+			conversationId: 1,
+		}
+	},
+}
+</script>
+```
+</docs>
+
+<template>
+	<div v-if="appEnabled && isVisible" class="related-resources">
+		<div class="related-resources__header">
+			<h5>{{ headerTranslated }}</h5>
+		</div>
+
+		<NcResource v-for="resource in resources"
+			:key="resource.itemId"
+			class="related-resources__entry"
+			:title="resource.title"
+			:subtitle="resource.subtitle"
+			:tooltip="resource.tooltip"
+			:url="resource.url" />
+	</div>
+</template>
+
+<script>
+import axios from '@nextcloud/axios'
+import { generateOcsUrl } from '@nextcloud/router'
+import { showError } from '@nextcloud/dialogs'
+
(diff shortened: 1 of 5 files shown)


ArtificialOwl:
> [](https://github.com/nextcloud/server/issues/28320)
> 
> @ArtificialOwl since this component will be used in other apps how do we fetch the related resources of a Talk conversation, Deck board, Calendar event, et cetera?

It should be working the same as for Files.
The url of the OCS endpoint is : `/related/{providerId}/{itemId}`

replace `providerId` by `"talk"` or `"deck"`.
replace `itemId` by the string or int Id that identify the current browsed item (room for talk, board for deck)

Pytal:
Keeping as draft for now

src/components/NcRelatedResourcesPanel/NcRelatedResourcesPanel.vue
```diff
@@ -0,0 +1,183 @@
...
+			this.$emit('has-resources', resources.length > 0)
+		},
+	},
+
+	created() {
+		this.fetchRelatedResources()
+	},
+
+	methods: {
+		async fetchRelatedResources() {
```
nickvergessen, line 154: Should abort when any of the provider/item info is missing?

src/components/NcRelatedResourcesPanel/NcResource.vue
```diff
@@ -0,0 +1,112 @@
...
+		<div v-tooltip="tooltip" class="resource__desc">
+			<h5>{{ title }}</h5>
+			<p v-if="subtitle">
+				{{ subtitle }}
+			</p>
+		</div>
+		<NcButton :aria-label="labelTranslated"
+			type="tertiary"
+			:href="url"
+			target="_blank">
```
nickvergessen, line 34: 👎🏼  this is a user decision and we should not force it.

PVince81:
how about renaming from "related resources list" to just "resources list" ? it's just a name change and it already makes it more generic. having a component that is able to list resources can be useful in its own way (even if they are not "related")

src/components/NcRelatedResourcesPanel/NcResource.vue
Pytal: User decision to be able to open the link to the resource?

Pytal:
> how about renaming from "related resources list" to just "resources list" ? it's just a name change and it already makes it more generic. having a component that is able to list resources can be useful in its own way (even if they are not "related")

There's no "related resources list" string do you mean renaming the component to `NcResourceList`?

src/components/NcRelatedResourcesPanel/NcResource.vue
nickvergessen: Yes, new tab/window or not is user's decision. Especially since all those links are internal we should not force _blank

PVince81:
> > how about renaming from "related resources list" to just "resources list" ? it's just a name change and it already makes it more generic. having a component that is able to list resources can be useful in its own way (even if they are not "related")
> 
> There's no "related resources list" string do you mean renaming the component to `NcResourceList`?

I meant that yes. But now going through the code I see that the logic is highly specific to that. I didn't think it would also contain the XHR requests for related resources and more logic. So the current code makes it very specific to related resources and is then reusable.

I still have mixed feelings for having this here, wondering if it shouldn't then be a different nextcloud repo and lib. On the other hand a specific lib for just 1-2 components feels like overkill. So perhaps we just move forward with this for now.

src/components/NcRelatedResourcesPanel/NcResource.vue
```diff
@@ -0,0 +1,111 @@
...
+		},
+		url: {
+			type: String,
+			required: true,
+		},
+	},
+
+	data() {
+		return {
+			labelTranslated: t('Open link to resource'),
```
PVince81, line 80: accessible name must be unique in the whole page (it's cool that you thought of adding one)

if possible, try adding the name in the text of the resource also inside the label

PVince81 (approved):
:+1:

only one comment about accessible name
can also be resolved separately in case it becomes complicated

Pytal:
> I still have mixed feelings for having this here, wondering if it shouldn't then be a different nextcloud repo and lib. On the other hand a specific lib for just 1-2 components feels like overkill. So perhaps we just move forward with this for now.

Yeah this solution as well as the alternative would incur the benefits and drawbacks detailed in https://github.com/nextcloud/nextcloud-vue-dashboard/issues/407

juliushaertl:
Note we branched off a stable6 branch and master will be the next major only compatible with Nextcloud 25+, so if needed please backport this pr also to stable6.

PVince81:
related resources is planned for NC 25 only, so probably no backport needed

juliushaertl (approved):

PR #3087, opened 2022-08-22
Add example with second modal on top

PVince81:
Add Modal example with another one on top to test focus trap.

The focus trap seems to work fine.

There's a bug though: when I submit the second dialog, instead of closing, I get this in the console:
`[Vue warn]: Error in nextTick: "NotFoundError: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node."`

This was an attempt to reproduce https://github.com/nextcloud/server/issues/33637

### DIFF
--- a/src/components/NcModal/NcModal.vue
+++ b/src/components/NcModal/NcModal.vue
@@ -79,8 +79,23 @@ export default {
 				<NcTextField label="First Name" :value.sync="firstName" />
 				<NcTextField label="Last Name" :value.sync="lastName" />
 				<NcButton
-					:disabled="!this.firstName || !this.lastName"
-					@click="closeModal"
+					:disabled="!firstName || !lastName"
+					@click="submitFirstModal"
+					type="primary">
+					Submit
+				</NcButton>
+			</div>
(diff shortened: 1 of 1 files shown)


CarlSchwan:
I wonder if we shouldn't use the native dialog element https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog instead of trying to mess in js the focus order

PVince81:
wow, the dialog element looks pretty recent, I never heard of it before.

the focus trap seemed fine though and I don't think we should switch tech at this point.
would still be good to try it out separately

PVince81:
I discovered that the case I tried to reproduce was different: it was a legacy non-vue dialog appearing on top of a vue dialog. In such case the focus trap of the parent will definitely not stop. Ideal solution: port all dialogs to Vue.

see: https://github.com/nextcloud/server/pull/33516#issuecomment-1225869625

but if not enough we might want to add a function to pause the focus trap on the current dialog.

PVince81:
abandoned/obsolete

PR #3096, opened 2022-08-24
Fix the active state for listitem

GretaD:
we have to force active state styling on mail because otherwise doesnt work. Trying to solve it here on the component


GretaD:
tentative solution here: https://github.com/nextcloud/mail/pull/7078

PR #3100, opened 2022-08-25
Improve multiselect

CarlSchwan:
- Add focus effect for better accessibility
- Increase font size to be 15px like the other component
- Update style to be more like the NcInputField

There are still some remaining issues that are not fixed by this PR,
notably the width of the multiselect component change when clicking
on it. This is a preexisting issue :(

image

### DIFF
--- a/src/components/NcMultiselect/index.scss
+++ b/src/components/NcMultiselect/index.scss
@@ -16,24 +16,6 @@
 		z-index: 2 !important;
 	}
 
-	// active state, force the input to be shown, we don't want
-	// the placeholder or the currently selected options
-	&.multiselect--active {
-		/* Opened: force display the input */
-		input.multiselect__input {
-			opacity: $opacity_full !important;
-			cursor: text !important;
-			// remove border radius on bottom opening
-			border-radius: var(--border-radius) var(--border-radius) 0 0;
-			display: block !important;
-		}
-
-		/* multiselect__limit hidden if active */
-		.multiselect__limit {
-			display: none;
-		}
-	}
-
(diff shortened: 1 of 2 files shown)


jancborchardt (approved):
Looks good!
It's kinda strange that the field itself is double height when open, but guess also preexisting issue.

CarlSchwan:
> Looks good! It's kinda strange that the field itself is double height when open, but guess also preexisting issue.

Yeah here is a before screenshot

image

src/components/NcMultiselect/index.scss
```diff
@@ -165,18 +152,38 @@
...
-			height: 100% !important;
 			border: none;
 			/* override hide to force show the placeholder */
 			/* only when not active */
 			cursor: pointer;
 			/* override inline styling of the lib */
-			padding: 7px 6px !important;
+			padding: 7px 4px !important;
+			display: none;
+            font-size: var(--default-font-size);
```
CarlSchwan, line 162:
```suggestion
		font-size: var(--default-font-size);
```

src/components/NcMultiselect/index.scss
```diff
@@ -165,18 +152,38 @@
...
+			height: 40px !important;
+		}
+
+		/* multiselect__limit hidden if active */
+		.multiselect__limit {
 			display: none;
 		}
 	}
 
+
```
CarlSchwan, line 186:
```suggestion
```

max-nextcloud (approved):
Had a quick pass over the code. Looks legit.

PR #3106, opened 2022-08-26
Fix emptyContent once more

jotoeri:
- Now went back to the action slot, so action and description can be properly formatted here on the component.
- The description also became a prop, thus beeing consistent with the title.
- Adapting resp. docs
- Adapting Usage within DashboardWidget

I considered alternatively to make the old description slot a flex-element for centering items, that however didn't allow me to put some margin between text and action here on the component. So apps would have needed to separately style the distance between.

|Before|After|
|---|---|
|Screenshot from 2022-08-26 09-03-28|Screenshot from 2022-08-26 08-50-34|


skjnldsv (approved):
Sure! :+1:

raimund-schluessler (approved):
Why not.

PR #3108, opened 2022-08-26
Fix modal button visibility

marcoambrosini:


marcoambrosini:
Not needed in the end

PR #3055, opened 2022-08-17
ListItem Design update

marcoambrosini:
Screenshot 2022-08-17 at 16 20 39

Signed-off-by: Marco Ambrosini <marcoambrosini@icloud.com>

PR #3110, opened 2022-08-26
Revert listItem bold title

jotoeri:
As of https://github.com/nextcloud/nextcloud-vue/pull/3055#issuecomment-1225273555, reverting the #3055.

### DIFF
--- a/src/components/NcListItem/NcListItem.vue
+++ b/src/components/NcListItem/NcListItem.vue
@@ -195,6 +195,55 @@
 	</NcListItem>
 </ul>
 ```
+
+### NcListItem not bold or fully bold
+```
+<ul style="width: 350px;">
+	<NcListItem
+		:title="'Title of the element'"
+		:bold="true"
+		:counter-number="1"
+		:compact="true" >
+		<template #icon>
+			<div class="icon-edit" />
+		</template>
+		<template #subtitle>
+			This one is with subtitle
+		</template>
+		<template #actions>
+			<NcActionButton>
+				Button one
+			</NcActionButton>
+			<NcActionButton>
+				Button two
+			</NcActionButton>
+		</template>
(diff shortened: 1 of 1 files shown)


raimund-schluessler (dismissed):

juliushaertl (approved):

marcoambrosini (changes requested):
There has been a misunderstanding, I was talking about the border radius in that comment, not the bolding of the title, which looks very good both in mail and talk. 
I suggest that to address the issue in collectives we have 2 separate props:
`boldTitle` that defaults true
`boldSubtitle` that defaults false
Sorry for the confusion

jotoeri:
https://deploy-preview-3110--nextcloud-vue-components.netlify.app/#/Components/NcListItems

jotoeri:
> `boldTitle` that defaults true
> `boldSubtitle` that defaults false

Oups, sorry. Didn't read again before changing.
Now its `bold` and `boldTitle`. I thought a bold Subtitle on its own is probably not very useful? Hope, that's also fine?

jancborchardt:
> I thought a bold Subtitle on its own is probably not very useful?

True indeed – maybe makes sense to only have "boldTitle" and "boldAll"?

jotoeri:
> True indeed – maybe makes sense to only have "boldTitle" and "boldAll"?

I could change that, yes, makes it however a breaking change then?

marcoambrosini:
> Now its bold and boldTitle. I thought a bold Subtitle on its own is probably not very useful? Hope, that's also fine?

Fine, but then let's bold everything when this is true :)

jotoeri:
> Fine, but then let's bold everything when this is true :)

Yes, that's what it does. There also is an example in the docs, but seems like this is broken currently unfortunately. 😢

Issue #3095, opened 2022-08-24
Update css vars with server ones


juliushaertl:
We should update css vars from server regularly

- [ ] FIgure out dark/light mode options

Quick dirty script that i used for testing:

```
#!/bin/bash

NEXTCLOUD=http://nextcloud.dev.local
TARGET=/tmp/vars-temp
TARGET_OUTPUT=styleguide/assets/variables.css

printf "\n\n/* Plain */\n" > $TARGET
curl $NEXTCLOUD/index.php/apps/theming/theme/default.css?plain=1 >> $TARGET

printf "\n\n/* Light */\n" >> $TARGET
printf "@media (prefers-color-scheme: light) {" >> $TARGET
curl $NEXTCLOUD/index.php/apps/theming/theme/light.css?plain=1 >> $TARGET
printf "}" >> $TARGET

printf "\n\n/* Dark */\n" >> $TARGET
printf "@media (prefers-color-scheme: dark) {" >> $TARGET
curl $NEXTCLOUD/index.php/apps/theming/theme/dark.css?plain=1 >> $TARGET
printf "}" >> $TARGET

cat $TARGET | sed -E "s/\-\-[^:]*:\s*url[(]'[^;]+'\);//g" > $TARGET_OUTPUT
```

raimund-schluessler:
There is a GH action now doing this.

PR #3112, opened 2022-08-29
Update variables.css from nextcloud server

vinicius73:
Related to: #3095

Just update `styleguide/assets/variables.css` using the script below in preparation to automate this process.

```sh
NEXTCLOUD=http://nextcloud.local
TARGET=/tmp/vars-temp
TARGET_OUTPUT=styleguide/assets/variables.css

printf "\n\n/* Plain */\n" > $TARGET
curl $NEXTCLOUD/index.php/apps/theming/theme/default.css?plain=1 >> $TARGET

printf "\n\n/* Light */\n" >> $TARGET
printf "@media (prefers-color-scheme: light) {" >> $TARGET
curl $NEXTCLOUD/index.php/apps/theming/theme/light.css?plain=1 >> $TARGET
printf "}" >> $TARGET

printf "\n\n/* Dark */\n" >> $TARGET
printf "@media (prefers-color-scheme: dark) {" >> $TARGET
curl $NEXTCLOUD/index.php/apps/theming/theme/dark.css?plain=1 >> $TARGET
printf "}" >> $TARGET

cat $TARGET | sed -E "s/\-\-[^:]*:\s*url[(]'[^;]+'\);//g" > $TARGET_OUTPUT
```

### DIFF
--- a/styleguide/assets/variables.css
+++ b/styleguide/assets/variables.css
@@ -1,54 +1,79 @@
+/* Plain */
 :root {
-	--color-main-text: #222;
-	--color-main-background: #fff;
-	--color-main-background-translucent: rgba(255, 255, 255, 0.97);
-	--gradient-main-background: var(--color-main-background) 0%, var(--color-main-background-translucent) 85%, transparent 100%;
-	--color-background-hover: whitesmoke;
+	--color-main-background: #ffffff;
+	--color-main-background-rgb: 255, 255, 255;
+	--color-main-background-translucent: rgba(
+		var(--color-main-background-rgb),
+		0.97
+	);
+	--color-main-background-blur: rgba(var(--color-main-background-rgb), 0.8);
+	--filter-background-blur: blur(25px);
+	--gradient-main-background: var(--color-main-background) 0%,
+		var(--color-main-background-translucent) 85%, transparent 100%;
+	--color-background-hover: #f5f5f5;
 	--color-background-dark: #ededed;
 	--color-background-darker: #dbdbdb;
 	--color-placeholder-light: #e6e6e6;
-	--color-placeholder-dark: #ccc;
+	--color-placeholder-dark: #cccccc;
 	--color-primary: #0082c9;
-	--color-primary-hover: #339bd4;
-	--color-primary-light: #e6f3fa;
-	--color-primary-light-hover: #dce9ef;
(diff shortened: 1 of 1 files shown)


juliushaertl (approved):

raimund-schluessler (changes requested):
This applies random dark styles from the server I guess. Most buttons have a dark background, hover effects are weird, action menus are black. There is more stuff, that is easily visible if one looks at the result in the docs once.

See the button component as example:
Screenshot 2022-08-30 at 10-50-31 Nextcloud Vue Style Guide


Edit: I realized this happens because my system/browser is set to dark mode. But if the styleguide reacts to this setting it should work properly (or be ignored completely, as it is now).

styleguide/assets/variables.css
```diff
@@ -1,54 +1,250 @@
...
+		--header-menu-profile-item-height: 66px;
+		--breakpoint-mobile: 1024px;
+		--primary-invert-if-bright: no;
+		--background-invert-if-dark: no;
+		--background-invert-if-bright: invert(100%);
+	}
+}
+
+/* Dark */
+@media (prefers-color-scheme: dark) {
```
raimund-schluessler, line 167: I would skip the dark color scheme and only update the light one. Otherwise the rest of the styleguide would need adjustments as well, because it looks weird at the moment. Maybe something for a follow-up to have dark mode supported in the styleguide.

juliushaertl: Agreed, we can add the dark one at some point if the style guide supports it.

vinicius73: I believe this is just exposing a problem we already have.
This css is from server, if it looks weird in styleguide, why not is look weird in nextcloud?

raimund-schluessler: Because the styleguide is not Nectcloud server. It does not bring all styles the server does and components are used without the correct containers there, i.e. an app must use NcAppContent as the root container, which the styleguide obviously does not.

vinicius73: I've removed the dark and light entries.
We can discuss it better later.

raimund-schluessler (approved):

skjnldsv:
We could also just load the https://github.com/nextcloud/server/blob/master/apps/theming/css/default.css file :)

PR #3115, opened 2022-08-29
Make action trigger button UI consistent.

vinicius73:
Reported in: https://github.com/nextcloud/text/issues/2829

## Actual

https://user-images.githubusercontent.com/1561347/187306331-d0cb06bc-7566-400a-8264-5070c73fb8d7.mp4

## New

https://user-images.githubusercontent.com/1561347/187306305-bc280a05-05ce-4c3f-a903-47c6d090326f.mp4

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -411,6 +411,107 @@ export default {
 </script>
 ```
 
+### Type variants
+
+```
+<template>
+	<div>
+		<NcActions :type="current">
+			<template #icon>
+				<SelectColor :size="20" />
+			</template>
+
+			<NcActionButton v-if="current" close-after-click @click="define(undefined)">
+				<template #icon>
+					<Delete :size="20" />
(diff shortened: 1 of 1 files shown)


juliushaertl (approved):

src/components/NcActions/NcActions.vue
```diff
@@ -1011,7 +1119,8 @@ export default {
 		return h('div',
 			{
 				class: [
-					'action-item',
+					'action-item action-item--default-popover',
```
raimund-schluessler, line 1122:
```suggestion
					'action-item',
```
The class added is nowhere used, no?

raimund-schluessler (approved):
I would've liked to reuse a class from the `NcButton` component to keep the button styling encapsulated in the `NcButton` component, but looking at its code there is no suitable class and adding one would mean quite a bit of refactoring. So 👍 from my side.

marcoambrosini (approved):

PR #3119, opened 2022-08-30
Use dark themed avatar on dark theme

CarlSchwan:
See https://github.com/nextcloud/server/pull/33752 for screenshots

### DIFF
--- a/src/components/NcAvatar/NcAvatar.vue
+++ b/src/components/NcAvatar/NcAvatar.vue
@@ -321,9 +321,9 @@ export default {
 			}
 
 			if (this.hasStatus && this.showUserStatus && this.showUserStatusCompact) {
-				return t('Avatar of {displayName}, {status}', { displayName: this.displayName || this.userId, status: this.userStatus.status })
+				return t('Avatar of {displayName}, {status}', { displayName: this.displayName ?? this.user, status: this.userStatus.status })
 			}
-			return t('Avatar of {displayName}', { displayName: this.displayName || this.userId })
+			return t('Avatar of {displayName}', { displayName: this.displayName ?? this.user })
 		},
 
 		canDisplayUserStatus() {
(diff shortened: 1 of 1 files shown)


src/components/NcAvatar/NcAvatar.vue
```diff
@@ -559,9 +559,11 @@ export default {
 		 * @return {string}
 		 */
 		avatarUrlGenerator(user, size) {
-			let url = '/avatar/{user}/{size}'
+			const darkTheme = window.getComputedStyle(document.documentElement)
```
juliushaertl, line 562: Can we get the computed style rather on the current element?

juliushaertl (approved):

src/components/NcAvatar/NcAvatar.vue
```diff
@@ -559,9 +559,11 @@ export default {
 		 * @return {string}
 		 */
 		avatarUrlGenerator(user, size) {
-			let url = '/avatar/{user}/{size}'
+			const darkTheme = window.getComputedStyle(this.$el)
+				.getPropertyValue('--background-invert-if-dark') === 'invert(100%)'
```
skjnldsv, line 563: Reeeaaally not fond of this! :see_no_evil:

CarlSchwan: same but I can't find a better way to do it :(

The possibly alternative I considered was using picture/img with srcset. But that would only work for setting the dark theme with media query but not with the explicit dark theme from the theming settings

skjnldsv: I had plans to implement a prop in the ITheme definition for each theme to specify if it's a bright or dark theme.
Didn't find the time, maybe we should for 25

That way we could use `body[data-theme-colour="dark"]` and `body[data-theme-colour="light"]` or the capabilities

CarlSchwan: That won't work with the system theme settings since that this is only client-side and we don't know in php which theme is used.

ChristophWurst (approved):

PR #3132, opened 2022-08-31
Add slot for dot indicator in NcListItem.vue

JuliaKirschenheuter:
Required for https://github.com/nextcloud/mail/issues/6982

Screenshots:
image
image

### DIFF
--- a/src/components/NcListItem/NcListItem.vue
+++ b/src/components/NcListItem/NcListItem.vue
@@ -21,102 +21,127 @@
 <docs>
 
 ### Default Usage
-```
-<ul>
-	<NcListItem
-		:title="'Title of the element'"
-		:bold="false"
-		:details="'1h'"
-		:counter-number="44"
-		counterType="highlighted">
-		<template #icon>
-			<NcAvatar :size="44" user="janedoe" display-name="Jane Doe" />
-		</template>
-		<template #subtitle>
-			In this slot you can put both text and other components such as icons
-		</template>
-		<NcCounterBubble #counter>
-			7
-		</NcCounterBubble>
-		<template #actions>
-			<NcActionButton>
-				Button one
-			</NcActionButton>
-			<NcActionButton>
-				Button two
-			</NcActionButton>
-			<NcActionButton>
-				Button three
-			</NcActionButton>
-		</template>
-	</NcListItem>
-	<NcListItem
-		:title="'Title of the element'"
-		:bold="false"
-		:force-display-actions="true"
-		:details="'1h'"
-		:counter-number="44"
-		counterType="highlighted">
-		<template #icon>
-			<NcAvatar :size="44" user="janedoe" display-name="Jane Doe" />
-		</template>
-		<template #subtitle>
-			In this slot you can put both text and other components such as icons
-		</template>
-		<NcCounterBubble #counter>
-			7
-		</NcCounterBubble>
-		<template #actions>
-			<NcActionButton>
-				Button one
-			</NcActionButton>
-			<NcActionButton>
-				Button two
-			</NcActionButton>
-			<NcActionButton>
-				Button three
-			</NcActionButton>
-		</template>
-	</NcListItem>
-	<NcListItem
-		:title="'Title of the element'"
-		:bold="false">
-		<template #icon>
-			<NcAvatar :size="44" user="janedoe" display-name="Jane Doe" />
-		</template>
-		<template #subtitle>
-			In this slot you can put both text and other components such as icons
-		</template>
-		<NcCounterBubble #counter>
-			7
-		</NcCounterBubble>
-		<template #actions>
-			<NcActionButton>
-				Button one
-			</NcActionButton>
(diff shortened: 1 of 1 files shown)


raimund-schluessler:
@JuliaKirschenheuter Could you please add an example to the docs how to use the indicator slot?

src/components/NcListItem/NcListItem.vue
```diff
@@ -215,7 +215,7 @@
 						:class="{ 'list-item-content__main--oneline': oneLine }">
 
 						<!-- First line, title and details -->
-						<div class="line-one">
+						<div v-if="!$slots.indicator" class="line-one">
```
raimund-schluessler, line 218: I am a bit puzzled by this. If there is an indicator slot, the title will be shown nowhere?

src/components/NcListItem/NcListItem.vue
```diff
@@ -243,6 +243,17 @@
 						</div>
 					</div>
 
+					<div v-if="$slots.indicator && showDetails" class="list-item-content__secondary">
```
raimund-schluessler, line 246: What happens if there is an indicator slot, but `showDetails` is false? Then neither the slot nor `details` or `title` will be shown. This seems not correct.

JuliaKirschenheuter:
> Wouldn't it make sense to put the indicator in line two, next to the counter, instead of creating a `list-item-content__secondary` element and leaving the rest untouched? Judging from the mockup in [nextcloud/mail#6982](https://github.com/nextcloud/mail/issues/6982) this seems most fitting.

You are right, thanks for your feedback!

src/components/NcListItem/NcListItem.vue
JuliaKirschenheuter: right

src/components/NcListItem/NcListItem.vue
JuliaKirschenheuter: right

jancborchardt (approved):
Looks great! :)

src/components/NcListItem/NcListItem.vue
```diff
@@ -233,12 +262,18 @@
...
-							<span v-if="showCounter" class="line-two__counter">
+							<!-- Counter and indicator -->
+							<span v-if="showAdditionalElements" class="line-two__additional_elements">
 								<NcCounterBubble v-if="counterNumber != 0"
+									class="line-two__counter"
 									:type="counterType">
 									{{ counterNumber }}
 								</NcCounterBubble>
+
+								<span class="line-two__indicator">
```
raimund-schluessler, line 273: Since this element has a margin, it will take space even if the slot is empty, which moves the CounterBubble to the left then.
```suggestion
								<span v-if="hasIndicator" class="line-two__indicator">
```
and add a `hasIndicator` variable, similar to e.g. `hasActions`.

JuliaKirschenheuter: please write why should i use hasIndicator (ok, besides additional span) and how should i define it, i  don't know right now

raimund-schluessler: I just named it accordingly to the `hasActions` and `hasSubtitle` variables. It is just meant to hide the `line-two__indicator` entirely if the slot is empty.
Just put
```js
 			if (this.hasIndicator !== !!this.$slots.indicator) { 
 				this.hasIndicator = !!this.$slots.indicator
 			}
```
here
https://github.com/nextcloud/nextcloud-vue/blob/01024eb24f66c0fd015c088e7695ed2567c986c9/src/components/NcListItem/NcListItem.vue#L582-L588

and 
```js
			hasIndicator: false,
```
here
https://github.com/nextcloud/nextcloud-vue/blob/01024eb24f66c0fd015c088e7695ed2567c986c9/src/components/NcListItem/NcListItem.vue#L453

I think this should do it.

JuliaKirschenheuter: ok, done
Thank you!

src/components/NcListItem/NcListItem.vue
```diff
@@ -672,8 +711,14 @@ export default {
 		color: var(--color-text-lighter);
 	}
 
-	&__counter {
-		margin: 2px 4px 0 4px;
+	&__additional_elements {
+		margin-top: 5px;
```
raimund-schluessler, line 715: If we change this, items without indicator start jumping on hover.
```suggestion
		margin: 2px 4px 0 4px;
```

JuliaKirschenheuter: i could't detect it, but will try to apply your solution

raimund-schluessler: > i could't detect it, but will try to apply your solution

Happens for me in the docs when hovering the first ListItem in the example both on Firefox 103 and Chrome 104. The text moves slightly downwards.

raimund-schluessler (approved):

GretaD (approved):
looks good

PR #3139, opened 2022-08-31
[PoC] Use github source css for styleguide

skjnldsv:
Seeing https://github.com/nextcloud/nextcloud-vue/pull/3112

Thoughts?


styleguide/assets/additional.scss
```diff
@@ -1,4 +1,5 @@
-@import 'server.css';
+@import url('https://nextcloud.github.io/server/apps/theming/css/default.css');
+@import url('https://nextcloud.github.io/server/core/css/server.css');
```
skjnldsv, line 2: This is auto deployed by github pages btw
https://github.com/nextcloud/server/settings/pages

styleguide/assets/additional.css
```diff
@@ -1,9 +1,8 @@
-@import 'server.css';
+@import url('https://nextcloud.github.io/server/apps/theming/css/default.css');
+@import url('https://nextcloud.github.io/server/core/css/server.css');
```
vinicius73, line 2: The issue is track and versioning css changes.

skjnldsv: We keep updating it manually to master anyway :shrug:

juliushaertl: Only adds the dependency on keeping the sass compiled css files in the repo, but think that is fine.

skjnldsv: We already have that depepdency. There is also tests on server to make sure this file is up to date.

juliushaertl: Sure, just wanted to have that mentioned if we decide to move that out and make it part of the release process rather than having build assets in the repo. :) I'm definitely all in for doing that now so we have more up to date css on the vue components. 👍

CarlSchwan (approved):

styleguide/assets/additional.css
CarlSchwan: > We already have that depepdency. There is also tests on server to make sure this file is up to date.

We have a task to remove the compiled assets from server (hopefully soon), so this might break at some point.

skjnldsv: No, this one needs to be in for error pages.
But luckily it should not change that frequently as css server variables should be somewhat stable

PR #3140, opened 2022-08-31
Improve text fields

CarlSchwan:
- Fix random name being used since it makes sending form cumbersome and
  also doesn't provide the label hover effect
- Expose more properties from the underlying <input> element

Required for https://github.com/nextcloud/server/pull/33781


ChristophWurst (approved):
Makes sense

raimund-schluessler (approved):

PR #3145, opened 2022-09-01
7.0.0-beta.0

juliushaertl:
Migration guide is likely to be extended with the next beta releases

# Changelog

## [v7.0.0-beta.0](https://github.com/nextcloud/nextcloud-vue/tree/v7.0.0-beta.0) (2022-09-01)

[Full Changelog](https://github.com/nextcloud/nextcloud-vue/compare/v6.0.0-beta.6...v7.0.0-beta.0)

### :boom: Breaking changes

- This component library release requires Nextcloud 25 or later
- We changed the container layout to adapt to the design changes with Nextcloud 25
- The body is no longer the main scroll area but instead the app content
	- Please do careful testing of your user interface, especially in regards to the following components and scroll behaviour:
		- NcContent
		- NcAppNavigation
		- NcAppContent
		- NcAppSidebar

#### Migration guide

- Scroll listeners need to be adapted to watch for the scroll on the NcAppContent container instead of the main document
- If you are using sticky headers in your app you might need to adjust the css top property as the app content container is now the scrollable element. For any contained sticky that had a header height top spacing before you should be able to set that to 0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,27 @@
 
 All notable changes to this project will be documented in this file.
 
+## [v7.0.0-beta.0](https://github.com/nextcloud/nextcloud-vue/tree/v7.0.0-beta.0) (2022-09-01)
+
+[Full Changelog](https://github.com/nextcloud/nextcloud-vue/compare/v6.0.0-beta.6...v7.0.0-beta.0)
+
+### :boom: Breaking changes
+
+- This component library release requires Nextcloud 25 or later
+- We changed the container layout to adapt to the design changes with Nextcloud 25
+- The body is no longer the main scroll area but instead the app content
+	- Please do careful testing of your user interface, especially in regards to the following components and scroll behaviour:
+		- NcContent
+		- NcAppNavigation
+		- NcAppContent
+		- NcAppSidebar
+
+#### Migration guide
+
+- Scroll listeners need to be adapted to watch for the scroll on the NcAppContent container instead of the main document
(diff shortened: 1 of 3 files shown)


PVince81 (approved):
:+1:

raimund-schluessler:
> The body is no longer the main scroll area but instead the app content 

If the app content is scrolled instead of body, I am a bit worried, that the "tap status bar in Safari iOS to scroll to top" does not work anymore, which I think is a really useful feature. Will need to be checked, though.

juliushaertl:
>> The body is no longer the main scroll area but instead the app content
> If the app content is scrolled instead of body, I am a bit worried, that the "tap status bar in Safari iOS to scroll to top" does not work anymore, which I think is a really useful feature. Will need to be checked, though.

There might also be some unexpected scrolling behaviours with the keyboard open on iOS but that was agreed to be accepted with @jancborchardt. The tap to top we hand't noticed yet.

GretaD (approved):

raimund-schluessler:
> > The body is no longer the main scroll area but instead the app content
> 
> If the app content is scrolled instead of body, I am a bit worried, that the "tap status bar in Safari iOS to scroll to top" does not work anymore, which I think is a really useful feature. Will need to be checked, though.

As expected, if body is not the main scroll area anymore, "tap to top" on iOS does not work anymore. I have to admit, I find this very annoying, because I find this feature very useful and use it often. Of course, this is a personal preference of an iOS user, but could you elaborate a bit, why this decision was made?

@jancborchardt @nimishavijay

PR #3149, opened 2022-09-01
Use events provided by floating-vue

raimund-schluessler:
This cleans up the `NcPopover` component and makes use of the events provided by floating-vue. This way we can remove the workaround of using the watcher.


raimund-schluessler:
\backport to stable6

vinicius73 (approved):

juliushaertl (approved):

Issue #3053, opened 2022-08-17
Actions keyboard navigation regression


skjnldsv:
I noticed the keyboard nav is broken, when opening, the first item should be focused, it is not the case anymore.

1. Open the menu
2. Use arrow down to go to the second item
3. See it doesn't work

You have to tab once to get into the first item THEN you can use arrows keys again

cc @raimund-schluessler

raimund-schluessler:
Likely a regression due to https://github.com/nextcloud/nextcloud-vue/pull/2911.

raimund-schluessler:
Fixed with floating-vue@1.0.0-beta.19 and https://github.com/nextcloud/nextcloud-vue/pull/3504.

PR #3151, opened 2022-09-01
Focus first action after opening actions menu

raimund-schluessler:
This is a dirty hack to solve #3053 and get keyboard navigation in the Actions menu to work without pressing tab once.

I have honestly no idea why a timeout is necessary here. All my attempts of using nextTick or listening for floating-vue events (even with #3149 included) did not succeed. The menu is open, and the button is there and defined, but still `focus()` does nothing when called without a timeout directly after the menu was open. The timeout duration is chosen arbitrarily, a timeout of 1 ms works for me as well (0 does not).

Maybe this inspires someone to find the correct solution, if not, I think the PR at least does not make it worse :see_no_evil:

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -1003,6 +1003,7 @@ export default {
 						boundary: this.boundariesElement,
 						container: this.container,
 						popoverBaseClass: 'action-item__popper',
+						noAutoFocus: true,
 					},
 					// For some reason the popover component
 					// does not react to props given under the 'props' key,
@@ -1015,6 +1016,7 @@ export default {
 						boundary: this.boundariesElement,
 						container: this.container,
 						popoverBaseClass: 'action-item__popper',
+						noAutoFocus: true,
 					},
 					on: {
 						show: this.openMenu,


raimund-schluessler:
This might be a problem of floating-vue: https://github.com/Akryum/floating-vue/issues/872 actually, which automatically gives the focus to the popover. I don't see a way to disable this behavior, so the timeout might be our best option for now.

src/components/NcActions/NcActions.vue
```diff
@@ -829,7 +829,9 @@ export default {
 			if (focusElement) {
 				this.removeCurrentActive()
 				const liMenuParent = focusElement.closest('li.action')
-				focusElement.focus()
+				setTimeout(() => {
+					focusElement.focus()
+				}, 10)
```
vinicius73, line 834: After change the state, it is not affect the view (DOM) immediately.
To "wait" for it, just use [`this.nextTick(() => {})`](https://v2.vuejs.org/v2/api/#vm-nextTick)

```
this.$nextTick(() => focusElement.focus())
```

vinicius73 (changes requested):

vinicius73:
> This might be a problem of floating-vue: [Akryum/floating-vue#872](https://github.com/Akryum/floating-vue/issues/872) actually, which automatically gives the focus to the popover. One can actually see, that the first action has focus for a short amount of time, and then loses it again. I don't see a way to disable this behavior, so the timeout might be our best option for now.

I have a problem indirectly related to this.

I didn't find the root cause, but I guess it is related to multiple focus traps activate at the same time. 

`focusElement.focus()` is causing recursive calls of focus trap.

image

image

src/components/NcActions/NcActions.vue
raimund-schluessler: I know, and I tried this. But it does not work. Even with multiple nextTick.

vinicius73: 😨

raimund-schluessler:
> I have a problem indirectly related to this.

I created a new issue for this #3154.

skjnldsv:
https://github.com/Akryum/floating-vue/issues/907

skjnldsv:
It's fixed without this PR :)

skjnldsv:
> It's fixed without this PR :)

AH, not really.
But you don't need the no-auto-focus. It's disabled by default :rocket:

skjnldsv:
So, it is being focused, but pressing tab focus the first element again :thinking: 
I guess there might be some interference. This is a very very minor issue though

raimund-schluessler:
> > It's fixed without this PR :)
> 
> AH, not really. But you don't need the no-auto-focus. It's disabled by default 🚀

Indeed, it works as it is now with https://github.com/Akryum/floating-vue/pull/894. I just checked before touching this PR again, and it didn't, but that was probably a caching issue, I guess.

> So, it is being focused, but pressing tab focus the first element again :thinking: 

What do you mean? After opening the Actions menu, keydown or tab moves to the second action, as it was before we moved to floating-vue. This is how I would expect it to work.

> I guess there might be some interference. This is a very very minor issue though

I was just checking my stale PRs in this repo, and thought it would be an easy fix. Which it turned out to be, since it was in fact working already. Hence, closing the PR.

skjnldsv:
Thanks @raimund-schluessler ! :rocket:

PR #3171, opened 2022-09-05
Release v6.0.0-beta.7

GretaD:
trying again to see if the changelog will be correctly generated this time.


GretaD:
the compare is from v7beta0 to v6-beta7, https://github.com/nextcloud/nextcloud-vue/runs/8187822450?check_suite_focus=true
 so the change log is wrong again, is it ok if i just remove the pr that are not on v6 and release it? Or am i doing smth wrong?

PR #3178, opened 2022-09-05
Right-align nav item utils

ChristophWurst:
Follow-up to https://github.com/nextcloud/nextcloud-vue/pull/3165.

| Before | After
|---|---|
| Bildschirmfoto vom 2022-09-05 19-42-24 | Bildschirmfoto vom 2022-09-05 19-39-01 |


raimund-schluessler (approved):

marcoambrosini (approved):

jancborchardt:
Actually seems a bit too far off to the right now. Better would be to right-align it with the right side of the more-actions icon?

Experimented with 12px margin-right for the counter instead of 2px – but note this needs to be reduced again on hover/focus.

After this PR | With increased margin-right
-|-
image|image

PR #3190, opened 2022-09-07
Allow 'search' as type prop for NcTextField

artonge:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search

### DIFF
--- a/src/components/NcTextField/NcTextField.vue
+++ b/src/components/NcTextField/NcTextField.vue
@@ -173,6 +173,7 @@ export default {
 				'email',
 				'tel',
 				'url',
+				'search',
 			].includes(value),
 		},
 


CarlSchwan (approved):

marcoambrosini (changes requested):
This adds a second clear button at the trailing end. Not sure how we should go forward here. 

Screenshot 2022-09-12 at 08 23 27

One drawback is that this native button cannot be focused.

Maybe if the devs pass in the "search" type we can capture it instead of forwarding it to the input component and take care of the functionality and looks ourselves. So basically displaying the trailing close button and the leading search icon. 
Any other ideas?

artonge:
> This adds a second clear button at the trailing end. Not sure how we should go forward here.

Nice catch. I have added a condition to the trainling icon slot. What do you think ?

marcoambrosini:
I think that our close button is better because
- click target it is big (44px)
- The cursor becomes a pointer when hovering it
- it is focusable
- proper style - uses a mdi icon

https://user-images.githubusercontent.com/26852655/189596979-44598361-819b-49cd-a599-c9c0638bda1d.mov

What would be the advantages of using the native one?

artonge:
> I think that our close button is better because
> 
>     * click target it is big (44px)
> 
>     * The cursor becomes a pointer when hovering it
> 
>     * it is focusable
> 
>     * proper style - uses a mdi icon
> 
> 
> Screen.Recording.2022-09-12.at.09.29.17.mov
> 
> What would be the advantages of using the native one?

If the goal is to have a proper type for the input, and if the browser's close icon is not removable, then I don't see any other solution.

@jancborchardt any opinion on the accessibility benefit of using a proper type for search input ? Maybe it is not worth the trouble

src/components/NcTextField/NcTextField.vue
```diff
@@ -126,7 +126,8 @@ export default {
 		@input="handleInput">
 		<!-- Default slot for the leading icon -->
 		<slot />
-		<template slot="trailing-button-icon">
+		<!-- Trailing icon slot, except for search type input as the browser already adds a trailing close icon -->
+		<template v-if="type !== 'search'" slot="trailing-button-icon" >
```
skjnldsv, line 130:
```suggestion
		<template v-if="type !== 'search'" slot="trailing-button-icon">
```

szaimen (approved):

PR #3191, opened 2022-09-07
Fix the NcPopoverMenu component

CarlSchwan:
This just overwrites some style from the server

Before 
image

After: some spacing between the items and not a ridiculously big border-radius

### DIFF
--- a/src/components/NcPopoverMenu/NcPopoverMenu.vue
+++ b/src/components/NcPopoverMenu/NcPopoverMenu.vue
@@ -58,5 +58,6 @@ export default {
 ul {
 	display: flex;
 	flex-direction: column;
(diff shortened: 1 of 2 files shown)


raimund-schluessler:
Fine with me. But wouldn't it be better to migrate server from NcPopoverMenu (which is deprecated) to NcActions?

CarlSchwan:
Definitively better but also more work

Yesterday, I ported groupfolders app away from the NcPopoverMenu and I might do the same to notificationsupdate tomorrow

ChristophWurst (approved):
:+1: :elephant:

blizzz (approved):
looks nice

PR #3192, opened 2022-09-07
Limit max width of the settings

CarlSchwan:
This option only apply to the title and description by default and is opt-in for the content.

### DIFF
--- a/src/components/NcSettingsSection/NcSettingsSection.vue
+++ b/src/components/NcSettingsSection/NcSettingsSection.vue
@@ -33,15 +33,16 @@ This component is to be used in the settings section of nextcloud.
 	<NcSettingsSection
 		title="Two-Factor Authentication"
 		description="Two-factor authentication can be enforced for all users and specific groups."
-		doc-url="https://docs.nextcloud.com/server/19/go.php?to=admin-2fa">
-		Your settings here
+		doc-url="https://docs.nextcloud.com/server/19/go.php?to=admin-2fa"
+		:limit-width="true">
+		<p>Your settings here</p>
 	</NcSettingsSection>
 </template>
 ```
 </docs>
 
 <template>
-	<div class="settings-section">
+	<div class="settings-section" :class="{'settings-section--limit-width': limitWidth}">
 		<h2 class="settings-section__title">
 			{{ title }}
 			<a v-if="hasDocUrl"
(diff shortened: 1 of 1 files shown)


szaimen (approved):

src/components/NcSettingsSection/NcSettingsSection.vue
```diff
@@ -85,6 +86,16 @@ export default {
...
 		},
+		/**
+		 * Limit the width of the setting's content
+		 *
+		 * By default only the title and description have a limit, use this
+		 * property to also apply this to the rest of the content.
+		 */
+		limitWidth: {
+			type: Boolean,
+			default: false,
```
skjnldsv, line 97: Why not by default? :)

skjnldsv (approved):

CarlSchwan:
There is probably more elements we want to limit the width than just p, h1, h2, ... like the notecard components, input fields, and more

The idea of @skjnldsv to make it on by default, is probably a good one since nc7.0 is a breaking release anyway. We just need to document somewhere that if someone is using NcSettingSection to check if they might want to disable this behavior (e.g. activity app)

src/components/NcSettingsSection/NcSettingsSection.vue
CarlSchwan: Done. I was unsure if this was a good idea but after a night sleeping on it, it indeed make sense since nc7 is a breaking change release.

src/components/NcSettingsSection/NcSettingsSection.vue
```diff
@@ -117,12 +130,17 @@ export default {
 		border-bottom: 1px solid var(--color-border);
 	}
 
+	&--limit-width > * {
+		max-width: 900px;
```
skjnldsv, line 134: $maxWidth

skjnldsv:
```suggestion
		max-width: $maxWidth;
```

jancborchardt:
Ok, sounds good. The admin Logging app is definitely one which needs an exception, but I think it's written in React anyway @icewind1991?

CarlSchwan:
> Ok, sounds good. The admin Logging app is definitely one which needs an exception, but I think it's written in React anyway @icewind1991?

it is, same with the groupfolder app. At some point we need to move that to vue but this can probably wait :)

PR #3197, opened 2022-09-08
Fix NcAppNavigationItem, pass correct boundaries-element NcActions

### LINKED ISSUE #3187: NcAppNavigationItem:  Invalid prop: type check failed for prop "boundariesElement". Expected Element, got String with value "#content-vue"
Using v7 with Mail I get 

```
[Vue warn]: Invalid prop: type check failed for prop "boundariesElement". Expected Element, got String with value "#content-vue".

found in

---> <NcActions>
       <RouterLink>
         <NcAppNavigationItem>
           <NavigationAccount> at src/components/NavigationAccount.vue
             <NcAppNavigation>
               <Navigation> at src/components/Navigation.vue
                 <Content>
                   <Home> at src/views/Home.vue
                     <App> at src/App.vue
                       <Root> vue.runtime.esm.js:4573
```

## Steps to reproduce
1) Open Mail
2) Add an account
3) Open the account's actions menu in the navigation

eneiluj:
Pass a dom element to the boundaries-element prop of NcActions.

closes #3187

### DIFF
--- a/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
+++ b/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
@@ -178,7 +178,7 @@ Just set the `pinned` prop.
 					ref="actions"
 					class="app-navigation-entry__actions"
 					container="#app-navigation-vue"
-					boundaries-element="#content-vue"
+					:boundaries-element="actionsBoundariesElement"
 					:placement="menuPlacement"
 					:open="menuOpen"
 					:force-menu="forceMenu"
@@ -492,6 +492,9 @@ export default {
 
 		undoButtonAriaLabel() {
 			return t('Undo changes')
+		},
+		actionsBoundariesElement() {
+			return document.querySelector('#content-vue') || undefined
 		},
 	},
 


src/components/NcAppNavigationItem/NcAppNavigationItem.vue
```diff
@@ -493,6 +493,9 @@ export default {
 		undoButtonAriaLabel() {
 			return t('Undo changes')
 		},
+		actionsBoundariesElement() {
+			return document.querySelector('#content-vue')
```
raimund-schluessler, line 497: What happens if the element is not found? Does it fall back to undefined, hence body in NcActions?

eneiluj: I didn't check exactly what happens inside but it works fine with an invalid query selector. I guess you're right, prop is null so the default value is used. I'll check that.

eneiluj: @raimund-schluessler Good thing you asked :grin: Default prop value is actually only used if the prop is undefined.

New problem: NcAppNavigationItem's `actionsBoundariesElement` is computed too early so #content-vue does not exist yet.

I didn't face any use case where it is problematic that a nav item action's boundaries are the <body>.
I wonder if specifying the boundaries in this case is really necessary. Do you know that? Or who might know?

GretaD (approved):

PR #3202, opened 2022-09-08
Fix overwriting placeholder property in NcDatetimePicker

Chartman123:
This fixes an issue in Forms where the placeholder for NcDatetimePicker cannot be overwritten by passing some text to the property: https://github.com/nextcloud/forms/issues/1296 

Signed-off-by: Christian Hartmann <chris-hartmann@gmx.de>

### DIFF
--- a/l10n/messages.pot
+++ b/l10n/messages.pot
@@ -122,6 +122,24 @@ msgstr ""
 msgid "People & Body"
 msgstr ""
 
+msgid "Pick a date"
+msgstr ""
+
+msgid "Pick a date and a time"
+msgstr ""
+
+msgid "Pick a month"
+msgstr ""
+
+msgid "Pick a time"
+msgstr ""
+
+msgid "Pick a week"
+msgstr ""
+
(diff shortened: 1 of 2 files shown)


jotoeri:
Can you remove the `this.` from the default-translations, too? So from `this.t('Pick a time')` make `t('Pick a time')`. Then run `npm run l10n:extract` and commit the pot with it.

Then, regarding the components, we have in this PR:
- Fixing translations of the Placeholder of DatePicker
- Adding the possibility to pass custom Placeholders

jotoeri (approved):

CarlSchwan (approved):

Chartman123:
Should we also backport this to the current releases?

skjnldsv:
> Can you remove the `this.` from the default-translations, too? So from `this.t('Pick a time')` make `t('Pick a time')`.

This broke the component: fix in https://github.com/nextcloud/nextcloud-vue/pull/3405

PR #3204, opened 2022-09-08
Fix loading avatar status

CarlSchwan:
Use asset/inline in webpack config and add an example to show that this works

Fix https://github.com/nextcloud/server/issues/33890


szaimen (approved):

ChristophWurst (approved):
:+1: :elephant:

PR #3219, opened 2022-09-10
Allow opting out of automatic NcModal remounting

Pytal:
Automatic remounting behaviour of NcModal is not always desired so we allow passing `null` to prevent this and to keep NcModal mounted at the initial mount point on the DOM

More context in https://github.com/nextcloud/nextcloud-password-confirmation/pull/387


PVince81 (approved):
:+1: makes sense considering the reasons

GretaD (approved):

PR #3225, opened 2022-09-12
Fix a word puzzle in NcDashboardWidget component

CarlSchwan:
Instead of having a prop that allows customizing one part of a sentence, use a prop to allow changing the full sentences.

Part of the fix for https://github.com/nextcloud/activity/issues/919

This an API break so only for nc-7

### DIFF
--- a/l10n/messages.pot
+++ b/l10n/messages.pot
@@ -83,13 +83,10 @@ msgstr ""
 msgid "Hide password"
 msgstr ""
 
-msgid "items"
-msgstr ""
-
 msgid "Message limit of {count} characters reached"
 msgstr ""
 
-msgid "More {dashboardItemType} …"
+msgid "More items …"
 msgstr ""
 
 msgid "Next"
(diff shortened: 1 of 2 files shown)


raimund-schluessler (approved):

raimund-schluessler:
Need to extract the l10n changes.

Also, nc6 was not released yet, so we can still backport breaking changes as well, I would say.

eneiluj (approved):
:+1: 
Sorry about that :grin:

nickvergessen:
🚀

Jerome-Herbinet:
Hello, 
I've noticed that the translated wording isn't used in front-end ; is the issue actually fixed ?
My instance is Nextcloud 25.0.3.

thumb-Clipboard - 9 février 2023 11_44



Cc : @nickvergessen @julien-nc @raimund-schluessler

nickvergessen:
Each app needs to adapt it manually.

nickvergessen:
And the activity app is still using the legacy component:
https://github.com/nextcloud/activity/blob/master/package.json#L84

Issue #3233, opened 2022-09-12
NcAppNavigationItem: actions menu only displayed on hover on mobile


mejo-:
In latest version, the actions menu in `NcAppNavigationItem` is only displayed on hover. While that's nice on desktop, it should not be the case on mobile (where there's no hover).

I guess `!isMobile` should be added to https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/NcAppNavigationItem/NcAppNavigationItem.vue#L172

mejo-:
Closing as discussed in #3234

PR #3234, opened 2022-09-12
NcAppNavigationItem: always display actions menu on mobile

mejo-:
Fixes: #3233

Signed-off-by: Jonas <jonas@freesources.org>

### DIFF
--- a/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
+++ b/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
@@ -169,7 +169,7 @@ Just set the `pinned` prop.
 			<!-- Counter and Actions -->
 			<div v-if="hasUtils && !editingActive"
 				class="app-navigation-entry__utils"
-				:class="{'app-navigation-entry__utils--display-actions': forceDisplayActions || menuOpenLocalValue}">
+				:class="{'app-navigation-entry__utils--display-actions': isMobile || forceDisplayActions || menuOpenLocalValue}">
 				<div v-if="$slots.counter"
 					class="app-navigation-entry__counter-wrapper">
 					<slot name="counter" />


raimund-schluessler:
Not sure about this. The actions are shown when the `NcAppNavigationItem` is active. This means after you selected the item, you can see and click the actions. This should work just fine on mobile as well.

For `NcAppNavigationItem`s that are not supposed to be clickable, you can force the actions to be always visible via `forceMenu`.

I guess this needs input from @jancborchardt.

mejo-:
@raimund-schluessler the problem is that selecting an item might navigate to its details view (e.g. in Collectives). In these cases there's no such state as selecting an item but still viewing the `AppNavigation` on mobile.

nimishavijay:
Have to agree with @raimund-schluessler here, with limited horizontal space on mobile, always showing the action menu can cut off what little text is visible. In https://github.com/nextcloud/nextcloud-vue/issues/2340#issuecomment-910196304 we decided that the menu icon should be visible on mobile only when appropriate, but not as a part of the component by default. So I think it makes more sense for the action menu to stay hidden by default, and depending on how often it would be used, it can be made visible for a particular app.
cc @jancborchardt @marcoambrosini

mejo-:
Will go with this for now.

marcoambrosini:
Yes, In `listItem` we have the ability to force-display the actions, but it is not a default. I think we could do smth like that here too.
https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/NcListItem/NcListItem.vue#L425

jancborchardt:
Agree with @nimishavijay as well, and on mobile there's 2 options:
- Either people manually open the menu again to reach the actions
- Or the detail view always includes the actions as well. This is e.g. how we already do it in Mail, and also how we should ideally do it elsewhere.

Also compare Android or iOS: In message lists, it's not like every entry has a menu. There you can reach the actions via slide or long-press, and the actions of an element are also always inside the detail view.

So it's fine if only the currently selected (and hovered/focused) element has the actions menu shown. :)

PR #3236, opened 2022-09-13
Update design of related resources panel

Pytal:
Before | After
--- | ---
image | image

- [x] Using icons instead of repeated subline
- [x] Have a short explanatory text below the heading as per mockup
- [x] Whole entry should be clickable

`coreApps` and `appswebroots` data was added to resolve some console errors in the styleguide

### DIFF
--- a/src/components/NcRelatedResourcesPanel/NcResource.vue
+++ b/src/components/NcRelatedResourcesPanel/NcResource.vue
@@ -22,18 +22,16 @@
 
 <template>
 	<li class="resource">
-		<div v-tooltip="tooltip" class="resource__desc">
-			<h5>{{ title }}</h5>
-			<p v-if="subtitle">
-				{{ subtitle }}
-			</p>
-		</div>
-		<NcButton :aria-label="labelTranslated"
+		<NcButton class="resource__button"
+			:aria-label="labelTranslated"
 			type="tertiary"
 			:href="url">
 			<template #icon>
-				<ArrowRight :size="20" />
+				<div class="resource__icon">
+					<img :src="icon">
+				</div>
 			</template>
+			{{ title }}
 		</NcButton>
 	</li>
 </template>
@@ -43,13 +41,10 @@ import NcButton from '../NcButton/index.js'
 import Tooltip from '../../directives/Tooltip/index.js'
 import { t } from '../../l10n.js'
 
-import ArrowRight from 'vue-material-design-icons/ArrowRight.vue'
-
 export default {
 	name: 'NcResource',
 
 	components: {
-		ArrowRight,
 		NcButton,
 	},
 
@@ -58,6 +53,10 @@ export default {
 	},
 
 	props: {
+		icon: {
+			type: String,
+			required: true,
+		},
 		title: {
 			type: String,
 			required: true,
@@ -90,22 +89,31 @@ export default {
 	align-items: center;
 	height: 44px;
 
-	&__desc {
-		padding: 8px;
-		line-height: 1.2em;
-		position: relative;
-		flex: 1 1;
-		min-width: 0;
+	// Override default NcButton styles
+	&__button {
+		width: 100%;
+		justify-content: flex-start;
+		padding: 0;
 
-		h5 {
-			white-space: nowrap;
-			text-overflow: ellipsis;
-			overflow: hidden;
-			max-width: inherit;
+		&:deep(.button-vue__text) {
+			font-weight: normal;
+			margin-left: 2px;
 		}
+	}
 
-		p {
-			color: var(--color-text-maxcontrast);
+	&__icon {
+		width: 32px;
+		height: 32px;
+		background-color: var(--color-text-maxcontrast);
+		border-radius: 50%;
+		display: flex;
+		align-items: center;
+		justify-content: center;
+
+		img {
+			width: 16px;
+			height: 16px;
+			filter: var(--background-invert-if-dark);
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


PVince81:
in the original mock up the circles of related resources were aligned with the ones on top

now, it seems the highlight needs some margin anyway to be able to appear, so perhaps it's rather the sharing items that need indenting ? I'm guessing that those don't yet have the new highlight style

PVince81:
it already looks much better!

PVince81 (approved):
:+1: ok to move forward though, indent topic can be addressed separately if needed

src/components/NcRelatedResourcesPanel/NcResource.vue
```diff
@@ -77,9 +88,19 @@ export default {
...
 	data() {
 		return {
+			icon: subtitleIconMap[this.subtitle],
 			labelTranslated: t('Open link to "{resourceTitle}"', { resourceTitle: this.title }),
 		}
 	},
+
+	computed: {
+		isDarkTheme() {
+			return this.colorMainText === '#D8D8D8'
```
ChristophWurst, line 98: this will certainly break at some point :see_no_evil:

ChristophWurst (approved):

src/components/NcRelatedResourcesPanel/NcResource.vue
```diff
@@ -77,9 +88,19 @@ export default {
...
+		},
```
nickvergessen, line 99:
```suggestion
```

src/components/NcRelatedResourcesPanel/NcResource.vue
```diff
@@ -22,33 +22,44 @@
...
-		<NcButton :aria-label="labelTranslated"
+		<NcButton class="resource__button"
+			:aria-label="labelTranslated"
 			type="tertiary"
 			:href="url">
 			<template #icon>
-				<ArrowRight :size="20" />
+				<div class="resource__icon">
+					<img :class="{ 'resource__icon--invert': isDarkTheme }"
+						:src="icon">
```
nickvergessen, line 32:
```suggestion
					<img :src="icon">
```

src/components/NcRelatedResourcesPanel/NcResource.vue
```diff
@@ -89,22 +110,27 @@ export default {
...
+		img {
+			width: 16px;
+			height: 16px;
 		}
 
-		p {
-			color: var(--color-text-maxcontrast);
+		&--invert {
+			filter: invert(1);
 		}
```
nickvergessen, line 134:
```suggestion
		filter: var(--background-invert-if-dark);
```

nickvergessen (dismissed):
Fix dark mode detection

src/components/NcRelatedResourcesPanel/NcResource.vue
nickvergessen: See comments in https://github.com/nextcloud/nextcloud-vue/pull/3236#pullrequestreview-1105320756 for solution

src/components/NcRelatedResourcesPanel/NcResource.vue
```diff
@@ -22,33 +22,44 @@
...
+/**
+ * Subtitles are implemented by providers with `\OCA\RelatedResources\IRelatedResourceProvider::setSubtitle()`
+ *
+ * FIXME icons should be provided by the backend https://github.com/nextcloud/related_resources/issues/15
+ */
+const subtitleIconMap = {
+	Calendar: imagePath('calendar', 'calendar.svg'),
+	'Deck board': imagePath('deck', 'deck.svg'),
+	Files: imagePath('files', 'app.svg'),
+	'Talk Room': imagePath('spreed', 'app.svg'),
```
nickvergessen, line 56: Totally agree with the fixme.
Also the backends should be in the respective apps, but other dicussion.

Anyway this here should be 
```suggestion
	'Talk conversation': imagePath('spreed', 'app.svg'),
```

We never use the word Room in the UI in talk

nickvergessen: Also this breaks when translating is fixed:
https://github.com/nextcloud/related_resources/issues/16

jancborchardt (changes requested):
Looks very nice already! Some additional feedback (all as per mockup):
- [ ] The heading and subline:
  - [ ] Can use even more whitespace above, e.g. 44px
  - [ ] Should be indented to be left-aligned with the rest of the text as per mockup
  - [ ] The heading "Related resources" can be bold for better scannability
- [ ] The entries
  - [ ] Should not be bold
  - [ ] Don't need the starting slash for Files
  - [ ] The text alignment with the share part up top as @PVince81 mentioned :)

Otherwise nice!

jancborchardt:
Ah and btw, the subline wording

> Anything shared eith the same group of people will show up here

is this still how the algorithm works? Or is it drastically different? @ArtificialOwl 
If so we might have to reword.

src/components/NcRelatedResourcesPanel/NcResource.vue
Pytal: Done in https://github.com/nextcloud/related_resources/pull/21 and https://github.com/nextcloud/related_resources/pull/22

Pytal:
Styles + screenshot updated

Pytal:
Requires release with upstream fixes

- https://github.com/nextcloud/related_resources/pull/21
- https://github.com/nextcloud/related_resources/pull/22
- https://github.com/nextcloud/related_resources/pull/24

PVince81:
@jancborchardt please recheck the screenshots

note: if we don't get feedback we'll merge anyway later today to get it shipped and further changes will need to happen separately

jancborchardt (approved):
Awesome work @Pytal, looks really great now! :)

PR #3241, opened 2022-09-13
Align app navigation item and caption

ChristophWurst:
Found in Calendar

| Before | After
|---|---|
| Bildschirmfoto vom 2022-09-13 15-55-49 | Bildschirmfoto vom 2022-09-13 15-53-46 |

### DIFF
--- a/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
+++ b/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
@@ -765,7 +765,7 @@ export default {
 	/* counter */
 	.app-navigation-entry__counter-wrapper {
 		// Add slightly more space to the right of the counter
-		margin-right: calc(var(--default-grid-baseline) * 2);
+		margin-right: calc(var(--default-grid-baseline) * 2); /* same as NcAppNavigationCaption */
 		display: flex;
 		align-items: center;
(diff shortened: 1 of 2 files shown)


src/components/NcAppNavigationItem/NcAppNavigationItem.vue
```diff
@@ -765,7 +765,7 @@ export default {
 	/* counter */
 	.app-navigation-entry__counter-wrapper {
 		// Add slightly more space to the right of the counter
-		margin-right: 2px;
+		margin-right: 2px; /* same as NcAppNavigationCaption */
```
vinicius73, line 768: I would to say to use a css variable here, although it is not possible to do it in a good way since the css of components are fully isolate.

vinicius73 (approved):

src/components/NcAppNavigationItem/NcAppNavigationItem.vue
ChristophWurst: It uses a var now thanks to a conflict :)

ChristophWurst:
Bildschirmfoto vom 2022-09-14 17-58-41

This got just worse with https://github.com/nextcloud/nextcloud-vue/pull/3250

I don't know how to fix this now. I'll leave this to a designer.

PR #3245, opened 2022-09-13
Forward the autofocus attribute to input field component

kesselb:
Make it possible to set the `autofocus` attribute on on TextField and PasswordField


PVince81 (approved):
:+1:

nickvergessen (approved):

PR #3252, opened 2022-09-14
Fix multiselect background color spilling outside the component

ChristophWurst:
If you put the multiselect somewhere other than an element with the main background color, it looked weird.

| Before | After |
|---|---|
| Bildschirmfoto vom 2022-09-14 11-50-55 | Bildschirmfoto vom 2022-09-14 11-53-22 |


GretaD (approved):
code looks good

szaimen (approved):

PR #3259, opened 2022-09-14
Mark resource button styles as important to fix styles in some contexts

Pytal:
Fix button styles in https://github.com/nextcloud/spreed/pull/7789


Pytal:
Would need another beta release

PVince81 (approved):
:+1:

GretaD (approved):

Issue #2975, opened 2022-08-09
Add live refresh for Avatar


Pytal:
Live refresh the user's avatar image on update

PR #3266, opened 2022-09-16
Live refresh image and displayname based avatar on update

Pytal:
Close https://github.com/nextcloud/nextcloud-vue/issues/2975

Listen to avatar https://github.com/nextcloud/server/blob/b9c002df0a3ab80c1612206cfb1f724d90e0da3a/apps/settings/src/components/PersonalInfo/AvatarSection.vue#L268 and displayname https://github.com/nextcloud/server/blob/b9c002df0a3ab80c1612206cfb1f724d90e0da3a/apps/settings/src/components/PersonalInfo/DisplayNameSection.vue#L66 update events to automatically refresh the image or displayname based avatar

### DIFF
--- a/src/components/NcAvatar/NcAvatar.vue
+++ b/src/components/NcAvatar/NcAvatar.vue
@@ -476,6 +476,8 @@ export default {
 
 	mounted() {
 		this.loadAvatarUrl()
+		subscribe('settings:avatar:updated', this.loadAvatarUrl)
+		subscribe('settings:display-name:updated', this.loadAvatarUrl)
 		if (this.showUserStatus && this.user && !this.isNoUser) {
 			if (!this.preloadedUserStatus) {
 				this.fetchUserStatus(this.user)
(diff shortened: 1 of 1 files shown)


Pytal:
Actually don't think this needs to be toggleable but what do others think?

src/components/NcAvatar/NcAvatar.vue
```diff
@@ -490,6 +492,8 @@ export default {
 	},
 
 	beforeDestroy() {
+		unsubscribe('settings:avatar:updated', this.loadAvatarUrl)
+		unsubscribe('settings:display-name:updated', this.loadAvatarUrl)
```
skjnldsv, line 496: I would have used the proper context ?

```suggestion
		unsubscribe('user:avatar:updated', this.loadAvatarUrl)
		unsubscribe('user:display-name:updated', this.loadAvatarUrl)
```

skjnldsv (approved):

szaimen (approved):
lgtm but didn't test

PR #3268, opened 2022-09-19
Bind NcActions props to NcPopover component

marcoambrosini:
Signed-off-by: Marco <marcoambrosini@icloud.com>

### DIFF
--- a/src/components/NcPopover/NcPopover.vue
+++ b/src/components/NcPopover/NcPopover.vue
@@ -85,7 +85,7 @@ The prop `:focus-trap="false"` help to prevent it when the default behavior is n
 	<Dropdown ref="popover"
 		:distance="10"
 		:arrow-padding="10"
-		v-bind="$attrs"
+		v-bind="$props"
 		:popper-class="popoverBaseClass"
 		v-on="$listeners"
 		@apply-show="afterShow"
@@ -120,6 +120,36 @@ export default {
 		focusTrap: {
 			type: Boolean,
 			default: true,
+		},
+
+		delay: {
+			type: Number,
+			default: 0,
+		},
+
+		handleResize: {
+			type: Boolean,
+			default: true,
+		},
+
+		shown: {
+			type: Boolean,
+			default: false,
+		},
+
+		placement: {
+			type: String,
+			default: 'auto',
+		},
+
+		boundary: {
+			type: String,
+			default: 'scrollParent',
+		},
+
+		container: {
+			type: String,
+			default: 'body',
 		},
 	},
 
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -1003,18 +1003,6 @@ export default {
 						container: this.container,
 						popoverBaseClass: 'action-item__popper',
 					},
-					// For some reason the popover component
-					// does not react to props given under the 'props' key,
-					// so we use both 'attrs' and 'props'
-					attrs: {
-						delay: 0,
-						handleResize: true,
-						shown: this.opened,
-						placement: this.placement,
-						boundary: this.boundariesElement,
-						container: this.container,
-						popoverBaseClass: 'action-item__popper',
-					},
 					on: {
 						show: this.openMenu,
 						'after-show': this.onOpen,


nickvergessen (approved):

raimund-schluessler (changes requested):
No, this doesn't work. Try setting the `open` prop for an NcActions component now:
Screenshot 2022-09-19 at 10-52-08 Nextcloud Vue Style Guide

The popover does not open anymore.

marcoambrosini:
so the problem here is that the Actions component doesn't actually declare the props. Declaring them fixes the issues. Should we use $attrs or $props? I think that $props is the way to go

raimund-schluessler:
> so the problem here is that the Actions component doesn't actually declare the props.

Do you mean the `NcPopover` component? If `props` works, I would go for it.

raimund-schluessler:
> Declaring them fixes the issues. Should we use $attrs or $props? I think that $props is the way to go

I don't know whether you have already adjusted the fix, but the `open` prop of `NcActions` still does not work.

GretaD (approved):
tested with contacts and the errors are gone, thanks a lot marco

raimund-schluessler:
> tested with contacts and the errors are gone, thanks a lot marco

Which errors actually? There is no bug description available. And it still breaks `open` and `placement` props.

nickvergessen:
In Talk action menus of the chat messages are cut off on lower widths
grafik

raimund-schluessler:
> In Talk action menus of the chat messages are cut off on lower widths grafik

My guess is that you set a `container` for the `NcActions` there, and it only gets "fixed" by this PR, because the forwarding to the `NcPopover` gets broken, and it falls back to `body`. Can you show me the code where this menu is generated, please?

nickvergessen:
https://github.com/nextcloud/spreed/pull/7876/files

Might be correct as @marcoambrosini changed the container in the meantime to the scroller?

raimund-schluessler:
> https://github.com/nextcloud/spreed/pull/7876/files
> 
> Might be correct as @marcoambrosini changed the container in the meantime to the scroller?

That is very likely the case. The scroller probably has a lower z-index than the sidebar, or overflow:hidden and then the menu is cut off. If you break the forwarding of the container prop with this PR here, `container` will be `body` and the menu is shown.

I propose to remove the container prop in spreed, or set it to `body` (or anything else which does not cut-off the menu), instead of breaking the `NcActions` component 😉

GretaD:
> Which errors actually? There is no bug description available. And it still breaks `open` and `placement` props.

[Vue warn]: $attrs is readonly. [Vue warn]: $listeners is readonly. and some other like this one.
Also, i dont have the menu problem on contacts or mail, its not cut off

raimund-schluessler:
> > Which errors actually? There is no bug description available. And it still breaks `open` and `placement` props.
> 
> [Vue warn]: $attrs is readonly. [Vue warn]: $listeners is readonly. and some other like this one. Also, i dont have the menu problem on contacts or mail, its not cut off

Hm, that seems indeed related. But not forwarding the props/attrs anymore is no solution either, I would say.

marcoambrosini:
Yep this is not fixing anything for now, just braking more.. And yeah the menus worked cause they were appended to the body. Pushing fix soon

raimund-schluessler:
> > Which errors actually? There is no bug description available. And it still breaks `open` and `placement` props.
> 
> [Vue warn]: $attrs is readonly. [Vue warn]: $listeners is readonly. and some other like this one. Also, i dont have the menu problem on contacts or mail, its not cut off

Following SO, this error is likely due to having multiple instances of Vue in your bundle, this is the SO thread:
https://stackoverflow.com/questions/53206078/vue-warn-listeners-and-attrs-is-readonly

Adding an alias to the webpack config might prevent this, like so:
https://v2.vuejs.org/v2/guide/installation.html#Runtime-Compiler-vs-Runtime-only

```js
module.exports = {
  // ...
  resolve: {
    alias: {
      'vue$': 'vue/dist/vue.esm.js' // 'vue/dist/vue.common.js' for webpack 1
    }
  }
}
```

marcoambrosini:
What do you think of this approach @raimund-schluessler ?

src/components/NcPopover/NcPopover.vue
```diff
@@ -121,6 +121,36 @@ export default {
...
+			default: false,
+		},
+
+		placement: {
+			type: String,
+			default: 'auto',
+		},
+
+		boundary: {
+			type: String,
```
raimund-schluessler, line 146: Floating vue expects a DOM element. This triggers an invalid prop type check now.
```
[Vue warn]: Invalid prop: type check failed for prop "boundary". Expected String, got HTMLBodyElement 

found in

---> <NcPopover> at src/components/NcPopover/NcPopover.vue
       <NcActions> at src/components/NcActions/NcActions.vue
         <TaskBody> at src/components/TaskBody.vue
           <Draggable>
             <TaskDragContainer> at src/components/TaskDragContainer.vue
               <General> at src/views/AppContent/General.vue
                 <Collections> at src/views/AppContent/Collections.vue
                   <NcAppContent> at src/components/NcAppContent/NcAppContent.vue
                     <NcContent> at src/components/NcContent/NcContent.vue
                       <App> at src/App.vue
                         <Root> tasks-main.js:260533:21
```

src/components/NcPopover/NcPopover.vue
```diff
@@ -121,6 +121,36 @@ export default {
...
+			default: 'scrollParent',
```
raimund-schluessler, line 147: This is to specific.

raimund-schluessler:
> What do you think of this approach @raimund-schluessler ?

Honestly, I don't think it is a good idea.
- It doesn't fix what I think you want to fix. The `$attrs/$listeners is readonly` error is still present when I run this PR. And I honestly think it has nothing to do with our `NcPopover` implementation, but with a duplication of the `vue` dependency, see https://github.com/nextcloud/nextcloud-vue/pull/3268#issuecomment-1254941633. When I add this piece of code
  ```
  webpackConfig.resolve.alias = {
	  "vue$": path.resolve('./node_modules/vue')
  }
  ```
  to my webpackConfig e.g. in Tasks, the `$attrs is readonly` warning is gone and the package is smaller, indicating that the duplication is gone. Btw, for vue3 I had to add this to the webpack config, otherwise the app wouldn't run at all, see here:
  https://github.com/nextcloud/webpack-vue-config/blob/800d666f1117523481f519d7b6a15e4fe49576fc/webpack.config.js#L117-L119
- Your changes still break stuff (invalid prop type check, to specific default, see my comments to the code).
- The changes make it harder to maintain our component, e.g. when floating-vue changes something in their props, we need to follow explicitly. And in many cases here we just forward the attrs.

So, in conclusion, I strongly vote to leave the component as it is, and rather look into deduplicating the vue dependency to fix the attrs issue.

marcoambrosini:
Fine by me :)

PR #3271, opened 2022-09-20
Fix initials color of new non-user avatars

ChristophWurst:
Ref https://github.com/nextcloud/mail/issues/7294
Ref https://github.com/nextcloud/server/pull/33752

| Before | After |
|---|---|
| Bildschirmfoto vom 2022-09-20 08-48-29 | Bildschirmfoto vom 2022-09-20 09-00-32 |

(Demo requires https://github.com/nextcloud/nextcloud-vue/pull/3270)

### DIFF
--- a/src/components/NcAvatar/NcAvatar.vue
+++ b/src/components/NcAvatar/NcAvatar.vue
@@ -129,7 +129,7 @@ export default {
 			:class="'avatardiv__user-status--' + userStatus.status" />
 
 		<!-- Show the letter if no avatar nor icon class -->
-		<div v-if="userDoesNotExist && !(iconClass || $slots.icon)" class="unknown">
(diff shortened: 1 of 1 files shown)


nickvergessen (approved):

src/components/NcAvatar/NcAvatar.vue
```diff
@@ -405,6 +405,12 @@ export default {
 			}
 			return style
 		},
+		initialsStyle() {
+			const { r, g, b } = usernameToColor(this.getUserIdentifier)
+			return {
+				color: `rgba(${r}, ${g}, ${b})`
```
tcitworld, line 411:
```suggestion
				color: `rgb(${r}, ${g}, ${b})`
```
Since no 4th argument.

tcitworld (changes requested):

raimund-schluessler (approved):

PR #3278, opened 2022-09-20
Properly align AppNavigationToggle with first navigation item

susnux:
Currently the toggle is set at `4px` while the app navigation items are placed at `8px` (using that variable), this looks slightly displaced.

So this PR moves the padding calculation into a variable and uses the variable for the toggle (besides the navigation items).

**Before / Currently:**
toggle slightly misplaced

**After / This PR:**
Toggle centered

### DIFF
--- a/src/components/NcAppNavigation/NcAppNavigation.vue
+++ b/src/components/NcAppNavigation/NcAppNavigation.vue
@@ -142,6 +142,14 @@ export default {
 }
 </script>
 
+<style lang="scss">
+.app-navigation,
+.app-content {
+	/** Distance of the app naviation toggle and the first navigation item to the top edge of the app content container */
+	--app-navigation-padding: #{$app-navigation-padding};
+}
+</style>
+
 <style lang="scss" scoped>
 .app-navigation {
 	// Set scoped variable override
@@ -184,7 +192,7 @@ export default {
 		display: flex;
 		flex-direction: column;
 		gap: var(--default-grid-baseline, 4px);
-		padding: calc(var(--default-grid-baseline, 4px) * 2);
+		padding: var(--app-navigation-padding);
 	}
 	&__content {
 		height: 100%;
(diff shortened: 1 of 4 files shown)


ChristophWurst:
I'm not able to see a difference between the before and after screenshot

susnux:
> I'm not able to see a difference between the before and after screenshot

Have a close look on the navigation toggle, on the `before` screenshot it is slightly of-center (4px displaced to the top).
As the navigation items use a padding of 8px => spaced 8px from the top, but the navigation toggle uses a margin of 4px.

ChristophWurst:
@GretaD please test how this aligns with the Groupware apps

GretaD:
on main
Screenshot from 2022-09-21 19-04-52

on this pr is slightly down(red background is added so i can tell im connected to nc/vue, ignore that)
if we want to keep this change, we can change the custom padding for search box to fit in. Same for mail with the button.
Screenshot from 2022-09-21 19-02-35

susnux:
My reason why I think fixing the alignment here would be better then manually align it in every project is that many apps use the default styling, so e.g. even the user settings (`/settings/users`) page is affected:

offset of toggle to first nav. item

skjnldsv:
This issue has been here for a while, talk shift it manually, photos too, having a proper standard would indeed be nice yeah :)

susnux:
Rebased this and used a css variable for easier adjustment as wished.

I think with current master being version 8 for NC28 we give apps enough time to adjust.

susnux:
Ping?

marcoambrosini (approved):

src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -142,6 +142,13 @@ export default {
 }
 </script>
 
+<style lang="scss">
+.app-navigation,.app-content {
```
ShGKme, line 146:
```suggestion
.app-navigation,
.app-content {
```

ShGKme (approved):

PR #3285, opened 2022-09-22
Don't explicitly foward attrs to NcPopover

raimund-schluessler:
Signed-off-by: Raimund Schlüßler <raimund.schluessler@mailbox.org>


raimund-schluessler:
Not explicitly forwarding the attrs also breaks it. Closing.

Issue #3198, opened 2022-09-08
color-text-maxcontrast needs to use darker value in new left navigation


jancborchardt:
Since the new left navigation is slightly transparent, we need to use a slightly darker `--color-text-maxcontrast` value when it’s used in there.

To be clear: Only for the usage in there, we should **not** increase the value of the variable overall, as then it will reduce the contrast/difference to `--color-main-text`.

Of course there are different backgrounds, but let’s start out by making sure it works for the default background.

With current `#767676` (maxcontrast) for subline | With better `#646464`
-|-
![image](https://user-images.githubusercontent.com/925062/189121169-29d195cf-3f42-4676-a08c-3f4068c05a11.png)|![image](https://user-images.githubusercontent.com/925062/189122098-ec3eaf99-2214-4e99-8db6-85551b083137.png)
![image](https://user-images.githubusercontent.com/925062/189122340-c08f27cf-12aa-495c-929e-351ca24db3a9.png)|![image](https://user-images.githubusercontent.com/925062/189122252-26d814a3-7a26-4336-9ea2-1095c927b7f2.png)

----

Also, `.line-two__subtitle` uses the outdated variable `var(--color-text-lighter)` – we only use `--color-main-text` and `--color-text-maxcontrast`, as per the design guidelines: https://docs.nextcloud.com/server/latest/developer_manual/design/foundations.html#text-color

jancborchardt:
@juliushaertl @CarlSchwan we should ideally fix this for the release as it relates to the accessibility of a very prominent element. What do you think of the proposal?

AndyScherzinger:
Sounds good to me 👍 
Also the change is very subtle - thus not impacting screenshots

Pytal:
This uses server-side calculated color variables so splitting up calculations with new client-side logic wouldn't be an option, seems that adding a new variable would be the way to go then @skjnldsv?

Would mean replacing `--color-text-lighter` in `NcListItem` with the new variable

jancborchardt:
@Pytal yep, sounds good! Note that `--color-text-lighter` and `--color-text-light` are actually not recommended to use. We use:
- `--color-main-text`
- `--color-text-maxcontrast`

[...]

PR #3307, opened 2022-09-28
Fix contrast of maxcontrast text on blurred background

Pytal:
Fix https://github.com/nextcloud/nextcloud-vue/issues/3198

Light theme before/after in https://github.com/nextcloud/nextcloud-vue/issues/3198

### Dark theme

Before (`#8C8C8C`) | After (`#919191`)
--- | ---
image | image
image | image

> Minimum 4.5 contrast ratio

Requires variables added in https://github.com/nextcloud/server/pull/34299

### DIFF
--- a/src/components/NcAppNavigation/NcAppNavigation.vue
+++ b/src/components/NcAppNavigation/NcAppNavigation.vue
@@ -137,8 +137,11 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-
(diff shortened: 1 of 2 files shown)


jancborchardt (approved):
Awesome, thank you! :)

ChristophWurst (approved):
:mag: :+1: :elephant:

Issue #3154, opened 2022-09-02
Recursive calls of focusTrap() in NcActions component


raimund-schluessler:
> This might be a problem of floating-vue: [Akryum/floating-vue#872](https://github.com/Akryum/floating-vue/issues/872) actually, which automatically gives the focus to the popover. One can actually see, that the first action has focus for a short amount of time, and then loses it again. I don't see a way to disable this behavior, so the timeout might be our best option for now.

I have a problem indirectly related to this.

I didn't find the root cause, but I guess it is related to multiple focus traps activate at the same time. 

`focusElement.focus()` is causing recursive calls of focus trap.

![image](https://user-images.githubusercontent.com/1561347/188007697-96ccf6cd-77c5-4387-acbe-71d4c3b3aa23.png)

![image](https://user-images.githubusercontent.com/1561347/188007801-e19849de-5c17-4143-8dc0-6413480f1f51.png)

_Originally posted by @vinicius73 in https://github.com/nextcloud/nextcloud-vue/issues/3151#issuecomment-1234755061_

raimund-schluessler:
@vinicius73 Since I think this is a problem different from the one approached in #3151, I created a new issue.

However, what I realized while debugging #3151 is that this:
https://github.com/nextcloud/nextcloud-vue/blob/1ef7b682202424ca1fdeaa9fdd5509486fefc4c0/src/components/NcActions/NcActions.vue#L774-L791
is called on every mouse move in the actions. Hence, your line in question
https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/NcActions/NcActions.vue#L832
is also called on nearly every move (depends a bit what is hovered). So this might create trouble with the focus trap, but I didn't check further so far.

Anyway, I think how we focus an Action on mouseover/move needs refactoring, since there is to much going on for a simple focus that should be called once and not on every mouse move. Maybe this could solve the focusTrap issue as well.

max-nextcloud:
This recursion happens if there are two competing focus traps. In text vinicius is fixing it by moving the Menubar (which had focus traps for the dropdowns) into the modal which was also using focus traps. Now one focus trap is inside the other which is fine. 

However this is not an option currently for NcActions in the sidebar when the modal is open. Sidebar lives outside of the modal and so do all of its children. So they compete with the focus trap of the modal.

Issue #3300, opened 2022-09-27
Slow menu opening since nc 7


skjnldsv:
![image](https://user-images.githubusercontent.com/14975046/192544946-73f6516f-e97a-4f71-bafb-dd9568b6bc6d.png)

400ms to open a menu is not acceptable, debugging the profiler gives the above culprit.
Something must have changed.

It request an animation frame and block the browser whch drop my fps to `2`. 
I have 45GB of free ram and 4 full cpu core available (intel i5, so hardware is not an issue)

@raimund-schluessler @marcoambrosini

raimund-schluessler:
There is not really any difference of `NcActions` between stable6 and master:
https://github.com/nextcloud/nextcloud-vue/compare/stable6#diff-a09afa053a2a82c8e733f6d785927c30bcb3f0969d82f5ccfda36861a4af553d

raimund-schluessler:
@skjnldsv Maybe this is related to the focus-trap issue https://github.com/nextcloud/nextcloud-vue/issues/3154#issuecomment-1259603481

skjnldsv:
Might be! :shrug:

raimund-schluessler:
@skjnldsv @vinicius73 Is this fixed with https://github.com/nextcloud/nextcloud-vue/pull/3347?

skjnldsv:
Should be yes

skjnldsv:
But this wasnot included in 7.0.0 :(

raimund-schluessler:
> But this wasnot included in 7.0.0 :(

Looks like 7.0.1 needs to come soon.

PR #3308, opened 2022-09-28
Use inject/provide to define popover/dropdown container

vinicius73:
It helps to define the container of all popovers in the component tree at the same time.

Solving part of #3154 

```vue
<template>
<div :id="randomID">
    <!-- any popover (NcActions, NcEmojiPicker, ...)  will use the randomID as container by default -->
</div>
</template>

<script>
import { defineComponent } from 'vue'
import { NC_KEY_POPOVER_CONTAINER } from '@nextcloud/vue'

const randomHash = () => (Math.ceil((Math.random() * 100000) + 1000)).toString(16)

export default defineComponent({
  data() {
    return {
       randomID: `custom-container-${randomHash()}`,
    }
  },
  
  provide () {
    return {
       [NC_KEY_POPOVER_CONTAINER]: `#${this.randomID}`,
    }
  }
})
</script>

```


vinicius73:
@raimund-schluessler this problem appears when we have more than one instance of `focus trap` active ate the same time (as mentioned https://github.com/nextcloud/nextcloud-vue/issues/3154 and https://github.com/focus-trap/focus-trap/issues/428) 

In `text` it happens when `text` is open in the `viwer` (a `NcModal`) and an option from menu bar is open (https://github.com/nextcloud/text/issues/2842)  

The solution was moving the container of `NcMenu`  (who uses `NcPopover` under the hood) to `MenuBar` element (https://github.com/nextcloud/text/pull/3066).

If the `focus trap` is created inside a children element of other `focus trap`, we do not have this error.

We've faced the same problem in `server` (https://github.com/nextcloud/nextcloud-vue/issues/3300)
If any viewer is open, interact with sidebar (specially in share options) will trigger the same error.

The viewer app already handles part of this problem

https://github.com/nextcloud/viewer/blob/2b5ed86ec105f65b8e3990373885f59dc9772749/src/views/Viewer.vue#L906-L913

Catching the sidebar event, and update its focus trap instance.

So, using a common container solve the problem without hard changes in next-cloud components.

---

This PR does not change any actual behavior of `NcPopover`. 

In some cases, we have a lot of `NcActions` or similar at the same context.

This change makes the app developers able to handle this problem (if necessary) without hard changes, until we work in a more robust solution (a singleton instance of `focus trap https://github.com/focus-trap/focus-trap/issues/428)

I've already prepared a PR for server

```diff
diff --git a/apps/files_sharing/src/views/SharingTab.provider.js b/apps/files_sharing/src/views/SharingTab.provider.js
new file mode 100644
index 00000000000..38fe212e57b
--- /dev/null
+++ b/apps/files_sharing/src/views/SharingTab.provider.js
@@ -0,0 +1,12 @@
+export const SHARING_TAB_ID = Symbol('menu::id')
+
+export const useSharingTabIDMixin = {
+	inject: {
+		$sharingTabID: { from: SHARING_TAB_ID, default: null },
+	},
+	computed: {
+		sharingTabIDSelector() {
+			return `#${this.$sharingTabID}`
+		},
+	},
+}
diff --git a/apps/files_sharing/src/views/SharingTab.vue b/apps/files_sharing/src/views/SharingTab.vue
index d1416a6c0d1..a5eea691ee4 100644
--- a/apps/files_sharing/src/views/SharingTab.vue
+++ b/apps/files_sharing/src/views/SharingTab.vue
@@ -21,7 +21,7 @@
   -->
 
 <template>
-	<div :class="{ 'icon-loading': loading }">
+	<div :id="randomID" :class="{ 'icon-loading': loading }">
 		<!-- error message -->
 		<div v-if="error" class="emptycontent" :class="{ emptyContentWithSections: sections.length > 0 }">
 			<div class="icon icon-error" />
@@ -89,6 +89,7 @@
 import { CollectionList } from 'nextcloud-vue-collections'
 import { generateOcsUrl } from '@nextcloud/router'
 import NcAvatar from '@nextcloud/vue/dist/Components/NcAvatar'
+import { NC_KEY_POPOVER_CONTAINER } from '@nextcloud/vue/dist/providers.js'
 import axios from '@nextcloud/axios'
 import { loadState } from '@nextcloud/initial-state'
 
@@ -120,8 +121,16 @@ export default {
 
 	mixins: [ShareTypes],
 
+	provide() {
+		return {
+			[NC_KEY_POPOVER_CONTAINER]: `#${this.randomID}`
+		}
+	},
+
 	data() {
 		return {
+			randomID: `SharingTab-${(Math.ceil((Math.random() * 900000) + 100)).toString(16)}`,
+
 			config: new Config(),
 
 			error: '',

```

raimund-schluessler:
I see, but I am still a bit puzzled, since they had a common container already, it was just not a provided div, but simply `body`, no?. So what I don't get is, why switching away from `body` solves this.

max-nextcloud:
> I see, but I am still a bit puzzled, since they had a common container already, it was just not a provided div, but simply `body`, no?. So what I don't get is, why switching away from `body` solves this.

My understanding is that if you have two focus traps sitting next to each other (for example as siblings) in the DOM they will compete for the focus. The focus can only be in one of the traps.
But if one focus trap is the parent of the other the focus can be in the child and therefore in both focus traps - so they are 'happy'.

If dropdown menus are attached to the body they are outside of any possible other focus trap (in particular the modal) and we have the competing situation. Attaching them to an element inside the current focus trap will make sure the focus can be in both traps at the same time.

So the problem could be solved by changing the container of all popovers inside a modal. However this can be tedious if there are multiple menus. This commit allows changing the default container for all popovers in the modal at once.

raimund-schluessler:
Ah, ok, I think now I got it. Thanks. I somehow was under the impression that it's about multiple `NcActions` fighting about the focus, for which it wouldn't help to give them a common container. But it's in fact about a `NcModal` and multiple `NcActions` within and giving the actions a container within the modal solves the issue.

So this PR just adds a convenience function for an app-specific solution that would be possible already (e.g. by touching every `NcActions` in a modal), right?

That's fine with me. But since it still requires devs to adjust their code, a note about this in the documentation for every component using the focus-trap (i.e. `NcActions`, `NcModal`, etc.) would be great.

raimund-schluessler (approved):

ChristophWurst (approved):
:+1: :elephant:

skjnldsv (changes requested):
Sorry, I think this whole approach is far too messy and complex. Maintaining this in the long run will be too cumbersome.

If multiple traps are fighting between each others, we should fix the issue at the root cause.

Here is my suggestion:
1. We create a service that handle all the registered focus traps
2. When a new focus trap is registered and activated, it pauses all the others. Example given:
    1. You open a Settings modal (like talk or mail)
    4. You open an Actions menu inside this modal
    5. Those two items will have thei rown focus traps
    6. The one that should have the priority should be the last activated, as they are (presumably) always initiated by a user action
    7. When the Actions menu gets closed, the previous trap can resume its activity
8. This should maybe require us to create a `@nextcloud/focus-trap`, where the first to register one initialize the manager.
  Storing all data in `window._nc_focus_trap_service` or something. The same way we do with the event bus. This is how we properly share a service across various independent services/apps.

PR #3323, opened 2022-09-30
Add undo trailing icon

Pytal:
Add an `"undo"` icon in addition to the `"close"` and `"arrowRight"` icons

### DIFF
--- a/src/components/NcTextField/NcTextField.vue
+++ b/src/components/NcTextField/NcTextField.vue
@@ -131,6 +131,7 @@ export default {
 		<template v-if="type !== 'search'" #trailing-button-icon>
 			<Close v-if="trailingButtonIcon === 'close'" :size="20" />
 			<ArrowRight v-else-if="trailingButtonIcon === 'arrowRight'" :size="20" />
+			<Undo v-else-if="trailingButtonIcon === 'undo'" :size="20" />
 		</template>
 	</NcInputField>
 </template>
@@ -141,6 +142,7 @@ import NcInputField from '../NcInputField/NcInputField.vue'
 
 import Close from 'vue-material-design-icons/Close.vue'
 import ArrowRight from 'vue-material-design-icons/ArrowRight.vue'
+import Undo from 'vue-material-design-icons/UndoVariant.vue'
 
 import { t } from '../../l10n.js'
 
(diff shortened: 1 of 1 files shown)


Pytal:
For some reason this affects the NcAppSidebar cypress tests

artonge:
Can we mark `trailingButtonIcon` as deprecated?

raimund-schluessler:
> For some reason this affects the NcAppSidebar cypress tests

No idea why, but the color of the favorite icon changed.
Before:
NcAppSidebar vue-subtitle_null-starred_true-compact_true-header_image-secondary_button-editable_false-base

After:
NcAppSidebar vue-subtitle_null-starred_true-compact_true-header_image-secondary_button-editable_false-actual

raimund-schluessler:
Could it be that a change in server now makes the cypress tests fail due to https://github.com/nextcloud/nextcloud-vue/pull/3139? The cypress tests for a PR that got merged yesterday without issues (https://github.com/nextcloud/nextcloud-vue/actions/runs/3159764008/jobs/5152100405) now also fails. I think it is less than ideal that external changes break the tests for this repository without interaction. There is no way of catching or debugging this.

@skjnldsv @marcoambrosini

raimund-schluessler:
I created a separate issue for this https://github.com/nextcloud/nextcloud-vue/issues/3325

marcoambrosini:
@Pytal We deliberately limited the functionality there to control the design a bit more. What trailing icon do you need there?

Pytal:
> Can we mark `trailingButtonIcon` as deprecated so we can remove it in a few major bump?

Did not mention deprecating anything?

Pytal:
> @Pytal We deliberately limited the functionality there to control the design a bit more. What trailing icon do you need there?

Would need the `undo-variant` icon from https://materialdesignicons.com, and had thought of adding another string `undo` to use this icon but thought that allowing any icon would prevent having to update the component each time a new icon is needed in the future

Pytal:
Should we add the `undo-variant` icon case or allow any custom icon @marcoambrosini?

marcoambrosini:
Yes I think we should add only `undo-variant` and proceed slowly keep things under control. cc @jancborchardt @nimishavijay

Pytal:
Review-ready

szaimen (approved):

marcoambrosini (approved):

PR #3324, opened 2022-10-01
Add advanced fields to color picker

Pytal:
Before | After
--- | ---
image | image

<details>
<summary>RGB and HSL</summary>

RGB
image

HSL
image

</details>

### DIFF
--- a/src/components/NcColorPicker/NcColorPicker.vue
+++ b/src/components/NcColorPicker/NcColorPicker.vue
@@ -106,6 +106,40 @@ export default {
 </style>
 ```
 
+* Using advanced fields including HEX, RGB, and HSL:
+
+```vue
+<template>
+	<div class="container0">
+		<NcColorPicker v-model="color"
+			:advanced-fields="true">
+			<NcButton> Click Me </NcButton>
+		</NcColorPicker>
+		<div :style="{'background-color': color}" class="color0" />
+	</div>
+</template>
+<script>
+export default {
+	data() {
+		return {
+			color: '#0082c9'
+		}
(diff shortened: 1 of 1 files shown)


szaimen (approved):
width looks strange but otherwise good:
image

marcoambrosini (approved):

Pytal:
Cypress failures unrelated, see https://github.com/nextcloud/nextcloud-vue/issues/3325

ChristophWurst:
FYI as seen in https://github.com/nextcloud/calendar/issues/5013 this change is not backwards compatible for apps that used the old picker

raimund-schluessler:
> FYI as seen in [nextcloud/calendar#5013](https://github.com/nextcloud/calendar/issues/5013) this change is not backwards compatible for apps that used the old picker

I am a bit puzzled. Before this change it wasn't possible to show the advanced fields at all. `:disable-fields="true"` did not offer to show them. Are you sure they were visible?

Pytal:
> I am a bit puzzled. Before this change it wasn't possible to show the advanced fields at all. `:disable-fields="true"` did not offer to show them. Are you sure they were visible?

Precisely, the fields were never visible before and this added them

Seems that the native color picker was used before https://github.com/nextcloud/calendar/commit/e7c489700bcca6d8bacab17585e4437ab7a67afa#diff-cb41e24846d5e15e56c4f885bb5c83fcf872d22c308ef7a529409334bef6c891L81-L88 @ChristophWurst

ChristophWurst:
My bad

PR #3335, opened 2022-10-03
Support inputmode for input fields

### LINKED ISSUE #3246: [NcInputField] Support inputmode
For mobile devices `inputmode` is very helpful to select the right keyboard.

Suggestion: 

```html
	<input :inputmode="inputmode" />
```

```js
	props: {
		inputmode: {
			type: String,
			default: null,
			validator(value) {
				return ['text', 'none', 'numeric', 'email', 'url'].includes(value)
			},
		},
	},

```

Ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode

dartcafe:
Resolves #3246

### DIFF
--- a/src/components/NcActionInput/NcActionInput.vue
+++ b/src/components/NcActionInput/NcActionInput.vue
@@ -87,6 +87,7 @@ For the multiselect component, all events will be passed through. Please see the
 					<input :id="id" type="submit" class="action-input__submit">
 
 					<input :type="type"
+						:inputmode="inputmode"
 						:value="value"
 						:placeholder="text"
 						:disabled="disabled"
@@ -143,6 +144,16 @@ export default {
 		icon: {
 			type: String,
 			default: '',
+		},
+		/**
+		 * inputmode attribute of the input field
+		 */
+		inputmode: {
+			type: String,
+			default: 'text',
+			validator(value) {
+				return ['text', 'none', 'numeric', 'email', 'url'].includes(value)
+			},
 		},
 		/**
 		 * type attribute of the input field
--- a/src/components/NcInputField/NcInputField.vue
+++ b/src/components/NcInputField/NcInputField.vue
@@ -33,6 +33,7 @@
 				ref="input"
 				class="input-field__input"
 				:type="type"
+				:inputmode="inputmode"
 				:disabled="disabled"
 				:placeholder="computedPlaceholder"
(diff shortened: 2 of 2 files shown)


marcoambrosini (changes requested):
I think we should keep using the `type` attribute for this kind of things as suggested by MDN documentation.

dartcafe:
`inputmode` helps to choose the appropiate virtual keyboard, when i.e. `type='text'` is to use, but also special characters like spaces or hyphens are allowed. This is very helpful on mobile devices.

A common use case are formatted numeric inputs like credit card numbers, IBAN, article numbers or similair.

It is an addition to the type `attribute` and does no harm. The default is null, so it should not be set.

marcoambrosini:
For that use case I think we should use something like `<TextField type="number" />` to be consistent with what we're already doing. It's not implemented yet but it should be enough to allow the `number` string in the `type` prop validator. 

https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/NcTextField/NcTextField.vue#L169

dartcafe:
> It's not implemented yet but it should be enough to allow the `number` string in the `type` prop validator.

Not sure, if I get you right. Since '2345 3765 7654 7695' is not a number, the type 'number' is wrong, following the specification of https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number.

> \<input\> elements of type number are used to let the user enter a number. They include built-in validation to reject non-numerical entries.

marcoambrosini:
Sorry for the late reply @dartcafe. Is that the only use-case that is not contemplated by the `type` attribute? If that is the case I would just accept `numeric`. For `email` and `url` we can just use `type`, right? 
My concern here is not making these components too complicated and limit the usage patterns. For example I would like to avoid having:
`<TextField inputmode="email" />`
in favor of:
`<TextField type="email" />`

dartcafe:
> Is that the only use-case that is not contemplated by the `type` attribute?

That is one use case, which helped me to get a better UX with article numbers and masked fields, where type=number was not possible.

Some other use cases you can find in [this article](https://css-tricks.com/everything-you-ever-wanted-to-know-about-inputmode/) and [this](https://css-tricks.com/better-form-inputs-for-better-mobile-user-experiences/). Maybe that helps to get an overview.

The point is, that inputmode can define a virtual keyboart explicitly, while type defines it implicitly. I guess there is a reason, why all mobile browsers support this living standard.

dartcafe:
closing in favour of #3485

PR #3336, opened 2022-10-04
Add minlength and maxlength to text field

Pytal:
Add `minlength` and `maxlength` to be forwarded to the input by `v-bind="$props"`

The part which adds the `autofocus` prop definition is an addendum to https://github.com/nextcloud/nextcloud-vue/pull/3245

Before this change `autofocus` wasn't contained in `$props` so it wasn't automatically forwarded by `v-bind="$props"`, adding the explicit definition now makes it get forwarded by the `v-bind` and also makes the prop show up in the docs

### DIFF
--- a/src/components/NcPasswordField/NcPasswordField.vue
+++ b/src/components/NcPasswordField/NcPasswordField.vue
@@ -95,7 +95,6 @@ export default {
 <template>
 	<NcInputField v-bind="$props"
 		ref="inputField"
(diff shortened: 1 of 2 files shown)


kesselb (approved):
LGTM :+1:

artonge (approved):

blizzz (approved):
:elephant:

Issue #3332, opened 2022-05-30
Focus search field when opening emoji picker


come-nc:
**Is your feature request related to a problem? Please describe.**
When clicking the emoji picker, I have to click the search field before I can start typing to find the right emoji. This is always frustrating.

**Describe the solution you'd like**
Focus the search field so I can type right away, as a lot of other applications do it (reference needed :thinking: ).

PR #3343, opened 2022-10-04
Fix emoji picker and Popover no-Auto-Focus

skjnldsv:
See https://github.com/Akryum/floating-vue/pull/894

Because the Popover always focuses after the event is emitted, we need to make sure to wait after the render (nextTick) to properly focus on another element.
Also implementing a way around the library to mitigate the focus priority until the new PR above is merged

fix https://github.com/nextcloud/nextcloud-vue/issues/3332


PVince81 (approved):
:+1:

won't be the first time

szaimen (approved):
works

PR #3367, opened 2022-10-15
Fix outdated variables

jancborchardt:
**Replace deprecated --color-text-lighter with --color-text-maxcontrast** & **Replace deprecated --color-text-light with --color-main-text**
We should never ever use `--color-text-lighter` (or "light") anymore.
- `color-text-light` → `color-main-text`
- `color-text-lighter` → `color-text-maxcontrast`

Could we fully deprecate the old variables?

---

**Progressive enhancement for outdated --color-primary-element-lighter**
The --lighter does not exist in new versions anymore, so currently a button with `type=""` (empty, as opposed to no type property or `type="secondary"` actually does not show any background. This fixes that.

### DIFF
--- a/src/components/NcActionInput/NcActionInput.vue
+++ b/src/components/NcActionInput/NcActionInput.vue
@@ -404,7 +404,7 @@ $input-margin: 4px;
 		padding: 7px 6px;
 
 		opacity: $opacity_full;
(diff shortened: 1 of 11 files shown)


marcoambrosini (approved):
🐘

skjnldsv (approved):

PR #3369, opened 2022-10-17
Add guest content component

ChristophWurst:
For pages like

Bildschirmfoto vom 2022-10-17 10-42-20

This could be used for login, login flow, 2FA, OAuth redirect pages, etc.

In a future version I could also image that there is a "fullscreen" or size prop to make the container expand to the available width/height.


JuliaKirschenheuter:
> In a future version I could also image that there is a "fullscreen" or size prop to make the container expand to the available width/height.

Sounds good!

JuliaKirschenheuter:
Why are we need a new component https://github.com/nextcloud/nextcloud-vue/blob/4fc02b3b13df82e9a6d49713dd2055c960f4e8a4/src/components/NcGuestContent/NcGuestContent.vue for this issue?

ChristophWurst:
Because there is no such component at the moment.

ChristophWurst:
Acute need for the component is triggered by https://github.com/nextcloud/mail/pull/7430.

GretaD (approved):
Code looks fine, i didnt test it with an app

st3iny (changes requested):
I like the idea!

However, it doesn't work out of the box with components inside our regular content container. Have a look at an example from Calendar:

### At the top
image

### Scrolled a bit down
image

IMO, `NcGuestContent` needs at least `margin: 0 auto` to center it and maybe `height: fit-content` to make it grow.

**EDIT:** The outer content container might also need changes to allow scrolling and prevent clipping at the bottom (taken from `calendar/src/views/Appointments/Booking.vue`):
```scss
#content.app-calendar {
	// Enable scrolling
	overflow: auto;

	// Fix box being cutoff at the bottom
	margin-bottom: 0;
	height: calc(var(--body-height) + var(--body-container-margin));
}
```

ChristophWurst:
> MO, `NcGuestContent` needs at least `margin: 0 auto` to center it and maybe `height: fit-content` to make it grow.

While you already have your fancy setup to check this :smiling_imp:, could you push those styles to my branch?

st3iny (approved):
## Screenshots from Calendar appointments feature

[scroll-guest-content.webm](https://user-images.githubusercontent.com/1479486/196402382-98b0c2a3-d547-48aa-aa30-c711098915ea.webm)

Screenshot 2022-10-18 at 12-10-38 Nextcloud

src/components/NcGuestContent/NcGuestContent.vue
```diff
@@ -0,0 +1,77 @@
...
+	min-width: 0;
+	border-radius: var(--border-radius-large);
+	box-shadow: 0 0 10px var(--color-box-shadow);
+	height: fit-content;
+	padding: 15px;
+	margin: 20px auto;
+}
+</style>
+
+<style lang="scss">
```
nickvergessen, line 68: This is not scoped and therefore breaks files app, settings, basically any page layout by missing the bottom border:
grafik

grafik

PR #3379, opened 2022-10-19
Bump version to 7.1.0

ChristophWurst:
The last major was v7.0.0 and I see there is a `stable7`, so I assume `master` changes are going to be released as either 7.1 or 8.0 depending on whether we have breaking changes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "@nextcloud/vue",
-	"version": "7.0.1",
+	"version": "7.1.0",
 	"lockfileVersion": 2,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "@nextcloud/vue",
-			"version": "7.0.1",
+			"version": "7.1.0",
 			"license": "AGPL-3.0",
 			"dependencies": {
 				"@nextcloud/auth": "^2.0.0",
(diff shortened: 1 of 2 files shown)


mejo-:
> Who created the stable7 branch?

That was me, after discussion in the fontenders room. But I'm a naive novice :grimacing: So I'm totally fine with removing the branch and continuing 7.0 development in master.

skjnldsv:
Well, no worries, just wondering if there is a underlying motive  :stuck_out_tongue_closed_eyes: 

We never did that in this repo. I am much more in favour of keeping this one as 7.x.x until 8.x.x as usual.
Otherwise it's just duplicating all the mess with backports etc etc

ChristophWurst:
Okay but while we are here, what if we bump to 7.1 and release this shiny new component https://github.com/nextcloud/nextcloud-vue/pull/3369?

mejo-:
> I am much more in favour of keeping this one as 7.x.x until 8.x.x as usual.
> Otherwise it's just duplicating all the mess with backports etc etc

Yep, let's do that. Unfortunately I lack the permissions to delete the protected branch `stable7`, so could someone with super powers do that?

ChristophWurst:
Okay then we don't release the component.

skjnldsv:
@ChristophWurst we will, we just never preemptively bumped versions on master before releasing.
Anyway, 7.1.0 it is! :rocket:

PR #3401, opened 2022-10-28
Implement cron job to synchronise server styling

### LINKED ISSUE #3325: Cypress tests broken due external changes to server repo
Probably due to #3139 the visual regression tests now get the server CSS live from the server repo. This randomly breaks the visual regression tests here.

Imho this is not a good idea, since  then unrelated changes randomly make the cypress tests fails and can't get merged, see e.g. #3323. Furthermore, we then don't know if it was a change here, or in the server that makes the tests fail.

@skjnldsv @CarlSchwan @vinicius73

skjnldsv:
Fix #3325 

Every night, we'll have a server sync and a PR will be created if any changes


raimund-schluessler:
That sounds good. But I think styleguide/assets/additional.css needs to be adjusted as well.

skjnldsv:
> That sounds good. But I think styleguide/assets/additional.css needs to be adjusted as well.

Done, completely forgot to change it :see_no_evil:

skjnldsv:
I purposefully shipped outdated css in this PR, so we can fire the cron job right after merging and see if everything works fine :)

szaimen (approved):
🐘

skjnldsv:
Fixed imports.

skjnldsv:
Ok, let's merge and check the workflow

PR #3404, opened 2022-10-29
Allow querying for resources with complex item ids

Pytal:
Query for resources with the item id as a query string parameter instead of a URL path segment to allow for complex ids

For https://github.com/nextcloud/related_resources/pull/124


PVince81 (approved):
:+1: hope it won't break backward compat though

Pytal:
> 👍 hope it won't break backward compat though

`related_resources` v1 major for `stable25` will keep the `/related/{providerId}/{itemId}` route, it is only dropped for NC26

artonge (approved):

PR #3408, opened 2022-10-31
Fix password_policy fallback handling

nickvergessen:
The defined fallback is only affecting when no core=>capabilities is there. But that is pretty unlikely while the password_policy entry can be missing often (app not installed, unit tests, ...)

Also null vs. undefined gives different results:
https://github.com/nextcloud/nextcloud-vue/blob/0204f6dd86ad16e9b57d8c5915ee15672d4a5e5a/src/components/NcPasswordField/NcPasswordField.vue#L354-L359

Fixes the failing JS unit test in: https://github.com/nextcloud/spreed/pull/8213

### DIFF
--- a/src/components/NcPasswordField/NcPasswordField.vue
+++ b/src/components/NcPasswordField/NcPasswordField.vue
@@ -127,7 +127,7 @@ import { generateOcsUrl } from '@nextcloud/router'
 import { t } from '../../l10n.js'
 import logger from '../../utils/logger.js'
 
-const defaultPasswordPolicy = loadState('core', 'capabilities', { passwordPolicy: null }).password_policy
+const defaultPasswordPolicy = loadState('core', 'capabilities', {}).password_policy || null
 
 export default {
 	name: 'NcPasswordField',


danxuliu (approved):
Not tested, but code looks good.

Should the handling of `null` and `undefined` be unified?

skjnldsv (approved):

src/components/NcPasswordField/NcPasswordField.vue
```diff
@@ -127,7 +127,7 @@ import { generateOcsUrl } from '@nextcloud/router'
 import { t } from '../../l10n.js'
 import logger from '../../utils/logger.js'
 
-const defaultPasswordPolicy = loadState('core', 'capabilities', { passwordPolicy: null }).password_policy
+const defaultPasswordPolicy = loadState('core', 'capabilities', { password_policy: null })?.password_policy || null
```
skjnldsv, line 130:
```suggestion
const defaultPasswordPolicy = loadState('core', 'capabilities', {}).password_policy || null
```

Should even be enough I think

nickvergessen: The question mark was needed for talks test, but maybe I messed up the compiling order between vue and talk when testing. But can retry tomorrow

nickvergessen: Seems you are right. the `|| null` basically seems to kill the error already.

nickvergessen:
> Should the handling of null and undefined be unified?

With the current approach they are.

src/components/NcPasswordField/NcPasswordField.vue
skjnldsv: It would fail if someone provided an initial state that is not an object. But I don't see why that would happen anyway :)

nickvergessen:
`(╯°□°)╯︵ ┻━┻`

PR #3410, opened 2022-10-31
Expand color picker width only when using advanced fields

Pytal:
Before | After
--- | ---
image | image

Advanced fields
image


mejo- (approved):
Looks sensible to me :ok_hand:

Failed cypress tests seem unrelated, given that the very same tests already fail on latest master.

szaimen (approved):

PR #3411, opened 2022-10-31
Fix dark themed color picker and use button component

Pytal:
Before | After
--- | ---
image | image
image | image


nickvergessen (approved):
Didn't test, but looks good

artonge (approved):

PR #3445, opened 2022-11-08
fix the size of the popper element

szaimen:
Fix https://github.com/nextcloud/server/issues/33839

| Before | After |
|---|---|
| image | image |

Signed-off-by: Simon L <szaimen@e.mail.de>

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -1179,7 +1179,7 @@ export default {
 .v-popper--theme-dropdown.v-popper__popper.action-item__popper .v-popper__inner {
 	border-radius: var(--border-radius-large);
 	padding: 4px;
-	max-height: calc(100vh - 16px);
+	max-height: calc(50vh - 16px);
 	overflow: auto;
 }
 </style>


juliushaertl:
I think the 50vh approach might also not work well if the header is larger (e.g. with a preview). Maybe we can change the popper positioning strategy so that the popover is not limited to the tab area but may also span over the full window height and be positioned to the left of the button?

juliushaertl:
> Is there no way for the popper element to automatically detect out-of-bound window ?

Might not work in a generic way, so maybe that is something that needs rather some fine tuning in the sharing sidebar implementation.

szaimen:
> I think the 50vh approach might also not work well if the header is larger (e.g. with a preview)

No, that is no problem as it will overlap the preview header then. Let me test this out as well...

szaimen:
@skjnldsv @juliushaertl both edge cases seem to work:
image

szaimen:
@skjnldsv @juliushaertl can you approve? :)

Pytal (approved):

szaimen:
One additional review needed :)

PVince81 (approved):
:+1: nice

Issue #3386, opened 2022-10-21
[7.0.1] "too much recursion" when opening NcModal and NcAppSettingsDialog


dartcafe:
When opening NcModal, it opens slowly and the following errors are thrown in  the JS console:

Test it with https://nextcloud-vue-components.netlify.app/#/Components/NcModal

![grafik](https://user-images.githubusercontent.com/26707476/197141941-31737efe-4e16-443d-a888-61dd09fdd773.png)

ChristophWurst:
Can confirm when I link `master` against Mail and open a settings modal

![Bildschirmfoto vom 2022-10-21 15-53-30](https://user-images.githubusercontent.com/1374172/197212510-2b17850d-d722-4305-85b8-60824a8fcd9d.png)

:eyes: @vinicius73

ChristophWurst:
Tracked in https://github.com/nextcloud/nextcloud-focus-trap/issues/2

dartcafe:
🎉

PR #3478, opened 2022-11-15
Define global focus trap

skjnldsv:
https://github.com/focus-trap/focus-trap/pull/826/

- Fix https://github.com/nextcloud/server/issues/35056
- Fix https://github.com/nextcloud/nextcloud-vue/issues/3386


szaimen (approved):
LGTM but didnt test

src/utils/focusTrap.js
```diff
@@ -0,0 +1,35 @@
...
+import { FocusTrap } from 'focus-trap'
+
+/**
+ * Return the default global focus trap stack
+ *
+ * @return {FocusTrap[]}
+ */
+export const getTrapStack = function() {
+	// Create global stack if undefined
+	Object.assign(window, { _nc_focus_trap: window._nc_focus_trap || [] })
```
artonge, line 32: Does this have to be done by every app ?
Wouldn't it make sense to put the global focus trap logic in `@nextcloud/focus-trap` ? This would ensure that no app forgets to use the global focus trap.

artonge: Looking at `NcPopover` and `NcModal` we already have it twice in this repo.

skjnldsv: > Wouldn't it make sense to put the global focus trap logic in `@nextcloud/focus-trap` ? This would ensure that no app forgets to use the global focus trap.

Is there any app manually using a focus trap?
If not I guess we're good with this in that library then.

Because we already have a hard time maintaining all those libraries. I don't want to add one more to my list :(

skjnldsv: Also, having an external lib to build and release and maintain deps just to have that wrapper is quite overkill :(

skjnldsv: I would delete @nextcloud/focus-trap once we have this in and working tbh

juliushaertl:
Did a quick test, but I still have an endless recursion on text with the session list popover

juliushaertl:
Ah wait I also need to link/build viewer i guess

juliushaertl (approved):
Tested and works with text/viewer

src/utils/focusTrap.js
artonge: Make sense :+1:

PR #3489, opened 2022-11-17
Add NcHeaderMenu

skjnldsv:
image
Since we're positioning it fixed on nextcloud the live example is not as great as this example, but it works out.

From https://github.com/nextcloud/server/blob/abaa3ef61b9a6177ca3fe965e837e6e9bca16a4b/core/src/views/UnifiedSearch.vue

### DIFF
--- a/src/components/NcHeaderMenu/NcHeaderMenu.vue
+++ b/src/components/NcHeaderMenu/NcHeaderMenu.vue
@@ -88,7 +88,7 @@ export default {
 			:id="`header-menu-${id}`"
 			class="header-menu__wrapper"
 			role="menu">
-			<div class="header-menu__content">
+			<div ref="content" class="header-menu__content">
 				<!-- @slot Main content -->
 				<slot />
 			</div>
@@ -98,7 +98,10 @@ export default {
 
 <script>
 import { directive as ClickOutside } from 'v-click-outside'
+import { createFocusTrap } from 'focus-trap'
+
 import excludeClickOutsideClasses from '../../mixins/excludeClickOutsideClasses/index.js'
+import { getTrapStack } from '../../utils/focusTrap.js'
 
 export default {
 	name: 'NcHeaderMenu',
@@ -146,24 +149,24 @@ export default {
 
 	data() {
 		return {
+			focusTrap: null,
 			opened: this.open,
+			shortcutsDisabled: window.OCP?.Accessibility?.disableKeyboardShortcuts?.(),
+
 			clickOutsideConfig: {
 				handler: this.closeMenu,
 				middleware: this.clickOutsideMiddleware,
 			},
-			shortcutsDisabled: window.OCP?.Accessibility?.disableKeyboardShortcuts?.(),
 		}
 	},
 
 	watch: {
 		open(newVal) {
 			this.opened = newVal
+			// Wait for component to finish rendering
 			this.$nextTick(() => {
-				if (this.opened) {
-					this.openMenu()
-				} else {
-					this.closeMenu()
-				}
+				// Trigger events
+				this.opened ? this.openMenu() : this.closeMenu()
 			})
 		},
 	},
@@ -199,6 +202,9 @@ export default {
 			this.opened = false
 			this.$emit('close')
 			this.$emit('update:open', false)
+
+			// Kill focus trap
+			this.clearFocusTrap()
 		},
 
 		/**
@@ -212,6 +218,17 @@ export default {
 			this.opened = true
 			this.$emit('open')
 			this.$emit('update:open', true)
+
+			/**
+			 * Triggered after the tooltip was visually displayed.
+			 *
+			 * This is different from the 'show' and 'apply-show' which
+			 * run earlier than this where there is no guarantee that the
+			 * tooltip is already visible and in the DOM.
+			 */
+			this.$nextTick(() => {
+				this.useFocusTrap()
+			})
 		},
 
 		onKeyDown(event) {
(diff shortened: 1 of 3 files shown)


src/components/NcHeaderMenu/NcHeaderMenu.vue
```diff
@@ -0,0 +1,315 @@
...
+	watch: {
+		open(newVal) {
+			this.opened = newVal
+			this.$nextTick(() => {
+				if (this.opened) {
+					this.openMenu()
+				} else {
+					this.closeMenu()
+				}
+			})
```
artonge, line 167: Why do we need to set `this.opened` when `openMenu` or `closeMenu` will do it again later ?
Would this work ?
```js
if (newVal) {
	this.openMenu()
} else {
	this.closeMenu()
}
```

src/components/NcHeaderMenu/NcHeaderMenu.vue
```diff
@@ -0,0 +1,315 @@
...
+  - along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+
+<docs>
+This component is made to be used in the Nextcloud top header.
+
+```
+<template>
+	<div id="nextcloud-header">
```
nickvergessen, line 28: This means the component can only be used exactly once.
Please check SettingsSections etc for automatic ID appending or better don't use and ID or require it to be set as a prop

nickvergessen (dismissed):
I was gifted a similar component in the notifications app by @CarlSchwan 
https://github.com/nextcloud/notifications/blob/master/src/Components/HeaderMenu.vue

Can we look that it can cover both cases?

src/components/NcHeaderMenu/NcHeaderMenu.vue
```diff
@@ -0,0 +1,315 @@
...
+		<NcHeaderMenu id="search"
+			aria-label="Search">
+			<template #trigger>
+				<Magnify />
+			</template>
+			<div>
+				<input placeholder="Search for files, comments, contacts..." type="search" style="width: 99%;" />
+				<NcEmptyContent
+					title="Search"
+					description="Start typing to search">
```
nickvergessen, line 38: very specific and not using translations?

src/components/NcHeaderMenu/NcHeaderMenu.vue
```diff
@@ -0,0 +1,315 @@
...
+```
+<template>
+	<div id="nextcloud-header">
+		<NcHeaderMenu id="search"
+			aria-label="Search">
+			<template #trigger>
+				<Magnify />
+			</template>
+			<div>
+				<input placeholder="Search for files, comments, contacts..." type="search" style="width: 99%;" />
```
nickvergessen, line 35: very specific and not using translations?

src/components/NcHeaderMenu/NcHeaderMenu.vue
```diff
@@ -0,0 +1,315 @@
...
+  -->
+
+<docs>
+This component is made to be used in the Nextcloud top header.
+
+```
+<template>
+	<div id="nextcloud-header">
+		<NcHeaderMenu id="search"
+			aria-label="Search">
```
nickvergessen, line 30: very specific and not using translations?

nickvergessen:
Sorry... I read the sample not the actual template...

skjnldsv:
> I was gifted a similar component in the notifications app by @CarlSchwan

Carl took mine from server :see_no_evil: 
This is actually the reason I am putting this one here :wink:

src/components/NcHeaderMenu/NcHeaderMenu.vue
skjnldsv: Because we keep a local state `opened`.
The prop is `open`. If we don't do that, it means the dev _have_ to provide a writeable prop (with `.sync` for example)

Like this it means the dev can manually trigger open/close from the outside or not provide it and let the component handle itself.

If the external `open` prop change, we trigger the same way as if the user would have opened or close the menu.
Changing the internal data variable `opened`

skjnldsv: I added comments, it's to properly fire the events too. But we make sure the component is updated before doing so

skjnldsv: https://github.com/nextcloud/server/blob/abaa3ef61b9a6177ca3fe965e837e6e9bca16a4b/core/src/views/UnifiedSearch.vue#L371

Here is an example. If we trigger too soon, we cannot focus the input.

artonge: Then `nextTick` needs to go in `openMenu`/`closeMenu` if we want to be consistent, no ?

Because here, I understand that the parent will receive the event before the next tick if it comes from here: https://github.com/nextcloud/nextcloud-vue/pull/3489/files/9c21762167bba03dd398cd3d5a0cc7a58b9e4f5d#diff-7af842de9d50674691572f5571b0849046baa1f66655a53bc2d1d6f51fd14ddcR77

src/components/NcHeaderMenu/NcHeaderMenu.vue
```diff
@@ -0,0 +1,359 @@
...
+			this.$emit('update:open', true)
+
+			/**
+			 * Triggered after the tooltip was visually displayed.
+			 *
+			 * This is different from the 'show' and 'apply-show' which
+			 * run earlier than this where there is no guarantee that the
+			 * tooltip is already visible and in the DOM.
+			 */
+			this.$nextTick(() => {
```
artonge, line 229: There is already one in `useFocusTrap`, is this one needed ?

src/components/NcHeaderMenu/NcHeaderMenu.vue
skjnldsv: Might be yeah :shrug:

skjnldsv: Do we want to properly do the `open` and `opened` differences in emitted events ?

src/components/NcHeaderMenu/NcHeaderMenu.vue
skjnldsv: Done

src/components/NcHeaderMenu/NcHeaderMenu.vue
skjnldsv: Done

Pytal (approved):
Apart from the minor eslint errors

skjnldsv:
> Apart from the minor eslint errors

Fixed!  :see_no_evil:

src/components/NcHeaderMenu/NcHeaderMenu.vue
```diff
@@ -0,0 +1,347 @@
...
+		 */
+		openMenu() {
+			// Open the menu
+			this.opened = true
+			this.$emit('open')
+			this.$emit('update:open', true)
+
+			// Wait for component to finish rendering
+			this.$nextTick(() => {
+				this.useFocusTrap()
```
artonge, line 224: `useFocusTrap` is still guarded twice by a `$nextTick` call

artonge (approved):
Minor nitpick, but good otherwise :)

Issue #3403, opened 2022-10-28
ActionInput date ignores given value when using strings


PVince81:
### Steps

1. Edit NcActionInput.vue
2. Search for the "Please pick a date"
3. Add `value="2022-10-31 00:00:00"` in the row
4. `npm run styleguide` and test it
5. Open the actions input

### Expected result
The date picker field uses the given value.

### Actual result
Always empty.

### Notes

I tried with other formats like "2022-10-31" and "10/31/2022" but the value is always ignored.

This is a regression as it used to work before, see https://github.com/nextcloud/server/issues/34695#issuecomment-1295236260

Tested with nextcloud-vue master and also v7.0.1.
Unsure if v6 is broken.

PVince81:
@Pytal can you investigate ?

PVince81:
still not working on server: https://github.com/nextcloud/server/issues/34926#issuecomment-1316649298

not sure what's different from the style guide

PVince81:
it's not working, tested with my steps above

maybe this got closed too early or as a mistake ?

PVince81:
need to be careful because it seems we're now using the NcDateTimePickerNative so not sure if it needs fixing too

PVince81:
after analysis we found that the native date picker only accepts `Date` objects and not `String`, so we might need to parse strings to keep compatible with the old NcDateTimePicker implementation

skjnldsv:
> I see that the example is using `v-model` instead of the `value` property. is it now required that everyone switches to `v-model` ?

v-model is like value. But with both get and set.
Also like `:value.sync=""` but it will listen to input or change depending on the html element.

PVince81:
so regarding "value" we saw that one needs to pass in a `Date` object, should we move forward and remove the `String` type from the implementation to make sure Vue will display warnings when used improperly ? @skjnldsv

skjnldsv:
Yes, removing should be enough to not mislead :)

PR #3491, opened 2022-11-17
Fix value type for DateTimePickerNative

skjnldsv:
Fix https://github.com/nextcloud/nextcloud-vue/issues/3403

### DIFF
--- a/src/components/NcDateTimePickerNative/NcDateTimePickerNative.vue
+++ b/src/components/NcDateTimePickerNative/NcDateTimePickerNative.vue
@@ -159,16 +159,17 @@ export default {
 
 	props: {
 		/**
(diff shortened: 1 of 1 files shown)


PVince81:
@szaimen it won't. In that ticket/PR we switched to `Date`. The reason it was broken was because it was wrongly set to use a `String`

PVince81 (approved):
:+1:

artonge (commented):
Need to update the jsDoc of this method: https://github.com/nextcloud/nextcloud-vue/blob/ba7c3a42c5133d495a3a80603874aa6289d2c70a/src/components/NcDateTimePickerNative/NcDateTimePickerNative.vue#L333

PR #3516, opened 2022-11-24
Add trap stack util to public API

Pytal:
For use in server and apps


GretaD (approved):

skjnldsv:
Can you describe a location we would use that in server? or app?
Assuming all the elements requiring a focus trap are vue components, doing this should not be necessary

skjnldsv (changes requested):
Can you point or explain a context where we would use that in server or an app?
Assuming all the elements requiring a focus trap are vue components, doing this should not be necessary

Pytal:
> Can you point or explain a context where we would use that in server or an app? Assuming all the elements requiring a focus trap are vue components, doing this should not be necessary

Could be used in https://github.com/nextcloud/server/pull/35381 as the user and contacts menus are not in Vue

skjnldsv:
I would migrate them to vue and use the `HeaderMenu` then, no?

Pytal:
> I would migrate them to vue and use the `HeaderMenu` then, no?

The menus have been using the legacy js implementation for awhile so if the component is flexible enough then sounds good!

Would then need https://github.com/nextcloud/nextcloud-vue/pull/3489 first

skjnldsv:
There was an attempt: 
- https://github.com/nextcloud/server/pull/22274

Pytal:
Not needed as menus will use `NcHeaderMenu`

PR #3517, opened 2022-11-25
Allow accessible labels for NcActionInput

Pytal:
Brought over from NcInputField and switched `labelVisible` default to `true`

https://github.com/nextcloud/nextcloud-vue/blob/1dc8ed09c47d482a7a9cf8ae89babbd15f209aee/src/components/NcInputField/NcInputField.vue#L147-L155
https://github.com/nextcloud/nextcloud-vue/blob/1dc8ed09c47d482a7a9cf8ae89babbd15f209aee/src/components/NcInputField/NcInputField.vue#L167-L175
https://github.com/nextcloud/nextcloud-vue/blob/1dc8ed09c47d482a7a9cf8ae89babbd15f209aee/src/components/NcInputField/NcInputField.vue#L370-L382

### Example

image

### DIFF
--- a/src/components/NcActionInput/NcActionInput.vue
+++ b/src/components/NcActionInput/NcActionInput.vue
@@ -29,6 +29,7 @@ For the multiselect component, all events will be passed through. Please see the
 <NcActions>
 	<NcActionInput icon="icon-edit" value="This is an input" />
 	<NcActionInput icon="icon-edit">This is the placeholder</NcActionInput>
+	<NcActionInput icon="icon-edit" label="Visible label">Input with visible label</NcActionInput>
 	<NcActionInput icon="icon-close" :disabled="true" value="This is a disabled input" />
 	<NcActionInput icon="icon-edit" type="date">Please pick a date</NcActionInput>
 	<NcActionInput icon="icon-edit" type="multiselect" :options="['Apple', 'Banana', 'Cherry']">Please pick a fruit</NcActionInput>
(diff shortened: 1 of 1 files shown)


src/components/NcActionInput/NcActionInput.vue
```diff
@@ -171,6 +190,22 @@ export default {
...
+		 * The hidden input label for accessibility purposes.
+		 */
+		label: {
+			type: String,
+			default: undefined,
+		},
+		/**
+		 * We normally have the label hidden visually and use it for
+		 * accessibility only. If you want to have the label visible just above
+		 * the input field pass in true to this prop.
```
PVince81, line 203: from my understanding, I thought we always wanted the labels visible

this is also why we have made them visible

so defaulting to false would be just for compatibility?

Pytal: Default was decided in https://github.com/nextcloud/nextcloud-vue/pull/2868, perhaps @marcoambrosini could detail the reasoning behind this?

Login page explicitly opts-in to using the visible label https://github.com/nextcloud/server/blob/be0facaba478235ee5f0608e36bf806570298a18/core/src/components/login/LoginForm.vue#L59-L73

src/components/NcActionInput/NcActionInput.vue
```diff
@@ -85,22 +86,32 @@ For the multiselect component, all events will be passed through. Please see the
...
-						:class="{ focusable: isFocusable }"
-						class="action-input__input"
-						@input="onInput"
-						@change="onChange">
-					<!-- allow the custom font to inject a ::before
-						not possible on input[type=submit] -->
-					<label v-show="!disabled" :for="id" class="action-input__label">
-						<ArrowRight :size="20" />
-					</label>
+					<div class="action-input__input-label-container">
```
JuliaKirschenheuter, line 89: Should we probably create a more-understandable html structure, because now label ``` action-input__text-label ``` belongs to input ``` action-input__input-container ``` which is inside of ``` action-input__input-container ```. 

And input ``` action-input__submit``` belongs to label ``` action-input__label ``` which is inside of ```action-input__input-label-container``` and  ```action-input__input-label-container```. 

A bit confusing i think

JuliaKirschenheuter (changes requested):

src/components/NcActionInput/NcActionInput.vue
JuliaKirschenheuter: *if it works we can leave it like this for a day we have more time

Pytal: Good point @JuliaKirschenheuter! A refactor of the HTML structure here would be out-of-scope so we can do this in a separate PR

JuliaKirschenheuter (approved):

src/components/NcActionInput/NcActionInput.vue
marcoambrosini: I think that's what I first did but I then reverted for backwards compatibility yes after discussing it. It's totally fine for me to have it visible by default

Pytal: Defaulted `labelVisible` to `true`, `NcInputField` would have to stay hidden by default for v7 at least

src/components/NcActionInput/NcActionInput.vue
```diff
@@ -84,23 +85,33 @@ For the multiselect component, all events will be passed through. Please see the
...
-						class="action-input__input"
-						@input="onInput"
-						@change="onChange">
-					<!-- allow the custom font to inject a ::before
-						not possible on input[type=submit] -->
-					<label v-show="!disabled" :for="id" class="action-input__label">
-						<ArrowRight :size="20" />
-					</label>
+					<div class="action-input__container">
+						<label v-if="label !== undefined"
```
skjnldsv, line 89:
```suggestion
						<label v-if="label"
```
Otherwise would be displayed with `null`, or empty strings

skjnldsv (approved):
Minor comment

PR #3530, opened 2022-12-01
Fix focus returning to top of page instead of actions toggle button

Pytal:
Fix focused element when tabbing after reaching the last actions menu item

### Before

https://nextcloud-vue-components.netlify.app/#/Components/NcActions?id=ncactions-1 (Scroll down to "Multiple actions")

Focus returns to the top of the page

### After

https://deploy-preview-3530--nextcloud-vue-components.netlify.app/#/Components/NcActions?id=ncactions-1 (Scroll down to "Multiple actions")

Focus returns to the actions menu toggle button


PVince81 (approved):
:+1:

skjnldsv (approved):

PR #3532, opened 2022-12-01
Add button role for NcButton with href

Pytal:
Attribute `role="button"` added for anchor element in NcButton with `href` and fixed the invalid `type="button"` attr being added to the DOM when using NcButton as a link https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-type

`router-link` is rendered as a button element https://v3.router.vuejs.org/api/#tag

### DIFF
--- a/src/components/NcButton/NcButton.vue
+++ b/src/components/NcButton/NcButton.vue
@@ -202,7 +202,6 @@ button {
 		v-bind="rootElement"
 		:class="buttonClassObject"
 		:aria-label="ariaLabel"
-		:type="nativeType"
 		:disabled="disabled"
 		v-on="$listeners">
 		<span class="button-vue__wrapper">
@@ -321,6 +320,7 @@ export default {
 				return {
 					is: 'router-link',
 					tag: 'button',
+					type: this.nativeType,
 					to: this.to,
 					exact: this.exact,
 					...this.$attrs,
@@ -335,6 +335,7 @@ export default {
 			}
 			return {
 				is: 'button',
+				type: this.nativeType,
 				...this.$attrs,
 			}
 		},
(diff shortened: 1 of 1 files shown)


JuliaKirschenheuter (approved):

raimund-schluessler:
@Pytal Are you sure, this is necessary? Looking at https://github.com/nextcloud/server/pull/35527 I suppose this is about passing `href` to the `NcButton`. But that should work already by passing the `propsData` here:
https://github.com/nextcloud/nextcloud-vue/blob/4e8546981baa32b3c1ec34825fa3a7f02db182d7/src/components/NcActions/NcActions.vue#L949
At least in the docs, a single `NcActionLink` with `href` set correctly becomes a `NcButton` of type link with the `href` present.

Pytal:
This is for the attribute `role="button"` from [this comment](https://github.com/nextcloud/server/pull/35527#discussion_r1036811493) as it is not included in `propsData` @raimund-schluessler, open to ideas if another way is better?

raimund-schluessler:
> This is for the attribute `role="button"` from [this comment](https://github.com/nextcloud/server/pull/35527#discussion_r1036811493) as it is not included in `propsData` @raimund-schluessler, open to ideas if another way is better?

If `role="button"` is necessary, maybe it's a good idea to add it to `NcButton` by default? @skjnldsv what do you think?

Pytal:
Probaby here
https://github.com/nextcloud/nextcloud-vue/blob/288d451dc68303ee4f8b6ef491277ec60921e321/src/components/NcButton/NcButton.vue#L330-L334
then?

Will wait for input from @skjnldsv

skjnldsv:
> If `role="button"` is necessary, maybe it's a good idea to add it to `NcButton` by default? @skjnldsv what do you think?

Absolutely :+1:

artonge (approved):

PR #3539, opened 2022-12-02
Fix icon repeat and position for AppNavigationItem

skjnldsv:
I don't really understand how no one experienced this before...

EDIT: ah, ok.
Some icons do not start with `icon-`
image

| Before | After |
|:---------:|:------:|
|||

### DIFF
--- a/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
+++ b/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
@@ -682,6 +682,8 @@ export default {
 			width: $clickable-area;
 			height: $clickable-area;
 			background-size: $icon-size $icon-size;
+			background-repeat: no-repeat;
+			background-position: $icon-margin center;
 		}
 
 		.app-navigation-entry__title {


raimund-schluessler:
> I don't really understand how no one experienced this before...

Maybe everyone ideally uses material design icons 😉 🙈

raimund-schluessler (approved):

szaimen:
Why is the css not scoped?

skjnldsv:
> Why is the css not scoped?

No clue, not relevant to this PR :see_no_evil:

szaimen:
@skjnldsv are you sure that it is unrelated? wouldn't scoping the css not disable the impact on legacy components or do I understand something incorrectly?

skjnldsv:
> I understand something incorrectly?

Yes, this is not the issue. The issue is those two missing css rules that are implied because icons classes starts with `icon-`, but this ule is not true for legacy files icons
e.g. `nav-icon-trashbin`

Feel free to try [in the vue styleguide](https://nextcloud-vue-components.netlify.app/#/Components/App%20containers/NcAppNavigation?id=ncappnavigationitem) with `<NcAppNavigationItem title="My title" icon="nav-icon-systemtagsfilter" />`
image

szaimen (approved):
Fine by me then 👍

Issue #3550, opened 2022-12-07
[BITV]: Add title prop to components for native tooltips


JuliaKirschenheuter:
Part of https://github.com/nextcloud/nextcloud-vue/issues/2503

  - [ ] NC Vue components: (extend functionality)
      -  NcActionTextEditable (not needed for files && files_sharing)
      -  NcActionInput (not needed for files && files_sharing)
      -  NcButton
      -  NcAvatar
      -  NcAppSidebar
      - NcActions
      - NcActionLink

ChristophWurst:
Items moved into https://github.com/nextcloud/server/issues/35770 and https://github.com/nextcloud/server/issues/35654 respectively.

PR #3569, opened 2022-12-13
Adapt avatar and NcButton to correct showing of native tooltip

JuliaKirschenheuter:
Resolves https://github.com/nextcloud/nextcloud-vue/issues/3550

A11y issue


ChristophWurst:
> Resolves #3550

The ticket mentions five components, this PR touches two. Please clarify.

ChristophWurst:
Title adjusted to reflect the actual changes.

src/components/NcAvatar/NcAvatar.vue
```diff
@@ -374,7 +387,12 @@ export default {
 			if (this.ariaLabel !== null) {
 				return this.ariaLabel
 			}
-
+			if (this.title) {
+				if (this.hasStatus && this.showUserStatus && this.showUserStatusCompact) {
+					return t('Avatar of {title}, {status}', { title: this.title ?? this.user, status: this.userStatus.status })
+				}
+				return t('Avatar of {title}', { title: this.title ?? this.user })
```
PVince81, line 394: `this.title` will always be defined here because you exclude this if block by checking if `this.title` is defined

maybe set the fallback to user display name earlier

please double check if `this.user` is the display name and not the user id

st3iny (commented):
Code looks good. What about the other 2 components (ref what Christoph asked)?

JuliaKirschenheuter:
> The ticket mentions five components, this PR touches two. Please clarify.

Adapted NcAppSidebar and NcButton, thnaks. Other components won't be used for files && files_sharing. I will create other PR for other componets at the right time.

ChristophWurst:
Could you please create it now?

The title of this PR is once again misleading if you are saying that the title prop change is only partially done with your changes.

src/components/NcAvatar/NcAvatar.vue
JuliaKirschenheuter: this.user is the display name and not the user id

JuliaKirschenheuter: > this.title will always be defined here because you exclude this if block by checking if this.title is defined

make sense. I will leave comp prop ```avatarAriaLabel``` as it already was, just added a small tooltip (in form of ```title```) to ```<div>```. Detailed description will be found in ```aria-label```

marcoambrosini (approved):

ChristophWurst (approved):

PR #3575, opened 2022-12-16
Add `NcSavingIndicatorIcon`

susnux:
An icon to indicate the saving state of a document, file... as used by the *text* and the *office* app.
The icon is based on the material design icon `Check` combined with `CircleMedium`.

I think as this is now used by two apps we should provide this for other apps as well so they do not need to reinvent the wheel.

For design review see this PR which introduces the icon for the text app: https://github.com/nextcloud/text/pull/3572


src/components/NcSavingIndicatorIcon/NcSavingIndicatorIcon.vue
```diff
@@ -0,0 +1,130 @@
...
+  -->
+
+<docs>
+# Usage
+
+```
+<template>
+<div>
+	<div>
+		<NcSavingIndicatorIcon :title="`${savingState === 'error' ? 'Error saving' : savingState} file`"
```
nickvergessen, line 28: This should be translated

susnux: This is just the documentation

nickvergessen: Oh, sorry

skjnldsv (approved):
Approved by the design team, good from me too  :+1:

raimund-schluessler (approved):

PR #3612, opened 2023-01-04
feat(AppNavigationItem): Add href support

skjnldsv:
Vue router does NOT support external href

```vue
<NcAppNavigationItem title="Files" href="/index.php/apps/files" />
<NcAppNavigationItem title="Nextcloud" href="https://nextcloud.com" />
```

### DIFF
--- a/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
+++ b/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
@@ -89,6 +89,7 @@ prevent the user from collapsing the items.
 	</template>
 </NcAppNavigationItem>
 ```
+
 ### Editable element
 Add the prop `:editable=true` and an edit placeholder if you need it. By default
 the placeholder is the previous title of the element.
@@ -97,6 +98,7 @@ the placeholder is the previous title of the element.
 <NcAppNavigationItem title="Editable Item" :editable="true"
 	editPlaceholder="your_placeholder_here" icon="icon-folder" @update:title="function(value){alert(value)}" />
 ```
+
 ### Undo element
 Just set the `undo` and `title` props. When clicking the undo button, an `undo` event is emitted.
 
@@ -104,6 +106,16 @@ Just set the `undo` and `title` props. When clicking the undo button, an `undo` 
 <NcAppNavigationItem :undo="true" title="Deleted important entry" @undo="alert('undo delete')"  />
 
 ```
+
+### Link element
+Href that start by http will be treated as external and opened in a new tab
+```
+<div>
+	<NcAppNavigationItem title="Files" href="/index.php/apps/files" />
+	<NcAppNavigationItem title="Nextcloud" href="https://nextcloud.com" />
+</div>
+```
+
 ### Pinned element
 Just set the `pinned` prop.
 ```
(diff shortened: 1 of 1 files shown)


src/components/NcAppNavigationItem/NcAppNavigationItem.vue
```diff
@@ -296,6 +309,15 @@ export default {
 			default: '',
 		},
 
+		/**
+		 * A direct link. This will be used as the `href` attribute.
+		 * This will ignore any `to` prop being defined.
+		 */
+		href: {
+			type: String,
+			default: '#',
```
raimund-schluessler, line 318:
```suggestion
			default: null,
```
Otherwise the later `if (this.to && !this.href)` clause won't work as expected.

raimund-schluessler (changes requested):
See https://github.com/nextcloud/nextcloud-vue/pull/3612/files#r1061352128

src/components/NcAppNavigationItem/NcAppNavigationItem.vue
skjnldsv: Forgot to change it! Well caught :wink:

src/components/NcAppNavigationItem/NcAppNavigationItem.vue
```diff
@@ -586,6 +608,13 @@ export default {
 				this.$refs.actions.$refs.menuButton.$el.blur()
 			}
 		},
+
+		/**
+		 * Is this an external link
```
artonge, line 613: External means not from the same domain, no ?
Here it looks like you just check if it has a protocol. Also, there is probably some domains that starts with `'http'`

artonge: Also, explicitly check if `href` is not `null`: `href !== null`.

src/components/NcAppNavigationItem/NcAppNavigationItem.vue
```diff
@@ -469,7 +491,7 @@ export default {
 
 		// This is used to decide which outer element type to use
 		navElement() {
-			if (this.to) {
+			if (this.to && !this.href) {
```
artonge, line 494: Same, check if `href` is explicitely `null`

src/components/NcAppNavigationItem/NcAppNavigationItem.vue
skjnldsv: Why not check against invalid href (e.g. empty strings?)

skjnldsv: > External means not from the same domain, no ?

Well, yes and no, a nextcloud instance can have multiple domains.
That's why I checked for http only :)

artonge: I double-checked, and it should not cause an issue here, but I am always afraid of implicit comparisons. :fearful:  We have some of them in the backend that caused issues. Maybe in the front-end too, but can't remember a specific example. I usually tend to avoid them as it reduce easy mistake, and cognitive load.

image

artonge: > Well, yes and no, a nextcloud instance can have multiple domains.

Indeed

If it is only a protocol check, then I would call it that way and use a regexp to check for `'://'`. Something around `/https?:\/\//` ?

skjnldsv: > I double-checked, and it should not cause an issue here, but I am always afraid of implicit comparisons.

This is not a comparison, this is a falsy check. 
image

artonge (approved):

PR #3614, opened 2023-01-05
Fix action input styles

Pytal:
Before | After
--- | ---
image | image
image | image


PVince81 (approved):
:+1:

raimund-schluessler (approved):

raimund-schluessler:
MIght be a good idea to use an `NcInputField` for the action input fields to get the same styling.

skjnldsv (approved):

PR #3617, opened 2023-01-05
feat(AppNavigationItem): Allow native title attribute control

skjnldsv:
From nc25 (not vue):
Capture d’écran_2023-01-05_17-10-04

Since we're moving away from tooltips libraries.
```
<NcAppNavigationItem name="My title" icon="icon-category-enabled" />
<NcAppNavigationItem name="My title" title="My custom title" icon="icon-category-enabled" />
```

### DIFF
--- a/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
+++ b/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
@@ -33,7 +33,7 @@
 
 ```vue
 	<template>
-		<NcAppNavigationItem title="My title">
+		<NcAppNavigationItem name="My title">
 			<template #icon>
 				<Check :size="20" />
 			</template>
@@ -52,7 +52,7 @@
 * With a spinning loader instead of the icon:
 
 ```
-<NcAppNavigationItem title="Loading Item" :loading="true" />
+<NcAppNavigationItem name="Loading Item" :loading="true" />
 ```
 
 ### Element with actions
@@ -62,7 +62,7 @@ button will be automatically created.
 ```vue
 	<template>
 		<div id="app-navigation-vue"><!-- Just a wrapper necessary in the docs. Not needed when NcAppNavigation is correctly used as parent. -->
-			<NcAppNavigationItem title="Item with actions">
+			<NcAppNavigationItem name="Item with actions">
 				<template #icon>
 					<Check :size="20" />
 				</template>
@@ -79,7 +79,7 @@ button will be automatically created.
 						</template>
 						Delete
 					</NcActionButton>
-					<NcActionLink title="Link" href="https://nextcloud.com">
+					<NcActionLink name="Link" href="https://nextcloud.com">
 						<template #icon>
 							<OpenInNew :size="20" />
 						</template>
@@ -110,7 +110,7 @@ Just nest the counter in a template within `<NcAppNavigationItem>` and add `#cou
 
 ```vue
 	<template>
-		<NcAppNavigationItem title="Item with counter">
+		<NcAppNavigationItem name="Item with counter">
 			<template #icon>
 				<Folder :size="20" />
 			</template>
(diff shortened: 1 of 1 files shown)


raimund-schluessler:
I think I don't understand this change. We now have two `title` string props, and when one is evaluating to `false` we use the other one. That doesn't seem to be reasonable to me. You could just set `title` correctly where `NcAppNavigationItem` is used, no?

skjnldsv:
> You could just set `title` correctly where `NcAppNavigationItem` is used, no?

Well, title is being used as the string rendered in the entry.
It's also being used as the tooltip. So if I want to specify the tooltip only, I am stuck :(

Properly using title would be best, but hey, this is the least I can do without doing a breaking change and renaming the content to something different than  `title` :(

raimund-schluessler (approved):

skjnldsv:
> Maybe it would be clearer, if we name the new prop something with `tooltip`, either just `tooltip` or `tooltipString` or so. Besides this nitpick, fine with me.

OR, we introduce a new `name` prop, and if not defined, we fallback to title.
That way, since title was already use fo the tooltip, there is no breaking change, but we can deprecate using `title` for the entry name ? :thinking:

skjnldsv:
Done, please review again @raimund-schluessler

src/components/NcAppNavigationItem/NcAppNavigationItem.vue
```diff
@@ -438,7 +456,7 @@ export default {
 	emits: [
 		'update:menuOpen',
 		'update:open',
-		'update:title',
```
raimund-schluessler, line 441: This should stay. We still emit title.

src/components/NcAppNavigationItem/NcAppNavigationItem.vue
```diff
@@ -140,14 +144,14 @@ Just set the `pinned` prop.
...
 			class="app-navigation-entry">
-			<!-- Icon and title -->
+			<!-- Icon and name -->
 			<a v-if="!undo"
 				class="app-navigation-entry-link"
 				:aria-description="ariaDescription"
 				:aria-expanded="opened.toString()"
 				:href="href || '#'"
 				:target="isExternal(href) ? '_blank' : ''"
-				@focus="handleFocus"
```
raimund-schluessler, line 150: Did you remove this by accident?

src/components/NcAppNavigationItem/NcAppNavigationItem.vue
skjnldsv: Well, this is for documentation only.
Doesn't have any effect for vue2 and should most likely not be documented as it's deprecated now

src/components/NcAppNavigationItem/NcAppNavigationItem.vue
skjnldsv: Yeeeep :see_no_evil:

src/components/NcAppNavigationItem/NcAppNavigationItem.vue
raimund-schluessler: Hm, fine with me. Vue3 will complain, but since it'll be a major anyway, it's probably ok.

src/components/NcAppNavigationItem/NcAppNavigationItem.vue
```diff
@@ -265,11 +270,25 @@ export default {
...
+		 */
+		name: {
+			type: String,
+			// TODO: Make it required in the next major release (see title prop)
+			default: '',
+		},
+
+		/**
+		 * The title attribute of the element.
+		 *
```
raimund-schluessler, line 284: Lint complains.
```suggestion
		 *
```

src/components/NcAppNavigationItem/NcAppNavigationItem.vue
skjnldsv: Perfect, one reason to make sure this is removed in the next major! :stuck_out_tongue_closed_eyes:

src/components/NcAppNavigationItem/NcAppNavigationItem.vue
skjnldsv: It's the little things :p

raimund-schluessler:
Rebased.

PR #3621, opened 2023-01-06
fix(actions): Fix cypress actions

skjnldsv:
It has stopped recording our runs for 2 months now........

To use latest v4 or above (v5), I think we need cypress >=10


skjnldsv:
Fixed on #3609

PR #3630, opened 2023-01-07
Add NcIconSvgWrapper to public API

Pytal:
image

Unnecessary SVG title element addition is dropped as screenreaders will read the `aria-label` on `span` so we also reduce system resource usage

### DIFF
--- a/l10n/messages.pot
+++ b/l10n/messages.pot
@@ -62,9 +62,6 @@ msgstr ""
 msgid "Error getting related resources"
 msgstr ""
 
-msgid "Error parsing svg"
-msgstr ""
-
 msgid "External documentation for {title}"
 msgstr ""
 
--- a/src/components/NcIconSvgWrapper/NcIconSvgWrapper.vue
+++ b/src/components/NcIconSvgWrapper/NcIconSvgWrapper.vue
@@ -20,19 +20,84 @@
   -
 -->
 
+<docs>
+### Description
+
+Render raw SVG string icons.
+
+### Example
+
+```vue
+<template>
+	<div class="grid">
+		<NcButton>
+			<template #icon>
+				<NcIconSvgWrapper :svg="closeSvg" title="Close" />
+			</template>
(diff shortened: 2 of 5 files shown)


skjnldsv (approved):
Muuuch cleaner! Thank you Chris! :rocket:

skjnldsv:
There are two followups while we're at it:
- We recently found out that we could also use [DOMPurify](https://github.com/cure53/DOMPurify) as a sanitizer. That would avoid myself maintaining `@skjnldsv/sanitize-svg` :grin: 
- Ideally we could support the `size` and `fillColor` prop, so we could also use that like a `vue-material-design-icon` ?
  https://github.com/robcresswell/vue-material-design-icons/blob/681f4550c368d1acecb2fd3b63b2790f842598b2/build.ts#L13-L51
- Another additional sanity check would be to use [is-svg](https://www.npmjs.com/package/is-svg) to ensure the string is an svg before even sanitising it :thinking:

raimund-schluessler:
>    * We recently found out that we could also use [DOMPurify](https://github.com/cure53/DOMPurify) as a sanitizer. That would avoid myself maintaining `@skjnldsv/sanitize-svg` grin

`DOMPurify.sanitize()` works fine for SVGs too, I use it in an SVGViewer component (that will be replaced by this one here once it's out :wink:). I already wondered why `@skjnldsv/sanitize-svg` exists :see_no_evil:

>    * Ideally we could support the `size` and `fillColor` prop, so we could also use that like `vue-material-design-icon` ?
>       https://github.com/robcresswell/vue-material-design-icons/blob/681f4550c368d1acecb2fd3b63b2790f842598b2/build.ts#L13-L51

That would be nice :+1

raimund-schluessler (approved):

skjnldsv:
> I already wondered why `@skjnldsv/sanitize-svg` exists

You and me both lol
We searched for a solution and forked the first one we found working.

PR #3641, opened 2023-01-11
v7.4.0

Pytal:
https://github.com/nextcloud/nextcloud-vue/compare/v7.3.0...master


PVince81 (approved):
:+1:

skjnldsv (changes requested):
I would prefer to be safe and do a beta first please
https://github.com/nextcloud/nextcloud-vue/pull/3643

PR #3656, opened 2023-01-14
Set placement of select dropdown

Pytal:
Allow setting the placement of the dropdown to open above the input in addition to the default of opening below

image

This also updates the dropdown position automatically to keep the element within the viewport when e.g. scrolling

Before | After
--- | ---
image | image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
 			"version": "7.4.0",
 			"license": "AGPL-3.0",
 			"dependencies": {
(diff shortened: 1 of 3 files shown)


JuliaKirschenheuter (approved):

artonge (approved):

Pytal:
Minor updates + example for the styleguide!

marcoambrosini (approved):

Issue #3677, opened 2023-01-25
[BITV]: Rework native tooltips on NcAvatar


JuliaKirschenheuter:
## Problem

To show a tooltip on a several components we currently use v-tooltip. An example can be found in https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/NcAvatar/NcAvatar.vue#L112

v-tooltip is not accessible and we want to migrate to native tooltips (title attribute) instead. An overview can be found in https://github.com/nextcloud/nextcloud-vue/issues/2503.

## Solution / ToDo:

- [ ]  Add a `title` prop to NcAvatar component that will render as `title` attribute on the HTML element. Remove custom tooltips.
- [ ] Show valid value for `aria-label` every time.

PR #3678, opened 2023-01-25
Replace custom toolip with native one on Avatar component

JuliaKirschenheuter:
Fixes https://github.com/nextcloud/nextcloud-vue/issues/3677

- [x] Remove custom tooltips.
- [x] Remove unneeded props `aria-label` and `title`.

|  Before   | After                                                        |
|-------------|---------------------------------------------------------------|
| Screenshot from 2023-01-25 15-26-14 | Screenshot from 2023-01-25 15-22-51 | 
| Screenshot from 2023-01-25 15-25-58 |  Screenshot from 2023-01-25 15-23-03 |

### DIFF
--- a/src/components/NcAvatar/NcAvatar.vue
+++ b/src/components/NcAvatar/NcAvatar.vue
@@ -109,13 +109,12 @@ export default {
 </docs>
 <template>
 	<div ref="main"
-		v-tooltip="tooltip"
 		v-click-outside="closeMenu"
 		:class="{
 			'avatardiv--unknown': userDoesNotExist,
 			'avatardiv--with-menu': hasMenu
 		}"
-		:title="title"
+		:title="tooltip"
 		:style="avatarStyle"
 		class="avatardiv popovermenu-wrapper"
 		:tabindex="hasMenu ? '0' : undefined"
@@ -172,7 +171,6 @@ export default {
 import NcPopover from '../NcPopover/index.js'
 import NcPopoverMenu from '../NcPopoverMenu/index.js'
 import NcLoadingIcon from '../NcLoadingIcon/index.js'
-import Tooltip from '../../directives/Tooltip/index.js'
 import usernameToColor from '../../functions/usernameToColor/index.js'
 import { userStatus } from '../../mixins/index.js'
 import { t } from '../../l10n.js'
@@ -215,7 +213,6 @@ export default {
 
 	directives: {
 		ClickOutside,
-		tooltip: Tooltip,
 	},
 	components: {
 		DotsHorizontal,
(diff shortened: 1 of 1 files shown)


src/components/NcAvatar/NcAvatar.vue
```diff
@@ -327,17 +324,6 @@ export default {
 			default: null,
 		},
 
-		/**
-		 * Declares a native tooltip when not null
-		 *
-		 * requires disableTooltip not to be set to true
-		 * requires tooltipMessage not to be provided
-		 */
-		title: {
```
nickvergessen, line 336: Since you just added the title everywhere and now you drop it here, would it make sense to keep it and just fall back from tooltip to title if tooltip is empty, or something alike?

JuliaKirschenheuter: yes, good ideas come afterwards (reason for this issue). I will follow your suggestion.

ChristophWurst (approved):

src/components/NcAvatar/NcAvatar.vue
```diff
@@ -384,11 +381,8 @@ export default {
...
 	computed: {
 		avatarAriaLabel() {
-			if (!this.hasMenu) {
-				return
-			}
-			if (this.ariaLabel !== null) {
-				return this.ariaLabel
+			if (!this.displayName) {
+				return t('Avatar')
 			}
```
skjnldsv, line 386: Please also remove the unused `ariaLabel` prop/variable then

skjnldsv (changes requested):

src/components/NcAvatar/NcAvatar.vue
```diff
@@ -384,11 +381,8 @@ export default {
 	},
 	computed: {
 		avatarAriaLabel() {
-			if (!this.hasMenu) {
-				return
-			}
```
Pytal, line 389: I believe this should be kept as `aria-label` should only be present on interactive elements as documented in https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label

src/components/NcAvatar/NcAvatar.vue
JuliaKirschenheuter: I've looked through my changes last time and there are only 2 Apps with NcAvatar to change: 

- [ ] files_sharing, https://github.com/nextcloud/server/pull/35655/files
- [ ] privacy, https://github.com/nextcloud/privacy/pull/848/files

I would like to remove unneeded attrs completely because it is better and clearer implementation.

Pytal (approved):

PR #3680, opened 2023-01-26
Add aria-labels to colorpicker buttons

raimund-schluessler:
This fixes the missing text / aria-labels warnings from the NcColorpicker.

### DIFF
--- a/src/components/NcColorPicker/NcColorPicker.vue
+++ b/src/components/NcColorPicker/NcColorPicker.vue
@@ -188,6 +188,7 @@ export default {
 			<div class="color-picker__navigation">
 				<NcButton v-if="advanced"
 					type="tertiary"
+					:aria-label="ariaBack"
 					@click="handleBack">
 					<template #icon>
 						<ArrowLeft :size="20" />
@@ -195,6 +196,7 @@ export default {
 				</NcButton>
 				<NcButton v-if="!advanced"
 					type="tertiary"
+					:aria-label="ariaMore"
 					@click="handleMoreSettings">
 					<template #icon>
 						<DotsHorizontal :size="20" />
@@ -213,7 +215,7 @@ export default {
 <script>
 import NcButton from '../NcButton/index.js'
 import NcPopover from '../NcPopover/index.js'
-import l10n from '../../mixins/l10n.js'
+import { t } from '../../l10n.js'
 import GenColors from '../../utils/GenColors.js'
 
 import ArrowLeft from 'vue-material-design-icons/ArrowLeft.vue'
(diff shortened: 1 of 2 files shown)


src/components/NcColorPicker/NcColorPicker.vue
```diff
@@ -284,6 +284,8 @@ export default {
 		return {
 			currentColor: this.value,
 			advanced: false,
+			ariaBack: t('Back'),
+			ariaMore: t('More options'),
```
skjnldsv, line 288: Are we not allowed to have those in the template? I remember something like this, no idea if we fixed it or not
I think we did. I see a few `t` functions above in the template ?

raimund-schluessler: The l10n extraction does not work if it's in the template, unless its in curly brackets apparently. No idea why, though, or how to fix it.

raimund-schluessler: It's done like this for `NcAppSidebar` and other components as well. See here:
https://github.com/nextcloud/nextcloud-vue/blob/f23f8e42f6f1b4c04ffe6c6ee1a77004f4a2e517/src/components/NcAppSidebar/NcAppSidebar.vue#L455-L457

Pytal (approved):

src/components/NcColorPicker/NcColorPicker.vue
skjnldsv: Yes, no, I mean, I remember something like that, but I've seen a lot of translations in template lately, including still in this PR: https://github.com/nextcloud/nextcloud-vue/blob/e4b181c454e4dcbde393fb39a56b57e3cac1ac7e/src/components/NcColorPicker/NcColorPicker.vue#L207-L209

My question was more:
1. if still required, we should also fix the other `t` in the template
2. if not, then we can just plainly move the `ariaBack` and `ariaMore` back into the template

raimund-schluessler: It is still required, but the line you mention works for some reason. You can find the term in the pot file and l10n:extract finds changes to the term:
https://github.com/nextcloud/nextcloud-vue/blob/82c46cb1ec80032877c18f0531bbf689727be885/l10n/messages.pot#L35-L36

So, this works:
```vue
<template>
    <element>
 	{{ t('Choose') }} 
    </element>
</template>
```

This does not:
```vue
<template>
    <element :prop="t('Choose')">
</template>
```

Don't ask me why this does not work. The status quo and how we do it in this PR is the best we can at the moment.

skjnldsv: Good to know!! :)

skjnldsv (approved):

PR #3690, opened 2023-01-27
Fix keyboard navigation of modal slides

Pytal:
We handle keydown events on the global window object
https://github.com/nextcloud/nextcloud-vue/blob/58b5c435979784c5143c1b539e52c9ada826981f/src/components/NcModal/NcModal.vue#L536
and click events on the previous and next arrow buttons
https://github.com/nextcloud/nextcloud-vue/blob/58b5c435979784c5143c1b539e52c9ada826981f/src/components/NcModal/NcModal.vue#L264
https://github.com/nextcloud/nextcloud-vue/blob/58b5c435979784c5143c1b539e52c9ada826981f/src/components/NcModal/NcModal.vue#L296

So when keyboard focus is on the previous or next button and we hit `Enter`, the click handler of either previous or next will fire and the global handler will always call `this.next()`

Focused element | Result
--- | ---
Previous button | Back then forward to the same slide
Next button | Forward two slides

### DIFF
--- a/src/components/NcModal/NcModal.vue
+++ b/src/components/NcModal/NcModal.vue
@@ -612,7 +612,6 @@ export default {
 			case 37: // left arrow
 				this.previous(e)
 				break
-			case 13: // enter key
 			case 39: // right arrow
 				this.next(e)
 				break


szaimen (approved):
LGTM but didnt test

src/components/NcModal/NcModal.vue
```diff
@@ -612,7 +612,6 @@ export default {
 			case 37: // left arrow
 				this.previous(e)
 				break
-			case 13: // enter key
```
artonge, line 615: So pressing enter when nothing is focused, will not go to the next slide anymore ?

Maybe we can add a listener for enter and stop the propagation ? Something like:

```vue
<NcButton v-show="hasPrevious"
	...
	@click="previous"
	@keydown.enter.stop="previous">
	...
</NcButton>
```

Pytal: 🤔 the modal is focus trapped so some element will always be focused on open

After it's opened though the user could click on some empty space in the modal which gives focus to `body`, but having to click empty space so that `Enter` goes to the next slide would be strange UX

JuliaKirschenheuter (approved):

src/components/NcModal/NcModal.vue
skjnldsv: We could also debounce this.
But stopping the propagation should be the solution. To avoid firing further events listeners.

artonge (approved):

src/components/NcModal/NcModal.vue
Pytal: Using debounce here to allow for the specific case of clicking empty space to put focus on `body` first before navigating seems unintuitive

Right arrow navigation seems more consistent and predictable

PR #3695, opened 2023-01-30
fix: Export NcHighlight through default package export

juliushaertl:
Otherwise importing with `import { NcHighlight } from ‘@nextcloud/vue'` fails


julien-nc:
> Seems like no one wanted to use it until now smilesee_no_evil

Or everyone used it like that:
`import NcHighlight from '@nextcloud/vue/dist/Components/NcHighlight.js'`
That's what I was doing :grin:.

julien-nc (approved):

PR #3701, opened 2023-01-31
Adjust style of NcSelect to match other input elements

susnux:
* Border should be primary color on hover
* Border should be stay primary color if active
* Unify the style of the `no-options` slot with `NcMultiselect`

### DIFF
--- a/src/components/NcSelect/NcSelect.vue
+++ b/src/components/NcSelect/NcSelect.vue
@@ -965,6 +965,15 @@ export default {
 		margin-right: 2px;
 	}
 
+	&.vs--open .vs__dropdown-toggle {
+		border-color: var(--color-primary);
+		border-bottom-color: transparent;
+	}
+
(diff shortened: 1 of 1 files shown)


raimund-schluessler (approved):

src/components/NcSelect/NcSelect.vue
```diff
@@ -983,7 +983,20 @@ export default {
...
 
+.v-select {
+	&:not(.vs--open) .vs__dropdown-toggle:hover {
+		border-color: var(--color-primary);
+	}
+
+	&.vs--open .vs__dropdown-toggle {
+		border-color: var(--color-primary);
+		border-bottom-color: transparent;
+	}
```
Pytal, line 994: Looks good!

Let's merge these declarations into the `.v-select.select` block above as well :)

susnux: Done!

artonge (approved):

Pytal (approved):

PR #3708, opened 2023-02-01
Add link picker in RichContentEditable with tribute

julien-nc:
This integrates the vue-richtext's link picker in RichContentEditable with TributeJS. The trigger is "/". The item list is obtained from a vue-richtext function call.

The workflow:
* type /
* choose a link provider
* this opens a modal which contains either
    * a generic search (using one or multiple unified search providers)
    * a custom picker element
* user does the work which results in getting a link
* the link is inserted in the RCE content

As picking a link is asynchronous, I used a trick to make it work with Tribute:
A dummy invisible element is inserted immediately when opening the link picker (when selecting a tribute dropdown item). Once the picker returns the link (with a promise), the dummy element is replaced by the link.

:warning: For the link picker to get the provider list, the page in which RCE is used must have dispatched `RenderReferenceEvent`. This will be documented in Nextcloud's general link reference chapter. Maybe it's worth mentioning it in RCE's doc as well.

We could make this a little bit more generic (passing data and callback as RCE props) but we figured (with @juliushaertl) the link picker was strongly depending on unified search and vue-richtext and cannot really be used with other data sources anyway. What do you think?

https://user-images.githubusercontent.com/11291457/216076359-5c435ab4-544d-4d31-a33c-a8acf71e08ae.mp4

### DIFF
--- a/src/components/NcRichContenteditable/NcRichContenteditable.vue
+++ b/src/components/NcRichContenteditable/NcRichContenteditable.vue
@@ -148,6 +148,7 @@ import NcAutoCompleteResult from './NcAutoCompleteResult.vue'
 import richEditor from '../../mixins/richEditor/index.js'
 import Tooltip from '../../directives/Tooltip/index.js'
 import { emojiSearch, emojiAddRecent } from '../../functions/emoji/index.js'
+import { linkProviderSearch, getLink } from '../../functions/linkPicker/index.js'
 
 import Tribute from 'tributejs/dist/tribute.esm.js'
 import debounce from 'debounce'
@@ -227,6 +228,14 @@ export default {
 			type: Boolean,
 			default: true,
 		},
+
+		/**
+		 * Enable or disable link autocompletion
+		 */
+		linkAutocomplete: {
+			type: Boolean,
+			default: true,
+		},
 	},
 
 	emits: [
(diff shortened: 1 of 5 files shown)


juliushaertl (approved):

juliushaertl:
Uncommited l10n changes. Run `npm run l10n:extract`.

raimund-schluessler:
> Uncommited l10n changes. Run `npm run l10n:extract`.

We have an awesome github command for this now (which should probably be mentioned in the L10n check).

raimund-schluessler:
> /l10n-update

And which doesn't seem to work anymore, whereas it did just last week 🙈

juliushaertl:
@julien-nc Mind to rebase?

juliushaertl:
@ShGKme Maybe yiiz can also have a look since this would be relevant for Talk with https://github.com/nextcloud/spreed/pull/4333 :)

src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -381,9 +414,40 @@ export default {
 		if (this.emojiTribute) {
 			this.emojiTribute.detach(this.$el)
 		}
+		if (this.linkTribute) {
+			this.linkTribute.detach(this.$el)
+		}
 	},
 
 	methods: {
+		getLink(item) {
```
mejo-, line 423: I find it a bit confusing to have `this.getLink()` and global `getlink()` in the component, but that's only nitpicking.

mejo- (approved):
Code looks good to me, didn't test though.

PR #3714, opened 2023-02-02
Use proper default actions aria label in `NcListItem`

raimund-schluessler:
We currently overwrite the `aria-label` of the `NcActions` component used in the `NcListItem` component with an empty string, which leads to a `You need to fill either the text or the ariaLabel props in the button component.` triggered by `NcListItem` by default. This PR sets the default aria-label in `NcListItem` to the same value as used in `NcActions`.

### DIFF
--- a/src/components/NcListItem/NcListItem.vue
+++ b/src/components/NcListItem/NcListItem.vue
@@ -268,7 +268,7 @@
 						@click.prevent.stop="">
 						<NcActions ref="actions"
 							menu-align="right"
-							:aria-label="actionsAriaLabel"
+							:aria-label="computedActionsAriaLabel"
 							@update:open="handleActionsUpdateOpen">
 							<!-- @slot Provide the actions for the right side quick menu -->
 							<slot name="actions" />
@@ -281,7 +281,7 @@
 					@click.prevent.stop="">
 					<NcActions ref="actions"
 						menu-align="right"
-						:aria-label="actionsAriaLabel"
+						:aria-label="computedActionsAriaLabel"
 						@update:open="handleActionsUpdateOpen">
 						<!-- @slot Provide the actions for the right side quick menu -->
 						<slot name="actions" />
@@ -300,6 +300,7 @@
 <script>
 import NcActions from '../NcActions/index.js'
 import NcCounterBubble from '../NcCounterBubble/index.js'
+import { t } from '../../l10n.js'
 
 export default {
 	name: 'NcListItem',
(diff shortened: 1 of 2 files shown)


artonge:
I think this is on purpose for accessibility to not have multiple actions with the same aria label

raimund-schluessler:
> I think this is on purpose for accessibility to not have multiple actions with the same aria label

Well, every action has this label unless you overwrite it explicitly. So there are multiple actions with this label on the same page already. Not sure what's the benefit of setting an empty aria-label (besides triggering a warning about it 🙈).

skjnldsv (approved):

artonge:
> Not sure what's the benefit of setting an empty aria-label for some actions, but not all (besides triggering a warning about it see_no_evil).

Then maybe the fix would be to remove this default aria-label on the actions that have a default one ?

raimund-schluessler:
> Then the fix would be to remove the default aria-label on the actions that have a one ?

That would mean making the aria-label prop required, otherwise the `NcButton` component will complain about every actions component used. That would be a breaking change. And I think devs would then just end up setting the default aria-labels that we just removed in the lib. Input from the accessibility team would indeed be good.

Pytal (approved):
Seems fine to me :)

What do you think @JuliaKirschenheuter?

src/components/NcListItem/NcListItem.vue
```diff
@@ -396,7 +396,7 @@ export default {
 		 */
 		actionsAriaLabel: {
```
JuliaKirschenheuter, line 397: i would suggest to create computed prop `actionsAriaLabel` and create a value like `t('Actions for item with title {propertyName}', { propertyName: this.title})`
And drop `actionsAriaLabel` prop. That would be a small breaking change, i think with time we can replace all remaining places.

JuliaKirschenheuter (changes requested):

src/components/NcListItem/NcListItem.vue
raimund-schluessler: We could do that without a breaking change, by keeping the `actionsAriaLabel` prop and either use the prop if provided or the computed if not. Is this fine? Then I will adjust the PR.

JuliaKirschenheuter: A bit "duplication" but fine -> no breaking change.

raimund-schluessler: Done.

JuliaKirschenheuter (approved):

Issue #3687, opened 2023-01-27
Feature request: native title tooltip on NcListItem title


ShGKme:
It seems to be impossible to set the `title` attribute on `<NcListItem>`.

What about adding a `nativeTitle` prop or even more, adding title-tooltip feature for long titles to `NcListItem`?

Use case: display full title (from prop) in title (tooltip), when it is too long.

![Tooltip on a long conversation name with ellipses](https://user-images.githubusercontent.com/93392545/214914351-9ca30e36-6c0d-4476-90b6-b0610adaa260.png)

Related:
- https://github.com/nextcloud/nextcloud-vue/issues/2503
- https://github.com/nextcloud/spreed/pull/8605

nickvergessen:
https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/NcListItem/NcListItem.vue#L140

Documentation says it's there.
I think they are automatically passed on when they are not props. So simply using title works already.

ShGKme:
> https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/NcListItem/NcListItem.vue#L140
> 
> Documentation says it's there. I think they are automatically passed on when they are not props. So simply using title works already.

But `title` is a prop:

https://github.com/nextcloud/nextcloud-vue/blob/e65bd9247874264aa1d57d59d79f53391e31faa1/src/components/NcListItem/NcListItem.vue#L321-L327

It is used only as the text content of the title element:

https://github.com/nextcloud/nextcloud-vue/blob/e65bd9247874264aa1d57d59d79f53391e31faa1/src/components/NcListItem/NcListItem.vue#L230-L234

Antreesy:
Seems closed by https://github.com/nextcloud-libraries/nextcloud-vue/pull/4106

PR #3715, opened 2023-02-02
Implement slot for title in ListItem

Antreesy:
Signed-off-by: Maksim Sukharev <antreesy.web@gmail.com>

## Problem
Right now there is no stable opportunity to manipulate 'Title' sub-component (see issues and PR below) - only through the access to DOM element via `querySelector` or using `:deep` selector for scoped styles.

**Related:**
- https://github.com/nextcloud/nextcloud-vue/issues/3687
- https://github.com/nextcloud/spreed/pull/8605

## Proposal
Add the ability to integrate a developer-controlled sub-component via Vue `slot`. Proposed solution should not affect current realisations, because required prop `title` is provided as a fallback for slot.

image
- [x] Examples for documentation is provided

## :warning:  IMPORTANT
`.list-item` selector doesn't inherit `box-sizing: border-box` by default (If NcContent is not used). Fixed by this PR.

### DIFF
--- a/src/components/NcListItem/NcListItem.vue
+++ b/src/components/NcListItem/NcListItem.vue
@@ -84,6 +84,12 @@
 			<template #icon>
 				<NcAvatar :size="44" user="janedoe" display-name="Jane Doe" />
 			</template>
+			<template #title>
+				<span style="display: flex; color: var(--color-primary);">
+					Title of the element with content
+					<div class="icon-edit" />
+				</span>
+			</template>
 			<template #subtitle>
 				In this slot you can put both text and other components such as icons
 			</template>
@@ -169,6 +175,12 @@
 		<template #icon>
 			<div class="icon-edit" />
 		</template>
+		<template #title>
+			<span style="display: flex; color: var(--color-primary);">
+				Title of the element with content
+				<div class="icon-edit" />
+			</span>
+		</template>
 		<template #subtitle>
 			This one is with subtitle
 		</template>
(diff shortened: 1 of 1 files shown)


raimund-schluessler:
I am not really fond of this solution. I think it's overkill to make this a slot, just for having the option to set the (native) title attribute. What's wrong with simply using `<span class="line-one__title" :title="title">`?

Antreesy:
> What's wrong with simply using `<span class="line-one__title" :title="title">`?

The trick is, we don't need this all the time. For example, in Talk it's only implemented for long names, which are covered with ellipsis and aren't fully visible. So it is better to have controlled tag, where you can apply attributes, logic, styles, e.t.c. locally.

I also want to note that, as it's an open-source library, we shouldn't mess with components working good before changes. So, attribute `title` could conflict with someone else local realisation.

nickvergessen:
> For example, in Talk it's only implemented for long names, which are covered with ellipsis and aren't fully visible.

Which is a nice gimmic, but if that blocks a proper and performant solution I prefer to have the title just displayed all the time.

raimund-schluessler:
> The trick is, we don't need this all the time. For example, in Talk it's only implemented for long names, which are covered with ellipsis and aren't fully visible. So it is better to have controlled tag, where you can apply attributes, logic, styles, e.t.c. locally.

If that *really* is a requirement, we can also introduce a new prop for it. Implementation wise, we could go the same way as in https://github.com/nextcloud/nextcloud-vue/pull/3617, introduce a new `name` prop for what is now called `title` and make `name` fallback to the value of `title` if its not provided and use `title` as the native title value, to prevent a breaking change until the next major.

> I also want to note that, as it's an open-source library, we shouldn't mess with components working good before changes. So, attribute `title` could conflict with someone else local realisation.

We cannot possibly anticipate all possible ways someone is using this library. What is not documented, is not supported in my opinion. We frequently introduce new features and props. If we couldn't do this because it might break someones undocumented and unintended use-case, we would be left with an unmaintained library without progress, because every little change could break something.

Antreesy:
Superseded by https://github.com/nextcloud/spreed/pull/11455

PR #3724, opened 2023-02-06
No background for `tertiary-no-background` `NcActions`

### LINKED ISSUE #3720: NcActions with `tertiary-no-background` has background when open
When using the type `tertiary-no-background` I would expect the actions toggle to never have a background assigned, but when the actions are toggled the button gets the default background assigned.

This looks a bit odd in my case:
button with background overlays an input element

raimund-schluessler:
This removes the background for open `NcActions` of type `tertiary-no-background`.

Closes #3720.

| Type | Before | After |
|-|-|-|
| `tertiary` | Screenshot 2023-02-06 at 10-09-44 Nextcloud Vue Style Guide | Screenshot 2023-02-06 at 10-06-51 Nextcloud Vue Style Guide |
| `tertiary-no-background` | Screenshot 2023-02-06 at 10-10-43 Nextcloud Vue Style Guide | Screenshot 2023-02-06 at 10-07-20 Nextcloud Vue Style Guide |

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -1169,6 +1169,10 @@ export default {
 		--open-background-color: var(--color-success-hover);
 	}
 
+	&.action-item--tertiary-no-background {
+		--open-background-color: transparent;
+	}
+
 	&.action-item--open .action-item__menutoggle {
 		opacity: $opacity_full;
 		background-color: var(--open-background-color);


nimishavijay:
Looks good! One question: Are the 3 states (normal, hover and active/focus) well defined for this type of button? In the [deploy-preview component](https://deploy-preview-3724--nextcloud-vue-components.netlify.app/#/Components/NcButton) there seems to be no hover state? Or is that an unrelated issue? :)

raimund-schluessler:
> Looks good! One question: Are the 3 states (normal, hover and active/focus) well defined for this type of button? In the [deploy-preview component](https://deploy-preview-3724--nextcloud-vue-components.netlify.app/#/Components/NcButton) there seems to be no hover state? Or is that an unrelated issue? :)

This PR does not touch the `NcButton` component. So if you mean that an `NcButton` of type `tertiary-no-background` has no hover effect (neither text-color or background or similar), that's unrelated.

susnux (approved):
Thank you!

nimishavijay:
Oh sorry, you are right, I meant `NcButton` 
But seems like the `NcActions` component also doesn't have a hover state? There should be an explicit hover state indicated by, for eg. opacity of the icon :)

raimund-schluessler:
> Oh sorry, you are right, I meant `NcButton` But seems like the `NcActions` component also doesn't have a hover state? There should be an explicit hover state indicated by, for eg. opacity of the icon :)

That's correct. This is because `NcActions` uses `NcButton` internally. So unless we explicitly overwrite it (which we did for e.g. the background), the actions inherit the behaviour of the button. But this (missing) hover effect of a `tertiary-no-background` button is also not touched here 😉 

I would propose to discuss/fix this in a separate issue/PR, see #3727.

raimund-schluessler:
@nimishavijay It seems a hover feedback for `tertiary-no-background` was not desired, see https://github.com/nextcloud/nextcloud-vue/issues/3727#issuecomment-1422614541. Can we then merge this PR here now, removing the background-color?

nimishavijay:
Yes sounds good!

marcoambrosini (approved):

PR #3725, opened 2023-02-06
Simplify the render function of NcBreadcrumbs

raimund-schluessler:
This simplifies the render function of `NcBreadcrumbs` a bit. We need less `if` and less `concat` now. The behaviour stays the same, no visual changes.


artonge (approved):

susnux (approved):
:broom:

PR #3726, opened 2023-02-06
Move NcButton to render function

### LINKED ISSUE #3231: NcButton: icon-only button broken if using conditional template for text
Conditionally setting the text slot in a NcButton breaks the icon-only version of the button:

```
<NcButton>
    <template #icon>
        <SomeIcon :size="20" />
    </template>
    <template v-if="!isMobile">
        {{ title }}
    </template>
</NcButton>
```

This leads to empty buttons (with neither icon nor text) in some cases.

Apparently the button component is not re-rendered when the v-if condition changes.

Maybe @vinicius73 can explain better what's the problem here :wink:

raimund-schluessler:
This moves the `NcButton` component to use a render function. This gets rid of the reactivity caveat of using `$slots` outside the template or render function. With this, the button now also properly warns if text and aria-label become undefined during runtime and not only during mount. There are no visual changes resulting.

This changes also prepares for the migration to vue 3.

Closes #3231.

### DIFF
--- a/src/components/NcButton/NcButton.vue
+++ b/src/components/NcButton/NcButton.vue
@@ -197,24 +197,6 @@ button {
 ```
 </docs>
 
-<template>
-	<root-element class="button-vue"
-		v-bind="rootElement"
-		:class="buttonClassObject"
-		:aria-label="ariaLabel"
-		:disabled="disabled"
-		v-on="$listeners">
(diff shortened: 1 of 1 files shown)


marcoambrosini (approved):
👏

Pytal (approved):

PR #3728, opened 2023-02-06
Fix `NcButton` cursor style

raimund-schluessler:
Nextcloud server has a CSS rule `span { cursor: default; }` that would sometimes (depending on the order in which the bundle is loaded) overwrite the button cursor style. This should make it more explicit so it takes precedence over the server style.

### DIFF
--- a/src/components/NcButton/NcButton.vue
+++ b/src/components/NcButton/NcButton.vue
@@ -418,7 +418,8 @@ export default {
 
 	// Cursor pointer on element and all children
 	cursor: pointer;
-	& * {
+	& *,
+	span {
 		cursor: pointer;
 	}
 	border-radius: math.div($clickable-area, 2);


susnux:
So this is for `span` elements within the slots?

raimund-schluessler:
> So this is for `span` elements within the slots?

Well, it's supposed to be for the material design icon spans. But if you are worried that it messes with spans you put in the slot, I think it should also work without the :deep().

raimund-schluessler:
> So this is for `span` elements within the slots?

Well, it's supposed to be for the material design icon spans. But if you are worried that it messes with spans you put in the slot, I think it should also work without the :deep().

raimund-schluessler:
> No i just wondered why the `:deep` is required other than for nesting components.

Yes, I used it for the nested material design icons, just to be sure. But it seems to work without it as well, so I removed it now.

marcoambrosini (approved):

Pytal (approved):

susnux (approved):

PR #3729, opened 2023-02-06
fix(package): export more components to main package

juliushaertl:
Signed-off-by: Julius Härtl <jus@bitgrid.net>


raimund-schluessler:
The timezone picker isn't exported by accident, I would say.

The `NcInputField` is deliberately not exported, since it is an internal component used by `NcPasswordField` and `NcTextField`. As per the `NcInputField` docs:
https://github.com/nextcloud/nextcloud-vue/blob/27d09d66f582cedccdcb34223368d70daab06904/src/components/NcInputField/NcInputField.vue#L25-L29

juliushaertl:
> The `NcInputField` is deliberately not exported, since it is an internal component used by `NcPasswordField` and `NcTextField`. As per the `NcInputField` docs:

Ok, thanks. Then we'll need to migrate the usage in server of that one.

Pytal:
> @skjnldsv @Pytal Can you share some light on why two components were not exported for the main package bundle in

NcEllipsisedOption was originally only used by NcMultiselect but was also needed for the NcSelect* components so it was hoisted from the NcMultiselect folder in https://github.com/nextcloud/nextcloud-vue/pull/3435/commits/ed6f3ed2f82bb3a25246708fef3f8a5c6188103c#diff-f341f5a900ec8175943fcefd443dda0e5bd2dde34062b5461674f0d40b133175, if useful for developers then we can go ahead and export it as well

raimund-schluessler:
> > The `NcInputField` is deliberately not exported, since it is an internal component used by `NcPasswordField` and `NcTextField`. As per the `NcInputField` docs:
> 
> Ok, thanks. Then we'll need to migrate the usage in server of that one.

I guess that's to discuss with @skjnldsv.

juliushaertl:
Changed the comment a bit to be more clear about why this should not be used now.

raimund-schluessler (approved):

max-nextcloud (approved):

PR #3733, opened 2023-02-08
Fix throw when header menu contains no tabbable nodes

Pytal:
Fix focus trap error throw "Your focus-trap must have at least one container with at least one tabbable node in it at all times" with `fallbackFocus` option documented in https://github.com/focus-trap/focus-trap#createoptions

This occurs e.g. when opening the notifications menu when there are no notifications
image


raimund-schluessler (approved):
Works :+1:

JuliaKirschenheuter (approved):

JuliaKirschenheuter:
works!

Issue #3662, opened 2023-01-17
Tag search does not work anymore


nickvergessen:
Ref https://github.com/nextcloud/server/issues/36152

### Steps to reproduce

1. Create tags `test-1` until `test-10`
2. Go to any other place where tags can be selected (the tag management does not use the component yet), e.g. creating a workflow with automated tagging: https://github.com/nextcloud/files_automatedtagging
3. Search for `test-6` and try to pick it.

### Expected
Option is displayed

### Actual
The result list is empty

### Guess
The problem is that the custom filter is applied before the search:
https://github.com/nextcloud/nextcloud-vue/blob/4df172ae3869cec3733daa207308f6cf701aedbc/src/components/NcMultiselectTags/NcMultiselectTags.vue#L200-L205
https://github.com/nextcloud/nextcloud-vue/blob/84e7175ac2fd48c578187ae9de496ca3cbd498ac/src/components/NcSelectTags/NcSelectTags.vue#L261-L266

So we proceed with 5 items (default filter) into the search and then have no results left.

nickvergessen:
### Workaround
```js
:options-filter="() => true"
```
Can be used to remove the filter.

### Optimal solution
But of course it would be nice to have the filter apply after a potential search.

### Other solution

Another option would be to change the default filter to not filter when a search is performed if that would work:
```js
		optionsFilter: {
			type: Function,
			default: (_element, index) => index < 5 || this.search !== '',
		},
```

PVince81:
duplicate of https://github.com/nextcloud/server/issues/34395 ?

@Pytal

nickvergessen:
https://github.com/nextcloud/server/issues/36152 should be a duplicate of the server ticket.
Can keep one to remember the lib upgrade. But the vue repo ticket needs to stay :)

Pytal:
By previous design in https://github.com/nextcloud/nextcloud-vue/pull/2973 and https://github.com/nextcloud/nextcloud-vue/pull/2990

Since filtering is handled by the lib component, for NcSelectTags the best option is probably to use the https://vue-select.org/api/props.html#filter prop to truncate to 5 after the filtering logic

Should we do NcMultiselectTags as well @nickvergessen?

nickvergessen:
> Should we do NcMultiselectTags

Since it's deprecated we could also skip it and just ask to migrate to the new component.

Pytal:
Fixed in https://github.com/nextcloud/nextcloud-vue/pull/3747

PR #3747, opened 2023-02-14
Fix tag search

Pytal:
Fix https://github.com/nextcloud/nextcloud-vue/issues/3662 and make minor updates to `vue-select` prop handling

The `limit` prop is now used instead of `optionsFilter` to limit the available tags

Previously only 5 tags were being passed as options to `VueSelect`, since `VueSelect` handles searching internally any tag after the 5th one could not be searched for

Before | After
--- | ---
image | image

### DIFF
--- a/src/components/NcSelect/NcSelect.vue
+++ b/src/components/NcSelect/NcSelect.vue
@@ -669,7 +669,8 @@ export default {
 		/**
 		 * Key of the displayed label for object options
 		 *
-		 * Defaults to `'label'`
+		 * Defaults to the internal vue-select string documented at the link
+		 * below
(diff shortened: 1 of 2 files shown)


artonge (approved):

raimund-schluessler (approved):
Works.
But I have to admit, I find it weird to see only 5 tags in the dropdown and then new tags showing up when I filter. But that's a different story, I guess.

nickvergessen:
> But that's a different story, I guess.

It's a browser performance thing :)
Otherwise when you have thousands of tags (automated tagging from AI and things) they slow down the page loading

PR #3752, opened 2023-02-14
Add node-polyfill-webpack-plugin

### LINKED ISSUE #3696: Missing dependencies on node modules
After upgrading to 7.x, webpack build failed with several messages like:
````
Module not found: Error: Can't resolve 'assert' in '[...]/node_modules/@nextcloud/vue/dist'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.
````
Same for the "buffer" and "util" modules. As a workaround, I added those packages to my project:

`npm install assert buffer util`

The commonly suggested fallback configuration was not necessary.

Shouldn't these dependencies be handled by the nextcloud-vue package?

GretaD:
fixes #3696

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,6 +31,7 @@
 				"hammerjs": "^2.0.8",
 				"linkify-string": "^4.0.0",
 				"md5": "^2.3.0",
+				"node-polyfill-webpack-plugin": "^2.0.1",
 				"splitpanes": "^2.4.1",
 				"string-length": "^5.0.1",
 				"striptags": "^3.2.0",
@@ -2786,7 +2787,6 @@
 			"version": "3.1.0",
 			"resolved": "https://registry.npmjs.org/@jridgewell/resolve-uri/-/resolve-uri-3.1.0.tgz",
 			"integrity": "sha512-F2msla3tad+Mfht5cJq7LSXcdudKTWCVYUgw6pLFOOHSTtZlj6SWNYAp+AhuqLmWdBO2X5hPrLcu8cVP8fy28w==",
-			"dev": true,
 			"engines": {
 				"node": ">=6.0.0"
 			}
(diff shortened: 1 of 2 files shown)


GretaD:
i cannot run npm right now, will check tomorrow again

ChristophWurst:
I think this should be a peer dependency

Otherwise :+1:

GretaD:
> I think this should be a peer dependency
> 
> Otherwise +1

right, but we only have dependencies and devDependencies here, so should i switch to Dependencies?

ChristophWurst:
peerDependencies

raimund-schluessler:
> peerDependencies

While technically correct, listing this as a peer dependency is very likely a breaking change. Every app using nextcloud/vue will have to have this peer dependency in their dependency list. Otherwise npm will probably warn about it. I don't know how that behaves when node-polyfill-webpack-plugin is already installed as a dependency of another package (e.g. nextcloud/webpack-config), though.

But to keep it pragmatic, I would just list it as a dependency, to be honest.

ChristophWurst:
npm installs peers automatically, doesn't it?

adding it as dev dep is also fine :+1:

raimund-schluessler:
> npm installs peers automatically, doesn't it?

Indeed, it seems this feature came back with npm v7, it was gone since npm v3. As npm v7 is the lowest supported version here, it's probably fine. But I didn't test it.

ChristophWurst (approved):

GretaD:
> @GretaD You put it in `devDependencies` now. I think this will not work. And please order it alphabetically slightly_smiling_facesee_no_evil

yes, thats what christoph proposed and you did not object on your comment. Changing it again then

raimund-schluessler:
> > @GretaD You put it in `devDependencies` now. I think this will not work. And please order it alphabetically slightly_smiling_facesee_no_evil
> 
> 
> 
> yes, thats what christoph proposed and you did not object on your comment. Changing it again then

Ah, yes, sorry. I missed this. I think a development dependency will not work as they are afaik not installed for the peer. Did you try it with a dev dep and it worked?

GretaD:
> > > @GretaD You put it in `devDependencies` now. I think this will not work. And please order it alphabetically slightly_smiling_facesee_no_evil
> > 
> > 
> > yes, thats what christoph proposed and you did not object on your comment. Changing it again then
> 
> Ah, yes, sorry. I missed this. I think a development dependency will not work as they are afaik not installed for the peer. Did you try it with a dev dep and it worked?

no i didnt test it on devDep, sorry :see_no_evil:

raimund-schluessler:
Then let's hope adding it as a dependency works :slightly_smiling_face: :see_no_evil:

raimund-schluessler (approved):

Issue #3753, opened 2023-02-14
`npm run styleguide` broken


raimund-schluessler:
Running `npm run styleguide` on latest master gives this error at the moment:

```
 FAIL  Failed to compile

Module not found: Error: Can't resolve 'process/browser' in '/home/raimund/nextcloud/apps2/nextcloud-vue/node_modules/@nextcloud/dialogs/dist'
Did you mean 'browser.js'?
BREAKING CHANGE: The request 'process/browser' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
Cannot read properties of undefined (reading 'module')
```

I suspect this is related to the update of `@nextcloud/dialogs` from 3.2.0 to 4.0.0 in #3746.
@skjnldsv @Pytal @susnux Does one of you have an idea what causes this (and ideally how to fix it :wink: )?

PR #3754, opened 2023-02-14
fix: Add a webpackrule to resolve module js files in `node_modules`

susnux:
* Resolves: #3753 

This should probably go into the webpack config project, but until it is fixed there, we should fix it here.

Somewhere in a module js (ESM) file, probably by webpack-node-polyfills, `process/browser` without the required `.js` extension is imported. We do not use babel-loader on the node_modules and the default loader defaults to `resolveFullySpecified = true`, so this causes an issue. See also https://webpack.js.org/configuration/module/#resolvefullyspecified

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -110,6 +110,15 @@ webpackRules.RULE_RAW_SVG = {
 	type: 'asset/source',
 }
 
+webpackRules.RULE_NODE_MJS = {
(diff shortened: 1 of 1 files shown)


Pytal (approved):

raimund-schluessler (approved):
@susnux Awesome, thanks a lot! The local styleguide build works again.

Issue #3351, opened 2022-10-07
App navigation missing right border in high-contrast mode


nickvergessen:
Also it feels a bit weird, but the "highlighted" entries seem less highlighted:

Normal | High-contrast
---|---
![Bildschirmfoto vom 2022-10-07 12-30-50](https://user-images.githubusercontent.com/213943/194533736-927c6ca6-1933-4694-9a4d-dbffc4928fe9.png) | ![Bildschirmfoto vom 2022-10-07 12-30-35](https://user-images.githubusercontent.com/213943/194533731-8ff1bd11-5e03-4732-84ce-2be4d4c934ed.png)

PR #3756, opened 2023-02-15
fix: High contrast mode optimization

susnux:
* Resolved: #3351 

Add the missing right border on the app navigation (high contrast mode). Unify design with settings view where the app navigation has a border on the right on high contrast mode.

Changed the background color of the active element on high contrast mode, as the highlighting was barley visible.

Also applied the styles for the NcListItem to have a common styling across apps, as some apps, like the forms app, use list items instead of app navigation items (to allow a second title within the item).

Normal | High-contrast before | High-contrast now|
---|--- | ---
Screenshot_20230215_025927|Screenshot_20230215_030016|Screenshot_20230215_031029

### DIFF
--- a/src/components/NcAppNavigation/NcAppNavigation.vue
+++ b/src/components/NcAppNavigation/NcAppNavigation.vue
@@ -185,6 +185,13 @@ export default {
 	}
 }
 
+// add extra border for high contrast mode
+[data-themes*="highcontrast"] {
+	.app-navigation {
+		border-right: 1px solid var(--color-border);
+	}
+}
+
 // When on mobile, we make the navigation slide over the appcontent
 @media only screen and (max-width: $breakpoint-mobile) {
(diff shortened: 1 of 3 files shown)


jancborchardt (commented):
Very nice – I would even say your fix could be applied generally, not only for high contrast mode. Then the navigation would always have a slight border to the right, which looks a bit nicer.

For example, it looks especially strange on narrow screens when opening the navigation:

Currently | With your patch
-|-
image|image

It also looks very nice with the general style in bigger desktop view, with the background shining through:
image

What do you think @susnux? :)

nimishavijay (approved):
Agreed with Jan! This definitely makes sense here but we can also do it normally :)

JuliaKirschenheuter (approved):

jancborchardt:
@susnux mind sending another pull request which just adds a border in general? :)

PR #3757, opened 2023-02-15
RichContent autocomplete not required

jotoeri:
Since we don't need autocompletion on forms, we would just pass an empty function (as talk does btw., too :wink:). So why not doing this here.

### DIFF
--- a/src/components/NcRichContenteditable/NcRichContenteditable.vue
+++ b/src/components/NcRichContenteditable/NcRichContenteditable.vue
@@ -176,8 +176,9 @@ export default {
 
 		autoComplete: {
 			type: Function,
-			required: true,
+			default: () => [],
 		},
+
 		menuContainer: {
 			type: Element,
 			default: () => document.body,


src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -176,8 +176,9 @@ export default {
 
 		autoComplete: {
 			type: Function,
-			required: true,
+			default: function(){ return [] },
```
susnux, line 179: Make eslint happy:
```suggestion
			default: () => [],
```

jotoeri: Wouldn't that exectue and give the array instead of the function? 🤔

jotoeri: Seems to work  👍

susnux: Same as your code, but using the function shorthand.
Only for `type = Object` or `Array` the default value function is executed.

jotoeri: Ok. Didn't know about the different behaviour depending on the type. Thanks! 👍

susnux (approved):

raimund-schluessler (approved):

PR #3759, opened 2023-02-15
`NcSelect`: Selected options should not overflow the container

susnux:
If the selected option is longer than the element it should not overflow the NcSelect but use ellipsis.

before | with `min-width: 0` | with hidden disabled search
---|---|---
before|after|after-2

As you can see currently the option overflows the select, note the actions (dropdown toggle) are missing because they are placed outside the select underneath other elements on the right.

After fixing the overflow there is still this weird height issues, which happens because `.vs__selected-options` have `flex-wrap` set and contain the search input even if it is disabled (`searchable=false`).
This can be fixed by removing the search input from the visible dom when not used.

I already submitted a upstream PR, but I doubt that it will be merged soon.

### DIFF
--- a/src/components/NcSelect/NcSelect.vue
+++ b/src/components/NcSelect/NcSelect.vue
@@ -1007,6 +1007,36 @@ body {
 			}
 		}
 	}
+
+	// Hide search from dom if unused to prevent unneeded flex wrap
+	.vs__search[readonly] {
+		position: absolute;
+	}
+	// If search if hidden, ensure that the height of the search is the same
+	.vs__selected-options {
+		min-height: 40px; // 36px search height + 4px search margin
+	}
+
+	/**
+	 * Fix overlow of selected options
+	 * There is an upstream pull request, if it is merged and released remove this fix
+	 * https://github.com/sagalbot/vue-select/pull/1756
+	 */
+	.vs__selected-options {
+		min-width: 0;
+		.vs__selected {
+			min-width: 0;
+		}
+	}
+	&.vs--single {
+		&.vs--loading,
+		&.vs--open {
+			.vs__selected {
+				// Fix `max-width` for `position: absolute`
+				max-width: 100%;
+			}
+		}
+	}
 }
 
 .vs__dropdown-menu {


raimund-schluessler:
Fixing the width issue is fine, but the selected option still overflows:
image

Seems like `.vs__selected` needs a width fix as well.

But, hiding the `input` field prevents the dropdown from closing. It now only closes when you select an option, but not when you click outside anymore. This is an upstream issue: https://github.com/sagalbot/vue-select/issues/1665
As a workaround for the Tasks app, I set `position: absolute;` for the input field:
https://github.com/nextcloud/tasks/blob/205f6c410e30d82e656f6e065c10562473ebc9af/src/components/AppSidebar/MultiselectItem.vue#L162
This doesn't hide it really, but moves it out of the flow.

raimund-schluessler (changes requested):
See https://github.com/nextcloud/nextcloud-vue/pull/3759#issuecomment-1431910791. This breaks closing the dropdown.

susnux:
> Fixing the width issue is fine, but the selected option still overflows :framed_picture: 

This is caused by `position: absolute` added when opening the menu.
I fixed this by setting a `max-width`.

> But, hiding the `input` field prevents the dropdown from closing. It now only closes when you select an option, but not when you click outside anymore.

Ah did not noticed that, thank you! Fixed that as well.
But I also added `visibility: hidden` to prevent it from being announced by screen readers.

raimund-schluessler (approved):

susnux:
PR for the nextcloud fork of the vue-select: https://github.com/nextcloud-deps/vue-select/pull/18

(Forgot that we are using a custom fork and not the vanilla upstream)

mejo- (approved):

src/components/NcSelect/NcSelect.vue
```diff
@@ -1007,6 +1007,36 @@ body {
...
+	 * Fix overlow of selected options
+	 * There is an upstream pull request, if it is merged and released remove this fix
+	 * https://github.com/sagalbot/vue-select/pull/1756
+	 */
+	.vs__selected-options {
+		min-width: 0;
+		.vs__selected {
+			min-width: 0;
+		}
+	}
```
raimund-schluessler, line 1030: I just noticed that these two lines create issues when the `NcSelect` component has `:no-wrap="true"`set. See here:
Screenshot 2023-02-22 at 12-45-19 Nextcloud Vue Style Guide

raimund-schluessler: @susnux I created a new issue for this: https://github.com/nextcloud/nextcloud-vue/issues/3808

mejo-: Mh, shall we revert the PR for this release @raimund-schluessler?

raimund-schluessler: I don't really have a preference here.

mejo-: I don't think there's a need for rush, so let's try to get this solved before releasing 7.7.0. @susnux would you prefer to look for a fix or to revert the PR for 7.7.0?

susnux: I will look for a fix.

Issue #3743, opened 2023-02-12
Use `NcSelect` instead of `NcMultiselect` in the components


raimund-schluessler:
Since `NcMultiselect` got deprecated, we should stop using it in the components itself and use `NcSelect` instead.

Means, we need to migrate here:
- [x] `NcActionInput` #3760
- [x] `NcTimezonePicker` #3781
- [x] `NcSettingsSelectGroup` #4120
https://github.com/nextcloud/nextcloud-vue/blob/4ac3029b10dc54f68a451f3496bbf6bef5272a37/src/components/NcSettingsSelectGroup/NcSettingsSelectGroup.vue#L25-L41
- [x] `NcSearch` #4247
https://github.com/nextcloud/nextcloud-vue/blob/026afe9a4c8959be853cccbef825fe7b6da8398e/src/components/NcRichText/NcReferencePicker/NcSearch.vue#L3-L10
- [x] `NcProviderList` #4247
https://github.com/nextcloud/nextcloud-vue/blob/026afe9a4c8959be853cccbef825fe7b6da8398e/src/components/NcRichText/NcReferencePicker/NcProviderList.vue#L3-L8

@Pytal @susnux Just for reference.

Pytal:
`master` currently targets v7 major, so this breaking change will have to be done for v8

raimund-schluessler:
> `master` currently targets v7 major, so this breaking change will have to be done for v8

I think it doesn't have to be a breaking change if we do it right.

Pytal:
> I think it doesn't have to be a breaking change if we do it right.

Could review a non-breaking PR if you'd be so inclined then :)

raimund-schluessler:
@mejo- or @julien-nc Would one of you have time to move `NcSearch` and `NcProviderList` to use `NcSelect` as well? I think that would be nice, as `NcSelect` is also more accessbile.

julien-nc:
@raimund-schluessler Yes sure, done in #4247

raimund-schluessler:
Closed with #4247.

PR #3760, opened 2023-02-15
Use `NcSelect` for `NcActionInput` type multiselect

raimund-schluessler:
This replaces the deprecated `NcMultiselect` with the new `NcSelect` component for the `NcActionInput` component of type multiselect.

~Although it got better (the dropdown wasn't visible at all before),~ Like before, there currently still is a `z-index` issue for the select dropdown. This is because the `NcActions` component has a `z-index: 100000` and the select dropdown only got `9999`. Fixing this is tricky at the moment, because the dropdown gets appended to body without a custom class that could be targeted, which means we can only set the `z-index` for all select dropdowns globally, which I don't know whether we want this. The proper solution would be to enhance `NcSelect` to allow to set a custom dropdown class, but that needs an upstream fix.

| | Before | After |
|-|-|-|
| select closed | Screenshot 2023-02-15 at 21-48-03 Nextcloud Vue Style Guide | Screenshot 2023-02-15 at 21-47-11 Nextcloud Vue Style Guide |
| select open | image | Screenshot 2023-02-15 at 22-17-25 Nextcloud Vue Style Guide |

I propose we go ahead despite the z-index issue, since it is already better than before.

Fixes 1/3 of #3743.

### DIFF
--- a/src/components/NcActionInput/NcActionInput.vue
+++ b/src/components/NcActionInput/NcActionInput.vue
@@ -115,12 +115,14 @@ For the multiselect component, all events will be passed through. Please see the
 			}"
 			class="action-input"
 			@mouseleave="onLeave">
-			<!-- @slot Manually provide icon -->
-			<slot name="icon">
-				<span :class="[isIconUrl ? 'action-input__icon--url' : icon]"
-					:style="{ backgroundImage: isIconUrl ? `url(${icon})` : null }"
-					class="action-input__icon" />
-			</slot>
+			<span class="action-input__icon-wrapper">
+				<!-- @slot Manually provide icon -->
+				<slot name="icon">
+					<span :class="[isIconUrl ? 'action-input__icon--url' : icon]"
+						:style="{ backgroundImage: isIconUrl ? `url(${icon})` : null }"
+						class="action-input__icon" />
+				</slot>
+			</span>
 
 			<!-- form and input -->
 			<form ref="form"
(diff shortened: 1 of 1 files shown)


susnux:
Does not change much on the functionallity (still that z-index issues), but from the styling it now looks different.
Could we adjust the border color to match up with the other input types?

Same with the border radius, it looks rounder than the other input types, but this was even before so not related to this PR (it just catch my eye).

raimund-schluessler:
> Does not change much on the functionality (still that z-index issues), but from the styling it now looks different. Could we adjust the border color to match up with the other input types?
> 
> Same with the border radius, it looks rounder than the other input types, but this was even before so not related to this PR (it just catch my eye).

I would say, the other inputs need to be adjusted. Ideally, we should use an `NcTextField` here, then the borders and their radii should look the same. I would rather not touch the `NcSelect` styling, though. I can do that in a follow-up PR later.
The z-index problem is annoying, but it still works, since the actions menu simply shows a scrollbar and you can still select the options.

src/components/NcActionInput/NcActionInput.vue
```diff
@@ -406,7 +407,7 @@ $input-margin: 4px;
 
 	font-weight: normal;
 
-	&:deep(.material-design-icon) {
+	&:deep(.material-design-icon:not(.vs__open-indicator)) {
```
Pytal, line 410: Do these styles also incorrectly apply to the clear icon?

Pytal (commented):
ue

src/components/NcActionInput/NcActionInput.vue
raimund-schluessler: Yes, they did. I fixed it.

ChristophWurst (approved):

nickvergessen (approved):

nickvergessen:
Added a doc sample:
grafik

PR #3761, opened 2023-02-15
NcPopover: Fix `setReturnFocus` property

susnux:
Fixed the wrong JSDoc comment :

> Syntax error in type: import('focus-trap').Options['setReturnFocus'] eslint [jsdoc/valid-types](https://github.com/gajus/eslint-plugin-jsdoc#eslint-plugin-jsdoc-rules-valid-types)

Also added a default value to make eslint happy (even if undefined) and added the missing type information:

> Prop 'setReturnFocus' requires default value to be set. eslint: [vue/require-default-prop](https://eslint.vuejs.org/rules/require-default-prop.html)
> Prop "setReturnFocus" should define at least its type. eslint: [vue/require-prop-types](https://eslint.vuejs.org/rules/require-prop-types.html)

### DIFF
--- a/src/components/NcPopover/NcPopover.vue
+++ b/src/components/NcPopover/NcPopover.vue
@@ -126,10 +126,11 @@ export default {
 		/**
 		 * Set element to return focus to after focus trap deactivation
 		 *
-		 * @type {import('focus-trap').Options['setReturnFocus']}
+		 * @type {import('focus-trap').FocusTargetValueOrFalse}
 		 */
 		setReturnFocus: {
-			required: false,
+			default: undefined,
+			type: [Object, String, Function, Boolean],
 		},
 	},
 


Pytal:
🤔 TypeScript syntax in JSDoc should work https://github.com/nextcloud/eslint-config/pull/413

raimund-schluessler (approved):

susnux:
> thinking TypeScript syntax in JSDoc should work

This part of the PR is a workaround for an upstream bug: https://github.com/jsdoc-type-pratt-parser/jsdoc-type-pratt-parser/issues/141

src/components/NcPopover/NcPopover.vue
```diff
@@ -126,10 +126,12 @@ export default {
 		/**
 		 * Set element to return focus to after focus trap deactivation
 		 *
-		 * @type {import('focus-trap').Options['setReturnFocus']}
+		 * @type {import('focus-trap').FocusTargetValueOrFalse}
 		 */
 		setReturnFocus: {
 			required: false,
+			default: undefined,
+			type: [Object, String, Function, Boolean],
```
Pytal, line 134:
```suggestion
			type: [Object, String, Function, Boolean],
			default: undefined,
```
I believe the `required` can be removed now?

susnux: Fixed it

Pytal (approved):

Issue #497, opened 2019-07-21
Modal component doesn't work twice


StCyr:
Hi,

I'm using the Modal component to display attached images in the "social" application.

For some reasons, images aren't displayed when the user tries to display them a second time: the Modal component shows a black screen with only the "x" icon in the top-right corner.

Do i use the component wrong, or is it some kind of limitation?

Thanks for the help,

Cyrille

# Steps to reproduce the issue

1. Install the social app
2. have a timeline containing posts with attachments
3. click on a thumbnail to display an attachment 
4. Confirm the image is shown properly using the Modal component
5. Close the Modal
6. Click the thumbnail again
7. Confirm that the Modal component only displays a black screen

# Additional info

Here's where I use the component:

 https://github.com/nextcloud/social/blob/master/src/components/PostAttachment.vue

skjnldsv:
Any error in logs?

juliushaertl:
Hm, I can reproduce that with the modal example code in https://nextcloud-vue-components.netlify.com/#/Components?id=modal

Seems this is cause by using v-show to trigger the modal, switching to v-if fixes it.

@skjnldsv Any idea why this is an issue?

skjnldsv:
I would say because we do not detect the modal hide/show other than on mounted?
https://github.com/nextcloud/nextcloud-vue/blob/fe4e53f6b1335422571dca579149c7cee2078857/src/components/Modal/Modal.vue#L279

StCyr:
> 
> Seems this is cause by using v-show to trigger the modal, switching to v-if fixes it.
> 

It seems that I can't use $refs anymore when using v-if

```
        <masonry>
                <div v-for="(item, index) in attachments" :key="index">
                        <img :src="OC.generateUrl('/apps/social/document/get/resized?id=' + item.id)" @click="showModal(index)">
                </div>
                <modal v-if="modal" :has-previous="current > 0" :has-next="current < (attachments.length - 1)"
                        size="full" @close="closeModal" @previous="showPrevious"
                        @next="showNext">
                        <div class="modal__content">
                                <canvas ref="modalCanvas" />
                        </div>
                </modal>
        </masonry>
```

I'm suddenly getting `TypeError: "canvas is undefined"`  when I click on a masonry component after changing v-show to v-if...

skjnldsv:
> TypeError: "canvas is undefined"

Well you named your ref "modalCanvas" not "canvas" :p

StCyr:
[...]

PR #3762, opened 2023-02-15
NcModal: Allow modals to be triggered by `v-show`

susnux:
* Resolves: #497

When using `v-show` instead of `v-if` the modal is not removed from DOM when hiding, so the `mounted` hook is not called again and we must use the `updated` hook to reset the `showModal` property to allow the model to be shown multiple times.

### DIFF
--- a/src/components/NcModal/NcModal.vue
+++ b/src/components/NcModal/NcModal.vue
@@ -21,12 +21,15 @@
   -->
 
 <docs>
+For showing the modal you can use either `v-show` or `v-if` on the `NcModal`,
+depending on whether you require the Modal to stay within the DOM or not.
+
 ```vue
 <template>
 	<div>
 		<NcButton @click="showModal">Show Modal</NcButton>
 		<NcModal
-			v-if="modal"
+			v-show="modal"
 			@close="closeModal"
 			size="small"
 			title="Title"
@@ -563,6 +566,9 @@ export default {
 	destroyed() {
 		this.clearFocusTrap()
 		this.$el.remove()
+	},
+	updated() {
+		this.showModal = this.$refs.mask.style.display !== 'none'
 	},
 
 	methods: {


src/components/NcModal/NcModal.vue
```diff
@@ -564,6 +567,9 @@ export default {
 		this.clearFocusTrap()
 		this.$el.remove()
 	},
+	updated() {
+		this.showModal = this.$refs.mask.style.display !== 'none'
```
raimund-schluessler, line 571: This looks terribly wrong 🙈 But it does work, though.

susnux: Thats needed because a `v-show` on the `NcModal` sets both, mask and container, `display: none`.
Closing the modal sets `showModal` to `false`, but there is no place where it is set to `true` again.

So we need to check on the updated hook (when `v-show` is changed) if the mask is currently displayed or not.

raimund-schluessler: I understand why you need to do it, it just doesn't look "vue-ish" to me 😉

Have a look at my approach in #3767.

susnux:
> The PR works, but I wonder whether the "correct" solution would be to add a `show` prop and emit updates to it back to the parent component.

I think this would be the good solution, but would require more rewriting.
Probably the best would be to set it by default to `true` so it is not a breaking change (as then `v-if` works as before).

raimund-schluessler:
> > The PR works, but I wonder whether the "correct" solution would be to add a `show` prop and emit updates to it back to the parent component.
> 
> I think this would be the good solution, but would require more rewriting. Probably the best would be to set it by default to `true` so it is not a breaking change (as then `v-if` works as before).

Checkout https://github.com/nextcloud/nextcloud-vue/pull/3767. It is non-breaking and `v-if` works as before.

susnux:
I think https://github.com/nextcloud/nextcloud-vue/pull/3767 is the best non breaking solution, so I close this

PR #3767, opened 2023-02-16
Allow using v-show for `NcModal`

### LINKED ISSUE #497: Modal component doesn't work twice

raimund-schluessler:
This is an alternative approach to #3762 to fix #497. It is a bit more effort compared to #3762 because we need an additional prop and kind of duplicate the internal `showModal` state with the `show` prop to keep it non-breaking, but it feels more like vue to me.

For the next major version we could make the `show` prop required and get rid of the watcher and the internal `showModal` value, which would simplify it a lot.

### DIFF
--- a/src/components/NcModal/NcModal.vue
+++ b/src/components/NcModal/NcModal.vue
@@ -26,12 +26,13 @@
 	<div>
 		<NcButton @click="showModal">Show Modal</NcButton>
 		<NcModal
-			v-if="modal"
+			v-show="modal"
 			@close="closeModal"
 			size="small"
 			title="Title"
 			:outTransition="true"
 			:hasNext="true"
+			:show.sync="modal"
 			:hasPrevious="true">
 			<div class="modal__content">Hello world</div>
 		</NcModal>
@@ -464,12 +465,22 @@ export default {
 			type: Number,
 			default: 0,
 		},
+
+		/**
+		 * Whether the modal is to be shown.
+		 * Needs to be provided if NcModal is toggled with v-show.
+		 */
+		show: {
+			type: Boolean,
+			default: undefined,
+		},
 	},
 
 	emits: [
 		'previous',
 		'next',
 		'close',
+		'update:show',
 	],
 
 	data() {
@@ -530,6 +541,15 @@ export default {
 				this.focusTrap.updateContainerElements([contentContainer, ...elements])
 			}
 		},
+
+		/**
+		 * Watch the external show prop to update the internal showModal value
+		 *
+		 * @param {boolean} show whether the modal should be shown
+		 */
+		show(show) {
+			this.showModal = show
+		},
 	},
 
 	beforeMount() {
@@ -541,7 +561,11 @@ export default {
 		this.mc.destroy()
 	},
 	mounted() {
-		this.showModal = true
+		// Don't show the modal if the external show prop is exactly false.
+		if (this.show !== false) {
+			this.showModal = true
+			this.$emit('update:show', true)
+		}
 
 		// init clear view
 		this.useFocusTrap()
@@ -595,6 +619,7 @@ export default {
 			// do not fire event if forbidden
 			if (this.canClose) {
 				this.showModal = false
+				this.$emit('update:show', false)
 
 				// delay closing for animation
 				setTimeout(() => {


susnux:
I think we can get rid of the water already, simply replace all usage of `showModal` with the `show` prop, add a `v-show="show"` on the `mask` and set the default to `true`.

Then it should work like before (default open when used with `v-if`)  but also allow using `v-show`.

raimund-schluessler:
> I think we can get rid of the watcher and the internal state already, simply replace all usage of `showModal` with the `show` prop, add a `v-show="show"` on the `mask` and set the default to `true`.
> 
> Then it should work like before (default open when used with `v-if`) but also allow using `v-show`.

Let me try that.

raimund-schluessler:
> I think we can get rid of the watcher and the internal state already, simply replace all usage of `showModal` with the `show` prop, add a `v-show="show"` on the `mask` and set the default to `true`.
> 
> Then it should work like before (default open when used with `v-if`) but also allow using `v-show`.

Sorry, I think this doesn't work, because we have no way of updating the `show` prop internally without making it required. Then it would be breaking.

susnux:
> Sorry, I think this doesn't work, because we have no way of updating the `show` prop internally without making it required. Then it would be breaking.

I tried this, I do not see how this is breaking:
https://github.com/nextcloud/nextcloud-vue/compare/master...fix/ncmodal-show-prop

It works for `v-if` as before and additionally adds support to keep the modal in the dom by setting `show` to `false`.
Or am I missing something?

raimund-schluessler:
> > Sorry, I think this doesn't work, because we have no way of updating the `show` prop internally without making it required. Then it would be breaking.
> 
> I tried this, I do not see how this is breaking: [master...fix/ncmodal-show-prop](https://github.com/nextcloud/nextcloud-vue/compare/master...fix/ncmodal-show-prop)
> 
> It works for `v-if` as before and additionally adds support to keep the modal in the dom by setting `show` to `false`. Or am I missing something?

Mind creating a PR? I assumed the `show` prop wouldn't be updated if there was no `:show.sync="showModal"` set on the component. Then the out-transition would not be triggered when using `v-if`. But even if that's the case, probably no one would notice, since it's anyway unmounted. 🤔

raimund-schluessler:
> > > Sorry, I think this doesn't work, because we have no way of updating the `show` prop internally without making it required. Then it would be breaking.
> > 
> > 
> > I tried this, I do not see how this is breaking: [master...fix/ncmodal-show-prop](https://github.com/nextcloud/nextcloud-vue/compare/master...fix/ncmodal-show-prop)
> > It works for `v-if` as before and additionally adds support to keep the modal in the dom by setting `show` to `false`. Or am I missing something?
> 
> Mind creating a PR? I assumed the `show` prop wouldn't be updated if there was no `:show.sync="showModal"` set on the component. Then the out-transition would not be triggered when using `v-if`. But even if that's the case, probably no one would notice, since it's anyway unmounted. 🤔

@susnux I created a PR with your changes in https://github.com/nextcloud/nextcloud-vue/tree/fix/ncmodal-show-prop and when using `v-if` without `:show.sync="showModal"`, the transitions don't run anymore. I don't know how bad this is, but it could be considered a bug I guess.

susnux:
> the transitions don't run anymore. I don't know how bad this is, but it could be considered a bug I guess.

I noticed that, I am unsure about the badness of that too...
If it is too bad, I would stick with your changes and change the property to required with version 8.

raimund-schluessler:
Superseded by #3769.

Issue #3758, opened 2023-02-15
NcAppDetailsToggle missing with viewport width 1024px


mejo-:
When opening three-pane Nextcloud apps (with `NcAppNavigationList` and `NcAppContentList`) with a viewport width of 1024px, the `NcAppContentList` is inaccessible. It's closed due to https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/NcAppContent/NcAppContent.vue#L384-L386, but the `NcAppDetailsToggle` button is not displayed either for some reason.

When lowering the viewport width to 1023px, the `NcAppDetailsToggle` button becomes visible again.

When raising the viewport width to 1025px, `NcAppContentList` gets expanded again and all is good.

So the problem only exists with a viewport width of exactly 1024px.

This affects at least Mail and Collectives, possibly many other apps as well.

It has been reported originally here: https://github.com/nextcloud/collectives/issues/433

mejo-:
The following to PRs are required to fix this issue:

* https://github.com/nextcloud/nextcloud-vue/pull/3768
* https://github.com/nextcloud/server/pull/36751

PR #3768, opened 2023-02-16
Don't regard 1024px width viewport as mobile

mejo-:
* In IsMobileState.js we treat viewports smaller 1024px as mobile.
* In NcAppContent.vue we treat viewports starting with 1024px as mobile.

This results in weird behaviour with screens that have exactly 1024px width. E.g., on three-pane apps, NcAppContentList is hidden but the NcAppDetailsToggle button is not displayed yet.

This commit changes NcAppContent.vue to treat viewports smaller 1024px as mobile, to be in sync with IsMobileState.js.

This PR accompanies https://github.com/nextcloud/server/pull/36751.
    
Both are required to fix https://github.com/nextcloud/nextcloud-vue/issues/3758.

Fixes: #3758

### DIFF
--- a/src/components/NcAppContent/NcAppContent.vue
+++ b/src/components/NcAppContent/NcAppContent.vue
@@ -381,7 +381,7 @@ export default {
 			position: sticky;
 			top: var(--header-height);
 
-			@media only screen and (max-width: $breakpoint-mobile) {
+			@media only screen and (width < $breakpoint-mobile) {
 				display: none;
 			}
 		}
@@ -389,7 +389,7 @@ export default {
 		&-details {
 			overflow-y: auto;
 
-			@media only screen and (max-width: $breakpoint-mobile) {
+			@media only screen and (width < $breakpoint-mobile) {
 				min-width: 100%;
 			}
 		}


ChristophWurst (approved):
Should fix https://github.com/nextcloud/mail/issues/7990 too

raimund-schluessler:
Shouldn't we also adjust this here
https://github.com/nextcloud/nextcloud-vue/blob/8342cda8a128da49b421bd24edf3b1c928a152c6/src/components/NcAppNavigation/NcAppNavigation.vue#L189-L196
and here
https://github.com/nextcloud/nextcloud-vue/blob/c80a585a2eef2bd1057917271c1ce0d2dabd02a6/src/components/NcModal/NcModal.vue#L936
and here
https://github.com/nextcloud/nextcloud-vue/blob/c53599ed09b68d1e00c134c0fc2405848c162548/src/components/NcAppSidebar/NcAppSidebar.vue#L881

raimund-schluessler:
> Shouldn't we also adjust this here

Let's move on with this and adjust the rest later.

raimund-schluessler (approved):

PR #3769, opened 2023-02-16
Allow using v-show for `NcModal`, fix modal-wrapper transition

### LINKED ISSUE #497: Modal component doesn't work twice

raimund-schluessler:
This PR allows to toggle the modal visibility by setting the `show` prop of the modal, which will keep the modal in the DOM.
It also fixes the transition animation of the modal-wrapper when showing/hiding the modal with `v-if`.
Checkout the docs https://deploy-preview-3769--nextcloud-vue-components.netlify.app/#/Components/NcModal and enjoy a smooth modal wrapper animation 😉 

@susnux I think this is the best version of the solution now, as it allows to use `v-show / :show.sync="modal"` or `v-if`, fixes the transition animations for both and requires no watcher.

Fixes #497.

### DIFF
--- a/src/components/NcModal/NcModal.vue
+++ b/src/components/NcModal/NcModal.vue
@@ -21,12 +21,15 @@
   -->
 
 <docs>
+For showing the modal you can use either `:show.sync="showModal"` or `v-if` on the `NcModal`,
+depending on whether you require the Modal to stay within the DOM or not. Do not mix both, as this will break the out transition animation.
+
 ```vue
 <template>
 	<div>
 		<NcButton @click="showModal">Show Modal</NcButton>
 		<NcModal
-			v-if="modal"
+			:show.sync="modal"
 			@close="closeModal"
 			size="small"
 			title="Title"
@@ -134,8 +137,8 @@ export default {
 <template>
 	<div>
 		<NcButton @click="showModal">Show Modal</NcButton>
-		<NcModal v-if="modal" @close="closeModal" size="small">
-			<NcEmojiPicker container=".modal-mask" @select="select">
+		<NcModal v-if="modal" @close="closeModal" size="small" class="emoji-modal">
+			<NcEmojiPicker container=".emoji-modal" @select="select">
 				<NcButton>Select emoji {{ emoji }}</NcButton>
 			</NcEmojiPicker>
 		</NcModal>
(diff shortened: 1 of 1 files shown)


susnux:
The emojipicker does not work anymore within the modal, there is an error in the console:
> Uncaught (in promise) Error: Your focus-trap must have at least one container with at least one tabbable node in it at all times

raimund-schluessler:
> The emojipicker does not work anymore within the modal, there is an error in the console:
> 
> > Uncaught (in promise) Error: Your focus-trap must have at least one container with at least one tabbable node in it at all times

Weird. I will have a look.

raimund-schluessler:
> The emojipicker does not work anymore within the modal, there is an error in the console:
> 
> > Uncaught (in promise) Error: Your focus-trap must have at least one container with at least one tabbable node in it at all times

The problem is related to the `container` from of `NcEmojiPicker`. Without setting `container=".modal-mask"` it works as expected. I still don't know why, though.

raimund-schluessler:
The problem occurs if there are multiple modals in the DOM. This wasn't really possible before, since we always had to use `v-if`, so other modals got removed from the DOM after being hidden.
Now, with the first modal example using `v-show`, there is another modal in the DOM and setting `container=".modal-mask"` targets this first modal, which causes this error. Using `v-if` for the first modal example as well, makes the emoji picker work. I guess we need to be more specific with setting the correct container now that multiple modals might be in the DOM.

raimund-schluessler:
@susnux Everything should work now.

artonge (approved):

susnux (approved):

PR #3771, opened 2023-02-16
Fix nativeOn is only valid on components error, make rootElement computed

raimund-schluessler:
This fixes the warning
```
[Vue warn]: The .native modifier for v-on is only valid on components but it was used on <button>.
```
which is triggered for `NcButton` currently. This is a problem due to #3726.

### DIFF
--- a/src/components/NcButton/NcButton.vue
+++ b/src/components/NcButton/NcButton.vue
@@ -289,7 +289,7 @@ export default {
 		},
 	},
 
-	methods: {
+	computed: {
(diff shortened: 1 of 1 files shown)


Pytal (approved):
Not directly related but does `rootElement` not work as a computed property?

raimund-schluessler:
> Not directly related but does `rootElement` not work as a computed property?

Yes, it does. I changed it.

mejo- (approved):
Code looks good :relaxed:

PR #3773, opened 2023-02-17
Adjust emits option

raimund-schluessler:
No effect for vue2, required for vue3. Makes the migration in #3692 a tiny bit shorter.

### DIFF
--- a/src/components/NcHeaderMenu/NcHeaderMenu.vue
+++ b/src/components/NcHeaderMenu/NcHeaderMenu.vue
@@ -143,7 +143,9 @@ export default {
 
 	emits: [
 		'close',
+		'closed',
 		'open',
+		'opened',
 		'update:open',
 		'cancel',
 	],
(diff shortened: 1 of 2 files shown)


ShGKme (approved):
It also helps with documentation and IDE. For example, the latest WebStorm EAP adds autocomplete for `$emit`: https://blog.jetbrains.com/webstorm/2023/02/webstorm-2023-1-eap-4/#completion_for_custom_component_events_in_vue

Should we add ESLint rule to catch it?

[vue/require-explicit-emits](https://eslint.vuejs.org/rules/require-explicit-emits.html)

raimund-schluessler:
> Should we add ESLint rule to catch it?
> 
> [vue/require-explicit-emits](https://eslint.vuejs.org/rules/require-explicit-emits.html)

I don't know if it's worth adding the rule for vue2. It will come automatically with the vue3 migration.

ShGKme:
> I don't know if it's worth adding the rule for vue2. It will come automatically with the vue3 migration.

Documentation, IDE support.

This PR adds missed `emits`. ESLint may help not to miss them.

Pytal (approved):

PR #3774, opened 2023-02-17
Use component with :is

raimund-schluessler:
Vue3 limits the use of the custom elements `:is` special attribute to the reserved `<component>` tag only, see https://v3-migration.vuejs.org/breaking-changes/custom-elements-interop.html#customized-built-in-elements
Vue2 does not care. So we migrate already to keep #3692 simpler.

### DIFF
--- a/src/components/NcBreadcrumb/NcBreadcrumb.vue
+++ b/src/components/NcBreadcrumb/NcBreadcrumb.vue
@@ -39,7 +39,7 @@ This component is meant to be used inside a Breadcrumbs component.
 		@dragover.prevent="() => {}"
 		@dragenter="dragEnter"
 		@dragleave="dragLeave">
-		<element :is="tag"
+		<component :is="tag"
 			v-if="(title || icon) && !$slots.default"
 			:exact="exact"
 			:to="to"
@@ -51,7 +51,7 @@ This component is meant to be used inside a Breadcrumbs component.
 				<span v-if="icon" :class="icon" class="icon" />
 				<span v-else>{{ title }}</span>
 			</slot>
-		</element>
+		</component>
 		<NcActions v-if="$slots.default"
 			ref="actions"
 			type="tertiary"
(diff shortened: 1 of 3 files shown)


ShGKme (commented):
I really like this refactoring. Regardless of Vue 3 migration, using the dynamic component seems to be much more explicit, especially when `:is` is bound not directly but by `v-bind="object"`. At the same time, the purpose of the `is` special attribute is combability with templates in HTML documents, which is not the case.

However, this approach is used not only in these two components but in many others in a bit less straight way.

What about refactoring all of them?

## Binding `is` via `v-bind="object"`

### NcListItem

https://github.com/nextcloud/nextcloud-vue/blob/a7050634f805ae55826b504afe970b968b6ff670/src/components/NcListItem/NcListItem.vue#L202-L204

### NcAppNavigationItem

https://github.com/nextcloud/nextcloud-vue/blob/a7050634f805ae55826b504afe970b968b6ff670/src/components/NcAppNavigationItem/NcAppNavigationItem.vue#L212

### NcUserBubble

https://github.com/nextcloud/nextcloud-vue/blob/a7050634f805ae55826b504afe970b968b6ff670/src/components/NcUserBubble/NcUserBubble.vue#L79

## Direct using `is` on a component

### NcUserBubble

https://github.com/nextcloud/nextcloud-vue/blob/a7050634f805ae55826b504afe970b968b6ff670/src/components/NcUserBubble/NcUserBubble.vue#L72

raimund-schluessler:
@ShGKme Thanks for the feedback! Indeed, I missed the `NcUserBubble` component, I will change it there as well.

Regarding `NcAppNavigationItem` and `NcListItem` I would propose to ignore them for now, since they use `router-link` which anyway needs a deprecation migration, which I started in #3775 already.

raimund-schluessler:
> @ShGKme Thanks for the feedback! Indeed, I missed the `NcUserBubble` component, I will change it there as well.

Done.

ShGKme (approved):
✨

susnux (approved):

PR #3781, opened 2023-02-17
Use `NcSelect` for `NcTimezonePicker`

raimund-schluessler:
This PR moves `NcTimezonePicker` from using `NcMultiselect` to `NcSelect`. Since `NcSelect` does not support grouping, I had to work around this a bit, but the outcome is the same as when using `NcMultiselect`. I additionally enhanced the search function a bit, so one can now also search for the continent name and see all timezones e.g. in Europe, and searching for "Europe London", will give the correct timezone.

The styling probably needs a bit of adjustment still, because it currently uses the default `NcSelect` style with a gray/blue border, whereas the current timezone picker has no border. But maybe we want to update the style here and match with the other pickers, so I left it for now.

| Before | After |
|-|-|
| image | image |

Fixes 1/3 of #3743.

### DIFF
--- a/src/components/NcTimezonePicker/NcTimezonePicker.vue
+++ b/src/components/NcTimezonePicker/NcTimezonePicker.vue
@@ -41,17 +41,15 @@ export default {
 </docs>
 
 <template>
-	<NcMultiselect :value="selectedTimezone"
+	<NcSelect :value="selectedTimezone"
 		:options="options"
 		:multiple="false"
-		:group-select="false"
+		:clearable="false"
 		:placeholder="placeholder"
-		group-values="regions"
-		group-label="continent"
-		track-by="timezoneId"
+		:selectable="isSelectable"
+		:filter-by="filterBy"
 		label="label"
-		open-direction="above"
-		@input="change" />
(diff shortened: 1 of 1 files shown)


jancborchardt (changes requested):
Very cool! :) I think the blue border around it is actually nice. 

3 things:
- [ ] Style-wise it coukd look a bit more like the ActionMenu component, with the entries not expanding the whole width, but having a bit of margin and having a birder-radius
- [ ] The contrast of the highlight seems to not be enough for accessibility, here we should also do it however we make the ActionMenu entry hover/focus accessible cc @JuliaKirschenheuter @Pytal
- [ ] Detail: The height of the co tainer should (if possible) always cut the list at ½ of the last entry, so that one entry is cut off and it is obvious to people that the list continues. :)

jancborchardt (approved):
Sounds good, thanks for opening the issue, approved design-wise then. :)

susnux (approved):

raimund-schluessler:
I adjusted the search code a bit, so that the continent labels are shown if one region matches, just like with the old picker.

PR #3784, opened 2023-02-18
Use `NcTextField` for `NcActionInput`

raimund-schluessler:
This PR introduces the `NcTextField` and `NcPasswordField` to the `NcActionInput` component to unify the appearance.

| Before | After |
|-|-|
| Screenshot 2023-02-22 at 10-42-12 Nextcloud Vue Style Guide | Screenshot 2023-02-22 at 10-43-10 Nextcloud Vue Style Guide |

I had to enhance the `NcInputField` component a bit, to allow type `number` and enable adding a class to its input field.

@jancborchardt @nimishavijay Could you especially have a look at the color picker please? Before we used an input field of type color which gave the native colorpicker, now we use the colorpicker from nextcloud/vue.

### DIFF
--- a/src/components/NcActionInput/NcActionInput.vue
+++ b/src/components/NcActionInput/NcActionInput.vue
@@ -184,23 +184,47 @@ For the multiselect component, all events will be passed through. Please see the
 							{{ label }}
 						</label>
 						<div class="action-input__input-container">
-							<input :id="inputId"
-								:type="type"
+							<NcPasswordField v-if="type==='password'"
+								:id="inputId"
 								:value="value"
-								:placeholder="text"
+								:label="text"
 								:disabled="disabled"
-								:aria-label="ariaLabel"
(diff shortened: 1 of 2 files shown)


mejo- (approved):
Code looks good to me :+1:

susnux (approved):
And design looks good! Finally the actions have an (almost) consistent look.

jancborchardt (approved):

jancborchardt (commented):
> > Could you especially have a look at the color picker please? Before we used an input field of type color which gave the native colorpicker, now we use the colorpicker from nextcloud/vue.

Main question here would be if it's still accessible. In other places we tend to go with native solutions since they often are the closest to being accessible. Is it possible to make the element look like it does in your proposal but still use the input type color?

Also cc @Pytal @JuliaKirschenheuter regarding accessibility.

Issue #3808, opened 2023-02-22
`NcSelect` with `:no-wrap="true"` styling broken


raimund-schluessler:
@susnux I just noticed that these lines
https://github.com/nextcloud/nextcloud-vue/blob/9dd000a8f8251dc8fa501f995326411497084bf6/src/components/NcSelect/NcSelect.vue#L1025-L1030
added in #3759 create issues when the `NcSelect` component has `:no-wrap="true"`set. See here:
![Screenshot 2023-02-22 at 12-45-19 Nextcloud Vue Style Guide](https://user-images.githubusercontent.com/2496460/220611641-f2789437-48d1-4850-b4ee-cde8dc4339bd.png)

mejo-:
To discuss this here rather than [in the closed PR](https://github.com/nextcloud/nextcloud-vue/pull/3759#discussion_r1114236353):

I think we should get this sorted out before releasing 7.7.0. @susnux would you prefer to look for a fix or to revert the PR for 7.7.0?

susnux:
Is there a date 7.7.0 is scheduled for? Maybe I find some time today, but no later than tomorrow for looking into this.

mejo-:
> Is there a date 7.7.0 is scheduled for? Maybe I find some time today, but no later than tomorrow for looking into this.

That's perfectly fine, thanks a lot @susnux. Releasing tomorrow should be alright.

PR #3811, opened 2023-02-22
fix(NcSelect): Do not squash selected items with `no-wrap` set

susnux:
* Resolves: #3808 

I think this resolves the issue, does this looks like you would expect it to look with `no-wrap` set?

before | after
---|---
image | image

https://user-images.githubusercontent.com/1855448/220661348-1ed78048-cb97-49f3-ac1e-8bb8a04b47f7.mp4


jancborchardt (approved):
> does this look like you would expect it to look with `no-wrap` set?

Yup, I would say so! Better than truncating each one of the elements. :)

mejo- (approved):

mejo-:
Thanks a lot for the quick fix @susnux :heart:

PR #3813, opened 2023-02-23
Adjust `NcDatetimePicker` styling

### LINKED ISSUE #3809: `NcDatetimePicker`s do not match the consistent NcAction style
With #3784 (thanks to @raimund-schluessler ) almost all `NcAction...` have a consistent style , except the `DatePicker`.

image

* The default `NcDatetimePicker`
  * does not use a material icon so the icon stands out from the other actions.
  * does not have the same border style
  * the right margin of the icon looks smaller than the others (but might be only because of the different icon style)
* The native `NcDatetimePicker`
  * <del>looks like there is too much margin on the right side of the icon (compared to the other icons, like for password input or the submit on text inputs)</del> (not possible, see comments below)

raimund-schluessler:
This PR adjust the `NcDatetimePicker` styling to bring it inline with the other picker and input styles. It also replaces the icons used with material design icons.

| Before | After |
|-|-|
| Screenshot 2023-02-23 at 13-52-58 Nextcloud Vue Style Guide | Screenshot 2023-02-23 at 13-53-23 Nextcloud Vue Style Guide |
| ActionInputBefore | ActionInputAfter |

Closes #3809.

### DIFF
--- a/src/components/NcDatetimePicker/NcDatetimePicker.vue
+++ b/src/components/NcDatetimePicker/NcDatetimePicker.vue
@@ -140,13 +140,16 @@ export default {
 		@select-year="handleSelectYear"
 		@select-month="handleSelectMonth"
 		@update:value="$emit('update:value', value)">
-		<template v-if="showTimezoneSelect" #icon-calendar>
-			<NcPopover :open.sync="showTimezonePopover"
+		<template #icon-calendar>
+			<NcPopover v-if="showTimezoneSelect"
+				:open.sync="showTimezonePopover"
 				open-class="timezone-popover-wrapper">
 				<template #trigger>
-					<button class="datetime-picker-inline-icon icon-timezone icon"
+					<button class="datetime-picker-inline-icon"
 						:class="{'datetime-picker-inline-icon--highlighted': highlightTimezone}"
-						@mousedown.stop.prevent="() => {}" />
+						@mousedown.stop.prevent="() => {}">
+						<Web :size="20" />
+					</button>
 				</template>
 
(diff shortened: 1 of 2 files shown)


mejo- (approved):

susnux (approved):
The border looks darker than the others, but I think this is because of #3814

raimund-schluessler:
> The border looks darker than the others, but I think this is because of #3814

Yes, exactly. I rebased the branch with #3814 in and updated the screenshot. It looks the same now.

jancborchardt:
@raimund-schluessler just a super detail → when used in the action menu, is the text of the datepicker a bit too far down vertically inside the input? And the "calendar" icon on the right of the input is also not the same position for both of the actions for some reason?
image

raimund-schluessler:
> when used in the action menu, is the text of the datepicker a bit too far down vertically inside the input?

Might be, but I think this comes from the datepicker styling. See the standalone datepicker:
Screenshot 2023-02-27 at 11-29-01 Nextcloud Vue Style Guide
But to be honest, I find it difficult to say what's the correct position, so I would leave it up to you to fix it 🙈 

> And the "calendar" icon on the right of the input is also not the same position for both of the actions for some reason?

Yes, it is. And until now, I thought we had no chance to fix this, as it's an icon coming from the browser. But I just noticed it reacts to the padding of the input field. So I will prepare a PR to fix it.

raimund-schluessler:
@jancborchardt See https://github.com/nextcloud/nextcloud-vue/pull/3826.

PR #3816, opened 2023-02-23
Make `NcVNodes` component non-functional

raimund-schluessler:
Functional components are not supported anymore with vue 3, so we make the `VNodes` component non-functional already. The performance impact should be negligible, since the component is only used to show the appsidebartabs icons.


src/components/NcVNodes/NcVNodes.vue
```diff
@@ -1,5 +1,5 @@
 <!--
- - @copyright Copyright (c) 2020 Raimund Schlüßler <raimund.schluessler@mailbox.org>
+ - @copyright Copyright (c) 2022 Raimund Schlüßler <raimund.schluessler@mailbox.org>
```
Pytal, line 2:
```suggestion
 - @copyright Copyright (c) 2023 Raimund Schlüßler <raimund.schluessler@mailbox.org>
```
📅

Pytal (approved):

skjnldsv (approved):
:+1:

PR #3818, opened 2023-02-25
Rounded select dropdown options

Pytal:
Before | After
--- | ---
image | image

### DIFF
--- a/src/components/NcSelect/NcSelect.vue
+++ b/src/components/NcSelect/NcSelect.vue
@@ -1044,6 +1044,7 @@ body {
 
 .vs__dropdown-menu {
 	border-color: var(--color-primary);
+	padding: 4px;
 
 	&--floating {
 		width: max-content;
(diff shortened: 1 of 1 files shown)


raimund-schluessler (approved):
Looks great and works well, with the timezone picker as well:
Screenshot 2023-02-25 at 08-01-50 Nextcloud Vue Style Guide

I just have a minor nitpick :slightly_smiling_face:

src/components/NcSelect/NcSelect.vue
```diff
@@ -1044,6 +1044,7 @@ body {
 
 .vs__dropdown-menu {
 	border-color: var(--color-primary);
+	padding: 5px;
```
raimund-schluessler, line 1047: For the `Actions` component we use a value of `4px`
https://github.com/nextcloud/nextcloud-vue/blob/602c9073bdb3333353744561c897dc3c9d426875/src/components/NcActions/NcActions.vue#L1192

which makes sense, because together with the `border-radius: 6px` used below, it adds up to `var(--border-radius-large) = 10px`.

nimishavijay (approved):
Awesome! Looks way better now and completely agree with the note from @raimund-schluessler, nice catch! :)

jancborchardt (approved):
Beautiful! :) Just see @raimund-schluessler's notes

PR #3827, opened 2023-02-27
Don't fail if `NcButton` default slot contains not only text

### LINKED ISSUE #3824: `NcButton` fails to compile without direct text content in default slot
Issue raised because of computed value `text()` in `NcButton`, which doesn't consider result of computed value `hasText()` properly

Template | Result
--- | ---
image | ---
image | image

### replace [line](https://github.com/nextcloud/nextcloud-vue/blob/2c949c5bde580dabf0acf7250185fa24d80b9cf7/src/components/NcButton/NcButton.vue#L313): 
```
const text = this.$slots.default?.[0]?.text.trim()
```
### with:
```
const text = this.$slots.default?.[0]?.text?.trim()
```

raimund-schluessler:
This fixes a runtime warning if the `default` slot of `NcButton` does not contain text only, e.g. `<p>Test</p>`.

Closes #3824.

### DIFF
--- a/src/components/NcButton/NcButton.vue
+++ b/src/components/NcButton/NcButton.vue
@@ -310,7 +310,7 @@ export default {
 	 * @return {object|undefined} The created VNode
 	 */
 	render(h) {
-		const text = this.$slots.default?.[0]?.text.trim()
+		const text = this.$slots.default?.[0]?.text?.trim?.()
 
 		const hasText = !!text
 		const hasIcon = this.$slots?.icon


artonge (approved):

Antreesy (approved):
trim`?`.() seems like overdo for me, but Ok

raimund-schluessler:
> trim`?`.() seems like overdo for me, but Ok

Yes, maybe. I don't have a strong opinion here, just wanted to play it safe 🙈🙂

PR #3829, opened 2023-02-27
NcModal's left column width increase for french and other "large words" languages

Jerome-Herbinet:

NcModal width increase to compensate NcModal's left column width increase.

Some examples for too thing left column in French :


raimund-schluessler:
@Jerome-Herbinet Can you please remove the changes to `package-lock.json`?

nimishavijay:
Good observation! Checking the width, increasing the left column 300px leaves a lot of space to the right, at least when the language is set to English. Does this PR affect the width for all languages? If yes, then we can go with a smaller number like 256px. If this affects only certain languages then it is all good! :)

raimund-schluessler:
It affects all languages.

jancborchardt (commented):
Would an increase by 50px also be enough? And the additional modal width increase I’d say we leave out?

cc modal & navigation pro @marcoambrosini

marcoambrosini:
How about we make this fullscreen??

Jerome-Herbinet:
@jancborchardt, I tested with only 50px before doing the PR, and IMO, it's not enough ...
But, as @jancborchardt says, we can leave the main modal with unchanged, why not. 
I'm waiting for your instructions.
@raimund-schluessler, I'll remove the mentioned file.

Jerome-Herbinet:
OK @raimund-schluessler ; file deleted with a rebase.

Jerome-Herbinet:
Are you finally waiting for me to do a test with different widths ?

marcoambrosini:
I think that settings are so complex that they deserve the entire screen. There's really no need to truncate anything in my opinion and I can't think of any benefit in showing other parts of the app interface while changing the settings, it's just a distraction. 

https://user-images.githubusercontent.com/26852655/225382948-db9e4c83-0da9-4728-9f2c-d984efcc994d.mov

src/components/NcModal/NcModal.vue
```diff
@@ -936,7 +936,7 @@ export default {
 	&--large {
 		.modal-container {
 			max-width: 90%;
-			width: 900px;
+			width: 1000px;
```
marcoambrosini, line 939:
```suggestion
			width: 900px;
```

src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
```diff
@@ -384,7 +384,7 @@ export default {
 		position: relative;
 	}
 	&__navigation {
-		min-width: 200px;
+		min-width: 300px;
```
marcoambrosini, line 387:
```suggestion
		min-width: 400px;
```

marcoambrosini (changes requested):
After testing it with @nimishavijay we decided that we could go with `400px` for the navigation, which collapses when `isMobile` is true. The width of the content was already too much anyway

Jerome-Herbinet:
> After testing it with @nimishavijay we decided that we could go with `400px` for the navigation, which collapses when `isMobile` is true. The width of the content was already too much anyway

Done

marcoambrosini (approved):

Jerome-Herbinet:
> > Done
> 
> Nice! @Jerome-Herbinet could you please post updated screenshots so it's easier for everyone to review? :)

Hi @nimishavijay, below is the before and after display. 
**However**, I don't understand one thing after accepting @marcoambrosini's changes.
I have, in the "Files changed" tab only the NcAppSettingsDialog.vue file, while there was also the NcModal.vue file changed (missing). Can you tell me what's going on ? Did I forget to do something, or did I do it wrong?

Before : 


After :

jancborchardt (changes requested):
> After testing it with @nimishavijay we decided that we could go with 400px for the navigation

@marcoambrosini @nimishavijay judging from the latest screenshot, this looks way unbalanced, with nav almost as wide as content.

If this is still relevant and not fixed elsewhere, something between 250 and 300px should be enough.

Jerome-Herbinet:
> > After testing it with @nimishavijay we decided that we could go with 400px for the navigation
> 
> @marcoambrosini @nimishavijay judging from the latest screenshot, this looks way unbalanced, with nav almost as wide as content.
> 
> If this is still relevant and not fixed elsewhere, something between 250 and 300px should be enough.

@jancborchardt : 
Reduced to 300px (with of the navigation left column), but, as a reminder, note that global modal width hasn't been compensated.

jancborchardt:
> Reduced to 300px (with of the navigation left column), but, as a reminder, note that global modal width hasn't been compensated.

Seems good to me. What does it mean that it needs to be compensated? cc @marcoambrosini

Jerome-Herbinet:
> > Reduced to 300px (with of the navigation left column), but, as a reminder, note that global modal width hasn't been compensated.
> 
> Seems good to me. What does it mean that it needs to be compensated? cc @marcoambrosini 

I mean that i f you enlarge the left column, after, you have less space (less width) for content. So if the original left column was 200px wide  and that its now 300px wide, we should make the modal 100px larger in order to guaranty that content keepa as much space (width) as before. What do you think ?

ShGKme:
What about making the column (and modal) optionally flexible to fit content? With hard limits. So that in any language, when applicable, the column's width would be enough for the longest option but not more.

szaimen (approved):
Just tested and 300px looks good to me :)

marcoambrosini:
@nimishavijay it's fine to ellipsize user names or email subjects but not something like setting sections. I've never seen that.  I'd advise to wrap the text regardless of the column width we choose.

Jerome-Herbinet:
I just saw that the left column is larger than before, for example in the Talk's settings modal, in Nextcloud 28. I don't think it is due to my PR because it's still not merged. Can I close it ?

ShGKme:
With new `NcDialog` design, navigation now is not limited anymore. It is as wide as needed for the widest item. Though, probably this is also an issue, because in the worth case, there is no place for content at all.

Jerome-Herbinet:
> With new `NcDialog` design, navigation now is not limited anymore. It is as wide as needed for the widest item. Though, probably this is also an issue, because in the worth case, there is no place for content at all.

Wouldn't it be possible to specify a maximum width for the left-hand column?

ShGKme:
> Wouldn't it be possible to specify a maximum width for the left-hand column?

It is possible. It's just another issue.

Jerome-Herbinet:
> > Wouldn't it be possible to specify a maximum width for the left-hand column?
> 
> It is possible. It's just another issue.

@ShGKme I made some tests and couldn't reproduce the issue that you described ; do you have any precise example (screenshot) ?

Jerome-Herbinet:
> image

OK @ShGKme, thanks, it's clear now for me. @jancborchardt is this a bug or a feature ? 
Shouldn't we get a "max-with" coupled with an "ellipsis"  in order to avoid that ?

Jerome-Herbinet:
> > image
> 
> OK @ShGKme, thanks, it's clear now for me. @jancborchardt is this a bug or a feature ? Shouldn't we get a "max-with" coupled with an "ellipsis" in order to avoid that ?

Not that great : 
Capture d’écran du 2024-04-29 15-44-11

jancborchardt:
> > image
> 
> OK @ShGKme, thanks, it's clear now for me. @jancborchardt is this a bug or a feature ? 
> Shouldn't we get a "max-with" coupled with an "ellipsis"  in order to avoid that ?

Yes absolutely, otherwise the interface breaks with just 1 faulty text.

Also, texts and translations need to be kept aptly short. E.g. "Peripheriques" instead of "Choissisez les peripheriques" etc. if that works.

Jerome-Herbinet:
> > > image
> > 
> > 
> > OK @ShGKme, thanks, it's clear now for me. @jancborchardt is this a bug or a feature ?
> > Shouldn't we get a "max-with" coupled with an "ellipsis"  in order to avoid that ?
> 
> Yes absolutely, otherwise the interface breaks with just 1 faulty text.
> 
> Also, texts and translations need to be kept aptly short. E.g. "Peripheriques" instead of "Choissisez les peripheriques" etc. if that works.

@jancborchardt can you test my last commit ?

jancborchardt (commented):
Considering the overall size of the settings window and the idea that we want the wording there to be short, I would go for max-width 200px instead. Otherwise looks fine. :)

Jerome-Herbinet:
@jancborchardt ; I've taken a look at NC 27 to 30, and it seems that the problem has been fixed since NC 28 ; the left panel is larger and the modal is well sized IMO. I close this PR.

PR #3833, opened 2023-02-28
NcSelect: Only hide search input if disabled and an element was selected

susnux:
This fixes a regression where the placeholder text is not shown if `NcSelect` is configured to be not searchable, in this case the input element was hidden but this also hides the placeholder.

### DIFF
--- a/src/components/NcSelect/NcSelect.vue
+++ b/src/components/NcSelect/NcSelect.vue
@@ -1008,13 +1008,14 @@ body {
 		}
 	}
 
-	// Hide search from dom if unused to prevent unneeded flex wrap
-	.vs__search[readonly] {
-		position: absolute;
-	}
-	// If search if hidden, ensure that the height of the search is the same
 	.vs__selected-options {
+		// If search is hidden, ensure that the height of the search is the same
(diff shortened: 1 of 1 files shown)


raimund-schluessler (approved):

src/components/NcSelect/NcSelect.vue
```diff
@@ -1008,13 +1008,14 @@ body {
 		}
 	}
 
-	// Hide search from dom if unused to prevent unneeded flex wrap
-	.vs__search[readonly] {
-		position: absolute;
-	}
-	// If search if hidden, ensure that the height of the search is the same
 	.vs__selected-options {
+		// If search if hidden, ensure that the height of the search is the same
```
Chartman123, line 1012:
```suggestion
		// If search is hidden, ensure that the height of the search is the same
```
Just a little typo here that could be fixed :)

Chartman123 (approved):

src/components/NcSelect/NcSelect.vue
susnux: Fixed :)

PR #3847, opened 2023-03-02
v7.7.3

julien-nc:
Changelog is coming


raimund-schluessler:
@julien-nc This definitely needs to be 7.8.0.

PR #3852, opened 2023-03-03
Display related resources error message in panel

Pytal:
| | Screenshot |
--- | ---
**Before (Top-right toast + empty panel area)** | image image
**After** | image

### DIFF
--- a/src/components/NcRelatedResourcesPanel/NcRelatedResourcesPanel.vue
+++ b/src/components/NcRelatedResourcesPanel/NcRelatedResourcesPanel.vue
@@ -47,7 +47,7 @@ export default {
 	<div v-if="appEnabled && isVisible" class="related-resources">
 		<div class="related-resources__header">
 			<h5>{{ headerTranslated }}</h5>
-			<p>{{ descriptionTranslated }}</p>
+			<p>{{ description }}</p>
 		</div>
 
 		<NcResource v-for="resource in resources"
@@ -64,10 +64,10 @@ export default {
 <script>
 import axios from '@nextcloud/axios'
 import { generateOcsUrl } from '@nextcloud/router'
-import { showError } from '@nextcloud/dialogs'
-import { t } from '../../l10n.js'
 
 import NcResource from './NcResource.vue'
+
+import { t } from '../../l10n.js'
 
 export default {
 	name: 'NcRelatedResourcesPanel',
@@ -103,6 +103,7 @@ export default {
 	},
 
 	emits: [
+		'has-error',
 		'has-resources',
 	],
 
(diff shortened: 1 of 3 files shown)


src/components/NcRelatedResourcesPanel/NcRelatedResourcesPanel.vue
```diff
@@ -103,22 +114,27 @@ export default {
...
 		'has-resources',
 	],
 
 	data() {
 		return {
 			appEnabled: OC?.appswebroots?.related_resources !== undefined,
-			headerTranslated: t('Related resources'),
-			descriptionTranslated: t('Anything shared with the same group of people will show up here'),
+			headerTranslated: this.t('Related resources'),
+			descriptionTranslated: this.t('Anything shared with the same group of people will show up here'),
```
raimund-schluessler, line 125: This breaks the l10n extraction. See above, the strings got removed.

src/components/NcRelatedResourcesPanel/NcResource.vue
```diff
@@ -77,7 +82,7 @@ export default {
 
 	data() {
 		return {
-			labelTranslated: t('Open link to "{resourceTitle}"', { resourceTitle: this.title }),
+			labelTranslated: this.t('Open link to "{resourceTitle}"', { resourceTitle: this.title }),
```
raimund-schluessler, line 85: Please keep it as it is, the l10n extraction does not work for the mixin.

raimund-schluessler (changes requested):
Please see the l10n problems.

artonge (approved):

Pytal:
@nimishavijay @jancborchardt any additional design suggestions?

nimishavijay:
Looks good to me! ccing @jancborchardt to double check the wording :)

jancborchardt (approved):
Looks good to me as well! :)

src/components/NcRelatedResourcesPanel/NcRelatedResourcesPanel.vue
```diff
@@ -103,22 +103,33 @@ export default {
...
-			return !this.loading && this.resources.length > 0
+			if (this.loading) {
+				return false
+			}
+			return this.error ?? this.resources.length > 0
+		},
+
+		description() {
+			if (this.error) {
+				return t('Error getting related resources. Please contact your system administrator if you have any questions.')
```
nickvergessen, line 130: From my side of things this is not correct. When the 400 error occurs with Talks 1-1 chats there is nothing I as sysadmin was able to do about it.

nickvergessen: Further details in the comments over there: https://github.com/nextcloud/spreed/pull/9256#issuecomment-1497089458

Now that logging is added it actually makes sense, but i find it weird that the error is not using the normal showError anymore

PR #3854, opened 2023-03-03
Fix select styles

Pytal:
Ensure that the styles are always applied correctly

Before | After
--- | ---
image | image

> The "Before" image shows the share select in https://github.com/nextcloud/server `master` after opening a file in viewer

### DIFF
--- a/src/components/NcSelect/NcSelect.vue
+++ b/src/components/NcSelect/NcSelect.vue
@@ -1044,29 +1044,30 @@ body {
 }
 
 .vs__dropdown-menu {
-	border-color: var(--color-primary);
-	padding: 4px;
+	border-color: var(--color-primary) !important;
+	padding: 4px !important;
 
 	&--floating {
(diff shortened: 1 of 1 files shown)


susnux (approved):

JuliaKirschenheuter (commented):
should we use `!important` in all places? I can for possible other solution

JuliaKirschenheuter (approved):

skjnldsv (approved):

skjnldsv:
> > should we use `!important` in all places? I can for possible other solution
> 
> Ideally, we should not use it at all. I know it's sometimes necessary, or rather easier, but I would prefer to limit its usage to the places where we have issues. 

We could be using a scoped styling with vdeep. It should have a higher selector level?

susnux:
I agree `!important` should be used rarely, but for the moment I can only think of one possible "bigger" issue with this:
The enforcement of the primary color on the border, as this makes it hard to assign an error state for custom validation of `NcSelect`.

PR #3857, opened 2023-03-04
Don't use deprecated v-deep

raimund-schluessler:
This migrates the usage of the deprecated `v-deep` directive in `NcSearch`.


julien-nc (approved):
Thanks

skjnldsv (approved):

PR #3866, opened 2023-03-06
Allow setting modal size when registering a custom picker component

julien-nc:
Design review outcome: Some custom picker components don't look good in a "large" `NcModal`.

This allows to pass the desired modal size when registering a custom picker component.
This size is then used by `NcReferencePickerModal`.

### DIFF
--- a/src/components/NcRichText/NcReferencePicker/customPickerElements.js
+++ b/src/components/NcRichText/NcReferencePicker/customPickerElements.js
@@ -23,7 +23,15 @@ const isCustomPickerElementRegistered = (id) => {
 	return !!window._vue_richtext_custom_picker_elements[id]
 }
 
-const registerCustomPickerElement = (id, callback, onDestroy = (el) => {}) => {
+const getCustomPickerElementSize = (id) => {
+	const size = window._vue_richtext_custom_picker_elements[id]?.size
+	if (['small', 'normal', 'large', 'full'].includes(size)) {
+		return size
+	}
+	return null
+}
+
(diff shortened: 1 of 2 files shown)


src/components/NcRichText/NcReferencePicker/customPickerElements.js
```diff
@@ -61,4 +70,5 @@ export {
 	renderCustomPickerElement,
 	destroyCustomPickerElement,
 	isCustomPickerElementRegistered,
+	getCustomPickerElementSize,
```
juliushaertl, line 73: Do we need this exported actually?

julien-nc: Yes, to be able to use it in
https://github.com/nextcloud/nextcloud-vue/pull/3866/files#diff-73dea770ee4a8a41a18ccabe45c3e8e129477d246dc93871d950e9240c6f5aafL42

It is not exported in index.js.

juliushaertl (approved):

juliushaertl:
@julien-nc Max/Jonas are out today, do you have an app that can be used for testing this already around so we can pull in someone else for another review today?

Antreesy (approved):
Seems fine by code, but didn't tested it yet

mejo- (approved):
Code changes look good to me. Didn't test locally though.

julien-nc:
@juliushaertl Any app implementing a custom smart picker component:
- integration_giphy
- integration_openstreetmap
- integration_openai
- integration_replicate
- text_templates
- files

PR #3874, opened 2023-03-09
Use render function in NcAppSidebarTabs

raimund-schluessler:
This PR moves the `NcAppSidebarTabs` component to use a render function.


ShGKme:
> This fixes the reactivity issues of the old implementation and prepares the component for vue3

Render functions API was much changed in Vue 3. Maybe leave the template approach and update the communication mechanism? At a glance, the problem is about ContainerComponent-ContentComponents communications. We may use `provide/inject`. Also, it will help to avoid the usage of `$parent` and `querySelector`.

raimund-schluessler:
> > This fixes the reactivity issues of the old implementation and prepares the component for vue3
> 
> Render functions API was much changed in Vue 3.

I know, I adjusted the component for vue3 in #3692 already. This PR here is kind of an (enhanced) backport of my vue3 adjustments. Once this PR here is in, I will adjust my vue3 changes accordingly.

> Maybe leave the template approach and update the communication mechanism? At a glance, the problem is about ContainerComponent-ContentComponents communications.

Well, the PR is here and works, so why not use it?

> We may use `provide/inject`. Also, it will help to avoid the usage of `$parent` and `querySelector`.

`provide/inject` is not reactive in vue2. It should work for vue3, though.

raimund-schluessler:
> In both Vue 3 and Vue 2.7, there is a `computed` function that is recommended to provide reactive values.

I wasn't aware that `computed` exists in vue2. But using it might be a breaking change, since we don't require vue 2.7 at the moment, as far as I know.

susnux:
> since we don't require vue 2.7 at the moment, as far as I know.

We have Vue 2.7 as a dependency for quite some time, so this would not be a breaking change.

raimund-schluessler:
> > since we don't require vue 2.7 at the moment, as far as I know.
> 
> 
> 
> We have Vue 2.7 as a dependency for quite some time, so this would not be a breaking change.

Fine with me. I can check that out in a follow-up PR.

src/components/NcAppSidebar/NcAppSidebarTabs.vue
```diff
@@ -192,89 +111,176 @@ export default {
...
+		 * https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html
 		 */
-		updateTabs() {
-			if (!this.$slots.default) {
-				this.tabs = []
-				return
+		 if (!this.$slots.default) {
+			this.tabs = []
+			return
+		}
```
ShGKme, line 182: This is a rendering side-effect.

ShGKme (commented):
> The reactivity is currently broken. See the adjusted example of adding a tab by `v-if`. It does not work for master, but does here. And simulating reactivity of the children/slots is quite a mess in my opinion.

There is no way reactivity is broken by using a template instead of a render function. After the build, it will be a render function anyway.

In the previous solution, the problem was that the `tabs` property was updated manually according to the `$slots.default` on `mounted` and dummy `children` prop updated. It didn't guarantee the update was handled.

The current solution works because `tabs` is not being updated during the rendering. It works because of a side-effect in rendering, not because it is a render function.

---

# Alternative solutions

## 1. The current solution in a template - call `updateTabs` in the template

This will do the same as what is render function is doing in this PR. But won't require the component's rewriting.

Though this should not happen in Vue, this is a rendering side-effect.

See the tip in [Vue Docs / Guide / Essentials](https://vuejs.org/guide/essentials/template-syntax.html#calling-functions)

## 2. Use the `beforeUpdate` hook

When the slot content changes, the component re-renders, and the `beforeUpdate` hook is triggered. This is the place for updating the state depending on `$slots`.

```js
export default {
  // ...
  beforeUpdate() {
    this.updateTabs()
  }
  // ...
}
```

## 3. Vue-way solution - `provide/inject`

By design in Vue, components should not depend on the slot content. A slot is just a place where the content is rendered by the parent. Slot content is not a prop. (An exception is render-level components like Transition and KeepAlive, but not general UI components)

Instead, `NcAppSidebarTabs` and `NcAppSidebarTab` may communicate by `provide/inject` so that each `NcAppSidebarTabs` will know how many `NcAppSidebarTab` it has not by `$slots` rendering result, but directly from `NcAppSidebarTab`. Or by `$parent`, but it is less implicit and not recommended as a components communication tool.

---

> Furthermore, calling `this.$slots` in vue3 outside a template or render function will yield you a warning about lacking reactivity.

This is not a problem for the Vue-way solution.

raimund-schluessler:
I am not dogmatic about this. If you fix the reactivity and especially make it work for vue 3, I am perfectly fine with that.

src/components/NcAppSidebar/NcAppSidebarTabs.vue
raimund-schluessler: > This is a rendering side-effect.

Depends a bit. `this.tabs` is not used in the render function itself, it's just necessary to activate the correct tab when using the keyboard.

marcoambrosini:
I have to say that I find it extremely tedious to work on components with render functions. If as @ShGKme says there are multiple ways to do this without using one, I'd definitely prefer that.

raimund-schluessler:
> I have to say that I find it extremely tedious to work on components with render functions. If as @ShGKme says there are multiple ways to do this without using one, I'd definitely prefer that.

I understand your preference of templates over render functions. But as long as no one provides a solution without the usage of a render function, especially for the use with vue3, we can either live with the bug persisting, or go with the solution we have at hand :wink: 

I don't really see any sensible reason to not use a render function now when it fixes the issue, and instead leave the problem open, just because we don't like render functions. And if we decide to use a render function now, we can still switch back to a template, once someone cares to fix it this way. Anyway, that's just my personal view.

skjnldsv:
> I have to say that I find it extremely tedious to work on components with render functions. If as @ShGKme says there are multiple ways to do this without using one, I'd definitely prefer that.

Sometimes it will be unavoidable.
Components like those are more and more complex. If we want to keep performances and upgradability to a certain level, then render functions are not easy to get around.

While we benefit from vue sfc and that's the reason we did not went for react, there is also a reason react have this render approach much more standardized :)

I think it's fine for those big components here. I actually find it easier to handle and less hackish with a render function than with a complex sets of templates/computed and watchers

skjnldsv:
> By design in Vue, components should not depend on the slot content. A slot is just a place where the content is rendered by the parent. Slot content is not a prop. (An exception is render-level components like Transition and KeepAlive, but not general UI components)

That's not entirely true.
In fact, this fancy tab info extraction based on slot if heavily inspired by a tutorial one of the Vue dev made a while back :)
Slots should not be altered, but there is no malpractice in extracting data from said slots and using them in the parent component. A lot of the UI libraries you can find out there uses similar approaches to create smart rendering blocks/components. :)

ShGKme:
> That's not entirely true.
In fact, this fancy tab info extraction based on slot if heavily inspired by a tutorial one of the Vue dev made a while back 😉

Sorry, I may not have expressed myself clearly. I don't propose leaving this approach at all. Only using `provide/inject` instead of direct slots content processing.

> I think it's fine for those big components here. I actually find it easier to handle and less hackish with a render function than with complex sets of templates/computed and watchers

If I understand correctly, the main reason to rewrite the component to render the function is to make it more compatible with Vue 3. (Because The slots reactivity problem is solvable by adding this hook)

```js
  beforeUpdate() {
    this.updateTabs()
  }
```

To make this component fully compatible with Vue 3, we may instead use `provide/inject` for communication.

I created this draft of `provide/inject` proposal with a template-based solution: https://github.com/nextcloud/nextcloud-vue/pull/3891, so you can see better what I propose.

In my opinion, for this case, it's easier and not hackish or complex sets of templates/computed and watchers.

@skjnldsv what do you think about this proposed solution?

raimund-schluessler:
Superseded by https://github.com/nextcloud/nextcloud-vue/pull/3891.

PR #3882, opened 2023-03-13
feat(NcRichText): Support icon class for smart picker integrations

nickvergessen:
For https://github.com/nextcloud/spreed/pull/9053

Before | After
---|---
grafik | grafik


julien-nc (approved):
Tested with https://github.com/nextcloud/spreed/pull/9053
Works fine except one class should be adjusted in spreed's search results.

Antreesy (approved):
Should works properly, except from one thing: if we'll have both an `entry.thumbnailUrl` and an `entry.icon` (as a fallback). it could be better to switch their order

Issue #3461, opened 2022-11-10
[7.x] NcAppSidebarTabs prevents conditional NcAppSidebarTab


dartcafe:
If the server responses for the store are delayed, the sidebar tabs will not load anymore, because they seem to be discarded as invalid children. 

I noticed that effect before, but it only occurred ometimes and I did not analyze that until now. Since moving all backend calls to a centralized api layer, this happens regulary (unless the server responds fast enough, as it seems). 

Static loading with `v-show=` does not solve this problem, as far as I could see (tested it, same effect). Not sure, if the static loading is the cause or `v-if=`.

Static loading causes the tabs to load, but because of `v-show`, they stay empty.
![grafik](https://user-images.githubusercontent.com/26707476/201035563-1c893114-ce35-48ac-a66e-851173533cf4.png)

Any idea how to prevent that? **This is a blocker for polls!**

**Wanted Behavior:** 
NcAppSidebarTab is not discarded as invalid child, whan loaded conditionally.

Affected code (partly): 
https://github.com/nextcloud/polls/blob/97509e53a1a487ae3985170061d26df9287b551b/src/js/views/SideBar.vue#L24-L36

https://github.com/nextcloud/polls/blob/97509e53a1a487ae3985170061d26df9287b551b/src/js/views/SideBar.vue#L89-L97

```
Ignoring invalid tab 
Object { tag: "vue-component-67-NcAppSidebarTab", data: {…}, children: undefined, text: undefined, elm: section#tab-comments.app-sidebar__tab, ns: undefined, context: {…}, fnContext: undefined, fnOptions: undefined, fnScopeId: undefined, … }
​asyncFactory: undefined
​asyncMeta: undefined
​children: undefined
​componentInstance: Object { _uid: 117, _isVue: true, __v_skip: true, … }
​componentOptions: Object { Ctor: VueComponent(options)
, propsData: {…}, tag: "NcAppSidebarTab", … }
​context: Object { _uid: 77, _isVue: true, __v_skip: true, … }
​data: Object { attrs: {}, scopedSlots: {…}, on: undefined, … }
​elm: <section id="tab-comments" class="app-sidebar__tab" data-v-268dd44a="" aria-labelledby="comments" tabindex="0" role="tabpanel" data-v-7339f406="" aria-hidden="true">
​fnContext: undefined
​fnOptions: undefined
​fnScopeId: undefined
​isAsyncPlaceholder: false
​isCloned: false
​isComment: false
[...]

PR #3891, opened 2023-03-15
Fix dynamic tabs registration and a11y in NcAppSidebarTabs with provide/inject

ShGKme:
Fixes: #3461

Currently, `NcAppSidebarTabs` gets content passed slot, goes throw all nodes, and detects `NcAppSidebarTab` instances. But it happened once on mounted and does not support tabs update.

In this solution:
1. `NcAppSidebarTabs` provides:
   - `registerTab` method
   - `unregisterTab` method
   - `activeTab` property (replacement for `$parent`), could be reverted
2. All `NcAppSidebarTab` register and unregister themself on `created` and `beforeDestroy`
3. After registration `NcAppSidebarTabs` has the tab's data directly from the `NcAppSidebarTab`s

Sorting works the same way it was before. But there is no warnings if a user passes not-tabs as content.

This solution is fully compatible with Vue 3 as it is and Composition API.

### DIFF
--- a/src/components/NcAppSidebar/NcAppSidebar.vue
+++ b/src/components/NcAppSidebar/NcAppSidebar.vue
@@ -31,22 +31,33 @@ include a standard-header like it's used by the files app.
 
 ```vue
 <template>
-	<NcAppSidebar
-		title="cat-picture.jpg"
-		subtitle="last edited 3 weeks ago">
-		<NcAppSidebarTab name="Settings" id="settings-tab">
-			<template #icon>
-				<Cog :size="20" />
-			</template>
-			Settings tab content
-		</NcAppSidebarTab>
-		<NcAppSidebarTab name="Sharing" id="share-tab">
-			<template #icon>
-				<ShareVariant :size="20" />
-			</template>
-			Sharing tab content
-		</NcAppSidebarTab>
-	</NcAppSidebar>
+	<div>
+		<NcCheckboxRadioSwitch :checked.sync="hideFirstTab">
+			The first tab is hidden
+		</NcCheckboxRadioSwitch>
+		<NcAppSidebar
+			title="cat-picture.jpg"
+			subtitle="last edited 3 weeks ago">
+			<NcAppSidebarTab v-if="!hideFirstTab" name="Fist tab" id="first-tab">
+				<template #icon>
+					<Cog :size="20" />
+				</template>
+				New tab
+			</NcAppSidebarTab>
+			<NcAppSidebarTab name="Settings" id="settings-tab">
+				<template #icon>
+					<Cog :size="20" />
+				</template>
+				Settings tab content
+			</NcAppSidebarTab>
+			<NcAppSidebarTab name="Sharing" id="share-tab">
+				<template #icon>
+					<ShareVariant :size="20" />
+				</template>
+				Sharing tab content
+			</NcAppSidebarTab>
+		</NcAppSidebar>
+	</div>
 </template>
 <script>
 	import Cog from 'vue-material-design-icons/Cog'
@@ -57,7 +68,84 @@ include a standard-header like it's used by the files app.
 			Cog,
 			ShareVariant,
 		},
+		data() {
+			return {
+				hideFirstTab: true,
+			}
+		},
 	}
+</script>
+```
+
+### One tab
+
+```vue
+<template>
+	<div>
+		<NcAppSidebar
+			title="cat-picture.jpg"
+			subtitle="last edited 3 weeks ago">
+			<NcAppSidebarTab name="Settings" id="settings-tab">
+				<template #icon>
+					<Cog :size="20" />
+				</template>
+				New tab
+			</NcAppSidebarTab>
+		</NcAppSidebar>
+	</div>
+</template>
+<script>
+import Cog from 'vue-material-design-icons/Cog'
+
+export default {
+	components: {
+		Cog,
+	},
+}
+</script>
+```
+
+### One or two tabs with condition
+
+```vue
+<template>
+	<div>
+		<NcCheckboxRadioSwitch :checked.sync="hideFirstTab">
+			The first tab is hidden
+		</NcCheckboxRadioSwitch>
+		<NcAppSidebar
+			title="cat-picture.jpg"
+			subtitle="last edited 3 weeks ago">
+			<NcAppSidebarTab v-if="!hideFirstTab" name="Settings" id="settings-tab">
+				<template #icon>
+					<Cog :size="20" />
+				</template>
+				Settings
+			</NcAppSidebarTab>
+			<NcAppSidebarTab name="Sharing" id="share-tab">
+				<template #icon>
+					<ShareVariant :size="20" />
+				</template>
+				Sharing tab content
+			</NcAppSidebarTab>
+		</NcAppSidebar>
+	</div>
+</template>
+<script>
+import Cog from 'vue-material-design-icons/Cog'
+import ShareVariant from 'vue-material-design-icons/ShareVariant'
+
+export default {
+	components: {
+		Cog,
+		ShareVariant,
+	},
+	data() {
+		return {
+			hideFirstTab: true,
+		}
+	},
+}
 </script>
 ```
 
--- a/src/components/NcAppSidebar/NcAppSidebar.vue
+++ b/src/components/NcAppSidebar/NcAppSidebar.vue
@@ -31,53 +31,47 @@ include a standard-header like it's used by the files app.
 
 ```vue
 <template>
-	<div>
-		<NcCheckboxRadioSwitch :checked.sync="hideFirstTab">
-			The first tab is hidden
-		</NcCheckboxRadioSwitch>
-		<NcAppSidebar
-			title="cat-picture.jpg"
-			subtitle="last edited 3 weeks ago">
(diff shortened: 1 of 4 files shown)


src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -57,10 +66,87 @@ include a standard-header like it's used by the files app.
...
+					<Cog :size="20" />
+				</template>
+				New tab
+			</NcAppSidebarTab>
+		</NcAppSidebar>
+	</div>
+</template>
+<script>
+import Cog from 'vue-material-design-icons/Cog'
+import ShareVariant from 'vue-material-design-icons/ShareVariant'
```
raimund-schluessler, line 97:
```suggestion
```
Unused.

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -57,10 +66,87 @@ include a standard-header like it's used by the files app.
...
+
+export default {
+	components: {
+		Cog,
+		ShareVariant,
```
raimund-schluessler, line 102:
```suggestion
```

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -57,10 +66,87 @@ include a standard-header like it's used by the files app.
...
+}
+</script>
+```
+
+### One or two tabs with condition
+
+```vue
+<template>
+	<div>
+		<input type="checkbox" v-model="hideFirstTab"> The first tab is hidden
```
raimund-schluessler, line 113:
```suggestion
		<NcCheckboxRadioSwitch :checked.sync="hideFirstTab">The first tab is hidden</NcCheckboxRadioSwitch>
```

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -31,22 +31,31 @@ include a standard-header like it's used by the files app.
...
-		</NcAppSidebarTab>
-		<NcAppSidebarTab name="Sharing" id="share-tab">
-			<template #icon>
-				<ShareVariant :size="20" />
-			</template>
-			Sharing tab content
-		</NcAppSidebarTab>
-	</NcAppSidebar>
+	<div>
+		<input type="checkbox" v-model="hideFirstTab"> The first tab is hidden
```
raimund-schluessler, line 35:
```suggestion
		<NcCheckboxRadioSwitch :checked.sync="hideFirstTab">The first tab is hidden</NcCheckboxRadioSwitch>
```

src/components/NcAppSidebar/NcAppSidebarTabs.vue
```diff
@@ -216,64 +210,38 @@ export default {
 		 */
 		updateActive() {
 			this.activeTab = this.active
-				&& this.tabs.findIndex(tab => tab.id === this.active) !== -1
+			&& this.tabs.findIndex(tab => tab.id === this.active) !== -1
```
raimund-schluessler, line 213:
```suggestion
			&& this.tabs.some(tab => tab.id === this.active)
```

src/components/NcAppSidebar/NcAppSidebarTabs.vue
```diff
@@ -121,6 +117,16 @@ export default {
 		currentTabIndex() {
 			return this.tabs.findIndex(tab => tab.id === this.activeTab)
```
raimund-schluessler, line 118:
```suggestion
			return this.sortedTabs.findIndex(tab => tab.id === this.activeTab)
```
The keyboard navigation is currently messed up because of this.

raimund-schluessler: It also needs to be adjusted in the `focus*Tab` methods below.

src/components/NcAppSidebarTab/NcAppSidebarTab.vue
```diff
@@ -68,12 +70,19 @@ export default {
 	],
 
 	computed: {
-		// TODO: implement a better way to force pass a prop fromm Sidebar
 		isActive() {
-			return this.$parent.activeTab === this.id
+			return this.activeTab === this.id
```
raimund-schluessler, line 74: :+1:

src/components/NcAppSidebar/NcAppSidebarTabs.vue
```diff
@@ -216,64 +210,38 @@ export default {
...
-			this.tabs = tabs.sort((a, b) => {
-				const orderA = a.order || 0
-				const orderB = b.order || 0
-				if (orderA === orderB) {
-					return OC.Util.naturalSortCompare(a.name, b.name)
-				}
-				return orderA - orderB
-			})
+		registerTab(tab) {
+			this.tabs.push(tab)
```
raimund-schluessler, line 230: We could still validate the tabs before pushing them to the `tabs` array like above:
```js
				if (IsValidString(tab?.name)
					&& IsValidStringWithoutSpaces(tab?.id)
					&& (IsValidStringWithoutSpaces(tab?.icon) || tab?.$slots?.icon))
```

raimund-schluessler (changes requested):
I like the approach and I think that we can waive the warning when tabs and non-tabs are mixed. The keyboard navigation needs fixing, but that's an easy fix.

raimund-schluessler:
It currently does not work in the Tasks app, and I didn't figure out yet why.

raimund-schluessler:
> It currently does not work in the Tasks app, and I didn't figure out yet why.

For some reason, the injected `activeTab` value is not reactive when I test it in the Tasks app. Reverting to `return this.$parent.activeTab === this.id` fixes this. But I have no idea why though, especially since it works in the netlify docs. :shrug:

src/components/NcAppSidebarTab/NcAppSidebarTab.vue
raimund-schluessler: I just realized that this doesn't work in the Tasks app. I have no idea why, though.

src/components/NcAppSidebar/NcAppSidebarTabs.vue
raimund-schluessler: We could do the sorting here already. Then we don't need the computed prop.

raimund-schluessler:
@ShGKme I pushed a few commits to fix some issues that I found, because I wanted to test the changes with my vue3 branch. If you don't like the changes, feel free to discard them.

raimund-schluessler:
I checked that the approach also works for vue3. A few adjustments are of course necessary, but it works fine for vue3 as well, see https://github.com/nextcloud/nextcloud-vue/pull/3692.

src/components/NcAppSidebarTab/NcAppSidebarTab.vue
raimund-schluessler: @ShGKme I think it would be a good idea to revert this change for now. We can still use it when we figure out why it's broken currently. For vue3 it works fine, though.

raimund-schluessler:
@ShGKme Ping :slightly_smiling_face:

ShGKme:
> @ShGKme Ping 🙂

Pong 👀. I'll work on this today.

ShGKme:
Rebased onto master and rewrote WIP commit

src/components/NcAppSidebarTab/NcAppSidebarTab.vue
ShGKme: It could be a problem with getter is old Vue is used. Updated with classic getter instead of `computed`.

ShGKme:
## Finished fix

- Removed mixing tabs and not tabs content test
- Replace `computed` with classic getter in case that was the problem with the Tasks app
- Fix activating tabs after registration/unregistration (active prop in prior)
- `NcAppSidebarTab`'s icon slot rendering method moved to `NcAppSidebarTab` as a public Tab's method
- `NcAppSidebarTab`'s icon slot default content rendering also moved to `NcAppSidebarTab` component (otherwise slot and default slot content were defined in a different components).
- All exposed `NcAppSidebarTab` instance properties defined in `expose`

## Fix a11y

- Set active tab's `tabindex` to `0` instead of `undefined`. It allows to navigate to tabs with keyboard (`Tab`, `SHIFT+Tab`). Previously you could move from tab to tab's content but couldn't move back...
- Handle `Home` and `End` keypress to navigate to the first/last tab according to w3

See: https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html

@raimund-schluessler Don't you know if previous behavior was intended? It wasn't possible to focus tabs navigation with `Tab` of `Shift+Tab`.

Tab

## Updated docs with JSDocs for events and default slots and examples

DynamicTabs

Activating

CustomOrder

ShGKme:
Tested in Talk

TabsTalk

skjnldsv (approved):
Awesome work!

raimund-schluessler (approved):
Works nicely now with Tasks as well!

raimund-schluessler:
Shall we squash the commits or merge as is?

ShGKme:
> Shall we squash the commits or merge as is?

I think, we may squash commits between the first `WIP` and finished fixing issue in `fix dynamic tabs registration`. But the last two other commits are less related to the original problem of the PR: a11y and general docs, I'd keep them separated.

I squashed with `Co-authored-by`, hope GH understands this :D

PR #3893, opened 2023-03-15
Replace v-click-outside with vueuse

susnux:
Replace `v-click-outside` with directive from `VueUse`:
* VueUse supports both vue2 and vue3
* VueUse is actively maintained
* It supports also query selectors instead of only class names to be ignored

Also dropped the directive form files where it is not used.

We can also reuse this dependency for #3894


raimund-schluessler:
This will be a breaking change if we remove/change the mixin, no?

susnux:
> This will be a breaking change if we remove/change the mixin, no?

It depends how you define it, I would the mixins are part of the not documented API and as such part of the private API.
So changing private API should be ok with a minor version change, no?

src/components/NcHeaderMenu/NcHeaderMenu.vue
```diff
@@ -155,14 +155,18 @@ export default {
...
-				middleware: this.clickOutsideMiddleware,
-			},
 		}
 	},
 
+	computed: {
+		clickOutsideConfig() {
+			return [
+				this.closeMenu,
+				this.clickOutsideOptions
```
raimund-schluessler, line 165: Missing comma.
```suggestion
				this.clickOutsideOptions,
```

raimund-schluessler:
> > This will be a breaking change if we remove/change the mixin, no?
> 
> It depends how you define it, I would the mixins are part of the not documented API and as such part of the private API. So changing private API should be ok with a minor version change, no?

Seems to be alright. The respective mixin is not used, at least in the nextcloud org (besides in nextcloud/vue):
https://github.com/search?q=org%3Anextcloud%20excludeClickOutsideClasses&type=code

Pytal (approved):

src/components/NcHeaderMenu/NcHeaderMenu.vue
susnux: fixed

susnux:
Rebased

PR #3895, opened 2023-03-15
feat: Add slot for description of `NcEmptyContent`

susnux:
* Follow up on #3791

Like the `title` this allows to override the default behavior of the description by providing a slot for it.
This way you could pass in formatted content (e.g. using emphasis / strong emphasis ...).

Note: Maybe move the slot inside the paragraph element?

### DIFF
--- a/src/components/NcEmptyContent/NcEmptyContent.vue
+++ b/src/components/NcEmptyContent/NcEmptyContent.vue
@@ -48,7 +48,8 @@ export default {
 ```
 
 You can also customize the name using the `#name` slot
-and add actions.
+and add actions. But to keep the style consistent across Nextcloud
+consider only using header elements as the root elements for the name slot.
 
 ```
 <template>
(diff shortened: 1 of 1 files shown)


raimund-schluessler (approved):
Fine with me, although I am a bit skeptical whether we should allow this much adjustments, or rather keep it consistent.

susnux:
I know what you mean, maybe we should move the slot inside the `<p>` to restrict the content to inline elements.

And maybe the title should (have been) restricted to a prop accepting only `h` elements for different heading levels but no custom elements?

raimund-schluessler:
> I know what you mean, maybe we should move the slot inside the `<p>` to restrict the content to inline elements.
> 
> 
> 
> And maybe the title should (have been) restricted to a prop accepting only `h` elements for different heading levels but no custom elements?

I think this would be to much coding effort to restrict it. Maybe we simply give a hint in the description of the slot.

max-nextcloud (approved):
Ok with me if there's a need for this.

susnux:
Rebased this and added the discussed documentation comments and moved the slot inside the paragraph element.

Issue #3719, opened 2023-02-05
Actions are missing when using the Composition API


HassoSigbjoernson:
There are no errors, the components are just missing. The rest appears to work normally.

I'm using vue 2.7, which supports the Composition API by default, and nextcloud-vue 7.5.

**App.vue (Options API) - actions present**
```
<template>
	<NcContent app-name="vueexample">
		<NcAppNavigation>
			<template #list>
				<NcAppNavigationItem icon="icon-user" title="This is a user">
					<template slot="actions">
						<NcActionButton icon="icon-edit" @click="alert('Edit')">
							Edit
						</NcActionButton>
					</template>
				</NcAppNavigationItem>
			</template>
		</NcAppNavigation>

		<NcAppContent>
			<div>This is the content</div>
			
		<NcActions>
		    <NcActionButton>
			    Delete
	    		<template #icon>
				    <Delete :size="20" />
			    </template>
		    </NcActionButton>
		</NcActions>
	
		</NcAppContent>
	</NcContent>
</template>

<script>
import NcContent from '@nextcloud/vue/dist/Components/NcContent'
import NcAppContent from '@nextcloud/vue/dist/Components/NcAppContent'
import NcAppNavigation from '@nextcloud/vue/dist/Components/NcAppNavigation'
import NcAppNavigationItem from '@nextcloud/vue/dist/Components/NcAppNavigationItem'
import NcActionButton from '@nextcloud/vue/dist/Components/NcActionButton'
import NcActions from '@nextcloud/vue/dist/Components/NcActions'
import Delete from 'vue-material-design-icons/Delete'
export default {
	name: 'App',
	components: {
		NcContent,
		NcAppContent,
		NcAppNavigation,
		NcAppNavigationItem,
		NcActions,
		NcActionButton,
		Delete
	},
	data() {
		return {
		}
	},
}
</script>
```

**App.vue (Composition API) - actions missing**
```
<template>
	<NcContent app-name="vueexample">
		<NcAppNavigation>
			<template #list>
				<NcAppNavigationItem icon="icon-user" title="This is a user">
					<template slot="actions">
						<NcActionButton icon="icon-edit" @click="alert('Edit')">
							Edit
						</NcActionButton>
					</template>
				</NcAppNavigationItem>
			</template>
		</NcAppNavigation>

		<NcAppContent>
			<div>This is the content</div>
			
		<NcActions>
		    <NcActionButton>
			    Delete
	    		<template #icon>
[...]

PR #3896, opened 2023-03-15
fix: Render actions when using the composition API

susnux:
* Resolves: #3719

The setup function of the composition API does not pass components by name but by instance to the vue render function `h`.
This results in `componentOptions.tag` being undefined, so for this case we need to check the `Ctor` instead.

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -902,7 +902,7 @@ export default {
 		 * become problematic later on.
 		 */
 		const actions = (this.$slots.default || []).filter(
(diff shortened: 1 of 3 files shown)


raimund-schluessler (approved):
Still works as expected, but I didn't check if it fixes the issue.

susnux:
@raimund-schluessler I have added an unit test to ensure this issue is fixed and we do not reintroduce this later as a regression.

skjnldsv (approved):

ShGKme (approved):

PR #3900, opened 2023-03-17
Add accessible search field to emoji picker

Pytal:
Before | After
--- | ---
image | image

### DIFF
--- a/src/components/NcEmojiPicker/NcEmojiPicker.vue
+++ b/src/components/NcEmojiPicker/NcEmojiPicker.vue
@@ -109,12 +109,27 @@ This component allows the user to pick an emoji.
 			:show-preview="showPreview"
 			:title="previewFallbackTitle"
 			v-bind="$attrs"
-			@select="select" />
+			@select="select">
+			<template #searchTemplate="slotProps">
+				<NcTextField ref="search"
+					class="search"
+					:value.sync="search"
+					:label="t('Search')"
+					:label-visible="true"
+					:placeholder="i18n.search"
+					trailing-button-icon="close"
+					:trailing-button-label="t('Clear search')"
+					:show-trailing-button="search !== ''"
+					@trailing-button-click="clearSearch(); slotProps.onSearch(search);"
+					@update:value="slotProps.onSearch(search)" />
+			</template>
+		</Picker>
 	</NcPopover>
 </template>
 
 <script>
 import NcPopover from '../NcPopover/index.js'
+import NcTextField from '../NcTextField/index.js'
 import { t } from '../../l10n.js'
 
 import { Picker, EmojiIndex } from 'emoji-mart-vue-fast'
@@ -123,8 +138,9 @@ import data from 'emoji-mart-vue-fast/data/all.json'
 export default {
 	name: 'NcEmojiPicker',
 	components: {
-		Picker,
 		NcPopover,
+		NcTextField,
+		Picker,
 	},
 	props: {
 		/**
@@ -178,8 +194,9 @@ export default {
 	data() {
 		return {
 			emojiIndex: new EmojiIndex(data),
+			search: '',
 			i18n: {
-				search: t('Search'),
+				search: t('Search emoji'),
 				notfound: t('No emoji found'),
 				categories: {
 					search: t('Search results'),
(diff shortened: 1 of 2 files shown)


szaimen (changes requested):
clicking the x on the right doesnt seem to reset the search results.
image

Pytal:
> What I find a bit confusing is that the input and the first emoji seems to be selected at the same time which is obviously not the case. Otherwise works fine now :)

Seems like this is done in `emoji-mart-vue-fast` itself and the idea was to allow navigating the emojis while still focused onthe input? But this only works when using the built-in search input and not when overriding it

raimund-schluessler:
> > What I find a bit confusing is that the input and the first emoji seems to be selected at the same time which is obviously not the case. Otherwise works fine now :)
> 
> Seems like this is done in `emoji-mart-vue-fast` itself and the idea was to allow navigating the emojis while still focused onthe input? But this only works when using the built-in search input and not when overriding it

This behaviour is anyway a bit weird. While searching, `keydown.right` moves the cursor in the input field *and* the emoji selected at the same time. But `keydown.left` *first* moves the emoji selected and *then* the cursor in the input field. Both that it does two things and then inconsistent is unexpected to me.

However, now `keydown.left/right` does nothing anymore, even after one tabbed into the emoji list. Would be nice if this could be changed.

raimund-schluessler:
> However, now `keydown.left/right` does nothing anymore, even after one tabbed into the emoji list. Would be nice if this could be changed.

Might be a limitation of `emoji-mart-vue-fast`. Their example with a custom search behaves the same.

raimund-schluessler (approved):

JuliaKirschenheuter (commented):
As far i can see Emoji Mart Vue has some a11y problems (incl weird arrow navigation behavior) and to fix them is this is beyond our power ;( for now there is too less time to fix such small part (only status message).
Probably a quick / dirty solution would be to take off this part and just leave a usual input field. What do you mean?

Pytal:
> Probably a quick / dirty solution would be to take off this part and just leave a usual input field. What do you mean?

You mean drop these changes and keep the built-in search input?

skjnldsv (approved):

src/components/NcEmojiPicker/NcEmojiPicker.vue
```diff
@@ -109,12 +109,27 @@ This component allows the user to pick an emoji.
...
 			v-bind="$attrs"
-			@select="select" />
+			@select="select">
+			<template #searchTemplate="slotProps">
+				<NcTextField ref="search"
+					class="search"
+					:value.sync="search"
+					:label="t('Search')"
+					:label-visible="true"
+					:placeholder="slotProps.i18n.search"
```
skjnldsv, line 119: do you actually need to use `slotProps` ?
```suggestion
					:placeholder="i18n.search"
```

Pytal: Good point

Now using `i18n` directly while keeping the `slotProps.onSearch()` which is necessary for emoji search https://github.com/serebrov/emoji-mart-vue/blob/a8ac66cffd2bd08f64371add84ea71b7ef620dca/src/utils/picker.js#L97-L104

src/components/NcEmojiPicker/NcEmojiPicker.vue
```diff
@@ -109,12 +109,27 @@ This component allows the user to pick an emoji.
...
+					class="search"
+					:value.sync="search"
+					:label="t('Search')"
+					:label-visible="true"
+					:placeholder="i18n.search"
+					trailing-button-icon="close"
+					:trailing-button-label="t('Clear search')"
+					:show-trailing-button="search !== ''"
+					@trailing-button-click="clearSearch(); slotProps.onSearch(search);"
+					@update:value="slotProps.onSearch(search)" />
```
raimund-schluessler, line 124: Shall we maybe also show a `Magnify` icon to indicate the field is for searching?
```suggestion
					@update:value="slotProps.onSearch(search)">
					<Magnify :size="16" />
				</NcTextField>
```

JuliaKirschenheuter:
> You mean drop these changes and keep the built-in search input?

No, of coarse not. Each improvement is very welcome! I mean that we have to do a review of emoji pickershortly before next a11y review and if emoji picker would have bigger a11y issues from there side than we could "take it away" for some version if fixes would be very height. Of coarse we can discuss any (other) conceivable solution with Michael.

JuliaKirschenheuter (approved):

src/components/NcEmojiPicker/NcEmojiPicker.vue
Pytal: Was considered, though the other search fields, e.g. unified search and contacts search, do not have a magnify icon so will keep it consistent here :)

PR #3907, opened 2023-03-20
Fix missing NcVNodes import in NcListItem

julien-nc:
@raimund-schluessler When using `NcListItem` in https://github.com/julien-nc/integration_miro/blob/main/src/components/talk/SendModal.vue#L31-L49 we get this error:

image

This appeared with `@nextcloud/vue` 7.8.1.

If I'm right about the fix, this PR might need some more changes to deal with the undefined `router-link` in `<component :is="to ? 'router-link' : 'NcVNodes'"`

### DIFF
--- a/src/components/NcListItem/NcListItem.vue
+++ b/src/components/NcListItem/NcListItem.vue
@@ -304,6 +304,7 @@
 <script>
 import NcActions from '../NcActions/index.js'
 import NcCounterBubble from '../NcCounterBubble/index.js'
+import NcVNodes from '../NcVNodes/index.js'
 import { t } from '../../l10n.js'
 
 export default {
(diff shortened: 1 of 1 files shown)


raimund-schluessler (approved):

julien-nc:
@raimund-schluessler Thanks for the review. How about `router-link`? Should it be imported? If so, how?

raimund-schluessler:
> This appeared with `@nextcloud/vue` 7.8.1.

You are right, I forgot to import it, and didn't notice it because in the docs all `Nc*` components are globally imported :see_no_evil: 

> If I'm right about the fix, this PR might need some more changes to deal with the undefined `router-link` in `<component :is="to ? 'router-link' : 'NcVNodes'"`

I don't think so. `router-link` is a built-in component from `vue-router`. I think it doesn't need to be imported individually.

PR #3915, opened 2023-03-23
fix(NcRichContenteditable): Fix pasting XML and HTML content

### LINKED ISSUE #3264: NcRichContentEditable can not handle HTML a like input
Open the sample
https://nextcloud-vue-components.netlify.app/#/Components/NcRichContenteditable

Enter `3 < 4` and check the posted content below and in the other form

Bildschirmfoto vom 2022-09-16 16-15-55

Also the case for comments on the files app and talk descriptions (as reported in https://github.com/nextcloud/spreed/issues/7309 )

nickvergessen:
Fix #3264 
Fix https://github.com/nextcloud/spreed/issues/9104

### DIFF
--- a/src/components/NcRichContenteditable/NcRichContenteditable.vue
+++ b/src/components/NcRichContenteditable/NcRichContenteditable.vue
@@ -482,16 +482,16 @@ export default {
 				return
 			}
 
-			const html = clipboardData.getData('text')
+			const text = clipboardData.getData('text')
 			const selection = window.getSelection()
 
 			// If no selection, replace the whole data
 			if (!selection.rangeCount) {
-				this.updateValue(html)
+				this.updateValue(text)
+				return
 			}
 
 			// Generate text and insert
-			const text = this.parseContent(html)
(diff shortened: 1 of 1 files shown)


src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -488,13 +488,13 @@ export default {
...
 			if (!selection.rangeCount) {
 				this.updateValue(html)
+				return
 			}
 
 			// Generate text and insert
-			const text = this.parseContent(html)
 			const range = selection.getRangeAt(0)
 			selection.deleteFromDocument()
-			range.insertNode(document.createTextNode(text))
```
skjnldsv, line 497: No, this is a security issue.
Parsing the content ensure we are only rendering what we support. It's been a while, but I do remember this is very important to keep as is.

skjnldsv (changes requested):

julien-nc (approved):
Seems safe to me too.

skjnldsv (approved):

PR #3916, opened 2023-03-23
fix(NcRichContenteditable): Fix tribute opening and eating input after pasting text

### LINKED ISSUE #3913: Text eating on shift+enter after pasting url (this twice) when post started with a mention
When writing a chat message that starts with a mention, all written text apart of the mentions is being erased after twice pasting a URL AND hitting shift+enter. Please see this video for better understanding:

https://user-images.githubusercontent.com/2184312/226970275-5a2d93aa-5e97-4ae4-9869-aa26288d0018.mp4

You see after the pasting the second link and hitting shift+enter, the text disappeared, leaving only the mention.

cc @nickvergessen

nickvergessen:
Pasting content will pop up tribute automatically,
when the message starts with a trigger

	: for emojis
	@ for mentions
	/ for link picker

So we detach tribute before pasting and then
attach it again afterwards, so the opened tribute
doesn't eat the rest of your message until you press
Enter and just "autocomplete" the tribute that has
been actively hanging in the top left of the screen

Fix #3913 

Based on top of #3915 as it is in the same lines

### DIFF
--- a/src/components/NcRichContenteditable/NcRichContenteditable.vue
+++ b/src/components/NcRichContenteditable/NcRichContenteditable.vue
@@ -388,17 +388,13 @@ export default {
 
 	mounted() {
 		this.autocompleteTribute = new Tribute(this.autocompleteOptions)
-		this.autocompleteTribute.attach(this.$el)
-
 		if (this.emojiAutocomplete) {
 			this.emojiTribute = new Tribute(this.emojiOptions)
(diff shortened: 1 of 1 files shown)


julien-nc:
Works fine. Ready to approve after a decision is take regarding https://github.com/nextcloud/nextcloud-vue/pull/3915/files#r1146092180

nickvergessen:
Seems we found a better solution in #3917

PR #3923, opened 2023-03-23
feat(Emoji): Suggest emojis based on text smiles

mejo-:
See https://github.com/nextcloud/nextcloud-vue/pull/3924#issuecomment-1482504034 for more details

## Todo

* [x] :warning: This PR is untested! Properly test it.

PR #3924, opened 2023-03-24
fix(NcRichContenteditable): Fix tribute emoji complete interfering unexpectedly aka. 🇨🇨 

### LINKED ISSUE #3825: Emoji picker suggestions are unexpected when using ":)" and ":D"
When typing `:)` and `:D`, two very common text emojis, the suggestions by the emoji picker are very off. And when you just press "enter", you get very misleading emojis as a result:

:) | :D 
-|-
image|image

Expected suggestions:
- :) = :slightly_smiling_face: `:slightly_smiling_face:`
- :D = :smile: `:smile:`

nickvergessen:
* Fix #3825 

## 🪦 Cocos-Island Meme

Bildschirmaufzeichnung vom 2023-03-24, 07-16-01

Opposed to #3923 I fixed it in `NcRichContenteditable` so that other components like the emoji picker itself are not interfered and still operate normally.

### DIFF
--- a/src/components/NcRichContenteditable/NcRichContenteditable.vue
+++ b/src/components/NcRichContenteditable/NcRichContenteditable.vue
@@ -247,6 +247,7 @@ export default {
 
 	data() {
 		return {
+			textSmiles: [],
 			tribute: null,
 			autocompleteOptions: {
 				// Allow spaces in the middle of mentions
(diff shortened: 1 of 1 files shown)


src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -273,16 +274,32 @@ export default {
 				// Where to inject the menu popup
 				menuContainer: this.menuContainer,
 				// Popup mention autocompletion templates
-				menuItemTemplate: item => `<span class="tribute-container-emoji__item__emoji">${item.original.native}</span> :${item.original.short_name}`,
+				menuItemTemplate: item => {
+					if (this.textEmojis.includes(item.original)) {
+						return item.original
+					}
```
skjnldsv, line 280:
```suggestion
					// Return the raw string if there is no match
					if (this.textEmojis.includes(item.original)) {
						return item.original
					}
```

src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -388,6 +405,13 @@ export default {
 	},
 
 	mounted() {
+		const emojiCharacters = ['D', 'P', 'S', 'X', ')', '(', '/']
+		this.textEmojis = []
+		emojiCharacters.forEach((char) => {
+			this.textEmojis.push(':' + char)
+			this.textEmojis.push(':-' + char)
+		})
```
skjnldsv, line 413: Please add comments about what's going on here

src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -273,16 +274,32 @@ export default {
...
 					emojiAddRecent(item.original)
 					return item.original.native
 				},
 				// Pass the search results as values
-				values: (text, cb) => cb(emojiSearch(text)),
+				values: (text, cb) => {
+					const emojiResults = emojiSearch(text)
+					if (this.textEmojis.includes(':' + text)) {
+						emojiResults.unshift(':' + text)
+					}
```
skjnldsv, line 300: Please add comments about what's going on here

skjnldsv, line 300: Please add comments about what's going on here

src/components/NcRichContenteditable/NcRichContenteditable.vue
skjnldsv, line 413: Please add comments about what's going on here

src/components/NcRichContenteditable/NcRichContenteditable.vue
skjnldsv, line 280:
```suggestion
					// Return the raw string if there is no match
					if (this.textEmojis.includes(item.original)) {
						return item.original
					}
```

skjnldsv (commented):
See comments

src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -388,6 +414,17 @@ export default {
 	},
 
 	mounted() {
+		/**
+		 * Populate the list of text smiles we want to offer via Tribute.
+		 * We add the colon `:)` and colon-dash `:-)` version for each of them.
+		 */
+		const smilesCharacters = ['D', 'P', 'S', 'X', ')', '(', '/']
```
ChristophWurst, line 421: what about lowercase p and s? :p

nickvergessen: done

skjnldsv (approved):
Sure!

ChristophWurst (approved):
Looks good

mejo- (changes requested):
To be honest, I think that #3923 is the cleaner approach. It doesn't affect the `EmojiPicker` component either, the search function is only used by emoji autocompletion implementions in `NcRichContenteditable` and Text/Collectives. That's also why I think that it's the proper place to fix this.

mejo- (approved):
Let's get this in nevertheless.

PR #3927, opened 2023-03-25
fix(NcBreadcrumb): cursor and native title

skjnldsv:
Capture d’écran_2023-03-25_12-54-49

When navigating, clicking the last (current) directory should reload.
You can watch the click using `click.native`
```vue
<NcBreadcrumb v-for="(section, index) in sections"
	:key="section.dir"
	:aria-label="ariaLabel(section)"
	:native-title="ariaLabel(section)"
	v-bind="section"
	@click.native="onClick(section.to)">
	<template v-if="index === 0" #icon>
		<Home :size="20" />
	</template>
</NcBreadcrumb>
```


raimund-schluessler:
> When navigating, clicking the last (current) directory should reload. You can watch the click using `click.native`
> 
> ```
> <NcBreadcrumb v-for="(section, index) in sections"
> 	:key="section.dir"
> 	:aria-label="ariaLabel(section)"
> 	:native-title="ariaLabel(section)"
> 	v-bind="section"
> 	@click.native="onClick(section.to)">
> 	<template v-if="index === 0" #icon>
> 		<Home :size="20" />
> 	</template>
> </NcBreadcrumb>
> ```

This does not seem to be related to the changes in this PR :thinking:

raimund-schluessler:
Maybe we should go the same way as in https://github.com/nextcloud/nextcloud-vue/pull/3617 and add a `name` prop as replacement for the `title` prop instead (and make `title` behave as the native title).

skjnldsv:
> This does not seem to be related to the changes in this PR thinking

Nah, it's just a quick heads up for people that didn't know you could watch click on a NcBreadCrumb using vue router.

> Maybe we should go the same way as in #3617 and add a `name` prop as replacement for the `title` prop instead (and make `title` behave as the native title).

Sure, but this is a breaking change :)

raimund-schluessler:
> > Maybe we should go the same way as in #3617 and add a `name` prop as replacement for the `title` prop instead (and make `title` behave as the native title).
> 
> Sure, but this is a breaking change :)

Not if you add a fallback as you did in #3617 😉

skjnldsv:
Done @raimund-schluessler. Given the Breadcrumbs also use the `NcActions`, I had to fix the `NcActionLink` title usage and provided this to all the other `NcActions` too :rocket:

skjnldsv:
Hum, because title is too much bound to the `NcAction*`, I don't think I can achieve that without being a breaking change.
Because it means if I fallback the name from title, then the title will always be defined and the actions will always be seen like this 
image


I'll think about it next week :see_no_evil: 
Sorry for the ping everyone

raimund-schluessler:
> Done @raimund-schluessler. Given the Breadcrumbs also use the `NcActions`, I had to fix the `NcActionLink` title usage and provided this to all the other `NcActions` too rocket

Well, that escalated quickly :see_no_evil: 

> Hum, because title is too much bound to the `NcAction*`, I don't think I can achieve that without being a breaking change. Because it means if I fallback the name from title, then the title will always be defined and the actions will always be seen like this image

I pushed a commit that I think fixes the issue:
grafik


Please have a look and feel free to discard my changes, in case you think they don't work.

raimund-schluessler (approved):

src/components/NcBreadcrumb/NcBreadcrumb.vue
```diff
@@ -171,6 +186,16 @@ export default {
 		}
 	},
 	computed: {
+		/**
+		 * TODO: drop on the 8.0.0 major, see title/name prop
+		 */
+		nameTitleFallback() {
+			if (this.name === null) {
```
raimund-schluessler, line 193: We need to be able to set an empty `name` for the `NcBreadcrumb` that shows the hidden breadcrumbs and only shows an `NcActions` dropdown without name.

src/mixins/actionText.js
```diff
@@ -61,6 +74,17 @@ export default {
 	],
 
 	computed: {
+		/**
+		 * TODO: drop on the 8.0.0 major, see title/name prop
+		 */
+		nameTitleFallback() {
+			if (this.name === null && this.title) {
```
raimund-schluessler, line 81: The `title` prop was not required, so the new `name` prop should not be required either. We only warn if there is a `title` but no `name` (it's of course valid to do so, but we want to draw attention to this change, so some false warnings might be ok).

skjnldsv:
I added a fix for the inline actions rendering text

skjnldsv:
> @skjnldsv I added a commit showing what I mean. As always, discard it if you don't like it slightly_smiling_face

Perfect!!! :heart:

PR #3932, opened 2023-03-27
Fix NcRichText style

julien-nc:
@juliushaertl We missed something when moving `RichText` in here as `NcRichText`.
The `richtext.scss` file is not used anymore. It had to be imported separately when `RichText` was used.
We could now simply import it in `NcRichText`.

@raimund-schluessler No problem with importing it, right?

Another problem raised in https://github.com/nextcloud/spreed/issues/9163 is that links have the `rich-text--component` class and they apparently used to have the `rich-text--external-link` (with `useMarkdown` set to false). This leads to them not being underlined anymore.
@juliushaertl Apparently this appeared after `@nextcloud/vue-richtext` v2.0.4. Does it ring a bell? Any intentional change in the link class?

Fix nextcloud/spreed#9163

### DIFF
--- a/src/components/NcRichText/NcRichText.vue
+++ b/src/components/NcRichText/NcRichText.vue
@@ -238,6 +238,8 @@ export default {
 }
 </script>
 <style scoped>
+@import './richtext.scss';
+
 a:not(.rich-text--component) {
 	text-decoration: underline;
 }
--- a/src/components/NcRichText/NcRichText.vue
+++ b/src/components/NcRichText/NcRichText.vue
@@ -238,6 +238,7 @@ export default {
 }
 </script>
 <style scoped>
+/* stylelint-disable-next-line scss/at-import-partial-extension */
 @import './richtext.scss';
 
 a:not(.rich-text--component) {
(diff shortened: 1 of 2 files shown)


raimund-schluessler:
> @raimund-schluessler No problem with importing it, right?

I think that should be fine.

juliushaertl:
> Another problem raised in https://github.com/nextcloud/spreed/issues/9163 is that links have the rich-text--component class and they apparently used to have the rich-text--external-link (with useMarkdown set to false). This leads to them not being underlined anymore.

I think the class should still be there pulled in from https://github.com/nextcloud/nextcloud-vue/blob/fix/noid/ncrichtext-style/src/components/NcRichText/autolink.js#L69
https://github.com/nextcloud/nextcloud-vue/blob/fix/noid/ncrichtext-style/src/components/NcRichText/autolink.js#L20

>  Apparently this appeared after @nextcloud/vue-richtext v2.0.4. Does it ring a bell? Any intentional change in the link class?

Seems quite odd, nothing I'd be aware of, but might be that another app had the style imported and now that isn't the case anymore?

nickvergessen (approved):
Fixes missing underlining for links to github etc.
But links to 127.0.0.1 are still not clickable :(

Antreesy (approved):

julien-nc:
@juliushaertl Just wanna make sure you agree with this change :grin:. Are there potentially other specific cases (other components) for which the class should not be overridden?

PR #3935, opened 2023-03-28
fix(NcRichText): Match IP addresses as links

nickvergessen:
Fix https://github.com/nextcloud/spreed/issues/9187

Before | After
---|---
Bildschirmfoto vom 2023-03-28 12-31-20 | Bildschirmfoto vom 2023-03-28 12-31-27

### DIFF
--- a/src/components/NcRichText/helpers.js
+++ b/src/components/NcRichText/helpers.js
@@ -1,4 +1,16 @@
+/**
+ * Regex pattern to match links to be resolved by the link-reference provider
+ *
+ * @type {RegExp}
+ */
 export const URL_PATTERN = /(\s|^)(https?:\/\/)((?:[-A-Z0-9+_]+\.)+[-A-Z]+(?:\/[-A-Z0-9+&@#%?=~_|!:,.;()]*)*)(\s|$)/ig
 
-// FIXME See if we can merge those two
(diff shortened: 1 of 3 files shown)


src/components/NcRichText/helpers.js
```diff
@@ -1,4 +1,4 @@
-export const URL_PATTERN = /(\s|^)(https?:\/\/)((?:[-A-Z0-9+_]+\.)+[-A-Z]+(?:\/[-A-Z0-9+&@#%?=~_|!:,.;()]*)*)(\s|$)/ig
+export const URL_PATTERN = /(\s|^)(https?:\/\/)((?:[-A-Z0-9+_]+\.)+[-A-Z0-9]+(?:\/[-A-Z0-9+&@#%?=~_|!:,.;()]*)*)(\s|$)/ig
```
julien-nc, line 1: IMO this should not be changed. local urls should not be resolved, anyway the server reference manager refuses to resolve.

julien-nc (changes requested):

src/components/NcRichText/helpers.js
nickvergessen: This is not about the reference manager. it's the "make clickable".
Also posting a fully public IP is totally fine from my POV.

julien-nc (approved):

PR #3936, opened 2023-03-28
fix(NcButton): ellipsis text on small width

skjnldsv:
| Before | After |
|:---------:|:------:|
|Capture d’écran_2023-03-28_13-17-25|Peek 28-03-2023 13-16|

### DIFF
--- a/src/components/NcButton/NcButton.vue
+++ b/src/components/NcButton/NcButton.vue
@@ -429,7 +429,7 @@ export default {
 	&__wrapper {
 		display: inline-flex;
 		align-items: center;
-		justify-content: space-around;
+		justify-content: center;
 		width: 100%;
 	}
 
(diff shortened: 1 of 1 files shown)


src/components/NcButton/NcButton.vue
```diff
@@ -430,6 +430,7 @@ export default {
 		display: inline-flex;
 		align-items: center;
 		justify-content: space-around;
```
raimund-schluessler, line 432:
```suggestion
		justify-content: center;
```
Otherwise the wide button on the docs looks weird:
grafik

With this change it looks fine:
grafik

skjnldsv: Great catch!

artonge (approved):

PR #3945, opened 2023-04-01
Improve `NcCheckboxRadioSwitch` and use for `NcAppSidebarTabs`

### LINKED ISSUE #603: Sidebar tabbed navigation normal / hover / focus / active states.
I think we should apply the same rules that are going to be implemented in the app navigation since what we have now looks a bit confusing. We came to a conclusion on that here: #425

susnux:
* Resolves #603

I split this into separate commits to make reviewing easier as the PR is quite huge.
The first part adjusts the styling of the checkbox radio switch to be used with for the sidebar tabs (adjust borders and add an icon slot).

### Before
The hover colors are confusing, primary color is applied to multiple elements, the border is only applied to not-selected elements.

https://user-images.githubusercontent.com/1855448/229255077-560d4054-8cbf-4207-a4e5-4902bf887f8d.mp4

<details>

### With improvements (added borders)
The border is applied around the whole group, the hover color is fixed to be primary on the checked element and default hover color otherwise.

https://user-images.githubusercontent.com/1855448/229255140-24e1603a-c0cf-436c-83f8-053bfbe85533.mp4

</details>

### With improvements
The border is applied around the whole group, the hover color is fixed to be primary on the checked element and default hover color otherwise.
Added rounded borders (last commit) to match the mockup in the linked issue. 

https://user-images.githubusercontent.com/1855448/229257472-4cc21a34-1956-414e-b872-4c2ce742f709.mp4

## Part 2: Use `NcCheckboxRadioSwitch` for `NcAppSidebarTabs`

before | after
---|---
image|image

Outdated initial version:
<details>

before | after
---|---
image|image

https://user-images.githubusercontent.com/1855448/229258351-8b5406c9-33ec-4eab-94eb-2de81def30f4.mp4

</details>

### DIFF
--- a/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
+++ b/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
@@ -610,6 +610,9 @@ $spacing: 4px;
 		// remove borders between elements
 		&:not(:last-of-type) {
 			border-bottom: 0!important;
+			.checkbox-radio-switch__label {
+				margin-bottom: 2px;
+			}
 		}
 		&:not(:first-of-type) {
 			border-top: 0!important;
@@ -629,6 +632,9 @@ $spacing: 4px;
 		// remove borders between elements
 		&:not(:last-of-type) {
 			border-right: 0!important;
+			.checkbox-radio-switch__label {
+				margin-right: 2px;
+			}
 		}
 		&:not(:first-of-type) {
 			border-left: 0!important;
--- a/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
+++ b/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
@@ -199,16 +199,16 @@ export default {
 		}"
 		:style="cssVars"
 		class="checkbox-radio-switch">
+		<input :id="id"
+			:checked="isChecked"
+			:disabled="disabled"
+			:indeterminate="indeterminate"
+			:name="name"
+			:type="inputType"
+			:value="value"
+			class="checkbox-radio-switch__input"
+			@change="onToggle">
 		<label :for="id" class="checkbox-radio-switch__label">
-			<input :id="id"
-				:checked="isChecked"
-				:disabled="disabled"
-				:indeterminate="indeterminate"
-				:name="name"
-				:type="inputType"
-				:value="value"
-				class="checkbox-radio-switch__input"
-				@change="onToggle">
 			<NcLoadingIcon v-if="loading" class="checkbox-radio-switch__icon" />
 			<component :is="checkboxRadioIconElement"
 				v-else-if="!buttonVariant"
@@ -507,7 +507,6 @@ $spacing: 4px;
 
 	&__label {
 		display: flex;
-		position: relative;
 		align-items: center;
 		user-select: none;
 		min-height: $clickable-area;
@@ -536,9 +535,14 @@ $spacing: 4px;
 		}
 	}
 
-	&:not(&--disabled) &__label:hover,
-	&:not(&--disabled) &__label:focus-within {
-		background-color: var(--color-primary-light);
+	&:not(&--disabled, &--checked):focus-within &__label,
+	&:not(&--disabled, &--checked) &__label:hover {
+		background-color: var(--color-background-hover);
+	}
+
+	&--checked:not(&--disabled):focus-within &__label,
+	&--checked:not(&--disabled) &__label:hover {
+		background-color: var(--color-primary-light-hover);
 	}
 
 	// Switch specific rules
@@ -571,9 +575,12 @@ $spacing: 4px;
 			border-bottom-right-radius: var(--border-radius-large);
 		}
 
-		// avoid double borders between elements
-		& + &:not(&.checkbox-radio-switch--checked) {
-			border-top: 0;
+		// remove borders between elements
+		&:not(:last-of-type) {
+			border-bottom: 0!important;
+		}
+		&:not(:first-of-type) {
+			border-top: 0!important;
 		}
 		& + &.checkbox-radio-switch--checked {
 			// as the selected element has all borders:
@@ -592,9 +599,12 @@ $spacing: 4px;
 			border-bottom-right-radius: var(--border-radius-large);
 		}
 
-		// avoid double borders between elements
-		& + &:not(&.checkbox-radio-switch--checked) {
-			border-left: 0;
+		// remove borders between elements
+		&:not(:last-of-type) {
+			border-right: 0!important;
+		}
+		&:not(:first-of-type) {
+			border-left: 0!important;
 		}
 		& + &.checkbox-radio-switch--checked {
 			// as the selected element has all borders:
@@ -608,16 +618,15 @@ $spacing: 4px;
 		// better than setting border-radius on labels (producing a small gap)
 		overflow: hidden;
 
+		&:focus-within {
+			border: 2px solid var(--color-primary);
+		}
+
 		&--checked {
 			font-weight: bold;
-			border: 2px solid var(--color-primary-element-light);
-
-			&:hover {
-				border: 2px solid var(--color-primary);
-			}
 
 			label {
-				background-color: var(--color-background-dark);
+				background-color: var(--color-primary-light);
 			}
 		}
 	}
--- a/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
+++ b/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
@@ -2,8 +2,9 @@
   - @copyright Copyright (c) 2021 John Molakvoæ <skjnldsv@protonmail.com>
   -
   - @author John Molakvoæ <skjnldsv@protonmail.com>
+  - @author Ferdinand Thiessen <opensource@fthiessen.de>
   -
-  - @license GNU AGPL version 3 or any later version
+  - @license AGPL-3.0-or-later
   -
   - This program is free software: you can redistribute it and/or modify
   - it under the terms of the GNU Affero General Public License as
@@ -107,7 +108,7 @@ export default {
 			</NcCheckboxRadioSwitch>
 		</div>
 		<h4>Vertically</h4>
-		<div>
+		<div style="width: fit-content">
 			<NcCheckboxRadioSwitch
 				:button-variant="true"
 				:checked.sync="sharingPermission"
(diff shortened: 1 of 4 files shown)


jancborchardt (commented):
Looks much much better!

@susnux a question: On part 1 "with improvements, should the single-line radio buttons not have more border-radius so they are pill-style? (Like we do for the View switcher in Forms currently.)

susnux:
@jancborchardt 
> a question: On part 1 "with improvements, should the single-line radio buttons not have more border-radius so they are pill-style? (Like we do for the View switcher in Forms currently.)

I used your and @nimishavijay comment here as a reference for how it should look like: https://github.com/nextcloud/nextcloud-vue/issues/603#issuecomment-1236878789

jancborchardt:
@susnux ah thank you for the explanation! :) I would just say let’s give it a bit more border-radius so that when it’s just one line it ends up being pill-style like all our buttons (and the current switcher style in Forms), as otherwise it will look out of place?

The radius should be 22px probably, or `calc(var(--default-clickable-area) / 2)`. Then it should be nicely pill-style for single-line buttons and also work well for higher ones like in the sidebar.

susnux:
@jancborchardt 
> I would just say let’s give it a bit more border-radius so that when it’s just one line it ends up being pill-style like all our buttons (and the current switcher style in Forms), as otherwise it will look out of place?

I changed that, it does now look like this:
pill like buttons

raimund-schluessler:
> @jancborchardt
> 
> > I would just say let’s give it a bit more border-radius so that when it’s just one line it ends up being pill-style like all our buttons (and the current switcher style in Forms), as otherwise it will look out of place?
> 
> I changed that, it does now look like this: pill like buttons

I think that is to much. `calc(var(--default-clickable-area) / 2) i.e. 22px` works better, I think. And the outer element needs a radius `2px` larger than the inner one, to account for the border-thickness. Otherwise, there will be gaps at the corners.

src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
```diff
@@ -492,49 +540,83 @@ $spacing: 4px;
...
-	&-switch.checkbox-radio-switch--disabled.checkbox-radio-switch--checked &__icon {
+	&-switch.checkbox-radio-switch--disabled.checkbox-radio-switch--checked &__icon > * {
 		color: var(--color-primary-element-light);
 	}
 
+	&--button-variant {
+		--border-radius-nc-radio-switch: var(--border-radius-large);
+
+		&-h-grouped {
+			--border-radius-nc-radio-switch: var(--border-radius-pill);
```
jancborchardt, line 572: Here I would say to use `calc(var(--default-clickable-area) / 2)` instead.

This will result in a consistent border-radius of 22px.
- Then it looks pill-style for single-line switches like in Forms.
- And for double-line (like with icon above text like in the Files sidebar) it will also look nice and not have a huge radius


@susnux Hope that clears it up, sorry for the confusion. :)

src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
```diff
@@ -545,17 +627,20 @@ $spacing: 4px;
...
-		& + &:not(&.checkbox-radio-switch--checked) {
-			border-left: 0;
+		// remove borders between elements
+		&:not(:last-of-type) {
+			border-right: 0!important;
+		}
+		&:not(:first-of-type) {
+			border-left: 0!important;
 		}
 		& + &.checkbox-radio-switch--checked {
```
raimund-schluessler, line 645: I think this rule needs to go. It makes the center tab move when activating the right one.

src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
jancborchardt: Ah, just saw that @raimund-schluessler already commented the same – that's exactly what I meant, thanks! :)

src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
```diff
@@ -545,17 +627,20 @@ $spacing: 4px;
 
 	&--button-variant-h-grouped {
 		&:first-of-type {
-			border-top-left-radius: var(--border-radius-large);
-			border-bottom-left-radius: var(--border-radius-large);
+			border-top-left-radius: var(--border-radius-nc-radio-switch);
```
raimund-schluessler, line 630: These values need to be `2px` larger than the ones for the inner buttons to account for the border-thickness.

raimund-schluessler: Just for reference, here you see this gap:
image

susnux:
@jancborchardt @raimund-schluessler Do we want the same styling on the vertical group? That would look like this:

image

Or do we want to keep the previous design for the vertical buttons?

image

src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
raimund-schluessler: > The rule below was the problem, as it is now longer needed.

Yeah, I just couldn't comment there as it was not edited in this PR 🙈

src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
susnux: :+1:

src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
susnux: > Then it looks pill-style for single-line switches like in Forms.

BTW we use `--border-radius-pill` on forms too, so this might look a bit different when using this.

susnux:
Ok I fixed the outer border, used the radius for both vertical and horizontal button groups and fixed the inner border margin. I think it is now ready to go.

It looks like this:

https://user-images.githubusercontent.com/1855448/229516798-44b4b9a8-6310-4937-9536-68507c665931.mp4

jancborchardt (approved):
Good stuff, let's go with that! :)

raimund-schluessler:
@susnux works nicely now in the docs. I will test it in a real app tomorrow.
One of the unit tests still needs a fix, though.

skjnldsv:
Capture d’écran_2023-04-04_08-18-04

I find it a bit weird to have those pill rounded borders next to each others :)
I didn't thought that much into it, but couldn't we have straight borders when there is another button next to it and leave the rounded pill for the first and last items?
image

I'll let the design team take the decision, it's juste a thought :)

skjnldsv (approved):
Besides one comment, looks really good! Nice job @susnux !

marcoambrosini (commented):
Hi there, design-wise I would suggest to:

- loose the border
- add a 4px gap between the buttons
- increase the vertical padding around the labels 

https://user-images.githubusercontent.com/26852655/229856459-f4f52857-c92a-4673-bb08-519be7046f0a.mov

susnux:
I agree on the gap, I think it looks better :)

But I personally do not like the missing border, it feels like something is missing visually.

nimishavijay:
Agreed with @susnux that a border looks nice! We should also keep in mind that in some apps there are more than 3 sidebar tabs which would look fairly crowded on small screens (this is a bigger issue, but that's another discussion altogether). So I suggest that we keep the spacing to 2px. What do you think @marcoambrosini ?

jancborchardt:
I think the improved vertical spacing is good @marcoambrosini!

Agree with @susnux the border is needed, especially to cover cases where there are elements other than the radio group horizontally or vertically. E.g. in Forms there is the Share button and sidebar toggle next to it.

Regarding the gap I'm not sure mostly for space reasons. But I'd be fine with anything that's decided. :)

susnux:
Also there is no space for padding, the tabs have to fit into 300px to 500px (our current limits of the sidebar menu).

For a "real" app you will get following on the minimum size and three tabs:

2px gap | 4px gap
---|---
screenshot|screenshot

So I would go along with @nimishavijay on that 2px.

jancborchardt (approved):
Looks great now I think! :)

raimund-schluessler (approved):
Works well with the Tasks app. The failing test still needs a fix, but I have no idea, why it fails, honestly. Seems the `NcCheckboxRadioSwitch` does not emit the `onToggle` event in the tests, but it works well in a real app.

tests/unit/components/NcAppSidebar/NcAppSidebarTabs.spec.js
```diff
@@ -115,22 +115,22 @@ describe('NcAppSidebarTabs.vue', () => {
...
-			it('display all the 3 elements in li', () => {
-				const liList = wrapper.findAll('nav>ul>li')
+			it('display all the 3 elements', () => {
+				const liList = wrapper.findAll('nav>.checkbox-radio-switch')
 				expect(liList.length).toEqual(3)
 			})
 			it('emit "update:active" event with the selected tab id when clicking on a tab', () => {
-				const lastLink = wrapper.find('nav>ul>li:last-of-type>a')
+				const lastLink = wrapper.find('nav>.checkbox-radio-switch:last-of-type>label')
 				lastLink.trigger('click')
```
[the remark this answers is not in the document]
raimund-schluessler: > Maybe `await`ing the trigger fixes the test?

No, it does not :see_no_evil:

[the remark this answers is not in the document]
ShGKme: More info: a click on the `<label>` in `<NcCheckboxRadioSwitch>` triggers a `click` on `<input>`, but does not trigger a `change` event.

Seems to be either jsdom or VTU bug, but all related bug reports are fixed.

*continues investigating*

nimishavijay:
Well, I believe that looking like radio buttons is kind of the idea! :) @ShGKme 

Adding a border makes sure that:
- The items are visually grouped together
- It doesn't look too similar to a button, which is usually used to do some action. This component will be used to switch between multiple views
- The alignment of the item is clear. Without the border, it is difficult to determine where the leftmost item begins, thereby making it look like the item is not aligned neatly with the text above or below it (for example, in Marco's video, it seems like "Search" is not neatly aligned with `cat-picture.jpg / last edited 3 weeks ago`, and adding a border would help)

susnux:
Why I prefer the current style (borders and all round):
* It is consistent with the other styles (we always have rounded elements
* The border groups belonging elements
* It makes clear where the element ends, so you know where to click (I think this is an important point regarding accessibility

raimund-schluessler:
@jancborchardt I think we need a design decision here, and someone who feels confident to press "Merge" :wink:

jancborchardt:
I agree with the points brought up especially regarding accessibility and grouping.

For simplicity, I suggest:
- We go with this design now and merge it, rounded and border around, as it is much better than what is there now and in line with the rest of our design
- If we want to adjust its occurrence in the right sidebar navigation (as tabs) to use no border, we can very simply do that.

So I’ll merge it to unblock the situation. :v:

PR #3958, opened 2023-04-05
Fix : Disable hover behaviour on quota information

Jerome-Herbinet:

Disables background color and pointer cursor while hovering this area with mouse.

### DIFF
--- a/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
+++ b/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
@@ -747,6 +747,18 @@ export default {
 		flex-wrap: wrap;
 		box-sizing: border-box;
 		width: 100%;
+		
+		&.app-navigation-entry__settings-quota {
+			.app-navigation-entry:hover {
+				background-color: transparent;
+				a {
+					cursor: default;
+					* {
+					cursor: default;
+					}
+				}
+			}
+		}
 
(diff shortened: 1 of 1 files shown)


src/components/NcAppNavigationItem/NcAppNavigationItem.vue
```diff
@@ -747,6 +747,18 @@ export default {
 		flex-wrap: wrap;
 		box-sizing: border-box;
 		width: 100%;
+		
+		&.app-navigation-entry__settings-quota {
```
szaimen, line 751: Shouldnt this rather go into thr server repo? Also cc @skjnldsv

szaimen:
> Do I need here to do a "npm run build" ?

No, not needed here

src/components/NcAppNavigationItem/NcAppNavigationItem.vue
nickvergessen: Hmpf, that being said it's broken already in 26

skjnldsv:
Not the right place, besides, the hover is here for a reason.
If you click it refreshes the quota. So there is an action.
Until this is removed (when all views are migrated to vue and we can rely on files events only), the hover should stay :)

Jerome-Herbinet:
> 

OK @skjnldsv , I didn't know that... 
So I have a suggestion: You should at least add a tooltip when you move your mouse over the page, for example: "Refresh" ... because I think there are a lot of people like me who didn't understand the usefulness of this.

PR #3962, opened 2023-04-07
Remove the smart picker modal padding

julien-nc:
* Remove it to let custom picker components set it if needed.
* Adjust the padding of the generic search.
* Make one class more explicit

This allows, for example, custom picker components that have infinite scrolling (like the Giphy integration one) to remove bottom padding/margin which looks better.

As we can't make sure nextcloud/vue is updated everywhere at once and all the providers are adjusted (I'll make sure most of them are :grin: ), we might temporarily have:
* double padding with adjusted providers displayed in apps using an old nextcloud/vue release
* no padding at all for old providers displayed in apps using a nextcloud/vue release which includes those changes

Not that bad IMO.

### DIFF
--- a/src/components/NcRichText/NcReferencePicker/NcReferencePickerModal.vue
+++ b/src/components/NcRichText/NcReferencePicker/NcReferencePickerModal.vue
@@ -153,7 +153,6 @@ export default {
 
 <style lang="scss" scoped>
 .reference-picker-modal--content {
-	padding: 12px 16px 16px 16px;
 	width: 100%;
 	display: flex;
 	flex-direction: column;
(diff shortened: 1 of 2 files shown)


mejo- (approved):
@julien-nc seems sensible to me. Could you provide a (non-comprehensive) list of apps with picker componenents that are not under your umbrella? :grimacing:

julien-nc:
@mejo- Thanks for the review. So far I'm not aware of many providers implementing a custom picker component.
Appart from mines, there is the Files one implemented by @juliushaertl and that's about it I think.

juliushaertl (approved):

PR #3973, opened 2023-04-13
Don't overwrite router-link href in NcBreadcrumb

raimund-schluessler:
We currently wrongfully overwrite the `href` attribute of the `router-link` in `NcBreadcrumb` with the default `href` prop value. This makes the link rendered by `router-link` to not have its `href` attribute correctly set, which prevents the default link behavior. Navigating still works though, so I guess this is why we didn't notice yet. This PR fixes it by only setting the `href` attribute from the prop if it's not a `router-link`.

### DIFF
--- a/src/components/NcBreadcrumb/NcBreadcrumb.vue
+++ b/src/components/NcBreadcrumb/NcBreadcrumb.vue
@@ -42,7 +42,7 @@ This component is meant to be used inside a Breadcrumbs component.
 		<component :is="tag"
 			v-if="(nameTitleFallback || icon) && !$slots.default"
 			:title="title"
-			v-bind="{ ...$attrs, ...to && { to, exact }, ...!to && { href } }"
+			v-bind="linkAttributes"
 			v-on="$listeners">
 			<!-- @slot Slot for passing a material design icon. Precedes the icon and title prop. -->
 			<slot name="icon">
@@ -200,6 +200,14 @@ export default {
 		 */
 		tag() {
 			return this.to ? 'router-link' : 'a'
+		},
+
+		/**
+		 * The attributes to pass to `router-link` or `a`
+		 */
+		linkAttributes() {
+			// If it's a router-link, we pass `to` and `exact`, otherwise only `href`
+			return this.to ? { to: this.to, exact: this.exact, ...this.$attrs } : { href: this.href, ...this.$attrs }
 		},
 	},
 	methods: {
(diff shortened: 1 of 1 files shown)


skjnldsv (approved):

src/components/NcBreadcrumb/NcBreadcrumb.vue
```diff
@@ -41,11 +41,8 @@ This component is meant to be used inside a Breadcrumbs component.
 		@dragleave="dragLeave">
 		<component :is="tag"
 			v-if="(nameTitleFallback || icon) && !$slots.default"
-			:exact="exact"
-			:to="to"
-			:href="href"
 			:title="title"
-			v-bind="$attrs"
+			v-bind="{ ...$attrs, ...to && { to, exact }, ...!to && { href } }"
 			v-on="$listeners">
```
ShGKme, line 46: This expression seems to be too complicated to me for template. Especially spreading boolean values.

What about making a computed property with `if` statement?

ShGKme (commented):
Instead of fixing it in the `NcBreadcrumb`, shouldn't we fix it in `NvBreadcrumbs`, where wrong props are passed?

https://github.com/nextcloud/nextcloud-vue/blob/f46bba55b191d25497878c41e00c1feb42ab36d4/src/components/NcBreadcrumbs/NcBreadcrumbs.vue#L552-L572

raimund-schluessler:
> Instead of fixing it in the `NcBreadcrumb`, shouldn't we fix it in `NvBreadcrumbs`, where wrong props are passed?
> 
> https://github.com/nextcloud/nextcloud-vue/blob/f46bba55b191d25497878c41e00c1feb42ab36d4/src/components/NcBreadcrumbs/NcBreadcrumbs.vue#L552-L572

The lines you show here are not related to the problem. They are for showing the hidden crumbs in a dropdown menu. `NcBreadcrumb` is used explicitly by the dev and hence definitely needs a fix. We can of course, as a follow-up, also adjust these lines to not pass `href` to `NcActionRouter`, but `NcActionRouter` does not care if its provided an `undefined` `href`, so I am not sure it's worth the effort.

ShGKme:
> The lines you show here are not related to the problem. They are for showing the hidden crumbs in a dropdown menu.

Yep, sorry, didn't look carefully

ShGKme (approved):

PR #3974, opened 2023-04-14
fix(NcHeaderMenu): fix unnecessary filter invert

skjnldsv:
- This was here for legacy menus that were still using icon classes
- Because this is a vue component, it means apps using it are already using svg (notifications, settings menu, contacts menu, search... etc) 

This also fixes the weird avatar invert bug
Fix https://github.com/nextcloud/server/issues/37379


szaimen:
@skjnldsv i suppose this fixes https://github.com/nextcloud/server/issues/37379?

szaimen (approved):
LGTM but didnt test

PR #3984, opened 2023-04-18
feat(NcRichContenteditable): Allow to pass in an icon URL

nickvergessen:
- Required for https://github.com/nextcloud/spreed/issues/9188
- Talk PR to show it in action https://github.com/nextcloud/spreed/pull/9332

### DIFF
--- a/src/components/NcRichContenteditable/NcAutoCompleteResult.vue
+++ b/src/components/NcRichContenteditable/NcAutoCompleteResult.vue
@@ -67,6 +67,10 @@ export default {
 			type: String,
 			required: true,
 		},
+		iconUrl: {
+			type: String,
+			default: null,
+		},
 		source: {
 			type: String,
 			required: true,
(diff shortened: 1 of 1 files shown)


src/components/NcRichContenteditable/NcAutoCompleteResult.vue
```diff
@@ -67,6 +67,10 @@ export default {
 			type: String,
 			required: true,
 		},
+		iconUrl: {
```
raimund-schluessler, line 70: Since this is the avatar url, I would also call it `avatarUrl`.

nickvergessen: Then it conflicts with the computed and the CSS avatar option is also called `icon` so I tried to keep it simple and inline

Antreesy: We could come up with something like `predefinedUrl`, as we provide here a ready-to-download URL

Antreesy (approved):
Works properly, shouldn't break anything

PR #3990, opened 2023-04-19
Improve breadcrumbs accessibility

### LINKED ISSUE #3951: Create right html structure for NcBreadcrumbs.vue and NcBreadcrumb.vue
Related ticket: https://github.com/nextcloud/server/issues/36918

The breadcrumb navigation represents a list and should best be implemented as a list.

To be fixed in:

- [ ] https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/NcBreadcrumbs/NcBreadcrumbs.vue
- [ ] https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/NcBreadcrumb/NcBreadcrumb.vue

TODO:

- [ ] styles for narrow screens: when the screen is too narrow, the menu appear and breadcrumbs items should be hidden.
- [ ] html structure: list which is inside of the `nav` navigation.
- [ ] screen reader visibility for narrow screens: when the screen is too narrow, the breadcrumbs items and the actions menu should be hidden (`aria-hidden="true"`).

See the right way of implementation in: https://github.com/nextcloud/server/issues/36918

raimund-schluessler:
This improves the accessibility of `NcBreadcrumbs`:

- the bread crumbs are now wrapped in a `nav`
- the breadcrumbs are now a list, each crumb is a `li`
- the Action menu holding the hidden crumbs got an `aria-hidden` atrribute

Closes #3951.

### DIFF
--- a/src/components/NcBreadcrumb/NcBreadcrumb.vue
+++ b/src/components/NcBreadcrumb/NcBreadcrumb.vue
@@ -29,7 +29,7 @@ This component is meant to be used inside a Breadcrumbs component.
 </docs>
 
 <template>
-	<div ref="crumb"
+	<li ref="crumb"
 		class="vue-crumb"
 		:class="{'vue-crumb--hovered': hovering}"
 		:[crumbId]="''"
@@ -68,7 +68,7 @@ This component is meant to be used inside a Breadcrumbs component.
 			<slot />
 		</NcActions>
 		<ChevronRight class="vue-crumb__separator" :size="20" />
-	</div>
+	</li>
 </template>
 
 <script>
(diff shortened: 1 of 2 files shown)


AndyScherzinger (approved):

skjnldsv (changes requested):
It's missing one important step:
- [ ] The hidden breadcrumbs should still be rendered and hidden visually, so the nav does not change for screen readers

skjnldsv:
> No, that was already the case before, so I didn't mention it as a change. See the markup with two crumbs hidden:

Oooh, I did not know that! :+1:

skjnldsv (approved):

PR #3991, opened 2023-04-20
Order tags by last used

Pytal:
Last used tags will now be sorted to the top of the list

## TODO
- [x] Use webdav lib in place of custom xml parsing


src/components/NcSelectTags/api.js
```diff
@@ -21,95 +22,61 @@
...
+	))
 }
 
-const parseXml = (xml) => {
-	let dom = null
-	try {
-		dom = (new DOMParser()).parseFromString(xml, 'text/xml')
-	} catch (e) {
-		console.error('Failed to parse xml document', e)
+export const fetchTags = async () => {
```
nickvergessen, line 50: There are instances with thousands of tags, so loading all of them might not scale well and/or even break the ui

Pytal: This change does not affect the number of tags returned from the server endpoint which is the same as before

Any changes to the number of tags returned, i.e. pagination, would be out-of-scope

Is there something within the scope of this PR that you'd suggest?

src/components/NcSelectTags/NcSelectTags.vue
```diff
@@ -337,10 +339,28 @@ export default {
...
+					lastUsedTags.push(tag)
+					continue
+				}
+				remainingTags.push(tag)
+			}
+
+			const sortByLastUsed = (a, b) => {
+				return lastUsedOrder.indexOf(a.id) - lastUsedOrder.indexOf(b.id)
+			}
+			lastUsedTags.sort(sortByLastUsed)
```
artonge, line 359:
```suggestion
			const lastUsedTags = lastUsedOrder.map(tagId => tags.find(tag => tag.id === tagId))
			const remainingTags = []
```

Or simply:

```suggestion
			this.sortedTags = tags.sort((a, b) => {
				return lastUsedOrder.indexOf(a.id) - lastUsedOrder.indexOf(b.id)
			})
```

artonge (approved):

susnux:
While this does not make it that worse, I am still not happy with having business logic inside the component library.
For me it makes more sense to have design / components separated from business logic and instead provide such components from the logic packages.

like e.g. a group select should not be in the vue library which should be more or less independent from API changes but lets say in `@nextcloud/provisioning` or something like that.

src/components/NcSelectTags/NcSelectTags.vue
susnux: :+1:

PR #4007, opened 2023-04-26
feat(NcRichContenteditable): Support iconUrl on mention bubbles

nickvergessen:
* Contributes to https://github.com/nextcloud/spreed/issues/9188#issuecomment-1520658853

Before | After
---|---
grafik | Bildschirmfoto vom 2023-04-26 13-06-18

### DIFF
--- a/src/components/NcRichContenteditable/NcMentionBubble.vue
+++ b/src/components/NcRichContenteditable/NcMentionBubble.vue
@@ -59,6 +59,10 @@ export default {
 			type: String,
 			required: true,
 		},
+		iconUrl: {
+			type: [String, null],
+			default: null,
+		},
 		source: {
 			type: String,
 			required: true,
@@ -70,6 +74,10 @@ export default {
 	},
 	computed: {
 		avatarUrl() {
+			if (this.iconUrl) {
+				return this.iconUrl
+			}
+
 			return this.id && this.source === 'users'
 				? this.getAvatarUrl(this.id, 44)
 				: null


Antreesy (approved):

src/components/NcRichContenteditable/NcMentionBubble.vue
```diff
@@ -59,6 +59,10 @@ export default {
 			type: String,
 			required: true,
 		},
+		iconUrl: {
+			type: [String, null],
+			default: null,
+		},
```
ShGKme, line 65: Very similar component [NcUserBubble](https://nextcloud-vue-components.netlify.app/#/Components/NcUserBubble) has similar prop called `avatarImage`.

Also, `icon` name is usually used for Material Design Icons or CSS `.icon-class` in Nextcloud Vue. Including the `icon` prop of this component - it is a CSS class for an icon too.

I'd consider rename it to either `url` (as in `NcAvatar`) or `avatarUrl` (like in `NcUserBubble`).

nickvergessen: I used iconUrl, as that is the prop name in userData which is handed over to tribute when autocompleting mention suggestions.
Having to have 2 different properties with the same name or adding another mapping in between felt wrong to me, but yeah I also tripped a bit.

nickvergessen: Ref https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/NcRichContenteditable/NcAutoCompleteResult.vue#L66-L73

nickvergessen: And that refs to https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/NcRichContenteditable/NcRichContenteditable.vue#L325

ShGKme: > And that refs to https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/NcRichContenteditable/NcRichContenteditable.vue#L325

But this is from `linkOptions`, not `autocompleteOptions`. There it is more "icon"

image

ShGKme: Ok, I see now

PR #4016, opened 2023-04-28
Add aria-hidden attr to the whole navigation depending on a sidebar state

JuliaKirschenheuter:
Fix an old sidebar navigation on server.

This fix is only for (where `app-navigation-vue` is used):
- [x] Files
- [x] Photos
- [x] Users
- [x] Apps

* Fixes: https://github.com/nextcloud/server/issues/36990
* Fixes: https://github.com/nextcloud/server/issues/36926

## Summary

The goal of this fix: don't allow screen reader navigation through a sidebar if sidebar is closed. No visual changes.

After:
Peek 2023-04-28 20-57

### DIFF
--- a/src/components/NcAppNavigation/NcAppNavigation.vue
+++ b/src/components/NcAppNavigation/NcAppNavigation.vue
@@ -59,15 +59,16 @@ emit('toggle-navigation', {
 		role="navigation"
 		:class="{'app-navigation--close':!open }">
 		<NcAppNavigationToggle :open="open" @update:open="toggleNavigation" />
-		<slot />
+		<div :aria-hidden="ariaHidden" class="app-navigation__content">
+			<slot />
+			<!-- List for Navigation li-items -->
+			<ul class="app-navigation__list">
+				<slot name="list" />
+			</ul>
 
-		<!-- List for Navigation li-items -->
-		<ul class="app-navigation__list">
-			<slot name="list" />
-		</ul>
-
-		<!-- Footer for e.g. AppNavigationSettings -->
-		<slot name="footer" />
+			<!-- Footer for e.g. AppNavigationSettings -->
+			<slot name="footer" />
+		</div>
 	</div>
 </template>
 
@@ -90,6 +91,11 @@ export default {
 		return {
 			open: true,
 		}
+	},
+	computed: {
+		ariaHidden() {
+			return this.open ? 'false' : 'true'
+		},
 	},
 
 	watch: {
@@ -156,10 +162,6 @@ export default {
 	-moz-user-select: none;
 	-ms-user-select: none;
 	user-select: none;
-	display: flex;
-	flex-direction: column;
-	flex-grow: 0;
-	flex-shrink: 0;
 	background-color:  var(--color-main-background-blur, var(--color-main-background));
 	-webkit-backdrop-filter: var(--filter-background-blur, none);
 	backdrop-filter: var(--filter-background-blur, none);
@@ -182,6 +184,13 @@ export default {
 		flex-direction: column;
 		gap: var(--default-grid-baseline, 4px);
 		padding: calc(var(--default-grid-baseline, 4px) * 2);
+	}
+	&__content {
+		height: 100%;
+		display: flex;
+		flex-direction: column;
+		flex-grow: 0;
+		flex-shrink: 0;
 	}
 }
 


src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -91,6 +93,14 @@ export default {
 			open: true,
 		}
 	},
+	computed: {
+		ariaHidden() {
+			if (this.open === true) {
+				return 'false'
+			}
+			return 'true'
```
raimund-schluessler, line 101:
```suggestion
			return (!this.open).toString()
```

src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -59,15 +59,17 @@ emit('toggle-navigation', {
...
-		<slot />
 
-		<!-- List for Navigation li-items -->
-		<ul class="app-navigation__list">
-			<slot name="list" />
-		</ul>
+		<div :aria-hidden="ariaHidden" class="app-navigation__content">
+			<slot />
+			<!-- List for Navigation li-items -->
+			<ul ref="navigationList" class="app-navigation__list">
```
raimund-schluessler, line 66: The ref is not used, no?
```suggestion
			<ul class="app-navigation__list">
```

src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -59,15 +59,17 @@ emit('toggle-navigation', {
...
+				<slot name="list" />
+			</ul>
 
-		<!-- Footer for e.g. AppNavigationSettings -->
-		<slot name="footer" />
+			<!-- Footer for e.g. AppNavigationSettings -->
+			<slot ref="footer" name="footer" />
```
raimund-schluessler, line 71: Same here
```suggestion
			<slot name="footer" />
```

ShGKme: Also, it is not possible to set ref on a slot. `<slot>` is a special template tag, not an HTML element or a Vue component.

src/components/NcAppNavigation/NcAppNavigation.vue
ShGKme: I'd personally prefer `return this.open ? 'true' : 'false'`.

It is the same as `(!this.open).toString()`, but more clear to see, that we have `'true'` or `'false'` string value here.

src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -183,6 +193,13 @@ export default {
 		gap: var(--default-grid-baseline, 4px);
 		padding: calc(var(--default-grid-baseline, 4px) * 2);
 	}
+	.app-navigation__content {
+		height: 100%;
+		display: flex;
+		flex-direction: column;
+		flex-grow: 0;
+		flex-shrink: 0;
```
ShGKme, line 201: These styles are moved from `.app-navigation`, which previously was the container of slots and the list, to the new container. Could we now remove them from the `.app-navigation` or they are still required?

src/components/NcAppNavigation/NcAppNavigation.vue
raimund-schluessler: I don't mind, both is fine.

src/components/NcAppNavigation/NcAppNavigation.vue
JuliaKirschenheuter: it is hard to test all apps and the risk to break something is high. I could remove https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/NcAppNavigation/NcAppNavigation.vue#L161-L162 but would like leave it to be safe.

src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -183,6 +189,13 @@ export default {
 		gap: var(--default-grid-baseline, 4px);
 		padding: calc(var(--default-grid-baseline, 4px) * 2);
 	}
+	&__content {
+		height: 100%;
+		display: flex;
+		flex-direction: column;
+		flex-grow: 0;
+		flex-shrink: 0;
+	}
```
artonge, line 198: Isn't it out of scope of this PR?

JuliaKirschenheuter: no, styles for wrapper (`app-navigation__content`) has to be fixed

ShGKme: > Isn't this out of scope of this PR?

Adding a new container for content breaks the layout set by the original container

```
# before

.app-navigation   --- a flex container
|-- default slot     --- flex container's content
|-- ul               --- flex container's content
|-- footer slot      --- flex container's content

# after

.app-navigation              --- an old flex container
|-- .app-navigation__content   --- the new flex container
|-- |-- default slot             --- flex container's content
|-- |-- ul                       --- flex container's content
|-- |-- footer slot              --- flex container's content
```

So all flex container-related styles should be moved.

> it is hard to test all apps and the risk to break something is high.

@JuliaKirschenheuter It seems to me, that it cannot be broken. The content was in the flex container, it is still in the flex container with the same layout. The old container now has only one element.

The only thing that may be broken is if some app set styles to the root of the component. But it will break anyway now.

JuliaKirschenheuter: you are right, thanks a lot!

src/components/NcAppNavigation/NcAppNavigation.vue
JuliaKirschenheuter: done

ShGKme (approved):
🐸

artonge (approved):

src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -156,10 +162,6 @@ export default {
 	-moz-user-select: none;
 	-ms-user-select: none;
 	user-select: none;
-	display: flex;
-	flex-direction: column;
-	flex-grow: 0;
-	flex-shrink: 0;
```
nickvergessen, line 162: This needs reverting it heavily breaks the talk left sidebar:
grafik

skjnldsv: @JuliaKirschenheuter :warning:

ShGKme: Yes, but not all of them.

These styles are flex container styles. They are moved.

```css
display: flex;
flex-direction: column;
```

while these are flex elements styles. They should be only here.

```css
flex-grow: 0;
flex-shrink: 0;
```

ShGKme: > this seems absolutely not related to this PR

50% of these styles are flex container styles. In this PR there was created a new flex container for all content, and all flex container styles were moved to the new flex container. So, they are related.

Although, another 50% were flex element styles and really not related.

JuliaKirschenheuter: Ok, this what i feared: it is hard to test every app. Thank you @ShGKme for explanation and fixing!

PR #4017, opened 2023-04-28
Use `NcActions` component for `NcAvatar`

### LINKED ISSUE #1683: Avatar popover glitch when opened the first time

### LINKED ISSUE #1762: Avatar menu alignment and sizing troubles
### Server side

On the server side in the sharing panel, there's a regression when upgrading from 3.2.0 to 3.7.2.

#### Before
image

#### After
image

### Talk app

In the Talk app the icon completely disappeared. This is because "Content.vue" from nextcloud-vue is applying `box-sizing: border-box` globally with v-deep! We better use explicit styling / overrides for this one in the avatar menu to make sure it can work everywhere.

image

When setting `box-sizing: content-box` the screenshot looks the same as on the server.

### Todos:

- [ ] add explicit `box-sizing: content-box` in the avatar menu item
- [ ] sort out alignment between icon and text

raimund-schluessler:
This PR migrates the `NcAvatar` component from using the deprecated `NcPopoverMenu` to `NcActions`.

| Before | After |
|-|-|
| Screenshot 2023-04-29 at 13-06-41 Tasks - Nextcloud | Screenshot 2023-04-29 at 13-09-07 Tasks - Nextcloud |

Closes #1683, closes #1762

### DIFF
--- a/src/components/NcAvatar/NcAvatar.vue
+++ b/src/components/NcAvatar/NcAvatar.vue
@@ -112,7 +112,8 @@ export default {
 		v-click-outside="closeMenu"
 		:class="{
 			'avatardiv--unknown': userDoesNotExist,
-			'avatardiv--with-menu': hasMenu
+			'avatardiv--with-menu': hasMenu,
+			'avatardiv--with-menu-loading': contactsMenuLoading
 		}"
 		:title="tooltip"
 		:style="avatarStyle"
@@ -120,7 +121,7 @@ export default {
 		:tabindex="hasMenu ? '0' : undefined"
 		:aria-label="avatarAriaLabel"
 		:role="hasMenu ? 'button' : undefined"
-		v-on="hasMenu ? { click: toggleMenu } : {}"
+		@click="toggleMenu"
 		@keydown.enter="toggleMenu">
 		<!-- @slot Icon slot -->
 		<slot name="icon">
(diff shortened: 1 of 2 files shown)


ShGKme (changes requested):
The loading indicator is visible only on hover

NcAvatar

raimund-schluessler:
> The loading indicator is visible only on hover
> 
> NcAvatar [ NcAvatar ](https://user-images.githubusercontent.com/25978914/235452981-83008cbf-f1f0-4782-8bd3-2ae1d837d992.gif) [ ](https://user-images.githubusercontent.com/25978914/235452981-83008cbf-f1f0-4782-8bd3-2ae1d837d992.gif)

Fixed.

src/components/NcAvatar/NcAvatar.vue
```diff
@@ -710,28 +707,38 @@ export default {
...
-		.icon-more {
+		:deep(.action-item__menutoggle) {
 			cursor: pointer;
 			opacity: 0;
 		}
 		&:focus,
-		&:hover {
-			.icon-more {
+		&:hover,
+		&#{&}-loading {
```
nickvergessen, line 721: What's that magic woodoo selector doing? 🪄

skjnldsv:
> NcAvatar

I see something odd on the gif too.
On hover, the three dot should probably not be seen over the avatar, no?
The contrast is quite problematic. It should be full background colour with the three dot svg on top of it?

raimund-schluessler:
> > NcAvatar
> I see something odd on the gif too. On hover, the three dot should probably not be seen over the avatar, no?

I think I don't really understand. When the menu is closed, and one hovers the avatar, the three dots icon shows up. This is how it is now already 🤔 

> The contrast is quite problematic. It should be full background colour with the three dot svg on top of it?

I don't know where this gif comes from, but I think there is something off with it. For me, the three dots have far more contrast, in an app and in the docs:

Screenshot 2023-05-09 at 09-32-43 Nextcloud Vue Style Guide

marcoambrosini (approved):

src/components/NcAvatar/NcAvatar.vue
ShGKme, line 721: Do we really need `.avatardiv--with-menu.avatardiv--with-menu-loading` here?

src/components/NcAvatar/NcAvatar.vue
```diff
@@ -710,28 +707,38 @@ export default {
...
 		}
+		:deep() {
+			.button-vue,
+			.button-vue__icon {
+				height: var(--size);
+				min-height: var(--size);
+				width: var(--size) !important;
+				min-width: var(--size);
+			}
+		}
```
ShGKme, line 741: Can we add a size/small prop to `NcButton`instead of deep internal classes overrides?

ShGKme (approved):
Works fine now

src/components/NcAvatar/NcAvatar.vue
ShGKme: `&` - parent selector, equals to `.avatardiv--with-menu`
`#{expression}` - interpolation, `#{$}` equals to value of `&`

So
```scss
.avatardiv {
    &--with-menu {
        &#{&}-loading {
            /* ... */
        }
    }
}
```

compiles to

```scss
.avatardiv--with-menu.avatardiv--with-menu-loading {
  /* ... */
}
```

PR #4036, opened 2023-05-04
fix: Avoid showing the reference list if no results were found

juliushaertl:
Fixes https://github.com/nextcloud/text/issues/3824

When there is no result for the references we should hide the container and also not fail early on trying to access an undefined object.


julien-nc (approved):
:+1:

nickvergessen (approved):

PR #4039, opened 2023-05-04
fix: Fix regression to still display references

juliushaertl:
Turns out I messed up testing of #4036 with the case to show references, this is the follow up fix to actually still show the link preview when expected

Before this change the link preview was only shown if the data was provided through the prop, not if it was dynamically fetched from the provided text.


nickvergessen (approved):

julien-nc (approved):
Ouch, I should have tested #4036

ShGKme:
@juliushaertl ready to merge and release?

PR #4044, opened 2023-05-05
Revert new AppSidebarTabs design

skjnldsv:


nickvergessen (approved):

raimund-schluessler:
If this change is generally desired, just not for version 7, how about branching-of a stable7 branch and reverting it just there?

marcelklehr (approved):

nickvergessen:
> If this change is generally desired, just not for version 7, how about branching-of a stable7 branch and reverting it just there?


That is exactly what we did, this PR here is against stable7

raimund-schluessler:
> > If this change is generally desired, just not for version 7, how about branching-of a stable7 branch and reverting it just there?
> 
> That is exactly what we did, this PR here is against stable7

Sorry, I missed that 🙈

skjnldsv:
> Is the revert on the `NcCheckboxRadioSwitch` required, too?

I kinda don't want to take risks tbh :shrug:

PR #4052, opened 2023-05-07
Migrate from deprecated title prop

raimund-schluessler:
Since master is `8.0.0` now, we can merge all breaking changes scheduled for the next major.

This PR removes the `nameTitleFallback` and makes the `name` prop required for `NcActions*`, `NcAppNavigationItem` and `NcBreadcrumb*`.

### DIFF
--- a/src/components/NcBreadcrumbs/NcBreadcrumbs.vue
+++ b/src/components/NcBreadcrumbs/NcBreadcrumbs.vue
@@ -552,8 +552,7 @@ export default {
 				const href = crumb.componentOptions.propsData.href
 				const disabled = crumb.componentOptions.propsData.disableDrop
 				const title = crumb.componentOptions.propsData.title
-				// TODO: Remove this fallback once nameTitleFallback is removed from NcBreadcrumb
-				const name = crumb.componentOptions.propsData.name || title
+				const name = crumb.componentOptions.propsData.name
 
 				// Decide whether to show the breadcrumbs as ActionRouter or ActionLink
 				let element = 'NcActionLink'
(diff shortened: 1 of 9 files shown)


skjnldsv (approved):
Could you be any more awesome Raimund? :heart_eyes: 
:bow:

src/components/NcBreadcrumbs/NcBreadcrumbs.vue
```diff
@@ -573,7 +572,6 @@ export default {
 					props: {
 						href,
 						title,
-						name: '', // TODO: Remove this once nameTitleFallback is removed from actionText.js mixin
```
nickvergessen, line 576: Shouldn't this be passing on the name, like:
```
						name,
```

So that name is set in NcActionLink and NcActionRouter elements?

raimund-schluessler: No, the `name` prop of `NcBreadcrumb` does not map to the `name` prop of the `NcActions*`, it goes into the default slot of the actions, as we do it some lines below.
We could of course change that, and map `breadcrumb:name`->`actionLink:name` and `breadcrumb:default-slot` -> `actionLink:default-slot`, but then we would also need to introduce an actions slot for `NcBreadcrumb`, since the actions currently go into the default slot. But I would say that this is out of scope for this PR here and something for a follow-up, if we decide to do that.

nickvergessen: Ah, was not aware.
No, don't need to add more stuff to it. When it worked the same before we can just continue

Pytal (approved):

ShGKme:
Should we start to make breaking changes list or migration guide?

nickvergessen:
A migration guide similar to the OCP changes sounds like a wunderful idea!

Issue #4020, opened 2023-04-30
Add info type to NcNoteCard


moan0s:
As discussed in https://github.com/nextcloud/server/pull/37953 a info-type of NcNoteCard would be needed for certain types of communication to the user

As described by @nimishavijay we could use
 
- --color-primary for the icon and border
- --color-primary-light for the background

susnux:
Fixed with https://github.com/nextcloud/nextcloud-vue/pull/4063

PR #4063, opened 2023-05-08
Add info NoteCard

moan0s:
This is a proposal for #4020 and linked to https://github.com/nextcloud/server/pull/37953

I am net in any way experienced with this, so feel free to point out any errors. I am also not set on the colors, it's just what felt kinda right (but again, I am not a frontend dev)

### DIFF
--- a/src/components/NcNoteCard/NcNoteCard.vue
+++ b/src/components/NcNoteCard/NcNoteCard.vue
@@ -20,9 +20,10 @@
 
 <docs>
 This component is made to display additional information to the user. It is
-available in three versions:
+available in four versions:
 
 - success: Display a successful message
+- info: Display an informational message
 - warning: Display a warning to the user. This indicate that the action they want
 - error: Display an error message. For example
 
@@ -40,6 +41,10 @@ When using an error type,
 
 	<NcNoteCard type="success">
 		<p>You won</p>
+	</NcNoteCard>
+
+	<NcNoteCard type="info">
+		<p>For your information</p>
 	</NcNoteCard>
 </div>
 ```
@@ -66,6 +71,7 @@ When using an error type,
 import CheckboxMarkedCircle from 'vue-material-design-icons/CheckboxMarkedCircle.vue'
 import AlertDecagram from 'vue-material-design-icons/AlertDecagram.vue'
 import Alert from 'vue-material-design-icons/Alert.vue'
+import Information from 'vue-material-design-icons/Information.vue'
 
 export default {
 	name: 'NcNoteCard',
@@ -77,7 +83,7 @@ export default {
 		type: {
 			type: String,
 			default: 'warning',
-			validator: type => ['success', 'warning', 'error'].includes(type),
+			validator: type => ['success', 'info', 'warning', 'error'].includes(type),
 		},
 		showAlert: {
 			type: Boolean,
@@ -98,6 +104,8 @@ export default {
 				return AlertDecagram
 			case 'success':
 				return CheckboxMarkedCircle
+			case 'info':
+				return Information
 			case 'warning':
 				return Alert
 			default:
@@ -110,6 +118,8 @@ export default {
 				return 'var(--color-error)'
 			case 'success':
 				return 'var(--color-success)'
+			case 'info':
+				return 'var(--color-info)'
 			case 'warning':
 				return 'var(--color-warning)'
 			default:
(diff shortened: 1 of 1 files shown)


src/components/NcNoteCard/NcNoteCard.vue
```diff
@@ -110,6 +114,8 @@ export default {
 				return 'var(--color-error)'
 			case 'success':
 				return 'var(--color-success)'
+			case 'info':
+				return 'var(--color-info)'
```
szaimen, line 118:
```suggestion
				return 'var(--color-primary)'
```

src/components/NcNoteCard/NcNoteCard.vue
```diff
@@ -143,6 +149,11 @@ export default {
 		--note-theme: var(--color-success);
 	}
 
+	&--info {
+		--note-background: rgba(var(--color-info-rgb), 0.1);
+		--note-theme: var(--color-info);
```
szaimen, line 154:
```suggestion
		--note-background: var(--color-primary-light);
		--note-theme: var(--color-primary);
```

styleguide/assets/default.css
```diff
@@ -26,6 +26,9 @@
   --color-success: #46ba61;
   --color-success-rgb: 70,186,97;
   --color-success-hover: #6ac780;
+  --color-info: #646be31a;
+  --color-info-rgb: 100, 107, 227;
+  --color-info-hover: #4347961a;
```
szaimen, line 31:
```suggestion
```

szaimen (approved):
Thanks for your PR! I think it should work with the changes below.

szaimen (changes requested):
Clicked the wrong button, see the feedback above

szaimen (approved):
Works and looks good. Thanks a lot! :)

I took the liberty to add it to the examples.

@moan0s for us being able to merge this, can you please fix DCO? See https://github.com/nextcloud/nextcloud-vue/pull/4063/checks?check_run_id=13318409670 for details

moan0s:
Oh yes, sorry I forgot that and did not check the CI. Should be fixed now

src/components/NcNoteCard/NcNoteCard.vue
```diff
@@ -143,6 +153,11 @@ export default {
 		--note-theme: var(--color-success);
 	}
 
+	&--info {
+		--note-background: var(--color-primary-light);
+		--note-theme: var(--color-primary);
```
jancborchardt, line 158: I think we need to introduce a separate variable `--color-info` here which uses `#006aa3` manually. Otherwise for Nextclouds themed in e.g. red, the info box looks like a warning box. 

This is the same issue for the infobox in Text for example, cc @max-nextcloud @mejo-

src/components/NcNoteCard/NcNoteCard.vue
```diff
@@ -23,6 +23,7 @@ This component is made to display additional information to the user. It is
 available in three versions:
```
raimund-schluessler, line 23:
```suggestion
available in four versions:
```

src/components/NcNoteCard/NcNoteCard.vue
```diff
@@ -23,6 +23,7 @@ This component is made to display additional information to the user. It is
...
 
 - success: Display a successful message
+- info: Display a informational message
```
raimund-schluessler, line 26:
```suggestion
- info: Display an informational message
```

src/components/NcNoteCard/NcNoteCard.vue
raimund-schluessler: I think `--color-primary-default` should work, but there is no `--color-primary-default-light` afaik.

susnux: `--color-primary-default` can be overridden by custom themes, so I think adding a `--color-info` (and `--color-info-rgb` + `--color-info-hover` respectively) would be the best way to go.

mejo-: Oh, good catch @jancborchardt! I agree that we should introduce a `--color-info` variable! I opened https://github.com/nextcloud/text/issues/4161 to track this issue in Text.

mejo- (approved):
Code looks good. Thanks a lot for tackling this @moan0s.

moan0s:
Thanks for doing all the real work, I very much appreciate it!

max-nextcloud:
Not backported due to the dependency on Nextcloud 27.

susnux:
@max-nextcloud we can add a fallback on the css vars on stable7 so it will work on all versions. (If there is a need for this)

max-nextcloud:
@susnux I was thinking the same but then decided to add the fallback inside text: https://github.com/nextcloud/text/commit/c564511d9459e5bbbd0685a768920e44ebba9506 . Pretty happy with that as it will also be easy to backport whereas using the NoteCard component would have required more changes.

PR #4064, opened 2023-05-08
Fix styles for long tab names in `NcAppSidebarTabs`

Antreesy:
* Follow-up to #3945 
* Fix strange text-wrapping for long tab names

Example from Talk:
🏚️ Before | 🏡 After
---|---
image | image


raimund-schluessler (approved):

raimund-schluessler:
In a follow-up, we should also fix the icons positions, as they are not perfectly aligned in the center, due to some strange margins, probably necessary for the `NcCheckboxRadioSwitch`.

Antreesy:
> In a follow-up, we should also fix the icons positions, as they are not perfectly aligned in the center

Yes, the margin for icon should be reset, probably here:
https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue#L532

Or here:
https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue#L622

But I'm not sure, if it could be done without :deep selector

jancborchardt (approved):
Looks good, agree with the icon alignment issue!

nickvergessen:
Kinda sad to see the participant count dropped due to this, but yeah it's better then 3 lines.
Maybe we can find another place in talk then.

susnux:
> Kinda sad to see the participant count dropped due to this, but yeah it's better then 3 lines. Maybe we can find another place in talk then.

How about putting it as a counter bubble on the icon? (did not test this but I think it could look nice)

susnux (approved):

nickvergessen:
> How about putting it as a counter bubble on the icon? (did not test this but I think it could look nice)

Bubbles don't look too good with 3 digits anymore, but yeah

Antreesy:
Another approve round for icons centering, please take a look again at last commit 👀 
image

susnux:
@Antreesy the icon part is already handled here:
https://github.com/nextcloud/nextcloud-vue/pull/4065

Antreesy:
Reabsed onto master, revert changes for `NcCheckboxRadioSwitch`
So, first three approvals are valid again =)

raimund-schluessler (approved):

PR #4079, opened 2023-05-10
fix(NcSelect): Fix disabled state of NcSelect with dark mode

nickvergessen:
Before | After
---|---
Background is not our background color and therefor also used in dark mode | background is our color variable so perfectly fine in dark mode
grafik | Bildschirmfoto vom 2023-05-10 09-57-25
Bildschirmfoto vom 2023-05-10 10-08-40 | Bildschirmfoto vom 2023-05-10 10-09-31

### DIFF
--- a/src/components/NcSelect/NcSelect.vue
+++ b/src/components/NcSelect/NcSelect.vue
@@ -945,6 +945,9 @@ body {
 	--vs-state-disabled-color: var(--color-text-maxcontrast);
 	--vs-state-disabled-controls-color: var(--color-text-maxcontrast);
 	--vs-state-disabled-cursor: not-allowed;
+	--vs-disabled-bg: var(--color-background-dark);
+	--vs-disabled-color: var(--color-text-maxcontrast);
+	--vs-disabled-cursor: not-allowed;
 
 	/* Borders */
 	--vs-border-color: var(--color-border-maxcontrast);
@@ -998,13 +1001,13 @@ body {
 		margin-right: 2px;
 	}
 
-	&.vs--open .vs__dropdown-toggle {
-		border-color: var(--color-primary);
+	&.vs--open .vs__dropdown-toggle:not([disabled]) {
+		border-color: var(--color-primary-element);
 		border-bottom-color: transparent;
 	}
 
-	&:not(.vs--open) .vs__dropdown-toggle:hover {
-		border-color: var(--color-primary);
+	&:not(.vs--open) .vs__dropdown-toggle:hover:not([disabled]) {
+		border-color: var(--color-primary-element);
 	}
 
 	&--no-wrap {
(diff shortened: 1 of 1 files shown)


nickvergessen:
Found while fixing https://github.com/nextcloud/server/pull/38170

Antreesy (approved):
:seal:

src/components/NcSelect/NcSelect.vue
```diff
@@ -945,6 +945,9 @@ body {
 	--vs-state-disabled-color: var(--color-text-maxcontrast);
 	--vs-state-disabled-controls-color: var(--color-text-maxcontrast);
 	--vs-state-disabled-cursor: not-allowed;
+	--vs-disabled-bg: var(--color-background-dark);
+	--vs-disabled-color: var(--color-text-maxcontrast);
+	--vs-disabled-cursor: not-allowed;
```
Pytal, line 950: 🤔 the `--vs-state-disabled-*` variables are already set above and are aliased to these variables internally https://github.com/nextcloud-deps/vue-select/blob/9294fdcbf39380fd5cc766144b18b6f86cf107f4/src/css/global/states.css#L13C35-L15

Is this solvable without adding non-publicly documented variables https://vue-select.org/guide/css.html#available-css-variables-3-18?

nickvergessen: I have no clue why, but it fixes it. As per https://github.com/nextcloud/nextcloud-vue/pull/4079#discussion_r1189544610 it could be the order, since `--vs-disabled-bg: var(--vs-state-disabled-bg);` is not a permanent redirect, but it writes the currently value of the second var into the first one. So maybe we only replace the `--vs-state-disabled-bg` after that and therefore also need to overwrite the first var

nickvergessen: Or well:
https://github.com/nextcloud/nextcloud-vue/blob/93f9ab752f7d64dd14afbe6d549f0e1fab01d6ae/src/components/NcSelect/NcSelect.vue#L525
I'm pretty sure that is exactly what is going on

susnux: I think this happens because:
`--vs-disabled-bg: var(--vs-state-disabled-bg);`  from vue-select if on the `:root` selector.

And our `--vs-state-disabled-bg: var(--color-background-dark);` is on the `body` selector.

So when `--vs-disabled-bg` is set, the `--vs-state-disabled-bg` variable holds the old value.

AndyScherzinger (approved):

src/components/NcSelect/NcSelect.vue
susnux: So I think it has the same cause as this one: https://github.com/nextcloud/server/pull/36462

PR #4084, opened 2023-05-11
enh(NcActions): Navigation link actions

Pytal:
Automatically remove attributes that are incompatible with accessible navigation when all action links are website navigation links

### References

- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-haspopup
- https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -969,6 +969,16 @@ export default {
 			action => action?.componentOptions?.tag || action?.componentOptions?.Ctor?.extendOptions?.name
 		)
 
+		const isNavLink = (action) => {
+			const componentName = action?.componentOptions?.Ctor?.extendOptions?.name ?? action?.componentOptions?.tag
+			return (
+				componentName === 'NcActionLink'
+				&& action?.componentOptions?.propsData?.href?.startsWith(window.location.origin)
+			)
+		}
+		// Automatically detect whether all actions are website navigation links
+		const isNav = actions.every(isNavLink)
+
 		/**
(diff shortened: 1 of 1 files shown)


src/components/NcActions/NcActions.vue
```diff
@@ -725,6 +725,16 @@ export default {
...
+
+		/**
+		 * Enable if actions are website navigation links
+		 *
+		 * Removes attributes incompatible with accessible navigation
+		 */
+		isNav: {
+			type: Boolean,
+			default: false,
+		},
```
skjnldsv, line 737: It can probably be done automatically by checking the domain?
I'm scared of the amount of props we have everywhere :see_no_evil:

JuliaKirschenheuter (approved):

JuliaKirschenheuter (approved):

Pytal:
> I haven't realized that would look that complex 🙈 I can approve, but I'm really unsure

Either way is fine with me, just the initial reason for not detecting it automatically in the first place was that the overhead didn't seem worth it

Let me know which way you'd think is best and I'll update as needed :)

Pytal:
Which one do you think is best @skjnldsv?

skjnldsv:
> Which one do you think is best @skjnldsv?

I don't know.
I feel both solutions have their own dilemmas :/

artonge (approved):
Auto-detecting looks ok to me.
I would add a comment, though.

PR #4103, opened 2023-05-12
Files : Left panel dropdown improvement for NC 26

Jerome-Herbinet:
* Resolves:  [#38155](https://github.com/nextcloud/server/issues/38155)

(from @Clementine46)

**/!\ Testing is welcome in order to make sure it's not doing side effects elsewhere in the UI.**

## Before : 
[before.webm](https://github.com/nextcloud/nextcloud-vue/assets/33763786/653d4208-0dd7-4d4f-a32c-0177ce6988e3)

## After  : 
[after.webm](https://github.com/nextcloud/nextcloud-vue/assets/33763786/d5a919ca-28ce-49ec-9106-28dbccf317e6)

### DIFF
--- a/src/components/NcAppNavigationItem/NcAppNavigationIconCollapsible.vue
+++ b/src/components/NcAppNavigationItem/NcAppNavigationIconCollapsible.vue
@@ -82,9 +82,6 @@ export default {
 	z-index: 105; // above a, under button
 	color: var(--color-main-text);
 	right: 0;
-	&:hover{
-		color: var(--color-primary-element);
-	}
 	&--open {
 		color: var(--color-main-text);
 		&:hover{
--- a/src/components/NcAppNavigationItem/NcAppNavigationIconCollapsible.vue
+++ b/src/components/NcAppNavigationItem/NcAppNavigationIconCollapsible.vue
@@ -81,7 +81,7 @@ export default {
 	position: absolute;
 	z-index: 105; // above a, under button
 	color: var(--color-main-text);
-
+	right: 0;
 	&:hover{
 		color: var(--color-primary-element);
 	}
--- a/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
+++ b/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
@@ -868,8 +868,7 @@ export default {
 	&.app-navigation-entry--no-icon,
 	&:hover, &:focus {
 		a .app-navigation-entry-icon {
-			// hides the icon
-			visibility: hidden;
+			visibility: visible;
 		}
 		.icon-collapse {
 			//shows the triangle button


Jerome-Herbinet:
Also see the other PR for Nextcloud 25 https://github.com/nextcloud/server/pull/38217

nimishavijay (changes requested):
Super nice! I would only suggest that the hover effect doesn't need to use the primary color as we don't use that pattern anywhere else yet, the hover feedback can just be the change in the background color :) other than that it looks great!

nimishavijay:
I meant the hover effect under the dropdown button icon, on hover the icon changes to the primary color.

image

I just checked the other apps, and you are right, nothing has been changed by you. However, it's not intended in the first place. So I suggest on hover we could just have the background color changing, and the color of the icon can remain the same (--color-main-text). What do you think? :)

Jerome-Herbinet:
> I meant the hover effect under the dropdown button icon, on hover the icon changes to the primary color.
> 
> image
> 
> I just checked the other apps, and you are right, nothing has been changed by you. However, it's not intended in the first place. So I suggest on hover we could just have the background color changing, and the color of the icon can remain the same (--color-main-text). What do you think? :)

OK understood @nimishavijay :-) It's done.

nimishavijay (approved):
If the hover effect has been changed, then it is good to go! nice work! :)

skjnldsv (approved):

skjnldsv (approved):

GretaD:
this pr breaks mail, and most probably other apps that have an actionmenu on side bar

Screenshot from 2023-05-17 12-14-29

Jerome-Herbinet:
> this pr breaks mail, and most probably other apps that have an actionmenu on side bar
> 
> Screenshot from 2023-05-17 12-14-29

@GretaD, 
Thanks for your feedback. 
IMO it seems complicated to separate behaviour in nextcloud-vue to apply styles only in "Files" context (obviously, I don't find there any specific class or id to target, relative to Files context).
Instead, what do you think about trying to override CSS styles in apps.scss (core) ?
Any other suggestion ? 

@nimishavijay @jancborchardt @juliushaertl

szaimen:
I think this needs to be handled by this component. If both features are used at the same time, it should not coolide but instead be shown next to each other.

ChristophWurst:
Apps are reaching RC state for Nextcloud 27 today. Can this be fixed quickly or should we revert and redo cleanly?

Jerome-Herbinet:
I think things look more delicate and complex than I thought before starting to work on it. 
I propose to close this RP for now (and retry quietly later).
@szaimen @ChristophWurst

szaimen:
Let me try one thing. If that does not work we can revert this change.

Jerome-Herbinet:
Note also that the problem could be the same in https://github.com/nextcloud/server/pull/38338 event if styles are elsewhere.

szaimen:
This fixes the problem in my testing: https://github.com/nextcloud/nextcloud-vue/pull/4134

Let me know if it also for you :)

PR #4104, opened 2023-05-12
fix(NcSettingsSelectGroup): Fix "this.getValueObject is not a function"

nickvergessen:
Tried to use the component in an app, but it fails to load.

- [x] So I wrote a docs sample and it had the same issue.
- [x] Fixed the bug
- [x] Remove unused property
- [ ] Try to extend the example (but I need to fake the response of the following request, do we have a sample how to achieve that?
    - GET /ocs/v2.php/cloud/groups/details?search=&limit=10
- [ ] Migrate the component from deprecated NcMultiselect to NcSelect

### DIFF
--- a/src/components/NcSettingsSelectGroup/NcSettingsSelectGroup.vue
+++ b/src/components/NcSettingsSelectGroup/NcSettingsSelectGroup.vue
@@ -28,35 +28,47 @@ Simple group selection option
 
 ```vue
 <template>
-	<NcSettingsSelectGroup
-		label="Group selection in admin settings">
-	</NcSettingsSelectGroup>
+	<div>
+		<NcSettingsSelectGroup
+			v-model="groups"
+			label="Group selection in admin settings">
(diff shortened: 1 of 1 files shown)


raimund-schluessler:
>    * [ ]  Try to extend the example (but I need to fake the response of the following request, do we have a sample how to achieve that?
>       
>       * GET /ocs/v2.php/cloud/groups/details?search=&limit=10

This function from `NcSelectTags` might help. In the vue docs, some fake tags are returned there:

https://github.com/nextcloud/nextcloud-vue/blob/55b2b9f4611a59815a6be1aca7d97fa45c07b9f1/src/components/NcSelectTags/api.js#L91-L111

nickvergessen:
> The id prop is also not used, and the component seems to be in a bad shape generally.

Thanks, will see if we use the id instead or drop it

> This function from NcSelectTags might help. In the vue docs, some fake tags are returned there:

Thanks a lot 🙇🏼‍♂️

PR #4109, opened 2023-05-13
Add aria-expanded to settings accordion button

Fenn-CS:
Mark settings button that triggers the expand/collaps of the side nav settings with aria-expanded attribute to improve accesibility.

Resolves : https://github.com/nextcloud/server/issues/37136

### DIFF
--- a/src/components/NcAppNavigationSettings/NcAppNavigationSettings.vue
+++ b/src/components/NcAppNavigationSettings/NcAppNavigationSettings.vue
@@ -27,6 +27,9 @@
 		<div id="app-settings__header">
 			<button class="settings-button"
 				type="button"
+				:aria-expanded="open ? 'true' : 'false'"
+				aria-controls="app-settings__content"
+				:aria-label="ariaLabel"
 				@click="toggleMenu">
 				<Cog class="settings-button__icon" :size="20" />
 				<span class="settings-button__label">{{ title }}</span>
@@ -76,6 +79,9 @@ export default {
 				this.closeMenu,
 				this.clickOutsideOptions,
 			]
+		},
+		ariaLabel() {
+			return t('Open settings menu')
 		},
 	},
 	methods: {
--- a/l10n/messages.pot
+++ b/l10n/messages.pot
@@ -143,6 +143,9 @@ msgstr ""
 msgid "Open navigation"
 msgstr ""
 
+msgid "Open settings menu"
+msgstr ""
+
 msgid "Password is secure"
 msgstr ""
 


artonge (approved):

src/components/NcAppNavigationSettings/NcAppNavigationSettings.vue
```diff
@@ -27,6 +27,9 @@
 		<div id="app-settings__header">
 			<button class="settings-button"
 				type="button"
+				:aria-expanded="open"
+				aria-controls="app-settings__content"
+				aria-label="Open settings menu"
```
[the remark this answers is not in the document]
nickvergessen:
```suggestion
				:aria-label="t('Open settings menu')"
```

susnux (approved):

src/components/NcAppNavigationSettings/NcAppNavigationSettings.vue
```diff
@@ -27,6 +27,9 @@
 		<div id="app-settings__header">
 			<button class="settings-button"
 				type="button"
+				:aria-expanded="open"
+				aria-controls="app-settings__content"
+				aria-label="t('Open settings menu')"
```
raimund-schluessler, line 32: Unfortunately, this does not work. It needs to be a computed property, otherwise the l10n extraction does not find the string (seems to be a long standing bug in the script). You can see that the string did not get extracted into the pot file.

raimund-schluessler (changes requested):
See my comment.

src/components/NcAppNavigationSettings/NcAppNavigationSettings.vue
nickvergessen: And you are missing the `:` to make Vue realise the string inside is to be interpreted as JS code:

```suggestion
				:aria-label="t('Open settings menu')"
```

Fenn-CS: Right, thanks! Fixed.

raimund-schluessler: As I said, this needs to be a computed property.

nickvergessen: yeah, both need to be done. Sorry for the confusion

src/components/NcAppNavigationSettings/NcAppNavigationSettings.vue
```diff
@@ -27,6 +27,9 @@
 		<div id="app-settings__header">
 			<button class="settings-button"
 				type="button"
+				:aria-expanded="open"
+				aria-controls="app-settings__content"
+				:aria-label="t('Open settings menu')"
```
raimund-schluessler, line 32:
```suggestion
				:aria-label="ariaLabel"
```
and below in the computed section add:
```js
ariaLabel() {
	return t('Open settings menu')
},
```
then run `npm run l10n:extract` to extract the string into the translation files.

Fenn-CS: Thank for clarifying, fixed.

src/components/NcAppNavigationSettings/NcAppNavigationSettings.vue
```diff
@@ -27,6 +27,9 @@
 		<div id="app-settings__header">
 			<button class="settings-button"
 				type="button"
+				:aria-expanded="open"
```
raimund-schluessler, line 30: I am not entirely sure, but I think this needs to be a string, instead of a `Boolean`. At least we do it in every other place like this:
```suggestion
				:aria-expanded="open.toString()"
```

For example see:
https://github.com/nextcloud/nextcloud-vue/blob/b8e7742fce4024beeaf6c6b35fc32220c452ff2c/src/components/NcAppNavigationItem/NcAppNavigationItem.vue#L228
Sorry, that I didn't notice this earlier.

Fenn-CS: Fixed

raimund-schluessler (approved):

PR #4120, opened 2023-05-15
Improve `NcSettingsSelectGroup` - Convert to `NcSelect`

susnux:
Adding documentation, move to NcSelect and drop unused property
* Moved from deprecated `NcMultiselect` to `NcSelect` (Ref: https://github.com/nextcloud/nextcloud-vue/issues/3743 )
* Debounce search API calls (200ms) to prevent spamming the API while typing
* Cache empty search groups (the ones loaded when the search dropdown is opened) for current page
* Load display names of initial groups (those which were given through the `value` property).
* Removed unused `hint` property
* Added `placeholder` property and added a hidden label element
      The `label` property is still used for the placeholder for backwards compatibility.
* Added documentation for the events and added an example (Add axios mocking so we can emulate querying for groups on the netlify docs)

### DIFF
--- a/src/components/NcSettingsSelectGroup/NcSettingsSelectGroup.vue
+++ b/src/components/NcSettingsSelectGroup/NcSettingsSelectGroup.vue
@@ -3,6 +3,7 @@
   -
   - @author Julius Härtl <jus@bitgrid.net>
   - @author Greta Doci <gretadoci@gmail.com>
+  - @author Ferdinand Thiessen <opensource@fthiessen.de>
   -
   - @license GNU AGPL version 3 or any later version
   -
@@ -21,28 +22,51 @@
   -
   -->
 
+<docs>
+```vue
 <template>
-	<NcMultiselect :value="inputValue"
-		:options="groupsArray"
-		:options-limit="5"
-		:placeholder="label"
-		track-by="id"
-		label="displayname"
-		class="multiselect-vue"
-		:multiple="true"
-		:close-on-select="false"
-		:tag-width="60"
-		:disabled="disabled"
-		@input="update"
-		@search-change="findGroup">
-		<template #noResult>
-			<span>{{ t( 'No results') }}</span>
-		</template>
-	</NcMultiselect>
+	<section>
+		<NcSettingsSelectGroup v-model="groups" placeholder="Select user groups" label="The hidden label" />
+		<NcSettingsSelectGroup v-model="groups" :disabled="true" label="Also a fallback for the placeholder" />
+		<div>You have selected: <code>{{ groups }}</code></div>
+	</section>
+</template>
+<script>
+export default {
+	data() {
+		return {
+			groups: []
+		}
+	}
+}
+</script>
+<style scoped>
+section * {
+	padding: 6px 0px;
+}
+</style>
+```
+</docs>
+
+<template>
+	<div>
+		<label v-if="label" :for="id" class="hidden-visually">{{ label }}</label>
+		<NcSelect :value="inputValue"
+			:options="groupsArray"
+			:placeholder="placeholder || label"
+			:input-id="id"
+			:limit="5"
+			label="displayname"
+			:multiple="true"
+			:close-on-select="false"
+			:disabled="disabled"
+			@input="update"
+			@search="findGroup" />
+	</div>
 </template>
 
 <script>
-import NcMultiselect from '../../components/NcMultiselect/index.js'
+import NcSelect from '../../components/NcSelect/index.js'
 import { t } from '../../l10n.js'
 import l10n from '../../mixins/l10n.js'
 import GenRandomId from '../../utils/GenRandomId.js'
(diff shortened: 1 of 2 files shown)


raimund-schluessler:
Supersedes #4104, I guess.

susnux:
:face_exhaling: Somehow I missed that PR

But yes, basically this is fixing the same issue (+ more).

raimund-schluessler:
Is it the desired behaviour that the groups load only after one types into the input field? This way, no groups show up when one just opens the `NcSelect` with the arrow-down. But I can imagine that loading all groups by default might be to much on large instances!?

susnux:
@raimund-schluessler it is (and was) limited to 10 groups per search / api call. I have now pushed another commit for caching at least the initial groups, so that if you use the component multiple time only one API call for an *empty* search is performed.
Not sure if session storage is overkill, but the library does not use any vue store (like Pinia, etc).

nickvergessen:
All good, I can close mine

raimund-schluessler (approved):

src/components/NcSettingsSelectGroup/NcSettingsSelectGroup.vue
```diff
@@ -111,7 +138,7 @@ export default {
 	},
 	computed: {
 		inputValue() {
-			return this.getValueObject()
+			return this.getValueObject
```
nickvergessen, line 141: This change is only needed to make it usable in general, right?
Would it make sense to backport it to 7.11.x?

susnux: Yes, I would partially backport this

src/components/NcSettingsSelectGroup/NcSettingsSelectGroup.vue
```diff
@@ -131,8 +158,25 @@ export default {
...
+		update(updatedValue) {
+			const value = updatedValue.map((element) => element.id)
+			/** Emitted when the groups selection changes<br />**Payload:** `value` (`Array`) - *Ids of selected groups */
+			this.$emit('input', value)
+		},
+		/**
+		 * Load initial groups if not already loaded
+		 */
+		async initGroups() {
+			if (Object.keys(this.groups).length === 0) {
```
nickvergessen, line 170: When groups are selected they don't have their display name:
Bildschirmfoto vom 2023-05-17 07-06-09

As soon as I click into the field once, the displayname is loaded and used:
Bildschirmfoto vom 2023-05-17 07-06-14

you can easily create groups with diverging displaynames with occ:
```
occ group:add --display-name "Display name" `uuidgen`
```

nickvergessen: The "problem" is a search per selected group id might not be acceptable.
So I guess we should require that id+displayname of currently selected groups are given as a property.

nickvergessen (changes requested):

src/components/NcSettingsSelectGroup/NcSettingsSelectGroup.vue
nickvergessen: If you need an app for testing, you can use the announcementscenter with this branch:
https://github.com/nextcloud/announcementcenter/pull/648 
Setting is in the additional settings block

susnux: @nickvergessen I fixed the problem of the missing display name as well as I add a debounce to the input so that it does not spam the API with requests while typing

susnux:
All comments resolved :)

nickvergessen (changes requested):
Selection is still somewhat broken. I can also reproduce on styleguide

1. Manually change the `groups: []` and add `groups: ['foo']`

grafik

2. Open the dropdown
3. No more options are shown
4. Clear the selection using the "X" on foo
5. Force reload
6. Selection is not usable anymore

src/components/NcSettingsSelectGroup/NcSettingsSelectGroup.vue
```diff
@@ -126,15 +162,64 @@ export default {
...
+			// Run the watch handler also when the component is initially mounted
+			immediate: true,
+		},
+	},
+	/**
+	 * Load groups matching the empty query to reduce API calls
+	 */
+	async mounted() {
+		// version scoped to prevent issues with different library versions
+		const savedGroups = window.sessionStorage.getItem(`${appName}:${appVersion}/initialGroups`)
```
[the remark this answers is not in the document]
susnux: They should not interfere as only the "empty" search is cached. Moreover this is exactly what it should do: Cache the empty search for multiple elements within the same app.

For example think of the users list, each user has a group select, so without this or with binding to the ID we would dispatch the same search for every user.

nickvergessen (approved):
Seems to work now, yes

src/components/NcSettingsSelectGroup/NcSettingsSelectGroup.vue
nickvergessen: Can you do that? :)

nickvergessen: @susnux please backport to version 7, then we need to release it and update the version in server stable27 and stable26

PR #4128, opened 2023-05-17
Lazy load NcDatetimePicker in NcActionInput

pulsejet:
Mostly for discussion at the current stage. Related https://github.com/nextcloud/server/pull/38329

`NcDatetimePicker` depends on `NcTimezonePicker`, which in turn depends on `@nextcloud/calendar-js`, which makes it bundle unnecessarily large. One solution is to lazy load this particular component only if needed.


susnux:
There is no activity for over a year. Please feelfree to reopen if still needed!

PR #4134, opened 2023-05-17
fix toggle overlapping other stuff

szaimen:
Fix https://github.com/nextcloud/nextcloud-vue/pull/4103#issuecomment-1551128985

### DIFF
--- a/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
+++ b/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
@@ -166,7 +166,7 @@ prevent the user from collapsing the items.
 						<OpenInNew :size="20" />
 					</template>
 				</NcActionLink>
-			</template>		
+			</template>
 			<template>
 				<NcAppNavigationItem name="AppNavigationItemChild1" />
 				<NcAppNavigationItem name="AppNavigationItemChild2" />
@@ -189,7 +189,7 @@ prevent the user from collapsing the items.
 			Pencil,
 		},
 	}
-	</script>	
+	</script>
 ```
 
 ### Editable element
--- a/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
+++ b/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
@@ -137,15 +137,59 @@ Just nest the counter in a template within `<NcAppNavigationItem>` and add `#cou
 Wrap the children in a template with the `slot` property and use the prop `allowCollapse` to choose wether to allow or
 prevent the user from collapsing the items.
 
-```
-<NcAppNavigationItem name="Item with children" :allowCollapse="true" :open="true">
+```vue
 	<template>
-		<NcAppNavigationItem name="AppNavigationItemChild1" />
-		<NcAppNavigationItem name="AppNavigationItemChild2" />
-		<NcAppNavigationItem name="AppNavigationItemChild3" />
-		<NcAppNavigationItem name="AppNavigationItemChild4" />
+		<NcAppNavigationItem name="Item with children" :allowCollapse="true" :open="true">
+			<template #icon>
+				<Folder :size="20" />
+			</template>
+			<template #counter>
+				<NcCounterBubble>
+					99+
+				</NcCounterBubble>
+			</template>
+			<template #actions>
+				<NcActionButton @click="alert('Edit')">
+					<template #icon>
+						<Pencil :size="20" />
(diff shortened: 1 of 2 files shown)


Jerome-Herbinet (approved):
LGTM after a fast test :
[Capture vidéo du 17-05-2023 14:10:04.webm](https://github.com/nextcloud/nextcloud-vue/assets/33763786/5ba13ea7-d9f2-4e34-8265-3f0a846f0da2)

May some other reviewers test it ?

GretaD:
if this is the result we want, then works for me too
Screenshot from 2023-05-17 15-07-28

GretaD (approved):

raimund-schluessler:
> if this is the result we want, then works for me too Screenshot from 2023-05-17 15-07-28

I would tend to say that the collapse button should be on the very right:
Screenshot 2023-05-17 at 15-18-59 Nextcloud Vue Style Guide

Minor reordering of the elements would make this work.

raimund-schluessler:
> > Minor reordering of the elements would make this work.
> 
> Can you submit a patch? I am not sure where to adjust this...

I can try, but I don't have a dev environment at hand right now.

szaimen:
> > > Minor reordering of the elements would make this work.
> > 
> > 
> > Can you submit a patch? I am not sure where to adjust this...
> 
> Should work now. See the enhanced example in the docs. I didn't test it in an app though.

Thanks a lot! :)

src/components/NcAppNavigationItem/NcAppNavigationItem.vue
```diff
@@ -859,20 +903,21 @@ export default {
...
 	&.app-navigation-entry--no-icon,
 	&:hover, &:focus {
 		a .app-navigation-entry-icon {
 			visibility: visible;
 		}
 		.icon-collapse {
-			//shows the triangle button
-			visibility: visible;
+			// shows the triangle button
+			display: initial;
```
szaimen, line 920: I think one of this changes now reverted the idea of https://github.com/nextcloud/nextcloud-vue/pull/4103...

szaimen: cc @Jerome-Herbinet

szaimen: Really? Interesting!

szaimen: I see

jancborchardt (changes requested):
Like noted in the other PRs: It would be good if the arrow is pointing down also in closed state.
It pointing to the right is unexpected – pointing down in both states would make it look like a standard HTML select element.

raimund-schluessler:
> Like noted in the other PRs: It would be good if the arrow is pointing down also in closed state. It pointing to the right is unexpected – pointing down in both states would make it look like a standard HTML select element.

I wouldn't say it's entirely unexpected. Even Google still uses this pattern in their apps sometimes (e.g. Googlemail Web). But they indeed seem to switch to an up/down arrow for open/closed state, e.g. in their expandable list view:
https://m3.material.io/components/lists/guidelines#f6fe9c4e-7498-420a-b820-d12a2b59fc99
unnamed

So, how about we do this as well? Menu collapsed: arrow down, Menu open: arrow up?

raimund-schluessler:
> So, how about we do this as well? Menu collapsed: arrow down, Menu open: arrow up?

I implemented it like this. Please have a look.

jancborchardt (approved):
Yep, looks good now!

raimund-schluessler:
> The dropdown icon when collapsed is not shown permanently:
> without hover 	with hover
> image 	image
> 
> cc @Jerome-Herbinet

How about we simply show it all the time (when collapsed and open). I would find it confusing to only show it when collapsed and hide it when open.

src/components/NcAppNavigationItem/NcAppNavigationItem.vue
```diff
@@ -859,20 +903,21 @@ export default {
...
 /* Makes the icon of the collapsible element disappear
-*  When hovering on the root element */
+ * When hovering on the root element
+ */
 .app-navigation-entry--collapsible {
-	//shows the triangle button
+	// hides the triangle button
 	.icon-collapse {
-		visibility: hidden;
+		display: none;
```
szaimen, line 911: @raimund-schluessler can the above be addressed by simply setting this to display:initial?

PR #4135, opened 2023-05-17
Fix alignment of nav contents

pulsejet:
Padding was only applicable to the list, but should apply to the footer too.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
footer-before | footer_after

### DIFF
--- a/src/components/NcAppNavigation/NcAppNavigation.vue
+++ b/src/components/NcAppNavigation/NcAppNavigation.vue
@@ -174,7 +174,7 @@ export default {
 	}
 
 	//list of navigation items
-	& > ul,
+	&__content > ul,
 	&__list {
 		position: relative;
 		height: 100%;


pulsejet:
This doesn't play nice with scrollbars ...

Pytal (approved):

susnux (approved):

pulsejet:
What's the general policy on merging after approval? Is it okay to merge to master after 2 approvals or wait for the maintainer to do it?

src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -174,7 +174,7 @@ export default {
 	}
 
 	//list of navigation items
-	& > ul,
```
raimund-schluessler, line 177: @pulsejet Are you sure we can remove this rule? Maybe we should keep it, just in case? :thinking:

raimund-schluessler (approved):

PR #4152, opened 2023-05-24
collapse sections by default to make the styleguide a bit lucid

szaimen:
Problem: currently all sections are uncollapsed by default which makes each page very long and you cannot filter easily with your eyes without having to scroll a lot. Solution: collapse all sections by default which allows for easier filtering and scrolling. Then you can still view the sections that you are interested in by clicking the buttons.

One example:
| Before | After |
|---|---|
| nextcloud-vue-components netlify app_ |  deploy-preview-4152--nextcloud-vue-components netlify app_ |

### DIFF
--- a/styleguide.config.js
+++ b/styleguide.config.js
@@ -45,8 +45,8 @@ module.exports = async () => {
 			},
 		}),
 
-		exampleMode: 'expand',
-		usageMode: 'expand',
+		exampleMode: 'collapse',
+		usageMode: 'collapse',
 
 		components: 'src/components/*/*.vue',
 		getComponentPathLine(componentPath) {


jancborchardt (approved):
@szaimen I think you switched around the before/after screenshots ("after" should be all collapsed)?

But yes, as discussed I totally agree, sooooo much better and easier to find what you need, and then you can click to view the props and code of what you want.

nickvergessen:
I always wondered why it's shown by default and then collapsible :D

nickvergessen (approved):

nickvergessen:
We can't expand it when a specific section was linked, can we?

szaimen:
> We can't expand it when a specific section was linked, can we?

I fear I have not seen a feature in the styleguide docs that allows this :/

PR #4157, opened 2023-05-25
fix: make property "force-title" work

datenangebot:
- "name" and "title" were mixed up, so no titles were shown even if option "force-names" was set

### ☑️ Resolves

* Fix #…

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
SCR-20230525-mfxw | SCR-20230525-mfnv

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -153,11 +153,11 @@ export default {
 </script>
 ```
 
-### Multiple actions with 2 items inline AND forced names
+### Multiple actions with 2 items inline AND forced titles
 
 ```vue
 <template>
-	<NcActions :force-name="true" :inline="2">
+	<NcActions :force-title="true" :inline="2">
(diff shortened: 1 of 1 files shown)


raimund-schluessler:
This is not a mix-up. This is a breaking change scheduled for nextcloud-vue@8. Please see https://github.com/nextcloud/nextcloud-vue/pull/4106.

raimund-schluessler:
> > This is not a mix-up. This is a breaking change scheduled for nextcloud-vue@8. Please see #4106.
> 
> Ah, okay. Great. So the docs are faster than the releases... The docs tell me to use `:force-name="true"`, but that does not work.

Yes, that is indeed a bit unfortunate, but we have to live with it at the moment. The docs target master, which is unreleased and will become nextcloud-vue@8.

The latest release is 7.11.6 from the `stable7` branch, where #4106 is not included and you still have to use `force-title`.

PR #4158, opened 2023-05-25
adjust active eleement visibility in nclistitem

szaimen:


szaimen:
@marcoambrosini can you help me to get an active element as example into the docs? This would make testing a bit easier :)

nimishavijay:
Looks good in the screenshot, but as Marco said the dot needs to be adjusted, it can be changed to white :)

jancborchardt:
Looks good, and yep as @nimishavijay said switching the dot to white on the active element is good, so it’s symmetric to the "message for you" counter bubble color.

marcoambrosini:
@szaimen just add the `.active` class to the outer element :)

szaimen:
> @szaimen just add the `.active` class to the outer element :)

thanks for the pointer! Works great! :)

JuliaKirschenheuter:
Probably i haven't understood a problem, but why we could't add an `.active` class to three dots icons if an element is active?

szaimen:
> Probably i haven't understood a problem, but why we could't add an `.active` class to three dots icons if an element is active one?

I fear the problem is that this should probably be done automatically for embedded components? Also there is not active state for nccounterbubble for example afaik...

marcoambrosini:
Shouldn't we add this active state to both elements?

marcoambrosini:
I take that back @szaimen, I misunderstood the context

szaimen:
@JuliaKirschenheuter would you mind taking this over? I am unfortunatley stuck on how to proceed here.

marcoambrosini:
I can take over :)

szaimen:
> I can take over :)

Thanks a lot Marco! 💙🙏

marcoambrosini:
I just reworked a bit. @skjnldsv @raimund-schluessler @susnux  do you think this would cover all the use cases? Maybe we should also include a prop to force-primary style in the cases where the element is not a router link? 

Screenshot 2023-08-11 at 13 46 55

szaimen:
> I just reworked a bit. @skjnldsv @raimund-schluessler @susnux do you think this would cover all the use cases?

The screenshot looks good to me :)

Can you commit your changes? They arent yet afaics

jancborchardt (approved):
Looks good design-wise! :)

marcoambrosini:
@skjnldsv it seems I can't push to this repo, could you please grant me access? 

Screenshot 2023-08-16 at 13 27 42

skjnldsv:
> we should also include a prop to force-primary style

an `active` prop would be nice sure :)

skjnldsv:
> @skjnldsv it seems I can't push to this repo, could you please grant me access?


You failed to accept the invitation, it's sent again. :rocket: 
https://github.com/orgs/nextcloud-libraries/invitation

marcoambrosini:
@szaimen could you please take a look? :)

szaimen:
Tested in the docs and on hover the 3-dots button looks like it does not fit the contrast requirements:
image

jancborchardt:
@szaimen yeah, the icon needs to invert then to the same as text color (white in the case of default blue primary color).

marcoambrosini:
That is because the element is not activated unless the route is matched. I'm adding a "forceactive" prop now

marcoambrosini (approved):

szaimen (commented):
Tested and seem to work however the nclistitem compact mode looks a bit strange now. Not sure though if this is caused by 

### this change:
This PR: image
### master: 
image

marcoambrosini:
Should be fixed now @szaimen!

szaimen (commented):
LGTM! :)

Issue #2027, opened 2021-06-10
Text string in "Comments" text box


Valdnet:
Steps:

1. Select file details.
2. Go to the "Comments" tab.

The text string in the "Comments" text box is too long. In English, it ends with an arrow, while in other translations it can go beyond the text field.

![obraz](https://user-images.githubusercontent.com/47037905/121510094-94711780-c9e7-11eb-8fd0-2d46e044adaa.png)

![obraz](https://user-images.githubusercontent.com/47037905/121510258-ba96b780-c9e7-11eb-9101-c6c7d6d5f1a5.png)

App:
Nextcloud 22.0.0 beta 2 and 21.0.2
Comments 1.11.0

Browser:
Firefox 89.0

solracsf:
Fixed by https://github.com/nextcloud/server/pull/40294 (v28 only)

ShGKme:
Fixed by #4408

Issue #3853, opened 2023-03-03
Placeholder size not taken into account when sizing


SystemKeeper:
Talk 16:
![image](https://user-images.githubusercontent.com/1580193/222824742-3bb66273-85e2-4c45-8fdc-06cf02199d8f.png)

Talk 15:
![image](https://user-images.githubusercontent.com/1580193/222824861-fad6c9d3-d47c-46c5-addb-5fbfa99d91a1.png)

When testing on https://nextcloud-vue-components.netlify.app/#/Components/NcRichContenteditable this happens:
![image](https://user-images.githubusercontent.com/1580193/222825820-798a88bb-b3b2-406a-81ef-66601c0eb4c0.png)

ChristophWurst:
Server/comments
![Bildschirmfoto vom 2023-04-19 15-01-22](https://user-images.githubusercontent.com/1374172/233083479-fd01752c-b23b-4269-ba31-df0b000fe614.png)

szaimen:
As reference: https://github.com/nextcloud/nextcloud-vue/pull/2601

SystemKeeper:
Position absolute was introduced only recently at https://github.com/nextcloud/nextcloud-vue/pull/3832

PR #3832, opened 2023-02-28
Fix rich content placeholder position

raimund-schluessler:
This fixes the position of the cursor in the `NcRichContenteditable` component after one types text and deletes it again. I also adjusted the padding a bit, to bring the text to the center. Closes #3828.

Before:
Before

After:
After

PR #4163, opened 2023-05-27
fix: comments box overlapping

solracsf:
### Resolves

Fix https://github.com/nextcloud/nextcloud-vue/issues/2027

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/NcRichContenteditable/NcRichContenteditable.vue
+++ b/src/components/NcRichContenteditable/NcRichContenteditable.vue
@@ -725,7 +725,6 @@ export default {
 	&--empty:before {
 		content: attr(placeholder);
 		color: var(--color-text-maxcontrast);
-		position: absolute;
 	}
 
 	&[contenteditable='false']:not(&--disabled) {


ShGKme (commented):
I'm not sure this is a good way to fix overlapping. The fix should not change the size of the block.
And it breaks the cursor position in FF fixed by https://github.com/nextcloud/nextcloud-vue/pull/3832

Maybe cut the text in a placeholder with an ellipsis? 

Also, the default placeholder is truly too long now with `:`, `@`, and `/` commands. Maybe we can make it shorter or display it not in the placeholder.

solracsf:
Yeah, that PR introduced this.
https://github.com/nextcloud/nextcloud-vue/issues/3853

Text can be cutted of since there are essential informations on it. But, another way would be to:
- Put a small text as placeholder like `New comment`
- and on mouse hover or similar, git all that info about `:`, `@`, and `/` commands

szaimen (approved):
Good from my side

solracsf:
Closing after https://github.com/nextcloud/server/pull/40294

PR #4217, opened 2023-06-14
fix(NcSelect): Disabled colors

Pytal:
### ☑️ Resolves

- Lighten the disabled background and text colors slightly to distinguish the disabled state from the regular state

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/NcSelect/NcSelect.vue
+++ b/src/components/NcSelect/NcSelect.vue
@@ -955,11 +955,11 @@ body {
 	--vs-line-height: var(--default-line-height);
 
 	/* Disabled State */
-	--vs-state-disabled-bg: var(--color-background-dark);
+	--vs-state-disabled-bg: var(--color-background-hover);
 	--vs-state-disabled-color: var(--color-text-maxcontrast);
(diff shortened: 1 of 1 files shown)


susnux:
Looks better but I think the visual difference is still not that good. How about also changing the (text) color for selected elements?

(e.g. like this, top is normal enabled and bottom is disabled)
image
```css
.vs--disabled  .vs__search, .vs--disabled .vs__selected {
}
```

szaimen:
> Looks better but I think the visual difference is still not that good. How about also changing the (text) color for selected elements?
> 
> (e.g. like this, top is normal enabled and bottom is disabled) image
> 
> ```css
> .vs--disabled  .vs__search, .vs--disabled .vs__selected {
>   color: var(--color-text-maxcontrast);
> }
> ```

I agree! The proposal looks good to me!

szaimen (approved):
LGTM

susnux (approved):

PR #4219, opened 2023-06-15
NcDatetime: New component for displaying timestamps as time relative from now

### LINKED ISSUE #13: Moment

### LINKED ISSUE #2089: Improve Moment.vue, e.g. to get short timestamps
To show the time in different apps, we use Moment.vue, which we create "locally" for a certain app. We dont have a general component for that. Would be nice if we have a nc/vue component and also to improve a bit the way we show the time/year.

Considering other clients out there, we do it a bit differently. We show minutes, hours, years as a full word.

Would be nice if we only show m for minutes, and then hours as _h_, maybe up to 5h or 10h and then we show the exact time instead of counting 24h. For the previews day, we show date and month (i.e Jul 6) and for year ( 11/04/2020 )

susnux:
### ☑️ Resolves

* Resolves #13
* Maybe resolves #2089

This implements a component showing for displaying timestamps like *x seconds ago* without the need of huge libraries like *moment.js*. It is based on the `Intl` browser API, which is supported by all of our browser targets.

Tests included.

### 🖼️ Screenshots

image

### 🏁 Checklist


susnux:
:information_source:  Fixed the tests by enforcing `UTC` as the timezone, sadly this needs to be done by setting the environment on the node script as setting
```
process.env.TZ = 'UTC'
```
does not to work

src/components/NcDatetime/NcDatetime.vue
```diff
@@ -0,0 +1,221 @@
...
+			required: true,
+		},
+		/**
+		 * The format used for displaying, or if relative time is used the format used for the title (optional)
+		 *
+		 * @type {Intl.DateTimeFormatOptions}
+		 */
+		format: {
+			type: Object,
+			default: () => ({ timeStyle: 'medium', dateStyle: 'short' }),
```
marcoambrosini, line 117: Wouldn't it make sense to separate these into two separate props?

marcoambrosini (approved):
Very nice addition. Tested in the docs and it works nicely

Would it be possible to have a prop that allows to show just "a few seconds ago" until the first minute instead of each and every second?

src/components/NcDatetime/NcDatetime.vue
susnux: Which two properties? title format and main format?
Or do you mean `timeStyle` and `dateStyle`?

If the latter one, then I do not think so. As there are more options possible I just allowed every `Intl.DateTimeFormatOptions` option (which allow finer control over the style by setting the format of `weekday, hour, month...` directly).

susnux:
> Would it be possible to have a prop that allows to show just "a few seconds ago" until the first minute instead of each and every second?

Sure, something like `:ignore-seconds="true"`?

susnux:
> Sure, something like `:ignore-seconds="true"`?

@marcoambrosini This is now implemented :)

skjnldsv (approved):

Chartman123:
Will this have any effect on https://github.com/nextcloud/nextcloud-moment/issues/407?

susnux:
> Will this have any effect on [nextcloud-libraries/nextcloud-moment#407](https://github.com/nextcloud-libraries/nextcloud-moment/issues/407)?

@Chartman123 Not really, `@nextcloud/moment` is still supported, this only adds a new component which solves the most common problem when displaying time(stamps). It is not based on moment because it is not required (would introduce much larger dependency and Moment.js is not developed anymore so for new projects it is not recommended).

PR #4221, opened 2023-06-15
Fix NcModal does set the background color but not the text color

### LINKED ISSUE #2405: Modal not dark mode compatible (on public pages)
The Modal component sets the background color without setting the font color.
However on public pages the font is always set black on the body, so in the modal you then have black on dark grey.
Bildschirmfoto von 2021-12-14 10-10-34

I think when setting the background color it should also automatically set the font color:
https://github.com/nextcloud/nextcloud-vue/blob/da0352c89c34cc8e58d09153560317c813d58cb7/src/components/Modal/Modal.vue#L721

susnux:
### ☑️ Resolves

* Fix #2405

As it sets the background color it should also set the foreground color as it can not assume the default will match the contrast criteria.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/NcModal/NcModal.vue
+++ b/src/components/NcModal/NcModal.vue
@@ -915,6 +915,7 @@ export default {
 		transition: transform 300ms ease;
 		border-radius: var(--border-radius-large);
 		background-color: var(--color-main-background);
+		color: var(--color-main-text);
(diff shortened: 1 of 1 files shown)


marcoambrosini (approved):
👌

raimund-schluessler (approved):

PR #4222, opened 2023-06-15
Fix: Leftovers of `name` prop migration + fix `@` in user ids (richEditor)

### LINKED ISSUE #1825: User mention in comment field does not work for users with @ in name
First I thougt its solved with NC21. Selction works but we have users with special chars as "@" in their usernames. When I select one of those users the "@" inside the username is removed and so the user mention does not work

_Originally posted by @roadrunnerjb in https://github.com/nextcloud/server/issues/19523#issuecomment-810150432_

susnux:
### ☑️ Resolves

* Resolves #1825
* First commit is a follow up on #4106 

The first commit solves some issues / leftovers where the property was renamed but the example code was not changed.
The second commit solves an issues were mentioning users with `@` in there userid was not possible (`@` was removed).
Also added a note about the property renaming on the changelog previously only `name` -> `title` was mentioned and not `label` -> `title`.

New tests included.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/NcRichContenteditable/NcAutoCompleteResult.vue
+++ b/src/components/NcRichContenteditable/NcAutoCompleteResult.vue
@@ -32,10 +32,10 @@
 			</div>
 		</div>
 
-		<!-- Label and subline -->
+		<!-- Title and subline -->
 		<span class="autocomplete-result__content">
-			<span class="autocomplete-result__label">
-				{{ label }}
+			<span class="autocomplete-result__title" :title="title">
+				{{ title }}
 			</span>
 			<span v-if="subline" class="autocomplete-result__subline">
 				{{ subline }}
@@ -51,7 +51,7 @@ export default {
 	name: 'NcAutoCompleteResult',
 
 	props: {
-		label: {
+		title: {
 			type: String,
 			required: true,
 		},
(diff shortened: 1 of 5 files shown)


src/components/NcRichContenteditable/NcAutoCompleteResult.vue
```diff
@@ -51,7 +51,7 @@ export default {
 	name: 'NcAutoCompleteResult',
 
 	props: {
-		label: {
+		title: {
```
raimund-schluessler, line 54: This change was intentional, because we agreed to use the term `title` only if if relates to the HTML attribute `title`, which it does not here.
Is there a deeper problem with renaming the prop here?

marcoambrosini (approved):

src/components/NcRichContenteditable/NcAutoCompleteResult.vue
```diff
@@ -32,10 +32,10 @@
 			</div>
 		</div>
 
-		<!-- Label and subline -->
+		<!-- Title and subline -->
 		<span class="autocomplete-result__content">
-			<span class="autocomplete-result__label">
-				{{ label }}
+			<span class="autocomplete-result__title">
```
Antreesy, line 37:
```suggestion
			<span class="autocomplete-result__title" :title="title">
```
The only thing I could suggest to justify this change, and to provide ability to see long names fully:
image

susnux:
@Antreesy nice! Add the title prop :)

Antreesy (approved):
:seal:

PR #4247, opened 2023-06-22
Use NcSelect in NcRichtext

julien-nc:
refs #3743 

Switch to `NcSelect` in `NcProviderList` and `NcSearch` from `NcRichText`.

I couldn't figure out a way to avoid an issue with NcSearch. When selecting a "Load more..." item, even if we actually select the item while searching for more results, the selection jumps to the top and then back to the "more" item.

Not so pretty...but not so bad.

https://github.com/nextcloud/nextcloud-vue/assets/11291457/90b2c16c-3789-4a07-af7d-5e265d01711c

### DIFF
--- a/src/components/NcRichText/NcReferencePicker/NcProviderList.vue
+++ b/src/components/NcRichText/NcReferencePicker/NcProviderList.vue
@@ -1,32 +1,33 @@
 <template>
 	<div class="provider-list">
-		<NcMultiselect ref="provider-select"
+		<NcSelect ref="provider-select"
 			v-model="selectedProvider"
 			class="provider-list--select"
-			track-by="id"
+			input-id="provider-select-input"
 			label="title"
 			:placeholder="multiselectPlaceholder"
 			:options="options"
-			:internal-search="false"
-			:clear-on-select="true"
-			:preserve-search="true"
-			:option-height="44"
-			@search-change="query = $event"
+			:append-to-body="false"
+			:clear-search-on-select="true"
+			:clear-search-on-blur="() => false"
+			:filterable="false"
+			@search="onSearch"
 			@input="onProviderSelected">
-			<template #option="{option}">
+			<template #option="option">
 				<div v-if="option.isLink" class="provider">
 					<LinkVariantIcon class="link-icon" :size="20" />
 					<span>{{ option.title }}</span>
 				</div>
 				<div v-else class="provider">
 					<img class="provider-icon"
-						:src="option.icon_url">
+						:src="option.icon_url"
+						:alt="providerIconAlt">
 					<NcHighlight class="option-text"
 						:search="query"
 						:text="option.title" />
 				</div>
 			</template>
-		</NcMultiselect>
+		</NcSelect>
 		<NcEmptyContent class="provider-list--empty-content">
 			<template #icon>
 				<LinkVariantIcon />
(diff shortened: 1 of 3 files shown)


Pytal (approved):

Pytal:
> I couldn't figure out a way to avoid an issue with NcSearch. When selecting a "Load more..." item, even if we actually select the item while searching for more results, the selection jumps to the top and then back to the "more" item.

lmk if it works with https://github.com/nextcloud-deps/vue-select/pull/20 :)

src/components/NcRichText/NcReferencePicker/NcProviderList.vue
```diff
@@ -80,9 +82,9 @@ export default {
 	},
 	methods: {
 		focus() {
-			this.$nextTick(() => {
-				this.$refs['provider-select']?.$el?.focus()
-			})
+			setTimeout(() => {
```
raimund-schluessler, line 85: Why a timeout? What happens if it takes longer?

src/components/NcRichText/NcReferencePicker/NcSearch.vue
```diff
@@ -189,14 +187,16 @@ export default {
 			this.resultsBySearchProvider = resultsBySearchProvider
 		},
 		focus() {
-			this.$refs['search-select']?.$el?.focus()
+			setTimeout(() => {
```
raimund-schluessler, line 190: Same as above.

raimund-schluessler (approved):
Looks good. Just the timeout issue makes me wondering.

src/components/NcRichText/NcReferencePicker/NcProviderList.vue
julien-nc: > Why a timeout?

Because it does not work when doing it on the next tick.

> What happens if it takes longer?

It fails to focus :grin: 

Do you know how focusing on the input element could be improved?

julien-nc:
> lmk if it works with https://github.com/nextcloud-deps/vue-select/pull/20 :)

Interesting. I'll check if it helps.

julien-nc:
@Pytal Much better with `resetFocusOnOptionsChange` :blue_heart: Looking forward to it.
It's used and set to `false` in NcSearch even if it's not in NcSelect yet. Are we fine with that?


https://github.com/nextcloud/nextcloud-vue/assets/11291457/6eae59da-9245-4328-af88-ea214606311a

julien-nc:
Added another commit to make sure the option keys are unique. (It's just setting the `id` attribute of the option objects, I didn't wanna define the [getOptionKey prop](https://vue-select.org/api/props.html#getoptionkey)).

Pytal:
> @Pytal Much better with `resetFocusOnOptionsChange` 💙 Looking forward to it. It's used and set to `false` in NcSearch even if it's not in NcSelect yet. Are we fine with that?

Great! We can merge the bump https://github.com/nextcloud/nextcloud-vue/pull/4264 too!

Pytal (approved):

susnux (approved):
👍

src/components/NcRichText/NcReferencePicker/NcProviderList.vue
raimund-schluessler: If nextTick does not work, I don't have a better solution atm.

PR #4266, opened 2023-06-25
fix(NcModal): fix focus-trap fallback

### LINKED ISSUE #4265: Uncaught (in promise) Error: Your focus-trap must have at least one container with at least one tabbable node in it at all times
```js
Uncaught (in promise) Error: Your focus-trap must have at least one container with at least one tabbable node in it at all times
    at updateTabbableNodes (focus-trap.esm.js:328:1)
    at Object.activate (focus-trap.esm.js:590:1)
    at NcModal.js:2:1
    at u (NcModal.js:2:1)
    at Generator.<anonymous> (NcModal.js:2:1)
    at Generator.next (NcModal.js:2:1)
    at j (NcModal.js:2:1)
    at r (NcModal.js:2:1)
```

https://github.com/focus-trap/focus-trap/blob/a4b3948c7fce410841dc247227cbaa1292d2efcb/index.js#L286-L355

cc @susnux as you worked on this area before?

skjnldsv:
Fix #4265 

Sometimes the content is empty or not ready yet.

> `fallbackFocus` {HTMLElement | SVGElement | string | () => HTMLElement | SVGElement | string}: By default, an error will be thrown if the focus trap contains no elements in its tab order. With this option you can specify a fallback element to programmatically receive focus if no other tabbable elements are found. For example, you may want a popover's <div> to receive focus if the popover's content includes no tabbable elements. Make sure the fallback element has a negative tabindex so it can be programmatically focused. The option value can be a DOM node, a selector string (which will be passed to document.querySelector() to find the DOM node), or a function that returns any of these.

### DIFF
--- a/src/components/NcModal/NcModal.vue
+++ b/src/components/NcModal/NcModal.vue
@@ -180,7 +180,8 @@ export default {
 			role="dialog"
 			aria-modal="true"
 			:aria-labelledby="'modal-name-' + randId"
-			:aria-describedby="'modal-description-' + randId">
+			:aria-describedby="'modal-description-' + randId"
+			tabindex="-1">
 			<!-- Header -->
 			<transition name="fade-visibility" appear>
 				<div class="modal-header">
@@ -707,11 +708,14 @@ export default {
 			// wait until all children are mounted and available in the DOM before focusTrap can be added
 			await this.$nextTick()
 
-			// Init focus trap
-			this.focusTrap = createFocusTrap(contentContainer, {
+			const options = {
 				allowOutsideClick: true,
+				fallbackFocus: contentContainer,
 				trapStack: getTrapStack(),
-			})
+			}
+
+			// Init focus trap
+			this.focusTrap = createFocusTrap(contentContainer, options)
 			this.focusTrap.activate()
 		},
 		clearFocusTrap() {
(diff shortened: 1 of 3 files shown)


susnux:
@skjnldsv something is wrong with the package lock: *20,786 additions, 2 deletions*
Are you using the correct node version?

src/components/NcModal/NcModal.vue
```diff
@@ -707,11 +707,14 @@ export default {
 			// wait until all children are mounted and available in the DOM before focusTrap can be added
 			await this.$nextTick()
 
-			// Init focus trap
-			this.focusTrap = createFocusTrap(contentContainer, {
+			const options = {
 				allowOutsideClick: true,
+				fallbackFocus: contentContainer,
```
susnux, line 712: > Make sure the fallback element has a negative tabindex so it can be programmatically focused.

The mask (== content container) does not have a negative tab index.

skjnldsv: What if there is some focusable elements?
Will it still focus the first element or the mask then?

skjnldsv: Pushed

skjnldsv:
> @skjnldsv something is wrong with the package lock: _20,786 additions, 2 deletions_ Are you using the correct node version? As the lockfile version is also downgraded from 3 to 2 with this update.

Nice catch, haven't updated node on this terminal since the LTS merge

skjnldsv:
Preview looks good, and content still focusable :+1:

susnux (approved):

szaimen (approved):

skjnldsv:
https://github.com/nextcloud/nextcloud-vue/pull/4270

PR #4271, opened 2023-06-26
test if adjusting the font-face works

szaimen:


susnux:
I guess no longer needed?

PR #4274, opened 2023-06-26
fix: Ensure `extract-l10n` extracts strings from properties

### LINKED ISSUE #2062: Translations in Transifex
I noticed that there are a lot of text strings that are not translated in Transifex. The lack of translations results in a slightly underdeveloped NC instance.

@skjnldsv Is it possible to easily add these strings to be translated? This is a priority issue 😉.

susnux:
### ☑️ Resolves

* Resolves #2062 

There are still some (two at this moment) string which can not be extracted because they are used within properties like `:aria-label="t(...)"`.
To resolve this we need to use the HTML extractor on vue SFC files.

The huge diff is from applying our code style rules (tab vs spaces).

*BTW I tried `gettext-extractor-vue` written by the gitlab team, but that does not work as it converts all `t()` calls to `_t()` or different (compile the SFC), so the extractor does not know which strings to extract.*

### 🏁 Checklist

### DIFF
--- a/build/extract-l10n.js
+++ b/build/extract-l10n.js
@@ -1,28 +1,33 @@
-const { GettextExtractor, JsExtractors } = require('gettext-extractor');
+const { GettextExtractor, JsExtractors, HtmlExtractors } = require('gettext-extractor')
 
-let extractor = new GettextExtractor();
+const extractor = new GettextExtractor()
 
-extractor
-    .createJsParser([
-        JsExtractors.callExpression('t', {
-            arguments: {
-                text: 0,
-            }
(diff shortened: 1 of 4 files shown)


raimund-schluessler:
> There are still some (two at this moment) string which can not be extracted because they are used within properties like `:aria-label="t(...)"`.

As a workaround we can use computed props there. We do that in other places already.

skjnldsv (approved):

skjnldsv:
Nice!

raimund-schluessler (approved):

PR #3529, opened 2022-11-30
Migrate to vite / vitest / vitepress for building

susnux:
Created because of this comment: https://github.com/nextcloud/nextcloud-vue/pull/2959#issuecomment-1331760883

---

Based on [this comment](https://github.com/nextcloud/nextcloud-vue/pull/2625#issuecomment-1327605238) I migrated the project to use vite for bundling / building (without noticing this is already done in https://github.com/nextcloud/nextcloud-vue/pull/2959).

* [x] Migrate to [vite](https://vitejs.dev/) for building (so dependency on webpack anymore)
* [x] Migrate to [vitest](https://vitest.dev/) for unit tests (instead of jest)
* [x] Migrate to [vue-docgen](https://vue-styleguidist.github.io/docs/docgen-cli.html) + [vitepress](https://vitepress.vuejs.org/) for documentation
* [x] Migrate cypress to use vite (and updated cypress) 

One problem with this migration is the missing live preview of the Nc components, as `vitepress` is using vue3 but we are still using vue2. This leads dependency issues.
(I got it working locally but it requires experimental node features (import maps) to mount vue2 components within vitepress vue3).

PR #4276, opened 2023-06-26
Fix module entry point by migrating to vite for ESM building

susnux:
### ☑️ Resolves

This introduces vite for building the module entry point instead of webpack.
1. The previously generated module by webpack is **invalid** as it still uses `require` instead of `import` which is not defined for ES modules.
2. Build time is much shorter
3. The resulting file is only 600 kiB instead of 930 kiB

*(Different from #3529 as this just introduces vite for module building to **fix** the assets)*
### 🏁 Checklist


skjnldsv:
> Cannot we make a full migration to Vite? Or fix the problem with Webpack config, if possible.
> 
> I'm a bit afraid of having 2 (almost 3) different bundlers at the same time in terms of config support.

I think the goal is to do a step-by-step migration :)

susnux:
> Cannot we make a full migration to Vite?

That was / is my plan, but currently it does not work out that good because styleguide does not work that well with vite currently (see #3529 and #3529 ).
So I wanted to first migrate this part to fix the module bundle, and then later migrate the other part.

skjnldsv (approved):

package.json
```diff
@@ -15,9 +15,9 @@
   "scripts": {
     "dev": "webpack --node-env development --progress",
     "watch": "webpack --node-env development --progress --watch",
-    "watch:module": "LIBRARY_TARGET=module webpack --node-env development --progress --watch",
+    "watch:module": "NODE_ENV=production vite build --watch",
```
skjnldsv, line 18:
```suggestion
    "watch:module": "vite build --mode production --watch",
```

package.json
```diff
@@ -15,9 +15,9 @@
...
     "build": "webpack --node-env production --progress && npm run build:module",
-    "build:module": "LIBRARY_TARGET=module webpack --node-env production --progress",
+    "build:module": "NODE_ENV=production vite build",
```
skjnldsv, line 20:
```suggestion
    "build:module": "vite --mode production build",
```

package.json
```diff
@@ -32,7 +32,17 @@
...
+  "exports": {
+    ".": {
+      "import": "./dist/index.mjs",
+      "require": "./dist/index.js"
+    },
+    "./dist/install.js": "./dist/install.js",
+    "./dist/Components/*.js": "./dist/Components/*.js",
+    "./dist/Directives/*.js": "./dist/Directives/*.js",
+    "./dist/Functions/*.js": "./dist/Functions/*.js",
+    "./dist/Mixins/*.js": "./dist/Mixins/*.js"
```
skjnldsv, line 44:
```suggestion
    "./dist/": "./dist/",
```

package.json
```diff
@@ -57,6 +67,7 @@
     "@vueuse/components": "^10.0.2",
     "@vueuse/core": "^10.1.2",
     "clone": "^2.1.2",
+    "core-js": "^3.31.0",
```
skjnldsv, line 70: Is i really needed?

skjnldsv (changes requested):

cypress.config.mjs
```diff
@@ -1,11 +1,14 @@
 import { defineConfig } from 'cypress'
```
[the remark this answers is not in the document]
skjnldsv: I'd like to keep typescript here please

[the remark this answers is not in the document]
skjnldsv: I'd like to keep typescript here please.
I see the `esm` import issue in the commit comment, but I fail to see how that's relevant to this PR :thinking: 
We might not even need to specify the `__dirname` at all :)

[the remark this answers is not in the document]
susnux: > Why the rename from ts to mjs?

Short:
Because I made the translations a module.

Long:
Because `ts` is the same as the package type, so for the package it is commonjs Typescript. Every `import` is a `require`. Not sure if Cypress or webpack is the problem but even when changing the `webpackConfig` part to dynamically import `translations.mjs` it leads to.
> Error [ERR_REQUIRE_ESM]: require() of ES Module nextcloud-vue/resources/translations.mjs not supported.

Moreover Cypress supports `mjs` but not `mts`.
So either revert making the `resources/translations` helper a module or make the cypress config a module js file.

package.json
susnux: Yes to resolve the polyfills added by the corejs rollup plugin.

(I still think we should not add them in libraries but either globally or the calling app, but that is still to discuss https://github.com/nextcloud/standards/issues/10 )

package.json
susnux: BTW here it does work as we want to build in production to prevent that vue2 issue, but generally it does not work, as `--mode development` does not set the `NODE_ENV` (it sets it to 'production').

skjnldsv: Do you need the `NODE_ENV` ?

susnux: Thats how e.g. vue decides to show more verbose warnings

cypress.config.mjs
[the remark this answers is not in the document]
susnux: But yes that is unrelated to this PR and I will revert that change.

package.json
skjnldsv: Weird, it should also be assigning NODE_ENV
https://github.com/vitejs/vite/blob/2b1ffe86328f9d06ef9528ee117b61889893ddcc/packages/vite/src/node/plugins/define.ts#L27

skjnldsv: Just tested it and the NODE_ENV is indeed set properly with `--mode`
https://codesandbox.io/p/sandbox/throbbing-sky-x6gpw2

package.json
skjnldsv: None of the other apps/libs ships core-js
We removed it in the past, why do we want it back? What is different with vite that makes it necessary compared to webpack here?

package.json
susnux: This only works when **not** building in library mode:
https://github.com/vitejs/vite/blob/2b1ffe86328f9d06ef9528ee117b61889893ddcc/packages/vite/src/node/plugins/define.ts#L19

That is why I need to add those defines manually in the vite config:
https://github.com/susnux/nextcloud-vite-config/blob/91b2b564c6894daefe1ffe035a3d6fd339054c03/lib/baseConfig.ts#L112

And you can test that quite easy: Create a basic vite config, put `console.log(process.env)` on top of the config and run `vite --mode development build`:
```
% npx vite -c a.js --mode development build
mode is: production
vite v4.3.9 building for development...
✓ 1 modules transformed.
dist/test.mjs  0.06 kB │ gzip: 0.08 kB
✓ built in 53ms
```

```
% cat a.js
console.log("mode is:", process.env.NODE_ENV)
export default {
        build: {
                lib: {
                        entry: 'b.js',
                        formats: ['es'],
                },
        },
}
```

package.json
susnux: See comment above, resolved now :)

package.json
skjnldsv: > This only works when **not** building in library mode: [vitejs/vite@`2b1ffe8`/packages/vite/src/node/plugins/define.ts#L19](https://github.com/vitejs/vite/blob/2b1ffe86328f9d06ef9528ee117b61889893ddcc/packages/vite/src/node/plugins/define.ts#L19)

I love it ahaha, This is such a specific issue! Why do we always end up in the rabbit hole here :sob: :rofl:

cypress.config.mjs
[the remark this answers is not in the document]
susnux: Done, but it is not that easy see the commit message of the second commit.

package.json
susnux: Done with nextcloud-vite-config beta 13.

package.json
susnux: Oh. I really thought our babel config was set to `useBuiltIns: 'usage'`?
Well in this case lets disable polyfilling here :broom:

package.json
susnux: Not sure about this. When we build also the other entry points using vite we will get shared chunks which we probably do not want to export.
But for the moment I guess we can do it that way.

skjnldsv: I think it's more flexible yes

vite.config.mts
```diff
@@ -0,0 +1,79 @@
...
+
+// We need a callback config so we can access the vite build mode
+export default defineConfig((env) => {
+
+	const createConfig = createLibConfig(entryPoints, {
+		// Add our overrides to the config
+		config: overrides,
+		// Packages with paths imports should be added here to mark them as external as well
+		nodeExternalsOptions: {
+			include: [/^vue-material-design-icons\//, /^@nextcloud\/l10n\//, /^@mdi\/svg\//],
```
skjnldsv, line 65: Maybe we can consider marking all our packages as external `/^@nextcloud\/.*\//`

skjnldsv (approved):

vite.config.mts
susnux: :+1: 

They are external by default, but only if you import from them like `import { foo } from '@nextcloud/foo'` so only if you do not append any path like `@nextcloud/l10n/gettext`.

.eslintrc.js
```diff
@@ -14,4 +14,11 @@ module.exports = {
 	plugins: [
 		'cypress',
 	],
+	parserOptions: {
+		babelOptions: {
+		  plugins: [
+				'@babel/plugin-syntax-import-assertions',
+		  ],
+		},
+	},
```
juliushaertl, line 23: Indentation seems off ;)

susnux: Fixed :)

juliushaertl (approved):

PR #4287, opened 2023-06-28
Feature: Add visible inner label for `NcInputField`

### LINKED ISSUE #4126: RFC: Visible internal label for `NcInputField`s within the element
I pretty sure this is nothing new, probably someone already uses something like this, but I wanted to know your opinions on this idea:

Having a visual label is most often required, but also often the placeholder is misused for it.
As placeholders should specify what input is correct or expected but not necessarily what field it currently is.

I also noticed that our `NcSelect` does now has a height of 48px vs all other input elements (36px), sometimes this looks odd together so I resize inputs to match the 48px. Which gives 12px extra space.

Combining this two things I came up with the idea to have the label be inside the element and only shrink and move when entering text.

So this is how it could look like (just the "Old", "Good new" and "Disabled" ignore the other ones):
mockup

susnux:
### ☑️ Resolves

* Resolves #4126 

The inner label is now visible, to disable any label set `label-outside` to true, `label-visible` property is removed.
The label is shown instead of the placeholder if the input is empty and not focused, when focused the label is moved up and scaled but always shown.

Also the action input is adjusted for this change.

#### NcInputField
new

#### NcActionInput
new-action

### 🚧 Tasks

- Maybe adjust the datetime picker to this changes in a followup, @nextcloud/designers ?

### 🏁 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ All notable changes to this project will be documented in this file.
   - `label` property was renamed to `name` for `NcMentionBubble`
   - `name` propery is now required for `NcActions*`, `NcAppNavigationItem` and `NcBreadcrumb*`
   - See linked pull request for full migration guide
+- `NcInputField`: The `labelVisible` property was removed for accessibility it is required to always show a label.
+  You can still use the `labelOutside` property to remove the inner label from the component.
 
 ### :rocket: Enhancements
 - enh\(NcDatetime\): New component for displaying timestamps as time relative from now [\#4219](https://github.com/nextcloud-libraries/nextcloud-vue/pull/4219) ([susnux](https://github.com/susnux))
--- a/src/components/NcActionInput/NcActionInput.vue
+++ b/src/components/NcActionInput/NcActionInput.vue
@@ -23,6 +23,10 @@
 
 <docs>
 This component is made to be used inside of the [NcActions](#NcActions) component slots.
+
+It is recommended to not only provide a placeholder, but also a label.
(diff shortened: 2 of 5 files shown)


dartcafe:
I like it, but shouldn't there be a style guide to prevent, that every developer just follows his personal preference?

@ChristophWurst

marcoambrosini (commented):
@susnux  I still think that the label should move on the border as this will conflict with leading icons like search can we still change this?

susnux:
@marcoambrosini sure, I just got not much time to adjust this here, but it is still on my list :)

st3iny (commented):
Looks good to me. Is there anything left to be done/discussed?

susnux:
@marcoambrosini 
>  I still think that the label should move on the border as this will conflict with leading icons like search can we still change this?

I created an alternative with the label inside the border here: https://github.com/nextcloud-libraries/nextcloud-vue/pull/4394 . So you can compare both on the preview deployments.

I personally think that the label within the border is cleaner, is allows to have multiple input elements (input field, buttons, NcSelect) side by side and all having the same "border-height".
Also I do not think that the label conflicts with the icon, as the label will always be shorter (due to font size) when shown above than on initial state.

But this is a decision for designers :)

AndyScherzinger:
Label on the border would also match Material Design 2 and 3 btw which seems to be what we usually try to use or be inspired by.

susnux:
Close in favor of https://github.com/nextcloud-libraries/nextcloud-vue/pull/4394

PR #4344, opened 2023-07-18
feat(NcButton): Add `pressed` state for stateful buttons

susnux:
### ☑️ Resolves

* Alternative to #4280
* Resolves https://github.com/nextcloud/server/issues/37109

This simply adds a pressed state and enforces `primary` on pressed, and any other type (except primary) otherwise.
Not we should probably remove the button variant of the checkboxradioswitch for not grouped elements to prevent different styled components with the same function.

### 🖼️ Screenshots

https://github.com/nextcloud/nextcloud-vue/assets/1855448/b6716849-64ce-4bfb-93ac-1c28a0302db5

### 🏁 Checklist


src/components/NcButton/NcButton.vue
```diff
@@ -196,6 +196,66 @@ button {
...
+	computed: {
+		ariaLabel() {
+			return this.isFavorite ? 'Remove as favorite' : 'Add as favorite'
+		},
+	},
+	methods: {
+		toggleFavorite() {
+			this.isFavorite = !this.isFavorite
+		},
+	},
```
ShGKme, line 255: What about adding `update:pressed` event to the component?
Then `pressed` state will be synced and component's user won't need to handle clicks manually.

src/components/NcButton/NcButton.vue
```diff
@@ -352,6 +440,15 @@ export default {
...
 					...this.$listeners,
 					click: ($event) => {
+						// Update pressed prop on click if it is set
+						if (typeof this.$props.pressed === 'boolean') {
+							/**
+							 * Update the current pressed state of the button (if the `pressed` property was configured)
+							 *
+							 * @property {boolean} newValue The new `pressed`-state
+							 */
+							this.$emit('update:pressed', !this.$props.pressed)
```
ShGKme, line 450: Let's also list the event in the `emits` option

src/components/NcButton/NcButton.vue
```diff
@@ -352,6 +440,15 @@ export default {
 				on: {
 					...this.$listeners,
 					click: ($event) => {
+						// Update pressed prop on click if it is set
+						if (typeof this.$props.pressed === 'boolean') {
```
ShGKme, line 444:
```suggestion
						if (typeof this.pressed === 'boolean') {
```

src/components/NcButton/NcButton.vue
ShGKme, line 450:
```suggestion
							this.$emit('update:pressed', !this.pressed)
```

src/components/NcButton/NcButton.vue
ShGKme: It seems to me, we always use shorthand props access

susnux: ok

marcoambrosini (approved):
Nice work, thank you :)

JuliaKirschenheuter (approved):
haven't tested but looks good

ShGKme (approved):
Tested, works fine, looks good by code

ShGKme:
Cypress tests seems to be related

jancborchardt (approved):
Nice, looks good! :)

susnux:
> Cypress tests seems to be related

Yes as this is a visual change, the snapshots need to be updated, but the other tests pass :)

susnux:
Oh yes, I love that cypress on my system renders the snapshots differently than on CI (antialiasing)...

But now the tests are still flaky. As the cursor within the input "flashes" and that is causing the snapshots to differ (compare runner 1 and runner 2)

image

Issue #4193, opened 2023-06-07
[BITV] Buttons accessibility improvements


jancborchardt:
Update: I hid the previous conversation as we were going in the wrong direction. I'm updating the description with the specific accessibility problems that we have in the button component as per the call with the expert last week.  

To be addressed:

- [x] Focus state: currently there's not enough contrast between the background of the buttons and the dark outline we have. 
Solution: Add 2 outlines as focus state to buttons, both 2px thick. The inner should be `--color-main-text`, and the outer `--color-main-background`

<img width="811" alt="Screenshot 2023-06-30 at 11 12 46" src="https://github.com/nextcloud/nextcloud-vue/assets/26852655/fdcb465c-a3a9-4c50-9f11-d62dda7217ee">

- [x] Primary button hover state doesn't meet text contrast requirements (needs solution)
- [x] Special buttons don't meet text contrast requirements (needs solution)

marcoambrosini:
Why exactly do we need that @jancborchardt  @JuliaKirschenheuter ?  What would be the color of the text and of the outline?

nimishavijay:
@marcoambrosini the current style of the secondary button doesn't have enough contrast with the background, so it wouldn't meet a11y requirements.

For colors I think we could keep it similar to the screenshot, `--color-primary-element` text (in bold font) and `--color-border-maxcontrast` outline? it would look something like this:

![image](https://github.com/nextcloud/nextcloud-vue/assets/52440189/f608e499-2b67-4132-b2af-2e63c0d094c7)

The bold text and rounded corners are what differentiate it from an input field. What do you think?

marcoambrosini:
@nimishavijay or @JuliaKirschenheuter could you point me to the accessibility requirements that the secondary button is violating?

JuliaKirschenheuter:
Hi @marcoambrosini, 

> I've done some digging and there seems to be no requirement for button background contrast with the page background.
Here is one example: 

![Screenshot from 2023-06-16 15-27-41](https://github.com/nextcloud/nextcloud-vue/assets/6078378/0be57dfe-8d93-470c-8a9c-19e57892bce4)
https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html

well, it is important that component background have enough contrast with adjacent color outside the component. As well as button text with component background.
But i can't say i've completely understood all this examples and because of it we would like to have a meeting with Michael.

[...]

PR #4347, opened 2023-07-18
Set outline on focused radiobutton

JuliaKirschenheuter:
### ☑️ Resolves

* Fix https://github.com/nextcloud/server/issues/39413

Outline idea was taken from https://github.com/nextcloud/nextcloud-vue/issues/4193

### 🖼️ Screenshots

🏚️ Before
Peek 2023-07-17 10-35

🏡 After
Peek 2023-07-19 09-25

### 🏁 Checklist

### DIFF
--- a/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
+++ b/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
@@ -499,7 +499,9 @@ $spacing: 4px;
 
 	&:not(&--disabled) &__label:hover,
 	&:not(&--disabled) &__label:focus-within {
-		background-color: var(--color-primary-element-light);
+		border-color: var(--color-primary-element);
+		outline: 2px solid var(--color-main-text);
+		box-shadow: 0 0 0 4px var(--color-main-background);
 	}
 
 	// Switch specific rules
--- a/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
+++ b/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
@@ -499,7 +499,6 @@ $spacing: 4px;
 
 	&:not(&--disabled) &__label:hover,
 	&:not(&--disabled) &__label:focus-within {
-		border-color: var(--color-primary-element);
 		outline: 2px solid var(--color-main-text);
 		box-shadow: 0 0 0 4px var(--color-main-background);
 	}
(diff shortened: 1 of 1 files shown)


JuliaKirschenheuter:
should be done for master too

Pytal (approved):

susnux (approved):
:sparkles:

marcoambrosini (changes requested):
I don't think we should have this. Te radio button should display the double outline only when focused via keyboard like in `NcButton`. This pr shows the outline on them when hovered and focused.
cc @nextcloud/designers 

Current state as per this pr: 

https://github.com/nextcloud/nextcloud-vue/assets/26852655/a245f28b-8073-4ea6-beb9-41d30c58f293

JuliaKirschenheuter:
Thank you @marcoambrosini 
Then i will leave background color (as we had it already) and will set a double outline only on focus, right?

nimishavijay:
With this change, the outline is still seen when you click on an item (screenshot attached), which is not ideal. It should be shown only for keyboard navigation, similar to the conversation list in Talk. Would that be possible? :) 


https://github.com/nextcloud/nextcloud-vue/assets/52440189/44e04534-4081-4aaf-b7dc-f1d301029c39

Pytal:
> With this change, the outline is still seen when you click on an item (screenshot attached), which is not ideal. It should be shown only for keyboard navigation, similar to the conversation list in Talk. Would that be possible? :)

`:focus-visible` ?

marcoambrosini:
Yes `:focus-visible` @JuliaKirschenheuter

marcoambrosini:
You're right @JuliaKirschenheuter, there's no focus state on radio buttons.

So I would drop this outline altogether here as it's not an accessibility requirement.
In this case we have the filled circle already indicating the selected state of the radio button.

JuliaKirschenheuter:
> In this case we have the filled circle already indicating the selected state of the radio button.

@marcoambrosini we have to mark selected radio button, see an example under: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio

Peek 2023-07-24 15-49

> there's no focus state on radio buttons.

i will try to set `:focus-visible` on a label after my vacation

JuliaKirschenheuter:
I will finish this PR after my vacation

marcoambrosini:
Ok, I see. So it is possible to differentiate between keyboard and mouse navigation. Can we have a coworking call to get to the bottom of this @JuliaKirschenheuter when you're back? 
Happy holidays 🌴 😄

JuliaKirschenheuter:
After updating Nc Vue to ^8.0.0 and component refactoring focus / hover / selected state looking good ;)
Thanks @susnux!

How it look s now:

Peek 2023-08-11 11-05

Closing PR because not relevant anymore

PR #4349, opened 2023-07-18
chore: Drop `install` entry point and replace it with an Vue Plugin

susnux:
### ☑️ Resolves

The entry point is [unused](https://github.com/search?q=%22%40nextcloud%2Fvue%2Fdist%2Finstall%22&type=code) and just adds another 1.2MiB of assets for the package.
Instead this will add a Vue plugin as a named export.

:test_tube: Tests included

### 🏁 Checklist


tests/unit/plugin.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+import { NextcloudVuePlugin } from '../../src/plugin'
+import { createLocalVue, mount } from '@vue/test-utils'
+
+describe('Nextcloud Vue Plugin', () => {
+	'use strict'
+
```
ShGKme, line 28: Does `'use strict'` take effect in ESM?

tests/unit/plugin.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+	it('can be installed', () => {
+		const localVue = createLocalVue()
+		localVue.use(NextcloudVuePlugin)
+		const wrapper = mount({ render: (h) => h('NcButton') }, { localVue })
+		expect(wrapper.vm.$el.tagName).toBe('BUTTON') // If it could no be installed it would be 'NCBUTTON'
```
ShGKme, line 33: This test also tests that `NcButton` component renders `<button>` which in theory is not guaranteed and not the result of installing the plugin. 

> // If it could not be installed, it would be 'NCBUTTON'

If that is true, can we test `.not.toBe('NCBUTTON')` as it's said?

Or can we test it with `findComponent` or `wrapper.vm`?

ShGKme (commented):
Agree with install entrypoint removing

tests/unit/plugin.spec.ts
susnux: > his test also tests that NcButton component renders <button> which in theory is not guaranteed 

True, assuming button is the default, testing the negative version (not ncbutton) should be safer as it does not assume anything.

> If that is true, can we test .not.toBe('NCBUTTON') as it's said?

Yes :)

tests/unit/plugin.spec.ts
susnux: No ESM is always in strict mode. But I am not sure if we are running in ESM mode here as the Typescript is transpiled to JS and we currently have an implicit package type of CommonJS.

ShGKme: It should add `use strict` on transtiling. If it doesn't, we should enable it :D

ShGKme (approved):

juliushaertl (approved):

PR #4356, opened 2023-07-20
fix(NcListItem) - define a single place for NcActions to render

### LINKED ISSUE #3889: `NcListItem`: Can not access actions using keyboard navigation
If `force-display-actions` is not set the actions can not be accessed using the keyboard, there is no way to tab to the action menu.

You can test this on the Netlify page for `NcListItem`, e.g. this is a screen recording using chrome:
*(Note that you can focus the list item and the avatar, but the actions are only accessible if `force-display-actions` is set to `true`)*

https://user-images.githubusercontent.com/1855448/225318809-09b313aa-b85f-4778-bf32-dd02efa075da.mp4

Antreesy:
### ☑️ Resolves

* Fix #3889 
* I've noticed, that we may render two NcActions component in one NcListItem, but hide one of them with `display: none`
* Because of that, and because of behaviour of `handleTab()` method in NcListItem, we couldn't focus one of the two available action menus
* Not sure, that behaviour now is align with a11y requirements, but at least actions are accessible

### 🖼️ Screenshots

🏚️ Before

https://github.com/nextcloud/nextcloud-vue/assets/93392545/8ea06a24-f030-438a-b0c7-1bdb02e23a53

🏡 After

https://github.com/nextcloud/nextcloud-vue/assets/93392545/e54af34f-64d2-4b15-ae18-4c8fb5072b08

### 🚧 Tasks

- [ ] Code review
- [ ] Manual testing

### 🏁 Checklist

### DIFF
--- a/src/components/NcListItem/NcListItem.vue
+++ b/src/components/NcListItem/NcListItem.vue
@@ -267,7 +267,8 @@
 						</div>
 
 						<!-- Actions -->
-						<div v-show="displayActionsOnHoverFocus && !forceDisplayActions"
+						<div v-if="!forceDisplayActions"
+							v-show="displayActionsOnHoverFocus"
 							class="list-item-content__actions"
 							@click.prevent.stop="">
 							<NcActions ref="actions"
@@ -279,7 +280,7 @@
 						</div>
 					</div>
 					<!-- Actions -->
-					<div v-show="forceDisplayActions"
+					<div v-if="forceDisplayActions"
 						class="list-item-content__actions"
 						@click.prevent.stop="">
 						<NcActions ref="actions"


szaimen:
@Antreesy did you try starting from the bottom and using `shift + tab`? This does not seem to work as expected...

Antreesy:
> This does not seem to work as expected...

That's because of mentioned method, it considers only forward order:
https://github.com/nextcloud/nextcloud-vue/blob/ff4baa4b5e9cc2291d368b019638516d64a9d4df/src/components/NcListItem/NcListItem.vue#L563-L572

ShGKme (approved):
Looks good, works fine.
And I'd expect around 29% performance increase in Talk with this change.

Let's backport it

Antreesy:
@szaimen tabbing behaviour seems to be a different issue from the current PR. How about we merge it and solve `Shift+Tab` in a follow-up?

szaimen:
> @szaimen tabbing behaviour seems to be a different issue from the current PR. How about we merge it and solve `Shift+Tab` in a follow-up?

sounds okay to me

szaimen (approved):
Fine by me then

PR #4360, opened 2023-07-22
Automated styling update from server

nextcloud-command:
Auto-generated update of our locally cached css styling from
 * https://nextcloud.github.io/server/apps/theming/css/default.css
 * https://nextcloud.github.io/server/core/css/server.css
 * https://nextcloud.github.io/server/dist/icons.css


szaimen (approved):

susnux:
I think we should change the workflow as updating the styles breaks the snapshots, so we should always update the cypress snapshots.

szaimen:
> so we should always update the cypress snapshots.

Sounds good 👍

raimund-schluessler:
What happened to the text here? It looks very weird to me now. Is this desired?

See:
grafik

raimund-schluessler:
> I think we should change the workflow as updating the styles breaks the snapshots, so we should always update the cypress snapshots.

I don't think that's a good idea. Changes in server are not always correct and might as well introduce problems here. Automatically accepting visual changes coming from server defeats the purpose of the visual regression tests, I think.

susnux:
> What happened to the text here? It looks very weird to me now. Is this desired?

The font changed on the screenshot, old is Roboto new is something else (see the `c`).
I guess `system-ui` is defined to some different font on CI? Because locally it takes Roboto.

skjnldsv:
> > What happened to the text here? It looks very weird to me now. Is this desired?
> 
> The font changed on the screenshot, old is Roboto new is something else (see the `c`). I guess `system-ui` is defined to some different font on CI? Because locally it takes Roboto.

Yep, the trick to load Roboto doesn't work anymore, it needs to be installed system wide now

skjnldsv:
Tried something, didn't work :(

ShGKme (approved):

susnux:
If anyone of you has an idea to fix the cypress CI go for it.

Somehow we need to ensure one consistent font is used, e.g. ensure system-ui resolves to Roboto

raimund-schluessler:
> If anyone of you has an idea to fix the cypress CI go for it.
> 
> Somehow we need to ensure one consistent font is used, e.g. ensure system-ui resolves to Roboto

The cypress action used to upload the actual snapshots which one could then download and use as new base snapshots. But it doesn't do that anymore. Once can only find the case and diffs, but not the actual ones.

raimund-schluessler:
The screenshots are created in `cypress/snapshots` and cypress does not respect the settings in 
https://github.com/nextcloud-libraries/nextcloud-vue/blob/c2095bfbaa9877a5920e7375c94f75390090e0ba/cypress.config.ts#L40-L43

Maybe we should simply adjust the cypress action to upload the `screenshots` folder instead :thinking:

susnux:
> The screenshots are created in `cypress/snapshots` and cypress does not respect the settings in

That setting does not belong into `env` but inside the root

susnux (approved):

PR #4365, opened 2023-07-26
feat(`NcRichText`) add support of GFM (extended markdown)

Antreesy:
### ☑️ Resolves

* Add package [remark-gfm](https://github.com/remarkjs/remark-gfm) for further support of:
  * autolink literals
  * footnotes ??
  * strikethrough `~~`
  * tables ` -- | -- `
  * tasklists ` - [ ]` `- [x]`
* parsed text was hacked to insert non-breaking space between each two new lines. Thus span which rendered for a new line, have both width and height parameters, and is visible in the window

### 🖼️ Screenshots

image

[markdown-checkboxes.webm](https://github.com/nextcloud-libraries/nextcloud-vue/assets/93392545/a842bfb8-1e56-4777-9971-cb0c847694ab)

In Talk messages:
image

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/NcRichText/NcRichText.vue
+++ b/src/components/NcRichText/NcRichText.vue
@@ -79,6 +79,46 @@ textarea {
 </style>
 ```
 
+### Flavored Markdown
+
+This component can support [Github Flavored Markdown](https://github.github.com/gfm/).
+It adds such elements, as tables, task lists, strikethrough, and supports autolinks by default
+
+```vue
+<template>
+	<div>
+		<textarea v-model="text" />
+
+		<NcRichText :text="text" :use-extended-markdown="true"/>
+	</div>
+</template>
+<script>
+	export default {
+		data() {
+			return {
+				text: `## Try flavored markdown right now!
+
+~~strikethrough~~
+
+- [ ] task to be done
+- [x] task completed
+
+Table header | Column A | Column B
+-- | -- | --
+Table row | value A | value B
+`,
+			}
+		},
+	}
+</script>
+<style lang="scss">
+textarea {
+	width: 100%;
+	height: 200px;
+}
+</style>
+```
+
 ### Usage with NcRichContenteditable
 
 See [NcRichContenteditable](#/Components/NcRichContenteditable) documentation for more information
@@ -235,11 +275,13 @@ See [NcRichContenteditable](#/Components/NcRichContenteditable) documentation fo
 
 <script>
 import NcReferenceList from './NcReferenceList.vue'
+import NcCheckboxRadioSwitch from '../NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue'
 import { remarkAutolink } from './autolink.js'
 import { remarkPlaceholder, prepareTextNode } from './placeholder.js'
 
 import { unified } from 'unified'
-import markdown from 'remark-parse'
+import remarkParse from 'remark-parse'
+import remarkGfm from 'remark-gfm'
 import breaks from 'remark-breaks'
 import remark2rehype from 'remark-rehype'
 import rehype2react from 'rehype-react'
@@ -298,6 +340,11 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		/** Provide GitHub Flavored Markdown syntax */
+		useExtendedMarkdown: {
+			type: Boolean,
+			default: false,
+		},
 		autolink: {
 			type: Boolean,
 			default: true,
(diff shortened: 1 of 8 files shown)


susnux:
One thing I am not sure about is whether we want to different Markdown flavors, as this can be quite confusing.
So on the one side we would have the text and collectives apps running on CommonMark + Extensions and on the other side Talk.

Maybe we should make sure NcRichText and Text supports the same Markdown dialect?

Antreesy:
> Maybe we should make sure NcRichText and Text supports the same Markdown dialect?

That could be the best way to do that. Can you tell, where I could get a list of supported MD for Text / Collectives to compare with?

Antreesy:
Link to documentation:
- Github FM: https://github.github.com/gfm/
- CommonMark Spec: https://spec.commonmark.org/

So, GFM is based on CMS, including such extensions as autolink, strikethrough, tables and tasklists. I see they're activated in Text app as well: https://github.com/nextcloud/text/blob/cd39a0decac0a6f003d4bddff22faa70c3e71ae5/src/markdownit/index.js#L13-L16

From that point, flavor would be the same, although we might need to align syntax for some extensions (and not sure this make sense for the tables, as Text has injected table editor, which would be an overkill for Talk input)

To ensure we're supporting the same syntax, we could try to base NcRichText tests on  https://spec.commonmark.org/0.30/spec.json somehow, but I don't think it's needed, as libraries based on the same specs

Antreesy:
Because of using some React functionality at `master` branch to create nodes, I had troubles with passing an event from rendered checkbox to parent. Will make another attempt 

Apart from it, a basic extended markdown is working and evem possible to backport for master release, although I don't know, if it's considered as breaking change or not (original behavior was untouched, this goes under new functionality)

raimund-schluessler:
@Antreesy just out of curiosity, why not send this against `master`? Is there even an app migrated to vue 3, with which you could test the PR currently?

Antreesy:
> why not send this against `master`?

As i wrote above, it currently works with Vue 3 and doesn't work with Vue 2.7. As they require to handle code differently anyway, I'll try to implement it onto master and attach backport PR.

But for now it would be good at least to get a review on current changes

src/components/NcRichText/autolink.js
```diff
@@ -23,10 +23,10 @@ const NcLink = {
 	},
 }
 
-export const remarkAutolink = function({ autolink, useMarkdown }) {
+export const remarkAutolink = function({ autolink, useMarkdown, useExtendedMarkdown }) {
 	return function(tree) {
-
-		if (!useMarkdown || !autolink) {
+		// remark-gfm has its own autolink parser
+		if (useExtendedMarkdown || !useMarkdown || !autolink) {
```
ShGKme, line 29: Does it mean that with extended MD auto-link is always enabled even when autoLink === false?

src/components/NcRichText/NcRichText.vue
```diff
@@ -366,6 +438,27 @@ export default {
...
+									const id = 'markdown-input-' + GenRandomId(5)
+									const inputComponent = h(NcCheckboxRadioSwitch, {
+										attrs: {
+											...inputNode.data.attrs,
+											id,
+											disabled: !this.interactive,
+										},
+										on: {
+											'update:checked': (value) => {
+												this.$emit('interacted', { id, label, value })
```
ShGKme, line 455: Events are usually named as action name, e.g. `click`, not `clicked`.

```suggestion
												this.$emit('interact', { id, label, value })
```

ShGKme: But according to the payload, this event (or at least with this payload) is supposed to be only `check` state in todo change.

So I'd call it explicitly `check` or `todo-check` or something like that.

Or, if we expect more interactions in the future, kind of 

```js
this.$emit('interact', { for: 'todo', id, label, value })
this.$emit('interact:todo', { id, label, value })
```

ShGKme (commented):
Cool, seems to work fine. Some comments

src/components/NcRichText/autolink.js
Antreesy: Yes, there is no way to disable it

ShGKme (approved):

src/components/NcRichText/NcRichText.vue
ShGKme: > Sounds like a good idea, then? `this.$emit('interact', event)` for other elements with native events, if they ever appear `this.$emit('interact:todo', { id, label, value })` for checkboxes

I meant `interact` for any interaction in case we need to just handle all interactions, and `interaction:<name>` for a specific kind of interaction

susnux (approved):

Antreesy:
Aj, we forgot to put it into the 8.5.0 changelog 🙈

Pytal:
🙈 sorry, looks like this one was excluded by our config https://github.com/nextcloud-libraries/nextcloud-vue/blob/99a7c574e6f119b34be467afe354c7506d7f5669/.github/release.yml#L11 because of the dependencies label

PR #4366, opened 2023-07-26
NcButton: Add alignment property to change icon and text ordering

susnux:
### ☑️ Resolves

We sometimes need to change the alignment of icon and text of the button, currently this is dirty-hacked by using `:deep` selectors. Examples are the column headers on tables (file list, logreader, file picker).

### 🖼️ Screenshots

image

### 🏁 Checklist

### DIFF
--- a/src/components/NcButton/NcButton.vue
+++ b/src/components/NcButton/NcButton.vue
@@ -197,6 +197,41 @@ button {
 </style>
 ```
 
+### Alignment
+Sometimes it is required to change the icon alignment on the button, like for switching between pages as in following example:
+
+```vue
+<template>
+	<div>
+		<div style="display: flex; gap: 12px;">
+			<NcButton aria-label="Previous" type="tertiary">
+				<template #icon>
+					<IconLeft :size="20" />
+				</template>
+				Previous page
+			</NcButton>
+			<NcButton alignment="end-reverse" aria-label="Previous" type="primary">
+				<template #icon>
+					<IconRight :size="20" />
+				</template>
+				Next page
+			</NcButton>
+		</div>
+	</div>
+</template>
+<script>
+import IconLeft from 'vue-material-design-icons/ArrowLeft.vue'
+import IconRight from 'vue-material-design-icons/ArrowRight.vue'
+
+export default {
+	components: {
+		IconLeft,
+		IconRight,
+	},
+}
+</script>
+```
+
 ### Pressed state
 It is possible to make the button stateful by adding a pressed state, e.g. if you like to create a favorite button.
 The button will have the required `aria` attribute for accessibility and visual style (`primary` when pressed, and the configured type otherwise).
@@ -264,6 +299,17 @@ export default {
 	name: 'NcButton',
 
 	props: {
+		/**
+		 * Set the text and icon alignment
+		 *
+		 * @default 'center'
+		 */
+		alignment: {
+			type: String,
+			default: 'center',
+			validator: (alignment) => ['start', 'start-reverse', 'center', 'end', 'end-reverse'].includes(alignment),
+		},
+
 		/**
 		 * Toggles the disabled state of the button on and off.
 		 */
@@ -388,6 +434,20 @@ export default {
 			}
 			return this.type
 		},
+
+		/**
+		 * The flexbox alignment of the button content
+		 */
+		flexAlignment() {
+			return this.alignment.split('-')[0]
+		},
+
+		/**
+		 * If the button content should be reversed (icon on the end)
+		 */
+		isReverseAligned() {
+			return this.alignment.includes('-')
+		},
 	},
 
 	/**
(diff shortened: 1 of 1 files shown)


src/components/NcButton/NcButton.vue
```diff
@@ -388,6 +434,20 @@ export default {
...
+		 */
+		flexAlignment() {
+			return this.alignment.split('-')[0]
+		},
+
+		/**
+		 * If the button content should be reversed (icon on the end)
+		 */
+		isReverseAligned() {
+			return this.alignment.indexOf('-') !== -1
```
ShGKme, line 449:
```suggestion
			return this.alignment.includes('-')
```

src/components/NcButton/NcButton.vue
```diff
@@ -591,7 +667,8 @@ export default {
 
 	// Icon and text button
 	&--icon-and-text {
-		padding: 0 16px 0 4px;
+		padding-block: 0;
+		padding-inline: 4px 16px;
```
ShGKme, line 671: 💙 for rtl friendly properties

src/components/NcButton/NcButton.vue
```diff
@@ -197,6 +197,41 @@ button {
...
+				Previous page
+			</NcButton>
+			<NcButton alignment="end-reverse" aria-label="Previous" type="primary">
+				<template #icon>
+					<IconRight :size="20" />
+				</template>
+				Next page
+			</NcButton>
+		</div>
+	</div>
```
ShGKme, line 220: (More examples)

```suggestion
	<div style="display: flex; flex-direction: column; gap: 12px;">
		<NcButton aria-label="central (default)" type="secondary" wide>
			<template #icon>
				<IconLeft :size="20" />
			</template>
			central (default)
		</NcButton>
		<div style="display: flex; gap: 12px;">
			<div style="display: flex; flex-direction: column; gap: 12px; flex: 1">
				<NcButton alignment="start" aria-label="start" type="secondary" wide>
					<template #icon>
						<IconLeft :size="20" />
					</template>
					start
				</NcButton>
				<NcButton alignment="start-reverse" aria-label="start-reverse" type="secondary" wide>
					<template #icon>
						<IconRight :size="20" />
					</template>
					start-reverse
				</NcButton>
			</div>
			<div style="display: flex; flex-direction: column; gap: 12px; flex: 1">
				<NcButton alignment="end" aria-label="end" type="secondary" wide>
					<template #icon>
						<IconLeft :size="20" />
					</template>
					end
				</NcButton>
				<NcButton alignment="end-reverse" aria-label="end-reverse" type="secondary" wide>
					<template #icon>
						<IconRight :size="20" />
					</template>
					end-reverse
				</NcButton>
			</div>
		</div>
	</div>
```

image

jancborchardt: I think the "start-reverse" and "end" layouts should not exist though, they look off and should not be used or recommended.

src/components/NcButton/NcButton.vue
jancborchardt: This would be even better via variables :)
```suggestion
		padding-block: 0;
		padding-inline: var(--default-grid-baseline) calc(var(--default-grid-baseline) * 4);
```

ShGKme: Wow, we have spacing variables now!

src/components/NcButton/NcButton.vue
susnux: @jancborchardt start-reverse and end are the values for which this was feature is introduced (table headers sort icon should be on the opposite site of the alignment)

susnux: But as discussed with Jan, we should not promote those stypes for anything but table headers.

src/components/NcButton/NcButton.vue
susnux: Thank you, changed this to variables :)

marcoambrosini (commented):
are there other applications for trailing icons in buttons? I'm not really sure we should implement this just for pagination. Those should be link anyways and we could have a very simple wrapper of the button that just does pagination. 
What is the problem in tables?

susnux:
@marcoambrosini the sorting icon should normally be on the opposite side of the text alignment, or at least on the end instead of the inline start.
So we discuessed different approaches like adding a "trailing-icon" slot (not good, as we never want a button with two icons). Or introducing a wrapper component which hacks the alignment using `deep` selectors (this is seems not to be a good solution as it depends on private classes of the button, instead we would have to introduce something like `customClasses` property to allow to inject our own classes which then can be safely be using with `deep` selectors).

But allowing to align the button content using a property seemed to be the best and cleanest solution, instead of introducing a wrapper component.

There are even more examples on where this could be used, as at least the "start" alignment is quite common (have left aligned content but full width of clickable area).

src/components/NcButton/NcButton.vue
ShGKme: > @jancborchardt start-reverse and end are the values for which this was feature is introduced (table headers sort icon should be on the opposite site of the alignment)

Do you have `end`, not `end-reversed` button in file picker table?

marcoambrosini:
Prop is clean, but it's so clean that people might start think that it's fine to have those type of buttons elsewhere. That's my only worry, but let's do it!

skjnldsv (approved):

marcoambrosini (approved):

Issue #4200, opened 2023-06-08
Fix vertical alignment of text in nccounterbubble


szaimen:
As discussed with @jancborchardt and @nimishavijay, the text inside the counterbubble should be positioned right in the center and not a bit off.

See the following screenshot:
![image](https://raw.githubusercontent.com/nextcloud/documentation/02a3c92edab2984ef47758d0df8076fe3d4492c5/developer_manual/images/talk-counter-bubble.png)

szaimen:
As @nimishavijay and me found out this happens due to the font-family used. In our case it got fixed by using `sans-serif`. Maybe we should enforce this here?

jancborchardt:
The font should be the same as used elsewhere, otherwise that will look off instead.

How is the centering currently achieved? We can change the approach, either via vertical align, or padding, or however possible.

szaimen:
> The font should be the same as used elsewhere, otherwise that will look off instead.

It uses the first font that is installed on the system in this order: `-apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Cantarell, Ubuntu, "Helvetica Neue", Arial, sans-serif, "Noto Color Emoji", "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol`

> How is the centering currently achieved? We can change the approach, either via vertical align, or padding, or however possible.

The problem is indeed the font and not the centering itself. On macOS for example everything looks fine since a different system font, probably -apple-system or BlinkMacSystemFont is used.

See the following screenshot that shows this pretty clearly on my system:
![image](https://github.com/nextcloud/nextcloud-vue/assets/42591237/0ac6da14-376e-48f8-8c89-f9f972f98a37)

szaimen:
So it seems to come down to system font rendering here...

szaimen:
Should we enforce a specific font then here?

szaimen:
also cc @marcoambrosini

nimishavijay:
There were issues with using custom fonts in the past, if I remember correctly? My worry is that changing something like the font for this one element will both look out of place visually and not render correctly if that font is not supported/installed on the system.

marcoambrosini:
Who did the original screenshot? Could we gather more information about where it looks bad and try to find out why?

szaimen:
@marcoambrosini this is on my windoes 11 machine. It happens due to the system font used. Please read the conversation above.

[...]

PR #4367, opened 2023-07-26
NcButton: Allow to format text of button

susnux:
### ☑️ Resolves

Allow us to format the text of a button without dirty hacks like `:deep` selectors, I added the example why this is useful and where we should use this: Table headers.

Places where we have this and can remove all custom hacks are: all file tables in the files app, the logreader app, the file picker.

:warning: this is chained on #4366  (see first commit), should be reviewed and merged first

### 🖼️ Screenshots

image

### 🏁 Checklist

### DIFF
--- a/src/components/NcButton/NcButton.vue
+++ b/src/components/NcButton/NcButton.vue
@@ -319,6 +319,118 @@ export default {
 }
 </script>
 ```
+
+### Usage example: Sorting table columns
+The standard way to implement sortable table column headers should be like this:
+
+```vue
+<template>
+	<table>
+		<thead>
+			<tr>
+				<th :aria-sorted="sortedName" class="row-name">
+					<NcButton alignment="start-reverse"
+						:wide="true"
+						type="tertiary"
+						@click="sortName">
+						<template #icon>
+							<IconDown v-if="sortedName === 'ascending'" class="sort-icon" :size="20" />
+							<IconUp v-else-if="sortedName === 'descending'" class="sort-icon" :size="20" />
+						</template>
+						<span class="table-header">Name</span>
+					</NcButton>
+				</th>
+				<th :aria-sorted="sortedSize" class="row-size">
+					<NcButton alignment="end"
+						:wide="true"
+						type="tertiary"
+						@click="sortSize">
+						<template #icon>
+							<IconDown v-if="sortedSize === 'ascending'" class="sort-icon" :size="20" />
+							<IconUp v-else-if="sortedSize === 'descending'" class="sort-icon" :size="20" />
+						</template>
+						<span class="table-header">Size</span>
+					</NcButton>
+				</th>
+			</tr>
+		</thead>
+		<tbody>
+			<tr>
+				<td class="row-name">Lorem ipsum</td>
+				<td class="row-size">8 MiB</td>
+			</tr>
+		</tbody>
+	</table>
+</template>
+<script>
+import IconUp from 'vue-material-design-icons/MenuUp.vue'
+import IconDown from 'vue-material-design-icons/MenuDown.vue'
+
+export default {
+	components: {
+		IconUp,
+		IconDown,
+	},
+	data() {
+		return {
+			sortedName: null,
+			sortedSize: null,
+		}
+	},
+	methods: {
+		sortName() {
+			if (this.sortedName === 'ascending') {
+				this.sortedName = 'descending'
+			} else if (this.sortedName === 'descending') {
+				this.sortedName = null
+			} else {
+				this.sortedName = 'ascending'
+			}
+		},
+		sortSize() {
+			if (this.sortedSize === 'ascending') {
+				this.sortedSize = 'descending'
+			} else if (this.sortedSize === 'descending') {
+				this.sortedSize = null
+			} else {
+				this.sortedSize = 'ascending'
+			}
+		},
+	},
+}
+</script>
+<style>
+table {
+	table-layout: fixed;
+	width: 300px;
+}
+
+td.row-name {
+	padding-inline-start: 16px;
+}
+
+td.row-size {
+	text-align: right;
+	padding-inline-end: 16px;
+}
+
+.table-header {
+	font-weight: normal;
+	color: var(--color-text-maxcontrast);
+}
+
+.sort-icon {
+	color: var(--color-text-maxcontrast);
+	position: relative;
+	inset-inline: -10px;
+}
+
+.row-size .sort-icon {
+	inset-inline: 10px;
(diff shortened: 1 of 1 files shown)


jancborchardt (commented):
Compared to the normal buttons it seems like the text is a bit higher up here than there. Might be a non-issue since it’s only visible on hover / with background.

marcoambrosini (commented):
Same as for trailing icons, unless we have other use cases I would wrap the button component into a `tableHeaderButton` or something like that and make the style changes from there

susnux:
> Compared to the normal buttons it seems like the text is a bit higher up here than there. Might be a non-issue since it’s only visible on hover / with background.

Tested this with @szaimen on Window, it depends on your local font, we have a hack (1px margin) to center it with Segoe font which is aligned differently, but it breaks on other fonts (like I am using Roboto). See also #4200

skjnldsv (approved):

ShGKme:
> Same as for trailing icons, unless we have other use cases I would wrap the button component into a `tableHeaderButton` or something like that and make the style changes from there

We have many buttons in Talk with custom text styles: LeftSidbar, Settings button, Conversation, TopBar, CallTime and etc.

ShGKme (commented):
I agree with this change to allow custom content of the button (otherwise it should be a prop, not a slot).

But I'm not sure if it is the best way to customize text styles.

With this change, the usage of custom styles requires creating a new element just to set class:
```vue
<template>
  <NcButton class="my-button">
    <!-- There is no meaning in this span, only for setting a text's color -->
    <span class="my-button__text">Text</span>
  <NcButton>
</template>

<styles scoped>
.my-button__text {
  color: red;
}
</styles>
```

I'd prefer to have a `prop` to set inside styles or classes:

```vue
<template>
  <NcButton class="my-button" text-classes="my-button__text">
    Text
  <NcButton>
</template>

<styles scoped>
/* Also deep, but for my own classes, not NcButton's */
.my-button :deep(.my-button__text) {
  color: red;
}
</styles>
```

Or

```vue
<template>
  <NcButton class="my-button" text-styles="color: red">
    Text
  <NcButton>
</template>
```

susnux:
I like the idea of having `text-classes` property, but than we should also deprecate the default slot. As than it is a misuse of Vue slots.
We probably should move to a `text` property instead.

raimund-schluessler:
Once this is merged, one way or the other, we could then also adjust these lines
https://github.com/nextcloud-libraries/nextcloud-vue/blob/74a7e2816b9ccc80abcde647d6b7485d803379aa/src/components/NcBreadcrumb/NcBreadcrumb.vue#L339-L345

susnux:
The more I think about this, the more I think we should went with this one.

Why? Because we use the pattern of the default slot = content everywhere. Only moving the Button to `text` property seems non consistent, while I personally prefer that idea.
Moreover the HTML `<button>` also allows every other content inside, so it feels more native.

skjnldsv:
I still don't get it. You can just set a class to the button and change the text styling? `deep` only works because there is a scoped css being used, but you are free to use standard css? I don't think we actually needs this, using  :deep is not a `hack`, it's one way to ensure your css isn't scoped within a scoped `<style>`. Feel free to just use standard css?

I think we should keep things simple :)



> Once this is merged, one way or the other, we could then also adjust these lines
> 
> https://github.com/nextcloud-libraries/nextcloud-vue/blob/74a7e2816b9ccc80abcde647d6b7485d803379aa/src/components/NcBreadcrumb/NcBreadcrumb.vue#L339-L345

What does this do?
Put the icon at the end?
You can just use `direction: rtl` to adjust the text order and stay accessible

susnux:
Basically the problem is: Either we should allow to put any content inside the button, like the native html button allows, including styling. Or we should add a way to apply custom styles or classes to the **text** wrapper.
Currently you can only apply a class to the whole button and then have to **assume** internals of the button component, see below.

> using :deep is not a hack, it's one way to ensure your css isn't scoped within a scoped <style>

Yes but then you need to assume some internals of the button, and this is what you should not do.
Components should be a black box so that you do not depend on its implementation (because the class is assigned to the root element and not to the text wrapper).

>  You can just use direction: rtl to adjust the text order and stay accessible

No this is quite bad as you misuse it for stlying while you should use if for localization.

---

So a overall problem is missing consistancy with the components architecture. Sometimes you use the default slot for content and allow child nodes, sometimes a property. But for the button we use the slot but only allow text.
Meaning to be consistent we should either only allow text and then use a property together with a way to format that text, like the suggestes `text-classes` property, or allow any content within the button and simply use the default slot like it is done here.

skjnldsv:
To make sure I get it right, the use case (not talking about the wider discussion) is just about reveting text and the icon, right?

susnux:
> To make sure I get it right, the use case (not talking about the wider discussion) is just about reveting text and the icon, right?

No that is already implement in 8.x by using the `alignment` variable.

It is about formatting the text, like color or weight. See the example added, it then can be used for the files list header (e.g. in files app or file picker) or for the log reader header. Simply speaking for every table header without to have using any hacks that depend on private internals of the component.

susnux:
> The more I think about this, the more I think we should went with this one.
> 
> Why? Because we use the pattern of the default slot = content everywhere. Only moving the Button to `text` property seems non consistent, while I personally prefer that idea. Moreover the HTML `<button>` also allows every other content inside, so it feels more native.

susnux:
> The more I think about this, the more I think we should went with this one.

After seeing all that validator issues @ShGKme fixes at the moment, maybe go with the alternative PR. So that we do not allow arbitrary (and possibly wrong) content of the button but only text (and the icon).

What do you think?

ShGKme:
> `deep` only works because there is a scoped css being used, but you are free to use standard css? I don't think we actually needs this, using :deep is not a `hack`, it's one way to ensure your css isn't scoped within a scoped `<style>`. Feel free to just use standard css?
> 
> I think we should keep things simple :)

`deep` is a `hack` because it allows changing component's internal implementation (until we explicitly say that some class name is a part of the component's public interface and a way to customize it).

With `deep` any simple layout/class name refactoring brakes a layout.

marcoambrosini (approved):

susnux:
As this seems to get more positive feedback than the alternative https://github.com/nextcloud-libraries/nextcloud-vue/pull/4393

Shall we go with this one? :+1:  | :-1:  ?

jancborchardt:
@raimund-schluessler since you commented on the other one, what is your take here – approved or any feedback? :)

susnux:
Well then lets go with this one as it is less breaking?

PR #4368, opened 2023-07-27
fix(NcModal): prev/next color and sizing fix

skjnldsv:
| Before | After |
|:---------:|:------:|
|||

### DIFF
--- a/src/components/NcModal/NcModal.vue
+++ b/src/components/NcModal/NcModal.vue
@@ -882,11 +882,13 @@ export default {
 		z-index: 10000;
 		// ignore display: none
 		display: flex !important;
-		height: 35vw;
+		height: 35vh;
+		min-height: 300px;
 		position: absolute;
 		transition: opacity 250ms,
 			visibility 250ms;
-		color: var(--color-primary-element-text);
+		// hover the mask
+		color: white;
 
 		&:focus-visible {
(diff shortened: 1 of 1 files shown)


src/components/NcModal/NcModal.vue
```diff
@@ -882,11 +882,13 @@ export default {
...
 		display: flex !important;
-		height: 35vw;
+		height: 35vh;
+		min-height: 300px;
 		position: absolute;
 		transition: opacity 250ms,
 			visibility 250ms;
-		color: var(--color-primary-element-text);
+		// hover the mask
+		color: white;
```
szaimen, line 891: Is this mask always black?

szaimen (approved):
Thanks!

susnux (approved):

susnux:
BTW we need to fix the the actions and title, because it is lacking contrast

src/components/NcModal/NcModal.vue
skjnldsv: yep :+1: 
https://github.com/nextcloud-libraries/nextcloud-vue/blob/0e7db4c6a6518260396ebf47d6c786cd226e317b/src/components/NcModal/NcModal.vue#L743-L746

Even without dark mode, this is still black (with higher transparency, but still black)

PR #4393, opened 2023-08-02
feat(NcButton): Introduce `text` and `text-classes` properties

susnux:
### ☑️ Resolves

Alternative for #4367 

* `text` is the visible text of the button, the default slot for passing the text is now deprecated
* `text-classes` are additional classes injected to the text wrapper for custom styling
* Added example when this should be used (for table headers)

So basically this moves from misused slot to text property with allowing to inject custom classes that can be used later for custom styling.
*(Please note that I had to use the deprecated `v-deep` selector as vue-styleguidist does not support the recommended `:deep` selector).*

### 🖼️ Screenshots

https://github.com/nextcloud-libraries/nextcloud-vue/assets/1855448/61e02bf3-8e4a-43b3-93ca-2c44d7006ccc

### 🏁 Checklist


skjnldsv:
Conflicts

susnux:
@skjnldsv  resolved
---
BTW I would prefer this one over https://github.com/nextcloud-libraries/nextcloud-vue/pull/4367

marcoambrosini (commented):
from a visual perspective, I think these buttons should default as `tertiary-no-background`

src/components/NcButton/NcButton.vue
```diff
@@ -485,15 +635,16 @@ export default {
 	 * @return {object|undefined} The created VNode
 	 */
 	render(h) {
-		const text = this.$slots.default?.[0]?.text?.trim?.()
-
+		/** @slot **deprecated** Use the `text` property instead */
+		const text = this.text || this.$slots.default?.[0]?.text?.trim?.()
```
raimund-schluessler, line 639: Can we directly remove the slot in `master` which will be `8.x` and introduce the deprecation for `stable7`?

susnux: This will be huge in terms of breaking projects, not sure if we want to do this already for version 8 or wait for the next

raimund-schluessler: We have huge breaking changes in master already (e.g. the renaming from `title` to `name`) and the entire removal of some components. So, from my side, I wouldn't mind another breaking change.
The migration effort needs to be done anyway, the question is just, when does it suite you 😉

src/components/NcButton/NcButton.vue
```diff
@@ -346,6 +477,36 @@ export default {
...
+		 * </style>
+		 * ```
+		 */
+		textClasses: {
+			type: [String, Array],
+			validator(value) {
+				return typeof value === 'string' || (Array.isArray(value) && value.every((v) => typeof v === 'string'))
+			},
+			default: '',
+		},
```
ShGKme, line 508: Maybe allow all possible Vue classes types, like Object, array of objects and etc?

ShGKme: Even `[{a : true}, 'b', ['c', { d: true }]]` is valid in bot Vue 2 and Vue 3. So we can add `textClasses` as an array item to `class` property in render-function even without `additionalTextClasses`

src/components/NcButton/NcButton.vue
ShGKme: IMO, we should either allow using slot for button content. It would also solve custom text formatting problem by allowing to pass formatted text as a markup.
```vue
<NcButton>
  <span class="foo">My text</span> <span class="down">🔻</span>
</NcButton>
```

Or remove the content slot at all and make this breaking change with text.

During `beta` we can add check to `mounted` that `console.warn` if slot is passed to help with the migration.

susnux: After seeing all of your validator fixes @ShGKme  I think we should go with this one and only allow text content through props.

src/components/NcButton/NcButton.vue
susnux: Done :)

src/components/NcButton/NcButton.vue
susnux: Pushed a commit that warns when the default slot is used as suggested by @ShGKme 
Also added a test to remind us to remove that behavior when releasing (as soon as the version is set to stable it will fail if the default slot still works :wink: )

ShGKme: > After seeing all of your validator fixes @ShGKme I think we should go with this one and only allow text content through props.

For the record, the only place where `<button>` had invalid non-phrasal content was User Menu avatar-button-trigger, and plain html `<button>` was used there. There was no HTML validation issue with `NcButton` content.

---

In general case, I'd definitely prefer to have a full-functional slot content (that allows actual content, not render-text function). But as no app actually needed any true custom content, we can use text. It would make component more performant and simple.

Alternatively, we may add a new slot, but explicitly mark it as a custom content slot, not the default text content.

src/components/NcButton/NcButton.vue
```diff
@@ -447,6 +588,22 @@ export default {
...
+			if (typeof this.textClasses === 'string') {
+				return `button-vue__text ${this.textClasses}`
+			} else if (Array.isArray(this.textClasses)) {
+				return ['button-vue__text', ...this.textClasses]
+			} else {
+				return {
+					...this.textClasses,
+					'button-vue__text': true,
+				}
+			}
```
ShGKme, line 605:
```suggestion
			return ['button-vue__text', this.textClasses]
```

src/components/NcButton/NcButton.vue
```diff
@@ -478,22 +635,30 @@ export default {
 		},
 	},
 
+	mounted() {
+		// TODO: Remove with stable release
+		if (this.$slots?.default && this.$slots.default[0]?.text && this.$slots.default?.isComment !== true) {
```
ShGKme, line 640: This is guaranteed that `this.$slots` exists on a Vue component instance.
```suggestion
		// TODO: Remove with stable release
		if (this.$slots.default && this.$slots.default[0]?.text && this.$slots.default?.isComment !== true) {
```

src/components/NcButton/NcButton.vue
```diff
@@ -518,7 +683,7 @@ export default {
 					},
 				],
 				attrs: {
-					'aria-label': this.ariaLabel,
+					'aria-label': this.ariaLabel || text,
```
ShGKme, line 686: `this.ariaLabel` prop was removed, wasn't it?

susnux: No it is not currently. We use it to show a warning if it is not set :wink:

src/components/NcButton/NcButton.vue
susnux: Thank you, much cleaner!

src/components/NcButton/NcButton.vue
ShGKme: Oh, it was just moved to another place in props

src/components/NcButton/NcButton.vue
```diff
@@ -163,6 +163,11 @@ export default {
 			readonly: false,
 			style: 'icontext',
 		}
+	},
+	computed: {
+		buttonText() {
+			return this.style.indexOf('text') !== -1 ? 'Example text' : ''
+		}
```
ShGKme, line 170:
```suggestion
		buttonText() {
			return this.style.includes('text') ? 'Example text' : ''
		}
```

ShGKme (approved):
A huge painful breaking change but fine by me.

Let's see for other opinions.

PR #4394, opened 2023-08-02
Feature: Add visible inner label for `NcInputField` on border

susnux:
### ☑️ Resolves

Alternative to #4287 as requested by @marcoambrosini 

### 🖼️ Screenshots

https://github.com/nextcloud-libraries/nextcloud-vue/assets/1855448/20b1cb4e-9f5a-4e27-98c5-b3cc4a9e2124

### 🏁 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ All notable changes to this project will be documented in this file.
   - `name` propery is now required for `NcActions*`, `NcAppNavigationItem` and `NcBreadcrumb*`
   - See linked pull request for full migration guide
 - chore: Drop `install` entry point and replace it with an Vue Plugin [\#4349](https://github.com/nextcloud-libraries/nextcloud-vue/pull/4349) ([susnux](https://github.com/susnux))
+- `NcInputField`: The `labelVisible` property was removed for accessibility it is required to always show a label.
+  You can still use the `labelOutside` property to remove the inner label from the component.
 
 ### :rocket: Enhancements
 - enh\(NcDatetime\): New component for displaying timestamps as time relative from now [\#4219](https://github.com/nextcloud-libraries/nextcloud-vue/pull/4219) ([susnux](https://github.com/susnux))
--- a/src/components/NcActionInput/NcActionInput.vue
+++ b/src/components/NcActionInput/NcActionInput.vue
@@ -23,6 +23,10 @@
 
 <docs>
 This component is made to be used inside of the [NcActions](#NcActions) component slots.
+
+It is recommended to not only provide a placeholder, but also a label.
+The label should describe what input is expected and the placehold what format the input should have.
+
 All undocumented attributes will be bound to the input, the datepicker or the select component, e.g. `maxlength`, `not-before`.
 For the `NcSelect` component, all events will be passed through. Please see the `NcSelect` component's documentation for more details and examples.
 ```vue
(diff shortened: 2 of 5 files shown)


marcoambrosini (approved):
Looks very good and the text has more space around which makes it more readable. All the while using less height. It's true that the icon is not a problem with the other approach either.
I would probably avoid moving the label left in the transition when there's a leading icon. What do you think @susnux?

dartcafe:
I like it, but the initial state looks like the field already has a content, although it seems to be a placeholder/the label. I fear this could mislead users.

szaimen:
> I would probably avoid moving the label left in the transition when there's a leading icon. What do you think @susnux?

I agree. Otherwise looks good to me :)

susnux:
Thank you for your feedback, so far I would change it to:
1. Do not move label horizontal
2. Make label `--color-text-maxcontrast` so it does not look like content

szaimen:
> * Do not move label horizontal
> * Make label `--color-text-maxcontrast` so it does not look like content

Sounds good to me 👍

susnux:
@szaimen @dartcafe @marcoambrosini  I fixed the label color and the horizontal shift, it looks like this now:

https://github.com/nextcloud-libraries/nextcloud-vue/assets/1855448/994e4988-f080-4429-abe2-24a3d38f75b9

szaimen (approved):
Looks great! :)
But didnt review the code

nimishavijay:
Looks good! :)
I would suggest 2 very small changes:
- [ ] for the text fields without an icon there is still a horizontal shift of the label when it is focused, so if we do `margin-inline: 6px 0;` when focused and `margin-inline: 10px 0;` when unfocused this shift is gone
- [ ] for success and error variants, the color of the label when focused should be either `--color-success-text` or `--color-error-text`, currently it's the primary color for everything

With those fixes it would look like this:

https://github.com/nextcloud-libraries/nextcloud-vue/assets/52440189/bba84074-a998-45f6-87a7-85904429a65d


Also, on my system there is some text getting cut off, and it changes according to the zoom level.

1280x720 at 100%
image

1280x720 at 80% 
image

ccing @marcoambrosini as well :)

susnux:
>  for success and error variants, the color of the label when focused should be either --color-success-text or --color-error-text, currently it's the primary color for everything

Already was like this, the last commit "fixing stylelint" broke it :facepalm:
I try to fix this again.

JuliaKirschenheuter:
wow, that is great! Thanks a lot!

susnux:
@nimishavijay 

>     * [ ]  for the text fields without an icon there is still a horizontal shift of the label when it is focused, so if we do `margin-inline: 6px 0;` when focused and `margin-inline: 10px 0;` when unfocused this shift is gone

Fixed that, the label is now start-aligned with the input and with the focused label (So the first letter is always on the same x position)

>     * [ ]  for success and error variants, the color of the label when focused should be either `--color-success-text` or `--color-error-text`, currently it's the primary color for everything

Was a regression of the last commit, fixed again.

On question for me is: Should the text stay that color? Currently it will be max-contrast when not focused to have a visual hover / focus state (as the border color can not change).

> Also, on my system there is some text getting cut off, and it changes according to the zoom level.

Would you mind try again? I fixed the label height.

susnux:
@JuliaKirschenheuter 
> Could you please increase a contrast of labels on not focused state?

The contrast is currently 4.54 : 1
So it should fulfill the requirement :)

nimishavijay:
> Would you mind try again? I fixed the label height.

Works now 👍 

> Should the text stay that color? Currently it will be max-contrast when not focused to have a visual hover / focus state (as the border color can not change).

That's a good point. I would say that it could just stay `--color-text-maxcontrast` as the mouse cursor is changing to indicate hover feedback anyway. It would be inconsistent to change the color only for success and error variants. Since these variants are used not so often, it is okay to just have the mouse cursor change as the hover feedback :)

susnux:
Squashed the fixup commits

PR #4399, opened 2023-08-04
fix: remove NcMultiselect from styleguide config

raimund-schluessler:
The `NcMultiselect` component was removed from the library, this PR removes its leftover from the styleguide config.


ShGKme (approved):
🧹

susnux (approved):

PR #4406, opened 2023-08-09
Fix modal focus trap with additional elements and arrow navigation

ShGKme:
### ☑️ Resolves

* Fix https://github.com/nextcloud/server/issues/39415

Two problems with focus trap break navigation with modal.

**Problem 1:** When Viewer is opened with sidebar, focus on the sidebar is not possible
Source: `additionalTrapElements` are not used on init but only on change. 

**Problem 2:** `NcModal` intercept arrow navigation for slider (next/prev)

#### 🏚️ Before

https://github.com/nextcloud-libraries/nextcloud-vue/assets/25978914/abd9cf42-73bc-4cae-be4f-ebd49bb2b656

https://github.com/nextcloud-libraries/nextcloud-vue/assets/25978914/cae07f32-b44b-42c6-bb40-ee3dac0e59a5

#### 🏡 After

https://github.com/nextcloud-libraries/nextcloud-vue/assets/25978914/7bea1fb3-4618-486a-ab83-6bca3374ad2c

### 🚧 Tasks

- [x] Also use `additionalTrapElements` on init
- [x] On keyboard navigation handler check if there is an active element. 
  - If there is an element and it's outside the modal, then it is
    - either element from `additionalTrapElements`
    - or some pop over element like emoji picker or nc actions that also have its own navigation
  - **Alternative solution:** always disable arrow navigation if there is an active element

### 🏁 Checklist

### DIFF
--- a/src/components/NcModal/NcModal.vue
+++ b/src/components/NcModal/NcModal.vue
@@ -616,18 +616,26 @@ export default {
 			}
 		},
 
-		// Key Handlers
-		handleKeydown(e) {
-			switch (e.keyCode) {
-			case 37: // left arrow
-				this.previous(e)
-				break
-			case 39: // right arrow
-				this.next(e)
-				break
-			case 27: // escape key
-				this.close(e)
-				break
+		/**
+		 * @param {KeyboardEvent} event - keyboard event
+		 */
+		handleKeydown(event) {
+			if (event.key === 'Escape') {
+				return this.close(event)
+			}
+
+			const arrowHandlers = {
+				ArrowLeft: this.previous,
+				ArrowRight: this.next,
+			}
+			if (arrowHandlers[event.key]) {
+				// Ignore arrow navigation, if there is a current focus outside the modal.
+				// For example, when the focus is in Sidebar or NcActions's items,
+				// arrow navigation should not be intercept by modal slider
+				if (document.activeElement && !this.$el.contains(document.activeElement)) {
+					return
+				}
+				return arrowHandlers[event.key](event)
 			}
 		},
 
@@ -714,6 +722,9 @@ export default {
 				allowOutsideClick: true,
 				fallbackFocus: contentContainer,
 				trapStack: getTrapStack(),
+				// Esc can be used without stop in content or additionalTrapElements where it should not deacxtivate modal's focus trap.
+				// Focus trap is deactivated on modal close anyway.
+				escapeDeactivates: false,
 			}
 
 			// Init focus trap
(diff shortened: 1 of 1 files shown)


ShGKme:
There is one more problem: when the next or prev slide is a video, then tab navigation goes to visually hidden video controls. But this should be fixed in Viewer (I'll make a PR later)

AndyScherzinger (approved):

ShGKme:
It also fixes https://github.com/nextcloud/server/issues/39434

ShGKme:
PR is updated.

Only arrow navigation ignores elements outside (which is the problem in mentioned issue), while <kbd>ESC</kbd> still closes the modal as it was before, regardless of focus.

It is a problem for elements that uses <kbd>ESC</kbd> itself. For example, closing `NcSelect` by <kbd>ESC</kbd>. But I think, it's better to fix it on such components sides (with stop propagation of ESC keydown). Otherwise even focus on sidebar element after click prevents closing Viewer by <kbd>ESC</kbd>.

AndyScherzinger (approved):

src/components/NcModal/NcModal.vue
```diff
@@ -714,10 +722,13 @@ export default {
 				allowOutsideClick: true,
 				fallbackFocus: contentContainer,
 				trapStack: getTrapStack(),
+				// Esc can be used without stop in content or additionalTrapElements where it should not deacxtivate modal's focus trap.
+				// Focus trap is deactivated on modal close anyway.
+				escapeDeactivates: false,
```
Pytal, line 727:
```suggestion
```
Not used anymore?

src/components/NcModal/NcModal.vue
```diff
@@ -714,10 +722,13 @@ export default {
...
 			}
 
 			// Init focus trap
-			this.focusTrap = createFocusTrap(contentContainer, options)
+			this.focusTrap = createFocusTrap([contentContainer, ...this.additionalTrapElements], options)
```
Pytal, line 731: Making the `additionalTrapElements` watcher `immediate` probably also works but on init definitely seems more intuitive

Pytal (approved):

src/components/NcModal/NcModal.vue
ShGKme: It won't work. Watcher with `immediate: true` runs on an instance creation before rendering, while the focus trap must be initialized after the modal is actually mounted and shown (opened).

src/components/NcModal/NcModal.vue
ShGKme: Hmm... I'm not sure if it is required, I need to re-check.

But disabling `escapeDeactivates` is definitely safe:
- If a modal is closed by ESC, trap is deactivated anyway
- If a modal is not closed by ESC, the focus trap should also not be deactivated.

susnux (approved):

PR #4407, opened 2023-08-09
fix(NcRichContenteditable): add aria-placeholder

ShGKme:
### ☑️ Resolves

* Fix https://github.com/nextcloud/server/issues/37087

**The solution in this PR is different from the suggestion in the linked issue.**

Using `aria-describedby` requires a layout update, adding a wrapper element for `NcRichContenteditable` with the `div[contenteditable]` itself and a new visually hidden describing element. Though the layout of the element is not a part of its public interface, I think developers expect the component to be a directly `contenteditable` element and I prefer to keep current simple layout as long as it's possible.

A good and very simple alternative solution IMO is [`aria-placeholder`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-placeholder).

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 Tasks

- [x] Add `aria-placeholder`

### 🏁 Checklist

### DIFF
--- a/src/components/NcRichContenteditable/NcRichContenteditable.vue
+++ b/src/components/NcRichContenteditable/NcRichContenteditable.vue
@@ -157,6 +157,7 @@ export default {
 		}"
 		:contenteditable="canEdit"
 		:placeholder="placeholder"
+		:aria-placeholder="placeholder"
 		aria-multiline="true"
 		class="rich-contenteditable__input"
 		role="textbox"
--- a/tests/unit/components/NcRichContenteditable/NcRichContenteditable.spec.js
+++ b/tests/unit/components/NcRichContenteditable/NcRichContenteditable.spec.js
@@ -118,4 +118,17 @@ describe('NcRichContenteditable', () => {
 		expect(handlers.paste).toHaveBeenCalledTimes(1)
 		expect(handlers.blur).toHaveBeenCalledTimes(1)
 	})
+
+	it('should has accessible placeholder from placeholder prop', async () => {
+		const PLACEHOLDER = 'TEST_PLACEHOLDER'
+		const { wrapper } = mountNcRichContenteditable({
+			propsData: {
+				placeholder: PLACEHOLDER,
+			},
+		})
+		// Accessible placeholder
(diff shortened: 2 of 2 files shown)


AndyScherzinger (approved):

AndyScherzinger:
Not an expert in terms of BITV compliance of the solution, else looks good to me.

Pytal:
Makes sense though I suspect @michaelnissenbaum may not find this acceptable for similar reasons as on MDN image

Pytal:
Will likely have to go with one of the alternative solutions depending on his feedback

AndyScherzinger:
Can't we just put the text below the input field as suggested my Michael?

@jancborchardt

michaelnissenbaum:
I don't recommend using placeholders, whether in HTML (`placeholder`) or through `aria-placeholder`.  Instead, place the hint below the input field.

jancborchardt:
Yep, we have enough room in the comments interface, there can be a subline in color-text-maxcontrast.

ShGKme:
I'll make another PR with a new feature to add an always visible description for `NcRichContenteditable`. 

But making the placeholder visible by default as a description will break existing behavior and may not be suitable for any case. I think, in some cases, component users will want to keep using the placeholder as it currently is.

What do you think about merging this first to make the current placeholder accessible and then adding an additional description field?

susnux:
> Using `aria-describedby` requires a layout update

Isn't there a `aria-description` now? I remember seeing this for similar purpose.

Pytal (approved):

susnux (approved):

marcoambrosini:
@jancborchardt this is resulting with a hint implemented on all `NcRichContenteditable` instances. This is not acceptable, we just cannot have this everywhere. (think of Talk for example)

I don't think the comments instructions on how to use the input field are even needed, but regardless of that, let's treat the comments interface as a the only instance of this problem and solve it there? These `NcRichContenteditable` are usually in context with headings and other hints around them that are pretty clear and requiring an hint in all of them will mess up the UI in most places.

PR #4415, opened 2023-08-12
feat(NcDialog): New component (moved from `@nextcloud/dialogs`)

### LINKED ISSUE #631: Filepicker component
In #9 a filepicker component was mentioned as suggestion for a component, but I there is none yet afaik.

I think this would be a helpful improvement.

### LINKED ISSUE #4348: `AppSettingsDialog` does scroll the section list when clicking on entry
When clicking on one of the sections in the navigation of the app settings dialog not only the sections scroll, but also the sections list.

I can reproduce this every time if I have a window which height is smaller than the modal content height.

See this example with chromium:

https://github.com/nextcloud/nextcloud-vue/assets/1855448/ae0caff3-755c-4672-a838-1054e97c6061

susnux:
### ☑️ Resolves

* Fix #631 
* Fix #4348 (fixed by migrating AppSettingsDialog to NcDialog base)

Introduces NcDialog as a generic base Dialog component, while NcModal is simply a base for modals like the viewer, NcDialog provides a convenient interface to define dialogs (see example).
The new FilePicker of the `@nextcloud/dialogs` package is based on top of it.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 Tasks

- Maybe migrate to not use `<script setup>` ? I just copied it, but it would be possible to align with the other components in this library.

### 🏁 Checklist


src/components/NcRichText/NcRichText.vue
```diff
@@ -138,16 +138,15 @@ export default {
...
 				const matches = entry.match(/^\{([a-z\-_.0-9]+)\}$/i)
 				// just return plain string nodes as text
 				if (!matches) {
-					return prepareTextNode({ h, context }, entry)
+					return prepareTextNode({ h, context: this }, entry)
 				}
 				// return component instance if argument is an object
 				const argumentId = matches[1]
-				const argument = context.arguments[argumentId]
+				const argument = this.arguments[argumentId]
```
skjnldsv, line 149: Looks unrelated

susnux: Required to fix eslint errors which would break the CI test. (Comes from introducing Typescript for Vue components).
So I had to either rewrite the code as vanilla Javascript or fix this linting errors.

src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
```diff
@@ -75,9 +75,39 @@ export default {
...
 </docs>
 
+<template>
+	<NcDialog v-if="open"
+		:additional-trap-elements="additionalTrapElements"
+		:container="container"
+		:name="name"
+		:navigation-classes="['app-settings__navigation']"
+		class="app-settings"
+		size="large"
```
szaimen, line 85: shouldnt that better be small by default?

susnux: It is unchanged, it was previously also `large`. Otherwise the navigation would be never shown.

szaimen: what navigation? You mean the buttons at the right? Shouldnt they be positioned below the question?

susnux: No this is the app settings dialog, its about the sections on the left.
image

src/components/NcDialog/NcDialog.vue
```diff
@@ -0,0 +1,352 @@
...
+})
+
+const emit = defineEmits<{
+	/** Emitted when the dialog is about to close but the out transition did not finished yet */
+	(e: 'closing'): void
+	/** Emitted when the open state changed */
+	(e: 'update:open', v: boolean): void
+}>()
+
+const slots = useSlots()
```
ShGKme, line 215: I don't remember if it is supported in Vue 2.7, but in `SFC Setup` there is a `defineSlots` macro to define slots with TS.

ShGKme (commented):
Apart from adding a new component, this PR introduces `TS` and using `SFC Setup`. It requires more tooling updates:
- Webpack config update to support `vue-tsc` instead of using `ts-loader` with `tsc` and to build TS files
- ESLint rules Vue + TS + SFC Setup support
- Possibly, tsconfig update

Also, I'm a bit afraid of using `SFC Setup` in Vue 2. Although it was "backported" at the moment of releasing Vue 2.7, some things were not backported (e.g. Vue 3.3 new stuff) or works differently and has no documentation. And I guess it won't be changed for Vue 2.

What do you think about adding this component first without TS and maybe without SFC Setup but with JS and plain SFC (with Composition API)?

At least after full migration to Vite with another your PR, it will be easier to add first-class TypeScript support 😀

marcoambrosini (commented):
Hi, could with default with a size small or normal for the modal?

susnux:
Close in favor of https://github.com/nextcloud-libraries/nextcloud-vue/pull/4550

PR #4416, opened 2023-08-13
fix: adjust breadcrumbs component

raimund-schluessler:
### ☑️ Resolves

* This PR fixes an issue with the breadcrumbs component where the `getWidth` function errors because it uses `.elm` on a `vnode` which might not be defined. The PR is basically a backport of the changes I did for the `NcBreadcrumbs` component in the `vue3` branch. The PR still needs some cleanup and fine-tuning, but it seems to fix the error.

@susnux The hidden breadcumbs in the action menu currently don't work as they should. The breadcrumbs component might need to be adjusted, but I still need to figure out what's the problem.

### DIFF
--- a/src/components/NcBreadcrumbs/NcBreadcrumbs.vue
+++ b/src/components/NcBreadcrumbs/NcBreadcrumbs.vue
@@ -138,6 +138,7 @@ export default {
 
 <script>
 import NcActions from '../NcActions/index.js'
+import NcActionButton from '../NcActionButton/index.js'
 import NcActionRouter from '../NcActionRouter/index.js'
 import NcActionLink from '../NcActionLink/index.js'
 import NcBreadcrumb from '../NcBreadcrumb/index.js'
@@ -149,6 +150,7 @@ import IconFolder from 'vue-material-design-icons/Folder.vue'
 
 import debounce from 'debounce'
 import Vue from 'vue'
+import { Fragment } from 'vue-frag'
 
 const crumbClass = 'vue-crumb'
 
@@ -156,6 +158,7 @@ export default {
 	name: 'NcBreadcrumbs',
 	components: {
 		NcActions,
+		NcActionButton,
 		NcActionRouter,
 		NcActionLink,
 		NcBreadcrumb,
@@ -173,10 +176,6 @@ export default {
 	emits: ['dropped'],
 	data() {
 		return {
-			/**
-			 * The breadcrumbs which should be shown in a dropdown Actions menu.
-			 */
-			hiddenCrumbs: [],
 			/**
 			 * Array to track the hidden breadcrumbs by their index.
 			 * Comparing two crumbs somehow does not work, so we use the indices.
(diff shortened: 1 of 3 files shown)


raimund-schluessler:
> @susnux The hidden breadcumbs in the action menu currently don't work as they should. The breadcrumbs component might need to be adjusted, but I still need to figure out what's the problem.

This is partially fixed by second commit in the PR now. The hidden crumbs now correctly execute the bound listeners and navigating directories in the filepicker now also works for hidden crumbs. However, clicking a hidden crumb also navigates in the browser and the URL changes, which should not be the case. Maybe this can be skipped with a `preventDefault` on the listener, but I think adjusting `NcBreadcrumbs` to not navigate at all in such cases would be cleaner.

raimund-schluessler:
> However, clicking a hidden crumb also navigates in the browser and the URL changes, which should not be the case. Maybe this can be skipped with a `preventDefault` on the listener, but I think adjusting `NcBreadcrumbs` to not navigate at all in such cases would be cleaner.

@susnux I think the issue is that the hidden crumbs always render to an `ActionLink` or `ActionRouter` which makes the browser navigate. How about we allow also `ActionButton` in case the crumb has neither a `to` nor `href` set? I think we never really considered using `NcBreadcrumbs` in a non navigating context, like it's done in the Filepicker now.

src/components/NcBreadcrumbs/NcBreadcrumbs.vue
```diff
@@ -226,21 +230,21 @@ export default {
 		/**
 		 * Check that crumbs to hide are hidden
 		 */
-		this.delayedHideCrumbs()
+		this.$nextTick(() => {
+			this.hideCrumbs()
+		})
 	},
-	beforeDestroy() {
+	beforeUnmount() {
```
raimund-schluessler, line 237:
```suggestion
	beforeDestroy() {
```

src/components/NcBreadcrumbs/NcBreadcrumbs.vue
```diff
@@ -616,7 +664,7 @@ export default {
 	flex-grow: 1;
 	display: inline-flex;
```
skjnldsv, line 665:
```suggestion
	display: inline-flex;
	align-items: center;
```

skjnldsv (approved):

skjnldsv:
> I know you sure are busy with 27.1, but it would be great if you could have a look at this PR here, and check whether it fixes the issues with the breadcrumb bar in the new vue filepicker component. I think this should also be included for 27.1. because it otherwise could be seen as a regression.

Looks good to me! :)
Added one comment fixing a height issue

raimund-schluessler:
> Looks good to me! :) Added one comment fixing a height issue

Thanks for having a look. Could you check whether it fixes the issue with the crumbs in the filepicker for you? Anyway, I will cleanup the PR a bit and once @susnux had a look, we can merge it.

raimund-schluessler:
I cleaned up a bit. After a rebase and a squash this should be good to go.

susnux (approved):
Works perfectly now! Thank you :tada: 

Just squash and then ready to go :)

PR #4437, opened 2023-08-19
fix: `exports` field of package used deprecated syntax

susnux:
### ☑️ Resolves
Fixes eslint issues of libraries using the package like:
> "@nextcloud/vue/dist/Components/NcButton.js" is not found        n/no-missing-import

Using directory exports is deprecated since Node 14, instead wildcard exports should be used.
This causes eslint-plugin-n to fail because it does not support the deprecated syntax.

See server PR as example of failing lint: https://github.com/nextcloud/server/actions/runs/5908640226/job/16028330790?pr=39970

### 🏁 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
       "import": "./dist/index.mjs",
       "require": "./dist/index.cjs"
     },
-    "./dist/": "./dist/"
+    "./dist/*": "./dist/*"
   },
   "files": [
     "CHANGELOG.md",


ShGKme (approved):
Looks correct according to the current documentation.

The API was deprecated since Node 16 (though, I have found neither mentions about deprecation in Node 14 nor mentions about trailing slash pattern support).

https://nodejs.org/docs/latest-v16.x/api/packages.html#subpath-folder-mappings

Pytal (approved):

susnux:
@ShGKme 
> (though, I have found neither mentions about deprecation in Node 14 nor mentions about trailing slash pattern support).

See DEP0148: Deprecated since Node 14 and removed with Node 17:
https://nodejs.org/api/deprecations.html#dep0148-folder-mappings-in-exports-trailing-

skjnldsv:
Please set `enhancement` or `bug` label next time :)

PR #4451, opened 2023-08-23
fix(NcSelect): Use named export of VueSelect to prevent issues when imported in ESM projects

susnux:
### ☑️ Resolves

> TypeError: _nextcloud_vue_select__WEBPACK_IMPORTED_MODULE_4__.props is undefined

With mixed exports the default is not imported but the whole exports object, can be tested with a simple script.
Just name this script `some-name.mjs` and run `node ./some-name.mjs` (of cause in that directory must be a node_modules with `@nextcloud/vue-select` installed).

```js
import VueSelect from '@nextcloud/vue-select'
console.warn(VueSelect)
```

### 🏁 Checklist

### DIFF
--- a/src/components/NcSelect/NcSelect.vue
+++ b/src/components/NcSelect/NcSelect.vue
@@ -526,7 +526,7 @@ export default {
 </template>
 
 <script>
-import VueSelect from '@nextcloud/vue-select'
+import { VueSelect } from '@nextcloud/vue-select'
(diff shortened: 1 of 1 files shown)


skjnldsv (approved):

ShGKme (approved):
Makes sense

https://github.com/sagalbot/vue-select/blob/857753655adbebae612518aab072ed7f4a8deaa0/src/index.js#L4-L5

PR #4454, opened 2023-08-23
feat(NcModal): Allow to configure if the modal is closed when clicked outside

susnux:
### ☑️ Resolves

Currently the modal is closed if `canClose` is `true` and the user clicked outside the modal on the backdrop. This can now be configured with the `closeOnClickOutside` property (defaults to `true` for backwards compatibility).

This is especially helpful for implementing the dialogs.

### 🏁 Checklist

### DIFF
--- a/src/components/NcModal/NcModal.vue
+++ b/src/components/NcModal/NcModal.vue
@@ -252,7 +252,7 @@ export default {
 						spreadNavigation ? 'modal-wrapper--spread-navigation' : ''
 					]"
 					class="modal-wrapper"
-					@mousedown.self="close">
+					@mousedown.self="handleClickModalWrapper">
 					<!-- Navigation button -->
 					<transition name="fade-visibility" appear>
 						<NcButton v-show="hasPrevious"
@@ -420,6 +420,7 @@ export default {
 				return ['small', 'normal', 'large', 'full'].includes(size)
 			},
 		},
+
 		/**
 		 * Declare if the modal can be closed
 		 */
@@ -427,6 +428,16 @@ export default {
 			type: Boolean,
 			default: true,
 		},
+
+		/**
+		 * Close the modal if the user clicked outside of the modal
+		 * Only relevant if `canClose` is set to true.
+		 */
+		closeOnClickOutside: {
+			type: Boolean,
+			default: true,
+		},
+
 		/** Makes the modal backdrop black if true  */
 		dark: {
 			type: Boolean,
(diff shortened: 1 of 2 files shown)


marcoambrosini (commented):
Great addition :) 

Code looks good but I'm testing this in the docs and adding `close-on-click-outside="false"` doesn't work. Am I missing something?

susnux:
> Great addition :)
> 
> Code looks good but I'm testing this in the docs and adding `close-on-click-outside="false"` doesn't work. Am I missing something?

No `:close-on-click-outside="false"` should be enough:
See here, clicking close always work, clicking outside only works in the first example.

https://github.com/nextcloud-libraries/nextcloud-vue/assets/1855448/a41684a2-7980-4b76-b2d2-a7674b5d1777

Pytal (approved):

marcoambrosini (approved):

skjnldsv (approved):

src/components/NcModal/NcModal.vue
```diff
@@ -420,13 +420,24 @@ export default {
...
 		},
+
+		/**
+		 * Close the modal if the user clicked outside of the modal
+		 * Only relevant if `canClose` is set to true.
+		 */
+		closeOnClickOutside: {
+			type: Boolean,
+			default: true,
+		},
```
ShGKme, line 439: In a standard form of passing props Vue enforces HTML standard aka:

```html
<input /> <!-- enabled-->
<input disabled /> <!-- disabled -->
```

But boolean props with default `true` requires to always use longhand form over HTML Standard:

```vue
<NcModal /> <!-- closes on click -->
<NcModal close-on-click-outside /> <!-- also closes on click, passing bool prop does nothing -->
<NcModal :close-on-click-outside="false" /> <!-- changes behavior, doesn't close on click -->
<NcModal close-on-click-outside="false" /> <!-- casts string "false" to true and closes on click -->
```

because of that it is usually recommended to make bool props with `false` default value:

```vue
<!-- With default false boolean prop -->
<NcModal /> <!-- closes on click -->
<NcModal ignore-click-outside /> <!-- changes behavior, doesn't close on click -->
```

- https://eslint.vuejs.org/rules/prefer-true-attribute-shorthand.html
- https://eslint.vuejs.org/rules/no-boolean-default.html

susnux: Thank you :)
I know but did not wanted to make this a breaking change by changing the default behavior and a negated property seems to be confusing.
Maybe we can risk another breaking change?

PR #4456, opened 2023-08-24
fix(a11y): Remove nav and radios in sidebar tabs

Pytal:
### ☑️ Resolves

- For https://github.com/nextcloud/server/issues/37107

### 🚧 Tasks

- Remove redundant `nav` for element with tablist role
	image
	> MDN: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav
- Remove radios in favour of input with `type="button"`
	image
	> MDN: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/button

	> In theory, this could lead to issues.
	> —michaelnissenbaum

Ref: https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/#sc1_label

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppSidebar/NcAppSidebarTabs.vue
+++ b/src/components/NcAppSidebar/NcAppSidebarTabs.vue
@@ -28,7 +28,7 @@
 	<div class="app-sidebar-tabs">
 		<!-- tabs navigation -->
 		<!-- 33 and 34 code is for page up and page down -->
-		<nav v-if="hasMultipleTabs"
+		<div v-if="hasMultipleTabs"
 			role="tablist"
 			class="app-sidebar-tabs__nav"
 			@keydown.left.exact.prevent="focusPreviousTab"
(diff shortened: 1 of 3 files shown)


src/components/NcAppSidebar/NcAppSidebarTabs.vue
```diff
@@ -50,7 +50,7 @@
 				class="app-sidebar-tabs__tab"
 				:class="{ active: tab.id === activeTab }"
 				role="tab"
-				type="radio"
+				type="button"
```
susnux, line 53: This is not valid, NcCheckboxRadioSwitch only accepts checkbox radio and switch als property values.

susnux: Thats new :D

susnux (approved):

ShGKme (approved):
Works with a screen reader, but a bit buggy with both screen reading and Tab navigation sometimes, because each tab button actually is 2 focusable elements: `<input type="button">` and `<label>`. Though input has position absolute, it is still focusable.

ShGKme:
With Tab navigation you can see sometimes 2 different focus styles

tabs-1

ShGKme:
When actual button is button was selected, navigation with screen reader also goes through both elements.
The first is "Search tab button" and the second is "Search tab one of three"

tabs-2

PR #4461, opened 2023-08-24
fix(`NcRichText`) - keep newlines in rendered Markdown 

Antreesy:
### ☑️ Resolves

* Markdown eats newlines (`Enter`, `Shift+Enter`), because rendered `<span/>` element doesn't contain symbols, so it's `height === 0`
* For newline (several `\n\n` in a row) - we paste no-break space (unicode `\u00A0`) between to fill the line
* Also replaced ~~strikethrough~~ with `inline code`, as we don't support it yet

### 🖼️ Screenshots

🏚️ Before

[Screencast from 24.08.2023 14:27:11.webm](https://github.com/nextcloud-libraries/nextcloud-vue/assets/93392545/8fa83492-da04-4abe-91a7-c644cedff4fe)

🏡 After

[Screencast from 24.08.2023 14:26:17.webm](https://github.com/nextcloud-libraries/nextcloud-vue/assets/93392545/cb12c42c-6b0e-4071-bd42-ed1c40118249)

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/NcRichText/NcRichText.vue
+++ b/src/components/NcRichText/NcRichText.vue
@@ -41,7 +41,7 @@ export default {
 
 Local IP: http://127.0.0.1/status.php should be clickable
 
-Some examples for markdown syntax: **bold text** *italic text* ~~strikethrough~~`,
+Some examples for markdown syntax: **bold text** *italic text* \`inline code\``,
 			autolink: true,
 			useMarkdown: true,
 			args: {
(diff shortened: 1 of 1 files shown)


nickvergessen (approved):

Pytal:
Adding label for auto-generated release notes

dartcafe:
Not sure, if this matches the normal (official?) markdown bevahior: 
* an empty line should create a paragraph `<p></p>` around the previous lines
* a simple line break `<br>` is created while adding two spaces at the end of a line (sometimes with a backslash or `<br>`)

Fair enough: Normal users will not be aware of the 'two spaces at the end of a line' thing. But at least the paragraph renderimg makes sense.

Antreesy:
It's a workaround because of active rule `white-space-collapse: collapse`

Described by you behavior seems to match the current state: 
image

dartcafe:
The span tags are there for a reason? They look weird and unusual.

Antreesy:
Thought the same, but it's a result of current parsing process. They probably could be removed, but with some deep changes I wouldn't like to touch for a moment

PR #4486, opened 2023-08-30
fix(NcRichText) - don't strip `<tag>`-like content when parsing markdown

### LINKED ISSUE #3908: `NcRichContenteditable` pasting HTML alike content strips off text
Maybe related to #3264 but even worse:

Pasting HTML alike content strips content instead of replacing it with the HTML entities,
For example pasting
````markdown
```java
Test<Object> myMethod() {
}
```
````

Will result in:
````markdown
```java
Test myMethod() {
}
```
````

image

### PR DESCRIPTION
**Pull request by Antreesy**
### ☑️ Resolves

* Replace legacy package https://github.com/remarkjs/remark-external-links/releases/tag/9.0.1
* Fix NcAvatar render for shown initials, when use different variant of `white-space` rule (common for rich components)
* Ease HTML tree for markdown parsed content:
  * before each child node (also text node) was wrapped with span before inserted into parent node
  * now they are inserted directly
  * Revert some previous changes to restore behavior back to v7.12.2 and older (forced new lines):
    * https://github.com/nextcloud-libraries/nextcloud-vue/pull/4461
    * https://github.com/nextcloud-libraries/nextcloud-vue/pull/4470
  * Apply some base styling to markdown rendered content:
    * margin-bottom for separate text blocks
    * bold headings
    * aligned lists
    * stylised blockquotes
* Escape/unescape special symbols `<` and `>` before parsing:
  * Leading `<` could be treated like a XML tag start and be stripped with inner content (because of possible XSS attack)
  * Leading `>` should be unescaped to correctly parse blockquotes
* Fix #3908
* Fix https://github.com/nextcloud/spreed/issues/10412

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
commits 71c2cbddec02fa17f0a919938927641783cbb3ac, 836b7c897f1f523af590b84f91e196b6be79c828 | No visual changes
commit 2e457f3a24b7d389d441f867b12962b5bee7d753 | `NcAvatar` fix
image | image
commit 62fa807dc3bfd76c56e82f4616b882343af87ea6 | `ease HTML` fix
image | image
commit 22d0c6fa395e164d373c3c344e64f7e06b52c350 | Revert old commits, add base styling
image | image
commit 38606f27450fa2dbdb59d68a974e1ea066b737b5 | Escape XML when parsing markdown 
image | image

### 🚧 Tasks

- [ ] Code review
- [ ] Visual check
- [ ] Security check

### 🏁 Checklist

### CODE REVIEW

File: src/components/NcRichContenteditable/docsHelpers.js [hunk 1]
```diff
@@ -0,0 +1,95 @@
...
+			displayName: 'Test01',
+		},
+	},
+	Test02: {
+		component: 'NcUserBubble',
+		props: {
+			displayName: 'Test02',
+		},
+	},
+	// FIXME mentions below don't show up in the NcRichText examples
```
nickvergessen, line 76: they need to be posted as `@"Test Offline"` etc.

Antreesy: Seems like it, but the NcRichText doesn't render it with usual pattern
image

src/components/NcRichText/NcRichText.vue
```diff
@@ -192,6 +262,12 @@ export default {
 				.use(remarkPlaceholder)
 				.use(rehype2react, {
 					createElement: (tag, attrs, children) => {
+						// Special symbol ">" is transformed back for simple text nodes
```
DorraJaouad, line 265: "<" *

DorraJaouad (approved):
Tested and it works. 
It solves the issue of skipping HTML-like content though there might be a better solution as follow-up .

Antreesy:
Rebased onto master, split commits with documentation to different PR #4507

src/components/NcRichText/NcRichText.vue
```diff
@@ -217,15 +223,15 @@ export default {
...
-				.processSync(this.useMarkdown
-				// In order to correctly show newlines in Markdown,
-				// each newline contains a non-breaking space
-					? this.text.slice()
-						.replace(/\n>\n/g, '\n>\u00A0\n')
-						.replace(/\n{2,}/g, (match) => {
-							return '\n' + '\n\u00A0\n'.repeat(match.length - 1)
-						})
-					: this.text)
+				.processSync(this.text.slice()
```
ShGKme, line 226: What is the purpose of `slice` here?

src/components/NcRichText/NcRichText.vue
```diff
@@ -217,15 +223,15 @@ export default {
...
+					// escape special symbol "<" to not treat text as HTML
+					.replace(/</gmi, '&lt;')
+					// unescape special symbol ">" to parse blockquotes
+					.replace(/&gt;/gmi, '>')
+					// add a non-breaking space to each newline to show them
+					.replace(/\n>\n/g, '\n>\u00A0\n')
+					.replace(/\n{2,}/g, (match) => '\n' + '\n\u00A0\n'.repeat(match.length - 1)),
```
ShGKme, line 233: It works and it's not a part of this PR. But non-breaking space is still a space, it's additional character in the rendered text, it's being selected, copied... :(

ShGKme (commented):
Works weird with multiline new lines in quotes

Text | Render
---|---
image | image
image | image
image | image

src/components/NcRichText/NcRichText.vue
ShGKme: Copy from `NcRichText` output
image

Antreesy: We could then proceed with the same approach Github does it: add margin-bottom after some tags (p, ul, ol, blockquote), and get rid of new lines

src/components/NcRichText/NcRichText.vue
Antreesy: I thought `replace` is the mutating method 🙈

src/components/NcRichText/NcRichText.vue
ShGKme: Line1
Line2


Line3

ShGKme: It seems that GitHub just doesn't allow to add any number of new lines and works like MD here. `\n{2,}` works as a paragraph separator and just adds a paragraph with margin. Only single `\n` works as new line.

ShGKme: But we can do it in a separate PR later if it is not a part of this PR, don't feel blocked :D

src/components/NcRichText/NcRichText.vue
ShGKme: strings are immutable in js :P

src/components/NcRichText/NcRichText.vue
Antreesy: GitHub `markdown-body` styles for the reference

https://github.githubassets.com/assets/primer-15214e5f9104.css

Antreesy: Other usages of Markdown across NC apps:

https://github.com/nextcloud/spreed/blob/144f7a61de0689f72c27313538ea883ff28efe2b/src/components/MessagesList/MessagesGroup/Message/Message.vue#L1057
https://github.com/nextcloud/welcome/blame/ef1bdfe51aea6516ccbbd06b54e91439ee502510/src/views/Dashboard.vue#L117
https://github.com/nextcloud/integration_github/blob/585ea8e1fd5f93b44d13ddd8ac6d00080ececb0f/src/views/GithubIssuePrReferenceWidget.vue#L657

Additional spaces between `\n` should be removed, to restore initial parsing (before my first changes) and not break markdown in other apps unintentionally.

But regarding the styles, looks like everyone does it in theirs own way.
We could either stick with that, and don't define any additional styles in the vue-lib, or setup some common things, like `margin` or `color` alignments

Antreesy:
Rebased onto master with some changes



This returns a behavior, then Markdown parser eats newlines (`Enter`, `Shift+Enter`), but it seems to be a default behavior for parsers (see how Github also eats several newlines - edit this comment)

DorraJaouad (approved):

PR #4487, opened 2023-08-30
enh(a11y): Add nav to header menu

Pytal:
- For https://github.com/nextcloud/server/issues/37095

### Summary

- Add opt-in `nav`
- Remove role semantics, ref: https://inclusive-components.design/menus-menu-buttons/#themenuandmenuitemroles

### 🏁 Checklist

### DIFF
--- a/src/components/NcHeaderMenu/NcHeaderMenu.vue
+++ b/src/components/NcHeaderMenu/NcHeaderMenu.vue
@@ -75,6 +75,7 @@ export default {
 			ref="trigger"
 			class="header-menu__trigger button-vue"
 			:aria-label="ariaLabel"
+			:aria-describedby="description ? descriptionId : null"
 			:aria-controls="`header-menu-${id}`"
 			:aria-expanded="opened.toString()"
 			@click.prevent="toggleMenu">
@@ -82,6 +83,12 @@ export default {
 				is at least 16px. Usually mdi icon works at 20px -->
 			<slot name="trigger" />
 		</button>
+
+		<span v-if="description"
+			:id="descriptionId"
+			class="header-menu__description hidden-visually">
+			{{ description }}
+		</span>
 
 		<!-- Visual triangle -->
 		<div v-show="opened" class="header-menu__carret" />
@@ -152,6 +159,15 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+
+		/**
+		 * Additional visually hidden description text for the menu
+		 * open button
+		 */
+		description: {
+			type: String,
+			default: null,
+		},
 	},
 
 	emits: [
@@ -169,6 +185,7 @@ export default {
 			opened: this.open,
 			shortcutsDisabled: window.OCP?.Accessibility?.disableKeyboardShortcuts?.(),
 			triggerId: GenRandomId(),
+			descriptionId: GenRandomId(),
 		}
 	},
 
--- a/src/components/NcHeaderMenu/NcHeaderMenu.vue
+++ b/src/components/NcHeaderMenu/NcHeaderMenu.vue
@@ -64,12 +64,15 @@ export default {
 </docs>
 
 <template>
-	<div :id="id"
+	<component :is="wrapperTag"
+		:id="id"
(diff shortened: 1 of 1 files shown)


artonge (approved):

src/components/NcHeaderMenu/NcHeaderMenu.vue
```diff
@@ -139,6 +140,22 @@ export default {
...
+			default: false,
+		},
+
+		/**
+		 * aria-label attribute of the wrapper element
+		 */
+		wrapperAriaLabel: {
+			type: String,
+			default: null,
+		},
```
ShGKme, line 158: Do we need 2 different aria labels, one on the whole nav and one on the trigger?

ShGKme: Isn't it always supposed to be the same as trigger, so we can use `aria-labledby` here?

Pytal: Yes
image

See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role#labeling_landmarks

ShGKme: I still didn't understand... Could you make an example of when this component should have both `wrappedAriaLabel` and `ariaLabel`, and they are different?

Pytal: See the server draft https://github.com/nextcloud/server/pull/40168/files#diff-19af9ee8873f4260b63dd8d5974181daf4e53ab74d1d3ee84edbfd1975b215eaR27-R28

With reference to
image
> https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role#labeling_landmarks

and

image
> https://www.unimelb.edu.au/accessibility/aria/wai-aria-landmarks-and-screen-readers

when navigating landmark regions with <kbd>d</kbd> and without unique labels, screen readers will announce them both identically as "Navigation landmark"

With unique labels users may identofy them as distinct navigation landmarks i.e. read out as "App menu (navigation landmark)" and "Settings menu (navigation landmark)"

ShGKme: > With unique labels users may identify them as distinct navigation landmarks i.e. read out as "App menu (navigation landmark)" and "Settings menu (navigation landmark)"

But "App menu" and "Settings menu" will have different labels, because they are 2 different navigation - 2 different instances on `NcHeaderNav`.

ShGKme: In `NcHeaderNav` we have a navigation that is opened by a button. This button controls the navigation and its label describes the navigation.

If I understand you correctly, when we use 2 labels here, we need to always manually create and control this 2 labels. And the second label is not supposed to be anything except "open" + the first label, right? Like they are always "X menu" and "Open X menu".

In this case we can use only one label on the controlling button "X menu" and make navigation `labelledby` this button.

For example, for settings menu, when a user goes to the next landmark, screen reader reads it as:
```
Settings menu navigation landmark Open settings menu button collapsed
```

But with one label and `labelled-by` it is almost the same, it is still a navigation with a button:
```
Settings menu navigation landmark Settings menu button collapsed
```

Calling button just "Settings menu" without "Open" verb seems to be to be a common practice in both Nextcloud and w3c examples (https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/examples/menu-button-links/).

In the issue example link also `aria-labelledby` is used, but with header (there is no example with both opening button and labels :( ).

ShGKme: As a summarize I'm against using to an additional prop here because:
- They are always almost the same, "X menu" and "Open X Menu", but it is something on developer's responsibility, more to describe in the documentation
- One more string to translate for each nav
- It doesn't change much for a screen reader

Pytal:
`aria-describedby` added for extended description i.e. "Avatar of admin - do not disturb"

ShGKme (approved):

PR #4506, opened 2023-09-06
feat(NcEmptyContent)!: make empty content centered by default instead of 20vh margin

ShGKme:
### ☑️ Resolves

By default `NcEmptyContent` displays the content with fixed 20vh top margin. This doesn't look good sometimes. I think that could be useful to have centered `NcEmptyContent`.

For example, by default, it looks bad in comments, where the default margin is overridden now with `!important`. 

Current `!important` override | Default styles | New `centered`
---|---|---
image | image | image

Or in Talk.

Large screen (almost centered, but not) | Small screen (cut, but there is space)
---|---
image | image

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

Also still works:

NcAppSidebar | ReferencePicker
---|---
image | image

### 🚧 Tasks

- [x] Remove `margin-top: 20vh`
- [x] Make content centered
- [x] Add `flex: 1` to make it centered by default in flex parents
- [x] Remove unnecessary styles in other components that remove margin and check that nothing it breaking
  - `NcHeaderMenu`, `NcDashboardWidget` - custom styles, nothing breaking
  - `NcAppSidebar` - works fine by default because of flex
  - `NcReferencePicker`: `NcProviderList`, `NcRawLinkInput`, `NcSearch` - remove unnecessary `margin: 0`, works fine now by default

### 🏁 Checklist

### DIFF
--- a/src/components/NcDashboardWidget/NcDashboardWidget.vue
+++ b/src/components/NcDashboardWidget/NcDashboardWidget.vue
@@ -351,11 +351,9 @@ export default {
 <style scoped lang="scss">
 .dashboard-widget :deep(.empty-content) {
 	text-align: center;
-	margin-top: 0;
 	padding-top: 5vh;
 	&.half-screen {
 		padding-top: 0;
-		margin-top: 0;
 		margin-bottom: 1vh;
 	}
 }
--- a/src/components/NcEmptyContent/NcEmptyContent.vue
+++ b/src/components/NcEmptyContent/NcEmptyContent.vue
@@ -168,11 +168,19 @@ export default {
 	name: 'NcEmptyContent',
 
 	props: {
+		/**
+		 * A header message about an empty content shown
+		 * @example 'No comments'
+		 */
 		name: {
 			type: String,
 			default: '',
 		},
 
+		/**
+		 * Desription of the empty content
+		 * @example 'No comments yet, start the conversation!'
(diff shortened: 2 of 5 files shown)


skjnldsv:
It was the old behaviour, but to be honest, maybe just make it centered straight away?
No prop?

ShGKme:
> It was the old behaviour, but to be honest, maybe just make it centered straight away?
> No prop? I think it make sense to stop using it with this weird top margin

We can, but it will be a breaking change.

Centered works well when component's user stretch it and the way it's been streatched may depends on a specific layout. For example, `flex: 1 0;` for flex container, `height: 100%` in some cased for element, in another for a its flex container and etc...

AndyScherzinger (approved):

GretaD (approved):

ShGKme:
PR is updated. Now it is centered by default.

It works fine in all other `nextcloud/vue` components by default, I just removed unnecessary styles.

But it is a breaking change, in apps in parents without height and flex it will have not top margin now.

ShGKme:
Missclick ><

Pytal (approved):

AndyScherzinger:
> Edit: I guess the backport bot is still dead.

Yes, we didn't find the time yet...

PR #4511, opened 2023-09-08
chore(vue3): Migrate NcPopover

raimund-schluessler:
This migrates the `NcPopover` component to vue 3.


ShGKme (approved):

src/components/NcPopover/NcPopover.vue
```diff
@@ -106,10 +106,8 @@ The prop `:focus-trap="false"` help to prevent it when the default behavior is n
 	<Dropdown ref="popover"
 		:distance="10"
 		:arrow-padding="10"
-		v-bind="$attrs"
```
juliushaertl, line 109: I might not have the full picture but wouldn't those removals be a behavioral change and limit the usage of the component?

raimund-schluessler: They should be automatically inherited / fallthrough. @ShGKme correct me if I am wrong here.

juliushaertl: Ah interesting, good to know. Docs seem to proof https://vuejs.org/guide/components/attrs.html :)

juliushaertl (approved):

src/components/NcPopover/NcPopover.vue
ShGKme: For better understanding what's under the box.

In Vue 2 props and attrs where different parts of vnode. Then some params is passed to the node, it is passed either as prop or as attr. For example:

```vue
<Comp foo="foo" bar="bar" />
```

If only `foo` is a Comp's prop, then it is rendered as:
```js
h(Comp, { 
  props: { foo: 'foo' },
  attrs: { bar: 'bar' },
})
```

And no matter if bar is a prop for the root node of `Comp` or not. It is passed as attr and it is inheritted as an attr. First to $attrs, then to HTML attributes of the last node.

In Vue 3 they are flattened and passed just as params. So

```vue
<Comp foo="foo" bar="bar" />
```

compiles to the next code, no matter what is a prop, just vnode params:

```js
h(Comp, {
  foo: 'foo',
  bar: 'bar',
})
```

And then the node decides if it is a part `$props` or `$attrs`. Then all attrs are passed the same way on the Comp's root, event if it is a component.

ShGKme: image

PR #4512, opened 2023-09-09
chore(vue3): Migrate Nc*Field

raimund-schluessler:
This migrates the `Nc*Field` components to vue 3.


src/components/NcInputField/NcInputField.vue
```diff
@@ -260,7 +260,7 @@ export default {
 		},
 
 		hasLeadingIcon() {
-			return this.$slots.default
+			return isSlotPopulated(this.$slots.default?.())
```
ShGKme, line 263: **question (non-blocking):** checking slots

Do we need to change behavior here and check the content of the slot, not only if the slot is passed or not?

Also, it is not reactive, it won't work as a `computed`.

src/components/NcPasswordField/NcPasswordField.vue
```diff
@@ -96,7 +96,7 @@ export default {
 </docs>
 
 <template>
-	<NcInputField v-bind="{...$attrs, ...$props }"
+	<NcInputField v-bind="{ ...$props }"
```
ShGKme, line 99: **suggestion:** pass without cloning

Is there any purpose of creating an object here? It creates an object on each re-render...

```suggestion
	<NcInputField v-bind="$props"
```

src/components/NcInputField/NcInputField.vue
ShGKme, line 1: **issue (blocking):** breaking change according to `class` inheritance 

There is a kind of a breaking change.

https://v3-migration.vuejs.org/breaking-changes/attrs-includes-class-style.html

In Vue 2 `class` passed classes to a root node `<div>` and `inputClass` props passed classes to the `<input>`, because `class` was not a part of `$attrs`.

In Vue 3 it is, so both `class` and `inputClass` pass classes to `<input>` and there is no way to pass classes to the root node.

I think, a simple solution here is to manually pass `:class="$attrs.class"` to the root.

ShGKme (changes requested):

src/components/NcPasswordField/NcPasswordField.vue
raimund-schluessler: No, this is a leftover from my debugging. It works fine without cloning.

src/components/NcInputField/NcInputField.vue
raimund-schluessler: I think we nowhere check the slot content. So I would tend to keep it this way.

The reactivity is an issue though. But I think this is a general problem of using `$slots` outside the render function or template in vue 3.
One solution might be to use `isSlotPopulated(this.$slots.default?.())` directly in the template. But it's used three times, which makes it a bit ugly. The other option would be to move to a render function completely, but that would be overkill, I think. Other ideas are welcome, though.

ShGKme: What about just keeping slot existence?
```js
return !!this.$slots.default
```

raimund-schluessler: That doesn't work in vue 3 anymore (or at least I didn't get it to work). It also exists even if it just contains a comment or a Fragment that won't actually render anything. See here for example: https://github.com/vuejs/core/issues/4733

src/components/NcInputField/NcInputField.vue
raimund-schluessler: I defined a `class` prop and added `$props.class` to the root element, so `class` does not get forwarded to the input element.

src/components/NcInputField/NcInputField.vue
raimund-schluessler: @ShGKme See #4516.

ShGKme: > That doesn't work in vue 3 anymore (or at least I didn't get it to work).

Not completely. It still works fine to check if a slot was passed or not, which is enough in most cases. It doesn't work only to check if a passed slot has some content or not.

I understand using `isSlotPopulated` in our components like `<NcButton>` where we actually get text content from the slot. When it is not about text content, but with things like buttons/icons that are not generated, I'd prefer to force users to not pass empty slots.

```vue
<!-- simple check doesn't work because slots are passed but have no real content -->
<Comp>
  <template #foo>
    <!-- nothing -->
    <div v-if="false">Not rendered</div>
    {{ mayBeEmptyString }}
  </template>
  <template #bar></template>
</Comp>

<!-- But we should not use v-if on content when slot is optional -->
<Comp>
  <template #icon>
    <Icon v-if="withIcon" />
  </template>
</Comp>

<!-- Instead, we can use v-if to define is a slot is to be passed, and it works -->
<Comp>
  <template v-if="withIcon" #icon>
    <Icon />
  </template>
</Comp>
```

And I don't consider this a breaking change, because it's a Vue's change, not `NcInputField`'s

ShGKme: > @ShGKme See #4516.

Not sure what to see there...

"Both `!!this.$slots.default` and `!!this.$slots.default?.()` will basically always be true" - only for passed slot that we want to check for emptiness. Not for checking if a slot is passed:

check-slots

https://replit.com/@ShGKme/Vue3-Check-slots

src/components/NcInputField/NcInputField.vue
```diff
@@ -235,6 +235,13 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		/**
+		 * Class to add to the root component.
+		 */
+		class: {
+			type: [Object, String],
+			default: '',
+		},
```
ShGKme, line 244: Array is also a valide `class` value

```suggestion
		class: {
			type: [Object, String, Array],
			default: '',
		},
```

src/components/NcInputField/NcInputField.vue
raimund-schluessler:
```suggestion
			return !!this.$slots.default
```

src/components/NcInputField/NcInputField.vue
```diff
@@ -260,7 +267,7 @@ export default {
 		},
 
 		hasLeadingIcon() {
-			return this.$slots.default
+			return !!this.$slots.default
```
raimund-schluessler, line 270:
```suggestion
			return isSlotPopulated(this.$slots.default?.())
```

src/components/NcInputField/NcInputField.vue
raimund-schluessler: This does not work with the way we use `NcInputField` in `NcTextField` and `NcPasswordField`:

| `return !!this.$slots.default` (always true): | `return isSlotPopulated(this.$slots.default?.())` (works correctly):
|-|-|
| grafik | grafik |

I am open to any other working solution, but simply using `return !!this.$slots.default` here does not seem to do the trick. Or I am missing something.

ShGKme: Hmm, indeed. Let me check again.

src/components/NcTextField/NcTextField.vue
```diff
@@ -130,10 +132,9 @@ export default {
...
 
 <template>
-	<NcInputField v-bind="{...$attrs, ...$props }"
+	<NcInputField v-bind="$props"
 		ref="inputField"
 		:trailing-button-label="clearTextLabel"
-		v-on="$listeners"
 		@input="handleInput">
 		<!-- Default slot for the leading icon -->
 		<slot />
```
ShGKme, line 140: Because we use the `default` slot to pass an icon and shorthand syntax is used here, all the content described between `<NcInputField>` tags is always passed to the default slot (including empty lines, comments, and etc.). So `NcInputField` always receives and renders the default slot.

But if we use full syntax to pass only `<slot />` to the default `slot`, we can pass through the default slot only if a wrapper also has a slot content.

```suggestion
		<template v-if="$slots.default" #default>
			<slot />
		</template>
```

raimund-schluessler: But then we also need to put the icons in a template tag with a `v-if`.

So instead of this:
```vue
<NcTextField value="Text">
	<Icon1 v-if="icon1" />
	<Icon2 v-else-if="icon2" />
	<Icon3 v-else-if="icon3" />
</NcTextField>
```
one would need
```vue
<NcTextField value="Text">
	<template v-if="icon1 || icon2 || icon3" #default>
		<Icon1 v-if="icon1" />
		<Icon2 v-else-if="icon2" />
		<Icon3 v-else-if="icon3" />
	</template>
</NcTextField>
```
no?

I think I don't like this. It seems to add a lot of duplicate code and introduce an unnecessary breaking change.

ShGKme: > no?

Yes, in this case, where there can be a number of possible icons and at the same time there could be no icon at all - it won't work without such v-if.

```vue
<!-- Works, always has content -->
<NcTextField value="Text">
	<Icon1 v-if="icon1" />
	<Icon2 v-else-if="icon2" />
	<Icon3 v-else="icon3" />
</NcTextField>

<!-- Works, pass content if needed -->
<!-- Doesn't work with the current solution because `computed` doesn't react on `$slots`, only on render function inside it -->
<NcTextField value="Text">
	<template v-if="icon1" #default>
		<Icon1 />
	</template>
</NcTextField>

<!-- Doesn't work, but it's Vue 3's breaking change -->
<NcTextField value="Text">
	<Icon1 v-if="icon1" />
</NcTextField>

<!-- Doesn't work without "v-if="icon1 || icon2 || icon3" -->
<NcTextField value="Text">
	<template #default>
		<Icon1 v-if="icon1" />
		<Icon2 v-else-if="icon2" />
		<Icon3 v-else-if="icon3" />
		<!-- else no icon -->
	</template>
</NcTextField>
```

raimund-schluessler: I feel like this reactivity caveat will get back to us at some point. But I guess we can fix this later, as it is an general vue3 issue, if I understood correctly.

@ShGKme Is there a different solution to this problem then migrating to a render function? Seems a little overkill to me. Maybe don't make it a computed, but just use `$slots.default` in the template directly?

raimund-schluessler:
Rebased.

src/components/NcTextField/NcTextField.vue
raimund-schluessler: At least this works:
```vue
<NcTextField v-model:value="text1">
	<Magnify v-if="text1" />
</NcTextField>
```
although it's rather a quirk/side effect than actually intended.

ShGKme: I see 4 solutions here:
1. Just check for `!!$slots.default` in a template/method
   - It doesn't work for passing empty content like your example with 3 icons or when `v-if` is used on content and not on `<template v-slot>`
2. Call `isStopPopulated()` in template/method (the same effect as re-writing to render function)
   - Works in any case
   - A bit not great in performance, slot is rendered twice. But it should not be a problem for just an icon slot
3. Keep the current solution:
   - Slot is rendered twice only when changed
   - No reaction when `v-if` is used on `<template v-slot>`, but we can combine `isSlotPopulated` with `!!$slots.default` check in the template
4. Overkill solution - save `vnodes` from `slotVnodes = $slots.default()` and then reuse them - pass to `isSlotPopulated(slotVnodes)` and render as `<component :is="slotVnodes">` instead of `<slot>`
   - Solves all problems
   - Overkill 🙈
   
No is great for me :D

Speaking about only this specific component, I'd say that #1 is fine just because I don't see a case for `NcTextField` to dynamically add and remove an icon. It would move all the input content with padding, and it seems that it is usually static.

But if I'm wrong and there is a case, then I'm totally fine with using `isSlotPopulated()` as it is used in your solution (and maybe + check `!!$slots.default` in the template just in case).

For components with a large slot content, I'd go with the 4th solution to avoid this (possibly small) performance issue.

raimund-schluessler: I still think that # 1 is not an option as it forces devs to wrap the default slot content in a `<template>` and while for sure correct, I have never seen anyone doing so by default.
# 2 might be ok here, since it is a small component and the slot usually contains an icon only, which should be cheap to render.
# 3 is also fine with me. We can still adjust if it turns out to be an issue.
# 4 probably also fine, but as you said, a bit of an overkill here.

Honestly, I would tend towards # 3 at the moment, keep the issue in mind, but move on for now and fix it if it really becomes a problem.

skjnldsv (approved):

PR #4515, opened 2023-09-09
chore(vue3): Migrate Focus directive, globally import Tooltip in docs

raimund-schluessler:
This migrates the `Focus` directive to vue 3.

### DIFF
--- a/styleguide/preview.js
+++ b/styleguide/preview.js
@@ -1,0 +1,8 @@
+import { Tooltip } from '../src/directives/index.js'
+
+// The export here MUST be default or module.export
+// this is what is imported by the styleguide
+export default (app) => {
+  app.directive('tooltip', Tooltip)
+}
+
--- a/src/directives/Focus/index.js
+++ b/src/directives/Focus/index.js
@@ -21,7 +21,7 @@
  */
 
 export const directive = {
-	inserted(el) {
+	mounted(el) {
 		el.focus()
 	},
 }
--- a/src/index.js
+++ b/src/index.js
@@ -24,7 +24,7 @@ export * from './components/index.js'
 
 // Not yet adjusted for vue3
 // export * from './functions/index.js'
-// export * from './directives/index.js'
+export * from './directives/index.js'
 // export * from './mixins/index.js'
 
 export { NextcloudVuePlugin } from './plugin.ts'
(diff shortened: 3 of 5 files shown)


docs/directives.md
```diff
@@ -2,16 +2,25 @@
...
-Vue.directive('tooltip', Tooltip)
 ```
 
 The tooltip directive is based on v-tooltip. For a full feature list see the projects [README](https://github.com/Akryum/v-tooltip/blob/master/README.md#directive)
 
 In the template, use the `v-tooltip` directive:
 
 ```vue
-<a v-tooltip="'You have new messages.'">Hover me</a>
+    <template>
```
raimund-schluessler, line 12: I couldn't figure out how to globally import the `tooltip` directive for the styleguide, so I individually imported it for the examples.

ShGKme: Maybe that's what may help?
https://vue-styleguidist.github.io/Configuration.html#enhancepreviewapp

raimund-schluessler: Yes, thanks a lot. That's what I was looking for.

styleguide/preview.js
```diff
@@ -0,0 +1,8 @@
+import { Tooltip } from '../src/directives/index.js'
+
+// The export here MUST be default or module.export
+// this is what is imported by the styleguide
+export default (app) => {
+  app.directive('tooltip', Tooltip)
+}
+
```
ShGKme, line 8: Should we also add OC/OCA here to the `globalProperties` lioke it was in the `Vue.prototype` before?

ShGKme (approved):

styleguide/preview.js
raimund-schluessler: Maybe, but I don't know. I think we try to wrap its usage in libraries. But we can of course still add it here once we encounter a usage.

ShGKme: I meant this change:

https://github.com/nextcloud-libraries/nextcloud-vue/pull/4498/files#diff-a8c83951daf4c1d563181c22713c5162a6feae9effea0034e67bf9111c0b5b33

raimund-schluessler: Yes, I know. But OC and OCA is currently not used (directly) in the vue 3 version of nc/vue. The only place where we (will) use it is here:
https://github.com/nextcloud-libraries/nextcloud-vue/blob/3a9b9cd5f82aa571c0a94a9489f578f04d91faa8/src/components/NcAppSidebar/NcAppSidebarTabs.vue#L236

Maybe we cross that bridge (and add `OC` to vue) once we get there.

juliushaertl (approved):

Issue #841, opened 2020-01-29
[FR]: Typescript support


janis91:
It would be very beneficial to add TypeScript support through a declaration file. I am currently working on porting my app to Vue.js based on TypeScript. I personally feel lot more comfortable with TS than plain JS and I think it might support other app devs as well, when they think about using this shared component lib.

skjnldsv:
Hi @janis91. Thanks for raising this question!
Typescript rise the level entry rather quite high. We talked about it before and decided to not use it for our apps. Nonetheless, we do use it for our libraries `@nextcloud/xxxxxx`. So this questions is an interesting one to ask ourselves for the components library.

We don't get much contributors here and some components are already quite complicated.
How much work do you estimate to implement this here? Because we're all running on limited ressources and it's something to consider in the balance :)

cc @ChristophWurst

janis91:
Hi @skjnldsv.
I think there are a few things to consider:
Even though you don't have to use TS in your app, you can benefit from typings because of IDE autocompletion and checks based on them. So especially for libraries there is a benefit.

I'm not very familiar with authoring a component library or libraries that provide typings in general, so I can't say, if this is a big deal (considered the limited time). I think there are some alternatives:

1. Refactor / Rewrite the lib in TS to get types strongly connected to the code (lot of work initially; maybe some more work when developing in the future; potential contributors could be daunted, because they don't know TS, but on the other hand could also benefit from the typings, when contributing, because they'd get a lot more feedback directly, if everything fits code-wise)
2. Provide the declaration files loosely within the repo. (not much work initially; should be considered whenever something is changed API-wise; contributors might ignore this, but maybe no TS knowledge is needed to contribute)
3. Add typings to DefinitelyTyped. Much more loose from the repo. (like 2, but worse?)

I mean, as I am developing my app in TS anyway, I will have to provide some kind of declaration anyway within my project. Still, I won't type everything - just what I use / need.
I think it should be considered, moreover because Vue itself moves in the direction of TypeScript with version 3. This of course doesn't mean you'd have to use it in the future, but the community will be much more aware of it in the future, I guess.

Just my two cents.

skjnldsv:
> I think it should be considered, moreover because Vue itself moves in the direction of TypeScript with version 3.

Some reading about this? :)

janis91:
> Some reading about this? :)

https://medium.com/the-vue-point/plans-for-the-next-iteration-of-vue-js-777ffea6fabf

[...]

PR #3565, opened 2022-12-12
Migrate to vue 3 + vite / vitest / vitepress for building

susnux:
Based on https://github.com/nextcloud/nextcloud-vue/pull/2637 and https://github.com/nextcloud/nextcloud-vue/pull/3529

The migration to vite does only make sense using vue 3 (so the automatic docs generation works).

---

* [x] Migrate to [vite](https://vitejs.dev/) for building (so dependency on webpack anymore)
* [x] Migrate to [vitest](https://vitest.dev/) for unit tests (instead of jest)
* [x] Migrate to [vue-docgen](https://vue-styleguidist.github.io/docs/docgen-cli.html) + [vitepress](https://vitepress.vuejs.org/) for documentation
* [x] Migrate cypress to use vite (and updated cypress) 
* [x] :construction: vue 3 migration (under development by @raimund-schluessler )

Issue #174, opened 2018-12-22
Storybook for Vue


nicolad:
It will be nice to have integrated storybook.
More details here:
https://storybook.js.org/basics/guide-vue/
Here are some examples:
https://storybook.js.org/examples/

ChristophWurst:
@skjnldsv I think we discussed this already on IRC :)

skjnldsv:
@nicolad any reason you closed? :)

nicolad:
I didn't see any activity and supposed it's not relevant. Just reopened it.

splitt3r:
nextcloud.design wouldbe very cool :grin: justlike gitlab.design or atlassian.design :heart_eyes:

PR #4525, opened 2023-09-14
[next] Allow writing components using Typescript and provide type definitions for `NcButton`

susnux:
### ☑️ Resolves

* Start of #841 

Add the Typescript infrastructure to write components using Typescript and migrate a first example component (`NcButton`) to Typescript. The type definitions are generated and published.

### Todo
As discussed previously the legacy naming of our entry points (`components` vs `Components`) will lead to a crowded `dist` directory, as Typescript uses the source naming, but our entry points use the upper case naming.
Maybe we should unify those names, meaning either rename the source folder or the entry points (?)

### 🏁 Checklist


src/components/NcButton/NcButton.vue
```diff
@@ -554,7 +565,7 @@ export default {
 
 		// If we have a router-link, we wrap the button in it
 		if (this.to) {
-			return h(resolveComponent('router-link'), {
+			return h(RouterLink, {
```
ShGKme, line 568: **issue (blocking):** `vue-router` is not a dependency or peerDependency of `@nextcloud/vue`

package.json
```diff
@@ -110,6 +111,7 @@
     "@vue/vue3-jest": "^29.2.5",
     "babel-jest": "^29.0.1",
     "babel-loader-exclude-node-modules-except": "^1.2.1",
+    "babel-preset-typescript-vue3": "^2.0.17",
```
ShGKme, line 114: This preset seems to not be very widely-known and maintained with almost no installations, 5 stars, and rare updates...

susnux:
BTW currently the vue3 eslint config is a bit outdated, so there is no typescript config as for the vue2 config. So eslint fails currently.

src/components/NcButton/NcButton.vue
```diff
@@ -321,24 +321,32 @@ export default {
 ```
 </docs>
 
-<script>
+<script lang="ts">
 import isSlotPopulated from '../../utils/isSlotPopulated.js'
```
raimund-schluessler, line 325:
```suggestion
import isSlotPopulated from '../../utils/isSlotPopulated'
```

src/utils/isSlotPopulated.ts
```diff
@@ -20,18 +20,22 @@
  *
  */
 
-import { Fragment, Comment, Text } from 'vue'
+import { Fragment, Comment, Text, type VNode, type VNodeNormalizedChildren } from 'vue'
```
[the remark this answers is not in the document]
susnux: It is possible, but most of the time I write a extra types import to make it clearer what are types and what logic.

src/components/NcButton/NcButton.vue
raimund-schluessler: Well, I don't know, I just saw the error message from jest
```
 Cannot find module '../../utils/isSlotPopulated.js' from 'src/components/NcButton/NcButton.vue'
```

susnux: Good point, that needs to be fixed

src/components/NcButton/NcButton.vue
susnux: Done

package.json
susnux: Well the alternative would be copy-paste the rules from the preset, we only need this for vue styleguideist which still uses webpack+babel

src/components/NcButton/NcButton.vue
susnux: Solved :heavy_check_mark:

package.json
raimund-schluessler: Or we get rid of vue-styleguidist and webpack first and migrate to vitepress? But I guess we can also keep it for now and migrate later on.

package.json
```diff
@@ -109,21 +109,22 @@
...
     "@mdi/js": "^7.3.67",
     "@mdi/svg": "^7.0.96",
     "@nextcloud/babel-config": "^1.0.0",
     "@nextcloud/browserslist-config": "^3.0.0",
-    "@nextcloud/eslint-config": "github:nextcloud-libraries/eslint-config#fe67f6b84e3c13eb0ea2478f44939abc2fe7c99c",
+    "@nextcloud/eslint-config": "github:nextcloud-libraries/eslint-config#0d785eb6d1d62b91be1452759a615a476c87db4c",
     "@nextcloud/stylelint-config": "^2.3.1",
     "@nextcloud/vite-config": "github:nextcloud-libraries/nextcloud-vite-config#36ae8cf759a30e1b020d90a6c3585fa7b7734dc7",
     "@nextcloud/webpack-vue-config": "github:nextcloud/webpack-vue-config#1f4277c6ab14be63810f425ef6b762d8ca59697b",
+    "@vue/compiler-sfc": "^3.3.7",
```
raimund-schluessler, line 123: Do we really need to add this manually? The docs say, one doesn't:
https://vuejs.org/guide/scaling-up/tooling.html#vue-compiler-sfc

Installing it manually might break dependabot updates.

src/components/NcActionCheckbox/NcActionCheckbox.vue
```diff
@@ -118,7 +118,7 @@ export default {
 	},
 
 	methods: {
-		checkInput(event) {
+		checkInput(/* event */) {
```
raimund-schluessler, line 121: Is this necessary for typescript? Just asking out of curiosity.

src/components/NcSelect/NcSelect.vue
```diff
@@ -983,15 +983,12 @@ export default {
...
-				userSelect,
-				// Props to forward
-				...initialPropsToForward
-			} = this.$props
+			const initialPropsToForward = { ...this.$props }
+			// Props handled by this component
+			delete initialPropsToForward.inputClass
+			delete initialPropsToForward.noWrap
+			delete initialPropsToForward.placement
+			delete initialPropsToForward.userSelect
```
raimund-schluessler, line 991: This seems unrelated, no?

src/utils/isSlotPopulated.ts
```diff
@@ -20,18 +20,23 @@
...
- * @param {Array} vnodes The array of vnodes to check
+ * @param vnodes The array of vnodes to check
  */
-const isSlotPopulated = function(vnodes) {
-	return !!vnodes?.some(node => {
-		if (node.type === Comment) return false
-		if (node.type === Fragment && !isSlotPopulated(node.children)) return false
-		if (node.type === Text && !node.children.trim()) return false
+const isSlotPopulated = function(vnodes?: VNode[] | VNodeNormalizedChildren) {
+	return Array.isArray(vnodes) && vnodes.some(node => {
```
raimund-schluessler, line 32: Why do you prefer
```ts
Array.isArray(vnodes) && vnodes.some(...)
```
over
```js
!!vnodes?.some(...)
```
? Does it behave different in some cases?

package.json
susnux: Or even migrate to #174

src/components/NcActionCheckbox/NcActionCheckbox.vue
susnux: I can also remove it, but it is unused and otherwise ESLint will complain about unused parameter

src/components/NcSelect/NcSelect.vue
susnux: No this fixes ESLint which will complain about unused variables, because you create const variables `inputClass`, `noWrap` etc without using them.
So instead of creating unused variables, just delete the props we use.

package.json
raimund-schluessler: > Or even migrate to #174

Why not, but that looks like a lot more work than vitepress, no?

src/utils/isSlotPopulated.ts
susnux: Because vnodes is not always an array, you call the function like this `isSlotPopulated(node.children)` so the the parameter type for that call is `VNodeNormalizedChildren` which can be `null, string, VNode[], RawSlots`.

So when you want to call `some` on that parameter it will cause a Typescript error:

> Property 'some' does not exist on type 'string | VNode<RendererNode, RendererElement, { [key: string]: any; }>[] | VNodeArrayChildren | RawSlots'.
> Property 'some' does not exist on type 'string'.ts(2339)

So Array.isArray simply asserts that the parameter is an Array, so the following code will assume that vnodes can only be of type `VNode[]`.

src/components/NcActionCheckbox/NcActionCheckbox.vue
susnux: What would you prefer? Let it there as a comment or remove the unused argument?

raimund-schluessler: I don't mind. You can leave it as a comment.

src/utils/isSlotPopulated.ts
raimund-schluessler: Indeed, that's what I had in mind. But checking that's an array first seems like the better solution here. 👍

package.json
susnux: Yes that is what @ShGKme and I fear :sweat_smile: 
But it has some great benefits, as always the problem is missing time.

src/components/NcSelect/NcSelect.vue
susnux: (BTW I personally even think the explicit delete is more readable meaning it makes it clear what is expected to happen)

package.json
ShGKme: Does Vitepress mean to write all the documentation manually? Event props list...

raimund-schluessler: > Does Vitepress mean to write all the documentation manually? Even props list...

No, this will be extracted. @susnux started a first try in #3565, see https://deploy-preview-3565--nextcloud-vue-components.netlify.app
It already kind of works, but needs some adjustements and a fix for the components preview.

susnux: Yes instead of using vue-styleguideist which uses webpack / vuepress we can use the vue-styleguidist cli tools to extract the documentation as markdown which we then serve using vitepress, see the example:

https://github.com/vue-styleguidist/vue-styleguidist/tree/dev/examples/docgen-vite

susnux: I dropped that preset and added a simple config.

susnux:
Rebased and solved some issues after all components were migrated to Vue 3. So I would say this is ready for reviews :)

.eslintrc.js
```diff
@@ -21,4 +21,17 @@ module.exports = {
 			],
 		},
 	},
+	rules: {
+		'jsdoc/check-tag-names': [
+			'warn',
+			{
+				definedTags: [
+					'values', // required for vue-styleguideist
```
ShGKme, line 29: It is not required for styleguideist, it supports `@type` with string alternative `@type {'a'|'b'|'c'}`

ShGKme: And we don't have `@values` anymore

skjnldsv (approved):

.eslintrc.js
```diff
@@ -21,4 +21,17 @@ module.exports = {
...
+					'notExported',
+					'packageDocumentation',
+					'jest-environment',
```
raimund-schluessler, line 32: Jest is not used anymore. Can we get rid of this then?

raimund-schluessler (approved):

.eslintrc.js
susnux: Fixed

.eslintrc.js
susnux: Yes, fixed :)

ShGKme (approved):

PR #4542, opened 2023-09-18
chore(vue3): Migrate NcEmptyContent

raimund-schluessler:
This migrates `NcEmptyContent` to vue 3.

### DIFF
--- a/src/components/NcEmptyContent/NcEmptyContent.vue
+++ b/src/components/NcEmptyContent/NcEmptyContent.vue
@@ -146,11 +146,11 @@ export default {
 		</div>
 		<!-- @slot Optional name if not set as property, shall be enclosed by a header element -->
 		<slot name="name">
-			<h2 v-if="hasName" class="empty-content__name">
+			<h2 v-if="name !== ''" class="empty-content__name">
 				{{ name }}
 			</h2>
 		</slot>
-		<p v-if="hasDescription">
+		<p v-if="description !== '' || $slots.description">
 			<!-- @slot Optional formatted description rendered inside a paragraph -->
 			<slot name="description">
 				{{ description }}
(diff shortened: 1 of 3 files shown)


src/components/NcEmptyContent/NcEmptyContent.vue
```diff
@@ -187,7 +187,7 @@ export default {
 		 * Check if a description is given as either property or slot
 		 */
 		hasDescription() {
-			return this.description !== '' || this.$slots.description?.[0]
+			return this.description !== '' || this.$slots.description?.()?.[0]
```
skjnldsv, line 190: why would it be a function?

susnux: In vue3 `$slots` is a record of `slotname -> function`, so for scoped slots you would pass in the arguments of the slot.
See https://vuejs.org/api/component-instance.html#slots

ShGKme: If we need to check the content here, not if the slot just was passed, should we use `isSlotPopulated` here? Or just check `!!this.$slots.description` otherwise to not render description in computed?

raimund-schluessler: > Or just check `!!this.$slots.description` otherwise to not render description in computed?

That might be the best option here.

raimund-schluessler: Adjusted.

src/components/NcEmptyContent/NcEmptyContent.vue
```diff
@@ -228,8 +216,8 @@ export default {
 		background-size: 64px;
 
 		:deep(svg) {
-			width: 64px;
-			height: 64px;
+			width: 64px !important;
+			height: 64px !important;
```
raimund-schluessler, line 220: This comes from #4584.

ShGKme (approved):

mejo- (approved):

raimund-schluessler:
Resolved conflict.

PR #4543, opened 2023-09-18
chore(vue3): Nc*Field rename default to icon slot

raimund-schluessler:
@ShGKme What do you think of this approach:

We rename the `default` slot of `NcInputField` to `icon` and then simply use `!!$slots.icon` as you proposed. This way, one anyway has to wrap the content for the `icon` slot in a `template` tag and it is easier to explain that one has to use `v-if` on the `template` when toggling the existence of an icon.
This then makes everything reactive without the need for an overkill solution, no?

And I think renaming to `icon` anyway makes sense, since it makes it clearer that it's, well, for an icon.


skjnldsv (approved):

ShGKme (approved):
Thanks, seems to be a good solution to me

PR #4557, opened 2023-09-22
fix: Replace sanitize-svg with dompurify

raimund-schluessler:
### ☑️ Resolves

* This fixes  the display of `NcIconSvgWrapper` on iOS Safari. Safari wouldn't render the SVG if it has no `width` and `height` set.
* Moves from `skjnldsv/sanitize-svg` to `dompurify`
* Uses a `computed` instead of `beforeMount` and a method. @Pytal Any reason why this was a method? If yes, we can revert the last commit.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

Related to #4312 which introduced the problem, I think.

PS: If you wonder, why the dependency to `skjnldsv/sanitize-svg` is still in `package-lock.json`, this is because we depend on `nextcloud\dialogs@4.2.0` which has a dependency to `nextcloud/vue@7.12.4`.


susnux (approved):
Also better performance wise

skjnldsv (approved):

PR #4558, opened 2023-09-23
chore(next): Merge `master` into `next`

raimund-schluessler:
### ☑️ Resolves

* Manually merge `master` into `next` so we don't lose changes done to master.

I had to pin `nextcloud/dialogs` to `4.1.x`, because `4.2.x` depends on `nextcloud/vue` which is not yet available as a vue 3 version, obviously.


susnux:
> I had to pin `nextcloud/dialogs` to `4.1.x`, because `4.2.x` depends on `nextcloud/vue` which is not yet available as a vue 3 version, obviously.

I created https://github.com/nextcloud-libraries/nextcloud-vue/pull/4559 to fix this dependency circle.

skjnldsv (approved):

raimund-schluessler:
Pulled in more changes from `master`. I will adjust the `DateTimePickerNative` naming once this is merged here.

susnux (approved):

PR #4559, opened 2023-09-23
fix(NcSettingsSelectGroup): Prevent circular dependency with `@nextcloud/dialogs`

susnux:
### ☑️ Resolves

* Fix https://github.com/nextcloud-libraries/nextcloud-dialogs/issues/885

If there is a loading error, then we can simply add a error info text below the component.
This was the only place where we used the dialogs library, so by changing the error handling here we can drop the dependency and prevent a circular dependency issue.

### 🖼️ Screenshots
image

### 🏁 Checklist


raimund-schluessler:
This might indeed be the best solution.

raimund-schluessler (approved):

skjnldsv (approved):

PR #4560, opened 2023-09-23
chore!: Rename `Datetime` to `DateTime` in component names

susnux:
### ☑️ Resolves

Fix inconsistent naming
* `NcDatetime` is now called `NcDateTime` (not yet released to ok)
* `NcDatetimePicker` is now called `NcDateTimePicker` (breaking, but we make it consistent with DateTimeNative)

### 🏁 Checklist


skjnldsv (approved):

st3iny (approved):
That was quick :smile:

raimund-schluessler (changes requested):
That is inconsistent with the other component names.

susnux:
But what is a *Datetime*? It is a *Date* and *Time* picker so I would say *DateTimePicker* makes more sense.
At least it is more consistent with all other names like *AppNavigation* (and not *Appnavigation*) etc.

raimund-schluessler:
> But what is a _Datetime_? It is a _Date_ and _Time_ picker so I would say _DateTimePicker_ makes more sense. At least it is more consistent with all other names like _AppNavigation_ (and not _Appnavigation_) etc.

I considered datetime to be one word, but that is probably a German bias. I am fine with whatever consistent solution we use. We just have to revert and adjust it for `next` as well then.

skjnldsv:
DateTime is used in javascript
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat

PR #4562, opened 2023-09-24
chore(vue3): Migrate NcColorPicker

raimund-schluessler:


src/components/NcColorPicker/NcColorPicker.vue
```diff
@@ -159,32 +154,34 @@ export default {
...
+						@update:modelValue="pickColor" />
 				</div>
-				<Chrome v-if="advanced"
-					v-model="currentColor"
-					class="color-picker__advanced"
-					:disable-alpha="true"
-					:disable-fields="!advancedFields"
-					@input="pickColor" />
-			</transition>
+			</Transition>
```
ShGKme, line 184: This additional `<div>` wrapper breaks the transition

ShGKme (changes requested):

src/components/NcColorPicker/NcColorPicker.vue
raimund-schluessler: Indeed. I added it because I got a `VueCompilerError: <Transition> expects exactly one child element or component.`, but that was triggered by using `v-if="!advanced"` and `v-if="advanced"` instead of `v-if/v-else`. I fixed it now and the transition works again.

ShGKme (approved):

susnux (approved):

PR #4563, opened 2023-09-24
chore(vue3): Migrate NcEmojiPicker

raimund-schluessler:
This migrates `NcEmojiPicker` to vue 3 and pulls in two commits from `master`.

### DIFF
--- a/src/components/NcEmojiPicker/NcEmojiPicker.vue
+++ b/src/components/NcEmojiPicker/NcEmojiPicker.vue
@@ -124,10 +124,8 @@ This component allows the user to pick an emoji.
 </docs>
 
 <template>
-	<NcPopover :shown.sync="open"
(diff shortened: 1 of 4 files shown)


src/components/NcEmojiPicker/NcEmojiPicker.vue
```diff
@@ -393,6 +391,7 @@ export default {
 	background-color: var(--color-main-background) !important;
 	border: 0;
 	color: var(--color-main-text) !important;
+	display: flex !important;
```
raimund-schluessler, line 394: The component keeps overwriting `display: flex;` with `display: inline-block`, which breaks the scrolling. So I force it here for now.

ShGKme (approved):

mejo- (approved):

PR #4576, opened 2023-09-26
Set double outline to active action menu

JuliaKirschenheuter:
### ☑️ Resolves

* Part of https://github.com/nextcloud/server/issues/36966

See general requirements here: https://github.com/nextcloud-libraries/nextcloud-vue/issues/4193

Draft PR because in a moment i can't link NcVue and Server.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot from 2023-09-26 13-26-09 | Screenshot from 2023-09-26 13-25-20

### 🏁 Checklist

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -1264,6 +1264,8 @@ export default {
 
 	&.action-item--open .action-item__menutoggle {
 		background-color: var(--open-background-color);
+		outline: 2px solid var(--color-main-text);
+		box-shadow: 0 0 0 4px var(--color-main-background);
 	}
 }
 </style>
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -1265,7 +1265,9 @@ export default {
 	&.action-item--open .action-item__menutoggle {
 		background-color: var(--open-background-color);
 		outline: 2px solid var(--color-main-text);
-		box-shadow: 0 0 0 4px var(--color-main-background);
+		outline-offset: -4px;
+		border: 2px solid var(--color-main-background);
+		height: 44px;
 	}
 }
 </style>


raimund-schluessler:
This creates a lot of issues as it effectively increases the size of the button from 44x44px to 52x52px when active. The button then overlaps, or gets overlapped by other elements. If we really do it like this, I would consider it a breaking change that needs a lot of adjustments thoughout server and apps to not look weird.

Do we really have to do that by default for every active action menu? Couldn't this be only active when navigating with the keyboard of when some accessibility setting is active?

susnux:
> Do we really have to do that by default 

I think yes, for accessibility.

But maybe we can include it in the button?

JuliaKirschenheuter:
Hi @raimund-schluessler,

Thank you for your comment, you are right. But for a11y purposes we need to do something, there is no way around ;(

@susnux @raimund-schluessler @Pytal @nimishavijay would you approve such alternative solution?:

Screenshot from 2023-09-27 16-11-32

JuliaKirschenheuter:
> @susnux @raimund-schluessler @Pytal @nimishavijay would you approve such alternative solution?:
> 
> Screenshot from 2023-09-27 16-11-32

No, sorry, that won't work at already selected element ;( we have to move forwards with some kind of outline.

nimishavijay:
> But maybe we can include it in the button?

@JuliaKirschenheuter this seems like a good solution. Would it be possible to have 2 borders but inside the button instead of outside? the `outline` property in CSS results in something like this, what do you think?

image

raimund-schluessler:
These rules:
```css
  outline: 2px solid red;
  outline-offset: -4px;
  border: 2px solid green;
  height: 44px;
```
result in this:
grafik

(Ignore the colors, I just set them to make it more visible).

This way, the button keeps its size of `44px` width and height. I just don't know whether this is considered accessbile.

JuliaKirschenheuter:
Thank you @nimishavijay and @raimund-schluessler!

@raimund-schluessler what is not good in your eyes?

raimund-schluessler:
> @raimund-schluessler what is not good in your eyes?

Nothing in particular, I am just not an expert on accessibility 😉

raimund-schluessler (changes requested):
We also need to adjust these lines:
https://github.com/nextcloud-libraries/nextcloud-vue/blob/2e6b0f0588a9fdd77a89ee86b4e867c2c96dce70/src/components/NcButton/NcButton.vue#L707C19-L715

Otherwise, the toggle button gets a weird triple border effect when clicking.

And the transition for the `border-color` also needs to go:

https://github.com/nextcloud-libraries/nextcloud-vue/blob/2e6b0f0588a9fdd77a89ee86b4e867c2c96dce70/src/components/NcButton/NcButton.vue#L610

raimund-schluessler:
And would it be ok to switch the color of outline and border? Currently, with the darker color for the outline, it looks like the button was shrinking when clicked:

| Currently | Proposed changes |
|-|-|
| https://github.com/nextcloud-libraries/nextcloud-vue/assets/2496460/6d17f63e-a151-4260-897b-3281c9dbb161 | https://github.com/nextcloud-libraries/nextcloud-vue/assets/2496460/8b743224-6cdd-454d-a459-5d8de043df3d |

susnux:
> [...] Would it be possible to have 2 borders but inside the button instead of outside? the `outline` property in CSS results in something like this, what do you think?

@nimishavijay thank you, that was exactly what I meant :)
I really like the mockup and @raimund-schluessler proposal :+1: 

---

Slightly other topic, but we should generally agree on one box size, because it is already weird and cumbersome to use multiple components in a form.
Like buttons have a height of 44px (which is for a11y the recommended size), then we have input elements which use 38px (or 44px with label on the border), and then we have the NcSelect with a height of 48px.

If you use two or more of those together it looks not consistent, maybe we need a design review to agree on one box height (especially if you consider the border outline as the visible element boundaries)? Might be related to #4582

nimishavijay:
> And would it be ok to switch the color of outline and border? Currently, with the darker color for the outline, it looks like the button was shrinking when clicked

Ah, good point, but if we switch the colors it would be fine in light mode but look like it was shrinking in dark mode. The current colors were chosen because of https://github.com/nextcloud-libraries/nextcloud-vue/issues/4193#issuecomment-1614040777 , and since the button will look like it's shrinking in one of the 2 modes anyway, we can keep it like how it is right now, what do you think? :)

raimund-schluessler:
The colors get inverted in dark mode, so it should look similar in both modes. But that needs to be tested.

JuliaKirschenheuter:
> The current colors were chosen because of [#4193 (comment)](https://github.com/nextcloud-libraries/nextcloud-vue/issues/4193#issuecomment-1614040777) 

@nimishavijay @raimund-schluessler i would like to move with color combination we already had in other places to be identical with other components. Would it be ok for you?

JuliaKirschenheuter:
> Slightly other topic, but we should generally agree on one box size, because it is already weird and cumbersome to use multiple components in a form.

That is very right point! I would say we need this improvement! But better in other ticket / and after a11y certification.

raimund-schluessler:
> > The current colors were chosen because of [#4193 (comment)](https://github.com/nextcloud-libraries/nextcloud-vue/issues/4193#issuecomment-1614040777)
> 
> @nimishavijay @raimund-schluessler i would like to move with color combination we already had in other places to be identical with other components. Would it be ok for you?

Fine with me. But the box-shadow really needs to be fixed, it looks very odd with a triple border when clicking.

marcoambrosini (commented):
@JuliaKirschenheuter @AndyScherzinger I think that there's a misunderstanding here. This button has no active state or pressed state. It simply opens a menu, much like a sidebar toggle opens a sidebar. It does not require a visual change after a click.

JuliaKirschenheuter:
> @JuliaKirschenheuter @AndyScherzinger I think that there's a misunderstanding here. This button has no active state or pressed state. It simply opens a menu, much like a sidebar toggle opens a sidebar. It does not require a visual change after a click.

Hi @marcoambrosini i'm pretty sure that only a background indicator which we have right now is not enough. All users have to be able to see that menu is focused (=== and opened) not only via light grey background.

@michaelnissenbaum what would you say? Thank you!

marcoambrosini:
@JuliaKirschenheuter the background that we have right now is not strictly necessary, what matters is the contrast between the icon and it's background.

> All users have to be able to see that menu is focused (=== and opened)

The focus state of the button is already taken care of by the double outline of the NcButton component when tabbing through elements. 

Once the button is clicked, the focus should just move to the first element of the menu and all users will be able to see that the menu is opened, because they will see a big menu popping up upon pressing the button :) 
Visually treating the 3-dot button as if it's focused is actually misleading more than helpful here, as we would show two elements with a focused state (the menu button and the first element of the menu). 
As said above this is not a stateful button that has, so there's no need to change anything about it on click as far as I can see.

JuliaKirschenheuter:
@marcoambrosini Thank you for clarification! Agree with you!
But before i will close it i would like to have confirmation from Michael.

michaelnissenbaum:
@JuliaKirschenheuter  To confirm the proposed implementation, it would be great if I could have access to the implementation or a video showing how it will work. Then, I can provide my feedback.

JuliaKirschenheuter:
> @JuliaKirschenheuter To confirm the proposed implementation, it would be great if I could have access to the implementation or a video showing how it will work. Then, I can provide my feedback.

Dear @michaelnissenbaum 

I looks like this now: 

Peek 2023-10-04 14-41

michaelnissenbaum:
@JuliaKirschenheuter. Looks good to me.

JuliaKirschenheuter:
Closing PR because it is already good enough: https://github.com/nextcloud-libraries/nextcloud-vue/pull/4576#issuecomment-1746823155

PR #4578, opened 2023-09-26
fix(NcInputField): Adjust styling of the internal label

susnux:
### ☑️ Resolves

From today's design review call:
* Add note that the internal label is only recommended when the background matches
* Make border of the internal label rounded (needed to adjust the padding as well)
* Perfect pixel align label, placeholder and content

### 🖼️ Screenshots
(added background color to make the border more visible, e.g. like on the login)

🏚️ Before | 🏡 After
---|---
Screenshot 2023-09-26 at 15-55-49 Nextcloud Vue Style Guide|Screenshot 2023-09-26 at 15-55-20 Nextcloud Vue Style Guide

### 🏁 Checklist

### DIFF
--- a/src/components/NcInputField/NcInputField.vue
+++ b/src/components/NcInputField/NcInputField.vue
@@ -155,6 +155,8 @@ export default {
 		 * The input label, always provide one for accessibility purposes.
 		 * This will also be used as a placeholder unless the placeholder
 		 * prop is populated with a different string.
+		 *
+		 * Note: If the background color is not `--color-main-background` consider using an external label instead (see `labelOutside`).
 		 */
 		label: {
 			type: String,
@@ -345,7 +347,7 @@ export default {
 
 	&__input {
 		margin: 0;
-		padding-inline: 10px 6px; // align with label 8px margin label + 4px padding label - 2px border input
+		padding-inline: 12px 6px; // align with label 8px margin label + 6px padding label - 2px border input
 		height: 38px !important;
 		width: 100%;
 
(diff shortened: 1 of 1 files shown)


nimishavijay (approved):
Nice! Much better already!

ShGKme (approved):

jancborchardt (dismissed):
Much better! But shouldn’t we use variables in all those places, based on e.g. `var(--default-grid-baseline)`?

src/components/NcInputField/NcInputField.vue
```diff
@@ -465,16 +467,19 @@ export default {
 
 	&__input:focus + &__label,
 	&__input:not(:placeholder-shown) + &__label {
-		inset-block-start: -6px;
+		inset-block-start: -8px;
 		font-size: 13px; // minimum allowed font size for accessibility
+		font-weight: 500;
+		border-radius: 9px 9px 0 0; // height / 2 for pill style
```
marcoambrosini, line 473:
```suggestion
		border-radius: var(--default-grid-baseline) var(--default-grid-baseline) 0 0;
```

marcoambrosini (commented):
I'd go for 4px border-radius, not pill

src/components/NcInputField/NcInputField.vue
susnux: I am fine with both, pill style was the conclusion after the last design review call.

jancborchardt (approved):

PR #4587, opened 2023-09-28
chore(vue3): Migrate `NcSelect*` and related components

raimund-schluessler:
This migrates `NcSelect` and related components.

The components behave slightly different to the version in `master` currently, because I had to fall back to `vue-select` as our forked `nextcloud/vue-select` version lacks vue 3 support. Once this is done, we have to adjust the dependency. However, I would like to ignore this for now, as I need `NcSelect` to migrate the other components of `nextcloud/vue`.


src/mixins/index.js
```diff
@@ -22,6 +22,6 @@
 
 export { default as clickOutsideOptions } from './clickOutsideOptions/index.js'
 export { default as isFullscreen } from './isFullscreen/index.js'
-export { default as isMobile } from './isMobile/index.js'
-export { default as richEditor } from './richEditor/index.js'
+// export { default as isMobile } from './isMobile/index.js'
+// export { default as richEditor } from './richEditor/index.js'
```
raimund-schluessler, line 26: These mixins are not yet migrated, and would break the build.

package.json
```diff
@@ -96,7 +96,8 @@
     "unist-builder": "^4.0.0",
     "unist-util-visit": "^5.0.0",
     "vue": "^3.3.4",
-    "vue-material-design-icons": "^5.1.2"
+    "vue-material-design-icons": "^5.1.2",
+    "vue-select": "^4.0.0-beta.6"
```
raimund-schluessler, line 100: I fell back to the upstream library here, because nextcloud/vue-select does not have a vue3 version yet. Once we migrated this, we have to change the dependency here.

raimund-schluessler: See https://github.com/nextcloud-deps/vue-select/issues/23

src/components/NcListItemIcon/NcListItemIcon.vue
```diff
@@ -127,15 +127,14 @@ It might be used for list rendering or within the multiselect for example
...
-		<NcAvatar v-bind="$attrs"
+		:style="cssVars">
+		<!-- <NcAvatar v-bind="$attrs"
 			:disable-menu="true"
 			:disable-tooltip="true"
 			:display-name="displayName || name"
 			:is-no-user="isNoUser"
 			:size="avatarSize"
-			class="option__avatar" />
+			class="option__avatar" /> -->
```
raimund-schluessler, line 137: `NcAvatar` relies on `NcActions`, which relies on `NcSelect`. So I commented it out here, to not have a giant PR with actions, datepicker and select together. I will bring it back once `NcActions` are migrated.

src/components/NcSelect/NcSelect.vue
```diff
@@ -972,6 +971,7 @@ export default {
 				calculatePosition: this.localCalculatePosition,
 				filterBy: this.localFilterBy,
 				label: this.localLabel,
+				...this.$attrs,
```
raimund-schluessler, line 974: @ShGKme I don't know why, but without this it does not work.

ShGKme: Because of this undocumented behavior...

https://github.com/vuejs/core/pull/2011#issuecomment-685246268

If a component has some prop (`foo`), then an update event (`update:foo`) is not inherited by root component node, although it is not declared in `emits` and is a part of `$attrs`.

So in this case, a handler in attribute `onUpdate:modelValue` is passed neither as attr, nor as a component event listener to `VueSelect`, because there is a `modelValue` prop. 😣

ShGKme: Anyway, we should not pass all the `this.$attrs` here, but implicitly define and re-emit `update:modelValue` event, because only the `update:modelValue` is a problem, not all possible attrs.

Otherwise, there is a duplicating of attrs which is a problem for event handlers.

```vue
<NcSelect @open="handler" />
```

`handler` is called twice.

ShGKme (changes requested):
There is a problem with duplicating events (see comments above)

src/components/NcSelect/NcSelect.vue
raimund-schluessler: Interesting, thanks for figuring this out. I will have a look how to circumvent this.

ShGKme: > Interesting, thanks for figuring this out. I will have a look how to circumvent this.

We only need to define `update:modelValue` event in `emits` and manually proxy it by `@update:modelValue="$emit('update:modelValue', $event)"`

raimund-schluessler: Adjusted it. Seems to work fine now.

Pytal (commented):
👍 as an iterative migration without 100% feature+functional parity, but for reference please put a TODO in NcSelect.vue that all `@nextcloud/vue-select` specific improvements need to be reimplemented on top of the upstream vue-select beta branch

raimund-schluessler:
> 👍 as an iterative migration without 100% feature+functional parity, but for reference please put a TODO in NcSelect.vue that all `@nextcloud/vue-select` specific improvements need to be reimplemented on top of the upstream vue-select beta branch

I added a TODO that we have to switch to `nextcloud/vue-select` as soon as we migrated it to vue3. However, I wonder what's easier; use our branch, and migrate to vue 3, or use the upstream vue 3 version and implement our changes.

raimund-schluessler:
> 👍 as an iterative migration without 100% feature+functional parity, but for reference please put a TODO in NcSelect.vue that all `@nextcloud/vue-select` specific improvements need to be reimplemented on top of the upstream vue-select beta branch

@Pytal I just migrated `nextcloud\vue-select` to vue 3, please see https://github.com/nextcloud-deps/vue-select/pull/24. Using `npm link`, the `nextcloud/vue` styleguide works just fine with the vue 3 version.
In order to use it without `npm link`, we need a release to npm. Could you take care of this, and maybe release a alpha/beta version, so we can go ahead here?

src/components/NcSelect/NcSelect.vue
```diff
@@ -481,8 +481,8 @@ export default {
 			'select--no-wrap': noWrap,
 		}"
 		v-bind="propsToForward"
-		v-on="$listeners"
-		@search="searchString => search = searchString">
+		@search="searchString => search = searchString"
+		@update:model-value="(event) => $emit('update:modelValue', event)">
```
ShGKme, line 485:
```suggestion
		@search="search = $event"
		@update:model-value="$emit('update:modelValue', $event)">
```

src/components/NcSettingsSelectGroup/NcSettingsSelectGroup.vue
```diff
@@ -227,7 +227,7 @@ export default {
 		update(updatedValue) {
 			const value = updatedValue.map((element) => element.id)
 			/** Emitted when the groups selection changes<br />**Payload:** `value` (`Array`) - *Ids of selected groups */
-			this.$emit('input', value)
+			this.$emit('update:model-value', value)
```
ShGKme, line 230:
```suggestion
			this.$emit('update:modelValue', value)
```

src/components/NcSelectTags/NcSelectTags.vue
```diff
@@ -321,7 +321,7 @@ export default {
 				...propsToForward
 			} = this.$props
 
-			return propsToForward
+			return { ...propsToForward, ...this.$attrs }
```
ShGKme, line 324: Same as the previous component, or there is another reason to pass `$attrs`?

```suggestion
			return propsToForward
```

raimund-schluessler: No particular reason. I just forgot to adjust it.

ShGKme (approved):
Thanks!

raimund-schluessler:
Anyone up for a second review so we can go ahead here? Many other migrations depend on this.

ShGKme:
@susnux 👉👈

raimund-schluessler:
Rebased and squashed.

susnux (approved):

PR #4590, opened 2023-09-30
Added testcases for NcRelatedResourcesPanel Component

NitinRamnani:
### ☑️ Resolves

Added testcases for NcRelatedResourcesPanel Component and improved code coverage

### 🖼️ Screenshots

image

### 🚧 Tasks

- [x] Added testcases for NcRelatedResourcesPanel Component

### 🏁 Checklist

### DIFF
--- a/tests/unit/components/NcRelatedResourcesPanel/NcRelatedResourcesPanel.spec.js
+++ b/tests/unit/components/NcRelatedResourcesPanel/NcRelatedResourcesPanel.spec.js
@@ -1,0 +1,63 @@
+/**
+ * @copyright Copyright (c) 2023 Ferdinand Thiessen <opensource@fthiessen.de>
+ *
+ * @author Ferdinand Thiessen <opensource@fthiessen.de>
+ *
+ * @license AGPL-3.0-or-later
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+import { shallowMount } from '@vue/test-utils'
+import NcRelatedResourcesPanel from '../../../../src/components/NcRelatedResourcesPanel/NcRelatedResourcesPanel.vue'
+
+describe('NcRelatedResourcesPanel', () => {
+
+	let wrapper
+
+	beforeEach(() => {
+		wrapper = shallowMount(NcRelatedResourcesPanel, {
+			data: () => ({
+				appEnabled: true,
+				headerTranslated: 'Related resources',
+				loading: false,
+				error: null,
+				resources: [{ itemId: 'test', icon: 'test', title: 'test', url: 'http://test.com' }],
+			}),
+		})
+	})
+
+	it('it mounts', async () => {
+		expect(wrapper.html()).toBeTruthy()
+	})
+
+	it('tests fetchRelatedResources', async () => {
+		jest.mock('axios', () => ({
+			get: jest.fn(() => Promise.resolve({
+				data: {
+					ocs: {
+						data: [{ itemId: 'test', icon: 'test', title: 'test', url: 'http://test.com' }],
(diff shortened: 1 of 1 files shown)


tests/unit/components/NcRelatedResourcesPanel/NcRelatedResourcesPanel.spec.js
```diff
@@ -0,0 +1,63 @@
+/**
+ * @copyright Copyright (c) 2023 Ferdinand Thiessen <opensource@fthiessen.de>
+ *
+ * @author Ferdinand Thiessen <opensource@fthiessen.de>
+ *
```
ShGKme, line 5: This is supposed to be your contacts, not Ferdinand's :)

ShGKme: See: https://github.com/nextcloud/server/blob/master/contribute/HowToApplyALicense.md

tests/unit/components/NcRelatedResourcesPanel/NcRelatedResourcesPanel.spec.js
```diff
@@ -0,0 +1,63 @@
...
+	beforeEach(() => {
+		wrapper = shallowMount(NcRelatedResourcesPanel, {
+			data: () => ({
+				appEnabled: true,
+				headerTranslated: 'Related resources',
+				loading: false,
+				error: null,
+				resources: [{ itemId: 'test', icon: 'test', title: 'test', url: 'http://test.com' }],
+			}),
+		})
```
ShGKme, line 39: A test should never assign, change, or test a component's data or other parts or a component's implementation.

The main purpose of a test is to check that the component works as it is expected to work for its users: end users and developers (other components and modules). To test, that when an end-user or other units have a component instance and interact with it, it reacts as expected.

When a component is used, its data is never set outside. It is not a part of a component's interface.

tests/unit/components/NcRelatedResourcesPanel/NcRelatedResourcesPanel.spec.js
```diff
@@ -0,0 +1,63 @@
...
+					},
+				},
+			})),
+		}))
+		const fetchRelatedResourcesSpy = jest.spyOn(wrapper.vm, 'fetchRelatedResources')
+		wrapper.vm.appEnabled = true
+		wrapper.vm.$nextTick()
+		await wrapper.vm.fetchRelatedResources()
+		expect(fetchRelatedResourcesSpy).toHaveBeenCalled()
+		expect(wrapper.vm.resources).toEqual([{ itemId: 'test', icon: 'test', title: 'test', url: 'http://test.com' }])
```
ShGKme, line 61: Same problem here. Changing the component's data or calling its methods is not the result of the component. A component is not used to have any data changed, or its method to be called. All it is the component's implementation.

tests/unit/components/NcRelatedResourcesPanel/NcRelatedResourcesPanel.spec.js
```diff
@@ -0,0 +1,63 @@
...
+				loading: false,
+				error: null,
+				resources: [{ itemId: 'test', icon: 'test', title: 'test', url: 'http://test.com' }],
+			}),
+		})
+	})
+
+	it('it mounts', async () => {
+		expect(wrapper.html()).toBeTruthy()
+	})
```
ShGKme, line 44: Such tests are usually not very useful, they don't check anything specific. If a component doesn't mount, no test will pass anyway.

ShGKme (changes requested):
Thank you for your contribution!

Unfortunately, these tests check only the component's internal implementation, but not its usage.

Internal implementation tests don't help to make a unit more robust. It is not what component users actually use, they don't depend on it.

At the same time, test of internal implementation makes it more difficult to support the component. Now, even with a small internal refactoring like renaming data property, or changing the method would require the test to be rewritten. 

Instead, a test should check how a component is used.
- On arrangement, an instance can be created with 
  - some props
  - slot contents
  - mocked external dependencies, and etc.
  It is the same how other units use the component and it is its public interface. 
- Then, if needed, there can be an act, interaction with a component from outside:
  - Users interacting via triggering DOM events
  - Updating props
  - Emitting events from children
  - Sometimes calling its public methods
  - Changes in external dependencies
- In the end, there must be asserts of components "results":
  - Rendering and changes in DOM
  - Emitting events
  - Updating props of children (when they are not a part of the unit)
  - Side effects
  
If it follows the rules, test may help to make a component more robust. It helps to guarantee that the component instance, created with some props under certain conditions after a specific interaction has an expected result. And that it is still true after refactoring, bugfixes or adding new features.

susnux:
Thank you for your contribution!
I will close this for now due to lack of activity. But please feel free to reopen the pull request if you decide to continue to work on this. For this please see the review comments.

PR #4593, opened 2023-10-02
fix(NcAvatar): Ensure the aria label includes the status if there is any visible status indicator

susnux:
### ☑️ Resolves

* Fix https://github.com/nextcloud/server/issues/40713

No visible changes, just enforce the aria label to include the user status if there is any visible indicator (like when used in `NcSelect`).
Also translate `dnd` because that abbreviation is quite confusion (I had to search for it, as I was thinking about a pen and paper game :wink: ).

### 🏁 Checklist


src/components/NcAvatar/NcAvatar.vue
```diff
@@ -370,11 +370,8 @@ export default {
 	},
 	computed: {
 		avatarAriaLabel() {
-			if (!this.hasMenu) {
-				return
-			}
```
Pytal, line 375: We'd need to keep this as non-interactive elements should not have an `aria-label`, added in https://github.com/nextcloud-libraries/nextcloud-vue/pull/3618

Pytal (changes requested):

src/components/NcAvatar/NcAvatar.vue
susnux: Reverted and added a comment about the purpose of that check :)

tests/unit/components/NcAvatar/NcAvatar.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+import { mount } from '@vue/test-utils'
+import NcAvatar from '../../../../src/components/NcAvatar/NcAvatar.vue'
+
+describe('NcAvatar.vue', () => {
+	'use strict'
```
ShGKme, line 27: Does 'use strict' change here anything in a module?

tests/unit/components/NcAvatar/NcAvatar.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+			propsData: {
+				user: 'janedoe',
+				displayName: 'J. Doe',
+				preloadedUserStatus: status,
+			},
+		})
+
+		expect(wrapper.find('.avatardiv__user-status').exists()).toBe(true)
+		expect(wrapper.element.hasAttribute('aria-label')).toBe(true)
+		expect(wrapper.element.getAttribute('aria-label')).toBe('Avatar of J. Doe, online')
```
ShGKme, line 46:
```suggestion
		expect(wrapper.find('.avatardiv__user-status').exists()).toBe(true)
		expect(wrapper.attributes('aria-label')).toBe('Avatar of J. Doe, online')
```

src/components/NcAvatar/NcAvatar.vue
```diff
@@ -121,8 +121,10 @@ export default {
 		:tabindex="hasMenu ? '0' : undefined"
 		:aria-label="avatarAriaLabel"
 		:role="hasMenu ? 'button' : undefined"
-		@click="toggleMenu"
-		@keydown.enter="toggleMenu">
+		v-on="hasMenu ? {
+			click: toggleMenu,
+			keydown: toggleMenu,
+		} : null">
```
ShGKme, line 127: IMO, we can move `hasMenu` check to the events handler `toggleMenu`. Then we wouldn't need to implement Vue's functionality manually and template would be clearer without 4-lines computing directive.

susnux: No I explicitly moved it to the template, otherwise the listeners would always be added.
But when there is no menu we should not have any listeners, because that element is not interactive.

tests/unit/components/NcAvatar/NcAvatar.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+	})
+
+	it('aria label is set to include status even if status is do-not-disturb', async () => {
+		const status = {
+			icon: '',
+			status: 'dnd',
+			message: 'Playing DnD',
+		}
+
+		const wrapper = await mount(NcAvatar, {
```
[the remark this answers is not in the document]
susnux: I guess you mean `await new Promise(process.nextTick)`  ?

tests/unit/components/NcAvatar/NcAvatar.spec.ts
susnux: Copied from other test :/
No it does not.

tests/unit/components/NcAvatar/NcAvatar.spec.ts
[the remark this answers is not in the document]
ShGKme: > I guess you mean `await new Promise(process.nextTick)` ?

No, Vue's `nextTick`

https://vuejs.org/api/general.html#nexttick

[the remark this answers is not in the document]
ShGKme: But it depends on what are you need to wait there

tests/tsconfig.json
ShGKme, line 1: Looks like a mix of tabs/spaces in GitHub

Pytal (approved):

ShGKme (approved):

JuliaKirschenheuter (approved):

PR #4599, opened 2023-10-03
[next] feat(vitest): migrate from `jest` to `vitest`

raimund-schluessler:
This migrates the test runner from `jest` to `vitest` for the vue 3 `next` branch.

Edit: Fixed, see https://github.com/nextcloud-libraries/nextcloud-vue/pull/4599#issuecomment-1744812436 for details.


raimund-schluessler:
@susnux It seems that `vitest` stumbles upon a dot in the vitest baseConfig. Adjusting it like this:
https://github.com/nextcloud-libraries/nextcloud-vite-config/pull/18/commits/10dfa94f6c84c145cf2e10f3c6420b9cd042a3ee
makes the tests run. Is this dot there for a reason?

package.json
```diff
@@ -116,12 +116,10 @@
     "@nextcloud/browserslist-config": "^3.0.0",
     "@nextcloud/eslint-config": "github:nextcloud-libraries/eslint-config#fe67f6b84e3c13eb0ea2478f44939abc2fe7c99c",
     "@nextcloud/stylelint-config": "^2.3.1",
-    "@nextcloud/vite-config": "github:nextcloud-libraries/nextcloud-vite-config#vue3",
+    "@nextcloud/vite-config": "github:nextcloud-libraries/nextcloud-vite-config#10dfa94f6c84c145cf2e10f3c6420b9cd042a3ee",
```
susnux, line 119: You can also keep the branch name, it should then automatically update to the latest commit.
(the exact commit that is used will be shown in package-lock).

package.json
```diff
@@ -130,16 +128,14 @@
     "gettext-extractor": "^3.7.2",
     "gettext-parser": "^7.0.0",
     "glob": "^10.0.0",
-    "jest": "^29.0.1",
-    "jest-environment-jsdom": "29.7.0",
-    "jest-serializer-vue": "^3.1.0",
-    "jest-transform-stub": "^2.0.0",
+    "jsdom": "^22.1.0",
```
susnux, line 131: Just as a comment:
if it works than ok, but I recommend using happy-dom instead. As jsdom does not provide any new API like buffer etc (unrelated to vitest vs jest but got errors where some API was missing with jsdom).

package.json
raimund-schluessler: Yes, I know. I just wanted to be more specific, so it doesn't break in case we rebase `nextcloud-vite-config`. We anyway have to replace it after a final release of `nextcloud-vite-config`.

susnux (approved):

package.json
raimund-schluessler: I didn't try with `happy-dom`, but I can do that. I just took `jsdom`, because you did in #3565 😉

package.json
raimund-schluessler: But I can also go back to `vue3`. Should also be fine, unless we rebase and someone has to run `npm install`.

ShGKme:
I removed my review request here because I don't feel confident to take responsibility for migration to Vitest.
  
While it simplifies configuration by sharing Vite bundling config, works with ESM, and in some cases faster (but in some other slower because of re-initialization of each imported module in each test again), from my feelings, it's still quite a young tool with a small team. Even Vite, with all its support, still has many issues with almost every minor update.
  
At the same time, I don't have much experience with Vitest or its source to know its "special spots".

package.json
susnux: All good, just wanted to mention that.

package.json
susnux: No its fine :)

susnux: We can wait for a final release ;-)

ShGKme:
> I removed my review request here because I don't feel confident to take responsibility for migration to Vitest.
> 
> While it simplifies configuration by sharing Vite bundling config, works with ESM, and in some cases faster (but in some other slower because of re-initialization of each imported module in each test again), from my feelings, it's still quite a young tool with a small team. Even Vite, with all its support, still has many issues with almost every minor update.
> 
> At the same time, I don't have much experience with Vitest or its source to know its "special spots".

Though `@nextcloud/vue` is not large and we don't have many unit tests, so, we can go back if we find many issues.

tests/unit/components/NcAppSidebar/NcAppSidebarTabs.spec.js
```diff
@@ -83,7 +87,7 @@ describe('NcAppSidebarTabs.vue', () => {
 					},
 				})
 			})
-			it('Issues no warning.', () => {
+			it('Issues no warning.', (onWarning) => {
```
ShGKme, line 90: Shouldn't it be `({ onWarning })`?

susnux:
> Though @nextcloud/vue is not large and we don't have many unit tests, so, we can go back if we find many issues.

That is true and it is a lot easier to configure, especially as dependencies move to ESM and we already getting problems with them. As we can not externalize them for CJS build. Sometime in the future we will have to ESM only too.

tests/unit/components/NcAppSidebar/NcAppSidebarTabs.spec.js
raimund-schluessler: Could be. But these tests don't run anyway at the moment, because the underlying component is not yet migrated to vue3. So these adjustments are rather placeholders, than real fixes yet.

package.json
raimund-schluessler: I switched to `happy-dom` now and it seems to run slightly faster.

raimund-schluessler:
I rebased the vite-config to the latest changes in `main` and switched to `happy-dom`. Should be good now.

I think even though `vitest` is still quite young, its configuration is a lot easier than for `jest`. If we really find an unfixable problem, we can still go back.

PR #4611, opened 2023-10-05
WIP feat(NcColorPicker): add labeled color

ShGKme:
### ☑️ Resolves

* Fix https://github.com/nextcloud/server/issues/40688

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist


JuliaKirschenheuter (approved):

susnux:
Works but we should label the default palette in a follow up

ShGKme:
> Works but we should label the default palette in a follow up

The problem is that it is generated... Have no idea how to combine generation and naming, except having a list of 1000 color names and search for the nearest :D

susnux:
> The problem is that it is generated.

But it is not exported and only used in two places, both generate a palette of 6 colors (-> 18 colors total).
So we could either drop that function completely or provide a hardcoded list for that 18 colors.

ShGKme:
> > The problem is that it is generated.
> 
> But it is not exported and only used in two places, both generate a palette of 6 colors (-> 18 colors total). So we could either drop that function completely or provide a hardcoded list for that 18 colors.

Agree. I'll do it in follow-ups next week:
- Add example with new feature
- Add tests for a new feature
- Name all the colors from the default palette

PR #4629, opened 2023-10-09
Fix height of `label` and `input` on NcInputField

JuliaKirschenheuter:
### ☑️ Resolves

* https://github.com/nextcloud/server/issues/36977

Fix height for text spacing requirements. Checked with ARCToolkit.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot from 2023-10-09 17-12-08 | Screenshot from 2023-10-09 17-11-01

### 🏁 Checklist

### DIFF
--- a/src/components/NcInputField/NcInputField.vue
+++ b/src/components/NcInputField/NcInputField.vue
@@ -436,7 +436,7 @@ export default {
 		position: absolute;
 		margin-inline: 12px 0;
 		// fix height and line height to center label
-		height: 17px;
+		height: 25px;
 		max-width: fit-content;
 		line-height: 1;
 		inset-block-start: 12px;
@@ -468,7 +468,7 @@ export default {
 		inset-block-start: -6px;
 		font-size: 13px; // minimum allowed font size for accessibility
 		background-color: var(--color-main-background);
-		height: 14px;
+		height: 18px;
 		padding-inline: 4px;
 		margin-inline-start: 8px;
 


artonge (approved):

ShGKme:
What is the difference between `🏚️ Before` and `🏡 After`? 🤨

JuliaKirschenheuter:
> What is the difference between `🏚️ Before` and `🏡 After`? 🤨

truncated text

susnux:
Why is the text clipped at the bottom? Also the input and the placeholder text is placed not where it should be (not related to the PR but seen on the screenshots. Something wrong is going on there).
And also the moved label is not centered, if you compare with that [screenshot](https://user-images.githubusercontent.com/1855448/270682015-bdf0ac5d-a85c-4a33-a88b-ab0af6f8346a.png).

susnux:
The increased height now causes this visual bug where the background overflows the bottom border for a short moment:

https://github.com/nextcloud-libraries/nextcloud-vue/assets/1855448/204b6d65-d748-49b2-84e1-375dea018dbb

raimund-schluessler:
@JuliaKirschenheuter Your branch is quite a bit behind `master`. You have to please update it and resolve the conflicts, otherwise a discussion is difficult, because the values you want to adjust have changed in `master`.

But nonetheless, I can confirm the problem. It also occurs in the styleguide. It happens because this rule
https://github.com/nextcloud-libraries/nextcloud-vue/blob/481569be389da594f9c8e1966cbbee354873540d/src/components/NcInputField/NcInputField.vue#L441
is overwritte by this rule
https://github.com/nextcloud-libraries/nextcloud-vue/blob/481569be389da594f9c8e1966cbbee354873540d/src/components/NcInputField/NcInputField.vue#L475

JuliaKirschenheuter:
@susnux do we have any feedback from designers? How could we move forwards?

jancborchardt:
@JuliaKirschenheuter @susnux I am not sure where design input is needed here? The bug seems to be merely CSS where the text is somehow set to overflow: none (or something similar). There should be no heightening of the input, but only fixing of the cutoff and/or white container size.

JuliaKirschenheuter:
issue was solved here: https://github.com/nextcloud-libraries/nextcloud-vue/pull/4718

PR #4631, opened 2023-10-09
[next] feat(NcDateTimePicker): migrate to vue 3

raimund-schluessler:
This PR migrates `NcDateTimePicker` to vue 3 and also brings back `NcTimezonePicker`.

I had to move the scoping with `ScopeComponent` and the import of the datepicker `scss` file into the component, because styleguide would not properly import the style. In fact, it looks like the styleguide never compiles the `src/components/NcDateTimePicker/index.js` file, hence, never imports and scopes the component. Moving it into `NcDateTimePicker.vue` works fine, though. And it seems to be the cleaner solution to me anyway, instead of importing styles in a merely related file.


src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -105,7 +105,7 @@ picked time zone, but you will have to convert it yourself when necessary.
 			v-model="time"
 			type="datetime"
 			:show-timezone-select="true"
-			:timezone-id.sync="tz" /><br>
+			v-model:timezone-id="tz" /><br>
```
ChristophWurst, line 108: v-model is used twice?

raimund-schluessler: Yes, that's a change of vue 3, see here: https://v3-migration.vuejs.org/breaking-changes/v-model.html
`NEW: Multiple v-model bindings on the same component are possible now;`

ShGKme: With different arguments. First has default 'modelValue` argument and the second - `timezoneId`, so they set and listen for different props/events.

ChristophWurst (approved):

PR #4633, opened 2023-10-11
fix(NcAppNavigation): add focus trap on mobile and improve a11y

ShGKme:
### ☑️ Resolves

* A part of https://github.com/nextcloud/server/issues/36964

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
nav-before | nav-after

### 🚧 Tasks

- [x] Add focus trap, and activate it on **open** navigation on **mobile**, deactivate otherwise
- [x] Improve a11y a bit:
  - [x] Use `<nav>` instead of `role="navigation"` cc @JuliaKirschenheuter to confirm that this is fine
  - [x] Move the toggle button out from the navigation
  - [x] Add `ariaLabel` and `ariaLabelledBy` props to be set on navigation
- [x] Close navigation by <kbd>ESC</kbd> on mobile
- [x] Update tests
  - [x] Test by class instead of internal data property (testing `is closed or open` state by class also is not great, but we should not test by internal implementation as data property is not a part of the component interface). Alternative: `aria-hidden`.
  - [x] Add mobile state test
  - [x] Add a11y tests

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppNavigation/NcAppNavigation.vue
+++ b/src/components/NcAppNavigation/NcAppNavigation.vue
@@ -55,6 +55,7 @@ emit('toggle-navigation', {
 
 <template>
 	<div id="app-navigation-vue"
+		ref="appNavigationContainer"
 		class="app-navigation"
 		role="navigation"
 		:class="{'app-navigation--close':!open }">
@@ -77,8 +78,11 @@ emit('toggle-navigation', {
 <script>
 import NcAppNavigationToggle from '../NcAppNavigationToggle/index.js'
 import isMobile from '../../mixins/isMobile/index.js'
+import { getTrapStack } from '../../utils/focusTrap.js'
 
 import { emit, subscribe, unsubscribe } from '@nextcloud/event-bus'
+
+import { createFocusTrap } from 'focus-trap'
 
 export default {
 	name: 'NcAppNavigation',
@@ -92,6 +96,7 @@ export default {
 	data() {
 		return {
 			open: true,
+			focusTrap: null,
 		}
 	},
 	computed: {
@@ -103,6 +108,10 @@ export default {
 	watch: {
 		isMobile() {
 			this.open = !this.isMobile
+			this.toggleFocusTrap()
+		},
+		open() {
+			this.toggleFocusTrap()
 		},
 	},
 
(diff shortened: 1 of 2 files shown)


szaimen:
node tests seems to be failing

src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -135,9 +168,27 @@ export default {
...
+			if (this.isMobile && this.open) {
+				this.focusTrap.activate()
+			} else {
+				this.focusTrap.deactivate()
+			}
+		},
+
+		handleEsc() {
+			if (this.isMobile) {
+				this.toggleNavigation(false)
```
szaimen, line 189: This does not seem to work currently in my testing

szaimen (commented):
Tested and the focus trap seems to work :)

However the ESC key doesnt toggle the navigation currently


https://github.com/nextcloud-libraries/nextcloud-vue/assets/42591237/9037fba3-e4ff-4d16-a683-2ffe132fc45f

src/components/NcAppNavigation/NcAppNavigation.vue
szaimen: the reason was probably that I tested on a screen with higher res than 1024px (so ismobile was false)

ShGKme:
PR Status update: checking issue with Talk loosing tab navigation with focus trap

ShGKme:
The Talk issue is related but appears in any focus trap container with input with the trailing button visible only on focus.

What happens:
1. <kbd>Tab</kbd> -> Focus on input.
2. Now the "Clear" trailing button is visible.
3. <kbd>Tab</kbd> -> Focus the trailing button.
4. <kbd>Tab</kbd> -> issue
   1. Blurring focus from the trailing button hides it, and removes it from the DOM
   2. `focus-trap` with `MutationObserver` sees it and thinks that the current focused element is removed
   3. Then `focus-trap` moves the focus to the first element of the container

focus

Source:
https://github.com/focus-trap/focus-trap/blob/a49e9b247bb4d15e3a16ddcc9e6b847a62725c77/index.js#L862-L875

This behavior is not configurable.

IMO, Possible solutions could be:
- Fix it on the `focus-trap` side
- Do not remove the trailing button from the DOM, only hide it
- Hide the trailing button with a delay

ShGKme:
Another issue with the `NcAppNavigationItem`. Focus gets stuck on the last item

susnux (approved):

susnux:
Maybe report this upstream, but

> On Talk: Do not remove the trailing button from the DOM, only hide it

Sounds the best for now.

GretaD (approved):

JuliaKirschenheuter (approved):

AndyScherzinger (approved):

PR #4643, opened 2023-10-13
feat(NcIconSvgWrapper): allow to render raw svg paths

raimund-schluessler:
### ☑️ Resolves

This allows to render raw SVG paths which prevents the overhead of having to sanitize the SVG icon for use in `v-html`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -56,6 +56,7 @@
         "@babel/preset-typescript": "^7.22.5",
         "@cypress/vue2": "^2.0.1",
         "@fontsource/roboto": "^5.0.0",
+        "@mdi/js": "^7.3.67",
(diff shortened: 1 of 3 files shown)


susnux (dismissed):
Very nice, less overhead :sparkles: 

I think you need to add `@mdi/js` as a dependency.

Pytal (commented):
Can we reduce duplication by putting the identical props in a computed property then v-bind them ?

raimund-schluessler:
> Can we reduce duplication by putting the identical props in a computed property then v-bind them ?

Sure, why not. I adjusted it.

ShGKme (approved):

Pytal (approved):

PR #4645, opened 2023-10-13
fix(NcModal): Some modal improvments

### LINKED ISSUE #4627: NcModal close button is inaccessible with scrollable content
Issue:
- Probably a regression from #4350
- After re-ordering of DOM elements `modal-container__close` and `modal-container__content` content can overlap close button, making it inaccessible

Screenshots:
image
image

Possible solution:
- set `z-index: 1` for `modal-container__close`

_Originally posted by @Antreesy in https://github.com/nextcloud-libraries/nextcloud-vue/issues/4350#issuecomment-1752659310_

susnux:
### ☑️ Resolves

* Fix #4627 
* Fix https://github.com/nextcloud-libraries/nextcloud-vue/issues/3117

1. Revert hack to make unused buttons invisible instead of setting `display: none`
    The buttons are placed absolute, therefor they do not interact with the modal container and content.
    Meaning even if only on button is shown the modal is centered correctly.
2. Ensure the modal header does not overflow the modal content
    The modal height is max. 90%, but if the window is not that tall or the modal is tall, then the header might overlap the content. So fix this by setting the modal max. height to 100% - 2*modal-header (because of the centered placement).
3. Ensure the close button is always clickable and not covert by content. (simply setting a z-index on it).

### 🖼️ Screenshots
You can see the header no longer overflows the content. And the button is on top of the modal content even if the default content margin of 50px is overridden.

🏚️ Before | 🏡 After
---|---
Screenshot_20231013_171853 | Screenshot_20231013_171949

And for the revert of the visibility hack, here is a "after" screenshot with only on button, you can see it is still centered perfectly:
image

### 🏁 Checklist

### DIFF
--- a/src/components/NcModal/NcModal.vue
+++ b/src/components/NcModal/NcModal.vue
@@ -288,9 +288,6 @@ export default {
 						<NcButton v-show="hasPrevious"
 							type="tertiary-no-background"
 							class="prev"
-							:class="{
-								invisible: !hasPrevious
-							}"
 							:aria-label="prevButtonAriaLabel"
 							@click="previous">
 							<template #icon>
@@ -322,9 +319,6 @@ export default {
 						<NcButton v-show="hasNext"
 							type="tertiary-no-background"
 							class="next"
-							:class="{
-								invisible: !hasNext
-							}"
 							:aria-label="nextButtonAriaLabel"
 							@click="next">
 							<template #icon>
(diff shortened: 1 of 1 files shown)


Antreesy (approved):
Agree with changes

raimund-schluessler (approved):

PR #4646, opened 2023-10-13
[next] feat(NcActions*): migrate to vue 3

raimund-schluessler:
### ☑️ Resolves

This migrates the `NcActions` component to vue 3.


raimund-schluessler:
Rebased.

susnux (approved):

raimund-schluessler:
Resolved conflicts.

raimund-schluessler:
Another review would be very welcome :slightly_smiling_face:

raimund-schluessler:
Rebased and resolved conflicts.

src/components/NcActions/NcActions.vue
```diff
@@ -1008,59 +1019,44 @@ export default {
...
-						}),
-					},
+					class: 'action-item action-item--single',
+					'aria-label': ariaLabel,
+					title,
+					// If it has a menuName, we use a secondary button
+					type: this.type || (buttonText ? 'secondary' : 'tertiary'),
+					disabled: this.disabled || action?.props?.disabled,
+					ariaHidden: this.ariaHidden,
+					...action?.props,
```
ShGKme, line 1043: If I'm not wrong, passing `class` to action will re-define initial `action-item action-item--single` classes here

src/mixins/actionGlobal.js
```diff
@@ -19,14 +19,14 @@
...
+import { warn } from 'vue'
 
 export default {
 	before() {
 		// all actions requires a valid text content
 		// if none, forbid the component mount and throw error
-		if (!this.$slots.default || this.text.trim() === '') {
-			Vue.util.warn(`${this.$options.name} cannot be empty and requires a meaningful text content`, this)
+		if (!this.$slots.default() || this.text.trim() === '') {
+			warn(`${this.$options.name} cannot be empty and requires a meaningful text content`, this)
```
ShGKme, line 29: How does this part work? o_O

src/components/NcActions/NcActions.vue
```diff
@@ -1004,57 +1015,45 @@ export default {
...
 
-			let title = action?.componentOptions?.propsData?.title
+			let title = action?.props?.title
 			// Show a default title for single actions if none is present
 			if (!(this.forceName || title)) {
 				title = text
 			}
 
-			return h('NcButton',
+			return h(resolveComponent('NcButton'),
```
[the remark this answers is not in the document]
raimund-schluessler: You are right, we don't. I tried to import it via the string `'NcButton'` like in vue 2, but this didn't work anymore. Using the object `NcButton` does, though.

src/components/NcActions/NcActions.vue
```diff
@@ -1065,95 +1064,63 @@ export default {
...
-					: h('DotsHorizontal', {
-						props: {
-							size: 20,
-						},
-					})
-			)
-			return h('NcPopover',
+					: h(resolveComponent('DotsHorizontal'), { size: 20 })
+				)
+			return h(resolveComponent('NcPopover'),
```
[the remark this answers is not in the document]
raimund-schluessler: Fixed.

src/components/NcActions/NcActions.vue
```diff
@@ -960,18 +964,25 @@ export default {
...
+		const actions = []
+		// We have to iterate over all slot elements
+		this.$slots.default?.().forEach(vnode => {
+			if (this.isAction(vnode)) {
+				actions.push(vnode)
+				return
+			}
+			// If we encounter a Fragment, we have to check its children too
+			if (vnode?.type === Fragment) {
+				vnode?.children?.forEach?.(child => {
```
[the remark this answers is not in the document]
raimund-schluessler: Probably not, I just wanted to be safe. It works without checking for `null` as well. I changed it.

src/components/NcActions/NcActions.vue
```diff
@@ -702,7 +704,7 @@ export default {
 		 */
 		boundariesElement: {
 			type: Element,
-			default: () => document.querySelector('body'),
+			default: () => document?.querySelector('body'),
```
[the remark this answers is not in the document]
raimund-schluessler: I guess it's not. If I remember correctly, I saw errors here when first working on the vue 3 migration, but it works without checking for the existence of `document` now. Changed it.

src/components/NcActions/NcActions.vue
raimund-schluessler: You are right, it was overwritten. I fixed it now, but I find the code fec8239d740ae7868a2747ec0b228b54572ceea0 for this a bit ugly. :see_no_evil:  If you have a better solution, please let me know.

src/mixins/actionGlobal.js
raimund-schluessler: What exactly do you mean? I didn't write it, originally, I only adjusted it for vue 3.

src/components/NcActions/NcActions.vue
```diff
@@ -1008,59 +1019,45 @@ export default {
...
-						...(!!clickListener && {
-							click: (event) => {
-								if (clickListener) {
-									clickListener(event)
-								}
-							},
-						}),
-					},
+					...action?.props,
+					class: classes,
```
ShGKme, line 1038:
```js
class: ['action-item', 'action-item--single', action?.props?.class]
```

src/components/NcActions/NcActions.vue
```diff
@@ -1008,59 +1019,45 @@ export default {
...
-						// we bind it to execute on click and forward the click event
-						...(!!clickListener && {
-							click: (event) => {
-								if (clickListener) {
-									clickListener(event)
-								}
-							},
-						}),
-					},
+					...action?.props,
```
ShGKme, line 1037: Maybe this is what we need?
https://vuejs.org/api/render-function.html#mergeprops

src/components/NcActions/NcActions.vue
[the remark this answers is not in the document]
susnux: Maybe for unit tests, but if you unit test a DOM component you probably want some kind of DOM environment -> There should always be the document.

--- 

Well while writing this: There is one situation when the document is undefined: SSR.
Not sure if we ever will use SSR but it will fail here, because document is only defined after (and in) the mounted hook.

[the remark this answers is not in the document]
raimund-schluessler: Maybe we fix / adjust it when we do SSR and leave it like this until then.

src/components/NcActions/NcActions.vue
raimund-schluessler: Problem is that `action?.props?.class` can be `undefined` and we would end up with `['action-item', 'action-item--single', undefined]`.

src/components/NcActions/NcActions.vue
raimund-schluessler: Yes, that could be it. I will have a look.

src/components/NcActions/NcActions.vue
ShGKme: Is undefined a problem?

https://github.com/vuejs/core/blob/1789c1caac6fdd3ef3d676d628d7ad6e50888b96/packages/shared/src/normalizeProp.ts#L63-L82

raimund-schluessler: Ah, interesting. I thought to have seen an `undefined` in the class list in the DOM, but that's not the case. Let's use it like this then.

src/components/NcActions/NcActions.vue
raimund-schluessler: That is a nice helper. Thanks for pointing me to it. Cleans up the code a lot.

src/components/NcActions/NcActions.vue
raimund-schluessler: However, with `mergeProps` its superfluous now :slightly_smiling_face:

raimund-schluessler:
@nextcloud-libraries/developers Honestly, it is a bit demotivating that very few seem to care that the components finally support vue 3.

And at the risk of repeating myself: vue 2 reaches EOL in 45 business days. After that, it won't receive any more updates. Even if the risk for security issues in vue 2 is likely really low, it doesn't seem to be the best approach to not give developers at least the option to use supported libraries.

ShGKme (approved):

src/mixins/actionGlobal.js
ShGKme: Sorry, I missed that it was an old part. I just don't understand how it works in general, who and where call this `before()` custom option.

raimund-schluessler: Yes, I also find it a bit weird. Every `Action*` component, e.g. `ActionButton` uses this mixin to check that either `text` or the default slot are present.

PR #4650, opened 2023-10-15
fix(NcAppNavigationItem): Ensure primary styling when the item is active

### LINKED ISSUE #4571: NcAppNavigationItem: wrong color of action menu "three dots" for active item
When the item is active, the three dots icon for the actions menu has wrong color:

2023-09-25T12:35:22,036934733+02:00

Reproduced by manually adding class `active` to the `div.app-navigation-entry` element.

Ping @szaimen, since you worked on this recently.

susnux:
### ☑️ Resolves

* Fix #4571

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
NcAppNavigationItem-primary-on-active-route|image

### 🏁 Checklist


cypress/visual/AppNavigationItem.cy.ts
```diff
@@ -0,0 +1,36 @@
...
+				plugins: [router],
+			},
+			router,
+		})
+
+		cy.contains('.app-navigation-entry', 'Home').compareSnapshot('NcAppNavigationItem-primary-on-active-route')
+		cy.contains('.app-navigation-entry', 'Foo').compareSnapshot('NcAppNavigationItem-tertiary-non-active-route')
+
+		cy.contains('.app-navigation-entry', 'Home').find('.app-navigation-entry__actions').click()
+		cy.get('.app-navigation-entry--editing').compareSnapshot('NcAppNavigationItem-primary-on-active-route-menu-active')
```
ShGKme, line 34: I'm always afraid of screenshot testing 🙈 
Too many factors changes pixels...

cypress/visual/AppNavigationItem.cy.ts
```diff
@@ -0,0 +1,36 @@
+import { mount } from 'cypress/vue2'
+import Vue, { defineComponent } from 'vue'
+import VueRouter from 'vue-router'
+
+import NcAppNavigationItem from '../../src/components/NcAppNavigationItem/NcAppNavigationItem.vue'
+
+describe('NcAppNavigationItem', () => {
+	it('With router link is primary on active', () => {
```
ShGKme, line 8: `it With router link is primary on active` sounds weird for me =D

If the `spec` style is not used, maybe use `test` style?

ShGKme (approved):

cypress/visual/AppNavigationItem.cy.ts
ShGKme: So, CI says 

> Error: The "NcAppNavigationItem-primary-on-active-route" image is different. Threshold limit exceeded! differently

And we even cannot look at it on the GitHub to see if the test works differently here or the component doesn't work as expected.

susnux: currently trying to fix this, this happens if I create the snapshots on my local machine. CI uses ubuntu which has different color interpolation / aliasing. (not sure how this is called).

susnux: > I'm always afraid of screenshot testing 🙈 Too many factors changes pixels...

Yes but hard to otherwise test visual stuff :wink:

cypress/visual/AppNavigationItem.cy.ts
susnux: yes, makes sense :)

cypress/visual/AppNavigationItem.cy.ts
raimund-schluessler: We have the same issue for the `NcAppSidebar` tests. But you can download the snapshots generated by the cypress action and check them in as the base snapshots. That makes the tests pass on GitHub, but they of course still fail locally than. But that might be acceptable.

That being said, there seems to be something else wrong, that just the aliasing. These are the remote snapshots:
NcAppNavigationItem-primary-on-active-route
NcAppNavigationItem-primary-on-active-route-menu-active
NcAppNavigationItem-tertiary-non-active-route

susnux: @raimund-schluessler but all tests have passed, see changes for the checked-in snapshots which look fine.

BTW to create working snapshots I use something like:
`docker run -it -v .:/app --entrypoint /bin/sh -w app cypress/included:13.3.1 npm run cypress:update-snapshots`

raimund-schluessler: > @raimund-schluessler but all tests have passed, see changes for the checked-in snapshots which look fine.
> 
> BTW to create working snapshots I use something like: `docker run -it -v .:/app --entrypoint /bin/sh -w app cypress/included:13.3.1 npm run cypress:update-snapshots`

Right, sorry. I looked at the `diff`s instead of the actual ones :see_no_evil: All good in that case.

susnux: Ok :)
Feel free to leave a review

raimund-schluessler: > BTW to create working snapshots I use something like: `docker run -it -v .:/app --entrypoint /bin/sh -w app cypress/included:13.3.1 npm run cypress:update-snapshots`

Could be nice to put this in the readme.

raimund-schluessler (approved):

cypress/visual/AppNavigationItem.cy.ts
ShGKme: > Yes but hard to otherwise test visual stuff 😉

Manually, having more examples in the docs =D

Having tests is not the final goal, they should not make more troubles than profits.

From my experience, snapshot and visual regression tests often make more trouble, because they require reproducible environment (e.g. launch Cypress GUI to debug results in a different screenshot than launch without GUI), has no result in the CI log, hard to tell that goes wrong when there is no obvious difference, and developers always have an option to say "just a valid change, let's update snapshot".

PR #4652, opened 2023-10-15
[stable7] fix(l10n): Update translations with current master translations

susnux:
### ☑️ Resolves

* Fix https://github.com/nextcloud-libraries/nextcloud-vue/issues/4548

Update translations from current master as transifex does not allow automatically update multiple branches.
Done like this:
```
pushd l10n
for f in *; do msgmerge -N ../master-l10n/$f $f | sed '/^#~/d' | sed '/^$/N;/^\n$/D' > a; mv a $f; done
popd
```

### 🏁 Checklist


raimund-schluessler (approved):
Fine with me. But it is hard to tell, whether the translations really match, as we might not have backported every language change.

susnux:
> Fine with me. But it is hard to tell, whether the translations really match, as we might not have backported every language change.

They are only used if they fully match as `-N` disables fuzzy matching.
Just needed to fix the translations for some of the languages where they were broken.

skjnldsv (approved):

PR #4653, opened 2023-10-15
feat(NcSelect): Allow to filter users by email notation

### LINKED ISSUE #4491: NcSelect does not recognize email addresses when angle bracket notation is used
When searching a user with an email address while the email is entered in angle brackets, NcSelect should not filter out the matching options.

John john@example.org entered ✔️  | John <john@example.org> entered ❌ 
-|-
grafik|grafik

Expected behavior: When entering `John <john@example.org>` the user should be found as well as when entering without brackets. 

I would call it a bug, since upon a user's report, this worked in earlier versions, but I can't confirm this. https://github.com/nextcloud/polls/issues/3036

susnux:
### ☑️ Resolves

* Fix #4491 

Allows to filter by the email notation like `Jane <j.doe@example.com>`

### 🏁 Checklist

### DIFF
--- a/src/components/NcListItemIcon/NcListItemIcon.vue
+++ b/src/components/NcListItemIcon/NcListItemIcon.vue
@@ -139,11 +139,11 @@ It might be used for list rendering or within the multiselect for example
 		<div class="option__details">
 			<NcHighlight class="option__lineone"
 				:text="name"
-				:search="search" />
+				:search="searchParts[0]" />
 			<NcHighlight v-if="isValidSubname && isSizeBigEnough"
 				class="option__linetwo"
 				:text="subname"
-				:search="search" />
+				:search="searchParts[1]" />
 			<span v-else-if="hasStatus">
 				<span>{{ userStatus.icon }}</span>
 				<span>{{ userStatus.message }}</span>
@@ -313,6 +313,21 @@ export default {
 				'--height': this.avatarSize + 2 * margin + 'px',
 				'--margin': this.margin + 'px',
 			}
+		},
+
+		/**
+		 * Seperates the search property into two parts, the first one is the search part on the name, the second on the subname.
+		 * @return {[string, string]}
+		 */
+		searchParts() {
+			// Match the email notation like "Jane <j.doe@example.com>" with the email address as matching group
+			const EMAIL_NOTATION = /^([^<]*)<([^>]+)>?$/
+
+			const match = this.search.match(EMAIL_NOTATION)
+			if (this.isNoUser || !match) {
+				return [this.search, this.search]
+			}
+			return [match[1].trim(), match[2]]
 		},
 	},
 
--- a/src/components/NcSelect/NcSelect.vue
+++ b/src/components/NcSelect/NcSelect.vue
@@ -951,14 +951,19 @@ export default {
 		},
 
 		localFilterBy() {
+			// Match the email notation like "Jane <j.doe@example.com>" with the email address as matching group
+			const EMAIL_NOTATION = /[^<]*<([^>]+)/
+
 			if (this.filterBy !== null) {
 				return this.filterBy
 			}
 			if (this.userSelect) {
 				return (option, label, search) => {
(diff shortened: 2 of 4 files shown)


raimund-schluessler:
Might be, that I do something wrong, but it doesn't seem to work for me in the styleguide:
grafik

susnux:
> Might be, that I do something wrong, but it doesn't seem to work for me in the styleguide

@raimund-schluessler Ah yes there was an issue if there is no `subname` defined, fixed that:

https://github.com/nextcloud-libraries/nextcloud-vue/assets/1855448/a6da226c-8f2f-45ce-861a-c4d09c37ebdc

src/components/NcSelect/NcSelect.vue
```diff
@@ -951,14 +951,19 @@ export default {
...
 			if (this.filterBy !== null) {
 				return this.filterBy
 			}
 			if (this.userSelect) {
 				return (option, label, search) => {
-					return (`${label} ${option.subname}` || '')
-						.toLocaleLowerCase()
-						.indexOf(search.toLocaleLowerCase()) > -1
+					const match = search.match(EMAIL_NOTATION)
+					return (match && option.subname && option.subname.toLocaleLowerCase().indexOf(match[1].toLocaleLowerCase()) > -1)
```
raimund-schluessler, line 963:
```suggestion
					return (match && option.subname?.toLocaleLowerCase?.().indexOf(match[1].toLocaleLowerCase()) > -1)
```

raimund-schluessler (approved):
Works now as expected (in the docs).

Pytal (commented):
Is it possible to have the matching chars bolded like when typing an email without angle brackets?

susnux:
> Is it possible to have the matching chars bolded like when typing an email without angle brackets?

Yes but need to adjust `NcListItemIcon` then.

susnux:
@Pytal done


https://github.com/nextcloud-libraries/nextcloud-vue/assets/1855448/9272ff6e-4107-4a5e-b562-c6cdb0c68e4d

src/components/NcSelect/NcSelect.vue
susnux: Done

raimund-schluessler:
> > So, when typing <olivia@example in NcSelect, do the matching in NcSelect and set this.search to olivia@example without the token. This way, the highlighting should work, and we wouldn't need to duplicate the matching code in NcListItemIcon.
> 
> 
> 
> Yes but when typing `Olivia <olivia@example` **Olivia** would not be highlighted.
> 
> So I think this is also a feature on the `NcListItemIcon`

Ah, yes, I see. Valid point.

src/components/NcListItemIcon/NcListItemIcon.vue
```diff
@@ -314,6 +314,18 @@ export default {
...
+		/**
+		 * Seperates the search property into two parts, the first one is the search part on the name, the second on the subname.
+		 * @return {[string, string]}
+		 */
+		searchParts() {
+			// Match the email notation like "Jane <j.doe@example.com>" with the email address as matching group
+			const EMAIL_NOTATION = /^([^<]*)<([^>]+)>?$/
+
+			const match = this.search.match(EMAIL_NOTATION)
+			return this.isNoUser || !match ? [this.search, this.search] : [match[1].trim(), match[2]]
```
Pytal, line 327:
```suggestion
			if (this.isNoUser) {
				return [this.search, this.search]
			}
			return !match ? [this.search, this.search] : [match[1].trim(), match[2]]
```

susnux: done

Pytal (approved):

dartcafe:
@Pytal Any chance to get this backported to v7?

Pytal:
> @Pytal Any chance to get this backported to v7?

We don't normally backport features @dartcafe, so I'd recommend using the latest v8 of the lib :)

PR #4654, opened 2023-10-15
fix: missing trailing comma

raimund-schluessler:
### ☑️ Resolves

Fixes a missing trailing comma. Honestly, I am not really convinced there should be a comma, but it makes `eslint` happy :man_shrugging:


susnux (approved):

Antreesy (approved):
If eslint happy, I'm happy too

ShGKme:
> I am not really convinced there should be a comma, but it makes `eslint` happy 🤷‍♂️

There should, because it is the end of one object property value, so adding a new property would require modifying the first property by adding a comma

PR #4665, opened 2023-10-18
fix(NcTextField): do not pass all props to InputField BY stop passing $props

ShGKme:
### ☑️ Resolves

HTML Validation issue.

Currently `NcTextField` passes both `$attrs` and `$props` to the `NcInputField`.

```html
<NcInputField v-bind="{ ...$attrs, ...$props }">
```

While passing `$attrs` makes sense, passing `$props` is valid only when both components have **exactly** the same props. But they don't. `NcTextField` has its own prop `trailingButtonIcon` and may have more props in the future.

---

**Proposal:** same as `NcPopover`

Currently, all `NcInputField` props are defined as `NcTextField` props. But actually they are not `NcTextField` props. They are not used but `NcTextField`, it is not aware of most of them at all. It is used only for styleguidist documentation generation.

The proposal is to make component a transparent wrapper by passing attributes directly, defining only necessary props. 

**Alternative solution:** 

~~While I personally prefer this solution,~~ there is an alternative: filter out props on proxying.

See: https://github.com/nextcloud-libraries/nextcloud-vue/pull/4666

### 🖼️ Screenshots

`trailingbuttonicon` is not a valid `<input>` attribute.

image

### 🏁 Checklist

### DIFF
--- a/src/components/NcTextField/NcTextField.vue
+++ b/src/components/NcTextField/NcTextField.vue
@@ -130,9 +130,10 @@ export default {
 </docs>
 
 <template>
-	<NcInputField v-bind="{...$attrs, ...$props }"
-		ref="inputField"
-		:trailing-button-label="clearTextLabel"
+	<NcInputField ref="inputField"
+		:trailing-button-label="trailingButtonLabel"
+		:type="type"
+		v-bind="$attrs"
 		v-on="$listeners"
 		@input="handleInput">
 		<!-- Default slot for the leading icon -->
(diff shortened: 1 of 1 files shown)


szaimen:
Sorry I fear I cannot really judge the code changes

src/components/NcTextField/NcTextField.vue
```diff
@@ -171,11 +172,39 @@ export default {
 	inheritAttrs: false,
 
 	props: {
-		...NcInputField.props,
+		/**
+		 * Any NcInputField props
```
Pytal, line 176:
```suggestion
		 * Any [NcInputField](#/Components/NcInputField?id=ncinputfield) props
```

Pytal (approved):

src/components/NcTextField/NcTextField.vue
ShGKme: I didn't know that possible =D

ShGKme:
Closing in flavor of an alternative approach.

PR #4666, opened 2023-10-18
fix(Nc*Field): do not pass all props to InputField BY filtering $props

ShGKme:
### ☑️ Resolves

HTML Validation issue.

Currently `NcTextField` passes both `$attrs` and `$props` to the `NcInputField`.

```html
<NcInputField v-bind="{ ...$attrs, ...$props }">
```

While passing `$attrs` makes sense, passing `$props` is valid only when both components have **exactly** the same props. But they don't. `NcTextField` has its own prop `trailingButtonIcon` and may have more props in the future.

**Proposal:** same as `NcSelect`.

Filter props to proxy only original `NcInputField` props.

**Alternative solution:** do not use props for proxying.

See: https://github.com/nextcloud-libraries/nextcloud-vue/pull/4665

### 🖼️ Screenshots

`trailingbuttonicon` is not a valid `<input>` attribute.

image

### 🏁 Checklist

### DIFF
--- a/src/components/NcPasswordField/NcPasswordField.vue
+++ b/src/components/NcPasswordField/NcPasswordField.vue
@@ -128,6 +128,21 @@ import { generateOcsUrl } from '@nextcloud/router'
 import { t } from '../../l10n.js'
 import logger from '../../utils/logger.js'
 
+/**
+ * @typedef PasswordPolicy
+ * @property {object} api - The URLs to the password_policy app methods
+ * @property {string} api.generate - The URL to the password generator
+ * @property {string} api.validate - The URL to the password validator
+ * @property {boolean} enforceNonCommonPassword - Whether to enforce non common passwords
+ * @property {boolean} enforceNumericCharacters - Whether to enforce numeric characters
+ * @property {boolean} enforceSpecialCharacters - Whether to enforce special characters
+ * @property {boolean} enforceUpperLowerCase - Whether to enforce upper and lower case
+ * @property {number} minLength - The minimum length of the password
+ */
+
+/** @type {PasswordPolicy|null} */
+const passwordPolicy = loadState('core', 'capabilities', {}).password_policy || null
+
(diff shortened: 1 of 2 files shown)


szaimen:
Sorry I fear I cannot really judge the code changes

ChristophWurst (approved):
makes sense

ShGKme:
> makes sense

Thanks! Could you also have a look at this PR with an alternative (more preferred by me to be honest): https://github.com/nextcloud-libraries/nextcloud-vue/pull/4665

ShGKme:
Although an alternative approach with a classic transparent wrapper approach (with proxying only `$attrs`) seems more clean (and performant) for me and @susnux, I checked how IDEs handle the component, how Vue 3.3 works now with new implementation of `defineProps<WrappedProps & NewProps>`, discussed it with some experts and decided that this PR approach is better in terms of DX.

Even with `Vue 3.3 + SFC Setup + TS` it is still not possible to add typing for attributes for wrapper components.

Waiting for Vue to add something for defining attrs types for component wrappers 😭

src/components/NcPasswordField/NcPasswordField.vue
```diff
@@ -142,18 +158,32 @@ export default {
...
+
 		/**
-		 * Additional error message
-		 *
-		 * This will be displayed beneath the input field
+		 * Controls whether to display the trailing button.
 		 */
-		helperText: {
-			type: String,
-			default: '',
```
ShGKme, line 154: `helperText` removed, because it is exactly the same as in NcInputField and already defined

susnux (approved):
Makes sense, even if I think the other PR is cleaner :disappointed:

ShGKme:
> Makes sense, even if I think the other PR is cleaner 😞

Clean, fast, without any auto-complete in IDE :(

Pytal (approved):

PR #4671, opened 2023-10-19
fix(NcActions): Allow to set attributes to inline actions

susnux:
### ☑️ Resolves

Currently inline action do not receive the attributes set on them, this PR allows to set e.g. `data-test-id` to buttons for integration tests.

### 🏁 Checklist


skjnldsv (approved):

ShGKme (approved):
Makes sense to pass attributes, thought, we should try not use `data-test-id` here for testing, and use any semantic value (e.g. text).

PR #4695, opened 2023-10-24
[next] fix(NcListItemIcon): correctly use default slot in examples

raimund-schluessler:
### ☑️ Resolves

* This brings back the icons in the `NcListItemIcon` examples. I couldn't really figure out why it doesn't work when simply using `<template>`. It only works when not wrapping the default slot content, or when using `<template #default>`.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot 2023-10-24 at 10-05-28 Nextcloud Vue Style Guide | Screenshot 2023-10-24 at 10-05-14 Nextcloud Vue Style Guide

### DIFF
--- a/src/components/NcListItemIcon/NcListItemIcon.vue
+++ b/src/components/NcListItemIcon/NcListItemIcon.vue
@@ -32,9 +32,7 @@ It might be used for list rendering or within the multiselect for example
 ```vue
 	<template>
 		<NcListItemIcon name="User 1" :avatar-size="44">
-			<template>
-				<Account :size="20" />
-			</template>
+			<Account :size="20" />
 		</NcListItemIcon>
 	</template>
 	<script>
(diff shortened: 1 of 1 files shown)


ShGKme (approved):

ShGKme:
> * . I couldn't really figure out why it doesn't work when simply using `<template>`. It only works when not wrapping the default slot content, or when using `<template #default>`.

In Vue 3 `<template>` has special meaning only in combination with directives (`v-if`, `v-for`, `v-slot`). Otherwise, it works as an original HTML element `<template>`. (Though, I have no idea why anyone would need native `<template>` in a Vue template).

https://template-explorer.vuejs.org/#eyJzcmMiOiI8ZGl2PlxyXG4gIDx0ZW1wbGF0ZT5cclxuICAgIENvbnRlbnRcclxuICA8L3RlbXBsYXRlPlxyXG48L2Rpdj4iLCJvcHRpb25zIjp7fX0=

And there is no ESLint rule for that...

susnux (approved):

susnux:
> And there is no ESLint rule for that...

@ShGKme I think `vue/no-lone-template` is about that?

skjnldsv (approved):

ShGKme:
> > And there is no ESLint rule for that...
> 
> @ShGKme I think [`vue/no-lone-template`](https://eslint.vuejs.org/rules/no-lone-template.html) is about that?

I really searched 🙈 

Though, I guess it won't help in this case because it is in a code-block in a markdown...

PR #4699, opened 2023-10-26
Create outline for `focus-within` on RadioSwitch

JuliaKirschenheuter:
### ☑️ Resolves

* Fix https://github.com/nextcloud/server/issues/40706

- [x] Equalize styles for focused state: keyboard as well as mouse click
- [x] Apply styles for unchecked state

### 🖼️ Screenshots

|  Before   | After                                                        |
|-------------|---------------------------------------------------------------|
| image | Screenshot from 2023-10-26 13-28-29 |-------------|---------------------------------------------------------------|
| image | Screenshot from 2023-10-26 13-26-04 |-------------|---------------------------------------------------------------|
| Screenshot from 2023-10-26 13-34-43 | Screenshot from 2023-10-26 13-29-24 |-------------|---------------------------------------------------------------|
|  | |-------------|---------------------------------------------------------------|
| Screenshot from 2023-10-26 13-35-20 | Screenshot from 2023-10-26 13-29-14

### 🏁 Checklist

### DIFF
--- a/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
+++ b/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
@@ -620,8 +620,11 @@ export default {
 		margin: 4px $icon-margin;
 	}
 
-	&__input:focus-visible + &__content {
-		outline: 2px solid var(--color-primary-element) !important;
+	&__input:focus-visible + &__content,
+	&__input:focus-visible {
+		outline: 2px solid var(--color-main-text);
+		border-color: var(--color-main-background);
+		outline-offset: -2px;
 	}
 
 	&--disabled &__content {


susnux:
Does not work on the button style:

image

susnux:
Maybe something like `outline-offset: -2px;`?

JuliaKirschenheuter:
> Does not work on the button style:
image

Do you mean?:

Peek 2023-10-27 13-00

ShGKme:
> Maybe something like `outline-offset: -2px;`?

I always used `border+padding` to make an outline inside an element, never heard about this property 🙈

src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
```diff
@@ -620,8 +620,12 @@ export default {
 		margin: 4px $icon-margin;
 	}
 
-	&__input:focus-visible + &__content {
-		outline: 2px solid var(--color-primary-element) !important;
+	&__input:focus-visible + &__content,
+	&__content:active,
+	&__input:focus-within + &__content {
```
Pytal, line 625: Are the `:active` and `:focus-within` pseudo-classes necessary? As now the borders are visible when just clicking with a mouse and stay until clicking elsewhere

JuliaKirschenheuter: in my opinion we need it. without: 
Peek 2023-10-30 18-46

susnux (approved):
Looks good overall, but I agree with @Pytal (while it is not a blocker for me)

src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
susnux: I agree with @Pytal, I think it is a bit weird to have to focus outline also on click, we do not have it for buttons or other input elements.

Pytal: Was previously discussed with @michaelnissenbaum that only keyboard focus required focus borders but if you think this is better then fine with me :)

marcoambrosini:
@JuliaKirschenheuter focus-within is not needed and causes an unwanted behavior of showing the outline on click. We don't want that. I took the liberty to push a fix. Please review @Pytal @susnux @JuliaKirschenheuter 
Thanks :)

marcoambrosini (dismissed):
blocking this until we reach a proper solution

JuliaKirschenheuter:
> @JuliaKirschenheuter focus-within is not needed and causes an unwanted behavior of showing the outline on click. We don't want that.

I have to clarify that. Probably i had mistakes in thinking.

susnux (approved):
Checked again, works!

Pytal (approved):

skjnldsv:
Please set appropriate `fix` or `enhancement` labels next time :+1:

PR #4703, opened 2023-10-27
Add header prop to NcRelatedResourcesPanel

GretaD:
### ☑️ Resolves

* Needed for https://github.com/nextcloud/contacts/issues/3364

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A
related_before | Screenshot from 2023-10-27 14-33-16

### 🚧 Tasks

- [x] There are many ways how to do this, but for now we only need the app header to be dynamic so each one of the apps can add their own.

### 🏁 Checklist

- [ ] ⛑️ Tests are included or are not applicable
- [ ] 📘 Component documentation has been extended, updated or is not applicable

PR #4705, opened 2023-10-27
fix(NcRelatedResourcesPanel): t is not defined

raimund-schluessler:
### ☑️ Resolves

* Follow-up to #4703, fixes a `ReferenceError: t is not defined` error in the docs.


GretaD (approved):
right, thank you :)

raimund-schluessler:
> right, thank you :)

Hm, the docs don't render anything for the component nonetheless. But it is like this for `stable7` as well...

ShGKme (approved):

PR #4716, opened 2023-10-30
feat(NcAppNavigationItem): Add `active` property to set active state & use `aria-current`

susnux:
### ☑️ Resolves

When not using `vue-router` it is currently not possible to set the current navigation item.
Also in both cases the `aria-current´ attribute should be set.

### 🖼️ Screenshots

image

### 🏁 Checklist


raimund-schluessler (approved):

raimund-schluessler:
Needs a rebase since #4715 is in.

ShGKme (approved):

PR #4718, opened 2023-10-31
fix(NcInputField): Bring input field height to 44px and fixes its focus feedback

marcoambrosini:
* Resolves https://github.com/nextcloud-libraries/nextcloud-vue/issues/3376

### * bring input field height to 44px and fixes its focus feedback

https://github.com/nextcloud-libraries/nextcloud-vue/assets/26852655/cdbdd6ee-f6c2-4eea-985c-beb39d447a99

### * bring select height down to 44px and improves a bit the looks of the select chips

Screenshot 2023-11-07 at 10 21 13

### DIFF
--- a/src/components/NcInputField/NcInputField.vue
+++ b/src/components/NcInputField/NcInputField.vue
@@ -423,11 +423,8 @@ export default {
 	&__label {
 		position: absolute;
 		margin-inline: 14px 0;
-		// fix height and line height to center label
-		height: 17px;
 		max-width: fit-content;
-		line-height: 1;
-		inset-block-start: 14px;
+		inset-block-start: 11px;
 		inset-inline: 0;
 		// Fix color so that users do not think the input already has content
 		color: var(--color-text-maxcontrast);
@@ -451,14 +448,12 @@ export default {
 
 	&__input:focus + &__label,
 	&__input:not(:placeholder-shown) + &__label {
-		inset-block-start: -8px;
+		inset-block-start: -10px;
 		font-size: 13px; // minimum allowed font size for accessibility
 		font-weight: 500;
 		border-radius: var(--default-grid-baseline) var(--default-grid-baseline) 0 0;
 		background-color: var(--color-main-background);
-		height: 16px;
 		padding-inline: 5px;
-		padding-block-start: 2px;
 		margin-inline-start: 9px;
 
 		transition: height var(--animation-quick), inset-block-start var(--animation-quick), font-size var(--animation-quick), color var(--animation-quick);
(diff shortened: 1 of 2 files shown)


nimishavijay (changes requested):
Nice! Definitely an improvement. Suggestion: for the tags itself, we use a background color instead of more borders:
(also looks a bit off for single select)
image

When we show buttons and inputs next to each other, the alignment still seems to be an issue. when the field is focused, technically the button and input field are center aligned, but when unfocused it looks like it is a couple of pixels too low. What do you think?

https://github.com/nextcloud-libraries/nextcloud-vue/assets/52440189/9c65a1ea-0df2-4680-b05f-7a3f22def335



Also, the bottom of the label for the input field gets cut off on focus for some resolutions, and the close button on the right is not centered (dunno if it is related to this PR)

image

marcoambrosini:
>Suggestion: for the tags itself, we use a background color instead of more borders

@nimishavijay I tried that but it results in not enough contrast with the user avatars in user selects.

JuliaKirschenheuter:
Hi @marcoambrosini,

unfortunately the issue with truncation got worse ;(

Screenshot from 2023-11-02 15-43-53

Could you please use ARCToolkit tool with "Check Text Spacing" function to check the results? 

Screenshot from 2023-11-02 15-46-14

nimishavijay:
> I tried that but it results in not enough contrast with the user avatars in user selects.

Do we need a lot of contrast there though? I imagine it would similar to the user bubble when you are @ mentioning a user on Talk. The `x` button on each user, which I am assuming is the only clickable element, would have enough contrast against the background anyway, what do you think?
image

marcoambrosini:
@nimishavijay this is very bad in talk too. I guess we need to rethink the avatars again. Without even talking about accessibility, this is not acceptable from a visual design standpoint.
Screenshot 2023-11-06 at 18 28 18

I'll remove the border here and raise an issue with the avatars

marcoambrosini:
@JuliaKirschenheuter @nimishavijay could you please review again?

susnux (changes requested):
The trailing button is not centered anymore: 
image

JuliaKirschenheuter (commented):
Trancation is gone ;)
Screenshot from 2023-11-07 11-53-58


But a placeholder is not centered
Screenshot from 2023-11-07 11-54-09


I would say we can live with it, all design improvements can be done later

JuliaKirschenheuter (approved):

jancborchardt (approved):
Agree with @JuliaKirschenheuter :) Let’s merge it as is now and @marcoambrosini go for an improvement pull request.

marcoambrosini:
@JuliaKirschenheuter are you sure that you don't have any CSS interfering with the component there? 
It looks good in the style guide


https://github.com/nextcloud-libraries/nextcloud-vue/assets/26852655/ca278c31-8288-4259-9281-7d50e3bf477d

@susnux [here's a fix for the misaligned button](https://github.com/nextcloud-libraries/nextcloud-vue/pull/4770)

PR #4728, opened 2023-10-31
Change styles for focused Select

JuliaKirschenheuter:
### ☑️ Resolves

* Fix https://github.com/nextcloud/server/issues/41217

### 🖼️ Screenshots

🏚️ Before 

Peek 2023-10-31 14-05

🏡 After

Peek 2023-10-31 14-04

### 🏁 Checklist

### DIFF
--- a/src/components/NcSelect/NcSelect.vue
+++ b/src/components/NcSelect/NcSelect.vue
@@ -1113,12 +1113,14 @@ body {
 	}
 
 	&.vs--open .vs__dropdown-toggle {
-		border-color: var(--color-primary-element);
+		outline: 2px solid var(--color-main-background);
+		border-color: var(--color-main-text);
 		border-bottom-color: transparent;
 	}
 
 	&:not(.vs--disabled, .vs--open) .vs__dropdown-toggle:hover {
-		border-color: var(--color-primary-element);
+		outline: 2px solid var(--color-main-background);
+		border-color: var(--color-main-text);
 	}
 
 	&.vs--disabled {
@@ -1149,7 +1151,7 @@ body {
 			.vs__dropdown-toggle {
 				border-radius: 0 0 var(--vs-border-radius) var(--vs-border-radius);
 				border-top-color: transparent;
-				border-bottom-color: var(--color-primary-element);
+				border-bottom-color: var(--color-main-text);
 			}
 		}
 	}
@@ -1182,8 +1184,10 @@ body {
 }
 
 .vs__dropdown-menu {
-	border-color: var(--color-primary-element) !important;
+	border-color: var(--color-main-text) !important;
+	outline: 2px solid var(--color-main-background);
 	padding: 4px !important;
+	box-shadow: none;
 
 	&--floating {
 		/* Fallback styles overidden by programmatically set inline styles */
@@ -1196,7 +1200,6 @@ body {
 			border-radius: var(--vs-border-radius) var(--vs-border-radius) 0 0 !important;
 			border-top-style: var(--vs-border-style) !important;
 			border-bottom-style: none !important;
-			box-shadow: 0px -1px 1px 0px var(--color-box-shadow) !important;
 		}
 	}
 


raimund-schluessler:
This introduces a (thin) line between the input and the options, which was not there before:
grafik

And for the `NcSelectTags` component, the border/outline seems to be offset on the right:
grafik

src/components/NcSelect/NcSelect.vue
```diff
@@ -1086,12 +1086,16 @@ body {
 	}
 
 	&.vs--open .vs__dropdown-toggle {
-		border-color: var(--color-primary-element);
+		outline: 2px solid var(--color-main-text);
+		border-color: var(--color-main-background);
+		outline-offset: -2px;
```
susnux, line 1091:
```suggestion
		outline: 2px solid var(--color-main-background);
		border-color: var(--color-main-text);
```

Should fix the issue with the "line" being shown

marcoambrosini (commented):
Nice, a couple more things: 

There seems to be a border misalignment in one of your screenshots:

Screenshot 2023-11-01 at 12 54 19

Can we remove all box shadows from the select options container? 
Screenshot 2023-11-01 at 13 07 23


Thanks :)

JuliaKirschenheuter:
> Can we remove all box shadows from the select options container?

@marcoambrosini do you mean on hover? I can recognize this background only on hover

src/components/NcSelect/NcSelect.vue
JuliaKirschenheuter: Thanks!

JuliaKirschenheuter:
> And for the `NcSelectTags` component, the border/outline seems to be offset on the right:

Dear @raimund-schluessler 

unfortunately i can't reproduce it:
Screenshot from 2023-11-02 16-40-46

could you please try again?

@marcoambrosini could you please help me a bit to find box-shadow styles?

raimund-schluessler:
> > And for the `NcSelectTags` component, the border/outline seems to be offset on the right:
> 
> Dear @raimund-schluessler
> 
> unfortunately i can't reproduce it
> 
> could you please try again?

It seems to be fixed. I don't see it anymore either. However, the border at the bottom is still blue when the select is opened to the top. See this screenshot from the docs:
grafik

JuliaKirschenheuter:
@raimund-schluessler styles of my local styleguide instance are broken ;( I guess i've fixed an issue, could you please re-review? Thanks a lot!

src/components/NcSelect/NcSelect.vue
```diff
@@ -1152,7 +1155,7 @@ body {
 }
 
 .vs__dropdown-menu {
-	border-color: var(--color-primary-element) !important;
+	border-color: var(--color-main-text) !important;
 	padding: 4px !important;
```
raimund-schluessler, line 1159:
```suggestion
	padding: 4px !important;
	box-shadow: none;
```
This would remove the `box-shadow` as @marcoambrosini suggested (ignore the misalignment of the dropdown, unrelated):

| Before | After |
|-|-|
| Screenshot 2023-11-03 at 11-46-12 Nextcloud Vue Style Guide | Screenshot 2023-11-03 at 11-46-32 Nextcloud Vue Style Guide |

raimund-schluessler:
> @raimund-schluessler styles of my local styleguide instance are broken ;( I guess i've fixed an issue, could you please re-review? Thanks a lot!

The border is fixed now. 👍

src/components/NcSelect/NcSelect.vue
JuliaKirschenheuter: Thank you!

src/components/NcSelect/NcSelect.vue
```diff
@@ -1086,12 +1086,15 @@ body {
...
+		outline: 2px solid var(--color-main-background);
+		border-color: var(--color-main-text);
 		border-bottom-color: transparent;
 	}
 
 	&:not(.vs--disabled, .vs--open) .vs__dropdown-toggle:hover {
-		border-color: var(--color-primary-element);
+		outline: 2px solid var(--color-main-text);
+		border-color: var(--color-main-background);
+		outline-offset: -2px;
```
raimund-schluessler, line 1097: I think we should switch the colors here
```suggestion
		outline: 2px solid var(--color-main-background);
		border-color: var(--color-main-text);
```
Otherwise, the `NcSelect` will appear to resize when hovering and return to initial size on opening:

Before:

https://github.com/nextcloud-libraries/nextcloud-vue/assets/2496460/a30bf155-25e7-4307-84a4-abb089f94f52

After:

https://github.com/nextcloud-libraries/nextcloud-vue/assets/2496460/787efbcd-0d16-436f-950c-221f662c2c0e

src/components/NcSelect/NcSelect.vue
```diff
@@ -1182,8 +1184,10 @@ body {
 }
 
 .vs__dropdown-menu {
-	border-color: var(--color-primary-element) !important;
+	border-color: var(--color-main-text) !important;
+	outline: 2px solid var(--color-main-background);
```
raimund-schluessler, line 1188: This outline adds a not so nice gap between the select toggle and the dropdown-menu. But I am not sure we can fix this.
grafik

More obvious with adjusted outline color
grafik

@susnux Any idea here?

JuliaKirschenheuter: well, without this `outline` it looks odd too like from your video:

https://github.com/nextcloud-libraries/nextcloud-vue/assets/2496460/787efbcd-0d16-436f-950c-221f662c2c0e

I think it is better to keep `outline`. And all other improvements we can do after a11y certification

raimund-schluessler: > well, without this `outline` it looks odd too like from your video:

That's unfortunately true. One could fix it with an additional container around the dropdown and some CSS including a `margin-top: -2px;`, but that would require adjusting `@nextcloud/vue-select` which is out out scope here.

raimund-schluessler (approved):

src/components/NcSelect/NcSelect.vue
susnux: I think the vertical position of the drop down needs to be changed. But not sure this is possible without changing  the vue-select

susnux (approved):

src/components/NcSelect/NcSelect.vue
raimund-schluessler: > I think the vertical position of the drop down needs to be changed. But not sure this is possible without changing the vue-select

I tested this (manually). It won't help, since both elements, the trigger and the dropdown, have an outline.

JuliaKirschenheuter: =(
could we merge this pr and improve left things in next  future?

marcoambrosini (approved):

src/components/NcSelect/NcSelect.vue
JuliaKirschenheuter: ok, i will merge it now, got enough approves

PR #4730, opened 2023-10-31
add scoped to all ncappnavigation components

szaimen:
Close https://github.com/nextcloud/server/issues/41153


szaimen:
This might have broken NcAppNavigationNewItem:
image

szaimen:
@ShGKme I suppose this is not the fix to the linked problem? Am I missing something?

ShGKme:
> @ShGKme I suppose this is not the fix to the linked problem? Am I missing something?

It should work if all `nextcloud/vue` on a page is updated and has `scoped`.

Is there another app that adds these styles on the page?

Scoping styles with `scoped` is only a one-way scoping, meaning it saves this component from changing other elements on the page but doesn't save this component from being overridden by styles outside.
(+1 for reliable `<style module>`)

raimund-schluessler (changes requested):
> This might have broken NcAppNavigationNewItem:

This is still an issue. Since the styles are now scoped, they need to be duplicated for `NcAppNavigationNewItem` to apply there.

susnux:
Maybe share the styles as in a scss file and import as scoped styles in the components?

src/components/NcAppNavigationItem/NcAppNavigationItem.vue
```diff
@@ -771,208 +771,6 @@ export default {
...
-// Add more contrast for active entry
-[data-themes*='highcontrast'] {
-	.app-navigation-entry {
-		&:active {
-			background-color: var(--color-primary-element-light-hover) !important;
-		}
-	}
-}
+<style scoped lang="scss">
+@import '../../assets/NcAppNavigationItem';
```
susnux, line 775: `@import` is deprecated you should use `@use`

szaimen: done

szaimen: I tried this and it didnt work: https://github.com/nextcloud-libraries/nextcloud-vue/actions/runs/6734727700/job/18306263029

susnux (approved):

JuliaKirschenheuter (approved):
works!

PR #4734, opened 2023-11-01
fix(NcTimezonePicker): Allow to set ID of inner select component

susnux:
### ☑️ Resolves

* Required for https://github.com/nextcloud/server/issues/40697

Allows to the the ID of the inner select so it can be referenced by labels.

### 🏁 Checklist

### DIFF
--- a/src/components/NcTimezonePicker/NcTimezonePicker.vue
+++ b/src/components/NcTimezonePicker/NcTimezonePicker.vue
@@ -41,7 +41,8 @@ export default {
 </docs>
 
 <template>
-	<NcSelect :value="selectedTimezone"
+	<NcSelect :uid="uid"
+		:value="selectedTimezone"
 		:options="options"
 		:multiple="false"
 		:clearable="false"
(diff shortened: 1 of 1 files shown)


src/components/NcTimezonePicker/NcTimezonePicker.vue
```diff
@@ -81,6 +82,13 @@ export default {
 			type: String,
 			default: 'floating',
 		},
+		/**
+		 * ID of the inner vue-select element, can be used for labels like: `vs-${uid}__combobox`
+		 */
+		uid: {
+			type: [String, Number],
+			default: `tz-${(Math.random() + 1).toString(36).substring(7)}`,
+		},
```
ShGKme, line 91: `default` must be a function, otherwise all instances have random but the same ID.

BTW, there is `GenRandomId`: https://github.com/nextcloud-libraries/nextcloud-vue/blob/master/src/utils/GenRandomId.js

ShGKme (changes requested):

ShGKme (approved):

Pytal (approved):

PR #4736, opened 2023-11-01
fix(NcSidebarTab): Ensure there is an `aria-label` if the `aria-labelledby` element does not exist

susnux:
### ☑️ Resolves

* Fix https://github.com/nextcloud-libraries/nextcloud-vue/issues/3032

If only one sidebar tab is shown, the tab header is missing. Thus we can not reference it for `aria-labelledby` and have to use `aria-label` instead.

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppSidebar/NcAppSidebarTabs.vue
+++ b/src/components/NcAppSidebar/NcAppSidebarTabs.vue
@@ -36,8 +36,8 @@
 			@keydown.tab.exact.prevent.stop="focusActiveTabContent"
 			@keydown.home.exact.prevent.stop="focusFirstTab"
 			@keydown.end.exact.prevent.stop="focusLastTab"
-			@keydown.33.exact.prevent.stop="focusFirstTab"
-			@keydown.34.exact.prevent.stop="focusLastTab">
+			@keydown.page-up.exact.prevent.stop="focusFirstTab"
+			@keydown.page-down.exact.prevent.stop="focusLastTab">
 			<NcCheckboxRadioSwitch v-for="tab in tabs"
 				:key="tab.id"
 				:aria-controls="`tab-${tab.id}`"
@@ -90,6 +90,8 @@ export default {
 			unregisterTab: this.unregisterTab,
 			// Getter as an alternative to Vue 2.7 computed(() => this.activeTab)
 			getActiveTab: () => this.activeTab,
+			// Used to check whether the tab header is shown so the tabs can reference the tab header for `aria-labelledby` or not
+			isTablistShown: () => this.hasMultipleTabs,
 		}
 	},
 
--- a/src/components/NcAppSidebarTab/NcAppSidebarTab.vue
+++ b/src/components/NcAppSidebarTab/NcAppSidebarTab.vue
@@ -27,7 +27,8 @@
 	<section :id="`tab-${id}`"
 		:class="{'app-sidebar__tab--active': isActive}"
 		:aria-hidden="!isActive"
-		:aria-labelledby="`tab-button-${id}`"
+		:aria-label="isTablistShown() ? undefined : name"
+		:aria-labelledby="isTablistShown() ? `tab-button-${id}` : undefined"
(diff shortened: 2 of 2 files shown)


src/components/NcAppSidebar/NcAppSidebarTabs.vue
```diff
@@ -90,6 +90,8 @@ export default {
 			unregisterTab: this.unregisterTab,
 			// Getter as an alternative to Vue 2.7 computed(() => this.activeTab)
 			getActiveTab: () => this.activeTab,
+			// Used to check whether the tab header is shown so the tabs can reference the tab header for `aria-labelledby` or not
+			tablistShown: this.hasMultipleTabs,
```
ShGKme, line 94: This provide is not reactive. We need to provide a getter or a `computed`.

susnux: Is `hasMultipleTabs` not a computed?

susnux: Or do you mean it is required to be like this:
```
tablistShown: () => this.hasMultipleTabs
```

(`hasMultipleTabs` is a computed property)

ShGKme: > Is `hasMultipleTabs` not a computed?

This is a computed value, but not a computed reference object.

From JS side, it is just a boolean value `true/false`, so "provide" provides the value "true" by key "tablistShown". We need to provide a reference to a reactive mutable object or a computed ref.

> `tablistShown: () => this.hasMultipleTabs`

This will work in any Vue, but must be used as a function

`getTablistShown: () => this.hasMultipleTabs`

Or it can be computed ref:
`tablistShown: computed(() => this.hasMultipleTabs)`, but Raimund had a problem with this solution earlier (though, it should work in Vue 2.7+)

susnux: Ah I see Option API computed != reactive computes from Composition API :)

>  tablistShown: computed(() => this.hasMultipleTabs), but Raimund had a problem with this solution earlier (though, it should work in Vue 2.7+)

I like this one more, but looks also a bit like mixing Options API and Composition API.

ShGKme: > I like this one more, but looks also a bit like mixing Options API and Composition API.

Though Vue 3 reactivity with `reactive`, `ref`, `computed` was created for Composition API, it is not Composition API itself, not mixing. There is still no `setup` in the component, so no component is described with Composition API. This is just a reactivity function and a recommended way to provide reactive values in Vue 3 (if not count new toRef from 3.3 update).

https://vuejs.org/guide/components/provide-inject.html#working-with-reactivity

In Vue 3 there is even unwrapping for injected refs/computed in Options API.

src/components/NcAppSidebar/NcAppSidebarTabs.vue
```diff
@@ -90,6 +90,8 @@ export default {
 			unregisterTab: this.unregisterTab,
 			// Getter as an alternative to Vue 2.7 computed(() => this.activeTab)
 			getActiveTab: () => this.activeTab,
+			// Used to check whether the tab header is shown so the tabs can reference the tab header for `aria-labelledby` or not
+			tablistShown: () => this.hasMultipleTabs,
```
ShGKme, line 94: If we provide a getter function, I'd explicitly call it as a function, as `getActiveTab` above

```suggestion
			// Used to check whether the tab header is shown so the tabs can reference the tab header for `aria-labelledby` or not
			getTablistShown: () => this.hasMultipleTabs,
```

susnux: :+1:

ShGKme (approved):

AndyScherzinger (approved):

JuliaKirschenheuter (approved):

PR #4737, opened 2023-11-01
fix(NcAvatar): Improve initials generation to filter out special characters

susnux:
### ☑️ Resolves

* Fix https://github.com/nextcloud-libraries/nextcloud-vue/issues/2044

1. Filter user names to only use real characters (for all languages) to generate initials and not using special characters like e.g. `(`.
2. Use first & last name for initials instead of second name (e.g. `First Second Last` should be `FL` and not `FS`).

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image|image
image|image

### 🏁 Checklist

### DIFF
--- a/tests/unit/components/NcAvatar/NcAvatar.spec.ts
+++ b/tests/unit/components/NcAvatar/NcAvatar.spec.ts
@@ -20,7 +20,7 @@
  *
  */
 
-import { mount } from '@vue/test-utils'
+import { mount, shallowMount } from '@vue/test-utils'
 import { nextTick } from 'vue'
 import NcAvatar from '../../../../src/components/NcAvatar/NcAvatar.vue'
 
@@ -86,5 +86,49 @@ describe('NcAvatar.vue', () => {
 
 		expect(wrapper.find('.avatardiv__user-status').exists()).toBe(false)
 		expect(wrapper.attributes('aria-label')).toBe('Avatar of J. Doe')
+	})
+
+	it('should display initials for user id', async () => {
+		const wrapper = shallowMount(NcAvatar, {
+			propsData: {
+				user: 'Jane Doe',
+				isNoUser: true,
+			},
+		})
+		await nextTick()
+		expect(wrapper.text()).toBe('JD')
+	})
+
+	it('should display initials for display name property over user id', async () => {
+		const wrapper = shallowMount(NcAvatar, {
+			propsData: {
+				displayName: 'No User',
+				user: 'I am a group',
+				isNoUser: true,
+			},
+		})
+		await nextTick()
+		expect(wrapper.text()).toBe('NU')
+	})
+
+	describe('Fallback initials', () => {
+		it.each`
+			displayName             | initials | case
+			${''}                   | ${'?'}   | ${'empty user'}
+			${'Jane Doe'}           | ${'JD'}  | ${'display name property'}
+			${'Jane (Doe)'}         | ${'JD'}  | ${'special characters in name'}
+			${'jane doe'}           | ${'JD'}  | ${'lower case name'}
+			${'Jane Some Name Doe'} | ${'JD'}  | ${'middle names'}
+			${'Ümit Öçal'}          | ${'ÜÖ'}  | ${'non ascii characters'}
+			${'ジェーン ドー'}        | ${'ジド'} | ${'non latin characters'}
+		`('should display initials for $case ("$displayName" -> "$initials")', async ({ displayName, initials }) => {
+			const wrapper = shallowMount(NcAvatar, {
+				propsData: {
+					displayName,
+				},
+			})
+			await nextTick()
+			expect(wrapper.text()).toBe(initials)
+		})
 	})
 })
--- a/src/components/NcAvatar/NcAvatar.vue
+++ b/src/components/NcAvatar/NcAvatar.vue
@@ -173,10 +173,10 @@ export default {
 			v-bind="userStatusRole" />
 
 		<!-- Show the letter if no avatar nor icon class -->
-		<span v-if="userDoesNotExist && !(iconClass || $slots.icon)"
+		<span v-if="showInitials"
 			:style="initialsWrapperStyle"
 			class="avatardiv__initials-wrapper">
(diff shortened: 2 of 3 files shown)


tests/unit/components/NcAvatar/NcAvatar.spec.ts
```diff
@@ -20,11 +20,102 @@
...
+					propsData: {
+						...props,
+					},
+				})
+				expect(wrapper.exists()).toBe(true)
+
+				await nextTick()
+				const initials = wrapper.find('.avatardiv__initials')
+				expect(initials.exists()).toBe(true)
+				expect(initials.text()).toBe(expected)
```
ShGKme, line 114: IMO, it is not good to test span by class name `.avatardiv__initials` here, it is not a component interface. As all the avatar is a block with initials, it seems, we can just check the text in the avatar.

And is there a case where a wrapper could not exist in VTU?

```suggestion
				await nextTick()
				expect(wrapper.text()).toBe(expected)
```

tests/unit/components/NcAvatar/NcAvatar.spec.ts
```diff
@@ -20,11 +20,102 @@
...
+			})
+		}
+	})
+
```
ShGKme, line 118: These tests work fine, but manual generation of tests with a loop decrease DX a bit, for example, with testing GUI.

In WebStorm it doesn't work to go from test results to the test source, it cannot locate the generated test.

Both Jets and Vitest have `it.each` method to make tests by test cases, which seems more explicit to me.

With objects:

```suggestion
		it.each([
			{
				case: 'empty user',
				displayName: '',
				initials: '?',
			},
			{
				case: 'display name property',
				displayName: 'Jane Doe',
				initials: 'JD',
			},
			{
				case: 'special characters in name',
				displayName: 'Jane (Doe)',
				initials: 'JD',
			},
			{
				case: 'lower case name',
				displayName: 'jane doe',
				initials: 'JD',
			},
			{
				case: 'middle names',
				displayName: 'Jane Some Name Doe',
				initials: 'JD',
			},
			{
				case: 'non ascii characters',
				displayName: 'Ümit Öçal',
				initials: 'ÜÖ',
			},
			{
				case: 'non latin characters',
				displayName: 'ジェーン ドー',
				initials: 'ジド',
			},
		])('should display initials for $case ("$displayName" -> "$initials")', async ({ displayName, initials }) => {
			const wrapper = shallowMount(NcAvatar, {
				propsData: {
					displayName,
				},
			})
			await nextTick()
			expect(wrapper.text()).toBe(initials)
		})

		it('should display initials for user id', async () => {
			const wrapper = shallowMount(NcAvatar, {
				propsData: {
					user: 'Jane Doe',
					isNoUser: true,
				},
			})
			await nextTick()
			expect(wrapper.text()).toBe('JD')
		})

		it('should display initials for display name property over user id', async () => {
			const wrapper = shallowMount(NcAvatar, {
				propsData: {
					displayName: 'No User',
					user: 'I am a group',
					isNoUser: true,
				},
			})
			await nextTick()
			expect(wrapper.text()).toBe('NU')
		})
```

ShGKme: Or with a template table (I'd prefer this option as more beautiful):

```suggestion
it.each`
			displayName             | initials | case
			${''}                   | ${'?'}   | ${'empty user'}
			${'Jane Doe'}           | ${'JD'}  | ${'display name property'}
			${'Jane (Doe)'}         | ${'JD'}  | ${'special characters in name'}
			${'jane doe'}           | ${'JD'}  | ${'lower case name'}
			${'Jane Some Name Doe'} | ${'JD'}  | ${'middle names'}
			${'Ümit Öçal'}          | ${'ÜÖ'}  | ${'non ascii characters'}
			${'ジェーン ドー'}        | ${'ジド'} | ${'non latin characters'}
		`('should display initials for $case ("$displayName" -> "$initials")', async ({ displayName, initials }) => {
			const wrapper = shallowMount(NcAvatar, {
				propsData: {
					displayName,
				},
			})
			await nextTick()
			expect(wrapper.text()).toBe(initials)
		})

		it('should display initials for user id', async () => {
			const wrapper = shallowMount(NcAvatar, {
				propsData: {
					user: 'Jane Doe',
					isNoUser: true,
				},
			})
			await nextTick()
			expect(wrapper.text()).toBe('JD')
		})

		it('should display initials for display name property over user id', async () => {
			const wrapper = shallowMount(NcAvatar, {
				propsData: {
					displayName: 'No User',
					user: 'I am a group',
					isNoUser: true,
				},
			})
			await nextTick()
			expect(wrapper.text()).toBe('NU')
		})
```

ShGKme: In any case, I'd move those 2 tests from the loop over test cases, because they test different cases, not the initials generation algorithm.

ShGKme (approved):
Everything is good from the component side, but I'd discuss some changes in tests.

tests/unit/components/NcAvatar/NcAvatar.spec.ts
ShGKme: Links:
- https://jestjs.io/docs/api#testeachtablename-fn-timeout
- https://vitest.dev/api/#test-each

tests/unit/components/NcAvatar/NcAvatar.spec.ts
susnux: I agree better are test selectors like `data-testid` but in this case nothing is needed :)

tests/unit/components/NcAvatar/NcAvatar.spec.ts
susnux: Thank you! Did not know that something like this exists :see_no_evil: 
I also like the table style

ChristophWurst (approved):
Super nice!

If Nextcloud's back-end also still generates avatars we should at least create a ticket to also improve the handling there. This avoid inconsistencies between avatars generated on the front or the back.

raimund-schluessler (approved):

PR #4739, opened 2023-11-01
[next] feat(NcUserBubble): migrate to vue 3

raimund-schluessler:
### ☑️ Resolves

* This migrates `NcUserBubble` to vue 3.

### DIFF
--- a/src/components/NcUserBubble/NcUserBubble.vue
+++ b/src/components/NcUserBubble/NcUserBubble.vue
@@ -77,7 +77,7 @@ export default {
 
 </docs>
 <template>
-	<component :is="isPopoverComponent"
+	<component :is="!!$slots.default ? 'NcPopover' : 'NcUserBubbleDiv'"
 		trigger="hover focus"
 		:shown="open"
 		class="user-bubble__wrapper"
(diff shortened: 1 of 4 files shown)


susnux (approved):

susnux:
Works but noticed an error (we also have this on master so unrelated^^)

src/components/NcUserBubble/NcUserBubble.vue
```diff
@@ -91,7 +101,7 @@ This component has the following slot:
 					:style="styles.avatar"
 					:disable-tooltip="true"
 					:disable-menu="true"
-					v-bind="$props"
+					v-bind="$attrs"
```
ShGKme, line 104: I haven't tested this specific place, but this looks like a behavior change. Is it migration-related?

For example, prop `showUserStatus` prop is unused now.

Probably, we need to fix this on `main` first

raimund-schluessler: Indeed, that's a bit inconsistent. I changed it now to explicitly only forward the props that `NcAvatar` uses. If this is ok, I would send a PR to `master` as well.

raimund-schluessler: We should anyway wait until https://github.com/nextcloud-libraries/nextcloud-vue/pull/4743 is merged and backported before continuing here.

raimund-schluessler: See #4756 for the fix for `master`.

raimund-schluessler: With #4756 merged into `next`, the problem should be solved.

ShGKme (approved):

PR #4740, opened 2023-11-01
[next] fix(NcAvatar): don't show `false` for disabled tooltip

raimund-schluessler:
### ☑️ Resolves

* Fixes showing `false` as tooltip when the tooltip was disabled with `disable-tooltip`.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Bildschirmfoto vom 2023-11-01 21-45-19  | grafik

### DIFF
--- a/src/components/NcAvatar/NcAvatar.vue
+++ b/src/components/NcAvatar/NcAvatar.vue
@@ -474,7 +474,7 @@ export default {
 		},
 		tooltip() {
 			if (this.disableTooltip) {
-				return false
+				return null
 			}
 			if (this.tooltipMessage) {
 				return this.tooltipMessage


ShGKme (approved):

src/components/NcAvatar/NcAvatar.vue
```diff
@@ -474,7 +474,7 @@ export default {
 		},
 		tooltip() {
 			if (this.disableTooltip) {
-				return false
+				return null
```
susnux, line 477: Would make `undefined` not make more sense?

susnux (approved):
LGTM just a question :)

src/components/NcAvatar/NcAvatar.vue
raimund-schluessler: I have no preference here tbh. Both works. If you prefer `undefined`, that's fine with me.

susnux: No just a question :)

PR #4743, opened 2023-11-01
fix: NcUserBubble properties

### LINKED ISSUE #792: Make displayName optional
Must this be required? displayName gets replaced by user, if not set. 

https://github.com/nextcloud/nextcloud-vue/blob/76b45d870723810e21df367825e9bb3ddf38e2db/src/components/UserBubble/UserBubble.vue#L89-L92

susnux:
### ☑️ Resolves

* Fix #792 

1. Make `displayName` not required
    We already have a fallback so it is not required.
    But we should show a warning when user and display name are not set.
2. Fix URL prop validator to allow relative URLs
3. Fix examples
    * Examples does not need to pass computed property
    * `alert` is not defined in templates for Vue 2 (same with console)

### 🏁 Checklist

### DIFF
--- a/src/components/NcUserBubble/NcUserBubble.vue
+++ b/src/components/NcUserBubble/NcUserBubble.vue
@@ -123,6 +123,7 @@ export default {
 import NcUserBubbleDiv from './NcUserBubbleDiv.vue'
 import NcAvatar from '../NcAvatar/index.js'
 import NcPopover from '../NcPopover/index.js'
+import Vue from 'vue'
 
 export default {
 	name: 'NcUserBubble',
@@ -151,7 +152,7 @@ export default {
 		 */
 		displayName: {
 			type: String,
-			required: true,
+			default: undefined,
 		},
 		/**
 		 * Whether or not to display the user-status
@@ -277,6 +278,11 @@ export default {
 				},
 			}
 		},
+	},
+	mounted() {
+		if (!this.displayName && !this.user) {
+			Vue.util.warn('[NcUserBubble] At least `displayName` or `user` property should be set.')
+		}
 	},
 	methods: {
 		onOpenChange(state) {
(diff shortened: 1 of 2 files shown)


ShGKme (approved):

src/components/NcUserBubble/NcUserBubble.vue
```diff
@@ -158,10 +167,10 @@ export default {
 		url: {
 			type: String,
 			default: undefined,
-			validator: url => {
+			validator: (url) => {
 				try {
-					url = new URL(url)
-					return !!url
+					url = new URL(url, url?.startsWith?.('/') ? window.location.href : undefined)
```
Pytal, line 172:
```suggestion
					url = new URL(url, url?.startsWith?.('/') ? window.location.origin : undefined)
```
Links are relative to the origin in my testing

susnux: I do not think this is correct, e.g. Current page is `http://foo.bar/a/b.php` and the relative url is:
1. `/c`
2. `./c`

Using `origin` will result:
1. `http://foo.bar/c`
2. `http://foo.bar/c`

Using `href` will result:
1. `http://foo.bar/c`
2. `http://foo.bar/a/c`

So I would say `href` is correct as otherwise the relative URL `./c` would resolve to `root/c` instead of `/a/c`.

skjnldsv (approved):

src/components/NcUserBubble/NcUserBubble.vue
Pytal: I believe new URL() will work either way though the difference would be checking whether `url?.startsWith?.('/')` or `url?.startsWith?.('./')`, but in this case I don't think it matters

Pytal (approved):

src/components/NcUserBubble/NcUserBubble.vue
susnux: Yes but I am not sure if `http://foo.bar/../c` is valid

PR #4744, opened 2023-11-01
feat(NcActionButton): Allow pressed state on NcActionButton - similar to NcButton

susnux:
### ☑️ Resolves

* Fix https://github.com/nextcloud-libraries/nextcloud-vue/issues/425

NcButton already allows a checked state so this introduces this for the NcActionButton.
In checked state the `aria-pressed` attribute will be added and a primary border will indicate it visually.

### 🖼️ Screenshots

https://github.com/nextcloud-libraries/nextcloud-vue/assets/1855448/c90a7c12-e663-4cd6-8abd-5d3848088de1

### 🏁 Checklist

### DIFF
--- a/src/components/NcActionButton/NcActionButton.vue
+++ b/src/components/NcActionButton/NcActionButton.vue
@@ -171,18 +171,51 @@ export default {
 </script>
 ```
 
-You can set an "pressed" state, for example you have a toggle button, in this example the "fullscreen" button:
+### With different model behavior
+By default the button will act like a normal button, but it is also possible to change the behavior to a toggle button, checkbox button or radio button.
+
+For example to have the button act like a toggle button just set the `modelValue` property to the toggle state:
 
 ```vue
 <template>
 	<NcActions>
-		<NcActionButton :pressed="handRaised" @click="handRaised = !handRaised">
+		<NcActionButton :model-value.sync="fullscreen">
+			<template #icon>
+				<Fullscreen :size="20" />
+			</template>
+			Fullscreen
+		</NcActionButton>
+	</NcActions>
+</template>
+<script>
+import Fullscreen from 'vue-material-design-icons/Fullscreen.vue'
+
+export default {
+	components: {
+		Fullscreen,
+	},
+	data() {
+		return {
+			fullscreen: true,
+		}
+	},
+}
+</script>
+```
+
+Another example would be using it with checkbox semantics, to enable or disable features.
+This also allows tri-state behavior (`true`, `false`, `null`) in which case `aria-checked` will be either `true`, `false` or `mixed`.
+
+```vue
+<template>
+	<NcActions>
+		<NcActionButton :model-value.sync="handRaised" model-behavior="checkbox">
 			<template #icon>
 				<HandBackLeft :size="20" />
 			</template>
 			Raise hand
 		</NcActionButton>
-		<NcActionButton :pressed="fullscreen" @click="fullscreen = !fullscreen">
+		<NcActionButton :model-value.sync="fullscreen" model-behavior="checkbox">
 			<template #icon>
 				<Fullscreen :size="20" />
 			</template>
@@ -208,21 +241,56 @@ export default {
 }
 </script>
 ```
+
+It is also possible to use the button with radio semantics, this is only possible in menus and not for inline actions!
+
+```vue
+<template>
+	<NcActions>
+		<NcActionButton :model-value.sync="payment" model-behavior="radio" value="cash">
+			<template #icon>
+				<Cash :size="20" />
+			</template>
+			Pay with cash
+		</NcActionButton>
+		<NcActionButton :model-value.sync="payment" model-behavior="radio" value="card">
+			<template #icon>
+				<CreditCard :size="20" />
+			</template>
+			Pay by card
+		</NcActionButton>
+	</NcActions>
+</template>
+<script>
+import Cash from 'vue-material-design-icons/Cash.vue'
+import CreditCard from 'vue-material-design-icons/CreditCard.vue'
+
+export default {
+	components: {
+		Cash,
+		CreditCard,
+	},
+	data() {
+		return {
+			payment: 'card',
+		}
+	},
+}
+</script>
+```
 </docs>
 
 <template>
 	<li class="action" :class="{ 'action--disabled': disabled }" :role="isInSemanticMenu && 'presentation'">
-		<button class="action-button button-vue"
-			:class="{
-				'action-button--active': pressed,
+		<button :aria-label="ariaLabel"
+			:class="['action-button button-vue', {
+				'action-button--active': isChecked,
 				focusable: isFocusable,
-			}"
-			:aria-label="ariaLabel"
-			:aria-pressed="active"
+			}]"
 			:title="title"
-			:role="isInSemanticMenu && 'menuitem'"
 			type="button"
-			@click="onClick">
+			v-bind="buttonAttributes"
+			@click="handleClick">
 			<!-- @slot Manually provide icon -->
 			<slot name="icon">
 				<span :class="[isIconUrl ? 'action-button__icon--url' : icon]"
@@ -255,8 +323,8 @@ export default {
 
 			<!-- right arrow icon when there is a sub-menu -->
 			<ChevronRightIcon v-if="isMenu" :size="20" class="action-button__menu-icon" />
-			<CheckIcon v-else-if="pressed === true" :size="20" class="action-button__pressed-icon" />
-			<span v-else-if="pressed === false" class="action-button__pressed-icon material-design-icon" />
+			<CheckIcon v-else-if="isChecked === true" :size="20" class="action-button__pressed-icon" />
+			<span v-else-if="isChecked === false" class="action-button__pressed-icon material-design-icon" />
 
 			<!-- fake slot to gather inner text -->
 			<slot v-if="false" />
@@ -294,7 +362,7 @@ export default {
 		 * @todo Add a check in @nextcloud/vue 9 that this prop is not provided,
 		 * otherwise root element will inherit incorrect aria-hidden.
 		 */
-		 ariaHidden: {
+		ariaHidden: {
 			type: Boolean,
 			default: null,
 		},
@@ -317,11 +385,30 @@ export default {
 		},
 
 		/**
-		 * The pressed state of the button if it has a checked state
-		 * This will add the `aria-pressed` attribute and for the button to have a primary border in checked state.
+		 * The button's behavior, by default the button acts like a normal button with optional toggle button behavior if `modelValue` is `true` or `false`
+		 * But you can also set to checkbox button behavior with tri-state or radio button like behavior.
 		 */
-		 pressed: {
-			type: Boolean,
+		modelBehavior: {
+			type: String,
+			default: 'button',
+			validator: (behavior) => ['checkbox', 'radio', 'button'].includes(behavior),
+		},
+
+		/**
+		 * The buttons state if `modelBehavior` is 'checkbox' or 'radio' (meaning if it is pressed / selected)
+		 * Either boolean for checkbox and toggle button behavior or `value` for radio behavior
+		 */
+		modelValue: {
+			type: [Boolean, String],
+			default: null,
+		},
+
+		/**
+		 * The value used for the `modelValue` when this component is used with radio behavior
+		 * Similar to the `value` attribute of `<input type="radio">`
+		 */
+		value: {
+			type: String,
 			default: null,
 		},
 	},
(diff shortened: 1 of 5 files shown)


raimund-schluessler:
I think we went a bit away from showing the active state with a border at the left, no? But if it's approved, fine with me.

jancborchardt (dismissed):
Yep, that’s correct @raimund-schluessler – the specs in the issue are quite old and from a time where we used that style.

Since we are generally close to Material Design, what they do is show a checkmark on the right – see "list item trailing icon" at https://m3.material.io/components/menus/specs
Then we also do not have any issues with contrast, regardless of the color.

Also, not sure if related, but in that example there is a bit too much whitespace below the last entry. It should be exactly the same as on the left and right (4px likely), but seems more.

susnux:
> Also, not sure if related, but in that example there is a bit too much whitespace below the last entry. It should be exactly the same as on the left and right (4px likely), but seems more.

No this is unrelated and a different issue.

> Since we are generally close to Material Design, what they do is show a checkmark on the right

I am not sure about this one this will either be very "flashy" as the checkmark will disappear on disable and then the menu resizes or we add a quite huge margin for all button on the right side.
Maybe we can go with "activated" state of material design? This would be even somewhat consistent with NcButton which is primary on "activated", ref: https://m3.material.io/foundations/interaction/states/applying-states#f7145f12-7c29-4870-b587-b638833b2061

Example:

https://github.com/nextcloud-libraries/nextcloud-vue/assets/1855448/a574e040-c8b3-4a17-9623-05115f9072f8

susnux:
With an icon but without margin it would look like this:

https://github.com/nextcloud-libraries/nextcloud-vue/assets/1855448/81c8136f-90a2-4b3d-b6ac-bc8796355900

And with additional margin on "un-pressed":

https://github.com/nextcloud-libraries/nextcloud-vue/assets/1855448/18f5d43a-c54d-4ef2-ad13-34efe6e533f6

jancborchardt:
Thanks for the examples @susnux! :) I think the variant with the checkmark and the additional marking of course (so it doesn’t jump) is the best one.

The background color change wouldn’t pass contrast requirements unless we change it to primary-element background with white text – which we could also do, just like in Nextcloud Office or Text.

susnux:
> The background color change wouldn’t pass contrast requirements unless we change it to primary-element background 

@jancborchardt no this is fine, I tested that combination (the PR on server were we adjusted the colors so that common combinations work find (here it is primary-element-light + primary-element-light-text)

I am fine with the icon, the only down side is it can not be used in the button group example (see video)

raimund-schluessler:
We could do both. An icon and the background change.

JuliaKirschenheuter:
Thank you @susnux!

My impression regarding contrasts on selected / pressed button: if there is no other indicator (like check icon) then contrast have to be 3:1 (focused <-> unfocused, focused <-> surrounding background) like it was here https://github.com/nextcloud/text/issues/3860

Could we use another outline or background color?

susnux:
What is the suggested active state? Border does not work because it will be overridden by the focus state, as focus uses the border instead of outline. So only the fill color or icon is possible.
I think primary will be a bit too much?

JuliaKirschenheuter:
Ok, we can move with a "checked" icon. But then we have to notify a screen reader that something happened (checked, unchecked) via `aria-label` i think.

JuliaKirschenheuter:
> This is already done by `aria-pressed`. This PR is about the visual representation of it.

_Adding aria-pressed to an element with a role of [button](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role) turns the button into a toggle button. The aria-pressed attribute is only relevant for toggle buttons._

not sure that `aria-pressed` is ok for this case. Do we have a real toggle button inn that case?

susnux:
Yes this is about toggle button, like toggle between full screen view and windowed. Or toggle left align etc.

JuliaKirschenheuter:
> Yes this is about toggle button, like toggle between full screen view and windowed. Or toggle left align etc.

I've just asked myself because it is like a normal button inside a "normal" menu. Let's leave it like toggle button 👍

JuliaKirschenheuter (approved):
fine to me!

ShGKme:
What should it be used instead of `NcActionCheckbox`?

susnux:
> What should it be used instead of NcActionCheckbox?

I think from UX / design those are two different actions:
Checkbox / Radio is for settings while toggle button is for actions that have a state.

E.g. for the full screen example which triggers an action (switch to fullscreen) or text align.

ShGKme:
Maybe personal info visibility setting is a relevant example from the design (if we don't implement them as a radio button).

image

src/components/NcActionButtonGroup/NcActionButtonGroup.vue
```diff
@@ -29,19 +29,22 @@ This should be used sparingly for accessibility.
...
+					:pressed="alignment === 'c'"
+					@click="alignment = 'c'">
 					<template #icon>
 						<AlignCenter :size="20" />
 					</template>
 				</NcActionButton>
 				<NcActionButton aria-label="Align right"
-					@click="showMessage('Align Right')">
+					:pressed="alignment === 'r'"
+					@click="alignment = 'r'">
```
ShGKme, line 47: This example represents a radio group (only one of the buttons can be selected). But with `aria-pressed` there is no semantic of a radio group. With a screen reader, a user can hear whether "Aligned center" is pressed or not pressed, but not that there is a group of 3 possible alignments

According to the [specification](https://w3c.github.io/aria/#aria-pressed), it is supposed to be used on a toggle button

susnux: Maybe we need role=group on the button group component?

ShGKme: `role=group` should be added (doing it in another PR), but it doesn't add radio semantics to the group, "only one of options is selected".

susnux: But this is usage dependent if you want radio semantics or not, for this PR it is supporting toggle-buttons.
Having button groups as radio buttons is another topic (well part of the example here...). Because that would need changes on the button group component.

susnux: So for the moment I would like to got with this one (for `NcActionButton`) and do a follow up for `NcActionButtonGroup`. As this PR does not change anything (related to a11y) for the later one.

ShGKme: This PR adds a new prop for a new feature, which cannot be undone or changed much without a breaking change. That's why I wouldn't rush here without the need.

ShGKme:
I think, this can be useful for `NcActions` but in a bit different way to cover all the cases and a11y requirements. @susnux what do you think?

## Current usage

`NcActionButton` can be used in 3 ways:
1. An **inline** button like a normal button.
   - This could be a toggle button, so **pressed** state makes sense with `aria-pressed`
   - It is rendered as `NcButton` that already have a `pressed` feature.
     We only need a way to set `NcButton.pressed` via `NcActionButton`
   - Example - toolbar in text`. **Bold** and _Italic_ are toggle buttons
     image
2. In a menu as **a menu item** button with **radio behavior**
   - We have `NcActionRadio` already that represents a `role="menuitemradio"` and manages `aria-cheched`
   - But we want to have it in a button design, like in this PR
   - ? A problem here - how to determine a group ?
   - Examples: 
     Personal Settings: visibility scope | Text menu bar: callout, heading options
     ---|---
     image | image
2. In a menu as **a menu item** button with **checkbox behavior**
   - Same as p.2 but with a **menuitemcheckbox** role
   - No issues with making it a group
   - Examples: none
4. In a `NcActionButtonGroup`
   - Same as p.2 and p.3 but in a group
   - No problem with grouping a radio
   - Example - Table cell alignment in `text`:
     image
5. In a popup as a button in a dialog
   - Same as p.2 and p.3 but when it is not in a `role="menu"`
   - Examples: none

The most problematic part in my opinion - **menu item with a radio behavior**. Because there is no way to specify a group. My idea - allow a menu to have no more than one radio group (use `NcActionButtonGroup` to create a group). In future we can add horizontal layout to `NcActionButtonGroup`).

## Proposal

1. Add a new prop that represents pressed/checked state. For example, `selected` to be more universal and not associated with a specific type nor HTML attribute. Or `modelValue` =D
2. Add a new prop that defines if it is a radio or checkbox (= toggle button). Like `type` in `NcCheckboxRadioSwitch`. For example, `behavior: 'toggle' | 'radio'`.
2. Use it to define:
   - `NcButton`'s `pressed` prop in **inline buttons**
   - `role="menuitemcheckbox"`, `role="menuitemradio"` with `aria-checked` in menus with this new design
   - `aria-pressed` when used not in a menu with the same new design

## Alternative solution

1. Add button-like design to `NcActionCheckbox` and `NcActionRadio`, but it seems more complicated to me.
2. Add `NcActionButtonCheckbox` and `NcActionButtonRadio` to wrap `NcActionButton` instead of a new prop.

susnux:
@ShGKme I fully agree with your point and I think your proposal is the best way to implement this.
I decided to also support a real `value` for the radio semantics, this has some benefits like making it easier to interact with the component `:modelValue.sync` instead of `@click="state = 'x'"`.

Please feel free to review :)

src/components/NcActionButton/NcActionButton.vue
```diff
@@ -269,6 +381,34 @@ export default {
...
+			type: String,
+			default: 'button',
+			validator: (behavior) => ['checkbox', 'radio', 'button'].includes(behavior),
+		},
+
+		/**
+		 * The buttons state if `modelBehavior` is 'checkbox' or 'radio' (meaning if it is pressed / selected)
+		 * @type {boolean|any} Either boolean for checkbox and toggle button behavior or `radioValue` for radio behavior
+		 */
+		// eslint-disable-next-line vue/require-prop-types
```
susnux, line 399: Is there a proper way to set `type` to `any`?

ShGKme: I think it shouldn't be `any`. Like in native radio/checkboxes. Either `boolean` or `string`.

src/components/NcActionButton/NcActionButton.vue
```diff
@@ -269,6 +381,34 @@ export default {
...
+
+		/**
+		 * The button's behavior, by default the button acts like a normal button with optional toggle button behavior if `modelValue` is `true` or `false`
+		 * But you can also set to checkbox button behavior with tri-state or radio button like behavior.
+		 */
+		modelBehavior: {
+			type: String,
+			default: 'button',
+			validator: (behavior) => ['checkbox', 'radio', 'button'].includes(behavior),
+		},
```
ShGKme, line 393: **thought:**

For me, it is not clear what is the difference between `checkbox` and `button` if both supports `modelValue`.

src/components/NcActionButton/NcActionButton.vue
```diff
@@ -280,6 +420,66 @@ export default {
...
+				type: 'button',
+			}
+
+			if (this.isInSemanticMenu) {
+				// By default it needs to be a menu item in semantic menus
+				attributes.role = 'menuitem'
+
+				if (this.modelBehavior === 'radio') {
+					attributes.role = 'menuitemradio'
+					attributes['aria-checked'] = this.isChecked ? 'true' : 'false'
```
ShGKme, line 450: **proposal:** to make it a little bit simple to understand what the value is here.
`isChecked` is defined for all cases (with condition), but actually is used only here. And when it is used only here, it is as simple as one condition.

```suggestion
					attributes['aria-checked'] = this.modelValue === this.radioValue ? 'true' : 'false'
```

src/components/NcActionButton/NcActionButton.vue
ShGKme: If I understand correctly, the only difference is that `checkbox` support explicit `null` `modelValue`.

src/components/NcActionButton/NcActionButton.vue
susnux: Ok so like NcCheckboxRadioSwitch

src/components/NcActionButton/NcActionButton.vue
ShGKme: **thought:**

Together with `radioValue`, "string-value" behavior makes me feel that the component is overloaded with long-named props on the minimal usage of a radio:

```html
<NcActionButton :model-value.sync="checked" radio-value="left" model-behavior="radio" />
<NcActionButton :model-value.sync="checked" radio-value="middle" model-behavior="radio" />
<NcActionButton :model-value.sync="checked" radio-value="right" model-behavior="radio" />
```

What about making `radio-value` just `value` and `model-behavior="radio"`/model- behavior="checkbox" simplify to bool props `radio`/`checkbox` flags:

```html
<NcActionButton :model-value.sync="checked" value="left" radio />
<NcActionButton :model-value.sync="checked" value="middle" radio />
<NcActionButton :model-value.sync="checked" value="right" radio />
```

This brings undefined behavior when components have both `radio` and `checkbox`, but I think it doesn't make sense to pass both and won't be confusing. And no flag = just a button as usually.

And I don't see a case where we would need to dynamically change `model-behavior` to another value.

What do you think?

ShGKme (commented):
Everything looks cool, seems to work and covers all the use-cases having good a11y, thanks!

Just some nitpicks/thoughts

src/components/NcActionButton/NcActionButton.vue
susnux: Yes but there are more places, at least 3 which need the condition for radios

src/components/NcActionButton/NcActionButton.vue
susnux: I am not sure here, while it is shorter I think it removes some makes it less informative.
It is not bad but I think if we really want to make it short we should do it like with `value` and use a property similar to the native HTML one.

Meaning we could use `type` with `radio`, `checkbox`, `button` just like the `<input>` element.

src/components/NcActionButton/NcActionButton.vue
ShGKme: My bad, missed it

src/components/NcActionButton/NcActionButton.vue
ShGKme: With `type` it mixes with button native `type` and we may come back to ugly `native-type` here again. 

For example, we add dialog support on `NcActions` with a form (we already use it as a dialog form actually) when we want to have a submit button in the `NcActions`.

(The same with `value` actually, but `value` on button is barely used on buttons in JS and I guess many developers even don't remember what button value was for).

susnux: Makes sense, but then I would stick with `model-behavior` (or just `behavior`?)

ShGKme: I just realize that having button native `type` doesn't conflict with a radio/checkbox type. As A button cannot be at the same time e.g. `submit` and `radio`, so when we need `type="submit"` it only requires to add one more type to the list `radio|checkbox|button|submit`

Let's go with `type`?

susnux: Sounds good! Implemented like that and also forward to `NcButton::nativeType` for inline actions.

susnux:
@ShGKme could you give this a try for the 8.5.0 release?

src/components/NcActionButton/NcActionButton.vue
```diff
@@ -269,6 +383,35 @@ export default {
...
+		},
+
+		/**
+		 * The buttons state if `type` is 'checkbox' or 'radio' (meaning if it is pressed / selected)
+		 * Either boolean for checkbox and toggle button behavior or `value` for radio behavior
+		 */
+		modelValue: {
+			type: [Boolean, String],
+			default: null,
+		},
```
ShGKme, line 405: It's said "if `type` is 'checkbox' or 'radio'" but it also changes attributes on other types.

And I think it's worth mentioning that `null` (the default value) in a specific case (in a menu with type checkbox) actually means `mixed`, not `false`.

Do we actually need this `mixed` value? It is not supported by `NcButton`, and there is no UI for this value. If feels a bit complicated to me...

ShGKme:
> @ShGKme could you give this a try for the 8.5.0 release?

Yes, but it also requires https://github.com/nextcloud-libraries/nextcloud-vue/pull/5119

Otherwise it is not a11y for inline buttons...

src/components/NcActionButton/NcActionButton.vue
susnux: I added more documentation on this behavior

PR #4745, opened 2023-11-02
[NcAppSettingsDialog] Allow to add icons to the navigation sections

susnux:
### ☑️ Resolves

* Part of https://github.com/nextcloud-libraries/nextcloud-vue/issues/1441

This consists of 2 commits, the first one changes the way the sections are registered and the second on provides icon support.

#### Section registration
Previously the default slot content was iterated and the props of every vNode were used. This had two downsides:
1. `$slots` is not reactive -> changes will not propagate
2. We can not easily access slots of a section (like for the icon) because the slot will only be available when the vNode is mounted

So instead we now provide to functions to children:
`registerSection` and `unregisterSection` with that functions children (the sections) can register themself and update on change. Solving both issues: the reactivity and the ability to inject icons (as the registration is done after mount).

I personally also think it is easier to understand as no longer internals of vNodes need to be handled, but this is personal taste.

#### Icon support
There is a new slot `icon` on the settings section which allows to inject icons to the section name in the navigation.

### 🖼️ Screenshots
With this changes it loos like this:

without icons (looks like before) | with icons
---|---
Screenshot_20231102_021445|Screenshot_20231102_021453

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppSettingsSection/NcAppSettingsSection.vue
+++ b/src/components/NcAppSettingsSection/NcAppSettingsSection.vue
@@ -32,6 +32,7 @@
 <script>
 export default {
 	name: 'NcAppSettingsSection',
+	inject: ['registerSection', 'unregisterSection'],
 
 	props: {
 		name: {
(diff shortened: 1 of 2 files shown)


src/components/NcAppSettingsSection/NcAppSettingsSection.vue
```diff
@@ -54,6 +57,24 @@ export default {
 			return 'settings-section_' + this.id
 		},
 	},
+	// Reactive changes for section navigation
+	watch: {
+		id(newId, oldId) {
+			this.unregisterSection(oldId)
+			this.registerSection(newId, this.name, this.$slots?.icon)
```
raimund-schluessler, line 64: I am not sure this will keep working with vue 3.

raimund-schluessler (approved):

marcoambrosini (approved):

PR #4767, opened 2023-11-07
fix(NcAppNavifation): small screen support

ShGKme:
### ☑️ Resolves

* Fix https://github.com/nextcloud/server/issues/40954
* Chained on https://github.com/nextcloud-libraries/nextcloud-vue/pull/4633

There are 2 issues with a small screen:
1. On a very small screen (320px) the 300px navigation is too wide, the toggle button is out of viewport
2. The toggle button doesn't look good on top of content

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
before-2 | after-2

Toggle button on a different background:

image

image

image

### 🚧 Tasks

- [x] Add max width to make navigation smaller on a small screen
- [x] Add background to the toggle button

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppNavigation/NcAppNavigation.vue
+++ b/src/components/NcAppNavigation/NcAppNavigation.vue
@@ -57,7 +57,9 @@ emit('toggle-navigation', {
 	<div ref="appNavigationContainer"
 		class="app-navigation"
 		:class="{'app-navigation--close':!open }">
-		<NcAppNavigationToggle :open="open" @update:open="toggleNavigation" />
+		<div class="app-navigation__toggle-wrapper">
+			<NcAppNavigationToggle :open="open" @update:open="toggleNavigation" />
+		</div>
 		<nav id="app-navigation-vue"
 			:aria-hidden="open ? 'false' : 'true'"
 			:aria-label="ariaLabel || undefined"
@@ -206,8 +208,11 @@ export default {
 	// Set scoped variable override
 	// Using --color-text-maxcontrast as a fallback evaluates to an invalid value as it references itself in this scope instead of the variable defined higher up
 	--color-text-maxcontrast: var(--color-text-maxcontrast-background-blur, var(--color-text-maxcontrast-default));
+	// Left padding + toggle button + right padding from NcAppContent
+	--app-navigation-toggle-width: calc(var(--app-navigation-padding) + var(--default-clickable-area) + var(--default-grid-baseline));
 	transition: transform var(--animation-quick), margin var(--animation-quick);
 	width: $navigation-width;
+	max-width: calc(100vw - var(--app-navigation-toggle-width));
 	position: relative;
 	top: 0;
 	left: 0;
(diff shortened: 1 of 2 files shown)


ShGKme:
@nextcloud-libraries/designers Design review request :)

JuliaKirschenheuter (approved):
wooow 🥳

szaimen (approved):
LGTM from design perspective for now. Improvements should be made later

src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -135,9 +170,27 @@ export default {
...
+			if (this.isMobile && this.open) {
+				this.focusTrap.activate()
+			} else {
+				this.focusTrap.deactivate()
+			}
+		},
+
+		handleEsc() {
+			if (this.isMobile) {
+				this.toggleNavigation(false)
```
szaimen, line 191: Shouldnt this work always?

ShGKme: What do you mean?

ShGKme:
I forgot to mention, that it is chained on https://github.com/nextcloud-libraries/nextcloud-vue/pull/4633 and only the last one commit is related

nimishavijay:
Looks good, I would say we can merge.

Mostly the "different color background" case wouldn't happen, and that's the only case where there's maybe we can improve things. We can check the behavior on different apps and see if we need any enhancements for design, so we can merge now :)

nimishavijay (approved):

susnux (approved):

Pytal (approved):

src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -180,7 +233,16 @@ export default {
...
+	&__toggle-wrapper {
+		position: absolute;
+		top: 0;
+		left: 100%;
+		// "app-navigation-padding (default-clickable-area) default-clickable-area" to fit AppContentHeader margin
+		width: calc(var(--app-navigation-padding) + var(--default-clickable-area) + var(--default-grid-baseline));
+		height: calc(var(--app-navigation-padding) * 2 + var(--default-clickable-area));
+	}
+
+gri--default-grid-baseline of navigation items
```
Pytal, line 245: Maybe supposed to be a comment?

jancborchardt (approved):
Looks good from my side as well @ShGKme! :) Only a detail for follow-up: The toggle does not need an extra square background, it just needs a circular color-main-background as default

ShGKme:
> The toggle does not need an extra square background, it just needs a circular color-main-background as default

I wanted to make visible that this button is connected to the navigation sidebar and remove displaying  any background content like the "ex" here:

image

Do you think this is redundant or there is a better way to do this?

jancborchardt:
@ShGKme I would say it is redundant. It's just a floating button, the screenshot looks fine as is and in most cases the page is being interacted with and it's clear it's a button. :)

susnux:
The only problem with the sidebar button is a lot of users can not find it if it overlays page content (got complains for this quite often (I think because it had no background and the icon mix up with text underneath).

PR #4777, opened 2023-11-08
fix: Do not depend on uncompiled vue files as dependencies

susnux:
### ☑️ Resolves

We should not require any not compiled vue files as this breaks unit tests in apps, as you require the users to compile vue files first.
We had this discussion with @skjnldsv on dialogs before and decided there to bundle the vue-material-design-icons, so I would like to do the same here.


AndyScherzinger (approved):

mejo- (approved):
Looks sensible :+1:

PR #4779, opened 2023-11-08
fix(NcAppNavigation): remove background square connecting toggle and navigation

ShGKme:
### ☑️ Resolves

See: https://github.com/nextcloud-libraries/nextcloud-vue/pull/4767#issuecomment-1801812704

Returns transparency between the app navigation and the toggle. Only the toggle button has a background.

Before (no text between nav and toggle) | After (text is visible, only toggle has background)
---|---
image | image


susnux (approved):

jancborchardt (approved):
Looks good!

Btw, unrelated but the left navigation could use a border-right to separate for narrow screens. On desktop itnis separated by the glass-look, but on mobile there is no separation between the backgrounds.

PR #4798, opened 2023-11-10
fix(NcAppSettingsDialog): `unregisterSection` should remove the section instead of remove all other

susnux:
### ☑️ Resolves

There was a logic error where unregister removed all but the unregistered section.
Also fix some minor issues and only warn on duplicated name instead of error.

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
+++ b/src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
@@ -181,6 +181,7 @@ import { useIsMobile } from '../../composables/useIsMobile/index.js'
 import { t } from '../../l10n.js'
 
 import debounce from 'debounce'
+import Vue from 'vue'
 
 export default {
 
@@ -194,7 +195,7 @@ export default {
 	provide() {
 		return {
 			registerSection: this.registerSection,
-			unregisterSection: this.registerSection,
+			unregisterSection: this.unregisterSection,
 		}
 	},
 
@@ -256,7 +257,7 @@ export default {
 			scroller: null,
 			/**
 			 * Currently registered settings sections
-			 * @type {{ id: string, name: string, icon?: VNode[]  }}
+			 * @type {{ id: string, name: string, icon?: import('vue').VNode[] }[]}
 			 */
 			sections: [],
 		}
@@ -295,11 +296,6 @@ export default {
 		},
 	},
 
-	mounted() {
-		// Select first settings section
-		this.selectedSection = this.$slots.default[0].componentOptions.propsData.id
-	},
-
 	updated() {
 		// Check that the scroller element has been mounted
 		if (!this.$refs.settingsScroller) {
@@ -327,23 +323,33 @@ export default {
 				throw new Error(`Duplicate section id found: ${id}. Settings navigation sections must have unique section ids.`)
 			}
 			if (this.sections.some(({ name: otherName }) => name === otherName)) {
-				throw new Error(`Duplicate section name found: ${name}. Settings navigation sections must have unique section names.`)
+				Vue.util.warn(`Duplicate section name found: ${name}. Settings navigation sections must have unique section names.`)
 			}
 
 			const newSections = [...this.sections, { id, name, icon }]
 			// Sort sections by order in slots
 			this.sections = newSections.sort(({ id: idA }, { id: idB }) => {
-				const indexOf = (id) => this.$slots.default.indexOf(vnode => vnode?.componentOptions?.propsData?.id === id)
+				const indexOf = (id) => this.$slots.default?.findIndex?.(vnode => vnode?.componentOptions?.propsData?.id === id) ?? -1
 				return indexOf(idA) - indexOf(idB)
 			})
+
+			// If this is the first section registered, set it as selected
+			if (this.sections.length === 1) {
+				this.selectedSection = id
+			}
 		},
 
 		/**
-		 * Called when a new section is unregistered
+		 * Called when a section is unregistered to remove it from dialog
 		 * @param {string} id The section ID
 		 */
 		unregisterSection(id) {
-			this.sections = this.sections.filter(({ id: otherId }) => id === otherId)
+			this.sections = this.sections.filter(({ id: otherId }) => id !== otherId)
+
+			// If the current section is unregistered, set the first section as selected
+			if (this.selectedSection === id) {
+				this.selectedSection = this.sections?.[0]?.id ?? ''
+			}
 		},
 
 		/**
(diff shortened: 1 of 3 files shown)


nickvergessen:
I guess this fixes https://github.com/nextcloud/spreed/issues/10858 ?

nickvergessen:
> and only warn on duplicated name instead of error.

Not sure about this. Duplicate IDs are a HTML validation error and no one reads warnings :(

src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
```diff
@@ -322,23 +323,23 @@ export default {
...
 			if (this.sections.some(({ name: otherName }) => name === otherName)) {
-				throw new Error(`Duplicate section name found: ${name}. Settings navigation sections must have unique section names.`)
+				Vue.util.warn(`Duplicate section name found: ${name}. Settings navigation sections must have unique section names.`)
 			}
 
 			const newSections = [...this.sections, { id, name, icon }]
 			// Sort sections by order in slots
 			this.sections = newSections.sort(({ id: idA }, { id: idB }) => {
-				const indexOf = (id) => this.$slots.default.indexOf(vnode => vnode?.componentOptions?.propsData?.id === id)
+				const indexOf = (id) => this.$slots.default?.indexOf?.(vnode => vnode?.componentOptions?.propsData?.id === id) ?? -1
```
ShGKme, line 332: `indexOf` search for a value. `findIndex` is needed to find with a callback

src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
```diff
@@ -292,7 +293,7 @@ export default {
 
 	mounted() {
 		// Select first settings section
-		this.selectedSection = this.$slots.default[0].componentOptions.propsData.id
+		this.selectedSection = this.$slots.default?.[0]?.componentOptions?.propsData?.id ?? ''
```
ShGKme, line 296: Update `selectedSection` only on mounted makes `selectedSection` invalid on dynamic `register`/`unregister`

tests/unit/components/NcAppSettingsDialog/register-unregister.spec.ts
ShGKme, line 1: Usually test files are named the same as the unit to be tested, e.g. `NcAppSettingsDialog` and `describe` is used for a group of tests for a unit.

tests/unit/components/NcAppSettingsDialog/register-unregister.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+			},
+		})
+
+		wrapper.vm.registerSection('test_id', 'test_name')
+		wrapper.vm.registerSection('test_id2', 'test_name2')
+		expect(wrapper.vm.$data.sections).toHaveLength(2)
+		wrapper.vm.unregisterSection('test_id')
+		expect(wrapper.vm.$data.sections).toHaveLength(1)
+		expect(wrapper.vm.$data.sections[0]).toEqual({ id: 'test_id2', name: 'test_name2' })
+	})
```
ShGKme, line 118: IMO, we should test the final result here instead.

These tests are much relayed on the component implementation and test their internal implementation.

`vm.$data` should never be tested, it is not a part of a component's contract. Otherwise, we cannot rename `sections` property or refactor component to Composition API without breaking tests.

Although `registerSection` is provided, here it is not a public `provide` but an internal thing used to connect the dialog with the section components supposed to be used as an "organic whole".

We should test that navigation is rendered correctly while the components are used as the end developer uses them.

ShGKme (dismissed):
`selectedSection` and sorting don't work on dynamic sections rendering

ShGKme:
> I guess this fixes [nextcloud/spreed#10858](https://github.com/nextcloud/spreed/issues/10858) ?

This error is caused by a typo in a `provide`. "Register" is provided as `unregister`.
```js
	provide() {
		return {
			registerSection: this.registerSection,
			unregisterSection: this.registerSection,
		}
	},
```

tests/unit/components/NcAppSettingsDialog/register-unregister.spec.ts
susnux: The tests are already grouped in a folder with the component name, so one can split test files to logical parts like in this case only test register / unregister. So files are not too long -> better readability.

susnux:
> > I guess this fixes [nextcloud/spreed#10858](https://github.com/nextcloud/spreed/issues/10858) ?
> 
> This error is caused by a typo in a `provide`. "Register" is provided as `unregister`.

Yes that is part of this fixes :)

susnux:
> Not sure about this. Duplicate IDs are a HTML validation error and no one reads warnings :(

@nickvergessen duplicated IDs still throw an error, only the duplicated name is a warning now.

tests/unit/components/NcAppSettingsDialog/register-unregister.spec.ts
susnux: Makes it a little bit more complex as we need to wait for next tick, but I adjusted the tests as I agree with the idea :)

tests/unit/components/NcAppSettingsDialog/register-unregister.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+	props: { id: { type: String, default: 'test_id' } },
+	setup(props) {
+		const register = inject<(id: string, name: string)=> void>('registerSection')
+		onMounted(() => register?.(props.id, 'test_name'))
+	},
+})
+
+describe('NcAppSettingsDialog: Sections registration', () => {
+	it('injects register function to children', async () => {
+		const wrapper = mount(NcAppSettingsDialog, {
```
ShGKme, line 40: image

tests/unit/components/NcAppSettingsDialog/register-unregister.spec.ts
ShGKme: But this is a common practice in both this repository and Jest/Vitest in general to name a file `<unit>.spec.js` and use `describe` to describe a group of the tests in unit testing.

It allows to easily find a test for the component by name (you know that `NcFoo` is supposed to be tested in `NcFoo.spec.js` (with some exceptions like `foo.spec.js` in some old tests).

It creates better reports because different tests of one unit won't be in different places in the report but grouped, which is better for readability in the CI/terminal. Now adding a new test file would add a new block in the report with a new name. They won't be visually subgroups of one unit, even if they are located in one folder (folder structure doesn't affect the report).

ShGKme (approved):
Fine by issue solving.

But I'm not a fun of testing implementation in unit tests, because it doesn't help to check the unit and prevent further refactoring without rewriting the unit test.

Pytal (approved):

PR #4804, opened 2023-11-12
[next] fix(NcActions): find actions deeper nested

raimund-schluessler:
### ☑️ Resolves

* This fixes an issue that occured when actions are nested deeper than one level. This can happen e.g. when one puts a loop in a slot, as it happens in `NcDashboardWidgetItem`, i.e.:

```vue
<NcActions>
	<slot name="actions">
		<NcActionButton v-for="action in actions">
			{{ action.text }}
		</NcActionButton>
	</slot>
</NcActions>
```
These actions were nested two levels deep in two `Fragments` and were not found with the old logic, causing the action menu to be empty. I changed the logic to be recursive now, finding all actions.

As this occurred in `NcDashboardWidgetItem`, it is included here as well.


raimund-schluessler:
Could we get this reviewed?

ShGKme (approved):

raimund-schluessler:
@susnux Would you have time to review this PR, maybe?

raimund-schluessler:
Fixed the conflict. @susnux Mind reviewing this as well, please?

susnux (approved):

PR #4815, opened 2023-11-14
Fix tabs appearance

marcoambrosini:
### ☑️ Resolves

* Fix #…

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot 2023-11-14 at 18 47 54 | Screenshot 2023-11-14 at 18 30 22

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppSidebar/NcAppSidebarTabs.vue
+++ b/src/components/NcAppSidebar/NcAppSidebarTabs.vue
@@ -269,8 +269,8 @@ export default {
 	&__nav {
 		display: flex;
 		justify-content: stretch;
-		margin-top: 10px;
-		padding: 0 4px;
+		margin: 10px 8px 0 8px;
+		border-bottom: 1px solid var(--color-border);
 	}
 
 	&__tab {
@@ -314,7 +314,23 @@ export default {
 	}
 }
 
-:deep(.checkbox-radio-switch--button-variant.checkbox-radio-switch) {
-	border: unset;
+// Override checkbox-radio-switch styles so that it looks like tabs
+:deep(.checkbox-radio-switch--button-variant) {
+	border: unset !important;
+	border-radius: 0 !important;
+	.checkbox-content {
+		padding: var(--default-grid-baseline);
+		border-radius: var(--default-grid-baseline) var(--default-grid-baseline) 0 0 !important;
+		margin: 0 !important;
+		border-bottom: var(--default-grid-baseline) solid transparent !important;
+		.checkbox-content__icon--checked > * {
+			color: var(--color-main-text) !important;
+		}
+	}
+	&.checkbox-radio-switch--checked .checkbox-radio-switch__content{
+		background: transparent !important;
+		color: var(--color-main-text) !important;
+		border-bottom: var(--default-grid-baseline) solid var(--color-primary-element) !important;
+	}
 }
 </style>


jancborchardt (approved):
Looks good design-wise, any detail fixes can be done later or next release.

susnux:
So do we want to go back with the design? Should we also adjust the NcCheckboxRadioSwitch button grouped styling?

raimund-schluessler:
Why do we go back now? The design feedback in https://github.com/nextcloud/server/issues/41450 does not say anything about this, just mentions a small fix.

src/components/NcAppSidebar/NcAppSidebarTabs.vue
```diff
@@ -314,7 +314,23 @@ export default {
 	}
 }
 
-:deep(.checkbox-radio-switch--button-variant.checkbox-radio-switch) {
-	border: unset;
+// Override checkbox-radio-switch styles so that it looks like tabs
+:deep(.checkbox-radio-switch--button-variant) {
```
susnux, line 318: This completely changes the UI of the component, should we instead change that component?
Because this is now very likely to break on slight changes of the internals of NcCheckboxRadioSwitch.
For the same reason I feel quite uncomfortable with that the many `!important` statements.

raimund-schluessler: Could we first learn what the intended goal is here? Do we intend to go back with the design completely? Is it just about small fixes?

marcoambrosini: That said, I think we're trying to do way too many things with `NcCheckboxRadioSwitch`

jancborchardt:
> Why do we go back now? The design feedback in https://github.com/nextcloud/server/issues/41450 does not say anything about this, just mentions a small fix.

@raimund-schluessler the concrete issue is that it’s much too present and makes everything more busy. The "primary" style is ok for the primary action buttons, and it’s also good for the left navigation, but also having this in the right sidebar makes it a bit much. It’s a classic case of "if everything is important, nothing is". :) Does that explain it?

src/components/NcAppSidebar/NcAppSidebarTabs.vue
jancborchardt: @marcoambrosini can you make sure it doesn’t break in these 2 places:
- Sharing flow where we use the component vertically (right?)
- Forms on top right, where we use it for View/Edit/Results

jancborchardt: > Could we first learn what the intended goal is here? Do we intend to go back with the design completely? Is it just about small fixes?

Yes to both: It is about small fixes, and the active tab being way too present has been identified as a visual issue. Going to a middle ground between the old and new one seems a good way to go.

jancborchardt: > Could we first learn what the intended goal is here? Do we intend to go back with the design completely? Is it just about small fixes?

@raimund-schluessler Yes to both: It is about small fixes, and the active tab being way too present has been identified as a visual issue. Going to a middle ground between the old and new one seems a good way to go.

raimund-schluessler:
> > Why do we go back now? The design feedback in [nextcloud/server#41450](https://github.com/nextcloud/server/issues/41450) does not say anything about this, just mentions a small fix.
> 
> @raimund-schluessler the concrete issue is that it’s much too present and makes everything more busy. The "primary" style is ok for the primary action buttons, and it’s also good for the left navigation, but also having this in the right sidebar makes it a bit much. It’s a classic case of "if everything is important, nothing is". :) Does that explain it?

Thanks for the explanation.

I just wonder whether adjusting the color would be enough then, instead of going back to the old design and potentially breaking quite some things. E.g. one could use a similar style to a secondary button:

Screenshot 2023-11-14 at 12-14-23 Nextcloud Vue Style Guide

JuliaKirschenheuter (approved):

nickvergessen:
> I just wonder whether adjusting the color would be enough then, instead of going back to the old design and potentially breaking quite some things. E.g. one could use a similar style to a secondary button:

This was an intermediate solution, but did not work correctly:
https://github.com/nextcloud-libraries/nextcloud-vue/pull/4791
Maybe it was not used fully correct as it had primary color text on the secondary color or something, but at least it was non-trivial as well

jancborchardt:
> I just wonder whether adjusting the color would be enough then, instead of going back to the old design and potentially breaking quite some things. E.g. one could use a similar style to a secondary button

@raimund-schluessler as @nickvergessen mentioned this was actually not accessible, reason being that there was too little contrast between an active entry and a non-active entry. They were not distinguishable. The bar below on this design achieves that.

raimund-schluessler:
> > I just wonder whether adjusting the color would be enough then, instead of going back to the old design and potentially breaking quite some things. E.g. one could use a similar style to a secondary button
> 
> @raimund-schluessler as @nickvergessen mentioned this was actually not accessible, reason being that there was too little contrast between an active entry and a non-active entry. They were not distinguishable. The bar below on this design achieves that.

Hm, I guess we are stuck with the old, and now inconsistent, design then, since no one has a better solution 😕

src/components/NcAppSidebar/NcAppSidebarTabs.vue
ShGKme: > Are we sure this won't cause a11y issues? Cc @JuliaKirschenheuter

It only changes styles, so the only possible a11y issues could be about **state styles** and **contrast**. With the contrast everything is fine.

But with states it is not.

The current tab should have hover and active state styles.

Before | After
---|---
with-state | no-state

ShGKme (changes requested):
The current tab should also have hover and active state styles.

Before | After
---|---
with-state | no-state

ShGKme:
As was mentioned in the old PR (https://github.com/nextcloud-libraries/nextcloud-vue/pull/3945#issuecomment-1506964491), from the design perspective, I'd prefer to go with this PR's design.

IMO, the radio button switch doesn't look like a tab but presents a tab.

---

I also agree with @susnux that it should not be implemented by deep override of the radio switch. However, I think we can merge it to have the new design soon, and fix the implementation later.

jancborchardt:
@ShGKme since Marco is probably already out of office since he is on Japan time, could I ask you to add a hover style there which fits accessibility requirements?

ShGKme:
> @ShGKme since Marco is probably already out of office since he is on Japan time, could I ask you to add a hover style there which fits accessibility requirements?

Asked @JuliaKirschenheuter to check as she also worked on this problem with the previous tabs state styles.

JuliaKirschenheuter:
https://www.w3.org/WAI/ARIA/apg/patterns/tabs/
https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/

ShGKme:
> https://www.w3.org/WAI/ARIA/apg/patterns/tabs/ https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/

Effects from the link:
- The **current** tab has a line above
- **Any** tab has a border inside on **hover**
- **Any** tab has a border inside on **active** (after click)

test

JuliaKirschenheuter:
> > https://www.w3.org/WAI/ARIA/apg/patterns/tabs/ https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/
> 
> Effects from the link:
> 
> * The **current** tab has a line above
> * **Any** tab has a border inside on **hover**
> * **Any** tab has a border inside on **active** (after click)
> 
> test [ test ](https://user-images.githubusercontent.com/25978914/282782595-599347dd-b04d-47b7-9900-11daa6c45264.gif) [ ](https://user-images.githubusercontent.com/25978914/282782595-599347dd-b04d-47b7-9900-11daa6c45264.gif)

@jancborchardt could we create same styles for our Tabs?

ShGKme:
Note, in the w3 example, they use the same styles as used on keyboard navigation. 

So in this PR it would look like this. But better to have a gap between the current tab bottom border and the outline.

image

marcoambrosini:
@ShGKme  @JuliaKirschenheuter I have doubts about the need of this outline for the active state, let alone any hover effects. We've been told before that no hover effects are ever necessary for accessibility. 
The link you posted contains one accessible solution, not the only accessible solution.

raimund-schluessler:
Honestly, I feel like this change is a bit drastic and rushed, given that the problem to solve is the tab active being "much too present and makes everything more busy" (basically just a attention issue).

With the changes proposed we will introduce hover and active state issues, get an inconsistent design, probably conflicts with the checkbox-radio-switch component, and maybe more. And all this just to fix an attention issue (not even a11y, since this is fine, as far as I understand it). We have other places in the design where we use primary as a background, e.g. the active navigation entry
Files
yet, this seems to be ok, while also diverting the attention.

JuliaKirschenheuter:
> If I press Right, will I move to another tab or not? It is unknown because the user doesn't know if the tab button is active or not.
> 
> Probably I need to click on the element first. But there is also no hover effect to show that it is clickable (not about a11y).

Agree! But our light gray color like on hover won't help at all. If this is a requirement then it have to be solved differently. I can't find requirements regarding active styles here https://www.w3.org/WAI/ARIA/apg/patterns/tabs/. And would say we can merge it like it is for now.

PR #4825, opened 2023-11-15
Create scrollable content via `min-height` over sidebar and settings …

JuliaKirschenheuter:
Create scrollable content via `min-height` over sidebar and settings dialog on small screens

### ☑️ Resolves

* Fix https://github.com/nextcloud/server/issues/41504
As suggested https://github.com/nextcloud/server/issues/36965#issuecomment-1812599749

### 🖼️ Screenshots

🏚️ Before

Peek 2023-11-15 16-57

Peek 2023-11-15 16-56

🏡 After

Peek 2023-11-15 16-53

Peek 2023-11-15 16-52

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
+++ b/src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
@@ -395,6 +395,9 @@ export default {
 
 <style lang="scss" scoped>
 .app-settings {
+	&:deep(.dialog) {
+		min-height: 256px;
+	}
 	&__navigation {
 		min-width: 200px;
 		margin-right: 20px;
@@ -409,6 +412,7 @@ export default {
 		overflow-x: hidden;
 		padding: 24px;
 		width: 100%;
+		min-height: 256px;
 	}
 }
 
--- a/src/components/NcAppSidebar/NcAppSidebarTabs.vue
+++ b/src/components/NcAppSidebar/NcAppSidebarTabs.vue
@@ -323,8 +323,7 @@ export default {
 
 	&__content {
 		position: relative;
-		// take full available height
-		min-height: 0;
+		min-height: 256px;
 		height: 100%;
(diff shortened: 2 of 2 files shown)


Pytal (approved):

src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
```diff
@@ -395,6 +395,9 @@ export default {
 
 <style lang="scss" scoped>
 .app-settings {
+	&:deep(.dialog) {
```
susnux, line 398: I do not think we should add this here, but rather directly on the NcDialog component

susnux:
sorry lets revert you fixup comment, my comment was wrong. If we set this on the dialog even small text dialogs have at least that height which looks weird.
You initial version was already perfect  :see_no_evil:

susnux (approved):

jancborchardt:
Isn’t there being 2 scrollable contents right within each other quite a no-go? Having to scroll twice and the possibility of getting those mixed up (and have content hidden) is quite confusing.

ShGKme:
> Isn’t there being 2 scrollable contents right within each other quite a no-go? Having to scroll twice and the possibility of getting those mixed up (and have content hidden) is quite confusing.

Yes, there will be 2 scrollable contents.

But this is the easiest solution we found to make pages usable on a 320x256 screen (a11y requirement).

As an alternative to the settings dialog, we could remove `NcModal` header, that could give as some pixels. But I don't see any alternative for the sidebar.

Do you have any ideas? @jancborchardt

jancborchardt:
Will this 2-scrollable-areas situation only appear at these small sizes or before already?

If only at these small sizes I'd say it's fine. If it appears for everyone even at desktop size or a bit smaller screen (considering sidebar is only a part of it), we'd have to check.

PR #4761, opened 2023-11-06
feat: introduce `useIsMobile` and `useIsFullscreen` composables

ShGKme:
Both composables store and return reactive state in module-scope (aka global) and are updated on resize.

Old mixins are now using new composables and are marked `deprecated`.

**Fully compatible with Vue 3**

Usage example in Options API:
```vue
<script>
import { useIsMobile } from '@nextcloud/vue/dist/Composables/useIsMobile.js'

export default {
  setup() {
    return {
      isMobile: useIsMobile(),
    }
  },

  watch: {
    isMobile() {
      // handle
    },
  },
}
</script>

<template>
  <div v-if="isMobile">I'm on mobile</div>
</template>
```

### 🚧 Tasks

- [x] Add `useIsMobile` composable, that stores and returns `isMobile` state in module-scope (aka global)
- [x] Add composables to build
- [x] Refactor `isMobile` mixin to use new composable and mark deprecated
- [x] Migrate all components to use composable instead of the mixin
- [x] The same with `isFullscreen`

### 🏁 Checklist

- [x] ⛑️ Tests are included or are not applicable
- [ ] 📘 Component documentation has been extended, updated or is not applicable

PR #4829, opened 2023-11-16
fix: export composables

raimund-schluessler:
### ☑️ Resolves

* This exports the composables added in #4761 and ensures the correct license and copyright.


ShGKme (approved):
Thanks 💙

susnux (approved):

PR #4831, opened 2023-11-16
[next] feat(NcContent): migrate to vue 3

raimund-schluessler:
### ☑️ Resolves

* This migrates `NcContent` and `NcApp*` components to vue3. `NcAppSidebar*` is not included yet, since this is needs larger adjustments.

One notable change is that `NcContent` behaves a bit different now. This is due to vue 3 which does not replace the element to which the app is mounted, but inserts into it, see https://v3-migration.vuejs.org/breaking-changes/mount-changes.html#mounted-application-does-not-replace-the-element. Hence, the HTML structure of the apps will change a bit from
```html
<body>
	...
	<div id="content-vue" class="app-files content">
	</div>
</body>
```
to
```html
<body>
	...
	<main id="content" class="app-files">
		<div id="content-vue" class="app-files content">
		</div>
	</main>
</body>
```


tests/unit/components/NcAppNavigation/NcAppNavigation.spec.js
```diff
@@ -133,9 +133,9 @@ describe('NcAppNavigation.vue', () => {
 		})
 	})
 
-	describe('focus trap', () => {
-		// TODO
-	})
+	// describe('focus trap', () => {
+	// 	// TODO
+	// })
```
ShGKme, line 138: I forgot to do it ><

By the way, it is a problem for Vitest to not comment it?
To avoid conflict on backport from `master` later.

ShGKme (approved):

tests/unit/components/NcAppNavigation/NcAppNavigation.spec.js
raimund-schluessler: > By the way, it is a problem for Vitest to not comment it?

Unfortunately, yes. This is why I saw it 🙈 The test will fail with `test not found`.

susnux (approved):

PR #4832, opened 2023-11-16
Improve relative time handling

nickvergessen:
### Resolve

- 

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
See activity from "50 hours ago" and start dying as you are bad in math instead of knowing when a colleague did something | "2 days ago" helpfully obvious
Bildschirmfoto vom 2023-11-16 12-16-41 | grafik

### Background info

In stable27 Moment was used by the activity app, the breaking points are here:
https://github.com/moment/moment/blob/000ac1800e620f770f4eb31b5ae908f6167b0ab2/src/lib/duration/humanize.js#L4-L11

We could also go for those, but at least the 72 hours need to go.

### 🏁 Checklist

### DIFF
--- a/src/components/NcDateTime/NcDateTime.vue
+++ b/src/components/NcDateTime/NcDateTime.vue
@@ -197,7 +197,7 @@ export default {
 					return formatter.format(Math.round(minutes), 'minute')
 				}
 				const hours = minutes / 60
-				if (Math.abs(hours) <= 72) {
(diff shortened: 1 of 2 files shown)


susnux (approved):
Makes sense

raimund-schluessler (approved):

ShGKme (approved):

PR #4833, opened 2023-11-16
[next] feat(NcAppSidebar*): migrate to vue 3

raimund-schluessler:
### ☑️ Resolves

* This migrates `NcAppSidebar*` to vue 3.


src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -654,11 +655,8 @@ export default {
 		canStar() {
 			return this.isStarred !== null
 		},
-		hasFigure() {
-			return this.$slots.header || this.background
-		},
 		hasFigureClickListener() {
-			return this.$listeners['figure-click']
+			return this.$attrs['figure-click']
```
raimund-schluessler, line 659: This does not work anymore. Also see https://github.com/nextcloud-libraries/nextcloud-vue/pull/3692/files#r1276581301.

raimund-schluessler: Fixed in 7999dada18934a0416718fe8576fc362c05b8721 by removing `figure-click` from `emits` and use it as `$attrs.onFigureClick(event)` as suggested by @ShGKme.

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -732,7 +736,7 @@ export default {
 			 *
 			 * @type {Event}
 			 */
-			this.$emit('figure-click', e)
+			this.$attrs.onFigureClick(e)
```
ShGKme, line 739: We still should emit this event, otherwise it won't be visible in `vue-devtools`

src/components/NcAppSidebarTab/NcAppSidebarTab.vue
```diff
@@ -130,7 +130,7 @@ export default {
 		 * @return {import('vue').VNode[]}
 		 */
 		renderIcon() {
-			return this.$scopedSlots.icon?.()
+			return this.$slots.icon
```
ShGKme, line 133: Otherwise dynamic component slot will have no effect. But I haven't tested it.

```suggestion
			return defineComponent((props, { slots }) => this.$slots.icon?.() || slots.default())
```

ShGKme: Also, `this.$slots.icon` is a render function, so we need to rename a method and update JSDoc.

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -749,18 +749,17 @@ export default {
...
 
-		editName() {
+		async editName() {
 			/**
 			 * Emitted when the nameEditable value changes
 			 *
 			 * @type {boolean}
 			 */
-			this.$emit('update:nameEditable', true)
+			await this.$emit('update:name-editable', true)
```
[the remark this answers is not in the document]
raimund-schluessler: No, it does not. I don't know how this `await` ended up there :see_no_evil:

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -634,7 +635,7 @@ export default {
 		'opened',
 		'figure-click',
 		'update:starred',
-		'update:nameEditable',
+		'update:name-editable',
```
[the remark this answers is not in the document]
raimund-schluessler: Fine with me. I just aligned it with the other event names in this component. I will send a PR later to switch all event names from kebab to camel case.

src/components/NcAppSidebarTab/NcAppSidebarTab.vue
raimund-schluessler: The icon slot is reactive, if that is what you mean:

[Bildschirmaufzeichnung vom 2023-11-17, 09-20-19.webm](https://github.com/nextcloud-libraries/nextcloud-vue/assets/2496460/df97b015-d537-4d51-9147-cc7b8cd28344)


```vue
<template>
		<NcCheckboxRadioSwitch v-model:checked="iconType" value="magnify" name="icon_radio" type="radio">Magnify</NcCheckboxRadioSwitch>
		<NcCheckboxRadioSwitch v-model:checked="iconType" value="cog" name="icon_radio" type="radio">Cog</NcCheckboxRadioSwitch>
	<NcAppSidebar
		:starred="starred"
		name="cat-picture.jpg"
		subname="last edited 3 weeks ago">
		<NcAppSidebarTab name="Search" id="search-tab">
			<template #icon>
				<Cog v-if="iconType === 'cog'"  :size="20" />
				<Magnify v-else :size="20" />
			</template>
			Search tab content
		</NcAppSidebarTab>
		<NcAppSidebarTab name="Settings" id="settings-tab">
			<template #icon>
				<Cog :size="20" />
			</template>
			Settings tab content
		</NcAppSidebarTab>
		<NcAppSidebarTab name="Sharing" id="share-tab">
			<template #icon>
				<ShareVariant :size="20" />
			</template>
			Sharing tab content
		</NcAppSidebarTab>
	</NcAppSidebar>
</template>
<script>
	import Magnify from 'vue-material-design-icons/Magnify'
	import Cog from 'vue-material-design-icons/Cog'
	import ShareVariant from 'vue-material-design-icons/ShareVariant'

	export default {
		components: {
			Magnify,
			Cog,
			ShareVariant,
		},
		data() {
			return {
				starred: false,
				iconType: 'magnify',
			}
		},
		methods: {
			figureClick() {
				alert('figure clicked')
			},
		},
	}
</script>
```

raimund-schluessler: Is it ok to leave it as `return this.$slots.icon` or did I misunderstood your concern? If yes, then I would just change the function name to `renderFunctionIcon()` and adjust the JSDoc. Could you give me a hint, what the correct type would be?

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -733,6 +737,7 @@ export default {
 			 * @type {Event}
 			 */
 			this.$emit('figure-click', e)
+			this.$attrs.onFigureClick(e)
```
raimund-schluessler, line 740: I guess we need to add an eslint exception here @ShGKme

ShGKme: Yes, here this is an exception, we need to add eslint ignore next line.

An alternative is to add a prop to control the component behavior.

src/components/NcAppSidebarTab/NcAppSidebarTab.vue
ShGKme: > The icon slot is reactive, if that is what you mean:

Before, `VNodes` component was used, which not only rendered vnodes from from `vnodes` prop but also rendered its default content as a fallback content.

It was used to render `class-based` icon, when `icon` prop is provided instead of the `#icon` slot.

Now it is replaced with dynamic component `<component :is="render function from $slots.icon">`. Which renders vnodes from `renderIcon = $slots.icon`, but doesn't render its default slot.

So, `icon` prop doesn't work.

src/components/NcAppSidebar/NcAppSidebar.vue
raimund-schluessler: > An alternative is to add a prop to control the component behavior.

Yes, but this would mean a breaking change. Fine with me as well, I don't have a preference here.

src/components/NcAppSidebarTab/NcAppSidebarTab.vue
susnux: > `class-based` icon

I think we can finally get rid of icon classes with nextcloud-vue 9, no?

raimund-schluessler: > > `class-based` icon
> 
> 
> 
> I think we can finally get rid of icon classes with nextcloud-vue 9, no?

That would be nice. But let's not break it here. We can simply keep using `VNodes` for now.
Then we have a working version now and can still break it later.

raimund-schluessler: Fixed it by bringing back `VNodes`.

src/components/NcAppSidebar/NcAppSidebar.vue
raimund-schluessler: Fixed.

raimund-schluessler:
Should be good now. I will squash after reviews.

raimund-schluessler:
> I forgot to enable the visual regression tests, and cypress is not happy now.

Works now with 5244cb723f161c59b63aaa74abc8751ad214f6d4 by using `isSlotPopulated` for slot detection.

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -347,14 +354,14 @@ export default {
 		@after-leave="onAfterLeave">
 		<aside id="app-sidebar-vue" class="app-sidebar">
 			<header :class="{
-					'app-sidebar-header--with-figure': hasFigure,
+					'app-sidebar-header--with-figure': isSlotPopulated($slots?.header?.()) || background,
```
ShGKme, line 357: `$slots` is guaranteed to not be null, it is at least an empty object, null check is redundant here and in other places in the component.

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -732,7 +739,9 @@ export default {
 			 *
 			 * @type {Event}
 			 */
+			// eslint-disable-next-line vue/require-explicit-emits
 			this.$emit('figure-click', e)
+			this.$attrs.onFigureClick(e)
```
ShGKme, line 744: It calls the handler twice: once as a component event handler by `$emit` and once directly in `onFigureClick`.

src/components/NcAppSidebar/NcAppSidebar.vue
raimund-schluessler: Yeah, I wasn't sure it is guaranteed to be present and wanted to be safe. Fixed it.

src/components/NcAppSidebar/NcAppSidebar.vue
raimund-schluessler: I was under the impression it does not emit when it's not listed under `emits`. I removed `$attrs.onFigureClick` and it seems to work correct now.

ShGKme (approved):

susnux (approved):

PR #4836, opened 2023-11-16
Create scrollable content via `min-height` over sidebar and settings dialog on small screens

JuliaKirschenheuter:
Create scrollable content via `min-height` over sidebar and settings dialog on small screens

### ☑️ Resolves

* Fix https://github.com/nextcloud/server/issues/41542

### 🖼️ Screenshots

🏚️ Before 

image

🏡 After

Peek 2023-11-17 10-01

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
+++ b/src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
@@ -145,6 +145,7 @@ export default {
 		@update:open="handleCloseModal">
 		<template #navigation="{ isCollapsed }">
 			<ul :aria-label="settingsNavigationAriaLabel"
+				v-show="!isCollapsed"
 				:class="{ 'navigation-list': true, 'navigation-list--collapsed': isCollapsed }"
 				role="tablist">
 				<li v-for="section in sections" :key="section.id">


Pytal (approved):

ShGKme:
2 blocks with scrolls doesn't look good on a so small screen...

What about:
- Do not separate navigation and content on a small screen
- Just hide the navigation on a such small screen

cc @nextcloud-libraries/designers for ideas

JuliaKirschenheuter:
> * Just hide the navigation on a such small screen

That is a good idea! Actually this navigation is not important on small screens. @marcoambrosini @nimishavijay would it work for you?

susnux:
> Just hide the navigation on a such small screen

The navigation was introduced by accident on mobile screens, meaning we should just remove it then again like before.

susnux (approved):
Works but not sure if we do not want to simply `v-show="!isCollapsed"` but for the moment this fixes the issue, works and looks good!

src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
```diff
@@ -469,4 +470,12 @@ export default {
...
 }
 
+@media only screen and (max-width: 320px) {
+	.navigation-list {
+		&--collapsed {
+			display: none;
+		}
+	}
+}
+
```
susnux, line 480: this then is no longer needed?

src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
```diff
@@ -145,6 +145,7 @@ export default {
 		@update:open="handleCloseModal">
 		<template #navigation="{ isCollapsed }">
 			<ul :aria-label="settingsNavigationAriaLabel"
+				v-show="!isCollapsed"
```
szaimen, line 148: Isnt v-if usually better?

JuliaKirschenheuter: we can leave it for now with v-show

ShGKme (approved):

src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
ShGKme: In this case, it is fine, IMO. We just hide it on a small screen, not like having a different rendering branchs.

So, works like `display: none` in `@media` query =D

PR #4837, opened 2023-11-16
fix(cypress): delete unused files

raimund-schluessler:
### ☑️ Resolves

* This removes files that were erroneously committed in #4772 and increase the trouble checking out the repo due to too long file names, see https://github.com/nextcloud-libraries/nextcloud-vue/pull/4760#issuecomment-1814543709


szaimen (approved):
Thanks

susnux (approved):

PR #4841, opened 2023-11-16
fix(NcActions): adjust keyboard navigation

ShGKme:
### ☑️ Resolves

* Fix https://github.com/nextcloud/server/issues/41012
* https://www.w3.org/WAI/ARIA/apg/patterns/menubar/
* Briefly, **a primary keyboard navigation convention common across all platforms is:**
  * <kbd>TAB</kbd> and <kbd>SHIFT + TAB</kbd> should move focus between **components**
  * <kbd>Arrows ⬆️⬇️</kbd> should move focus between **elements inside a component**

See also: https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/

### 🚧 Tasks

- [x] Make <kbd>Arrows ⬆️⬇️</kbd> loop over elements
- [x] <kbd>TAB</kbd> and <kbd>SHIFT + TAB</kbd>:
  - [x] In **menu** and **navigation** should move to next/prev component
  - [x] In **popover like** actions with form elements should loop over elements
    - <kbd>ESC</kbd> closes actions

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
**Arrow** |
Don't loop | Loop
arrows-before | arrows-after
**TAB in Menu / Navigation** | 
Between elements | Between components
Tab-Menu-Before | Tab-Menu
**TAB in Popover-like** |
Submit is not focusable | Classic popover with focus trap
Tab-Before | Tab-Popover

### 🏁 Checklist

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -1154,8 +1154,7 @@ export default {
 		focusPreviousAction(event) {
 			if (this.opened) {
 				if (this.focusIndex === 0) {
-					// First element overflows to body-navigation (no preventDefault!) and closes Actions-menu
-					this.closeMenu()
+					this.focusLastAction(event)
 				} else {
 					this.preventIfEvent(event)
 					this.focusIndex = this.focusIndex - 1
@@ -1167,8 +1166,7 @@ export default {
 			if (this.opened) {
 				const indexLength = this.$refs.menu.querySelectorAll(focusableSelector).length - 1
 				if (this.focusIndex === indexLength) {
-					// Last element overflows to body-navigation (no preventDefault!) and closes Actions-menu
-					this.closeMenu()
+					this.focusFirstAction(event)
 				} else {
 					this.preventIfEvent(event)
 					this.focusIndex = this.focusIndex + 1
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -992,6 +992,8 @@ export default {
 			focusIndex: 0,
 			randomId: `menu-${GenRandomId()}`,
 			isSemanticMenu: false,
+			isSemanticNavigation: false,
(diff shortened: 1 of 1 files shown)


ShGKme:
**⚠️ Before merge should be tested in apps**

Pytal:
I recall that there was a reason the actions menu specifically didn't need to be focus trapped but this seems better :)

src/components/NcActions/NcActions.vue
```diff
@@ -1341,7 +1348,10 @@ export default {
 						boundary: this.boundariesElement,
 						container: this.container,
 						popoverBaseClass: 'action-item__popper',
-						setReturnFocus: this.$refs.menuButton?.$el,
+						// Menu and navigation should not have focus trap
+						// Tab should close the menu and move focus to the next UI element
```
Pytal, line 1352: nvm 🙈

Pytal (approved):

ShGKme:
> I recall that there was a reason the actions menu specifically didn't need to be focus trapped but this seems better :)

Currently it has a focus trap when it is used more like as a popover, but with action items. Like forms.

When it is used as a menu, list of buttons/links - it doesn't has `focus-trap`

src/components/NcActions/NcActions.vue
ShGKme: > nvm 🙈

I prefer https://volta.sh =D

susnux (approved):

JuliaKirschenheuter (approved):
🥳 works well

ShGKme:
Tested in:
- Server
  - Dashboard - Weather Status
  - Files - File list
  - Files - sidebar: sharing, versions
  - Users - new group + user edit
- Talk - issues in navigation, but seems unrelated
- Text - works weirdly, but I'd call it unrelated

PR #4861, opened 2023-11-21
enh: Split translations by components to only include needed strings in app bundles

### LINKED ISSUE #4411: Split translations by component
I think it would be nice if we could split translations by components when building, that way not every string for even unused components end up in you bundle.

cc @ShGKme

susnux:
### ☑️ Resolves

* Fix #4411 

On build time:
1. Translations are extracted
2. Not translated strings are discarded (fallback to English)
3. Translation objects are compressed
4. Translations are split by files that use that translation string

Then a vite plugin will intercept all imports for 'src/l10n.js' and replace the content with a wrapper that only injects the used translations of the importing file.

#### Results
Tested with logreader app:

##### Before
```
Building logreader for production
vite v4.5.0 building for production...
✓ 1367 modules transformed.
./css/logreader-main.css       99.63 kB │ gzip:  16.69 kB
./js/Splitter-9140b094.mjs      0.97 kB │ gzip:   0.59 kB │ map:     3.29 kB
./js/logreader-main.mjs     1,645.71 kB │ gzip: 390.74 kB │ map: 5,876.41 kB
```
##### After
```
Building logreader for production
vite v4.5.0 building for production...
✓ 1366 modules transformed.
./css/logreader-main.css      100.89 kB │ gzip:  16.89 kB
./js/Splitter-9140b094.mjs      0.97 kB │ gzip:   0.59 kB │ map:     3.29 kB
./js/logreader-main.mjs     1,492.89 kB │ gzip: 371.30 kB │ map: 6,063.92 kB
```

### 🏁 Checklist


ShGKme:
I have this build result and `.translations[""]` looks like it is not supposed to be `""`.

```js
import { getGettextBuilder } from "@nextcloud/l10n/gettext";
const gettext = getGettextBuilder().detectLocale().build();
const n = gettext.ngettext.bind(gettext);
const t = gettext.gettext.bind(gettext);
const register = (...chunks) => {
  chunks.forEach((chunk) => {
    if (!chunk.registered) {
      chunk.forEach(({ l: locale, t: translations }) => {
        const decompressed = Object.fromEntries(
          Object.entries(translations).map(([id, value]) => [
            id,
            {
              msgid: id,
              msgid_plural: value.p
              msgstr: value.v
            }
          ])
        );
        if (!gettext.gt.catalogs[locale]) {
          gettext.gt.catalogs[locale] = { messages: { translations: {} } };
        }
        gettext.gt.catalogs[locale].messages.translations[""] = { ...gettext.gt.catalogs[locale].messages.translations[""], ...decompressed };
      });
      chunk.registered = true;
    }
  });
};
```

susnux:
> What is weird, this import is generated for me with null symbol in the beginning of the path `\0`

Imports starting with 0 byte are imports from vite plugins, in this case it is imported by the vue2 plugin and not related to this PR.

susnux:
> I have this build result and `.translations[""]` looks like it is not supposed to be `""`.

This is the format gettext-parser and node gettext expect it, see also:
https://github.com/nextcloud-libraries/nextcloud-vue/blob/29b408a39a36f9729ce7e32b03a246895903097d/build/translations.js#L40

ShGKme:
> Imports starting with 0 byte are imports from vite plugins, in this case it is imported by the vue2 plugin

Yes, for virtual modules in Rollup. But only during the build. It should not be in the final `dist`, because this import is not a valid import.

> and not related to this PR.

Neither Server nor Talk can be built with this branch with an error here... And this error only appears if this new plugin is used.

build/l10n-plugin.mts
```diff
@@ -0,0 +1,123 @@
...
+		 * @param importer The file that imported the file
+		 */
+		resolveId(source, importer) {
+			if (source.startsWith('\0')) {
+				if (source === '\0l10n') {
+					// return our l10n main module containing all translations
+					return '\0l10n'
+				}
+				// dont handle other plugins imports
+				return false
```
ShGKme, line 65: Are you sure that we should ignore others here and not return to the same source for further resolving? 

```suggestion
				// dont handle other plugins imports
				return source
```

ShGKme (dismissed):
The build result doesn't work in Server and Talk, having invalid import.

build/l10n-plugin.mts
susnux: No, but false is also wrong. I mixed it up, from the docs:

> Returning null defers to other resolveId functions and eventually the default resolution behavior.
> Returning `false` signals that source should be treated as an external module and not included in the bundle.

susnux:
> The build result doesn't work in Server and Talk, having invalid import.

I tested with logreader that already uses vite so I did not noticed this error, but I can reproduce. This is caused by returning `false` instead of `null` for foreign internal modules. Will fix it.

build/l10n-plugin.mts
susnux: fixed :heavy_check_mark:

ShGKme (approved):
The `dist` looks good and correct without any extra translations, the translations on server and Talk work fine. Doesn't change a lot in size for 1 app, but multiplies for every app and library.

build/l10n-plugin.mts
```diff
@@ -0,0 +1,123 @@
...
+		 * This function injects the translation chunks by returning a module that exports one translation object per component
+		 * @param id The name of the module that should be loaded
+		 */
+		load(id) {
+			const match = id.match(/\0l10nwrapper\?source=(.+)/)
+			if (match) {
+				// In case this is the wrapper module we provide a module that imports only the required translations and exports t and n functions
+				const source = decodeURIComponent(match[1])
+				// eslint-disable-next-line @typescript-eslint/no-unused-vars
+				const imports = Object.entries(nameMap).filter(([paths, _value]) => paths.split(':').some((path) => source.endsWith(path))).map(([paths, alias]) => alias)
```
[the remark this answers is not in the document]
susnux: I tried to improve this and some other places with such lines.

AndyScherzinger (approved):

PR #4864, opened 2023-11-21
fix(NcAppNavigation): make toggle button not the first element for focus-trap

Antreesy:
### ☑️ Resolves

* Move `NcAppNavigationToggle` down the list
* Possible effect: when isMobile is active (from #4633), focus-trap grabs the first available element.
  * Before fix: toggle button
  * After fix: first element in navigation

### 🖼️ Screenshots

From Talk usecase of component:

🏚️ Before | 🏡 After
---|---
image | image

### DIFF
--- a/src/components/NcAppNavigation/NcAppNavigation.vue
+++ b/src/components/NcAppNavigation/NcAppNavigation.vue
@@ -57,9 +57,6 @@ emit('toggle-navigation', {
 	<div ref="appNavigationContainer"
 		class="app-navigation"
 		:class="{'app-navigation--close':!open }">
-		<div class="app-navigation__toggle-wrapper">
-			<NcAppNavigationToggle :open="open" @update:open="toggleNavigation" />
-		</div>
 		<nav id="app-navigation-vue"
 			:aria-hidden="open ? 'false' : 'true'"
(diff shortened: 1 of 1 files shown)


src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -76,6 +73,9 @@ emit('toggle-navigation', {
 			<!-- Footer for e.g. AppNavigationSettings -->
 			<slot name="footer" />
 		</nav>
+		<div class="app-navigation__toggle-wrapper">
+			<NcAppNavigationToggle :open="open" @update:open="toggleNavigation" />
+		</div>
```
ShGKme, line 78: This `div` wrapper can be removed (forgot to do it in https://github.com/nextcloud-libraries/nextcloud-vue/pull/4779)

ShGKme (approved):

src/components/NcAppNavigation/NcAppNavigation.vue
Antreesy: Positioned and works the same without div, updated

susnux (approved):

PR #4869, opened 2023-11-22
fix(NcAppNavigation): closed on mobile initially

ShGKme:
### ☑️ Resolves

When a window is resized to a mobile size, the app navigation closes. However, when the page was initially open on mobile, the navigation is open. It could be annoying because a user needs to close it to see the content.

See also discussion from: https://github.com/nextcloud/spreed/issues/10908#issuecomment-1822460276

### 🚧 Tasks

- [x] Make navigation closed not only after **switch to mobile**, but also **initially on mobile**

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppNavigation/NcAppNavigation.vue
+++ b/src/components/NcAppNavigation/NcAppNavigation.vue
@@ -93,12 +93,6 @@ export default {
 		NcAppNavigationToggle,
 	},
 
-	setup() {
-		return {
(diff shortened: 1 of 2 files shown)


szaimen (approved):
Makes sense but didnt test

Antreesy (approved):
Tested with Talk

jancborchardt (approved):
Agree, content should be given priority. :)

PR #4881, opened 2023-11-28
fix(NcDialog): Add slighly more padding on the bottom and fix example button order

susnux:
### ☑️ Resolves

On the OCDialogs.confirm and everywhere else we use
left: secondary action
right: primary action

So I adjusted the example for this.

And also add slightly more padding on the bottom in case there is only text shown, so that the padding on the left / right and bottom is the same.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot 2023-11-28 at 21-19-50 Nextcloud Vue Style Guide|Screenshot 2023-11-28 at 21-19-24 Nextcloud Vue Style Guide

### 🏁 Checklist

### DIFF
--- a/src/components/NcDialog/NcDialog.vue
+++ b/src/components/NcDialog/NcDialog.vue
@@ -45,15 +45,15 @@ export default {
 			lastResponse: 'None',
 			buttons: [
 				{
+					label: 'Cancel',
+					icon: IconCancel,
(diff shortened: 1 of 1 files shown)


raimund-schluessler (approved):
Eslint is not happy, otherwise fine with me.

marcoambrosini (approved):

PR #4884, opened 2023-11-29
enh(NcAppNavigationItem): Allow to pass SVG or path as icon property

susnux:
### ☑️ Resolves

Not only allow icon classes (somewhat deprecated) but also SVG or SVG paths as icon property.

### 🖼️ Screenshots
Screenshot 2023-11-29 at 15-34-46 Nextcloud Vue Style Guide

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
+++ b/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
@@ -33,18 +33,32 @@
 
 ```vue
 	<template>
-		<NcAppNavigationItem name="My name">
-			<template #icon>
-				<Check :size="20" />
-			</template>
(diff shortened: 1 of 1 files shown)


ShGKme:
Is it required for a specific place? Why not use the icon slot?

susnux:
> Why not use the icon slot?

I can do so. Whats then the plan with the icon props? Should we deprecate them as icon classes are deprecated and just recommend using icon slots with svg wrapper. Or should be move them to SVG support?

susnux:
Superseded by (IMHO better) approach by @ShGKme 
https://github.com/nextcloud-libraries/nextcloud-vue/pull/6348

PR #4889, opened 2023-11-29
fix(NcColorPicker): correct aria-label

ShGKme:
`aria-label` should be set directly on the element being labelled.

### ☑️ Resolves

* Fix https://github.com/nextcloud/server/issues/41904


Pytal (approved):

szaimen (approved):
LGTM but didnt test

susnux (approved):

PR #4894, opened 2023-11-30
feat(NcNoteCard): provide a slot for inserting a custom icon instead of default

Antreesy:
### ☑️ Resolves

* Allows to replace default icon with a custom one
* If no custom icon is provided, default is used as a fallback for a slot
* Reduce efforts for implementing a workarounds and extra wrappers, if custom icon is needed

### 🖼️ Screenshots

| 🏡 After
|---
| image

### 🏁 Checklist

### DIFF
--- a/src/components/NcNoteCard/NcNoteCard.vue
+++ b/src/components/NcNoteCard/NcNoteCard.vue
@@ -30,23 +30,42 @@ available in four versions:
 When using an error type,
 
 ```vue
-<div>
-	<NcNoteCard type="warning">
-		<p>This is dangerous</p>
-	</NcNoteCard>
+<template>
+	<div>
+		<NcNoteCard type="warning">
+			<p>This is dangerous</p>
+		</NcNoteCard>
 
-	<NcNoteCard type="error" heading="Error">
-		<p>The server is not happy and reported the following error</p>
-	</NcNoteCard>
+		<NcNoteCard type="error" heading="Error">
+			<p>The server is not happy and reported the following error</p>
+		</NcNoteCard>
 
-	<NcNoteCard type="success">
-		<p>You won</p>
-	</NcNoteCard>
+		<NcNoteCard type="success">
+			<p>You won</p>
+		</NcNoteCard>
 
-	<NcNoteCard type="info">
-		<p>For your information</p>
-	</NcNoteCard>
-</div>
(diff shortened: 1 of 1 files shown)


Antreesy:
> What is the usecase for this?

As always, a Talk repo. Styles of NoteCard component are fitting for Out-of-Office disclaimer to not reinvent this component from scratch, but I need to put a user's avatar instead of blue exclamation icon

image

ShGKme:
> > What is the usecase for this?
> 
> As always, a Talk repo. Styles of NoteCard component are fitting for Out-of-Office disclaimer to not reinvent this component from scratch, but I need to put a user's avatar instead of blue exclamation icon

Could avatar be the content of the note instead of an icon?

IMO, it is a very edge case for this component, and avatar should not be used as an icon.

susnux:
> Agree, but it's wider than just an avatar, and more a nice-to-have option (which we have at NcAction... components, for example)

Yes but for the avatar this would be the wrong place and should not be placed there.

And I am no designer but I think that this might make thing more inconsistent design wise (for all other cases). But as said I am not a designer.

susnux:
> image

@nextcloud-libraries/designers is this something we want?

If yes then lets go with it :)

szaimen:
> > image
> 
> @nextcloud-libraries/designers is this something we want?
> 
> If yes then lets go with it :)

what exaclty? (the screenshot is missing)

ShGKme:
> what exaclty? (the screenshot is missing)

Fixed

susnux:
> what exaclty?

Having a custom icon in the note card, or more specific should we omit the "info" icon if there is an avatar.

szaimen:
> Having a custom icon in the note card, or more specific should we omit the "info" icon if there is an avatar.

Yes, makes sense imho. But I would design that thing specifically like this: 
image

szaimen:
I just wanted to point out that the design for this specific feature could be improved. It is up to you to decide if such a design would make sense as separate component or to overwrite the design of the component in talk then...

Antreesy:
I don't see it as a wide use-case to make it a separate component. That's why I'd just put a slot as proposed by this PR and do the rest in the app itself.
Feature design is also up to Talk repo to discuss, so we can proceed with it there

DorraJaouad (approved):

PR #4902, opened 2023-12-01
enh: Add a full labelled color palette to be used as the default for the color picker

susnux:
### ☑️ Resolves

Lets provide a fully labelled color palette for the color picker by default.
When users provide a different palette its their task but for the default we can provide some color names (yes that are real color names for that colors).

### 🏁 Checklist

### DIFF
--- a/l10n/messages.pot
+++ b/l10n/messages.pot
@@ -14,6 +14,10 @@ msgstr ""
 msgid "a few seconds ago"
 msgstr ""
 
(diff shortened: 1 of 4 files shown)


raimund-schluessler (approved):
That will be fun for the translators.

szaimen (approved):
LGTM

ShGKme (approved):

PR #4903, opened 2023-12-01
enh(NcCheckboxRadioSwitch): Allow to set `aria-label`

susnux:
### ☑️ Resolves

If no text should be shown currently you have to provide the label as a `span` with `hidden-visually`, because the `aria-label` would not be forwarded correctly.

### 🏁 Checklist


susnux:
Alternative would be disable forwarding attr and filter for any `aria` attribute and move them to the input.

ShGKme (approved):

PR #4905, opened 2023-12-01
fix(NcListItem): Add focus-visible outline state

susnux:
### ☑️ Resolves

Add a focus-visible outline as the default one is not accessible (too low contrast).

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot 2023-12-01 at 17-50-47 Nextcloud Vue Style Guide|Screenshot 2023-12-01 at 17-51-23 Nextcloud Vue Style Guide
Screenshot 2023-12-01 at 17-50-27 Nextcloud Vue Style Guide|Screenshot 2023-12-01 at 17-51-51 Nextcloud Vue Style Guide

### 🏁 Checklist

### DIFF
--- a/src/components/NcListItem/NcListItem.vue
+++ b/src/components/NcListItem/NcListItem.vue
@@ -754,6 +754,11 @@ export default {
 		background-color: var(--color-background-hover);
 	}
 
+	&:focus-visible {
+		outline: 2px solid var(--color-main-text);
+		border: 2px solid var(--color-main-background);
+	}
+
 	&-content__wrapper {
 		display: flex;
 		align-items: center;
--- a/src/components/NcListItem/NcListItem.vue
+++ b/src/components/NcListItem/NcListItem.vue
@@ -708,6 +708,7 @@ export default {
 .list-item__wrapper {
 	position: relative;
 	width: 100%;
+	padding: 2px; // for inner focus-visible outline
 
 	&--active,
 	&:active,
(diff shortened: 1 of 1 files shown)


src/components/NcListItem/NcListItem.vue
```diff
@@ -754,6 +754,11 @@ export default {
 		background-color: var(--color-background-hover);
 	}
 
+	&:focus-visible {
+		outline: 2px solid var(--color-main-text);
+		border: 2px solid var(--color-main-background);
+	}
```
ShGKme, line 760: Can we make it inside the list item? like `outline-offset: -2`

There are always 2 annoying issues:
1. Outline of one item overlaps to the previous/next ones
2. Outline of the first/last items is cut by `overflow:scroll/hidden` container

Both issues can be solved by adding margins around. However, the second issue is a problem in a virtual list where we need to take into account this additional 4px. It is not possible with external libraries...

ShGKme: See: https://github.com/nextcloud/spreed/pull/10818

susnux: Adjusted it to have a consistent size, what do you think?

https://github.com/nextcloud-libraries/nextcloud-vue/assets/1855448/3b0da021-b818-4ce1-aaf1-898e3b82e319

ShGKme:
Styles seems to be overwritten by server styles:
https://github.com/nextcloud/server/blob/f778c6c34491031b5708bba786116dd72ec19aff/core/css/styles.scss#L36-L41

susnux:
> Styles seems to be overwritten by server styles:

There is a `:not`: see `:not(#content-vue)`. So they should not override them. But server stylings should be adjusted and probably should use the classname instead.

ShGKme:
> > Styles seems to be overwritten by server styles:
> 
> There is a `:not`: see `:not(#content-vue)`. So they should not override them. But server stylings should be adjusted and probably should use the classname instead.

Does it work for you on server?

JuliaKirschenheuter (approved):
looks good, please check the server styles for the possible conflicts

susnux:
Ready to review I would say :rocket: 


https://github.com/nextcloud-libraries/nextcloud-vue/assets/1855448/e7c61df9-9200-4dfd-a746-d072b7e84357

susnux:
Resolved as part of https://github.com/nextcloud-libraries/nextcloud-vue/pull/5198

PR #4907, opened 2023-12-03
enh(NcRichContenteditable): Add optional visual label for content editable

susnux:
### ☑️ Resolves

For consistent look of Nextcloud components this adds the same label as NcTextField / NcInputField / NcTextArea.

### 🖼️ Screenshots

https://github.com/nextcloud-libraries/nextcloud-vue/assets/1855448/a07e6492-003a-4c50-8226-07aa55d7436e

### 🏁 Checklist

### DIFF
--- a/src/components/NcRichContenteditable/NcRichContenteditable.vue
+++ b/src/components/NcRichContenteditable/NcRichContenteditable.vue
@@ -33,6 +33,7 @@ Try mentioning user @Test01 or inserting emoji :smile
 <template>
 	<div>
 		<NcRichContenteditable
+			label="Write a comment"
 			:value.sync="message"
 			:auto-complete="autoComplete"
 			:maxlength="100"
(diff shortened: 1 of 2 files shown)


susnux:
A use case could be: https://github.com/nextcloud/server/issues/41884

src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -212,28 +213,38 @@ export default {
...
+			@keydown.delete="onDelete"
+			@keydown.enter.exact="onEnter"
+			@keydown.ctrl.enter.exact.stop.prevent="onCtrlEnter"
+			@paste="onPaste"
+			@keyup.stop.prevent.capture="onKeyUp" />
+		<label v-if="label"
+			class="rich-contenteditable__label"
+			:for="id">
+			{{ label }}
+		</label>
```
ShGKme, line 246: I haven't checked all screen readers' support, but according to the standard, it is not valid to make a label for a div. We should use `aria-labelledby` here.

See: https://html.spec.whatwg.org/multipage/forms.html#category-label

susnux: Yes MDN also recommends using labelledby. Changed that.

ShGKme (approved):
Works fine. Was broken to me by JetBrains Grazie browser extension...

marcoambrosini (approved):

ShGKme:
> Note (more for myself, but maybe we should mark it in the changleog): additional `<div>` wrapper around the input breaks styles, currently applied directly with `<NcRichContenteditable class="custom-styling" />`

That is expected (was discussed in the [original issue](https://github.com/nextcloud/server/issues/41884#issuecomment-1837569214)). But formally, it is not a breaking change, because it doesn't change component's public interface and styles in Talk changed internal styles, not positioning styles.

PR #4909, opened 2023-12-04
fix(NcAppSidebar): add focus trap on mobile

ShGKme:
### ☑️ Resolves

* Fix https://github.com/nextcloud/server/issues/41839

On mobile view, the sidebar is full-screen. But it doesn't have a focus trap. So the focus may fall into the main content below the sidebar.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Focus is lost on the main content | It's a trap
sidebar-before | sidebar-after

### 🚧 Tasks

- [x] Add a focus trap on mobile
- [x] Add a header to the trap as it is fully visible

### For discussion

- [ ] Add a navigation toggle button to the trap so a user can open navigation above the focus trap
  - [ ] Alternative — hide toggle navigation button on mobile
- [ ] Close the sidebar by <kbd>ESC</kbd> on mobile

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppNavigationToggle/NcAppNavigationToggle.vue
+++ b/src/components/NcAppNavigationToggle/NcAppNavigationToggle.vue
@@ -22,18 +22,20 @@
  -
  -->
 <template>
-	<NcButton class="app-navigation-toggle"
-		type="tertiary"
-		:aria-expanded="open ? 'true' : 'false'"
-		:aria-label="label"
-		:title="label"
-		aria-controls="app-navigation-vue"
-		@click="toggleNavigation">
-		<template #icon>
-			<MenuOpenIcon v-if="open" :size="20" />
-			<MenuIcon v-else :size="20" />
-		</template>
-	</NcButton>
+	<div class="app-navigation-toggle-wrapper">
+		<NcButton class="app-navigation-toggle"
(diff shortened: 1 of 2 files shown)


ShGKme:
There is an issue with NcActions on mobile view now. But the same issue happens in `NcAppNavigation`, `NcModal`, and other components with the focus trap.

See: https://github.com/nextcloud-libraries/nextcloud-vue/issues/4910

JuliaKirschenheuter (approved):
Works, great!

do we need a focus here https://github.com/nextcloud-libraries/nextcloud-vue/blob/5e8aa0131d6ef992c60df08da6b1f8382fcdefd6/src/components/NcAppSidebar/NcAppSidebar.vue#L368-L369 ?

ShGKme:
> Works, great!
> 
> do we need a focus here
> 
> https://github.com/nextcloud-libraries/nextcloud-vue/blob/5e8aa0131d6ef992c60df08da6b1f8382fcdefd6/src/components/NcAppSidebar/NcAppSidebar.vue#L368-L369
> 
> ?

Yep, but it should also have a button role and have a label

ShGKme:
Rebased onto master

PR #4911, opened 2023-12-04
Enh/Make name longer when actions menu isn't visible

GVodyanov:
Make name longer when actions menu isn't visible by not displaying `app-navigation-entry__utils` when not needed

Useful for https://github.com/nextcloud/mail/issues/7171

### ☑️ Resolves

* Fix https://github.com/nextcloud/mail/issues/7171

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image |  image

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
+++ b/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
@@ -815,6 +815,10 @@ export default {
 		.app-navigation-entry__children {
 			background-color: var(--color-main-background);
 		}
+
+		.app-navigation-entry__utils {
+			display: flex;
+		}
 	}
 
 	// Show the actions on active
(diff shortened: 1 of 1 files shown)


src/components/NcAppNavigationItem/NcAppNavigationItem.vue
```diff
@@ -915,7 +919,7 @@ export default {
 
 /* counter and actions */
 .app-navigation-entry__utils {
-	display: flex;
+   	display: none;
```
ShGKme, line 922: This hides not only the actions but the `#counter` slot as well:
Before | After
---|---
image | image

ShGKme: Also, there is a linter error: `ESLint: Mixed spaces and tabs`. We use tabs for tabulation.

ShGKme (changes requested):
Hi Grigory!

Hiding actions makes sense, but we should not hide the counter. 

To make changes to `stable7` (if it is not a specific v7.x bug), the fix should be applied to the current `master` first and then backported to `stable7` if needed.

PR #4916, opened 2023-12-05
Make modal mask dark if some icons are shown outside

szaimen:
Close https://github.com/nextcloud/server/issues/41859


susnux (approved):
Maybe add a line to the docstring of the dark prop that it will be overridden with `true` if external buttons are used

szaimen:
> Maybe add a line to the docstring of the dark prop that it will be overridden with `true` if external buttons are used

done

PR #4926, opened 2023-12-06
enh(NcInputField): Support numeric values - if numeric also emit numeric

susnux:
### ☑️ Resolves

If the type is set to number and a number is passed as value also emit a number instead of a string.

It always bugs me that you have to write wrappers otherwise, example:
```vue
<template>
<NcInputField type="number" :value="someObject.number" @update:value="(v) => someObject.number = Number(v)" />
</template>
```

vs with this:
```vue
<template>
<NcInputField type="number" :value.sync="someObject.number" />
</template>
```

### 🏁 Checklist

### DIFF
--- a/src/components/NcInputField/NcInputField.vue
+++ b/src/components/NcInputField/NcInputField.vue
@@ -53,7 +53,7 @@ For a list of all available props and attributes, please check the [HTMLInputEle
 						'input-field__input--error': error,
 						'input-field__input--pill': pill,
 					}]"
-				:value="value"
+				:value="value.toString()"
 				v-on="$listeners"
 				@input="handleInput">
 			<!-- Label -->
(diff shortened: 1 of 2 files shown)


Pytal (approved):

ShGKme (approved):
Fine by me, but feels like reimplementing functionality that already exists in v9 because of Vue 3

susnux:
> Fine by me, but feels like reimplementing functionality that already exists in v9 because of Vue 3

Well then remove with v9 and vue 3 :)

src/components/NcInputField/NcInputField.vue
```diff
@@ -53,7 +53,7 @@ For a list of all available props and attributes, please check the [HTMLInputEle
 						'input-field__input--error': error,
 						'input-field__input--pill': pill,
 					}]"
-				:value="value"
+				:value="value.toString()"
 				v-on="$listeners"
 				@input="handleInput">
```
Antreesy, line 58: Just to raise awareness (as i spend quite a time to understand it :see_no_evil:):
- TextField don't listen for `@update:value`, but for `@input`
- As it wasn't modified, TextField always emits a string

Is it expected? As InputField is marked as internal component, we usually avoid to use it directly in the app
What should we do?
- update vue-lib docs, and 'allow' to use InputField?
- expand PR changes to TextField? (not sure about rest wrappers)

PR #4928, opened 2023-12-07
enh(NcDialog): Label dialog nav

Pytal:
### Summary

- For https://github.com/nextcloud/server/issues/41865

### 🏁 Checklist

### DIFF
--- a/src/components/NcDialog/NcDialog.vue
+++ b/src/components/NcDialog/NcDialog.vue
@@ -96,11 +96,14 @@ export default {
 		@close="handleClosed"
 		@update:show="handleClosing">
 		<!-- The dialog name / header -->
-		<h2 class="dialog__name" v-text="name" />
+		<h2 :id="navigationId" class="dialog__name" v-text="name" />
 		<div class="dialog" :class="dialogClasses">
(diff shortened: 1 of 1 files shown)


src/components/NcDialog/NcDialog.vue
```diff
@@ -287,6 +292,11 @@ export default defineComponent({
 		 */
 		const hasNavigation = computed(() => slots?.navigation !== undefined)
 
+		/**
+		 * The unique id of the nav element
+		 */
+		const navigationId = ref(GenRandomId())
```
susnux, line 298: Does this needs to be a reference, could it not be just a plain value?

susnux (approved):

src/components/NcDialog/NcDialog.vue
ShGKme: It could. It seems that making a property in Composition a `ref` is the same common practice as adding all the values to `data` in Options API, even if it is a constant.

(Though, in Composition it is easier (cleaner) to have constant properties).

ShGKme:
Remember to add `enhancement` label for new features and `bug` for fixing bugs, including accessibility or adjustments 😀

PR #4936, opened 2023-12-07
Enh/Make name longer when actions menu isn't visible

GVodyanov:
### ☑️ Resolves

* Fix https://github.com/nextcloud/mail/issues/7171

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
+++ b/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
@@ -314,7 +314,7 @@ Just set the `pinned` prop.
 				<!-- Counter and Actions -->
 				<div v-if="hasUtils && !editingActive"
 					class="app-navigation-entry__utils"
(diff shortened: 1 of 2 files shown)


GretaD (approved):
tested it with mail app :) works

src/components/NcAppNavigationItem/NcAppNavigationItem.vue
```diff
@@ -314,7 +314,7 @@ Just set the `pinned` prop.
 				<!-- Counter and Actions -->
 				<div v-if="hasUtils && !editingActive"
```
ChristophWurst, line 315: would it also be possible to not render the element at all instead of hiding visually?

GVodyanov: The CSS solution uses information on the hover state that vue doesn't have access to, so this is probably the most pragmatic way.

PR #4947, opened 2023-12-11

raimund-schluessler:
### ☑️ Resolves


raimund-schluessler:
@susnux Do you have an idea why cypress complains about typescript here?

```
DevTools listening on ws://127.0.0.1:35477/devtools/browser/566e977d-4408-4f26-b917-7152c8791cb7
[2404:1211/170837.045193:ERROR:object_proxy.cc(590)] Failed to call method: org.freedesktop.portal.Settings.Read: object_path= /org/freedesktop/portal/desktop: unknown error type: 
Missing baseUrl in compilerOptions. tsconfig-paths will be skipped
Your configFile is invalid: /home/runner/work/nextcloud-vue/nextcloud-vue/cypress.config.ts

It threw an error when required, check the stack trace below:

TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts" for /home/runner/work/nextcloud-vue/nextcloud-vue/cypress.config.ts
    at new NodeError (node:internal/errors:406:5)
    at Object.getFileProtocolModuleFormat [as file:] (node:internal/modules/esm/get_format:99:9)
    at defaultGetFormat (node:internal/modules/esm/get_format:142:36)
    at defaultLoad (node:internal/modules/esm/load:120:20)
    at ModuleLoader.load (node:internal/modules/esm/loader:396:13)
    at ModuleLoader.moduleProvider (node:internal/modules/esm/loader:278:56)
    at new ModuleJob (node:internal/modules/esm/module_job:65:26)
    at ModuleLoader.#createModuleJob (node:internal/modules/esm/loader:290:17)
    at ModuleLoader.getJobFromResolveResult (node:internal/modules/esm/loader:248:34)
    at ModuleLoader.getModuleJob (node:internal/modules/esm/loader:229:17)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async ModuleLoader.import (node:internal/modules/esm/loader:315:23)
    at async importModuleDynamicallyWrapper (node:internal/vm/module:431:15)
```

raimund-schluessler:
> @susnux Do you have an idea why cypress complains about typescript here?

Same issue as in https://github.com/nextcloud-libraries/nextcloud-vue/pull/4930.

src/components/NcRichContenteditable/NcRichContenteditable.vue
ShGKme, line 1: We should add class prop here as in `NcInputField`, otherwise passed class is not set on the root element.

ShGKme: And this component still uses removed `{ ...this.$listeners }`

raimund-schluessler: Yes, the `NcRich*` components are not migrated yet. That's why I didn't touch them here. PR for migration is in #4842.

ShGKme: Oh, sorry, I thought only NcRichText wasn't migrated...

susnux:
> @susnux Do you have an idea why cypress complains about typescript here?

No idea why this is caused by this changes but this means somehow ts-node is not called correctly. Or better: There is a mixup of ESM and CJS.
I guess because our `package.json` implicitly states commonjs but the configuration file is ESM, thus for ts-node it should be called `.mts` but cypress does not support this extension.

So there are two options:
1. Move configuration to `.mjs`
2. Make the package ESM by default (add `type: "module"` to the package.json).

For the next version / the `next` branch I would prefer the second solution, but when applied on current master then the first one so it is not a breaking change.

susnux (approved):

ShGKme:
> For the next version / the `next` branch I would prefer the second solution, but when applied on current master then the first one so it is not a breaking change.

What may break the second one?

raimund-schluessler:
I would propose to fix and merge #4930 first. Then it will work here as well.

susnux:
> What may break the second one?

I can not think of any but it changes the the package type thus I thought it might be considered breaking.

raimund-schluessler:
I recreated the PR without the problematic update to `typescript@5.3.3` (which isn't merged into `master` yet anyway) and pulled in all latest changes from `master`. I think it should be good to merge now.

ShGKme (approved):

PR #4951, opened 2023-12-11
fix(package): specify exports for composables

ShGKme:
### ☑️ Resolves

* See: https://github.com/nextcloud/spreed/issues/10927


nickvergessen (approved):
🙈

raimund-schluessler (approved):

PR #4953, opened 2023-12-12
fix(NcActions): intercept into current focus trap stack

### LINKED ISSUE #4910: `NcActions` doesn't work inside an element with a focus trap in a general case
`NcActions` handles `keydown` on the menu element. It works only when a menu item has a focus.

Because `NcActions` renders in the `body` by default, when it's used in a component with a focus trap, the focus trap blocks the focus and breaks keyboard navigation.

In a specific case, it can be adjusted by setting `container`. But not in a general case.

The current idea — provide a container for mounting from an element with a focus trap (overlapping elements like Navigation, Sidebar, Modal), and not to mount in the body by default.

ShGKme:
### ☑️ Resolves

* Resolves: #4910

When the component has its own focus trap, then it is managed by global trap stack by focus-trap.

However if the component has no focus trap and is used inside another focus trap - there is an issue. By default popover content is rendered in body or other container, which is likely outside the current focus trap containers. It results in broken behavior from focus-trap.

We need to pause all the focus traps for opening popover and then unpause them back after closing.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Header Menu | .
before-focus-trap-1 | after-focus-trap-1
Sidebar multi-level | .
before-focus-trap-2 | after-focus-trap-2

Focus trap still works.

Supports multi-level focus traps (header menu focus trap inside sidebar trap):

focus-trap-3

### 🏁 Checklist

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -813,6 +813,7 @@ p {
 import NcButton from '../NcButton/index.js'
 import NcPopover from '../NcPopover/index.js'
 import GenRandomId from '../../utils/GenRandomId.js'
+import { getTrapStack } from '../../utils/focusTrap.js'
 import { t } from '../../l10n.js'
 
 import Vue, { computed } from 'vue'
@@ -1009,6 +1010,7 @@ export default {
 			 * @type {'menu'|'navigation'|'dialog'|'tooltip'|''}
 			 */
 			actionsMenuSemanticType: '',
+			externalFocusTrapStack: [],
 		}
 	},
 
@@ -1035,6 +1037,10 @@ export default {
 
 			this.opened = state
 		},
+
+		opened() {
+			this.intersectIntoCurrentFocusTrapStack()
+		},
 	},
 
 	methods: {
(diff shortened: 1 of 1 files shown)


ShGKme:
Could be risky. A good review is appreciated 👀

JuliaKirschenheuter (commented):
sequence is unfortunately not fully correct ;( -> both directions: forwards and backwards

Peek 2024-01-26 08-15

JuliaKirschenheuter:
Doesn't work on Photos well ;( -> seems to be a regression, works on try.extcloud 

Peek 2024-01-26 09-01

Peek 2024-01-26 09-03

ShGKme:
> sequence is unfortunately not fully correct ;( -> both directions: forwards and backwards

I cannot reproduce. The `gif` looks exactly how it was before this PR. Could you check that `npm link` and rebuild was successful? 

On what page what it recorded?

ShGKme:
The issue happens when there are any inline actions. They are also counted in menu actions when determining what kind of menu we have.

Preparing another PR.

JuliaKirschenheuter (approved):
works, great!

susnux (approved):
works

PR #4956, opened 2023-12-13
fix(NcListItem): Ensure list item does not overflow wrapper

susnux:
### ☑️ Resolves

`width` is set to 100% but if `box-sizing` is not set to `border-box` it will not include the border and the padding.
This causes the list item to overflow its parents in some cases, you can for example see this in the new unified search of Nextcloud 28 (there are even more places).

To fix this simply set `box-sizing` to `border-box` of the list item.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screen Shot 2023-12-13 at 15 09 51|Screenshot 2023-12-13 at 15-09-32 Dashboard - Nextcloud

### 🏁 Checklist

### DIFF
--- a/src/components/NcListItem/NcListItem.vue
+++ b/src/components/NcListItem/NcListItem.vue
@@ -738,6 +738,7 @@ export default {
 
 // NcListItem
 .list-item {
+	box-sizing: border-box;
 	display: block;
 	position: relative;
 	flex: 0 0 auto;


ShGKme (approved):
I wonder if we can just make `box-sizing: border-box;` a global default...

Pytal (approved):

PR #4957, opened 2023-12-13
Make package ESM by default

susnux:
### ☑️ Resolves

* <del>Supersedes #4930</del>
* Fixes the issue in #4947

This makes the package using ESM by default (`type: "module"`) and adjusts all scripts used.
Some stuff like stylelint and eslint do not work correctly with ESM yet, so they are kept as commonjs scripts.

### 🏁 Checklist


susnux:
Basically this works and simplifies the code base, but it does not fix the Typescript issue

ShGKme (approved):

raimund-schluessler (approved):

susnux:
Do we want this for v8 or should I change the target branch to `next` for v9?

ShGKme:
> Do we want this for v8 or should I change the target branch to `next` for v9?

I think this is fine for v8. It should not affect the distributive and works fine on Server. Probably, we should test on more apps before just in case, e.g., Talk.

ShGKme:
Talk works including calls

skjnldsv:
> I think this is fine for v8. It should not affect the distributive and works fine on Server. Probably, we should test on more apps before just in case, e.g., Talk.

Switching package types is a breaking change, even if that does not break on our ecosystem :)
So v9 it is pls

ShGKme:
> Switching package types is a breaking change, even if that does not break on our ecosystem :)

Why?

skjnldsv:
> What does this change break?

The module resolution changes.
Pre-processors will now treat this package as `module`, not saying this will cause issues on our ecosystem, but I'm saying this is no patch release (and arguably, no minor either)

susnux:
Merged into Next for reasons mentioned by @skjnldsv

PR #4958, opened 2023-12-13
enh(aria): changed check color to adhere to 3:1 ratio

emoral435:
### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 Summary
Changes the check color depending on the background color chosen to either be white or black to achieve the 3:1 contrast ratio for ARIA - radio input button still has white border, so up to reviewers to decide if that should change

### 🏁 Checklist

- [X] ⛑️ Tests are included or are not applicable (N/A)
- [X] 📘 Component documentation has been extended, updated or is not applicable (N/A)

### DIFF
--- a/src/components/NcColorPicker/NcColorPicker.vue
+++ b/src/components/NcColorPicker/NcColorPicker.vue
@@ -173,7 +173,7 @@ export default {
 						:style="{ backgroundColor: color }"
 						class="color-picker__simple-color-circle"
 						:class="{ 'color-picker__simple-color-circle--active' : color === currentColor }">
-						<Check v-if="color === currentColor" :size="20" />
+						<Check v-if="color === currentColor" :size="20" :fill-color="contrastColor" />
 						<input type="radio"
 							class="hidden-visually"
 							:aria-label="name"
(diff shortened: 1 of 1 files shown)


emoral435:
For context - the choice between black and white was a heuristic, and I checked, all the buttons pass the 3:1 ratio, if not higher!

susnux:
Nice! But please fix the linting issues.

emoral435:
Fixed linting error : )

src/components/NcColorPicker/NcColorPicker.vue
```diff
@@ -386,6 +386,40 @@ export default {
...
+		 * gets the contrast color to use for the user chosen color
+		 *
+		 * @return {string} representing the hex value to change the check color to contrast with the background color
+		 *
+		 */
+		 contrastColor() {
+			const black = '#000000'
+			const white = '#FFFFFF'
+			return (this.calculateLuma(this.currentColor) > 0.5) ? black : white
+		},
```
ShGKme, line 422: This should be a computed property. It calculates a value that depends only on the component's state.

Using it as a method in the template makes the template less declarative and requires to check that this method is good in a performance aspect (it is called each time the component renders) and has no side effects.

emoral435: Agreed! Moving it to be a computed property, thanks!

susnux (approved):

Pytal (approved):

ShGKme:
Remember to add `enhancement` label for new features and `bug` for fixing bugs, including accessibility or adjustments 😀

PR #4959, opened 2023-12-14
fix(NcModal): apply box sizing to modal content

Antreesy:
### ☑️ Resolves

* Fix box-sizing property for nested components:
  * If NcModal is mounted inside of `#content-vue` element, as all other components, it inherits style from NcContent (see here
https://github.com/nextcloud-libraries/nextcloud-vue/pull/1026/commits/33f4af39f14648429650db6ca3319786188ec3a6)
  * But by default, NcModal is mounted in `body` element, where styles are not overwritten, and all element apart from [`<table>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/table), [`<select>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select), [`<button>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button) and  [`<input>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) elements have `box-sizing: content-box by [default](https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing)
  * This will affect `<div class="modal-container">`, `<div class="modal-container__content">`, and all that goes to `<slot />`

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist


ShGKme (approved):
I wonder if this counts as a breaking change.

Probably may break some modals that had content with styles for default border-box...

But I'd make this change, having different styles depending on mount point isn't good.

And +1 thing to think about global `box-sizing: border-box;` (https://github.com/nextcloud-libraries/nextcloud-vue/pull/4956 👀)

susnux:
> Probably may break some modals that had content with styles for default border-box...

I think every more complex modal will be broken as most of them are mounted to the body. So this needs to be considered.

Antreesy:
Maybe we put in on the `next` branch, as we should prepare a migration guide for it anyway? I'm assuming, that more or less complex modals including it already, so with moving to Vue 3 it wouldn't be a huge task to check styles

skjnldsv (approved):

skjnldsv (commented):
> @skjnldsv Does your approval mean we may not consider it as a breaking change and merge into v8? 😀

Nah, should be for v9 then

src/components/NcModal/NcModal.vue
```diff
@@ -937,6 +937,10 @@ export default {
 	width: 100%;
 	height: 100%;
 
+	:deep(*) {
+		box-sizing: border-box;
+	}
+
```
raimund-schluessler, line 943: While I think it's good to have this globally, I somehow feel this is not the correct place to put this rule. It will affect all components, but it's located in `NcModal`. Don't we have a global CSS file or so?

Antreesy: AFAIK, this rule was removed from global CSS styles and added to NcContent instead. This PR just aligns styles for all components, whether DOM elements are mounted inside NcContent or outside of it

ShGKme: > While I think it's good to have this globally, I somehow feel this is not the correct place to put this rule. It will affect all components, but it's located in `NcModal`. Don't we have a global CSS file or so?

But it applies styles only to the content of `.modal-wrapper`, not truly globaly

raimund-schluessler: > > While I think it's good to have this globally, I somehow feel this is not the correct place to put this rule. It will affect all components, but it's located in `NcModal`. Don't we have a global CSS file or so?
> 
> But it applies styles only to the content of `.modal-wrapper`, not truly globally

Ah, yes, sorry. I missed that it is `scoped`. Fine then.

raimund-schluessler (approved):

PR #4972, opened 2023-12-19
enh(NcAppNavigationSettings): added aria label prop

emoral435:
### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 Tasks / Summary
Added a prop (with a default value) so that we have better control for aria labels within the NcAppNavigationSetting component

### 🏁 Checklist

- [x] ⛑️ Tests are included or are not applicable (Not Applicable)
- [ ] 📘 Component documentation has been extended, updated or is not applicable (Updating / Updated)

### DIFF
--- a/src/components/NcAppNavigationSettings/NcAppNavigationSettings.vue
+++ b/src/components/NcAppNavigationSettings/NcAppNavigationSettings.vue
@@ -29,6 +29,7 @@
 				type="button"
 				:aria-expanded="open ? 'true' : 'false'"
 				aria-controls="app-settings__content"
+				:aria-label="inputAriaLabel"
 				@click="toggleMenu">
 				<Cog class="settings-button__icon" :size="20" />
 				<span class="settings-button__label">{{ name }}</span>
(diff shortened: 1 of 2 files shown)


ShGKme:
Linked issue was a duplication of 
- https://github.com/nextcloud/server/issues/41901

Fixed in 
- https://github.com/nextcloud-libraries/nextcloud-vue/pull/4932

The problem was that there was a `aria-label` was redundant as the button has a text value already.

emoral435:
Yep, just saw this right now! Closed

src/components/NcAppNavigationSettings/NcAppNavigationSettings.vue
```diff
@@ -29,6 +29,7 @@
 				type="button"
 				:aria-expanded="open ? 'true' : 'false'"
 				aria-controls="app-settings__content"
+				:aria-label="inputAriaLabel"
 				@click="toggleMenu">
 				<Cog class="settings-button__icon" :size="20" />
 				<span class="settings-button__label">{{ name }}</span>
```
ShGKme, line 35: Note: this button has a text value already in `span.settings-button__label {{ name }}`.

Element should not have `aria-label` and direct text at the same time, otherwise it has 2 different sources of label.

PR #4976, opened 2023-12-19
[next] feat(NcRich*): migrate to vue 3

raimund-schluessler:
### ☑️ Resolves

* Migrate `NcRich*` components to vue 3.


raimund-schluessler:
@Antreesy Thanks a lot for taking care of the remaining issues! Markdown support works now and I couldn't find any more issues. From my side, this PR is now ready to be merged.

When this gets merged, we have migrated *all* components to vue3 :rocket:

I will squash the commits into one after the PR is approved.

src/components/NcRichContenteditable/NcRichContenteditable.vue
ShGKme, line 1: One small breaking change — because of proxying `$attrs`, setting `class` on the component doesn't set class on the root element now. We can fix it by adding `class` to `props` and binding it manually.

ShGKme:
> When this gets merged, we have migrated _all_ components to vue3 🚀

Except `NcReferencePicker` 😶

raimund-schluessler:
> > When this gets merged, we have migrated _all_ components to vue3 🚀
> 
> Except `NcReferencePicker` 😶

This is included in this PR as well, since it sits in the `NcRichText` directory. I don't have an app that uses the picker though. So if one would test it, it would be great. But in doubt, it will be tested once the first app that uses it gets migrated. The first release of `nextcloud-vue@next` will be an alpha version meant for testing anyways.

src/components/NcRichContenteditable/NcRichContenteditable.vue
raimund-schluessler: Will do so.

src/components/NcRichText/NcReferencePicker/referencePickerModal.js
```diff
@@ -21,14 +21,12 @@ export async function getLinkWithPicker(providerId = null, isInsideViewer = unde
...
-				initialProvider,
-				isInsideViewer,
-			},
-		}).$mount(modalElement)
+		const view = createApp(NcReferencePickerModal, {
+			initialProvider,
+			isInsideViewer,
+		}).mount(modalElement)
 
+		// TODO adjust this for vue3
```
ShGKme, line 29: I meant this place. `view.$on` doesn't work

src/components/NcRichText/NcReferencePicker/referencePickerModal.js
```diff
@@ -21,14 +21,12 @@ export async function getLinkWithPicker(providerId = null, isInsideViewer = unde
...
 		view.$on('cancel', () => {
 			view.$destroy()
 			reject(new Error('User cancellation'))
```
ShGKme, line 32: I meant this place. `view.$on` doesn't work

raimund-schluessler:
@ShGKme I tried to adjust `referencePickerModal.js` in https://github.com/nextcloud-libraries/nextcloud-vue/pull/4976/commits/c24889549aa8d128195f6fd6e2b7e012fef6a0ca but I got stuck, since the `$destroy` method was removed with the following statement:

"`$destroy` instance method. Users should no longer manually manage the lifecycle of individual Vue components.", see https://v3-migration.vuejs.org/breaking-changes/#removed-apis

Do you have an idea how to fix `referencePickerModal` correctly?

ShGKme:
@raimund-schluessler there is `unmount` method in `createApp`'s app: https://vuejs.org/api/application.html#app-unmount

It can be used as a replacement for instance's `$destroy`

raimund-schluessler:
> @raimund-schluessler there is `unmount` method in `createApp`'s app: https://vuejs.org/api/application.html#app-unmount
> 
> It can be used as a replacement for instance's `$destroy`

Thanks for the hint. But I am a bit at a loss how to call it from  within `onCancel`. Would this work?
```js
		const view = createApp(NcReferencePickerModal, {
			initialProvider,
			isInsideViewer,
			onCancel() {
				destroyView()
				reject(new Error('User cancellation'))
			},
			onSubmit(link) {
				destroyView()
				resolve(link)
			},
		}).mount(modalElement)

		const destroyView = () => {
			view.unmount()
		}
```

src/components/NcRichText/NcReferencePicker/referencePickerModal.js
raimund-schluessler: If should be fixed now and works in the docs.

src/components/NcRichContenteditable/NcRichContenteditable.vue
raimund-schluessler: Done.

raimund-schluessler:
> > When this gets merged, we have migrated _all_ components to vue3 🚀
> 
> Except `NcReferencePicker` 😶

I applied the review comments and fixed another issue I found. I think it should work well now, a new review would be welcome.

src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -213,7 +213,7 @@ export default {
 </docs>
 
 <template>
-	<div class="rich-contenteditable">
+	<div class="rich-contenteditable" :class="class">
```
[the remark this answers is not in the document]
ShGKme: `$props.class` should work fine, so it won't mix it with `class` keyword

ShGKme, line 216:
```suggestion
	<div class="rich-contenteditable" :class="$props.class">
```

src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -362,12 +361,20 @@ export default {
 			type: Boolean,
 			default: true,
 		},
+
+		/**
+		 * CSS class to apply to the root element.
+		 */
+		class: {
+			type: String,
```
ShGKme, line 369:
```suggestion
		class: {
			type: [String, Array, Object],
```

src/components/NcRichText/NcReferencePicker/referencePickerModal.js
```diff
@@ -21,21 +21,22 @@ export async function getLinkWithPicker(providerId = null, isInsideViewer = unde
...
-				isInsideViewer,
+		const view = createApp(NcReferencePickerModal, {
+			initialProvider,
+			isInsideViewer,
+			onCancel() {
+				destroyView()
+				reject(new Error('User cancellation'))
+			},
+			onSubmit(link) {
+				destroyView()
```
ShGKme, line 32: Direct access to `view` should work here without function

```suggestion
			onSubmit(link) {
				view.unmount()
```

src/components/NcRichContenteditable/NcRichContenteditable.vue
raimund-schluessler: Done.

src/components/NcRichText/NcReferencePicker/referencePickerModal.js
raimund-schluessler: Indeed it does. Fixed.

src/components/NcRichContenteditable/NcRichContenteditable.vue
raimund-schluessler: Thanks for the idea. Works.

ShGKme (approved):
Everything seems to work fine.

Thanks for all the work 💙
And for resolving a large number of my comments 🙈

The last component in Vue 3 migration 🚀

src/components/NcRichContenteditable/NcRichContenteditable.vue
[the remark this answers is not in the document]
Antreesy: May we implement it also for master to avoid [styles breaking](https://github.com/nextcloud-libraries/nextcloud-vue/pull/4907#issuecomment-1856039941)?

Antreesy (approved):
LGTM from NcRich component general usage + markdown (apart from the code where I'm co-author)

src/components/NcRichContenteditable/NcRichContenteditable.vue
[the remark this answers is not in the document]
ShGKme: In `master` in Vue 2 it works the same by default, because `class` and `style` attributes don't count as `$attrs` and are not managed by `inheritAttrs`. So in master setting `class` on `<NcRichContenteditable>` sets it on the root anyway.

The change in this PR resolves breaking change between `master` and `next`, moving the class back to root.

susnux (approved):
Nice!

susnux:
Unrelated, but does anyone knows where this is coming from?
`Tribute was already bound to DIV`

raimund-schluessler:
> Everything seems to work fine.

Great.

> Thanks for all the work 💙 And for resolving a large number of my comments 🙈

Thanks for your comments, though. They helped a lot!

> The last component in Vue 3 migration 🚀

I think it's time for an alpha release, no? 🙂

PR #4983, opened 2023-12-21
enh(NcContent): Add skip content buttons

susnux:
### ☑️ Resolves

* Fix https://github.com/nextcloud/server/issues/41696

This adds skip content buttons for the Vue app when mounting `NcContent`.
It also handles the app navigation if it is not available (meaning do not add that button if there is no app navigation).
And it opens the app navigation when skipping to it.

### 🏁 Checklist


src/components/NcContent/NcContent.vue
```diff
@@ -72,18 +74,68 @@ It includes the Navigation, the App content and the Sidebar.
...
 	</div>
 </template>
 
 <script>
+import { emit } from '@nextcloud/event-bus'
+import { t } from '../../l10n.js'
+
+import NcButton from '../NcButton/NcButton.vue'
+// TODO: This is built-in for vue3 just drop the import
+import Teleport from 'vue2-teleport'
```
ShGKme, line 98: I'm a bit worried about this dependency. It has just 8 commits when it was created and doesn't seem to be actively used or maintained.

There is `vue-simple-portal` from the author of famous `vue-portal` library and the initiator of [adding a Teleport component to Vue](https://www.github.com/vuejs/rfcs/pull/112).

It has a little bit different interface (prop is called `selector` instead of `to`), but it is used by thousands of other packages and looks more reliable. 

https://github.com/LinusBorg/vue-simple-portal

src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -93,6 +93,9 @@ export default {
 		NcAppNavigationToggle,
 	},
 
+	// Injected from NcContent
+	inject: ['setAppNavigation'],
```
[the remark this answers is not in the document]
susnux: Well then either the files app is wrong or our documentation

> It _MUST_ be used as the main wrapper of your app.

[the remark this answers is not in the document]
susnux: But yes we can make this a bit more defensive

src/components/NcContent/NcContent.vue
```diff
@@ -72,18 +74,68 @@ It includes the Navigation, the App content and the Sidebar.
...
+	name: 'NcContent',
+	components: {
+		NcButton,
+		Teleport,
+	},
+	provide() {
+		return {
+			setAppNavigation: this.setAppNavigation,
+		}
+	},
```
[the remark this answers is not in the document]
susnux: I agree! This helps to understand what is happening

src/components/NcContent/NcContent.vue
susnux: Both packages have nearly the same amount of weekly downloads and both have their last real update 3 years ago.

src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -161,6 +166,14 @@ export default {
 		 * @param {boolean} [state] set the state instead of inverting the current one
 		 */
 		toggleNavigation(state) {
+			// Early return if alreay in that state
+			if (this.open === state) {
+				emit('navigation-toggled', {
+					open: this.open,
+				})
+				return
+			}
```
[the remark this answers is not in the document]
susnux: The behavior is not changes it just early returns, otherwise you have to wait for animation timeout.
Needed for users to know the state of the app navigation.

But it is no longer needed to be part of this PR (in the first version I waited for that event) but for e.g. screen readers it does not make any difference.

JuliaKirschenheuter:
But those scenarios https://github.com/nextcloud-libraries/nextcloud-vue/pull/4983#pullrequestreview-1793201530 are working for "skip to main content" well

susnux:
> Check "Skip to app navigation" one time -> navigate a bit -> check "Skip to app navigation" second time -> focus remains on a skip button and nothing happens.

Ah I see when the location hash is already set to that ID then the browser will not update the focus. I fixed this now :heavy_check_mark:

susnux:
> Check "Skip to app navigation" -> go to the app navigation and try to navigate backwards (Shift + Tab) -> navigation will move to the header, not to the app navigation

What do you mean? If you check "skip to navigation" then the focus should be on the navigation. If you press shift-tab then the next element in reverse order is focused (probably the settings button in the app menu?).

src/components/NcContent/NcContent.vue
ShGKme: Weekly download stats are shown only for the last year, when Vue 3 was released already for quite a long time. I guess most of the users of `vue2-teleport` searched for "Vue 2 teleport" as an alternative from Vue 3. But even if compare only weekly downloads last year, Linus' are a little bit more.

The full version (`portal-vue`) still has almost half a million downloads per week in comparison with 10-13k.

I'd consider a library from a Vue team member and author of well-known `portal-vue` less risky.

ShGKme:
> 2\. Check "Skip to app navigation" -> go to the app navigation and try to navigate backwards (Shift + Tab) -> navigation will move to the header, not to the app navigation

"Skip buttons" supposed to be one way — to skip the header or navigation. But user doesn't need to return to them on backward navigation.

In screen readers a user can return to the beginning with a special shortcut (CTRL+Home usually)

src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -78,21 +78,29 @@ emit('toggle-navigation', {
...
 	name: 'NcAppNavigation',
 
 	components: {
 		NcAppNavigationToggle,
 	},
 
+	// Injected from NcContent
+	inject: {
+		setHasAppNavigation: {
+			default: () => () => Vue.util.warn('NcAppNavigation is not mounted inside NcContent, this is probably an error.'),
```
ShGKme, line 99: If we show a warning here, we should add something to the Files app to not show the warning

src/components/NcContent/NcContent.vue
susnux: I am fine with both, for `next` it is then just one line more to adjust. I will change this.

src/components/NcAppNavigation/NcAppNavigation.vue
susnux: I would say it depends. Is there any good reason to not use `NcContent`?
If yes: Then we can remove the warning and adjust the docs.
If not: We should adjust files to use it.

src/components/NcContent/NcContent.vue
susnux: I adjusted it but this will require a bit more work as that library injects a wrapper around the teleported content, thus the server styling is broken.

susnux:
@nextcloud-libraries/designers this is unrelated to this PR (also on current master) but maybe we can take the opportunity and improve it.
When having the app navigation on an app the skip content buttons are hard to see because they are colored primary like the active navigation element element. Maybe we should add some background to the buttons to make them more visual distinguishable?

With background | current
---|---
Screenshot_20231221_191558|Screenshot_20231221_191715
Screenshot_20231221_191609|Screenshot_20231221_191705

marcoambrosini:
Good idea @susnux! Then I would suggest making it even bigger and more informative. I sketched out a suggestion. 
I can provide SVGs for the images of the focused parts. What do you think @nextcloud-libraries/designers ?
Screenshot from 2023-12-22 08-49-16

susnux:
@marcoambrosini that looks really amazing! Should we directly implement this or first fix this issue here and the improve when the SVGs are ready?

JuliaKirschenheuter:
@marcoambrosini would it be possible to backport so large improvement to stable28?

ShGKme:
> @marcoambrosini would it be possible to backport so large improvement to stable28?

I think this is not large as this part is visible only on initial keyboard navigation.

marcoambrosini:
Thanks @susnux :) 
I can finalize the SVGs the 27th first thing in the morning. Unfortunately, I'm AFK for today

src/components/NcAppNavigation/NcAppNavigation.vue
skjnldsv: There isn't, this is the only way we ask Devs to build Vue apps

JuliaKirschenheuter:
Hi @marcoambrosini 

could you please provide us a SVGs? 🙏

marcoambrosini:
Sorry for the delay, I lost access to mi gh account and missed this. 
Here's how it'd look in the end 

Screenshot from 2023-12-29 16-59-17

And here are the svgs

[Keyboard Navigation help.zip](https://github.com/nextcloud-libraries/nextcloud-vue/files/13791874/Keyboard.Navigation.help.zip)

JuliaKirschenheuter (approved):
Works, thank you!

susnux:
> And here are the svgs

@marcoambrosini I added the images to the skip buttons. It now looks like this:

https://github.com/nextcloud-libraries/nextcloud-vue/assets/1855448/7db6b93f-bb61-47eb-bcc2-eb1aa38bbe7c

JuliaKirschenheuter (commented):
Screenshot from 2024-01-11 17-51-18

something went wrong ;(

JuliaKirschenheuter:
Please remove heading `<h2>`, it should not be there (you could use `<span>` with same styles instead)

Screenshot from 2024-01-12 10-17-54

JuliaKirschenheuter (approved):
skip links are working fine. Can't see SVG's

marcoambrosini:
It looks like the svg is getting resized?

susnux:
> It looks like the svg is getting resized?

Forgot to set the size correctly, done now :heavy_check_mark:

susnux:
Screenshot_20240118_175720

ShGKme:
Now I can see the SVG, but button styles have changed (border styles on focus).

not sure if it is something on my side or not...

Before | After
---|---
image | image

ShGKme (approved):

marcoambrosini (commented):
Nice, just a couple of comments about the buttons:
- I would invert the order so that first comes navigation and then content
- I would make both of them tertiary

ShGKme:
> * I would make both of them tertiary

Maybe secondary is better here?

ShGKme:
Primary | Secondary | Tertiary
---|---|---
image | image | image
image | image | image

marcoambrosini:
@ShGKme since one f the two will always have a border, I don't think it's needed. But I'm fine with either.

susnux:
> Nice, just a couple of comments about the buttons

I pushed a commit to resolve both points.


https://github.com/nextcloud-libraries/nextcloud-vue/assets/1855448/da1e2763-fd54-4e02-92ae-18b3be2cb07a

ShGKme (approved):

PR #4999, opened 2023-12-27
fix(NcCheckboxRadioSwitch): fix shift+click on firefox

skjnldsv:
Fix https://github.com/nextcloud/server/issues/42202
Ref https://bugzilla.mozilla.org/show_bug.cgi?id=559506

I'm not sure, if anyone else have a better idea, please suggest it! :pray: :smiling_face_with_tear:

### DIFF
--- a/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
+++ b/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
@@ -277,6 +277,7 @@ export default {
 		v-on="isButtonType ? listeners : null">
 		<input v-if="!isButtonType"
 			:id="id"
+			:aria-labelledby="!isButtonType && !ariaLabel ? `${id}-label` : null"
 			:aria-label="ariaLabel || undefined"
 			class="checkbox-radio-switch__input"
 			:disabled="disabled"
@@ -293,7 +294,8 @@ export default {
 			:button-variant="buttonVariant"
 			:is-checked="isChecked"
 			:loading="loading"
-			:size="size">
+			:size="size"
+			@click.native="onToggle">
 			<template #icon>
 				<!-- @slot The checkbox/radio icon, you can use it for adding an icon to the button variant -->
 				<slot name="icon" />
--- a/src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
+++ b/src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
@@ -21,6 +21,10 @@
 -->
 
 <template>
+	<!--
+		label can't be used here because of shift+click firefox bug
+		https://bugzilla.mozilla.org/show_bug.cgi?id=559506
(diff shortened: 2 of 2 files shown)


emoral435 (approved):
I actually primarily use Firefox, and am working on an issue that needs this resolved, so this is perfect. LGTM, works on my dev env, and keeps the aria label logic clean :)

ShGKme:
> Ref https://bugzilla.mozilla.org/show_bug.cgi?id=151142

Isn't it this bug?
https://bugzilla.mozilla.org/show_bug.cgi?id=559506

src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
```diff
@@ -276,7 +276,7 @@ export default {
 		:type="isButtonType ? 'button' : null"
 		v-on="isButtonType ? listeners : null">
 		<input v-if="!isButtonType"
-			:id="id"
+			:aria-labelledby="!isButtonType && !ariaLabel ? id : null"
```
ShGKme, line 279: Isn't it a breaking change?

This component has a prop `id` which still has description `Unique id attribute of the input`.

But not an instance with `id` set won't render an `input` with this `id` but a `span`.

Example of breaking change:
```html
<!-- label doesn't work anymore -->
<label for="my-id">Custom label outside</label>

<NcCheckboxRadioSwitch id="my-id" />

<script>
// Doesn't work
document.getElementById('my-id').focus()
</script>
```

Maybe examples are edge cases, but they are supposed to work according to the current description `id attribute of the input`

And the same with `NcCheckboxContent` where the meaning of `id` prop has changed.

At least I'd change the prop descriptions.

---

As an alternative, I may propose to not use `id` from prop here, but generate a new local `id` and pass it to the `NcCheckboxContent` as a new prop.

skjnldsv:
> > Ref [bugzilla.mozilla.org/show_bug.cgi?id=151142](https://bugzilla.mozilla.org/show_bug.cgi?id=151142)
> 
> Isn't it this bug? [bugzilla.mozilla.org/show_bug.cgi?id=559506](https://bugzilla.mozilla.org/show_bug.cgi?id=559506)

Wrong paste, my bad!

src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
skjnldsv: Ah, right, I used the wrong id maybe, I'll have a second look this morning! Thanks @ShGKme  :)

skjnldsv: image

Here we go

ShGKme (approved):

ShGKme:
But I'd add a comment in the code with a link to the bug, because it seems non-obvious why a native element is not used here.

PR #5001, opened 2023-12-28
fix(NcCheckboxRadioSwitch): improve rendering and prevent unecessary elements

skjnldsv:
- Make the icon inert
- Do not render the content text if undefined

### DIFF
--- a/src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
+++ b/src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
@@ -37,7 +37,8 @@
 				'checkbox-content__icon--checked': isChecked,
 				[iconClass]: true
 			}"
-			:aria-hidden="true">
+			:aria-hidden="true"
+			inert>
 			<!-- @slot The checkbox/radio icon, you can use it for adding an icon to the button variant
 					@binding {bool} checked The input checked state
 					@binding {bool} loading The loading state
@@ -52,7 +53,7 @@
 			</slot>
 		</span>
 
-		<span :class="['checkbox-content__text', textClass]">
+		<span v-if="$slots.default" :class="['checkbox-content__text', textClass]">
 			<!-- @slot The checkbox/radio label -->
 			<slot />
 		</span>


ShGKme:
I don't understand... Icon is neither focusable nor interactive anyway, does `inert` change anything here?

Is it an alternative or a part of https://github.com/nextcloud-libraries/nextcloud-vue/pull/4999 ?

emoral435:
From what I am seeing, the `inert` attribute would just be a preventative in case the inserted slot is indeed focusable, no? It is not a breaking change, more of a preventative one from what I gather. As of now, I do not believe it hurts to add it on

emoral435 (approved):

ShGKme:
> the `inert` attribute would just be a preventative in case the inserted slot is indeed focusable, no?

It would be an a11y issue if there was an interactive focusable and yet inert element...

skjnldsv:
> It would be an a11y issue if there was an interactive focusable and yet inert element...

That's the point, sanitising what we allow in those slots

src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
```diff
@@ -51,7 +52,7 @@
 			</slot>
 		</span>
 
-		<span :class="['checkbox-content__text', textClass]">
+		<span v-if="$slots.default" :class="['checkbox-content__text', textClass]">
```
raimund-schluessler, line 55: I think this is not necessary (although it probably doesn't hurt either), since we have `display: none;` already if the span is empty:
https://github.com/nextcloud-libraries/nextcloud-vue/blob/8aec0a1f1848efb229401e2796537ba9fad8d449/src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue#L225-L228

Or do I miss something here?

skjnldsv: Just cleaner approach in general to really not have an extra unnecessary element.
It can quickly adds up in lists (like Files) :)

ShGKme: I'd call it a good practice to always check for slots on slot wrapper elements.

raimund-schluessler (approved):

PR #5007, opened 2023-12-29

raimund-schluessler:
### ☑️ Resolves


ShGKme (approved):

raimund-schluessler:
@susnux @skjnldsv Could we get this merged?

raimund-schluessler:
> Can you publish the first version of `9.0.0-alpha.0` to npmjs.org afterwards? Or isn't the package in a usable state yet?

Yes, we should do this. But I will leave this up to someone from the Nextcloud company.

`9.0.0-alpha.0` is usable, all components are migrated and it is definitely good enough to get the app migrations going. There are still a few things to do before an alpha release, but these are minor issue that should be quickly resolved, see https://github.com/nextcloud-libraries/nextcloud-vue/milestone/132

For a final release the list is a bit longer, see https://github.com/nextcloud-libraries/nextcloud-vue/milestone/118

PR #5009, opened 2023-12-29
fix(NcCheckboxRadioSwitch): use correct padding

### LINKED ISSUE #5008: `NcCheckboxRadioSwitch` with empty slot not round
The `NcCheckboxRadioSwitch` component when used without a default slot content does render a slightly ellipsoid background and is not round as expected. This looks a bit weird, e.g. in the new files app (I changed the background color to make it more obvious here):

grafik

I would propose to slightly reduce the left/right padding to make the background circular.

grafik

raimund-schluessler:
### ☑️ Resolves

* Fix #5008.

This adjusts the left/right padding of the label to the correct / better value of `10px`. The value of `14px` calculated before does not work here, as the element is not an icon with a width of `16px` but a label with a width of `24px`.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot 2024-01-02 at 10-48-22 Nextcloud Vue Style Guide | Screenshot 2024-01-02 at 13-06-26 Nextcloud Vue Style Guide

### DIFF
--- a/src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
+++ b/src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
@@ -30,6 +30,7 @@
 		:class="{
 			['checkbox-content-' + type]: true,
 			'checkbox-content--button-variant': buttonVariant,
+			'checkbox-content--has-text': !!$slots.default,
 		}">
 		<span :class="{
 				'checkbox-content__icon': true,
@@ -212,7 +213,7 @@ export default {
 	user-select: none;
 	min-height: $clickable-area;
 	border-radius: $clickable-area;
-	padding: 4px $icon-margin;
+	padding: 4px calc(($clickable-area - var(--icon-height)) / 2);
 	// Set to 100% to make text overflow work on button style
 	width: 100%;
 	// but restrict to content so plain checkboxes / radio switches do not expand
@@ -240,6 +241,10 @@ export default {
 		.checkbox-content__icon--checked > * {
 			color: var(--color-primary-element-text);
 		}
+	}
+
+	&--has-text {
+		padding-right: $icon-margin;
 	}
 
 	&:not(&--button-variant) {
--- a/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
+++ b/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
@@ -511,6 +511,7 @@ export default {
 		cssVars() {
 			return {
 				'--icon-size': this.size + 'px',
+				'--icon-height': (this.type === TYPE_SWITCH ? 16 : this.size) + 'px',
 			}
 		},
 


ShGKme:
If I understand the issue correctly, as alternative we may render `span.checkbox-content__text` only when there is actually content.

raimund-schluessler:
> If I understand the issue correctly, as alternative we may render `span.checkbox-content__text` only when there is actually content.

Not quite, as it is already hidden when no text is present for exactly this reason. See 
https://github.com/nextcloud-libraries/nextcloud-vue/blob/c9cb21982b5c23b65fc7552073a2456c287f312f/src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue#L225-L232

skjnldsv (dismissed):

src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
```diff
@@ -216,7 +216,7 @@ export default {
 	user-select: none;
 	min-height: $clickable-area;
 	border-radius: $clickable-area;
-	padding: 4px $icon-margin;
+	padding: 4px math.div($clickable-area - 24px, 2);
```
skjnldsv, line 219: Where is this 24 coming from?
Ideally this should be all calculated.

raimund-schluessler: You are right, it needs to be calculated, as it currently does not work for `type="switch"`. This value is the size of the icon set here:
https://github.com/nextcloud-libraries/nextcloud-vue/blob/dae4eca0f022724619c7b912133094d75c886769/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue#L498C11-L502

I guess it should look like this in the end (don't mind the background-colors, just set for visualization):
grafik

raimund-schluessler, line 219:
```suggestion
	padding: 4pxcalc(($clickable-area - var(--icon-size))/2);
```

raimund-schluessler: Fixed.

raimund-schluessler: Updated the screenshot in the initial post.

jancborchardt (changes requested):
Very nice detail fix! :) The checkboxes and radio buttons look great, but the toggle switches were better before as they fit exactly in the container radius-wise. :) That is, the container edge was always equidistant to the toggle/icon.

raimund-schluessler:
> Very nice detail fix! :) The checkboxes and radio buttons look great, but the toggle switches were better before as they fit exactly in the container radius-wise. :) That is, the container edge was always equidistant to the toggle/icon.

I see, I will fix this. We should also consistently set the same padding-right for all elments if there is text. I noticed that it looks weird to have a different padding on the right for switch and checkbox labels.

jancborchardt:
>  I noticed that it looks weird to have a different padding on the right for switch and checkbox labels.

@raimund-schluessler true – I think the padding-right for the versions with text looked better in the previous version and was consistent among switches and checkboxes, could that be?

raimund-schluessler:
> > I noticed that it looks weird to have a different padding on the right for switch and checkbox labels.
> 
> @raimund-schluessler true – I think the padding-right for the versions with text looked better in the previous version and was consistent among switches and checkboxes, could that be?

Yes, that is correct. Before we just set the same padding for switches and checkboxes on the left and right no matter what. I adjusted it now to have the same padding right as before if there is text and an adjusted padding left (and right without text) for checkboxes and radio switches to have the element circular without text. Switches did not change at all. The screenshot is updated to show the changes.

jancborchardt (approved):
Looks very nice! :)

skjnldsv (approved):

skjnldsv (commented):
Seems also relevant: https://github.com/nextcloud-libraries/nextcloud-vue/pull/5001

PR #5011, opened 2023-12-29
[next] fix(NcBreadcrumb): correctly emit drag events

raimund-schluessler:
### ☑️ Resolves

* Fixes opening the hidden breadcrumbs menu when dragging and hovering over the menu. The `dragenter` and `dragleave` events where not emitted, so the listener in `NcBreadcrumbs` was never called, hence, the hidden crumb menu not opened. Emitting the events fixes the issue.

### DIFF
--- a/src/components/NcBreadcrumbs/NcBreadcrumbs.vue
+++ b/src/components/NcBreadcrumbs/NcBreadcrumbs.vue
@@ -547,7 +547,6 @@ export default {
 				ref: 'actionsBreadcrumb',
 				key: 'actions-breadcrumb-1',
 				// Add handlers so the Actions menu opens on hover
-				onDragstart: this.dragStart,
 				onDragenter: () => { this.menuBreadcrumbProps.open = true },
 				onDragleave: this.closeActions,
 				// Make sure we keep the same open state
--- a/src/components/NcBreadcrumb/NcBreadcrumb.vue
+++ b/src/components/NcBreadcrumb/NcBreadcrumb.vue
@@ -151,8 +151,10 @@ export default {
 		},
 	},
 	emits: [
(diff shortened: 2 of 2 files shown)


skjnldsv (approved):

raimund-schluessler:
@susnux @ShGKme Could we get this merged?

src/components/NcBreadcrumbs/NcBreadcrumbs.vue
```diff
@@ -547,7 +547,6 @@ export default {
 				ref: 'actionsBreadcrumb',
 				key: 'actions-breadcrumb-1',
 				// Add handlers so the Actions menu opens on hover
-				onDragstart: this.dragStart,
```
ShGKme, line 550: Why should this be removed? Doesn't it act the same as in Vue 2 here?

raimund-schluessler: > Why should this be removed? Doesn't it act the same as in Vue 2 here?

It is redundant, and could be removed for `master` as well.

`dragStart` in `NcBreadcrumbs`:
https://github.com/nextcloud-libraries/nextcloud-vue/blob/d8d2621ca341e3d7e02bab8ec77c36ea3bbd854f/src/components/NcBreadcrumbs/NcBreadcrumbs.vue#L363-L365

and we set it anyway to
https://github.com/nextcloud-libraries/nextcloud-vue/blob/d8d2621ca341e3d7e02bab8ec77c36ea3bbd854f/src/components/NcBreadcrumb/NcBreadcrumb.vue#L37
in `NcBreadcrumb`.

ShGKme (approved):

PR #5012, opened 2023-12-30
[next] fix(test): correctly provide props in test

raimund-schluessler:
### ☑️ Resolves

* Fixes the way we provide props to the `mount` function in `modal.spec.js` for vue 3. 

@susnux Since the tests passed even though we didn't provide the props correctly, I assume the tests don't really test anything. I didn't change it here though, since it is the same in `master` and should probably be fixed there, I think.

### DIFF
--- a/tests/unit/components/NcModal/modal.spec.js
+++ b/tests/unit/components/NcModal/modal.spec.js
@@ -26,7 +26,7 @@ import NcModal from '../../../../src/components/NcModal/NcModal.vue'
 
 describe('NcModal', () => {
 	it('closes on click outside with `canClose`', async () => {
-		const wrapper = mount(NcModal, { propsData: { canClose: true, title: 'modal' } })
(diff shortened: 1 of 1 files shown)


susnux:
> Since the tests passed even though we didn't provide the props correctly

Yes but the test utils have a fallback so it still set the props correctly:
https://github.com/vuejs/test-utils/blob/b1c2388c3422115f0a7d2627fe308f44f4145462/src/createInstance.ts#L173

susnux (approved):

raimund-schluessler:
> Yes but the test utils have a fallback so it still set the props correctly: https://github.com/vuejs/test-utils/blob/b1c2388c3422115f0a7d2627fe308f44f4145462/src/createInstance.ts#L173

Ah, thanks for the hint. I wasn't aware of this. They apparently think ahead :laughing:

ShGKme (approved):

PR #5017, opened 2024-01-01
feat(NcActionInput): allow to append `NcSelect` to body

raimund-schluessler:
### ☑️ Resolves

* This allows to append the `NcSelect` in `NcActionInput` to body. This prevents clipping the select by the action menu.

Code-wise it would be simpler to just remove `:append-to-body="false"`, but this would introduce a breaking change, as it would set `append-to-body` to `true` by default for `NcSelect`. So I decided to add a prop instead.
Also, for vue 3 this is not necessary, since `v-bind="$attrs"` overwrites the props already set on the component.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
 Bildschirmfoto vom 2024-01-01 12-39-14 | Bildschirmfoto vom 2024-01-01 12-40-09

### DIFF
--- a/src/components/NcActionInput/NcActionInput.vue
+++ b/src/components/NcActionInput/NcActionInput.vue
@@ -100,6 +100,7 @@ For the `NcSelect` component, all events will be passed through. Please see the 
 				type="multiselect"
 				label="label"
 				track-by="id"
+				:append-to-body="true"
 				:multiple="true"
 				:options="[{label:'Apple', id: 'apple'}, {label:'Banana', id: 'banana'}, {label:'Cherry', id: 'cherry'}]">
 				<template #icon>
@@ -193,7 +194,7 @@ For the `NcSelect` component, all events will be passed through. Please see the 
 							:value="value"
 							:placeholder="text"
 							:disabled="disabled"
-							:append-to-body="false"
+							:append-to-body="$attrs.appendToBody || $attrs['append-to-body'] || false"
 							:input-class="{ focusable: isFocusable }"
 							class="action-input__multi"
 							v-bind="$attrs"


src/components/NcActionInput/NcActionInput.vue
```diff
@@ -385,6 +386,13 @@ export default {
 			type: Boolean,
 			default: true,
 		},
+		/**
+		 * Whether to append the `NcDateTimePicker` and `NcSelect` to body.
```
Pytal, line 390:
```suggestion
		 * Whether to append the `NcSelect` to body.
```

Pytal (approved):

src/components/NcActionInput/NcActionInput.vue
raimund-schluessler: Now that you write that, I realized that this PR does change the behavior of the component, as it is now not possible anymore to append the datepicker to body (since I defined a prop now, it is not in `attrs` anymore which we bind).

However, appending the datepicker to body is broken anyway:
image

Should I anyway keep the possibility to append it to body and forward `appendToBody` to the datepicker as well?

ShGKme:
> Also, for vue 3 this is not necessary, since `v-bind="$attrs"` overwrites the props already set on the component.

Then an alternative solution could be to emulate Vue 3 behavior:
```
v-bind="{ appendToBody: false, ...$attrs }"
```

It would allow avoiding adding a specific prop for `NcSelect`

src/components/NcActionInput/NcActionInput.vue
Pytal: 🤔 I would only forward it to NcSelect for now and then to datepicker after appendToBody is fixed for it

nimishavijay (approved):
Great fix :) As @marcoambrosini has mentioned we should generally move away from nesting components in an ActionMenu though :)

raimund-schluessler:
> v-bind="{ appendToBody: false, ...$attrs }"
> ```
> 
> It would allow avoiding adding a specific prop for `NcSelect`

I think this is the best solution, as it saves us from adding a prop. I changed it accordingly. It just has a tiny flaw, that one has to correctly name the attribute, as only `append-to-body` but not `appendToBody` will work.

src/components/NcActionInput/NcActionInput.vue
raimund-schluessler: With the solution from @ShGKme it will keep (not-)working as is for the datepicker.

nimishavijay (approved):

raimund-schluessler:
@nimishavijay You seem to have no write access to the repository, so I cannot merge yet 🙂 Is this on purpose?

src/components/NcActionInput/NcActionInput.vue
```diff
@@ -193,10 +194,9 @@ For the `NcSelect` component, all events will be passed through. Please see the
 							:value="value"
 							:placeholder="text"
 							:disabled="disabled"
-							:append-to-body="false"
 							:input-class="{ focusable: isFocusable }"
 							class="action-input__multi"
-							v-bind="$attrs"
+							v-bind="{ 'append-to-body': false, ...$attrs }"
```
ShGKme, line 199: `kebab-case` is for templates (I'm not sure it works in Vue 3)

```suggestion
							v-bind="{ appendToBody: false, ...$attrs }"
```

ShGKme (changes requested):

src/components/NcActionInput/NcActionInput.vue
raimund-schluessler: But then we also need to provide the prop as `:appendToBody`, otherwise the object spread/merge does not work.

ShGKme: Oh, indeed, Vue doesn't change the case in `$attrs`, only in `$props`...

But actually it works in any way.

`{ 'append-to-body': 1, appendToBody: 2 }` and `{ appendToBody: 1, 'append-to-body': 2 }` both set value 2, using the last value for the final prop value.

So both options works the same no matter if kebab or camel case was used in template.

ShGKme: So I'd call it a style preference just to keep camelCase in JS part

ShGKme (approved):

src/components/NcActionInput/NcActionInput.vue
raimund-schluessler: I think I don't get it. I tried it with camelCase for v-bind and it didn't work. Did I do something wrong?

ShGKme: Arr x2.

Vue 3 and Vue 2 works differently here... 

Vue 3 works as I described and the last one value wins.

In Vue 2 the first one wins.

ShGKme: So:
- `{ 'append-to-body': false, ...$attrs }` breaks with `<NcActionInput appendToBody>`
- `{ appendToBody: false, ...$attrs }` breaks with `<NcActionInput append-to-body>`

ShGKme: 🥲 ?

```js
:append-to-body="$attrs.appendToBody || $attrs['append-to-body'] || false"
```

raimund-schluessler: > ```
> :append-to-body="$attrs.appendToBody || $attrs['append-to-body'] || false"
> ```

Yeah, why not. Works for both `append-to-body` and `appendToBody` in the template. For vue 3 we anyway have to / can remove it, since it works as it there.

ShGKme (approved):

PR #5018, opened 2024-01-02
refactor(NcCheckboxRadioSwitch): move comment to have a single root node

ShGKme:
Having a comment directly in a template may cause issues in `@vue/test-utils` and other tools that consider such a component to have two root nodes in a fragment: a comment node and a `span` node.

Also, replaced `null` with `undefined` to remove minor typing issue.

No visual changes.

### 🏁 Checklist

### DIFF
--- a/src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
+++ b/src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
@@ -21,16 +21,16 @@
 -->
 
 <template>
-	<!--
-		label can't be used here because of shift+click firefox bug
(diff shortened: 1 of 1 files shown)


raimund-schluessler (approved):

skjnldsv:
> Having a comment directly in a template may cause issues in `@vue/test-utils` and other tools that consider such a component to have two root nodes in a fragment: a comment node and a `span` node.

Does it??? Omg, I really thought it would be smart enough :see_no_evil: 
Thanks for catching this :sweat_smile:

skjnldsv (approved):

PR #5045, opened 2024-01-09
fix(NcDateTimePicker): Use ISO week numbers if `showWeekNumber` is set

### LINKED ISSUE #5044: `NcDateTimePicker` week labels off by one
Originally reported in https://github.com/nextcloud/tasks/issues/2366.

See here:
Screenshot 2024-01-09 at 13-17-51 Aufgaben - Nextcloud

The first week of January should be the week with the starting year's first Thursday in it (the formal
ISO definition), see https://en.wikipedia.org/wiki/ISO_8601

In the screenshot above, the first week starts 7 days to early. What is labeled as `1` should be `53` of 2022.

susnux:
### ☑️ Resolves

* Fix #5044 

Use ISO week numbers by default, but still allow users to override it using the `formatter`  prop.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot_20240109_135501|Screenshot_20240109_135514

### 🏁 Checklist

### DIFF
--- a/src/components/NcDateTimePicker/NcDateTimePicker.vue
+++ b/src/components/NcDateTimePicker/NcDateTimePicker.vue
@@ -127,7 +127,7 @@ export default {
 		:append-to-body="appendToBody"
 		:clearable="clearable"
 		:format="format ? format : formatTypeMap"
-		:formatter="formatter"
+		:formatter="internalFormatter"
 		:lang="lang ? lang : defaultLang"
 		:minute-step="minuteStep"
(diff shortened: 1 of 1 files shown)


raimund-schluessler (approved):
Works well :+1:
There are just two trailing commas missing.

susnux:
>  There are just two trailing commas missing.

fixed

ShGKme (approved):

PR #5050, opened 2024-01-09
[next] chore(deps): bump `vue` to `3.4.13`

raimund-schluessler:
### ☑️ Resolves

* Bumps vue to 3.4.13.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -92,7 +92,7 @@
         "npm": "^9.0.0"
       },
       "peerDependencies": {
-        "vue": "^3.3.7"
+        "vue": "^3.4.13"
       }
     },
     "node_modules/@aashutoshrathi/word-wrap": {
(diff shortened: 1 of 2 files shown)


ShGKme:
> Unfortunately, the `NcAppSidebarTabs` component breaks and does not render the tab list initially.
> 
> For some reason, the value of the `hasMultipleTabs` computed of `NcAppSidebarTabs` is not updated in the template, although correctly calculated.

Hmm.. Interesting, I'll have a look later today.

Since Vue 3.4 was released 12 days ago, it already have had 7 patches with many bug fixes and 2 reverts 🙈

raimund-schluessler:
> Since Vue 3.4 was released 12 days ago, it already have had 7 patches with many bug fixes and 2 reverts 🙈

Yes, I saw that as well. Maybe we just need to wait for two weeks and the problem will be fixed by vue 3.4.14 😄🙈

raimund-schluessler:
3.4.8 didn't do the trick yet.

raimund-schluessler:
This really is a weird issue.

- Showing the tabs dynamically via user input / v-if works when going from 2 to 3 tabs. Hiding tabs always works, e.g. going from 3 to 2 to 1 to none renders the component correctly.
- Showing tabs does not work when going from 1 to 2 tabs, only adding a third tab makes all three appear.
- Putting `tabs.length` somewhere in the template, or using a computed that uses `tabs.length` as a number makes the component work. Converting `tabs.length` to a Boolean in a computed does not work, using it as a Boolean in the template still works.

Something is a bit off with the reactivity system :thinking:

raimund-schluessler:
3.4.10 didn't fix it either. I created a minimal reproduction example and reported it in the vue repository. See https://github.com/vuejs/core/issues/10082.

ShGKme:
A workaround - replace `created` with `mounted`, so it updates in a new tick. A bit less performant, but should be fine as we don't have many tabs usually.

raimund-schluessler:
> A workaround - replace `created` with `mounted`, so it updates in a new tick. A bit less performant, but should be fine as we don't have many tabs usually.

Interesting. I don't have a strong opinion here. I am fine with merging with the workaround or wait for the vue devs to handle the issue.
I would tend to use the workaround, since we don't know when the issue will be fixed in vue, though. We can go back to `created` once the issue is fixed.

ShGKme:
Vue 3.4.11 has a patch to fix the issue 🥳

raimund-schluessler:
Pushed vue 3.4.11 and reverted the change to NcAppSidebarTabs.

ShGKme (approved):
🥳

raimund-schluessler:
> Yes, I saw that as well. Maybe we just need to wait for two weeks and the problem will be fixed by vue 3.4.14 😄🙈

We are at `3.4.13` already which contains another fix for the issue. If we don't merge soon, we will reach `3.4.14` before merging as predicted :smile: :see_no_evil:

susnux (approved):

PR #5058, opened 2024-01-11
[next] fix(NcCheckboxContent): correctly check default slot

raimund-schluessler:
### ☑️ Resolves

* Fixes the detection whether `NcCheckboxContent` has text. Before, it would always be `true` leading to a wrong padding right.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
grafik | grafik

### DIFF
--- a/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
+++ b/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
@@ -301,8 +301,10 @@ export default {
 				<slot name="icon" />
 			</template>
 
-			<!-- @slot The checkbox/radio label -->
-			<slot />
+			<template v-if="!!$slots.default" #default>
+				<!-- @slot The checkbox/radio label -->
+				<slot />
+			</template>
 		</NcCheckboxContent>
 	</component>
 </template>


src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
```diff
@@ -166,6 +167,10 @@ export default {
 		},
 	},
 
+	methods: {
+		isSlotPopulated,
+	},
```
ShGKme, line 172: > eslint
>
> The "computed" property should be above the "methods" property on line 170

ShGKme (commented):
Isn't it fixable on `NcCheckboxRadioSwitch`'s side same as in some other component? To avoid tripple rendering.

```html
<template v-if="!!$slots.default" #default>
  <slot />
</template>
```

raimund-schluessler:
> Isn't it fixable on `NcCheckboxRadioSwitch`'s side same as in some other component? To avoid tripple rendering.
> 
> ```
> <template v-if="!!$slots.default" #default>
>   <slot />
> </template>
> ```

Yes, of course. Thanks for remembering that! Fixed it.

src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
```diff
@@ -302,7 +302,9 @@ export default {
 			</template>
 
 			<!-- @slot The checkbox/radio label -->
-			<slot />
+			<template v-if="!!$slots.default" #default>
+				<slot />
+			</template>
```
ShGKme, line 307: Not sure, but I'd expect `<!-- @slot` works only directly before `<slot>`

ShGKme: Cannot check on Netlify 🙈

raimund-schluessler: Yeah, probably. Adjusted.

ShGKme (approved):

skjnldsv (approved):

PR #5059, opened 2024-01-12
feat: Add NcUserStatusIcon

Pytal:
- For https://github.com/nextcloud/server/issues/42644

### Summary

Add new NcUserStatusIcon component for accessibility

### 🖼️ Screenshots

> No visual changes

### 🏁 Checklist

### DIFF
--- a/src/components/NcUserStatusIcon/NcUserStatusIcon.vue
+++ b/src/components/NcUserStatusIcon/NcUserStatusIcon.vue
@@ -1,0 +1,191 @@
+<!--
+  - @copyright 2024 Christopher Ng <chrng8@gmail.com>
+  -
+  - @author Christopher Ng <chrng8@gmail.com>
+  -
+  - @license AGPL-3.0-or-later
+  -
+  - This program is free software: you can redistribute it and/or modify
+  - it under the terms of the GNU Affero General Public License as
+  - published by the Free Software Foundation, either version 3 of the
+  - License, or (at your option) any later version.
+  -
+  - This program is distributed in the hope that it will be useful,
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+-->
+
+<docs>
+### Description
+
+This component displays a user status icon.
+
+### Example
+
+```vue
+<template>
+	<div class="row">
+		<NcUserStatusIcon status="online" />
+		<NcUserStatusIcon status="away" />
+		<NcUserStatusIcon status="dnd" />
+		<NcUserStatusIcon status="invisible" />
+	</div>
+</template>
+
+<style>
+.row {
+	display: flex;
+	gap: 10px;
+}
+</style>
+```
+</docs>
+
+<template>
+	<span v-if="activeStatus"
+		class="user-status-icon"
+		:class="{
+			'user-status-icon--invisible': ['invisible', 'offline'].includes(status),
+		}"
+		role="img"
+		:aria-hidden="ariaHidden"
+		:aria-label="ariaLabel"
+		v-html="activeSvg" /> <!-- eslint-disable-line vue/no-v-html -->
+</template>
+
+<script>
+import Vue from 'vue'
+import axios from '@nextcloud/axios'
+import { generateOcsUrl } from '@nextcloud/router'
+import { getCapabilities } from '@nextcloud/capabilities'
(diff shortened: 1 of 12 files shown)


src/components/NcUserStatusIcon/NcUserStatusIcon.vue
```diff
@@ -0,0 +1,173 @@
...
+			].includes(status),
+		},
+
+		/**
+		 * Set the `aria-hidden` attribute
+		 */
+		ariaHidden: {
+			type: String,
+			default: null,
+		},
```
ShGKme, line 113: I don't understand in which case this icon should and should not be `aria-hidden`...
Could you clarify it?

ShGKme (changes requested):
Great, definitely useful component!

ShGKme:
What about also support displaying emoji status icon in this component?

src/components/NcUserStatusIcon/NcUserStatusIcon.vue
ShGKme, line 113:
```suggestion
		/**
		 * Set the `aria-hidden` attribute
		 * @type {'true'|'false'}
		 */
		ariaHidden: {
			type: String,
			default: null,
			validator: (value) => ['true', 'false'].includes(value),
		},
```

src/components/NcUserStatusIcon/NcUserStatusIcon.vue
```diff
@@ -0,0 +1,173 @@
...
+		 * Set the user id to fetch the status
+		 */
+		user: {
+			type: String,
+			default: null,
+		},
+
+		/**
+		 * Set the status
+		 */
```
ShGKme, line 93:
```suggestion
		/**
		 * Set the status
		 * @type {'online'|'away'|'busy'|'dnd'|'invisible'|'offline'}
		 */
```

src/components/NcUserStatusIcon/NcUserStatusIcon.vue
```diff
@@ -0,0 +1,173 @@
...
+.row {
+	display: flex;
+	gap: 10px;
+}
+</style>
+```
+</docs>
+
+<template>
+	<span v-if="activeStatus" v-bind="attributes">
```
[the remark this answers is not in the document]
Pytal: Agreed!

src/components/NcUserStatusIcon/NcUserStatusIcon.vue
```diff
@@ -0,0 +1,173 @@
...
+			<template v-else-if="activeStatus === 'dnd'">
+				<path fill="none" d="M-4-4h24v24H-4V-4z" />
+				<path fill="var(--color-error)" d="M8 0C3.6 0 0 3.6 0 8s3.6 8 8 8 8-3.6 8-8-3.6-8-8-8z" />
+				<path fill="#fdffff" d="M5 6.5h6c.8 0 1.5.7 1.5 1.5s-.7 1.5-1.5 1.5H5c-.8 0-1.5-.7-1.5-1.5S4.2 6.5 5 6.5z" />
+			</template>
+			<template v-else-if="['invisible', 'offline'].includes(activeStatus)">
+				<path fill="none" d="M-4-4h24v24H-4V-4z" />
+				<path d="M8 0C3.6 0 0 3.6 0 8s3.6 8 8 8 8-3.6 8-8-3.6-8-8-8zm0 3.2c2.7 0 4.8 2.1 4.8 4.8s-2.1 4.8-4.8 4.8S3.2 10.7 3.2 8 5.3 3.2 8 3.2z" />
+			</template>
+		</svg>
```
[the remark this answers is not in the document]
Pytal: Inlining the markup would remove overhead but I agree that having them in separate files would be better in general ;)

All the svgs are just black when viewing in IDE though as the CSS variables aren't loaded 🤔

src/components/NcUserStatusIcon/NcUserStatusIcon.vue
Pytal: See this commit 1ccc246c which migrates NcIconSvgWrapper to NcUserStatusIcon. Since this new component does not need an explicit `name`/aria-label, as it derives it from the `activeStatus` automatically, we need to allow aria-hiding this explicitly which is needed for https://github.com/nextcloud-libraries/nextcloud-vue/blob/91c44a3b8fba7d18dcf0b6ba48862bac0fe55328/src/components/NcAvatar/NcAvatar.vue#L397-L398

Pytal:
> What about also support displaying emoji status icon in this component?

Not sure there is a large benefit as emojis are just strings, let me know if there is a strong reason to though :)

ShGKme (approved):

raimund-schluessler (approved):

raimund-schluessler:
The code looks fine. But the new server colors for success and warning look depressing.

PR #4550, opened 2023-09-20
feat(NcDialog): New component (moved from `@nextcloud/dialogs`)

susnux:
### ☑️ Resolves
* Fix #631 
* Fix #4348 (fixed by migrating AppSettingsDialog to NcDialog base)
* Replace https://github.com/nextcloud-libraries/nextcloud-vue/pull/4415 (without Typescript and fancy `<script setup>`).

Introduces NcDialog as a generic base Dialog component, while NcModal is simply a base for modals like the viewer, NcDialog provides a convenient interface to define dialogs (see example).
The new FilePicker of the `@nextcloud/dialogs` package is based on top of it.

### 🖼️ Screenshots
One example dialog:

image

### 🏁 Checklist

- [ ] ⛑️ Tests are included or are not applicable
- [x] 📘 Component documentation has been extended, updated or is not applicable

PR #5062, opened 2024-01-12
fix(NcDialog): allow to close NcDialog on click outside

Antreesy:
### ☑️ Resolves

* Follow-up from #4550
* It feels counter-intuitive now for some dialogs (ex-modals) to not be closed with outside click. That doesn't apply to small confirmation popups, but some general things should be reverted to old behavior with this:
  * NcAppSettingsDialog
  * FilePicker
  * e.t.c.

### 🚧 Tasks

- [ ] Alternative: make `closeOnClickOutside` true by default ?

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
+++ b/src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
@@ -268,6 +268,7 @@ export default {
 		dialogProperties() {
 			return {
 				additionalTrapElements: this.additionalTrapElements,
+				closeOnClickOutside: true,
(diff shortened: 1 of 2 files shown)


ShGKme (approved):

szaimen (approved):
Makes sense

susnux:
>  Alternative: make closeOnClickOutside true by default

I would say no for the filepicker we had this discussion somewhere. The close button is always visible - even if scrolled - on NcDialogs so users can explicitly close it.
This prevents erroneous aborting just because of clicking a bit too far outside.

PR #5064, opened 2024-01-13
enh(NcHeaderMenu): Close navigation header menu on focusout

Pytal:
- For https://github.com/nextcloud/server/issues/41875

### Summary

- Close the menu automatically on focusout
- Alternative way using keydown is not sufficient as users may switch focus to an external element by targetting it directly via e.g. mouse or screen reader

### 🏁 Checklist


ShGKme (approved):
Makes sense. Probably we should even do the same with all popovers without focus trap.

ShGKme:
Cool solution!

susnux (approved):

PR #5068, opened 2024-01-14
[next] chore(NcBreadcrumbs): simplify code

raimund-schluessler:
### ☑️ Resolves

This PR simplifies the code of `NcBreadcrumbs`.

- 825607a4490fe245b08dd6d8d6cd413c38f76224 changes the render function to use a single `Array.splice` instead of `slice -> push -> slice -> concat`. This should speed up the render function a bit. 
- 1ddda89109f2c481b7c30a9ff8fbd9c568297ba0 changes the `breadcrumbsRefs` object to be an array, which safes us from running `Object.values` at two spots.

I didn't send this PR against `master` to not introduce any issues in the stable branch. Also, merging changes to vue2 render functions into `next` is always a pain. In case you insist, we could of course also improve it for `master`.

No visual or functional changes.

Best viewed as https://github.com/nextcloud-libraries/nextcloud-vue/pull/5068/files?diff=unified&w=1


susnux (approved):
looks good to me

raimund-schluessler:
Resolved conflicts.

Antreesy (approved):
Seems legit, code changes should be equivalent
Haven't work with array refs in Vue 3, but know they should work :seal:

raimund-schluessler:
These random Cypress failures are annoying. Six runs now and always a different test failing.

susnux:
> These random Cypress failures are annoying. Six runs now and always a different test failing.

Yes that is super annoying, Cypress knows of that flaky tests for couple of month but is not considered to be something worth to fix.
@ShGKme mentioned PlayWright to me some time ago, since then I only use that for my personal projects, much less annoying :sweat_smile:

PR #5075, opened 2024-01-16
Remove unneeded `tab`, `tablist` and `aria-selected` roles from navigation

JuliaKirschenheuter:
### ☑️ Resolves

* For https://github.com/nextcloud/server/issues/42599
* See https://github.com/nextcloud/server/issues/42599#issuecomment-1879050422
* Navigation example: https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/navigation.html

### 🖼️ Screenshots

🏚️ Before | 🏡 After

No visual changes. Result: no errors through automatically a11y tests.

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
+++ b/src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
@@ -145,19 +145,17 @@ export default {
 		@update:open="handleCloseModal">
 		<template v-if="hasNavigation" #navigation="{ isCollapsed }">
 			<ul v-if="!isCollapsed"
-				:aria-label="settingsNavigationAriaLabel"
-				class="navigation-list"
-				role="tablist">
+				class="navigation-list">
 				<li v-for="section in sections" :key="section.id">
-					<a :aria-selected="section.id === selectedSection"
+					<a :aria-current="`${section.id === selectedSection}`"
 						:class="{
 							'navigation-list__link': true,
 							'navigation-list__link--active': section.id === selectedSection,
 							'navigation-list__link--icon': hasNavigationIcons,
 						}"
-						role="tab"
+						:href="`#settings-section_${section.id}`"
 						tabindex="0"
-						@click="handleSettingsNavigationClick(section.id)"
+						@click.prevent="handleSettingsNavigationClick(section.id)"
 						@keydown.enter="handleSettingsNavigationClick(section.id)">
 						<div v-if="hasNavigationIcons" class="navigation-list__link-icon">
 							<NcVNodes v-if="section.icon" :vnodes="section.icon" />
--- a/tests/unit/components/NcAppSettingsDialog/NcAppSettingsDialog.spec.ts
+++ b/tests/unit/components/NcAppSettingsDialog/NcAppSettingsDialog.spec.ts
@@ -55,8 +55,8 @@ describe('NcAppSettingsDialog: Sections registration', () => {
 
 		await nextTick()
 
-		expect(wrapper.findAll('[role="tab"]')).toHaveLength(1)
-		expect(wrapper.find('[role="tab"]').text()).toBe('test_name')
+		expect(wrapper.findAll('nav a')).toHaveLength(1)
+		expect(wrapper.find('nav a').text()).toBe('test_name')
 	})
 
 	it('can register a new section', async () => {
@@ -70,8 +70,8 @@ describe('NcAppSettingsDialog: Sections registration', () => {
 
 		wrapper.vm.registerSection('test_id', 'test_name')
 		await nextTick()
-		expect(wrapper.findAll('[role="tab"]')).toHaveLength(1)
-		expect(wrapper.find('[role="tab"]').text()).toBe('test_name')
+		expect(wrapper.findAll('nav a')).toHaveLength(1)
+		expect(wrapper.find('nav a').text()).toBe('test_name')
 	})
 
 	it('warn on register a already registered section name', async () => {
@@ -89,13 +89,13 @@ describe('NcAppSettingsDialog: Sections registration', () => {
 		// First call should be OK
 		wrapper.vm.registerSection('test_id', 'test_name')
 		await nextTick()
-		expect(wrapper.findAll('[role="tab"]')).toHaveLength(1)
+		expect(wrapper.findAll('nav a')).toHaveLength(1)
 		expect(spy).not.toHaveBeenCalled()
 
 		// Second one should unregister first and replace with this one, but show an error
 		wrapper.vm.registerSection('test_id_2', 'test_name')
 		await nextTick()
-		expect(wrapper.findAll('[role="tab"]')).toHaveLength(2)
+		expect(wrapper.findAll('nav a')).toHaveLength(2)
 		expect(spy).toHaveBeenCalled()
 	})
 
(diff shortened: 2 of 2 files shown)


src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
```diff
@@ -146,16 +146,13 @@ export default {
 		<template v-if="hasNavigation" #navigation="{ isCollapsed }">
 			<ul v-if="!isCollapsed"
 				:aria-label="settingsNavigationAriaLabel"
-				class="navigation-list"
-				role="tablist">
+				class="navigation-list">
 				<li v-for="section in sections" :key="section.id">
-					<a :aria-selected="section.id === selectedSection"
-						:class="{
+					<a :class="{
```
susnux, line 151:
```suggestion
					<a :aria-current="section.id === selectedSection"
						:class="{
```

Should we not keep the information of the current active one?

tests/unit/components/NcAppSettingsDialog/NcAppSettingsDialog.spec.ts
```diff
@@ -55,8 +55,8 @@ describe('NcAppSettingsDialog: Sections registration', () => {
 
 		await nextTick()
 
-		expect(wrapper.findAll('[role="tab"]')).toHaveLength(1)
-		expect(wrapper.find('[role="tab"]').text()).toBe('test_name')
+		expect(wrapper.findAll('.navigation-list__link')).toHaveLength(1)
```
susnux, line 58: @ShGKme mentioned to me we should rather use semantics than internal class names for testing, so in this case I guess like this?

```suggestion
		expect(wrapper.findAll('nav a')).toHaveLength(1)
```

ShGKme: Agree

src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
ShGKme: This attribute was not correct anyway 🙈
It should have been a string, not boolean in old Vue.

ShGKme: But I wonder if this should be navigation and links at all. It is not related to the actual navigation, there is no `href`.

But it has kind of navigation semantic.

I dunno

JuliaKirschenheuter: The aria-selected attribute indicates the current "selected" state for gridcell, option, row and tab roles.
> https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-selected

We have no tab roles anymore

JuliaKirschenheuter: > links at all. It is not related to the actual navigation, there is no href.

oohhh, good point. i have to replace it too

susnux: >  there is no href

I would say keep links, but add the href as `#${section.id}`.

> The aria-selected attribute indicates the current "selected" state for gridcell, option, row and tab roles.

Yes thats why I suggested `aria-current` :)

ShGKme: or `? 'true' : 'false'`

susnux: > With any aria-* it should be `(section.id === selectedSection).toString()`

Or short form `${section.id === selectedSection}`?

Pytal: > > there is no href
> 
> I would say keep links, but add the href as `#${section.id}`.

If we add anchor links then it wouldn't it be less accessible as the anchor at the end of the URL makes it seem like opening the same URL auto-opens the settings dialog to the same nav section when it doesn't

There is currently no special URL param to directly open the setting dialog either 🤔

ShGKme: > Or short form? (Not sure about readability)

Too many options, let's just migrate to Vue 3 finally =D

JuliaKirschenheuter: @susnux it should not be a link as @Pytal already said. Could you please explain your point why do we need `aria-current` for navigation items?

ShGKme: I think that it is best to make it a link. I agree with @Pytal thoughts, but it doesn't seem to me critical.

This list:
- is actually navigation (on the page). The only action click does is scrolling to the element with a specific ID
- has an item with a "current" state with also changes during scrolling

So semantically, it is very close to navigation with links from my PoV, rather than a button.

The only problem is that the dialog is not reopened on page reload with opened dialog... Do you think this is a serious issue? 

---

The best would be, it would also fix current keyboard navigation:
1. A user presses on button 3 "WebDAV" in navigation,
2. It scrolls to the section 3
3. A user presses tab or uses a screen reader
   - **Expected**: the focus is moved to the next element in WebDAV section
   - **Actual:** the focus is moved to the next navigation button

And this behavior is also very "link" behaviour.
---

But I'm also fine with buttons if we move the focus.

emoral435 (approved):

susnux (approved):

src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
```diff
@@ -145,17 +145,15 @@ export default {
 		@update:open="handleCloseModal">
 		<template v-if="hasNavigation" #navigation="{ isCollapsed }">
 			<ul v-if="!isCollapsed"
-				:aria-label="settingsNavigationAriaLabel"
-				class="navigation-list"
-				role="tablist">
+				class="navigation-list">
 				<li v-for="section in sections" :key="section.id">
-					<a :aria-selected="section.id === selectedSection"
+					<a :aria-current="section.id === selectedSection"
```
ShGKme, line 150: Choose your weapon

// no need in "false" actually
```suggestion
					<a :aria-current="section.id === selectedSection ? 'true' : undefined"
```

```suggestion
					<a :aria-current="section.id === selectedSection ? 'true' : 'false'"
```

```suggestion
					<a :aria-current="section.id === selectedSection ? 'location' : undefined"
```

```suggestion
					<a :aria-current="(section.id === selectedSection).toString()"
```

```suggestion
					<a :aria-current="`${section.id === selectedSection}`"
```

ShGKme:
- Button + manual focus - doesn't move focus (both general and screen reader)
- Link + prevent + manual focus - doesn't move focus (both general and screen reader)
- Link with native behavior - focus is moved, but with a very weird behavior. 
  It is moved though other elements, e.g. skip links of table header. And I cannot understand why...

For example, on this video, after click the screen reader focus is successfully moved on the header. However, it made many steps and went through other elements, like 5th user's action menu button (yellow square).

https://github.com/nextcloud-libraries/nextcloud-vue/assets/25978914/3a222d6a-c045-4661-8683-d6b8251cd7fb

I tried to disable the focus trap, remove debounce unfocus, play with the link and nothing helped.

If you don't have any ideas here, I'd propose to disable default link behavior for now with `.prevent` and fix it later.

JuliaKirschenheuter:
> disable default link behavior for now with `.prevent`

Following behavior will take place: link will be semantic link but will work as a button. That means focus will stay on links and strange focus behavior will be prohibited.

Let's do that!

susnux:
> * It is moved though other elements

Maybe because it is handled like navigation -> reset to tabstop 0 and then move to ID?

ShGKme:
> Maybe because it is handled like navigation -> reset to tabstop 0 and then move to ID?

I tried similar navigation on MDN and it worked fine...

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-haspopup

image

ShGKme (approved):

PR #5086, opened 2024-01-17
feat(NcPopover): provide a11y attributes to the trigger

ShGKme:
## ☑️ Resolves

* Fix https://github.com/nextcloud/server/issues/41904
* In the issue there is only 1 place, but we have this issue in literally all `NcPopover` usages except `NcActions`...

### A problem

1. `NcPopover` trigger button should have a11y attributes
   - `aria-haspopup="<role>"`
   - `aria-expanded="<is shown>"`
3. Trigger button is provided via slot in place, we don't control it
   ```html
   <NcPopover>
     <template #trigger>
       Any button here should have attrs
     </template>
   </NcPopover>
   ```
3. We don't want to control all the places where `NcPopover` is used to make sure its trigger button has attributes.

This implementation might look a bit complex, but it helps to make a lot of popovers more accessible.

### 1. Provide `NcPopover:trigger:attrs` from `NcPopover` to its `#trigger` slot's content

So any component used as a trigger may receive the attrs, required to be on the trigger

A small component `NcPopoverTriggerProvider` is created to provide content only to the trigger, not to the content.

### 2. Inject it in `NcButton`

`NcButton` is the most popular `NcPopover` trigger.

This may look incorrect that `NcButton` "knows" something about the popover. But as it is a part of one library, and they are used together very often, it is a plain solution to cover a lot of cases. Without requirements to change anything in the apps.

Covers:
- `NcColorPicker`, `NcEmojiPicker`, `NcActions`
- Most of custom `NcPopover` usages with `NcButton` trigger **in all Nextcloud apps**

```html
<NcPopover>
  <template #trigger>
    <!-- 🥳 Magically this button will have all the attrs --> 
    <NcButton>My custom popover</NcButton>
  </template>
  
  Popover Content
</NcPopover>
```

### 3. Provide the attributes as slot props

To manually set them in other cases
- Covers `NcDatetimePicker`, `NcUserBubble`
- Allow very custom popover triggers in some Nextcloud apps

And the same in `NcColorPicker`

```html
<NcPopover>
  <template #trigger="{ attrs }">
    <!-- 🙂 No magic, but we can still just assign attrs --> 
    <button v-bind="attrs">My Very Custom Button</button>
  </template>
  
  Popover Content
</NcPopover>

<NcColorPicker v-slot="attrs">
  <button v-bind="attrs">My Very Custom Color Picker Button</button>
</NcColorPicker>
```

### 4. Add `role` prop to specify the role of the popover

See: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-haspopup

## 🖼️ Screenshots

| Component / Trigger                                                                                                          | Popover                                                                                                              |
|---------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------|
| **NcDateTimePicker**                                                                                                | image |
| image | image |
| **NcColorPicker**                                                                                                   | image |
| image | image |
| **Custom NcColorPicker**                                                                                            | image |
| image | .                                                                                                                   |
| **NcEmojiPicker**                                                                                                   | image |
| image | image |
| **Custom EmojiPicker**                                                                                              | image |
| image | .                                                                                                                   |
| **NcUserBubble**                                                                                                    | image |
| image | .                                                                                                                   |
| **NcActions** - Dialog                                                                                              | image |
| image | image |
| **NcActions** - Navigation                                                                                          | image |
| image | image |
| **NcActions** - Menu                                                                                                | image |
| image | image |
| **NcPopover** - Custom                                                                                              | image | 
| image | .                                                                                                                   |

### What still have to be done manually in place

- For custom trigger button without `NcButton`: `v-slot="{ attrs }" + v-bind="attrs"`
- For `NcPopover` content:
  - `popupRole` on `<NcPopover>`
  - The same `role` on the content
  - `aria-label` or `aria-labelledby` for dialog content

This is where a warning in the `debug` mode helps to find the issue:

image

### Alternative solution

Using DOM API by `querySelector` find any `button` in the `#trigger` slot and mutate it.

### 🏁 Checklist

### DIFF
--- a/src/components/NcPopover/NcPopover.vue
+++ b/src/components/NcPopover/NcPopover.vue
@@ -36,18 +36,18 @@ open prop on this component;
 
 ### Examples
 
-#### With a `<button>` as a trigger:
+#### With a `<NcButton>` as a trigger:
 
 ```vue
 <template>
 	<div style="display: flex">
-		<NcPopover>
+		<NcPopover popup-role="dialog">
 			<template #trigger>
 				<NcButton>I am the trigger</NcButton>
 			</template>
 			<template>
-				<form tabindex="0" @submit.prevent>
-					<h2>this is some content</h2>
+				<form tabindex="0" role="dialog" aria-labelledby="popover-example-dialog-header-1" @submit.prevent>
+					<h2 id="popover-example-dialog-header-1">this is some content</h2>
 					<p>
 						Lorem ipsum dolor sit amet, consectetur adipiscing elit. <br/>
 						Vestibulum eget placerat velit.
@@ -89,7 +89,7 @@ The prop `:focus-trap="false"` help to prevent it when the default behavior is n
 ```vue
 <template>
 	<div style="display: flex">
-		<NcPopover container="body" :popper-hide-triggers="(triggers) => [...triggers, 'click']">
+		<NcPopover container="body" :popper-hide-triggers="(triggers) => [...triggers, 'click']" popup-role="dialog">
 			<template #trigger>
 				<NcButton>I am the trigger</NcButton>
 			</template>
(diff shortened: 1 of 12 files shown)


ShGKme:
I'll add tests later when most of a11y issues are fixed 👉👈

susnux (approved):

src/components/NcPopover/NcPopover.vue
```diff
@@ -161,8 +238,36 @@ export default {
...
+	data() {
+		return {
+			internalShown: this.shown,
+		}
+	},
+
+	computed: {
+		shownProxy: {
+			get() {
+				return this.shown
```
raimund-schluessler, line 256: @ShGKme Shouldn't this be
```vue
			return this.internalShown
``` 
? With `internal` it does not work for vue3.

PR #5091, opened 2024-01-18
fix(NcButton): use `a` as tag for router-link

### LINKED ISSUE #5089: `NcBreadcrumb` does not render `<a>` anymore when `to` is provided
In #5077 we messed up a bit and introduced a regression while trying to fix the breadcrumbs behaviour when neither `to` nor `href` are provided.

The issue is that before, the breadcrumbs would *always*  render an `<a>`, both for `to` and `href`, because that is what `<router-link>` does by default. So also when `to`  is provided, we would have an `<a>` with the correct `href` set in the DOM. This is what e.g. the files app also relies on, we want the breadcrumbs to be `<a>`s there, also if it is really a `<router-link>`.

Now, `NcButton` behaves different. It by default renders a `<button>`, just when an `href` is provided (and no `to`) it renders an `<a>`. This breaks the breadcrumbs behaviour, as the crumbs are not `<a>`s anymore, but `<button>`s. Navigation still works though, but one cannot open the crumb in a new tab like a link.

The question is now how to fix this.

- One option would be to adjust `NcButton` to render an `<a>` if `to` is provided. But that might classify as breaking change to `NcButton`, although one could also consider it a fix. I am not sure here.
- The more conservative, but also more complex option would be to introduce a prop to configure `NcButton` to render an `<a>` or a `<button>` when `to` is provided.

@ShGKme @susnux @emoral435 For input. We will need to fix this before the next patch or minor release to not mess up the files app.

raimund-schluessler:
### ☑️ Resolves

* This PR adjusts `NcButton` to render an `a` if the `to` prop is provided instead of a `button` as before. This is the expected behaviour (for me) and fixes #5089.

### DIFF
--- a/src/components/NcButton/NcButton.vue
+++ b/src/components/NcButton/NcButton.vue
@@ -620,7 +620,8 @@ export default {
 			this)
 		}
 
-		const renderButton = ({ navigate, isActive, isExactActive } = {}) => h((this.to || !this.href) ? 'button' : 'a',
+		const isLink = (this.to || this.href)
+		const renderButton = ({ href, navigate, isActive, isExactActive } = {}) => h(isLink ? 'a' : 'button',
 			{
 				class: [
 					'button-vue',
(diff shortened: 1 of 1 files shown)


ShGKme (approved):

ShGKme:
Thanks!

emoral435 (approved):
Fixes the issue / regression that I (sorry) ade, so it looks good to me :) nice catch!

raimund-schluessler:
> Fixes the issue / regression that I (sorry) ade, so it looks good to me :) nice catch!

Don't worry, I had a good share in creating the problem :laughing: :see_no_evil: And we found it before a release, so no harm was done anyway.

PR #5095, opened 2024-01-18
fix(NcSelect): remove visual gap on top of the list

ShGKme:
### ☑️ Resolves

* Fix https://github.com/nextcloud-libraries/nextcloud-vue/issues/4943

`outline` shows a border on all the sides, including the top, which looks like a gap.

Using `box-shadow` allows adding outline only on 3 sides, except the top.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/NcSelect/NcSelect.vue
+++ b/src/components/NcSelect/NcSelect.vue
@@ -1221,9 +1221,13 @@ body {
 
 .vs__dropdown-menu {
 	border-color: var(--color-main-text) !important;
-	outline: 2px solid var(--color-main-background);
+	outline: none !important;
+	box-shadow:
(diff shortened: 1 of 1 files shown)


ShGKme:
This annoyed me every day 🙈

raimund-schluessler (changes requested):
Nice, I wasn't aware that box-shadow could be used for this :+1: 

But it needs a minor adjustment to work properly when the select opens to the top, see here:
image

ShGKme:
@raimund-schluessler thanks!

raimund-schluessler (approved):

Pytal (approved):
🤯 available since Chrome 10!

PR #5097, opened 2024-01-18

raimund-schluessler:
### ☑️ Resolves

#5086 is not really easy to backport / merge into `next`. :see_no_evil:


src/components/NcPopover/NcPopoverTriggerProvider.vue
```diff
@@ -0,0 +1,54 @@
...
+		if (window.OC?.debug) {
+			// TODO: Adjust for vue 3. Since vue 3 components can have multiple root elements, this breaks.
+			/**
+			const rootElement = this.$el
+			console.debug(this.$el)
+			const innerElement = this.$el.querySelector('[aria-expanded][aria-haspopup]')
+			if (!rootElement.getAttribute('aria-expanded') && !innerElement) {
+				warn('It looks like you are using a custom button as a <NcPopover> or other popover #trigger. If you are not using <NcButton> as a trigger, you need to bind attrs from the #trigger slot props to your custom button. See <NcPopover> docs for an example.')
+			}
+			 */
```
raimund-schluessler, line 44: @ShGKme I don't really have an idea how to migrate this to vue 3. Do you know how to? If not, maybe we merge anyway for now, it's just a debug message after all.

src/components/NcPopover/NcPopover.vue
```diff
@@ -157,8 +234,36 @@ export default {
...
+	data() {
+		return {
+			internalShown: this.shown,
+		}
+	},
+
+	computed: {
+		shownProxy: {
+			get() {
+				return this.internalShown
```
raimund-schluessler, line 252: @ShGKme  I had to adjust this from `shown` to `internalShown` to make it work. See https://github.com/nextcloud-libraries/nextcloud-vue/pull/5086/files#r1458568969.

skjnldsv (approved):

ShGKme (approved):

src/components/NcPopover/NcPopoverTriggerProvider.vue
susnux: Does this really breaks for vue3? Are the elements not in the same placeholder that will be returned by `this.$el` in case of multi-root elements?

If it break, how about moving to NcPopover and assigning a template ref?

raimund-schluessler: One problem is that even linebreaks in the markup can lead to empty text nodes which have no `querySelector` function. Then it breaks (or simply doesn't warn after we merged #5104).

Using `ref` sounds like an option, though.

ShGKme: > If it break, how about moving to NcPopover and assigning a template ref?

I don't understand, how do you want to use a template ref here?

`ref` on `<NcPopoverTriggerProvider>` will be exactly the same object that we have in `this` in the `mounted`

Inside `NcPopoverTriggerProvider` there is no node to put a `ref` on.

raimund-schluessler: I will merge this now, I hope this is ok with you. These backports / merges get increasingly error prone the more commits we have in here. I will create an issue so we do not forget to fix this warning for vue 3 and try to figure something out.

raimund-schluessler: See https://github.com/nextcloud-libraries/nextcloud-vue/issues/5108

PR #5101, opened 2024-01-20
fix(NcHeaderMenu): mouse in now pointer

emoral435:
### 🖼️ Screenshots

🏡 After
---
firefox_UBN04LrlJO

### 🚧 Tasks

- make sure the the pointer is cursor on all elements that can be input into the #icon slot

### 🏁 Checklist

### DIFF
--- a/src/components/NcHeaderMenu/NcHeaderMenu.vue
+++ b/src/components/NcHeaderMenu/NcHeaderMenu.vue
@@ -73,18 +73,21 @@ export default {
 		class="header-menu"
 		v-on="listeners">
 		<!-- Trigger -->
-		<button :id="isNav ? triggerId : null"
+		<NcButton :id="isNav ? triggerId : null"
 			ref="trigger"
-			class="header-menu__trigger button-vue"
+			type="tertiary-no-background"
+			class="header-menu__trigger"
 			:aria-label="ariaLabel"
 			:aria-describedby="description ? descriptionId : null"
 			:aria-controls="`header-menu-${id}`"
 			:aria-expanded="opened.toString()"
 			@click.prevent="toggleMenu">
-			<!-- @slot Icon trigger slot. Make sure the svg path
+			<template #icon>
+				<!-- @slot Icon trigger slot. Make sure the svg path
 				is at least 16px. Usually mdi icon works at 20px -->
-			<slot name="trigger" />
-		</button>
+				<slot name="trigger" />
+			</template>
+		</NcButton>
 
 		<span v-if="description"
 			:id="descriptionId"
@@ -115,8 +118,14 @@ import GenRandomId from '../../utils/GenRandomId.js'
 import { clickOutsideOptions } from '../../mixins/index.js'
 import { getTrapStack } from '../../utils/focusTrap.js'
 
+import NcButton from '../NcButton/index.js'
+
 export default {
 	name: 'NcHeaderMenu',
+
+	components: {
+		NcButton,
+	},
 
 	directives: {
 		ClickOutside,
@@ -339,17 +348,9 @@ $externalMargin: 8px;
 	height: var(--header-height);
 
 	&__trigger {
-		display: flex;
-		align-items: center;
-		justify-content: center;
 		width: var(--header-height);
 		height: var(--header-height);
-		margin: 0;
-		padding: 0;
-		cursor: pointer;
 		opacity: .85;
-		background-color: transparent;
-		border: none;
 
 		// header is filled with primary or image background
 		filter: none !important;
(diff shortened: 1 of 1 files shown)


src/components/NcHeaderMenu/NcHeaderMenu.vue
```diff
@@ -354,6 +354,12 @@ $externalMargin: 8px;
 		// header is filled with primary or image background
 		filter: none !important;
 		color: var(--color-primary-text) !important;
+
+		// NOTE when migrating to Vue3, while not deprecated, a better solution would be to style the child slotted element
+		// LINK https://stackoverflow.com/a/55368933
+		:deep * {
+			cursor: pointer !important;
+		}
```
ShGKme, line 362: This is already done by `NcButton`: 
https://github.com/nextcloud-libraries/nextcloud-vue/blob/2bbb1b4eeffbc766341fe4b2d606f83ac7b2b971/src/components/NcButton/NcButton.vue#L728-L731

For some reason, button styles don't work on `ContactsMenu`.

ShGKme, line 362: This is already covered by `NcButton`. And in header menu icons it works where apps use `NcButton` as a trigger without `NcButton`. Would it be possible to also use `NcButton` instead of `<button>` in this component?

ShGKme: > NOTE when migrating to Vue3, while not deprecated, a better solution would be to style the child slotted element

BTW, `:slotted` would not be enough here. It works only on root elements in slot, like with `> *`.

emoral435: Hello, and happy 'beginning of the work week'! But I agree, 

> Would it be possible to also use <NcButton> instead of <button> in this component?

I think that using our own native button would be better. Will use in the new PR!

ShGKme (approved):

src/components/NcHeaderMenu/NcHeaderMenu.vue
```diff
@@ -339,14 +348,8 @@ $externalMargin: 8px;
...
-		display: flex;
-		align-items: center;
-		justify-content: center;
 		width: var(--header-height);
 		height: var(--header-height);
-		margin: 0;
-		padding: 0;
-		cursor: pointer;
 		opacity: .85;
 		background-color: transparent;
```
susnux, line 354: I think this is no longer needed due to `tertiary-no-background`

emoral435: Gotcha! Will remove right now :)

emoral435: Sorry for the delay - has been removed now :D

Pytal (changes requested):
Trigger button still needs adjustments to be styled the same as before

Before | After
--- | ---
image | image image

src/components/NcHeaderMenu/NcHeaderMenu.vue
```diff
@@ -363,10 +364,6 @@ $externalMargin: 8px;
 		opacity: 1;
 	}
 
-	&__trigger:focus-visible {
```
susnux, line 366: I think this needs to stay, see @Pytal comments

emoral435: I tried leaving this in to take out the border outline, but since the NcButton has the `!important` styling precedent on its outline and bordering styling, I had to use :deep styling in the newest PR.

`NcButton`
image

And this did not work :(
image

susnux: yes because both have the same specificity, you also can just increase it, currently the style is just on `.header-menu__trigger`, but it should work with `.header-menu .header-menu__trigger:focus-visible`.

Something like:
```scss
.header-menu {
    // ...
    #{&}__trigger:focus-visible {
        // here
    }
}
```

susnux: not sure if it is better just an option ;)

Pytal (approved):
Code looks code, style looks style!

susnux (approved):

src/components/NcHeaderMenu/NcHeaderMenu.vue
emoral435: Ahhh, only now do I get what you mean! I changed it to how you requested, actually, because I am going to fix this problem on the unified search button right now, and for code styling being consistant, I will use your solution (plus I try not to use `:deep` as much as possible)

Pytal:
There is a slight regression where the trigger button width of 50px is being overidden by NcButton width of 44px @emoral435

image

emoral435:
Gotcha 👍 will fix that right now, and send out the new version. Embarrassed that I did not notice this 🫠 thank you!

emoral435:
Regression fixed in nextcloud-libraries/nextcloud-vue#5129 ! @Pytal

PR #5103, opened 2024-01-22
enh(NcEmojiPicker): Always show skin tone selector + save selection

### LINKED ISSUE #5084: Ability to change skin tone for emojis
Currently it is not possible to change the skin tone for emojis in the picker, it is always the yellow default.

The library we use actually allows for this and the design looks exactly like Twitter, it’s just something we have to enable again: https://serebrov.github.io/emoji-mart-vue/

susnux:
### ☑️ Resolves

* Resolves #5084 

Always allow to select the skin tone used for emojis, this can be done by picking the skin tone on the left of the search bar.
The selected skin tone is saved on the local storage of the browser, so it is preserved during sessions.

Previously it was only possible to select a skin tone if the preview was enabled, but this has two problems:
1. Not always possible to select
2. :exclamation: The preview is for some emojis too crowed (too many shortcuts) so that the skin tone selector was not visible

Some minor other changes on components to implement this:
* Allow to limit NcColorPicker to the given palette
* Export `Color` class so we can reuse it
* Fix Color to provide `color` attribute required by the color picker (otherwise one would have to provide it manually)

### 🖼️ Screenshots

https://github.com/nextcloud-libraries/nextcloud-vue/assets/1855448/ad3b2c53-4215-4b3b-ba26-5a257a8ac3b0

### 🚧 Tasks

- We need this to make the skin tone selection reactive: https://github.com/serebrov/emoji-mart-vue/pull/294

### 🏁 Checklist


max-nextcloud (approved):
Code changes look good. Did not try it out though.

jancborchardt (commented):
Very nice! 2 questions:
- Since there are 5 modifiers + default yellow, could the color dropdown be a 3*2 grid instead of 4*2 with only 2 in the bottom row? That would look cleaner. :)
- Once this color is set, will it also be used in cases where you use `:` for emojis like in Text, Talk, Collectives, Mail, etc?

Thanks!

src/components/NcEmojiPicker/NcEmojiPicker.vue
```diff
@@ -187,17 +201,24 @@ This component allows the user to pick an emoji.
...
+import NcColorPicker from '../NcColorPicker/NcColorPicker.vue'
+import NcPopover from '../NcPopover/index.js'
+import NcTextField from '../NcTextField/index.js'
 
-// Shared emoji index for all NcEmojiPicker instances
+// Shared emoji index and skinTone for all NcEmojiPicker instances
 // Will be initialized on the first NcEmojiPicker creating
 let emojiIndex
 
+const storage = getBuilder(`${appName}`).persist(true).build()
```
ShGKme, line 220: I'm not sure it should be per-app storage. If a user has chosen the color in one app on one page, they would expect it to work on other pages.

jancborchardt: Yep, fully agree, should be saved per-account. :)

susnux: Then use a generic scope. :white_check_mark:

susnux:
> Since there are 5 modifiers + default yellow, could the color dropdown be a 32 grid instead of 42 with only 2 in the bottom row? That would look cleaner. :)

This comes from NcColorPicker and should probably be fixed there, would do so in a follow up.

jancborchardt:
> > Since there are 5 modifiers + default yellow, could the color dropdown be a 32 grid instead of 42 with only 2 in the bottom row? That would look cleaner. :)
> 
> This comes from NcColorPicker and should probably be fixed there, would do so in a follow up.

@susnux sounds good! :)

> Once this color is set, will it also be used in cases where you use `:` for emojis like in Text, Talk, Collectives, Mail, etc?

What about this one? Also cc @max-nextcloud for Text & Collectives, maybe you know?

ShGKme (approved):

susnux:
> Once this color is set, will it also be used in cases where you use `:` for emojis like in Text, Talk, Collectives, Mail, etc?

I need to test this but I think we need to adjust the `emojiSeach` function used in `NcRichContenteditable`. I will check an report.

susnux:
> I need to test this but I think we need to adjust the `emojiSeach` function used in `NcRichContenteditable`. I will check an report.

No it did not work, but adjusting `emojiSearch` worked:

So @jancborchardt all points resolved and working :tada: 


https://github.com/nextcloud-libraries/nextcloud-vue/assets/1855448/4967f99e-34fa-4966-9b1a-fea6c9012294

jancborchardt (approved):
Nice, awesome job @susnux! :)

src/components/NcEmojiPicker/NcEmojiPicker.vue
```diff
@@ -145,23 +145,37 @@ This component allows the user to pick an emoji.
...
+						:label="t('Search')"
+						:label-visible="true"
+						:placeholder="i18n.search"
+						trailing-button-icon="close"
+						:trailing-button-label="t('Clear search')"
+						:show-trailing-button="search !== ''"
+						@trailing-button-click="clearSearch(); slotProps.onSearch(search);"
+						@update:value="slotProps.onSearch(search)" />
+					<NcColorPicker palette-only
+						:palette="skinTonePalette"
```
Antreesy, line 169: Missing a container here for inner NcPopover. Because of that, skin tones are rendered always on 'body'

PR #5104, opened 2024-01-22
fix(NcPopoverTriggerProvider): Make debugging code more defensive

susnux:
### ☑️ Resolves

The code asserts that `this.$el` is always set or an `Element` but it is not, so in some cases errors like this could happen:

```
[Vue warn]: Error in mounted hook: "TypeError: this.$el.querySelector is not a function"
```

image

### 🏁 Checklist

### DIFF
--- a/src/components/NcPopover/NcPopoverTriggerProvider.vue
+++ b/src/components/NcPopover/NcPopoverTriggerProvider.vue
@@ -34,8 +34,8 @@ export default defineComponent({
 	mounted() {
 		if (window.OC?.debug) {
 			const rootElement = this.$el
-			const innerElement = this.$el.querySelector('[aria-expanded][aria-haspopup]')
-			if (!rootElement.getAttribute('aria-expanded') && !innerElement) {
+			const innerElement = this.$el?.querySelector?.('[aria-expanded][aria-haspopup]')
+			if (!rootElement?.getAttribute?.('aria-expanded') && !innerElement) {
 				Vue.util.warn('It looks like you are using a custom button as a <NcPopover> or other popover #trigger. If you are not using <NcButton> as a trigger, you need to bind attrs from the #trigger slot props to your custom button. See <NcPopover> docs for an example.')
 			}
 		}


raimund-schluessler (approved):
That is a good idea. We will still need to adjust it for vue 3, though.

src/components/NcPopover/NcPopoverTriggerProvider.vue
```diff
@@ -34,8 +34,8 @@ export default defineComponent({
 	mounted() {
 		if (window.OC?.debug) {
 			const rootElement = this.$el
-			const innerElement = this.$el.querySelector('[aria-expanded][aria-haspopup]')
-			if (!rootElement.getAttribute('aria-expanded') && !innerElement) {
+			const innerElement = this.$el?.querySelector?.('[aria-expanded][aria-haspopup]')
+			if (!rootElement?.getAttribute?.('aria-expanded') && !innerElement) {
```
ShGKme, line 38: Not important to the current PR, but have seen this in many places. Optional chaining `?.` works for a whole chain.

```js
const a1 = null
const a2 = { b: { c: { d() } } }
// This is enough to check for a1 == null
a1?.b.c.d()
a2?.b.c.d()
// This is unneeded and adds redundant huge code after transpiling
a1?.b?.c?.d?.()
```

Here it is guaranteed that if html element exists, it has `querySelector` and `getAttributes` methods.

ShGKme (changes requested):
In Vue 2 it doesn't fix the issue, because multiple root nodes are broken anyway (I forgot about this limitation for renderless components in Vue 2)

image

In Vue 3 it hides the error in the console, but removes this check and isn't more helpful than removing the check.

susnux:
> In Vue 2 it doesn't fix the issue, because multiple root nodes are broken anyway

It is not about multiple root nodes,I experienced that error on valid vue2 code.

ShGKme:
> It is not about multiple root nodes,I experienced that error on valid vue2 code.

Do you have an example?

susnux:
> Do you have an example?

Simplest one: Do not provide any `#trigger` slot content. Still valid vue code, but the value of `$el` will be an empty string.

susnux:
fixed by @ShGKme

PR #5109, opened 2024-01-22
Remove inaccessible opacity from SettingsSection

JuliaKirschenheuter:
### ☑️ Resolves

* For https://github.com/nextcloud/server/issues/42941

Opacity is a bad choice for a11y in general. Let's remove it.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot from 2024-01-22 11-21-09 | Screenshot from 2024-01-22 11-21-21

### 🏁 Checklist

### DIFF
--- a/src/components/NcSettingsSection/NcSettingsSection.vue
+++ b/src/components/NcSettingsSection/NcSettingsSection.vue
@@ -154,17 +154,17 @@ $maxWidth: 900px;
 		// make sure to properly align the icon with the text
 		margin: -$icon-margin;
 		margin-left: 0;
-		opacity: $opacity_normal;
+		color: var(--color-text-maxcontrast);
 
 		&:hover, &:focus, &:active {
-			opacity: $opacity_full;
+			color: var(--color-main-text);
 		}
 	}
 
 	&__desc {
 		margin-top: -.2em;
 		margin-bottom: 1em;
-		opacity: $opacity_normal;
+		color: var(--color-text-maxcontrast);
 		max-width: $maxWidth;
 	}
 }


szaimen (approved):

ShGKme (commented):
This has a design impact. IMO, it looked good with a bit gray text.

What about changing the color to:
```
```

Should be fine contrast-wise

image

image

src/components/NcSettingsSection/NcSettingsSection.vue
```diff
@@ -154,7 +154,7 @@ $maxWidth: 900px;
 		// make sure to properly align the icon with the text
 		margin: -$icon-margin;
 		margin-left: 0;
-		opacity: $opacity_normal;
+		color: var(--color-text-lighter);
```
ShGKme, line 157: By the way, this changes the help icon, not the text, and as a result, removes the hover effect from the help icon link. Is it intended?

JuliaKirschenheuter: opacity is on general case bad choice. Hover effect itself is not as important but if this hover effect is not accessible - there is a11y problem again ;(

susnux, line 157: This variable is deprecated and will shortly removed, use its replacement instead:
```suggestion
		color: var(--color-text-maxcontrast);
```

src/components/NcSettingsSection/NcSettingsSection.vue
```diff
@@ -164,7 +164,7 @@ $maxWidth: 900px;
 	&__desc {
 		margin-top: -.2em;
 		margin-bottom: 1em;
-		opacity: $opacity_normal;
+		color: var(--color-text-lighter);
```
susnux, line 167: This variable is deprecated and will shortly removed, use its replacement instead:
```suggestion
		color: var(--color-text-maxcontrast);
```

src/components/NcSettingsSection/NcSettingsSection.vue
susnux: Then maybe add a hover effect by making it `--color-main-text` on hover.

susnux (changes requested):
Please do not use `color-text-lighter`

ShGKme (approved):

emoral435 (approved):
Haven't tested this on my local dev env to check , but it looks good to me!

susnux (approved):

PR #5110, opened 2024-01-22
fix(NcPopover): check trigger a11y compatible with Vue 3

ShGKme:
### ☑️ Resolves

* Fix https://github.com/nextcloud-libraries/nextcloud-vue/issues/5108
* Alternative to https://github.com/nextcloud-libraries/nextcloud-vue/pull/5104 to keep check in Vue 3

Popover content is usually only the trigger. So no need to check content in a slot component.

Trigger's parent DOM node is accessible by `$refs` of Floating-Vue.

For Vue 3 only requires to change `refs` from Floating-Vue.

### 🏁 Checklist

### DIFF
--- a/src/components/NcPopover/NcPopover.vue
+++ b/src/components/NcPopover/NcPopover.vue
@@ -178,6 +178,7 @@ See: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/
 </template>
 
 <script>
+import Vue from 'vue'
 import { Dropdown } from 'floating-vue'
 import { createFocusTrap } from 'focus-trap'
 import { getTrapStack } from '../../utils/focusTrap.js'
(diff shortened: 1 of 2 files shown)


raimund-schluessler:
Shouldn't this trigger the warning? But it does not in the docs (unless they aren't built as debug, but I think they are). It also gives no warning for `master`.

```vue
<template>
	<div style="display: flex">
		<NcPopover>
			<!-- Take "attrs" from the slot props -->
			<template #trigger>
				<!-- Bind attrs as the button attrs -->
				<button>
					I am a custom button in the trigger2
				</button>
			</template>

			Hi! 🚀
		</NcPopover>
	</div>
</template>
```

ShGKme:
Works for me...

image

raimund-schluessler:
> Works for me...

Weird. I guess https://deploy-preview-5110--nextcloud-vue-components.netlify.app/#/Components/NcPopover is not debug mode then.

ShGKme:
> Weird. I guess https://deploy-preview-5110--nextcloud-vue-components.netlify.app/#/Components/NcPopover is not debug mode then.

It is... Let me check

ShGKme:
> Weird. I guess https://deploy-preview-5110--nextcloud-vue-components.netlify.app/#/Components/NcPopover is not debug mode then.

Oh, yes.

It is in `OC.debug` mode, but it uses `production` mode of Vue, so there is no Vue warnings.

raimund-schluessler (approved):
Works locally 👍

susnux (approved):

PR #5120, opened 2024-01-23
Fix additionalTrapElements to accept HTMLelements as well

GretaD:
### ☑️ Resolves

* Fix https://github.com/nextcloud/mail/pull/9144#issuecomment-1900748293

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 Tasks

- [x] Test if it works on mail, it works.

### 🏁 Checklist

### DIFF
--- a/src/components/NcDialog/NcDialog.vue
+++ b/src/components/NcDialog/NcDialog.vue
@@ -162,7 +162,14 @@ export default defineComponent({
 		/** Additional elements to add to the focus trap */
 		additionalTrapElements: {
 			type: Array,
-			validator: (arr) => Array.isArray(arr) && arr.every((element) => typeof element === 'string'),
+			validator: (arr) => {
+				return (
+					Array.isArray(arr) && arr.every(
+						(element) =>
+							typeof element === 'string' || element instanceof HTMLElement,
+					)
+				)
+			},
 			default: () => ([]),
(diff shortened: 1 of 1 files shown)


ShGKme (approved):

GretaD:
the error is gone, i can see the tool, but i cannot see the dropdown list. It is shown for a second and goes away. But im not sure if this issue with the drop down is connected to the trap element

src/components/NcDialog/NcDialog.vue
```diff
@@ -162,7 +162,14 @@ export default defineComponent({
 		/** Additional elements to add to the focus trap */
```
susnux, line 162: Maybe add that we accept the same as NcModal

susnux (approved):
Nice! But will not work. We need to pass it to `NcModal`.
This was forgotten :see_no_evil: So add `additionalTrapElements` to `modalProps`

GretaD:
> Nice! But will not work. We need to pass it to `NcModal`. This was forgotten 🙈 So add `additionalTrapElements` to `modalProps`

ah, ok, will do. test and then will merge this one

GretaD:
tested and works now :)

susnux (approved):

PR #5124, opened 2024-01-23

raimund-schluessler:
### ☑️ Resolves

This was one hell of a backport :see_no_evil: But I think I got everything working. And it also fixes the search box of `NcEmojiPicker` that currently does not work for `next`.

After this is merged, I want an alpha release of `nextcloud/vue@9.0.0-alpha.0` :wink:


susnux (approved):

Antreesy:
Ah, I had a prepared backport at my laptop for you 🙈

cypress/component/richtext.cy.ts
```diff
@@ -515,6 +568,49 @@ describe('NcRichText', () => {
...
+						text: testCases.map(i => i.input).join('\n'),
+						useExtendedMarkdown: true,
+					},
+				})
+
+				cy.get('ul').should('exist')
+				cy.get('li').should('have.length', testCases.length)
+				cy.get('li').each((item, index) => {
+					// Vue 2.7 renders three non-breaking spaces here for some reason
+					expect(item).have.text('   ' + testCases[index].output)
```
Antreesy, line 590:
```suggestion
					expect(item).have.text(testCases[index].output)
```
This test fails for me at least locally

Antreesy (approved):

cypress/component/richtext.cy.ts
ShGKme: Probably, should be `contains`

ShGKme (approved):

cypress/component/richtext.cy.ts
raimund-schluessler: Weird, the test seems to run fine on CI, no?

Antreesy: Seems to run fine, as we see it. But I believe, Grigorii's proposal should fit here better

susnux: I also have this issue on master locally.

PR #5130, opened 2024-01-24
fix: removed groupings in timezone picker

emoral435:
### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
firefox_zBqhJRiF04|firefox_9PTdGieRfq

### 🚧 Tasks

- [x] Removed the groupings from the TimeZone picker. This does two things:

1.  Removes the styling issue presented in the before picture (contrast ratio for the heading of the grouping not being correct. This is due to NcSelect marking the header as a non-selectable element due to us passing in the header as an object)
2. Now leaves us (for later!) to decide how to better organize the groupings better visually, as right now, I personally do not like the presentation of the areas. But that is me, personally, and we can always work on a better design later :)

### 🏁 Checklist

### DIFF
--- a/src/components/NcTimezonePicker/NcTimezonePicker.vue
+++ b/src/components/NcTimezonePicker/NcTimezonePicker.vue
@@ -115,15 +115,18 @@ export default {
 			/**
 			 * Since NcSelect does not support groups,
 			 * we create an object with the grouped timezones and continent labels.
+			 *
+			 * NOTE for now we are removing the grouping from the fields to fix an accessibility issue
+			 * in the future, other options can be introduced to better display the different areas
 			 */
 			let timezonesGrouped = []
 			Object.values(timezoneList).forEach(group => {
 				// Add an entry as group label
-				timezonesGrouped.push({
-					label: group.continent,
-					timezoneId: `tz-group__${group.continent}`,
-					regions: group.regions,
-				})
+				// timezonesGrouped.push({
+				// label: group.continent,
+				// timezoneId: `tz-group__${group.continent}`,
+				// regions: group.regions,
+				// })
 				timezonesGrouped = timezonesGrouped.concat(group.regions)
 			})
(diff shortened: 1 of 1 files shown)


emoral435:
Also - would love to have the opinion from all the reviewers. 

Would you personally leave the commented out code? This is a library, and so I am leaning on taking out the commented code (it was what was causing the problem in the first place), but due to the fact that we may later group the areas again, it MAY be used fr the logic of grouping areas (probably not, though, as all it does it create an object that NcSelect doesnt let you click on)

skjnldsv (approved):

Pytal:
> Would you personally leave the commented out code?

In general I would remove it, otherwise we'd have a lot of commented code laying around *just in case*

Pytal (approved):
Other than removing the commented code

ShGKme:
> In general I would remove it, otherwise we'd have a lot of commented code laying around _just in case_

+1 here. We always can re-take the code from history when it is needed

ShGKme (approved):
Sad, not we have no other easy solution with NcSelect for now

susnux:
> +1 here. We always can re-take the code from history when it is needed

Thats why we use a *version* control system :wink:

susnux (approved):

raimund-schluessler:
As I understand it, the problem was just about the constrast ratio, no? Why didn't we simply adjust the colors in the CSS of the picker? Or was there another issue that was not described?

ShGKme:
> As I understand it, the problem was just about the constrast ratio, no? Why didn't we simply adjust the colors in the CSS of the picker? Or was there another issue that was not described?

If I remember correctly, no.

The main issue is about semantics. We cannot have groups only visually, where group is only defined by disabled and designed as a header list item. Groups have to be really a group with a header inside.

cc @emoral435 @Pytal  @JuliaKirschenheuter to clarify, if I'm wrong

PR #5131, opened 2024-01-25
fix(NcAvatar): Remove span wrapper button semantics in favour of internal button components

Pytal:
### ☑️ Resolves

- For https://github.com/nextcloud/server/issues/42957

### Summary

<details>
<summary>Previous iteration</summary>

- Previously the tab sequence would go from wrapper span[role="button"] -> action menu button -> next element on page. Only the span focus would show visual focus while the action menu button had none so there would appear to be a gap in the tab sequence.
- W3 states that "a button element allows phrasing content as descendants, and does not allow interactive content or descendants with a tabindex attribute. Therefore, any elements specified with a role=button would follow the same descendant restrictions, and not allow any interactive content descendants, elements with a tabindex specified, or any elements with role values that are in the interactive content category" -https://www.w3.org/TR/html-aria/#allowed-descendants-of-aria-roles
- The buttons inside span[role="button"] would be classified as interactive content and so is invalid HTML
- Setting tabindex="-1" would violate the above requirement and only removes sequential focus but not focusability via click or programmatic .focus()
- The NcAvatar wrapper span acts as a button with role="button" + event listeners and for the purposes of this component should be the only interative element which programmatically activates NcActions. This makes direct interaction on the NcActions button redundant.
- All descendants of a role="button" element implictly have role="presenation" but because browsers ignore role="presentation" on focusable elements (the buttons) we have to set inert explicitly
- This inert fix corrects the invalid HTML and removes redundant behaviour

</details>

- Remove button semantics from span wrapper
- Fixes focus return on Escape from menu
- Uses button semantics from internal button components

### 🏁 Checklist

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -984,6 +984,7 @@ export default {
 		'close',
 		'focus',
 		'blur',
+		'click',
 	],
 
 	data() {
@@ -1222,6 +1223,17 @@ export default {
 		onBlur(event) {
 			this.$emit('blur', event)
 		},
+		onClick(event) {
+			/**
+			 * Event emitted when the menu toggle button is clicked.
+			 *
+			 * This is e.g. necessary for the NcAvatar component
+			 * which needs to fetch the menu items on click.
+			 *
+			 * @type {PointerEvent}
+			 */
+			this.$emit('click', event)
+		},
 	},
 
 	/**
@@ -1420,6 +1432,7 @@ export default {
 						on: {
 							focus: this.onFocus,
 							blur: this.onBlur,
+							click: this.onClick,
 						},
 					}, [
 						h('template', { slot: 'icon' }, [triggerIcon]),
--- a/src/components/NcAvatar/NcAvatar.vue
+++ b/src/components/NcAvatar/NcAvatar.vue
@@ -115,16 +115,8 @@ export default {
 			'avatardiv--with-menu': hasMenu,
 			'avatardiv--with-menu-loading': contactsMenuLoading
 		}"
-		:title="tooltip"
 		:style="avatarStyle"
-		class="avatardiv popovermenu-wrapper"
-		:tabindex="hasMenu ? '0' : undefined"
-		:aria-label="avatarAriaLabel"
-		:role="hasMenu ? 'button' : undefined"
-		v-on="hasMenu ? {
-			click: toggleMenu,
-			keydown: toggleMenu,
-		} : null">
+		class="avatardiv popovermenu-wrapper">
 		<!-- @slot Icon slot -->
 		<slot name="icon">
 			<!-- Avatar icon or image -->
@@ -137,10 +129,12 @@ export default {
 
 		<!-- Contact menu -->
 		<!-- We show a button if the menu is not loaded yet. -->
-		<NcButton v-if="hasMenu && !menu.length"
-			:aria-label="t('Open contact menu')"
+		<NcButton v-if="hasMenu && menu.length === 0"
 			type="tertiary-no-background"
-			class="action-item action-item__menutoggle">
+			class="action-item action-item__menutoggle"
+			:aria-label="avatarAriaLabel"
+			:title="tooltip"
+			@click="toggleMenu">
 			<template #icon>
 				<NcLoadingIcon v-if="contactsMenuLoading" />
 				<DotsHorizontal v-else :size="20" />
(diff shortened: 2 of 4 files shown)


ShGKme:
If I understand the issue, the problem is that we have `span[role="button"]` and inside another one `button`. The current solution makes the button inert. Why cannot we make the opposite, keep normal `NcActions` trigger button and remove `role=button` with `tabindex` from the span?

ShGKme:
Oh, We even don't use NcButton here as direct NcActions trigger...

Pytal:
> If I understand the issue, the problem is that we have `span[role="button"]` and inside another one `button`. The current solution makes the button inert. Why cannot we make the opposite, keep normal `NcActions` trigger button and remove `role=button` with `tabindex` from the span?

Structurally it doesn't make sense to me at least to have a child of the component i.e. NcActions/NcButton handle the menu activation

Having the root component wrapper element, which is needed for a lot of customized NcAvatar behaviour, handle menu toggle makes more sense in my opinion

ShGKme:
> Having the root component wrapper element, which is needed for a lot of customized NcAvatar behaviour, handle menu toggle makes more sense in my opinion

I didn't mean that we need to change the layout and to span/image inside the button. Only move role/tabindex/events to the button.

src/components/NcActions/NcActions.vue
```diff
@@ -976,6 +976,14 @@ export default {
 			type: Number,
 			default: 0,
 		},
+
+		/**
+		 * The `inert` attribute for the menu toggle button
+		 */
+		inert: {
+			type: Boolean,
+			default: false,
```
artonge, line 985: Will it ever be something else than `false`?

JuliaKirschenheuter (approved):
Thanks Chris for cleaning this up!

Works! One question: could we return a focus to the avatar (action button) when action menu is closed? Or is it another issue?

artonge (commented):
I think I agree with @ShGKme. Why can't we remove the `role="button"` from the `<span>`?

src/components/NcActions/NcActions.vue
ShGKme: It is `inert` in the current fix for `NcAvatar`

emoral435:
I actually personally agree with @Pytal here, I think that it makes more sense to have the wrapper for `NcAvatar` have the precedent over the button when it comes to being the one that accepts the focus. This also allows us to remove the focus from the `NcActions` component in the future, in case something similar happens.

emoral435 (approved):
Both solutions proposed works, all just a matter of preference 👍

Pytal:
> Works! One question: could we return a focus to the avatar (action button) when action menu is closed? Or is it another issue?

Automatically determined in NcActions here https://github.com/nextcloud-libraries/nextcloud-vue/blob/7cee513cf568a9ccc1691ffc82351c22302ba992/src/components/NcActions/NcActions.vue#L1387

ShGKme:
> Automatically determined in NcActions here

Automatic return focus from `NcActions` doesn't work here because this `NcActions` has no trigger button.

Pytal:
Moved the button semantics to the actual buttons rather than the wrapper span while preserving the same UX as before (slightly strange NcActions click triggering toggleMenu to fetch the latest contactsmenu data)

tests/unit/components/NcAvatar/NcAvatar.spec.ts
```diff
@@ -42,7 +42,7 @@ describe('NcAvatar.vue', () => {
 		await nextTick()
 
 		expect(wrapper.find('.avatardiv__user-status').exists()).toBe(true)
-		expect(wrapper.attributes('aria-label')).toBe('Avatar of J. Doe, online')
+		expect(wrapper.find('.action-item__menutoggle').attributes('aria-label')).toBe('Avatar of J. Doe, online')
```
susnux, line 45: still not a fan of testing using classes (better semantially like `button[aria-label]` or similar).
But the tests already do this so no blame just my 5 cents

susnux (approved):
Looks much better to me!

src/components/NcAvatar/NcAvatar.vue
```diff
@@ -115,16 +115,8 @@ export default {
...
 		}"
-		:title="tooltip"
 		:style="avatarStyle"
-		class="avatardiv popovermenu-wrapper"
-		:tabindex="hasMenu ? '0' : undefined"
-		:aria-label="avatarAriaLabel"
-		:role="hasMenu ? 'button' : undefined"
-		v-on="hasMenu ? {
-			click: toggleMenu,
-			keydown: toggleMenu,
```
Antreesy, line 126: Seems to be covered by `@click` event as well (Spacebar, Enter keys)

Antreesy (approved):
Seems legit, tested listeners and new template

src/components/NcAvatar/NcAvatar.vue
ShGKme: Yep, in real button should just work

PR #5133, opened 2024-01-25
fix(fields): adjust label line-height

ShGKme:
Before, it had the same line-height in docs, but server has different styles.
See: https://github.com/nextcloud-libraries/nextcloud-vue/pull/5132

Manually set the same line-height, which is the minimum allowed for a11y and was supposed to be set during development.

### ☑️ Resolves

* Fix https://github.com/nextcloud-libraries/nextcloud-vue/issues/5128

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist


susnux (approved):

JuliaKirschenheuter (approved):
checked text spacing and looks good ;)

PR #5138, opened 2024-01-25
fix(NcMentionBubble): inject component styles in NcRichContenteditable

### LINKED ISSUE #4901: `NcRichContenteditable`: @mention item is not rendered properly in other apps
> Am I right that now styles are not loaded from `.js` but loaded by `require('path-to.css')` inside modules?

Might be related to it?
- with import of `NcRichContenteditable` to the project also the component `NcMentionBubble` is bundled with id '357e6d0e' (which will be used for styles scope in `[data-v-357e6d0e]`)
- `NcRichContenteditable` is using mixin `richEditor` under the hood, which is in charge of rendering `NcMentionBubble`. It is also bundled in the project with id '8a961b36'
- `dist/assets/NcMentionBubble-<hash>.css` contains styles for scope `data-v-357e6d0e`, same as source `webpack://./node_modules/@nextcloud/vue/dist/assets/index.css` in production code
- dev environment seems to use mjs, where styles applied only for `[data-v-8a961b36]`

That causes the styles for rendered mentions to be broken, if lib is used in Talk, for example:
image

Content in a bundled app:
1. NcMentionBundle with id '357e6d0e'
2. Styles for id '8a961b36'
3. NcMentionBundle with id '8a961b36'
Screenshot from 2023-10-12 10-26-42

_Originally posted by @Antreesy in https://github.com/nextcloud-libraries/nextcloud-vue/issues/4386#issuecomment-1759250020_

Antreesy:
### ☑️ Resolves

* Fix #4901
  * :-1: `genSelectTemplate` returns a HTML template of NcMentionBubble component with scoped style attribute `data-v-123`, but these styles are not included in the bundled chunk for some reason
  * :-1: Because of that, NcRichContenteditable renders mention template as expected, but without applied styles
  * :+1: PR extracts NcMentionBubble styles to a mixin, so it could be included in every needed component to its scoped styles like this:
```css
@import '../../assets/NcMentionBubble';
:deep(.mention-bubble) {
	@include mention-bubble;
}
```
Drawback: because NcMentionBubble is still imported by richeditor mixin, styles in the compiled JS file are duplicated, but it's about 7 style rules, so lesser evil :imp: 

Tested with Talk chat, Files comments, and should be applicable for other apps

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image
image | image

_For Files, name/label is not provided on server-side -> unrelated to this PR_

### 🏁 Checklist

### DIFF
--- a/src/assets/NcMentionBubble.scss
+++ b/src/assets/NcMentionBubble.scss
@@ -1,0 +1,69 @@
+$bubble-height: 20px;
+$bubble-max-width: 150px;
+$bubble-padding: 2px;
+$bubble-avatar-size: $bubble-height - 2 * $bubble-padding;
+
+@mixin mention-bubble {
+	.mention-bubble--primary .mention-bubble__content {
+		color: var(--color-primary-element-text);
+		background-color: var(--color-primary-element);
+	}
+
+	.mention-bubble__wrapper {
+		max-width: $bubble-max-width;
+		// Align with text
+		height: $bubble-height - $bubble-padding;
+		vertical-align: text-bottom;
+		display: inline-flex;
+		align-items: center;
+	}
+
+	// Should be in both places
+	.mention-bubble__content {
+		display: inline-flex;
+		overflow: hidden;
+		align-items: center;
+		max-width: 100%;
+		height: $bubble-height;
+		-webkit-user-select: none;
+		user-select: none;
+		padding-right: $bubble-padding * 3;
+		padding-left: $bubble-padding;
+		border-radius: math.div($bubble-height, 2);
+		background-color: var(--color-background-dark);
+	}
+
+	.mention-bubble__icon {
+		position: relative;
+		width: $bubble-avatar-size;
+		height: $bubble-avatar-size;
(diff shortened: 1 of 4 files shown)


susnux:
This is a not that nice work around, the real but seems to be the way we use it + a bug in vite. As vite removes the chunk that imports the css :/ 
I am currently digging into vite source

Antreesy:
> I am currently digging into vite source 

https://vitejs.dev/config/build-options.html#build-csscodesplit

Could it be related?

susnux:
No the problem is vite removes chunks that only import css in the `generateBundle` hook, but we currently inject css only in the `renderChunk` hook which is called before.
I do not think it is a bug but a decision -> if you remove the chunks in `renderChunk` you would need to check every chunk twice.

The solution would be to fix the css plugin to also do the import in `generateBundle` but this might lead to duplicated code, but I will try.

Antreesy:
Superseded by https://github.com/nextcloud-libraries/nextcloud-vite-config/pull/112

PR #5142, opened 2024-01-25
fix(NcPopover): a11y attrs provide-inject reactivity

ShGKme:
### ☑️ Resolves

Though it must be fine in Vue 3.3+, for some reason, sometimes `computed` in `provide/inject` is not reactive.

It works in `docs`, in `server`, in `Talk`. But it doesn't work in `Text`...

We had a similar issue with provide/inject in `NcSidebarTabs` in past.

@raimund-schluessler I also included a small fix with `internalShown` based on your comment somewhere

### 🏁 Checklist

### DIFF
--- a/src/components/NcButton/NcButton.vue
+++ b/src/components/NcButton/NcButton.vue
@@ -440,9 +440,9 @@ export default {
 	name: 'NcButton',
 
 	inject: {
-		ncPopoverTriggerAttrs: {
+		getNcPopoverTriggerAttrs: {
 			from: 'NcPopover:trigger:attrs',
-			default: () => ({}),
+			default: () => () => ({}),
 		},
 	},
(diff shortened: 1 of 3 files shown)


ShGKme:
Maybe the issue is because Viewer is using not the latest Vue patch

Pytal (approved):

susnux (approved):
works but not that nice @raimund-schluessler probably do not port to vue3(?)

ShGKme:
> works but not that nice @raimund-schluessler probably do not port to vue3(?)

Getter function will work in Vue 3 too. Or what do you mean?

susnux:
> Getter function will work in Vue 3 too. Or what do you mean?

It also works without in Vue3

PR #5144, opened 2024-01-26
feat: Implement widget flag for requesting interactive view

### LINKED ISSUE #5090: Add interactive mode for widget registration and rendering
To cover https://github.com/nextcloud/server/issues/42550 we need to extend the link preview components and API methods to:

- [x] Allow registerWidget to let apps render a different view depending if an interactive widget was requested or not
- [x] Allow the NcReferenceList to indicate if the shown widgets should be interactive or not 
- [x] Figure out a generic way to only render widgets in the visible area
- [x] Check if we need a generic way to "enable" interactive mode on the fly to fix unexpected interactivity e.g. when scrolling through a talk room

juliushaertl:
### ☑️ Resolves

* Fix #5090 

Allow apps that display link previews to pass in if they want to request interactive link previews or not. We default to false to not have any breaking behaviour for apps. When apps register a preview render function with `registerWidget` they can then handle either the non-interactive or interactive view based on that flag. Apps not handling this will just fallback to the existing preview rendering that is currently in place already.

### Render interactive widgets as link preview

Just pass in the interactive property to the NcReferenceList component

```
<NcReferenceList :text="'https://nextcloud.com/f/1234'" :interactive="true" />
```

### Render interactive widgets as reference provider

```
registerWidget('deck-board', (el, { richObjectType, richObject, accessible, interactive }) => {
	if (interactive) {
		// render interactive component to el
		return
	}
	// existing custom render functionality
})

```

### 🏁 Checklist

### DIFF
--- a/src/components/NcRichText/NcReferenceList.vue
+++ b/src/components/NcRichText/NcReferenceList.vue
@@ -34,6 +34,10 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		interactive: {
+			type: Boolean,
+			default: false,
+		},
 	},
 	data() {
 		return {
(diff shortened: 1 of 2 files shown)


julien-nc (approved):
LGTM

susnux (approved):

PR #5146, opened 2024-01-26
feat(`NcListItem`): introduce 'details' slot

Antreesy:
### ☑️ Resolves

* Provide an option to use slot (for embedding icon / custom template) instead of prop string

### 🖼️ Screenshots

Screenshot from 2024-01-26 12-31-55

### 🚧 Tasks

- [ ] Design review
- [ ] Code review

### 🏁 Checklist

### DIFF
--- a/src/components/NcListItem/NcListItem.vue
+++ b/src/components/NcListItem/NcListItem.vue
@@ -87,13 +87,15 @@
 			:name="'This is an active element with normal counter'"
 			:bold="false"
 			:active="true"
-			:details="'1h'"
 			:counter-number="44">
 			<template #icon>
 				<NcAvatar :size="44" user="janedoe" display-name="Jane Doe" />
 			</template>
 			<template #subname>
 				In this slot you can put both text and other components such as icons
+			</template>
+			<template #details>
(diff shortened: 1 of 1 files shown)


marcoambrosini (approved):
Nice :)

src/components/NcListItem/NcListItem.vue
```diff
@@ -87,14 +87,16 @@
...
-			:details="'1h'"
 			:counter-number="44">
 			<template #icon>
 				<NcAvatar :size="44" user="janedoe" display-name="Jane Doe" />
 			</template>
 			<template #subname>
 				In this slot you can put both text and other components such as icons
 			</template>
+			<template #details>
+				<LinkIcon :size="16"/>
```
susnux, line 98: should it not be 20 as that's our default?

susnux (approved):

src/components/NcListItem/NcListItem.vue
Antreesy: Same as bubble icon below, and close to the font size. But that's up to end developer anyway, I believe

PR #5148, opened 2024-01-26
fix: app-navigation-toggle remains on screen on small viewports

emoral435:
### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
firefox_lHUQ9YnsiM|firefox_mBytSzpcFp

### 🚧 Tasks

This has the same changes from nextcloud/photos#2285

- When the viewport got small in the photos app, the app-navigation-button overflowed to the right of the screen, becoming inaccessible
- To fix this, since the button is positioned relative to the header, I removed some of the negative right positioning on the button

This allows the button to still be in view, while a little tight, and also still allows it to be found with keyboard navigation 👍
Any styling choices can be commented on !

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppNavigationToggle/NcAppNavigationToggle.vue
+++ b/src/components/NcAppNavigationToggle/NcAppNavigationToggle.vue
@@ -55,6 +55,11 @@ export default {
 	},
 
 	props: {
+		/**
+		 * Tracks whether the toggle has been clicked or not.
+		 * If it has been clicked, switches between the different MenuIcons
+		 * and emits a boolean indicating its opened status
+		 */
 		open: {
 			type: Boolean,
 			required: true,
@@ -69,6 +74,10 @@ export default {
 		},
 	},
 	methods: {
+		/**
+		 * Once the toggle has been clicked, emits the toggle status
+		 * so parent components can gauge the status of the navigation button
+		 */
 		toggleNavigation() {
 			this.$emit('update:open', !this.open)
 		},


emoral435:
Update: added some comments as well for the docs 👍

emoral435:
Gotcha. Will close this PR then!

emoral435:
Agreed ! Closed the PR, but decided to reopen it for the docs, but not the most important PR ever so this can always wait!

src/components/NcAppNavigationToggle/NcAppNavigationToggle.vue
```diff
@@ -45,6 +45,11 @@ import { t } from '../../l10n.js'
 import MenuIcon from 'vue-material-design-icons/Menu.vue'
 import MenuOpenIcon from 'vue-material-design-icons/MenuOpen.vue'
 
+/**
+ * The NcAppNavigationToggle component is used typically alongside another header navigation component.
+ * It emits the status of whether the button has been clicked or not.
+ *
+ */
```
ShGKme, line 52: It is supposed to be used only in `<NcAppNavigation>` and is not exported from the library.

We should probably remove it from the docs instead.

ShGKme (commented):
This component is supposed to be used only in `<NcAppNavigation>` and actually is not exported from the library.

So we should remove it from the docs at all instead. (in the styleguidist config)

susnux:
But I would keep the doc for internal purpose to understand the component itself.
We do not need to publish in styleguide, but keep for us.(?)

ShGKme:
> but keep for us.(?)

Not sure so detailed documentation is needed internally for so small a component and props like `open` for a toggle button.

susnux (approved):

JuliaKirschenheuter (approved):

PR #5149, opened 2024-01-26
fix: button size regression on small viewports fixed

emoral435:
### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image|firefox_I6ssjqt0MZ

### 🚧 Tasks
Regression for the aforementioned PR

- On small viewports, the button nested inside the wrapper for NcHeaderMenu stayed 50x50 px. This change ensures that the trigger button is the same size as the wrapper component on small viewports, which is 44x50 px. This now keeps the styling on the server consistent and working!

### 🏁 Checklist

### DIFF
--- a/src/components/NcHeaderMenu/NcHeaderMenu.vue
+++ b/src/components/NcHeaderMenu/NcHeaderMenu.vue
@@ -348,7 +348,7 @@ $externalMargin: 8px;
 	height: var(--header-height);
 
 	#{&}__trigger {
-		width: var(--header-height) !important;
+		width: 100% !important;
 		height: var(--header-height);
 		opacity: .85;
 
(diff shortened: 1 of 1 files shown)


Pytal (approved):
Tested and works!

src/components/NcHeaderMenu/NcHeaderMenu.vue
```diff
@@ -413,8 +413,8 @@ $externalMargin: 8px;
 	.header-menu {
 		width: $clickable-area;
 
-		&__trigger {
-			width: $clickable-area;
+		#{&}__trigger {
+			width: 100% !important;
```
ShGKme, line 417: Seems that this is 100% on any screen, so we can remove this from `@media` and set it above instead of `width: var(--header-height) !important;`

(haven't tested)

ShGKme: Motivation: to have less source or truth

emoral435: Tested this out on my local dev - it works! Sorry for getting back so late, I discarded this notification by accident 😅

firefox_thFKoqAqEa

susnux (approved):

PR #5150, opened 2024-01-26
fix(NcCheckboxRadioSwitch): Make sure to include current value in cheched array

susnux:
### ☑️ Resolves

The checkbox sets are currently broken as the click even does not set the checked state anymore thus the wrong value is emitted (tab + space works).

### 🏁 Checklist

### DIFF
--- a/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
+++ b/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
@@ -593,7 +593,7 @@ export default {
 
 			// If the initial value was a boolean, let's keep it that way
 			if (typeof this.checked === 'boolean') {
-				this.$emit('update:checked', !this.isChecked)
+				this.$emit('update:checked', !this.checked)
 				return
 			}
 
@@ -601,7 +601,12 @@ export default {
 			const values = this.getInputsSet()
 				.filter(input => input.checked)
 				.map(input => input.value)
-			this.$emit('update:checked', values)
+
+			if (values.includes(this.value)) {
+				this.$emit('update:checked', values.filter((v) => v !== this.value))
+			} else {
+				this.$emit('update:checked', [...values, this.value])
+			}
 		},
 
 		/**


src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
```diff
@@ -593,15 +593,20 @@ export default {
 
 			// If the initial value was a boolean, let's keep it that way
 			if (typeof this.checked === 'boolean') {
-				this.$emit('update:checked', !this.isChecked)
+				this.$emit('update:checked', !this.checked)
```
Antreesy, line 596: Shouldn't be done the same for TYPE_SWITCH?

Antreesy (approved):
Changes make sense, will give it a test a bit later

src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
ShGKme: Switch is supposed to have only `boolean` value, so should work

ShGKme (approved):
Explanation of the issue's root:
- Interesting regression of https://github.com/nextcloud-libraries/nextcloud-vue/pull/4999/files
- Before, `onToggle` listener was only on the `<input>` and a click triggered input via `<label>` (works for both **mouse** and **keyboard**)
- After, `onToggle` was on both `<input>` (for **keyboard**) and `<span>` (for **mouse**)
- For arrays, input values age get not from Vue instance state, but from DOM via `document.querySelector ` in `getInputsSet`
- What happens on mouse click - only `<span>` was clicked, the native `<input>` has not been changed, thus `getInputsSet` returns old values

Explanation of the fix:
- As native `<input>` is not changing natively, now we need to programmatically calculate new values

susnux:
> As native `<input>` is not changing natively, now we need to programmatically calculate new values

BTW I would prefer a native `label` but the firefox but...

ShGKme:
Sorry, I was tied and missed that this is actually exactly the same explanation that @susnux already made in the PR description 🙈

PR #5153, opened 2024-01-26
fix(NcActions): keyboard navigation

ShGKme:
### ☑️ Resolves

* Fix - see comment https://github.com/nextcloud-libraries/nextcloud-vue/pull/4953#pullrequestreview-1845179797

Fixes keyboard navigation in two cases:
1. When a menu type is determined incorrect because **inline actions** were taken into account
   - E.g. where is an inline button and we thought, it is in the menu
2. Correctly handle <kbd>Tab</kbd> blur, when there is no focusable element in the actions

There are quite a lot of changes.

Separated with 💙 into small commits to be easier to review.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
before-tab | after-tab

### 🏁 Checklist

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -698,8 +698,8 @@ export default {
 			</NcActions>
 		</p>
 
-		<h2>Popover</h2>
-		Has any elements, including text input element, or no buttons.
+		<h2>Dialog</h2>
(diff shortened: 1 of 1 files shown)


Pytal (approved):
Clean commits!

susnux (approved):

PR #5162, opened 2024-01-28
fix(NcSelect): list width on page scaling

### LINKED ISSUE #4943: [NcSelect] Minor presentation issue
When opening NcSelect, there is a small gap between the input and the suggestion box.
grafik

When zooming the page (here 170% with Firefox 121) , the suggestion box gets zoomed different from the other elements
grafik

ShGKme:
By default Vue-Select takes width for content list on element insert.
But if select is resized after opening (for example, on page scale), that width is not actual anymore.
Manually set width according to the toggle element on update.

The current solution is a bit glitchy when select has flexible width. Not sure, how to fix better.

### ☑️ Resolves

* Fix #4943

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
select-before | select-after
select-before-2 | select-after-2

### 🏁 Checklist

### DIFF
--- a/src/components/NcSelect/NcSelect.vue
+++ b/src/components/NcSelect/NcSelect.vue
@@ -968,6 +968,7 @@ export default {
 						Object.assign(dropdownMenu.style, {
 							left: `${x}px`,
 							top: `${y}px`,
+							width: `${component.$refs.toggle.getBoundingClientRect().width}px`,
 						})
 					})
 				}


raimund-schluessler (approved):
Not ideal, but works a lot better than before already 👍

susnux (approved):

PR #5167, opened 2024-01-29
[next] fix(docs): bring back NcAppSidebar in docs

raimund-schluessler:
### ☑️ Resolves

* This brings back the `NcAppSidebar` example in the docs and adjusts the style to actually show `NcContent`.
* It also fixes an issue in the styleguide with `Teleport` as the `#skip-actions` element was not present.

### DIFF
--- a/src/components/NcContent/NcContent.vue
+++ b/src/components/NcContent/NcContent.vue
@@ -34,6 +34,7 @@ It also will set the skip content buttons needed for accessibility.
 
 ```vue
 	<template>
+		<div id="skip-actions" />
 		<NcContent app-name="forms">
 			<NcAppNavigation>
 				<template #list>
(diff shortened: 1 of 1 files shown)


raimund-schluessler:
> > Adjust the styleguide to have a `<div id="#skip-actions">` element somewhere (preferred, but I don't know how).
> 
> I have not looked into it, but maybe just add this to the examples that use `NcContent`? Otherwise maybe (dirty!) inject it using `global.requires.js`.

That is a greate idea, thanks! I added an element and it makes the styleguide work. The skip actions is even functional. 👍

susnux (approved):
works but  styles are broken (like before)

raimund-schluessler:
Could we merge this and release `9.0.0-alpha.1`? Then I could proceed and merge the Tasks vue 3 migration https://github.com/nextcloud/tasks/pull/1971 as well.

ShGKme (approved):

PR #5176, opened 2024-01-29
fix(NcRichText): Make URL_PATTERN match localhost and URLs with ports

mejo-:
E.g. the file reference widget should work with links to Nextcloud on localhost.

Also add some basic unit tests to prevent regressions with the regexes in the future.

### 🏁 Checklist


ShGKme (approved):

src/components/NcRichText/helpers.js
```diff
@@ -3,14 +3,12 @@
  *
  * @type {RegExp}
  */
-export const URL_PATTERN = /(\s|^)(https?:\/\/)((?:[-A-Z0-9+_]+\.)+[-A-Z]+(?:\/[-A-Z0-9+&@#%?=~_|!:,.;()]*)*)(\s|$)/ig
+export const URL_PATTERN = /(\s|^)(https?:\/\/)([-A-Z0-9+_.]+(?::[0-9]+)?(?:\/[-A-Z0-9+&@#%?=~_|!:,.;()]*)*)(\s|$)/ig
+export const RN_AUTOLINK = /(\s|^)(https?:\/\/)([-A-Z0-9+_.]+(?::[0-9]+)?(?:\/[-A-Z0-9+&@#%?=~_|!:,.;()]*)*)(?=\s|\)|$)/ig
```
max-nextcloud, line 7: I don't understand what this is for. Seems very similar to the `URL_PATTERN` above. What does RN stand for?

max-nextcloud (approved):
Looks good to me. Thanks for adding the tests.

Maybe one thing to clarify (see below).

src/components/NcRichText/helpers.js
max-nextcloud: oh... and this is also not being tested. You might want to add it to the test to prevent regressions here as well.

mejo-: Oh, good catch. This was a leftover - removed now.

PR #5178, opened 2024-01-30
fix(NcActions): hotfix for custom children

ShGKme:
### ☑️ Resolves

* Hotfix for https://github.com/nextcloud-libraries/nextcloud-vue/issues/5171
* Particularly a regression of https://github.com/nextcloud-libraries/nextcloud-vue/pull/5153

In `NcActions` we check what `NcAction*` components were passed as direct children. But some apps (Text, Mail) pass `NcAction*` not as direct children of `NcActions`, but as a sub-child.

```html
<NcActions>
  <MyUserActionsList /> // renders <NcActionButton />
</NcActions>
```

The current `NcActions` implementation doesn't fully support this.
- In past it resulted in ignoring `inline` prop and incorrect render of single actions
- After mentioned PR it was completely broken:
  1. Menu is rendered
  2. There are no direct `NcAction*` interactive children (they are sub-children)
  3. Menu is considered a tooltip
  4. It should close on blur
  5. After opening, it focuses on an interactive element, causing blur and closing.

Unfortunately, in Vue 2 it is not possible to easily fix. When render function is executed, children are not rendered yet, so we don't know if they are to render `NcAction*` components or not.

This PR adds a hotfix that prevents component from being unusable. If children are not allowed custom components - consider menu a dialog with the default focus trap.

So the component works like it worked in the past. Yet it still has bad accessibility in this case.

#### Docs

image

🏚️ Before | 🏡 After
---|---
nc-actions-children-before | nc-actions-children-after

### 🚧 Tasks

- [x] Document `NCActions` children limitations
- [x] Add a hotfix - make NcActions be considered a semantic `dialog` if there are custom 

### 🏁 Checklist

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -1188,7 +1188,7 @@ export default {
 		 *
 		 * We need to pause all the focus traps for opening popover and then unpause them back after closing.
 		 */
-		 intersectIntoCurrentFocusTrapStack() {
+		intersectIntoCurrentFocusTrapStack() {
 			if (this.withFocusTrap) {
 				return
 			}
@@ -1415,7 +1415,11 @@ export default {
 			// When there is no focusable elements to handle Tab press from actions menu
 			// It requries manual closing
 			if (this.actionsMenuSemanticType === 'tooltip') {
-				this.closeMenu(false)
+				// Tooltip is supposed to have no focusable element.
+				// However, if there is a custom focusable element, it will be auto-focused and cause the menu to be closed on open.
+				if (this.$refs.menu && this.$refs.menu.querySelectorAll(focusableSelector).length === 0) {
+					this.closeMenu(false)
+				}
 			}
 		},
 		onClick(event) {
@@ -1491,7 +1495,22 @@ export default {
 		} else if (hasLinkAction) {
 			this.actionsMenuSemanticType = 'navigation'
 		} else {
-			this.actionsMenuSemanticType = 'tooltip'
+			// (!) Hotfix (!)
+			// In Vue 2 it is not easy to search for NcAction* in sub-component of a slot.
+			// When a menu is rendered, children are not mounted yet.
+			// If we have NcActions > MyActionsList > NcActionButton, only MyActionsList's vnode is available.
+			// So when NcActions has actions as non-direct children, here then we don't know about them.
+			// Like this menu has no buttons/links/inputs.
+			// It makes the menu incorrectly considered a tooltip.
+			const ncActions = actions.filter((action) => this.getActionName(action).startsWith('NcAction'))
+			if (ncActions.length === actions.length) {
+				// True tooltip
+				this.actionsMenuSemanticType = 'tooltip'
+			} else {
+				// Custom components are passed to the NcActions
+				// dialog is a universal fallback option
+				this.actionsMenuSemanticType = 'dialog'
+			}
 		}
 
 		const actionsRoleToHtmlPopupRole = {
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -807,6 +807,131 @@ p {
 }
 </style>
 ```
+
+## NcActions children limitations
+
+`<NcActions>` is supposed to be used with direct `<NcAction*>` children.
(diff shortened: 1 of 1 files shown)


src/components/NcActions/NcActions.vue
```diff
@@ -807,6 +807,119 @@ p {
 }
 </style>
 ```
+
+## NcActions children limitations
+
+`<NcActions>` is supposed to be used with direct `<NcAction*>` children.
+Thought, it will work when actions are not direct children but are wrapped into custom components, it has limitations:
```
raimund-schluessler, line 814:
```suggestion
Although it works when actions are not direct children but wrapped in custom components, it has limitations:
```

src/components/NcActions/NcActions.vue
```diff
@@ -1366,7 +1483,22 @@ export default {
 		} else if (hasLinkAction) {
 			this.actionsMenuSemanticType = 'navigation'
 		} else {
-			this.actionsMenuSemanticType = 'tooltip'
+			// (!) Hotfix (!)
+			// In Vue 2 it is not easy to search for NcAction* in sub-component of a slot.
+			// When a menu is rendered, children are not mounted yet.
+			// If we have NaActions > MyActionsList > NcActionButton, only MyActionsList's vnode is available.
+			// So when NcActions has actions as non-direct children, here the we don't know about them.
```
raimund-schluessler, line 1490:
```suggestion
			// If we have NcActions > MyActionsList > NcActionButton, only MyActionsList's vnode is available.
			// So when NcActions has actions as non-direct children, here then we don't know about them.
```

src/components/NcActions/NcActions.vue
```diff
@@ -1366,7 +1483,22 @@ export default {
...
+			// Like this menu has no buttons/links/inputs.
+			// It makes the menu incorrectly considered a tooltip.
+			const ncActions = actions.filter((action) => this.getActionName(action).startsWith('NcAction'))
+			if (ncActions.length === actions.length) {
+				// True tooltip
+				this.actionsMenuSemanticType = 'tooltip'
+			} else {
+				// Custom components are passed to the NcActions
+				// dialog is an universal fallback option
```
raimund-schluessler, line 1499:
```suggestion
				// dialog is a universal fallback option
```

src/components/NcActions/NcActions.vue
```diff
@@ -807,6 +807,119 @@ p {
...
+		</tr>
+		<tr>
+			<td>
+				<NcActions>
+					<MyUserActionButton />
+				</NcActions>
+			</td>
+			<td>
+				<NcActions>
+					<NcActionButton icon="icon-user-white">Button</NcActionButton>
```
raimund-schluessler, line 838: I know this is just an example for the docs, but we don't want to use the `icon` prop anymore 🙈

raimund-schluessler (approved):
These custom wrappers for `NcAction*` are a pain. I know they are necessary sometimes, but they really interfere with how `NcActions` works. And it got worse with the focus trap and a11y stuff.

However, the hotfix brings back the functionality for now, so 👍 

I just had a few nitpicks for the language 🙂

src/components/NcActions/NcActions.vue
ShGKme: Yeah, but I wanted an example to be smaller. We haven't marked them as deprecated though.

Antreesy: is `icon-user-white` a good idea to use? it's not visible with a light theme
image

Antreesy (approved):
Tested with styleguide only

susnux (approved):
Please fix DCO and the `fixup` commments

ShGKme:
> Please fix DCO and the `fixup` commments

`fixup` commits are not supposed to be merged without squash

src/components/NcActions/NcActions.vue
ShGKme: `--icon-use-[white/dark]` is defined on the user-preferred scheme. They require `data-theme` attrs on styleguidist page.

Moving to out preferred SVG in `#icon` slot then.

Template becomes huge, added `v-for`.

image

susnux:
> fixup commits are not supposed to be merged without squash

Thats what I meant: Squash the commits and add the sign off^^

PR #5181, opened 2024-01-30
chore: Update workflows

susnux:
* Drop `command-rebase` this is now native on Github
* Update `lint-eslint` from templates
* Update `lint-stylelint` from templates
* Replace `npm-test` with `node-test` from the templates
* Drop `node` because the build is already included in `node-test` so that will fail if build is not possible
* Update `npm-publish` 
  * Also adjust to automatically set correct tag for beta and alpha releases


mejo- (approved):

.github/workflows/npm-publish.yml
```diff
@@ -11,39 +11,51 @@ on:
...
   contents: read
+  packages: write
 
 jobs:
   publish:
     runs-on: ubuntu-latest
 
     name: Build and publish to npm
+    env:
+      RELEAS_GROUP: ${{ (contains(github.ref, 'beta') || contains(github.ref, 'alpha')) && 'next' || 'latest' }}
```
Antreesy, line 22: Typo?

mejo-:
Also it seems like the branch merge requirements need to be adjusted. It's waiting for required job "Node" which gets removed with this PR.

skjnldsv (approved):

susnux:
> branch merge requirements need to be adjusted#

Yes but waiting for this to be approved :wink:

.github/workflows/npm-publish.yml
susnux: Good catch!

PR #5184, opened 2024-01-30

raimund-schluessler:
### ☑️ Resolves

@ShGKme I didn't get #5178 to work properly with vue 3 yet. This needs some more work.


src/components/NcActions/NcActions.vue
```diff
@@ -807,6 +807,133 @@ p {
...
+			name: 'NcActionButtonList',
+			components: { Account },
+			render: () => h(Fragment, {}, [
+				h(resolveComponent('NcActionButton'), {}, { default: () => 'Button', icon: () => h(Account, { props: { size: 20 } })}),
+				h(resolveComponent('NcActionButton'), {}, { default: () => 'Button', icon: () => h(Account, { props: { size: 20 } })}),
+				h(resolveComponent('NcActionButton'), {}, { default: () => 'Button', icon: () => h(Account, { props: { size: 20 } })}),
+				h(resolveComponent('NcActionButton'), {}, { default: () => 'Button', icon: () => h(Account, { props: { size: 20 } })}),
+			]),
+		}
+	}
```
ShGKme, line 915: - Correctly pass `props`
- Bring back `DIV` instead of `Fragment` (to emulate more commonly used wrappers with HTML wrapper)
- Provide slot content via render-function, not vnodes 

```suggestion
import { h, resolveComponent } from 'vue'
import Account from 'vue-material-design-icons/Account.vue'

export default {
	components: {
		Account,

		MyUserActionButton: {
			name: 'MyUserActionButton',
			components: { Account },
			render: () => h(resolveComponent('NcActionButton'), null, { default: () => 'Button', icon: () => h('Account', { size: 20 }) }),
		},

		MyActionsList: {
			name: 'MyActionsList',
			components: { Account },
			render: () => h('div', null, {
				default: () => [
					h(resolveComponent('NcActionButton'), null, { default: () => 'Button', icon: () => h('Account', { size: 20 }) }),
					h(resolveComponent('NcActionButton'), null, { default: () => 'Button', icon: () => h('Account', { size: 20 }) }),
					h(resolveComponent('NcActionButton'), null, { default: () => 'Button', icon: () => h('Account', { size: 20 }) }),
					h(resolveComponent('NcActionButton'), null, { default: () => 'Button', icon: () => h('Account', { size: 20 }) }),
				],
			}),
		}
	}
```

ShGKme:
> @ShGKme I didn't get #5178 to work properly with vue 3 yet. This needs some more work.

The issue is not related to this PR. Such usage of `NcActions` didn't work before as well, because `findActions` cannot find actions from sub-components (they need to be rendered first). So we have the same issue that we always had in Vue 2 (but more noticeable).

So this PR only adds an example of this limitation to the docs.

We need to find a more generic solution to solve https://github.com/nextcloud-libraries/nextcloud-vue/issues/5171

src/components/NcActions/NcActions.vue
```diff
@@ -1091,7 +1218,7 @@ export default {
 		 * @return {boolean}
 		 */
 		isValidSingleAction(action) {
-			return ['NcActionButton', 'NcActionLink', 'NcActionRouter'].includes(this.getActionName(action))
+			return ['NcActionButton', 'NcActionLink', 'NcActionRouter'].some(valid => this.getActionName(action).startsWith(valid))
```
ShGKme, line 1221: Why change to `startsWith`?

raimund-schluessler: I thought we could introduce a naming convention for these wrappers. So if the components name starts with NcActions, we just pass it on. But that wasn't fully thought through and didn't work yet.

raimund-schluessler:
Yes, this didn't work at all with `next`. But I thought we should at least get the same behavior as for `master`. But due to `findActions`, wrapped actions do not show up at all.

src/components/NcActions/NcActions.vue
ShGKme: As an alternative, we could ask to add a custom option and check `$options`.

But this won't help in a general case. Developers may make wrappers for a list of actions

```html
<template>
  <div>
    <NcActionButton />
    <NcActionInput />
  </div>
</template>
```

Or dynamic actions. Like, I dunno,
```html
<template>
  <NcActionButton v-if="iLikeButtons" /> 
  <NcActionInput v-else-if="iDont" />
</template>
```

ShGKme:
> Yes, this didn't work at all with `next`. But I thought we should at least get the same behavior as for `master`. But due to `findActions`, wrapped actions do not show up at all.

I have found a small number of apps that require this feature. Text, Mail. I hope, we will find a solution until they migrate =D

src/components/NcActions/NcActions.vue
susnux: > Or dynamic actions. Like, I dunno

Have this in forms: https://github.com/nextcloud/forms/blob/e1cfbc737234e0bd3ff10f8a1222223a8093e254/src/views/Results.vue#L134

ShGKme: > Have this in forms:

This will work. I meant when those `<template>...</template>` are templates of new custom components.

```html
<template>
  <div>
    <NcActionButton />
    <NcActionInput />
  </div>
</template>

<script>
export default {
  name: 'MyAction'
}
</script>  
```

and then

```html
<template>
  <NcActions>
    <MyAction />
  </div>
</template>
```

raimund-schluessler:
Ok, I would propose that we do a simple merge of `master` into `next` here without trying to fix the underlying problem and then proceed to find a solution in a new PR. After all, it doesn't work in `next` anyway, so we don't make it worse.

ShGKme (approved):

raimund-schluessler:
Since `next` uses `vitest` to run the tests, we had to install `@vitest/coverage-v8`. I did this and also regenerated `package-lock.json` to get the latest dependencies.

ShGKme (approved):

susnux (approved):

PR #5187, opened 2024-01-30
fix(NcActions): adjust wrapper names

raimund-schluessler:
### ☑️ Resolves

If we adjust the action wrapper names to start with `NcAction` (or alternatively relax the "what is an action condition"), then we at least get the same (broken) behaviour for `next` as we get for `master`. Would that be an improvement?

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -830,7 +830,7 @@ Although it works when actions are not direct children but wrapped in custom com
 		<tr>
 			<td>
 				<NcActions>
-					<MyUserActionButton />
+					<NcActionUserButton />
 				</NcActions>
 			</td>
 			<td>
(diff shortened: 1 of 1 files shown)


ShGKme:
> If we adjust the action wrapper names to start with `NcAction` (or alternatively relax the "what is an action condition"), then we at least get the same (broken) behaviour for `next` as we get for `master`. Would that be an improvement?

I don't think we should provide such an interface for the component.

It doesn't solve the issue for wrappers over list of components or wrappers with dynamic content (see: https://github.com/nextcloud-libraries/nextcloud-vue/pull/5184#discussion_r1471863042).

It doesn't solve an issue with limitations anyway (e.g., it is important to separate `NcActionButton` from `NcActionButtonGroup` for both `inline` buttons and a11y).

A solution based on name is possible now already without prefix, component users can pass component with exact name `NcActionButton` and ect., when it is not a list and static.

While we don't have an app migrating to Vue 3 that suffers from this issue, I'd try to solve.

raimund-schluessler:
Ok. Fine with me.

PR #5198, opened 2024-01-31
fix(NcListItem): Remove actions from within `a` which caused invalid HTML output

susnux:
### ☑️ Resolves

* Fix https://github.com/nextcloud/server/issues/43237

It is not valid to put `button` inside `a` thus we need to refactor the NcListItem:
1. Unify the *actions* location within the source to be outside the main wrapper
2. Replace wrapper `a` with `div`
3. Make only the main content an `a` , this is required so users can right click + copy link or middle click -> native browser behavior.
4. Not related but fix *focus-visible* to have consistent style across all components.

*Hint: Hide white space changes when review*

### 🖼️ Screenshots

Could not spot visual changes except from focus visible.

https://github.com/nextcloud-libraries/nextcloud-vue/assets/1855448/de196e16-541e-4f5e-9517-23119fa4ceed

### 🏁 Checklist

### DIFF
--- a/src/components/NcListItem/NcListItem.vue
+++ b/src/components/NcListItem/NcListItem.vue
@@ -33,7 +33,7 @@
 			:counter-number="44"
 			counterType="highlighted">
 			<template #icon>
-				<NcAvatar :size="44" user="janedoe" display-name="Jane Doe" />
+				<NcAvatar disable-menu :size="44" user="janedoe" display-name="Jane Doe" />
 			</template>
 			<template #subname>
 				In this slot you can put both text and other components such as icons
@@ -62,7 +62,7 @@
 			:counter-number="44"
 			counterType="outlined">
 			<template #icon>
-				<NcAvatar :size="44" user="janedoe" display-name="Jane Doe" />
+				<NcAvatar disable-menu :size="44" user="janedoe" display-name="Jane Doe" />
 			</template>
 			<template #subname>
 				In this slot you can put both text and other components such as icons
@@ -89,7 +89,7 @@
 			:active="true"
 			:counter-number="44">
 			<template #icon>
-				<NcAvatar :size="44" user="janedoe" display-name="Jane Doe" />
+				<NcAvatar disable-menu :size="44" user="janedoe" display-name="Jane Doe" />
 			</template>
 			<template #subname>
 				In this slot you can put both text and other components such as icons
@@ -121,7 +121,7 @@
 			:counter-number="44"
 			counterType="highlighted">
 			<template #icon>
-				<NcAvatar :size="44" user="janedoe" display-name="Jane Doe" />
+				<NcAvatar disable-menu :size="44" user="janedoe" display-name="Jane Doe" />
 			</template>
 			<template #subname>
 				In this slot you can put both text and other components such as icons
@@ -146,7 +146,7 @@
 			:counter-number="44"
 			counterType="outlined">
 			<template #icon>
-				<NcAvatar :size="44" user="janedoe" display-name="Jane Doe" />
+				<NcAvatar disable-menu :size="44" user="janedoe" display-name="Jane Doe" />
 			</template>
 			<template #subname>
 				In this slot you can put both text and other components such as icons
@@ -170,7 +170,7 @@
 			:details="'1h'"
 			:counter-number="44">
 			<template #icon>
-				<NcAvatar :size="44" user="janedoe" display-name="Jane Doe" />
+				<NcAvatar disable-menu :size="44" user="janedoe" display-name="Jane Doe" />
 			</template>
 			<template #subname>
 				In this slot you can put both text and other components such as icons
@@ -191,7 +191,7 @@
 			:name="'Name of the element'"
 			:bold="false">
 			<template #icon>
-				<NcAvatar :size="44" user="janedoe" display-name="Jane Doe" />
+				<NcAvatar disable-menu :size="44" user="janedoe" display-name="Jane Doe" />
 			</template>
 			<template #subname>
 				In this slot you can put both text and other components such as icons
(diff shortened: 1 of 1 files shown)


src/components/NcListItem/NcListItem.vue
```diff
@@ -320,28 +331,30 @@
...
-					:class="{ 'list-item-content__wrapper--compact': compact }">
+				<div :class="['list-item-content__wrapper', {
+					'list-item-content__wrapper--compact': compact
+				}]">
 					<!-- @slot This slot is used for the NcAvatar or icon -->
 					<slot name="icon" />
 
 					<!-- Main content -->
-					<div class="list-item-content">
+					<a :id="anchorId || undefined"
```
Pytal, line 351: Child divs need to be switched to spans as `<a>` is inline

src/components/NcListItem/NcListItem.vue
```diff
@@ -320,28 +331,30 @@
...
+						tabindex="-1"
```
Pytal, line 352: No longer possible to activate the link?

src/components/NcListItem/NcListItem.vue
susnux: Are you sure? I read the specs that every content is allowed within `a` except for interactive content and such with `tabindex` specified.

src/components/NcListItem/NcListItem.vue
susnux: To keep current behavior of focusing the wrapper (`tabindex="0"`) but not introduce a second element that can be focused within.

src/components/NcListItem/NcListItem.vue
Pytal: Link to specs?

Pytal: If HTML validator passes then fine with me :)

src/components/NcListItem/NcListItem.vue
Pytal: Previously hitting enter on wrapper navigated to the link but now nothing happens

src/components/NcListItem/NcListItem.vue
susnux: Either: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a?retiredLocale=de#technical_summary
Or https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element

src/components/NcListItem/NcListItem.vue
ShGKme: I think, `<a>` should be focusable, but only `focus-within` should be set on the `NcListItem` wrapper. We probably need it anyway, because screen-reader navigation will navigate to the link itself.

Only elements with interactive `role` should have focus.

And if we have entire `NcListItem` focusable, we again have a focusable element inside a focusable element. Not sure if it is fine.

ShGKme: > hm, i think we need to make a real focus on the link but visual focus on a whole `NcListItem`, probably like: [nextcloud/server#41439 (comment)](https://github.com/nextcloud/server/pull/41439#issuecomment-1810201036)

Exactly the same solution as in Apps won't work here, because it will overflow the actions button.

But it could be the same as in files, there only file name is an actual link

susnux: > we again have a focusable element inside a focusable element

That is fine in some cases where you should be able to focus outer wrappers (e.g. in a grid) and then move the focus inside.

But I agree here we probably should not have the wrapper focused.

GretaD:
Hello @susnux please keep in mind that we are changing the design for mail. I will wait for this to be finished before a continue here: https://github.com/nextcloud-libraries/nextcloud-vue/pull/5209
But just so you know :)

susnux:
Ready to review I would say :rocket: 


https://github.com/nextcloud-libraries/nextcloud-vue/assets/1855448/e7c61df9-9200-4dfd-a746-d072b7e84357

src/components/NcListItem/NcListItem.vue
```diff
@@ -33,7 +33,7 @@
 			:counter-number="44"
 			counterType="highlighted">
 			<template #icon>
-				<NcAvatar :size="44" user="janedoe" display-name="Jane Doe" />
+				<NcAvatar disable-menu :size="44" user="janedoe" display-name="Jane Doe" />
```
ShGKme, line 36: Thanks!

JuliaKirschenheuter (approved):

ShGKme:
Compact is a little bit broken, the height is changed on hover.

compact

ShGKme (commented):
Works fine now with both Tab navigation and the screen reader quick navigation. The only issue I've found is the height on hover in compact mode

susnux:
> Compact is a little bit broken, the height is changed on hover.

@ShGKme I fixed this now:

https://github.com/nextcloud-libraries/nextcloud-vue/assets/1855448/3e75869c-f1e7-4f35-83fe-5cd7980e2952

emoral435 (approved):
LGTM!

src/components/NcListItem/NcListItem.vue
```diff
@@ -747,41 +732,54 @@ export default {
...
 .list-item {
 	box-sizing: border-box;
-	display: block;
+	display: flex;
 	position: relative;
 	flex: 0 0 auto;
 	justify-content: flex-start;
 	padding: 8px 10px;
+	// 4px padding for the focus-visible styles
+	margin: 4px;
```
Antreesy, line 741: That offset a list items content at 4px from the left border, cropping them
image

Maybe we should consider it like this, @susnux ?
```css
width: calc(100% - 8px);
```

PR #5202, opened 2024-02-01
chore: Make CHANGELOG.md content consistent styling

susnux:
Use consistent styling for the changelog, this makes the changelog pure Markdown without need for github mentions, just like most of its content already was.

Meaning
> fix xyz by @susnux in http://...

is now consistently
> fix xyz [\#000](http://example.com) ([susnux](https://github.com/susnux))"

---
regex: `by @([^ ]+) in ((https://github.com/)nextcloud-libraries/nextcloud-vue/pull/(\d+))`
replacement: `[\#$4]($2) \([$1]($3$1)\)`

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,7 +164,7 @@ index 0e3a6a705d..416b8b0fb9 100644
 - Generate the changelog content from the [release](https://github.com/nextcloud-libraries/nextcloud-vue/releases) page.
   Create a draft release, select the previous tag, click `generate` then paste the content to the `CHANGELOG.md` file
   1. adjust the links to the merged pull requests and authors so that the changelog also works outside of GitHub
-     by running `npm format:fix`.
+     by running `npm run prerelease:format-changelog`.
      This will apply this regex: `by @([^ ]+) in ((https://github.com/)nextcloud-libraries/nextcloud-vue/pull/(\d+))`
      Which this as the replacement: `[\#$4]($2) \([$1]($3$1)\)`
   2. use the the version as tag AND title (e.g `v4.0.1`)
--- a/build/format-changelog.mjs
+++ b/build/format-changelog.mjs
@@ -1,0 +1,24 @@
+/**
+ * SPDX-FileCopyrightText: Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+import { readFile, writeFile } from 'node:fs/promises'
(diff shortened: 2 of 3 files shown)


raimund-schluessler:
@susnux I think the changelog gets generated by an action. Could be that this would need an adjustment as well then.

susnux:
> I think the changelog gets generated by an action. Could be that this would need an adjustment as well then.

Currently people copy paste the content of the GitHub generate changelog output on the release page, the generate-changelog-action was outputting the styling correctly.

susnux:
Open for one year now. Your options:
1. keep -> approve :+1: 
2.  close -> react with a :-1:  on this

Antreesy:
> regex: ...
replacement: ...

I would also add it to README.md in https://github.com/nextcloud-libraries/nextcloud-vue#-releasing-a-new-version section, so we don't have to search for it 🙏🏽

susnux:
@Antreesy done

Antreesy (approved):

README.md
```diff
@@ -163,8 +163,11 @@ index 0e3a6a705d..416b8b0fb9 100644
   This will return a new version name, make sure it matches what you expect
 - Generate the changelog content from the [release](https://github.com/nextcloud-libraries/nextcloud-vue/releases) page.
   Create a draft release, select the previous tag, click `generate` then paste the content to the `CHANGELOG.md` file
-  1. use the the version as tag AND title (e.g `v4.0.1`)
-  2. add the changelog content as description (https://github.com/nextcloud-libraries/nextcloud-vue/releases)
+  1. adjust the links to the merged pull requests and authors so that the changelog also works outside of GitHub:
+     Apply this regex: `by @([^ ]+) in ((https://github.com/)nextcloud-libraries/nextcloud-vue/pull/(\d+))`
+     Which this as the replacement: `[\#$4]($2) \([$1]($3$1)\)`
```
ShGKme, line 168: I feel like an additional manual step will increase inconsistency instead...

Is it a problem is username is not a link? The PR is a link, where author is als available.

susnux: > Is it a problem is username is not a link? The PR is a link, where author is als available.

Yes author is not a valid link and also its pretty ugly when seen outside of GitHub (where you probably do not read it because there is the releases tab).
But you are right, its unnecessary work - so I created a script for it.

ShGKme (approved):

PR #5203, opened 2024-02-01
fix(NcActions): fix LI roles

ShGKme:
### ☑️ Resolves

Currently the menu in `NcActions` has the following structure:

```html
<ul role="POPUP_ROLE">
  <li role="LI_ROLE" />
</ul>
```

`POPUP_ROLE` can be - `menu | dialog | tooltip` or nothing (=`list`) in navigation:
- When it is a navigation, native roles works fine
- When it is a `menu`, `<li>` has role `presentation`, because there are `menuitem`-s inside.
- When it is `dialog` or `tooltip`, `<li>` had native `listitem` role because there is a list in the popup.

The problem is that in the last case `<UL>` is at the same time:
- A list items' container - should have `list` role
- A popup's container - should have `dialog` or `tooltip` role.

But it is not possible to have 2 roles. So, when it is a `dialog` or `tooltip`, list items are not in a **list**.

A proper solution should have been to wrap `UL` into another element.

```html
<div role="POPUP_ROLE">
  <ul>
    <li />
  </ul>
</div>
```

However, we decided that this is too risky change of the structure for apps and testing.

Temporal solution - remove `listitem` role in `dialogs` and `tooltips`.

### 🖼️ Screenshots

Example | 🏚️ Before | 🏡 After
---|---|---
image | image | image

### 🚧 Tasks

- [x] Better store HTML roles for different types of menu
- [x] Provide `NcActions` context with as menu type and LI role
  - Relaces existing `isSemanticMenu` to provide more important data
- [x] Add a composable to use this context
  - Replaces existing `inject`
- [x] Set list item role on all action items, except `NcActionSeparator` (it is always `separator`)

### P.S.

I haven't used existing **mixins** (`actionText`, `actionGlobal`) because they are not used on all the action items.

I haven't created a new **mixin**, because I decided, that we should move to **composables**.

### 🏁 Checklist

### DIFF
--- a/src/components/NcActionButton/NcActionButton.vue
+++ b/src/components/NcActionButton/NcActionButton.vue
@@ -331,7 +331,7 @@ export default {
 </docs>
 
 <template>
-	<li class="action" :class="{ 'action--disabled': disabled }" :role="isInSemanticMenu && 'presentation'">
+	<li class="action" :class="{ 'action--disabled': disabled }" :role="liRole">
 		<button :aria-label="ariaLabel"
 			:class="['action-button button-vue', {
 				'action-button--active': isChecked,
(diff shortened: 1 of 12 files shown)


ShGKme:
I forgot about this PR when was re-making it again in https://github.com/nextcloud-libraries/nextcloud-vue/pull/5381 🙈

PR #5209, opened 2024-02-02
feat(NcAppContent): Support full width lists

GretaD:
### ☑️ Resolves

* Ref https://github.com/nextcloud/mail/issues/8811

Having the possibility to switch to a full-width message/thread list. The three column design (nav + list + content) becomes two columns: nav + list or nav + content. The user will see the list. When they navigate to a message/thread, the content shows.

Adding a prop layoutMode to switch between views on NcAppContent and changing a bit the NcListItem, instead of putting the details inside the "line one" / "line two" we should split main content and details.

This allows to switch from flex-direction column to row for one-line layout without adjusting the DOM. This way we have the possibility to show the list in one line on mail, without breaking the other apps. 

When we are on list mode(oneLine) and content view is shown(aka thread on mail), we need to have both button, the back button and the navigation button. So, we had to change a bit how to show that on content view. We will need a better solution for those 2 buttons in the future, but for now it works apart from some edge cases.

### 🖼️ Screenshots
After
Screenshot from 2024-03-07 08-56-35

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppContent/NcAppContent.vue
+++ b/src/components/NcAppContent/NcAppContent.vue
@@ -78,16 +78,18 @@ The list size must be between the min and the max width value.
 
 		<template v-if="hasList">
 			<!-- Mobile view does not allow resizeable panes -->
-			<div v-if="isMobile"
-				:class="showDetails ? 'app-content-wrapper--show-details' : 'app-content-wrapper--show-list'"
-				class="app-content-wrapper app-content-wrapper--mobile">
-				<NcAppDetailsToggle v-if="hasList && showDetails" @click.native.stop.prevent="hideDetails" />
+			<div v-if="isMobile || layout === 'no-split'"
+				class="app-content-wrapper app-content-wrapper--no-split"
+				:class="{
+					'app-content-wrapper--show-details': showDetails,
+					'app-content-wrapper--show-list': !showDetails,
+					'app-content-wrapper--mobile': isMobile,}">
+				<NcAppDetailsToggle v-if="showDetails" @click.native.stop.prevent="hideDetails" />
+				<slot v-if="!showDetails" name="list" />
 
-				<slot name="list" />
-				<slot />
+				<slot v-else />
 			</div>
-
-			<div v-else class="app-content-wrapper">
+			<div v-else-if="layout === 'vertical-split'" class="app-content-wrapper">
 				<Splitpanes class="default-theme"
 					@resized="handlePaneResize">
 					<Pane class="splitpanes__pane-list"
@@ -108,9 +110,8 @@ The list size must be between the min and the max width value.
 				</Splitpanes>
 			</div>
 		</template>
-
 		<!-- @slot Provide the main content to the app content -->
-		<slot v-else />
+		<slot v-if="!hasList" />
 	</main>
 </template>
 
@@ -119,11 +120,11 @@ import NcAppDetailsToggle from './NcAppDetailsToggle.vue'
 import { useIsMobile } from '../../composables/useIsMobile/index.js'
 
 import { getBuilder } from '@nextcloud/browser-storage'
-import { emit } from '@nextcloud/event-bus'
 import { useSwipe } from '@vueuse/core'
 import { Splitpanes, Pane } from 'splitpanes'
 
 import 'splitpanes/dist/splitpanes.css'
+import { emit } from '@nextcloud/event-bus'
 
 const browserStorage = getBuilder('nextcloud').persist().build()
 
@@ -139,7 +140,6 @@ export default {
 		Pane,
 		Splitpanes,
 	},
-
 	props: {
 		/**
 		 * Allows to disable the control by swipe of the app navigation open state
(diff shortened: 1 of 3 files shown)


GretaD:
A comment from Ferdinand:

Currently we have:
line one
`   content | details
`
line two
`   content | details
`
but it would be easier to have
`content | details
`
with content:
`content one
`
`content two
`
And same with the details..

susnux:
> A comment from Ferdinand

Yes instead of putting the details inside the "line one" / "line two" we should split main content and details.
This allows to switch from `flex-direction` `collumn` to `row` for *one-line layout* without adjusting the DOM.

We only need to change the styles then.

st3iny (commented):
I tested this for regressions on Spreed and Calendar and it worked fine.

vite.config.mts.timestamp-1698257515050-21db1882926d9.mjs
ShGKme, line 1: This file should be removed, it is a temporary Vite file

src/components/NcAppContent/NcAppContent.vue
```diff
@@ -202,6 +209,16 @@ export default {
 			type: String,
 			default: null,
 		},
+		/**
+		 * Switch between current layout(vertical split) and no split layout. No split layout shows only 2 columns, navigation+list or navigation+content
+		 */
```
ShGKme, line 214: Navigation is not a part of this component. So `no-split` is 1-column.

```suggestion
		/**
		 * Content layout used when there is a list together with content:
		 * - `vertical-split` - a 2-column layout with list and default content separated vertically
		 * - `no-split` - a single column layout; List is shown when `showDetails` is `false`, otherwise the default slot content is shown with a back button to return to the list.
		 * On mobile screen `no-split` layout is forced.
		 */
```

src/components/NcAppContent/NcAppContent.vue
```diff
@@ -43,6 +43,13 @@ all the children of the NcAppContent component
 	</NcAppContent>
 </template>
 ```
+```vue
+<template>
+	<NcAppContent layout-mode="no-split">
+		<h2>Single-column main content, full width list</h2>
+	</NcAppContent>
+</template>
+```
```
ShGKme, line 52: `no-split` doesn't make sense in the example without the list, because `layout` is only used for `list + content`

src/components/NcListItem/NcListItem.vue
```diff
@@ -31,6 +31,7 @@
 			:active="true"
 			:details="'1h'"
 			:counter-number="44"
+			one-line
```
ShGKme, line 34: I'm not sure it should be in the default usage example. It could be confusing mixing with the default layout. I'd show it in a separated example in the docs.

image

src/components/NcListItem/NcListItem.vue
```diff
@@ -333,7 +331,8 @@
 			:class="{ 'list-item__wrapper--active' : isActive || active }">
 			<div ref="list-item"
 				class="list-item"
-				:class="{ 'list-item--compact': compact }"
+				:class="{ 'list-item--compact': compact,
+					'list-item--one-line': oneLine,}"
```
ShGKme, line 335: **nitpick**
```suggestion
				:class="{ 
					'list-item--compact': compact,
					'list-item--one-line': oneLine,
				}"
```

src/components/NcListItem/NcListItem.vue
```diff
@@ -767,12 +781,34 @@ export default {
...
+			display: flex;
+			flex-direction: row;
+			justify-content: end;
+		}
+	}
 
 	&__anchor {
 		display: flex;
-		flex: 1 0 auto;
+		flex: 1 0 10%;
```
ShGKme, line 808: This `10%` looks a bit suspicious. But I haven't found we we use it...

src/components/NcListItem/NcListItem.vue
```diff
@@ -767,12 +781,34 @@ export default {
...
 		align-items: center;
-		height: var(--default-clickable-area);
+		height: calc(var(--default-clickable-area) + 7);
```
ShGKme, line 810: We cannot add `7` with unit like `px`

GretaD: this shouldnt be there at all, not sure what i was trying to do. sorry :)

src/components/NcListItem/NcListItem.vue
GretaD: i changed it back to auto, and nothing was broken, so maybe we dont need to 10% anymore.

src/components/NcAppContent/NcAppContent.vue
```diff
@@ -43,6 +43,15 @@ all the children of the NcAppContent component
...
 ```
+```vue
+<template>
+	<NcAppContent layout="no-split">
+		<template #list>
+			<div>Full width list</div>
+		</template>
+	</NcAppContent>
+</template>
+```
```
ShGKme, line 54: This example also has only one slot content, it doesn't show a switch between list and content and looks like just "Single-column".

Making a good example here could be complex to fully emulate `NcContent`. Maybe only keep the description in the prop comment and leave an example of usages for now?

GretaD: ok, im removing the example

ShGKme (approved):
Everything seems to work fine, thanks for changes 😊

ChristophWurst (approved):
1. No changes for an app that doesn't change :+1: :pray: 
2. Works with https://github.com/nextcloud/mail/pull/9285 and a few minor adjustments :+1:

PR #5216, opened 2024-02-06
enh: added option to render icon and text for breadcrumb

emoral435:
For nextcloud/server#43604

### 🖼️ Screenshots

🏡 After
-
firefox_FroddX5muA
-

## Summary
* To prevent a breaking change, I decided to add a prop that user configures to enable the nationality to add text to accompany a icon breadcrumb that is a NcButton, and not an NcAction.
* Since default for iconText is false, all previous library users of NcBreadcrumb will not have breaking changes
* Library now allows user to have NcButton with just text, just icon, and now icon and text 

### 🏁 Checklist

### DIFF
--- a/src/components/NcBreadcrumb/NcBreadcrumb.vue
+++ b/src/components/NcBreadcrumb/NcBreadcrumb.vue
@@ -40,28 +40,19 @@ Renders a button element when given no redirection props, otherwise, renders <a/
 		@dragover.prevent="() => {}"
 		@dragenter="dragEnter"
 		@dragleave="dragLeave">
-		<NcButton v-if="(icon || $slots.icon) && !$slots.default"
+		<NcButton v-if="(name || icon || $slots.icon) && !$slots.default"
 			:title="title"
 			:aria-label="icon ? name : undefined"
 			type="tertiary"
 			v-bind="linkAttributes"
 			v-on="$listeners">
-			<template #icon>
+			<template v-if="$slots.icon || icon" #icon>
 				<!-- @slot Slot for passing a material design icon. Precedes the icon and name prop. -->
 				<slot name="icon">
 					<span :class="icon" class="icon" />
 				</slot>
 			</template>
-			<template v-if="forceIconText" #default>
-				{{ name }}
-			</template>
-		</NcButton>
-		<NcButton v-else-if="!$slots.default"
-			:aria-label="icon ? name : undefined"
-			type="tertiary"
-			v-bind="linkAttributes"
-			v-on="$listeners">
-			<template #default>
+			<template v-if="!($slots.icon || icon) || forceIconText" #default>
 				{{ name }}
 			</template>
 		</NcButton>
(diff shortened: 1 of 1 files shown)


src/components/NcBreadcrumb/NcBreadcrumb.vue
```diff
@@ -137,13 +140,21 @@ export default {
...
 		 */
 		icon: {
 			type: String,
 			default: '',
 		},
 
+		/**
+		 * Enable name to accompany the icon to create icon + text NcButton
+		 */
+		iconText: {
```
Pytal, line 153:
```suggestion
		forceIconText: {
```
`iconText` makes it seem like a string value when it is actually a boolean and `name` is the actual text that is displayed

Pytal (dismissed):

src/components/NcBreadcrumb/NcBreadcrumb.vue
emoral435: Agreed, that conveys the props usage better 👍

raimund-schluessler (changes requested):
These changes have a logic flaw and lead to the name being always shown.

src/components/NcBreadcrumb/NcBreadcrumb.vue
```diff
@@ -46,12 +46,15 @@ Renders a button element when given no redirection props, otherwise, renders <a/
...
 			v-bind="linkAttributes"
 			v-on="$listeners">
-			<template v-if="$slots.icon || icon" #icon>
+			<template v-if="icon || $slots.icon" #icon>
 				<!-- @slot Slot for passing a material design icon. Precedes the icon and name prop. -->
 				<slot name="icon">
 					<span :class="icon" class="icon" />
 				</slot>
 			</template>
+			<template v-if="($slots.icon || icon) && iconText" #default>
```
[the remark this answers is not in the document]
emoral435: Agreed, changed up logic in newest commit

emoral435:
@raimund-schluessler I agree with your requested changes. In order to simplify the logic, whilst not making any breaking changes, here are the logic design choices I made:
* First check if icon or iconSlot is passed, and there is no default slot being passed in
    * If so, load icon, then check to see if we want to load accompanying text
* If neither icon nor default slot is passed, then we need to have a fail safe. This fail safe is loading in the name text
* If default slot IS given, then the same logic as previous commits is used

raimund-schluessler:
> Does this align with your requested changes?

Your changes lead to the desired outcome, but they are quite a bit more complicated then they need to (especially the duplication of `NcButton`). I pushed a commit to show what i meant.

raimund-schluessler (approved):

emoral435:
> bit more complicated then they need to be
Ah, I see what you mean. I saw this problem while developing and I like this solution, but I have a question. 

```js
<NcButton v-if="(name || icon || $slots.icon) && !$slots.default"
```

`name` prop is required. So unless users explicitly put "", the check will never check if the icon prop or icon slot is inputted. From what I gathered, the checks in place where to always at least have some information shown in the button.
This new change means that if no default slot was input, but also no icon prop or icon slot, then the button will just be empty, no?

raimund-schluessler:
Without icon, icon slot and default slot, we still show the name. But yes, if that is also empty, nothing will be shown, because nothing was provided. I don't think that is an issue. NcButton works the same.

skjnldsv (approved):

PR #5218, opened 2024-02-06
feat: Split reference and smart picker functions from components

juliushaertl:
This change allows importing reference widget registration methods without pulling a large dependency chain of vue components in right away. That way one can reduce the bundle size of entrypoints quite a bit when moving required components to be lazy imported.

For tables this brings down the entrypoint from 4MB to 16KB

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot 2024-02-06 at 14 45 28 | Screenshot 2024-02-06 at 14 46 39

### 🏁 Checklist


juliushaertl:
Not entirely sure about the structure, which is any suggestions would be welcome to improve of course ;)

ShGKme (approved):
Cool!

Looks fine according to the code and the `dist` looks fine.

> Not entirely sure about the structure, which is any suggestions would be welcome to improve of course ;)

Currently there is no better place, I think.

An alternative would be to add a new folder, add it to build and to `package.exports`. I'm also not a fun of the current "separate by file type" structure. But we can rethink it for v9 or v10.

st3iny (approved):

juliushaertl:
Already present in next

PR #5219, opened 2024-02-07
feat(NcAppSidebar): move focus to sidebar on open and auto return focus on close

ShGKme:
### ☑️ Resolves

* Fix for https://github.com/nextcloud/server/issues/41877
* Can be tested with https://github.com/nextcloud/server/pull/42444

See description in commits.

### On desktop

1. Open sidebar
2. Focus is moved to the sidebar
3. Close sidebar
4. Focus is moved to the trigger button
5. For menu - special case, return to the menu trigger

https://github.com/nextcloud-libraries/nextcloud-vue/assets/25978914/a286b6a2-5a55-4d70-bd39-674016000bc4

### On mobile with focus-trap - also works

https://github.com/nextcloud-libraries/nextcloud-vue/assets/25978914/2dd59a40-0587-4c64-8e96-ebd0e70df85a

### When sidebar was open, but a new file or tab was selected — focus is moved to the tab, and the element to return is updated

https://github.com/nextcloud-libraries/nextcloud-vue/assets/25978914/6698351a-883b-4162-850f-6c250081f11d

### When the sidebar is initially open on page load - no focus move

https://github.com/nextcloud-libraries/nextcloud-vue/assets/25978914/7d940181-8ee8-43c0-88bd-50d42f7aa60e

### Also works for apps

https://github.com/nextcloud-libraries/nextcloud-vue/assets/25978914/0dab45a7-e618-4a11-9fcb-ca8962d60dae

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppSidebar/NcAppSidebar.vue
+++ b/src/components/NcAppSidebar/NcAppSidebar.vue
@@ -405,11 +405,12 @@ export default {
 								<!-- main name -->
 								<h2 v-show="!nameEditable"
 									:id="`app-sidebar-vue-${uid}__header`"
+									ref="header"
 									v-linkify="{text: name, linkify: linkifyName}"
 									:aria-label="title"
 									:title="title"
 									class="app-sidebar-header__mainname"
-									:tabindex="nameEditable ? 0 : undefined"
+									:tabindex="nameEditable ? 0 : -1"
 									@click.self="editName">
 									{{ name }}
 								</h2>
@@ -665,6 +666,7 @@ export default {
 			favoriteTranslated: t('Favorite'),
 			isStarred: this.starred,
 			focusTrap: null,
+			elementToReturnFocus: null,
 		}
 	},
 
@@ -690,7 +692,16 @@ export default {
 		},
 	},
 
+	created() {
+		this.preserveElementToReturnFocus()
+	},
+
 	mounted() {
+		// Focus sidebar on open only if it was opened by a user interaction
+		if (this.elementToReturnFocus) {
+			this.focus()
+		}
+
 		this.toggleFocusTrap()
 	},
 
@@ -701,6 +712,23 @@ export default {
 	},
 
 	methods: {
+		preserveElementToReturnFocus() {
+			// Save the element that had focus before the sidebar was opened to return back on close
+			if (document.activeElement && document.activeElement !== document.body) {
+				this.elementToReturnFocus = document.activeElement
+
+				// Special case for menus (NcActions)
+				// If a sidebar was opened from a menu item, we want to return focus to the menu trigger instead of the item
+				if (this.elementToReturnFocus.getAttribute('role') === 'menuitem') {
+					const menu = this.elementToReturnFocus.closest('[role="menu"]')
+					if (menu) {
+						const menuTrigger = document.querySelector(`[aria-controls="${menu.id}"]`)
+						this.elementToReturnFocus = menuTrigger
+					}
+				}
+			}
+		},
+
 		initFocusTrap() {
 			if (this.focusTrap) {
 				return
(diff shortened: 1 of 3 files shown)


ShGKme:
Checking Cypress

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -403,11 +404,13 @@ export default {
...
 								<!-- main name -->
 								<h2 v-show="!nameEditable"
+									:id="`app-sidebar-vue-${uid}__header`"
+									ref="header"
 									v-linkify="{text: name, linkify: linkifyName}"
 									:aria-label="title"
 									:title="title"
 									class="app-sidebar-header__mainname"
-									:tabindex="nameEditable ? 0 : undefined"
+									:tabindex="nameEditable ? 0 : -1"
```
Pytal, line 413:
```suggestion
									:tabindex="nameEditable ? 0 : undefined"
```
Setting tabindex="-1" will make it non-sequentially focusable but still click focusable which I don't think is desired

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -761,6 +793,10 @@ export default {
 			 * @type {HTMLElement}
 			 */
 			this.$emit('closed', element)
+
+			// Return focus to the element that had focus before the sidebar was opened
+			this.elementToReturnFocus?.focus({ focusVisible: true })
```
Pytal, line 798: Maybe note that this is only supported in Firefox currently?

src/components/NcAppSidebarTab/NcAppSidebarTab.vue
```diff
@@ -30,8 +30,8 @@
 		:aria-label="isTablistShown() ? undefined : name"
 		:aria-labelledby="isTablistShown() ? `tab-button-${id}` : undefined"
 		class="app-sidebar__tab"
-		tabindex="0"
-		role="tabpanel"
+		:tabindex="isTablistShown() ? 0 : -1"
```
Pytal, line 33:
```suggestion
		:tabindex="isTablistShown() ? 0 : undefined"
```
Similar to above, the tabindex should not be set at all otherwise it will still be click focusable

Pytal (approved):
Other than minor comments, nice!

src/components/NcAppSidebar/NcAppSidebar.vue
ShGKme: The recording in the PR was done in Chromium-based browser...

src/components/NcAppSidebar/NcAppSidebar.vue
ShGKme: It is actually important to have `-1` here. This element should not be in the keyboard focus sequence because it is not an interactive element.

But having no `tabindex` is also not correct, because this element receives programmatic focus.

`-1` allows making an element programmatically focusable without adding it to the Tab sequence.

For example, without it, there will be no outline on focus on `<h2>` and it will be broken in the `focus-trap`.

See also:

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

image

Pytal: Makes sense 👍

src/components/NcAppSidebar/NcAppSidebar.vue
Pytal: Guess MDN is not up-to-date then :(

src/components/NcAppSidebar/NcAppSidebar.vue
ShGKme: I actually spend hours trying to understand what can I focus here, and why the header didn't work =D

Pytal: That's how the tabindex crumbles 😢

ShGKme:
> Checking Cypress

`<Transition>` didn't work, `<transition>` works.

JuliaKirschenheuter (approved):
works great! Thanks a lot!

PR #5231, opened 2024-02-09
[next] fix(NcActionInput): listen to correct events

### LINKED ISSUE #5229: [next] `NcActionInput` does not listen to correct events
No idea how I could have overlooked this, but `NcActionInput` does not listen to the correct events. It still listens to `@change` and `@input` which got replace with `@update:model-value` or simply `v-model`.
I will send a PR later.

raimund-schluessler:
### ☑️ Resolves

* This PR fixes `NcActionInput` to listen to the correct events of its child components. Before, the events of its childcomponents would not trigger an update of the model value.
* Fix #5229

### DIFF
--- a/src/components/NcActionInput/NcActionInput.vue
+++ b/src/components/NcActionInput/NcActionInput.vue
@@ -137,7 +137,7 @@ For the `NcSelect` component, all events will be passed through. Please see the 
 </docs>
 
 <template>
-	<li class="action" :class="{ 'action--disabled': disabled }">
+	<li class="action" :class="[{ 'action--disabled': disabled }, $props.class ]">
 		<span :class="{
 				'action-input-picker--disabled': disabled,
 				'action-input--visible-label': labelOutside && label,
@@ -388,6 +388,13 @@ export default {
 		trailingButtonLabel: {
 			type: String,
 			default: t('Submit'),
+		},
+		/**
+		 * CSS class to apply to the root element.
+		 */
+		class: {
+			type: [String, Array, Object],
+			default: '',
 		},
 	},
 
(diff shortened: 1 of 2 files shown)


ShGKme (commented):
Makes sense.

I haven't tested, but it looks like `input` and `change` events doesn't work anymore with this PR. As they only proxied the same native events with native `Event` payload, we probably can just remove them from the `emits` declaration to bring them back. (Will work as a part of `v-bind="$attrs"`)

skjnldsv (approved):

raimund-schluessler:
> Makes sense.
> 
> I haven't tested, but it looks like `input` and `change` events doesn't work anymore with this PR. As they only proxied the same native events with native `Event` payload, we probably can just remove them from the `emits` declaration to bring them back. (Will work as a part of `v-bind="$attrs"`)

@ShGKme I removed them from the `emits` declaration and `@change` and `@input` are emitted now. Unfortunately, they are emitted twice, do you have an idea why? Could this be a problem with `NcTextField`?

raimund-schluessler:
Setting `inheritAttrs: false` for `NcActionInput` might then break things we want, like style or css  no?

ShGKme:
> Setting `inheritAttrs: false` for `NcActionInput` might then break things we want, like style or css no?

Yes, but using `v-bind="$attrs"` without `inheritAttrs: false` is always an issue, it duplicates everything. We can add `class` to `props` as we did in other places to set it on the root.

raimund-schluessler:
> > Setting `inheritAttrs: false` for `NcActionInput` might then break things we want, like style or css no?
> 
> Yes, but using `v-bind="$attrs"` without `inheritAttrs: false` is always an issue, it duplicates everything. We can add `class` to `props` as we did in other places to set it on the root.

Done.

emoral435 (approved):

ShGKme (approved):

PR #5235, opened 2024-02-09
fix(NcRichContenteditable): adjust styles and use CSS Modules to avoid global styles leak

ShGKme:
### ☑️ Resolves

* Fix https://github.com/nextcloud/mail/issues/9310
* Fix https://github.com/nextcloud/server/issues/43478

`NcRichContenteditable` has global CSS styles to style Tribute autocomplete. It works fine, until there are more than one version of those styles on the page. Styles should be scoped.

@szaimen @hamza221 @JuliaKirschenheuter Could you please check it in your apps? The bug was not reproducible for me...

## Screenshots

Note, there is no `data-v-<hash>` attrs in both cases.

Before | After
---|---
image | image

### 🚧 Tasks

- [x] Use CSS Modules for scoping styles of Tribute autocomplete (`scoped` doesn't work for non-component styles and doesn't truly scope from styles outside, having only one-way scoping)
- [x] Adjust styles, remove some unneeded classes and fix typos in CSS
  - Some issues were not noticeable before, because of other global styles
- [x] Add CSS Modules support for Styleguidist

### 🏁 Checklist

### DIFF
--- a/src/components/NcRichContenteditable/NcRichContenteditable.vue
+++ b/src/components/NcRichContenteditable/NcRichContenteditable.vue
@@ -575,7 +575,7 @@ export default {
 				// Class added to the menu container
 				containerClass: `${this.$style['tribute-container']} ${this.$style['tribute-container-autocomplete']}`,
 				// Class added to each list item
-				itemClass: `${this.$style['tribute-container__item']}`,
+				itemClass: this.$style['tribute-container__item'],
 
 			})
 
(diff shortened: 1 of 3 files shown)


JuliaKirschenheuter (approved):
works fine now! Thanks a lot!

susnux (approved):
Works and much cleaner!
(Funny how complex the webpack config is vs. no changes in the vite config^^)

ShGKme:
> (Funny how complex the webpack config is vs. no changes in the vite config^^)

Vite has pre-config for Rollup for "typical" bundling, but then you need to write your own plugin if you have something more special 😝

PR #5239, opened 2024-02-11
[next] fix(NcRichText*): listen to correct events form NcSelect

### LINKED ISSUE #5232: [next] `NcRichText*` listens to wrong event of `NcSelect`
It needs to listen to `@update:model-value` instead here:

https://github.com/nextcloud-libraries/nextcloud-vue/blob/01dd69f74cf1f51432209f10994b6127d3e97a41/src/components/NcRichText/NcReferencePicker/NcSearch.vue#L20

https://github.com/nextcloud-libraries/nextcloud-vue/blob/01dd69f74cf1f51432209f10994b6127d3e97a41/src/components/NcRichText/NcReferencePicker/NcProviderList.vue#L15

raimund-schluessler:
### ☑️ Resolves

- This corrects to which event `NcRichText*` components are listening from `NcSelect`.
- Fixes #5232


Antreesy (approved):
For the reference: should follow-up these migration changes: https://github.com/nextcloud-libraries/nextcloud-vue/commit/61bb1f9b2a6ce3ba71004d06ea8f2899be45e5d1

Seems legit, but didn't test it. Do you have a guide to reproduce it?

ShGKme (approved):

PR #5250, opened 2024-02-15
fix: regression in list item's extra slot

hamza221:
### ☑️ Resolves

- Fix https://github.com/nextcloud/mail/issues/9334

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/NcListItem/NcListItem.vue
+++ b/src/components/NcListItem/NcListItem.vue
@@ -394,20 +394,19 @@
 							</div>
 						</div>
 					</div>
+					<!-- Actions -->
+					<div v-show="forceDisplayActions || displayActionsOnHoverFocus"
+						class="list-item-content__actions"
+						@focusout="handleBlur">
+						<NcActions ref="actions"
+							:primary="isActive || active"
+							:aria-label="computedActionsAriaLabel"
+							@update:open="handleActionsUpdateOpen">
+							<!-- @slot Provide the actions for the right side quick menu -->
+							<slot name="actions" />
+						</NcActions>
+					</div>
 				</a>
-
-				<!-- Actions -->
-				<div v-show="forceDisplayActions || displayActionsOnHoverFocus"
-					class="list-item-content__actions"
-					@focusout="handleBlur">
-					<NcActions ref="actions"
-						:primary="isActive || active"
-						:aria-label="computedActionsAriaLabel"
-						@update:open="handleActionsUpdateOpen">
-						<!-- @slot Provide the actions for the right side quick menu -->
-						<slot name="actions" />
-					</NcActions>
-				</div>
 
 				<!-- @slot Extra elements below the item -->
 				<div v-if="$slots.extra" class="list-item__extra">
@@ -732,7 +731,7 @@ export default {
 // NcListItem
 .list-item {
 	box-sizing: border-box;
-	display: flex;
+	display: block;
 	position: relative;
 	flex: 0 0 auto;
 	justify-content: flex-start;


hamza221:
cc @GretaD  @susnux @ShGKme

susnux:
The "after" image also looks broken, for the `NcListItem` I would only expect two lines of text working and the tags to be shown on the very right.

hamza221:
> The "after" image also looks broken,

Yes the after is still a bit broken, the tags should be aligned with the lines
This is how it looks on nc.cloud
image

> for the `NcListItem` I would only expect two lines of text working and the tags to be shown on the very right.

Isn't the extra slot supposed to show elements below the lines

GretaD:
> > The "after" image also looks broken,
> 
> Yes the after is still a bit broken, the tags should be aligned with the lines This is how it looks on nc.cloud image
> 
> > for the `NcListItem` I would only expect two lines of text working and the tags to be shown on the very right.
> 
> Isn't the extra slot supposed to show elements below the lines

this is fixed on mail side, i already pushed a pr for that https://github.com/nextcloud/mail/pull/9336

src/components/NcListItem/NcListItem.vue
```diff
@@ -394,21 +394,20 @@
 							</div>
 						</div>
 					</div>
+					<!-- Actions -->
```
susnux, line 397: This **must not** be done! An interactive element (the `a`) must **not** contain any other interactive element (the `NcActions`).
We changed this just a couple of weeks ago.

susnux (changes requested):
This changes reverting a11y fixes and are not valid -> must not be done. You have to find a solution with the actions outside the anchor element

susnux:
I mean if it fits in the component fine, but we can not have nested interactive elements. This only works with some specific focus handling (e.g. how menu bars work, you can tab the first element but next tab will be outside the menu bar, you then need to use arrows to change focus).

GretaD:
Hello @susnux, this is still an issue on mail, i would get @jancborchardt @marcoambrosini @nimishavijay involved, what would be a good compromise :)

If it was up to me, I would keep it on the right side, with the argument, that its like this since couple of months and nobody complained about the space, which means that either people are fine with it, or its not used at all :), the tagging is only visual - you cannot filter a certain tag and now that we have different layouts, people can always choose a layout where the list is wide. But, of course, i would be fine to work on another solution.

marcoambrosini:
> An interactive element (the a) must not contain any other interactive element (the NcActions).

Are you sure that this is the case @susnux afaik an A cannot contain another A but it can contain buttons

ShGKme:
> e.g. most cards in card layouts

In valid implementation, they are not wrapped in `<a>`. They have a link inside and either use custom `onclick` (like we do in Files) or stretch the link in a pseudo element (like we do in Apps list).

But in both HTML validity and accessibility having nested interactive elements is not valid.

ShGKme:
See: https://github.com/nextcloud-libraries/nextcloud-vue/pull/5198

GretaD:
thank you for all the feedback,

@marcoambrosini how should we proceed then?

marcoambrosini:
I donn't know which would be the better approach. cc @ShGKme and @susnux

hamza221:
what we're trying to accomplish is just a third line for 'non interactive- non clickable "tags" ' would that be acceptable ? 
So move the slot outside the NcActions  rename it (thirdline? ) ?

PR #5266, opened 2024-02-20
[next] chore: Update and pin development dependencies

susnux:
Just update and also pin the used versions.


raimund-schluessler (approved):

ShGKme (approved):

susnux:
Rebased and pushed a commit fixing styleguid building (we really should replace that with something not using webpack, handling 2 bundlers in 1 project is really overkill)

PR #5269, opened 2024-02-20
feat(NcAppContent): Allow to set the page title

susnux:
### ☑️ Resolves

Sometimes - e.g. when you have multiple view - you want to adjust the page title, that is shown as the browsers tab name. This adds a property to set that value.

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppContent/NcAppContent.vue
+++ b/src/components/NcAppContent/NcAppContent.vue
@@ -50,6 +50,45 @@ The list size must be between the min and the max width value.
 	:list-max-width="45"
 >...</NcAppContent>
 ```
+
+#### Usage: Custom document title
+For accessibility reasons every document should have a `h1` heading,
+this is visually hidden, but required for a semantically correct document.
+You can use your app name or current view for the heading.
+
+Additionally you can set a custom document title, e.g. to show the current status.
+
+```vue
+<template>
+	<NcAppContent :pageHeading="heading ? 'Heading' : undefined" :pageTitle="title ? 'Title' : undefined" >
+		<NcCheckboxRadioSwitch type="switch" :checked.sync="title">
+			Toggle title
+		</NcCheckboxRadioSwitch>
+		<NcCheckboxRadioSwitch type="switch" :checked.sync="heading">
+			Toggle Heading
+		</NcCheckboxRadioSwitch>
+		<NcButton @click="reset">Reset</NcButton>
+	</NcAppContent>
+</template>
+
+<script>
+export default {
+	data() {
+		return {
+			heading: false,
+			title: false,
+		}
+	},
+	methods: {
+		reset() {
+			this.heading = false
+			this.title = false
+			document.title = ''
+		},
+	},
+}
+</script>
+```
 </docs>
 
 <template>
@@ -103,18 +142,22 @@ The list size must be between the min and the max width value.
 </template>
 
 <script>
-import NcAppDetailsToggle from './NcAppDetailsToggle.vue'
-import { useIsMobile } from '../../composables/useIsMobile/index.js'
-
 import { getBuilder } from '@nextcloud/browser-storage'
+import { emit } from '@nextcloud/event-bus'
+import { loadState } from '@nextcloud/initial-state'
 import { useSwipe } from '@vueuse/core'
 import { Splitpanes, Pane } from 'splitpanes'
-
-import 'splitpanes/dist/splitpanes.css'
-import { emit } from '@nextcloud/event-bus'
+import { useIsMobile } from '../../composables/useIsMobile/index.js'
 import { isRtl } from '../../utils/rtl.ts'
 
+import NcAppDetailsToggle from './NcAppDetailsToggle.vue'
+
+import 'splitpanes/dist/splitpanes.css'
+
 const browserStorage = getBuilder('nextcloud').persist().build()
+const { name: productName } = loadState('theming', 'data', { name: 'Nextcloud' })
+const activeApp = loadState('core', 'active-app', appName)
+const localizedAppName = loadState('core', 'apps', {})[activeApp]?.name ?? appName
 
 /**
  * App content container to be used for the main content of your app
@@ -217,6 +260,18 @@ export default {
 				return ['no-split', 'vertical-split', 'horizontal-split'].includes(value)
 			},
 		},
+
+		/**
+		 * Allow setting the page's `<title>`
+		 *
+		 * If a page heading is set it defaults to `{pageHeading} - {appName} - {productName}` e.g. `Favorites - Files - Nextcloud`.
+		 * When the page heading and the app name is the same only one is used, e.g. `Files - Files - Nextcloud` is shown as `Files - Nextcloud`.
+		 * When setting the prop then the following format will be used: `{pageTitle} - {pageHeading || appName} - {productName}`
+		 */
+		pageTitle: {
+			type: String,
+			default: null,
+		},
 	},
 
 	emits: [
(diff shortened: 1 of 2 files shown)


src/components/NcAppContent/NcAppContent.vue
```diff
@@ -268,6 +279,24 @@ export default {
...
 			}
 		},
+
+		realPageTitle() {
+			if (this.pageTitle) {
+				return this.pageTitle
+			}
+			if (this.pageHeading) {
+				return t('{view} - {app} - {product}', { view: this.pageHeading, app: appName, product: productName })
+			}
```
ShGKme, line 289: Does it mean that for custom `pageTitle` the developer has to manually build string in a format of `{view} - {app} - {product}`?

Shouldn't we fix at least `{app} - {product}` or `{product}` parts, allowing to provide the view name? Or provide a layout like in [unhead](https://unhead.unjs.io/usage/guides/title-template)?

Do you have an example of expected `pageTitle` values?

For example, if this is for `Files` and I open `Foo` folder on `Favorites`, that `pageTitle` is expected?
1. `Favorites`
1. `Favorites - Nextcloud`
1. `Favorites - Files - Nextcloud`
1. `Foo`
1. `Foo - Files - Nextcloud`
1. `Foo - Favorites - Nextcloud`
1. `Foo - Favorites - Files - Nextcloud`

src/components/NcAppContent/NcAppContent.vue
```diff
@@ -202,6 +205,14 @@ export default {
 			type: String,
 			default: null,
 		},
+
+		/**
+		 * Allow setting the page title (e.g. shown as the browsers tab name)
+		 */
```
ShGKme, line 211:
```suggestion
		/**
		 * Custom page's `<title>`, for example, ...example...
		 * {pageHeading} - {appName} - {productName} is used by default.
		 */
```

src/components/NcAppContent/NcAppContent.vue
ShGKme, line 289: `appName` seems to be a non-translated constant from the `package.json` file. Not sure it can be used as a page title

src/components/NcAppContent/NcAppContent.vue
```diff
@@ -268,6 +279,24 @@ export default {
 				},
 			}
 		},
+
+		realPageTitle() {
+			if (this.pageTitle) {
+				return this.pageTitle
+			}
+			if (this.pageHeading) {
+				return t('{view} - {app} - {product}', { view: this.pageHeading, app: appName, product: productName })
```
ShGKme, line 288: Does `{} - {} - {}` require translation? At least on server, it is not.

src/components/NcAppContent/NcAppContent.vue
```diff
@@ -115,17 +115,20 @@ The list size must be between the min and the max width value.
...
-import NcAppDetailsToggle from './NcAppDetailsToggle.vue'
-import { useIsMobile } from '../../composables/useIsMobile/index.js'
-
 import { getBuilder } from '@nextcloud/browser-storage'
 import { emit } from '@nextcloud/event-bus'
 import { useSwipe } from '@vueuse/core'
 import { Splitpanes, Pane } from 'splitpanes'
+import { useIsMobile } from '../../composables/useIsMobile/index.js'
+import { loadState } from '@nextcloud/initial-state'
+import { t } from '../../l10n'
```
skjnldsv, line 124:
```suggestion
import { t } from '../../l10n.js'
```

src/components/NcAppContent/NcAppContent.vue
susnux: The currently expected (but only implemented by a few apps) is `Favorites - Files - Nextcloud`.

ShGKme: > The currently expected (but only implemented by a few apps) is `Favorites - Files - Nextcloud`.

Then isn't `pageTitle` used as a kind of second level heading like `pageHeading`?

`{pageTitle} - {pageHeading} - {productName}`

susnux: We could also do it like that yes.

Maybe even standarize the visible headline like it is used on activity, so we could have the `h1` with the app name, the `h2` with the current view -> page title `view - app name - product name`.
But that would need more discussion.

src/components/NcAppContent/NcAppContent.vue
ShGKme: Anyway, the server has it without translation now.

src/components/NcAppContent/NcAppContent.vue
```diff
@@ -268,6 +284,24 @@ export default {
...
+				return t('{view} - {app} - {product}', { view: this.pageHeading, app: localizedAppName, product: productName })
+			}
+			return null
+		},
+	},
+
+	watch: {
+		realPageTitle() {
+			if (this.realPageTitle !== null) {
+				window.document.title = this.realPageTitle
```
ShGKme, line 302: nitpick
```suggestion
				document.title = this.realPageTitle
```

src/components/NcAppContent/NcAppContent.vue
```diff
@@ -268,6 +284,24 @@ export default {
...
+			}
+		},
```
ShGKme, line 304: Shouldn't it be `immediate`? Otherwise, it only works on changing title/heading but not setting.

Or do I miss something?

susnux: No you are right, fixed!

src/components/NcAppContent/NcAppContent.vue
ShGKme: It works fine when all the words are already translated and in RTL, space and dash doesn't change the direction.

image

src/components/NcAppContent/NcAppContent.vue
```diff
@@ -268,6 +284,27 @@ export default {
...
 			}
 		},
+
+		realPageTitle() {
+			if (this.pageTitle) {
+				return t('{view} - {app} - {product}', { view: this.pageTitle, app: this.pageHeading || localizedAppName, product: productName })
+			}
+			if (this.pageHeading) {
+				return t('{view} - {app} - {product}', { view: this.pageHeading, app: localizedAppName, product: productName })
+			}
```
ShGKme, line 294: These strings seem to not need translators.

It works fine when all the words are already translated, and in RTL, space and dash don't change the direction.

image

ShGKme (approved):
Everything looks fine, cool feature.

Only one old thing with the translations

src/components/NcAppContent/NcAppContent.vue
```diff
@@ -202,6 +207,17 @@ export default {
 			type: String,
 			default: null,
 		},
+
+		/**
+		 * Allow setting the page's `<title>`
+		 *
+		 * If a page heading is set it defaults to `{pageHeading} - {appName} - {productName}` e.g. `Favorites - Files - Nextcloud`.
+		 * When setting the prop then the following format will be used: `{pageTitle} - {pageHeading || appName} - {productName}`
```
ShGKme, line 215: Can you add an example here for the second option? I cannot figure out what it can be...

ShGKme: And docs for `pageHeading` also should be updated because now it not only setting `h1`, but also `<title>`

ShGKme: Also, does it mean that a page now can never have a 2-part title like `Talk - Nextcloud`, `Mail - Nextcloud`, `Contacts - Nextcloud` and etc.?

And the only way to have such a title is to have neither heading nor title (and no `h1` which is a11y issue).

ShGKme (dismissed):

src/components/NcAppContent/NcAppContent.vue
susnux: @ShGKme  I pushed a commit with a fix for this.
For `Talk - Nextcloud` simply set `pageHeading` to `Talk`.

ShGKme (approved):

PR #5271, opened 2024-02-20
fix: Center dashboard list items when there is no subtext

juliushaertl:
If we have no subtext the title shall be centered to the image

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot 2024-02-20 at 18 45 46 | Screenshot 2024-02-20 at 18 45 24

### 🏁 Checklist


ShGKme (approved):

susnux:
> Also works the way of aligning works as before with the flex layout just that we center within the minimum height.

No I meant it more like if it makes sense to only align centered if **all** elements have no sub text.

PR #5272, opened 2024-02-20
fix(NcReferenceList): Resolve relative URLs before fetching references

mejo-:
Resolve relative URLs and those without a location using `window.location` before fetching references for them. This fixes e.g. getting references for relative links to other pages in Collectives.

### 🏁 Checklist

### DIFF
--- a/src/components/NcRichText/NcReferenceList.vue
+++ b/src/components/NcRichText/NcReferenceList.vue
@@ -82,6 +82,9 @@ export default {
 				richObjectType: 'open-graph',
 			}
 		},
+		fullUrl() {
+			return new URL(this.text.trim(), window.location)
+		},
 	},
 	watch: {
 		text: 'fetch',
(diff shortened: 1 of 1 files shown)


juliushaertl (approved):

Antreesy:
Breaking messages in Talk, if couldn't resolve a URL:
Loading for all messages | Failing at more-than-just-link messages
--|--
image | image

Antreesy:
> The links in your screenshot already look like full URLs, so they shouldn't get altered by the changes from this PR.

They're not, but new URL creates an object for every text, whether it has a link or not.
BEFORE change, there was no attempt to resolve a reference, and messages don't have the spinner
Now, messages which includes both plain text and link, couldn't be loaded (third example)

I'm looking into how to fix that

mejo-:
Ah, so the problem is that that Talk passes text that contains more than just an URL to NcReferenceList. In Text, we check first if the passed text is merely an URL and only pass it in this case.

PR #5285, opened 2024-02-21
feat(NcAppSidebar): Add `subname` slot

susnux:
### ☑️ Resolves

This adds a `subname` slot for the same purpose as the `subname` prop, but this allows more advanced usage.
My use case: Add an icon with alt text / aria label.

Nextcloud files does not use the `starred` button anymore (design decision) but simply an action menu entry + an :star:  icon on the subname.
The problem is that the star icon (just a unicode symbol) is not announced by screen readers, so we need some accessible alternative. Fixed by allowing inline content as the slot content.

To implement this I also added an `inline` prop to the SVG icon wrapper to allow it to render icons inline (this is BTW something I really often saw people "hack" into the component using `deep` selectors to get rid of the `min-width` and `min-height`).

### 🖼️ Screenshots
Inline icon:
image

App sidebar:
image

### 🏁 Checklist


Pytal (approved):

skjnldsv (approved):
Thanks for taking the time to explain the background :)

PR #5290, opened 2024-02-22
fix(NcReferenceList): check if text contains valid URL before resolving

Antreesy:
### ☑️ Resolves

- Fix regression from #5272

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Loading | Result
image | image
image | image

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/NcRichText/NcReferenceList.vue
+++ b/src/components/NcRichText/NcReferenceList.vue
@@ -83,7 +83,11 @@ export default {
 			}
 		},
 		fullUrl() {
-			return new URL(this.text.trim(), window.location)
+			const matchArray = this.text.trim().match(new RegExp(URL_PATTERN))
+			if (matchArray) {
+				return new URL(matchArray[0], window.location).href
+			}
+			return null
 		},
 	},
 	watch: {
@@ -100,7 +104,7 @@ export default {
 				return
 			}
 
-			if (!(new RegExp(URL_PATTERN).exec(this.fullUrl.href))) {
+			if (!(new RegExp(URL_PATTERN).exec(this.fullUrl))) {
 				this.loading = false
 				return
 			}
@@ -116,7 +120,7 @@ export default {
 			})
 		},
 		resolve() {
-			const match = (new RegExp(URL_PATTERN).exec(this.fullUrl.href))
+			const match = (new RegExp(URL_PATTERN).exec(this.fullUrl))
 			if (this.limit === 1 && match) {
 				return axios.get(generateOcsUrl('references/resolve', 2) + `?reference=${encodeURIComponent(match[0])}`)
 			}


Antreesy:
@mejo- please, give it a test with Text and Collectives

src/components/NcRichText/NcReferenceList.vue
```diff
@@ -83,7 +83,11 @@ export default {
 			}
 		},
 		fullUrl() {
-			return new URL(this.text.trim(), window.location)
+			const matchArray = this.text.trim().match(new RegExp(URL_PATTERN))
```
mejo-, line 86: Unfortunately this will not work as `URL_PATTERN` matches only full URLs, not relative or absolute URLs without a location.

mejo- (changes requested):

src/components/NcRichText/NcReferenceList.vue
mejo-: In fact the whole idea of #5272 was to turn relative links (e.g. `../Target`) and absolute links (e.g. `/index.php/apps/collectives/Garden/Watering`) into full URLs with a location *before*  they're matched against `URL_PATTERN` :grimacing: 

Could you provide an example text of what is passed to `NcReferenceList` as property `text` which breaks currently?

Antreesy: `test message` - it tries to resolve a link: `https://nextcloud.local/index.php/call/test%20message`
`test message with a link https://nextcloud.local/index.php/call/w38mo5vy` - same: `https://nextcloud.local/index.php/call/test%20message%20with%20a%20link%20https://nextcloud.local/index.php/call/w38mo5vy`
`https://nextcloud.local/index.php/call/w38mo5vy` - resolves correctly

Antreesy: Maybe we could replace URL_PATTERN with some combination RELATIVE_URL_PATTERN?

mejo-: Would it be an option to sanitize the links inside Talk and make sure to only pass valid URLs to `NcReferenceList`?

Given that we cannot distinguish between valid relative links and mere text I only see two paths forward:

* Either make sure only links (be it full URLs, relative or absolute paths) are passed to NcReferenceList
* Or move the logic to transform relative/absolute paths into full URLs to Text. That would mean that relative/absolute paths from other apps no longer get regarded by NcReferenceList.

mejo-: > Maybe we could replace URL_PATTERN with some combination RELATIVE_URL_PATTERN?

That could work if use something like `^(\.){1,2}\/.*` as regex for `RELATIVE_URL_PATTERN`.

Antreesy: Did you mean one of these places?
https://github.com/nextcloud/text/blob/35be559ca27f384cd80e9e5fe76ee739c7c754e6/src/nodes/ParagraphView.vue#L105-L112
https://github.com/nextcloud/text/blob/35be559ca27f384cd80e9e5fe76ee739c7c754e6/src/components/Link/LinkBubbleView.vue#L148-L155

mejo-: Yeah, both probably. I'm fine with trying to handle this in Text. So let's revert my earlier PR and I'll look into it in Text.

mejo-:
I opened https://github.com/nextcloud-libraries/nextcloud-vue/pull/5291 to revert #5272.

PR #5298, opened 2024-02-23
chore(deps): use `@vueuse@10.7.2` and `vue@2.7.16`

### LINKED ISSUE #5296: [NcDialog] Cannot read properties of undefined (reading 'mounted')
cc @susnux 

```js
TypeError: Cannot read properties of undefined (reading 'mounted')
    at injectHook (vue.runtime.esm.js:3320:1)
    at vue.runtime.esm.js:3306:1
    at useMounted (index.mjs:501:1)
    at useSupported (index.mjs:509:1)
    at useResizeObserver (index.mjs:2518:1)
    at useElementSize (index.mjs:2676:1)
    at setup (NcDialog-652A8lMI.mjs:158:1)
    at invokeWithErrorHandling (vue.runtime.esm.js:3033:1)
    at initSetup (vue.runtime.esm.js:2457:1)
    at initState (vue.runtime.esm.js:5385:1)
logError @ vue.runtime.esm.js:3065
globalHandleError @ vue.runtime.esm.js:3057
handleError @ vue.runtime.esm.js:3024
invokeWithErrorHandling @ vue.runtime.esm.js:3040
initSetup @ vue.runtime.esm.js:2457
initState @ vue.runtime.esm.js:5385
Vue._init @ vue.runtime.esm.js:5714
VueComponent @ vue.runtime.esm.js:5849
createComponentInstanceForVnode @ vue.runtime.esm.js:4564
init @ vue.runtime.esm.js:4426
createComponent @ vue.runtime.esm.js:6591
createElm @ vue.runtime.esm.js:6545
patch @ vue.runtime.esm.js:7133
Vue._update @ vue.runtime.esm.js:3785
updateComponent @ vue.runtime.esm.js:3891
Watcher.get @ vue.runtime.esm.js:3462
Watcher.run @ vue.runtime.esm.js:3538
flushSchedulerQueue @ vue.runtime.esm.js:4141
(anonymous) @ vue.runtime.esm.js:3159
flushCallbacks @ vue.runtime.esm.js:3081
Promise.then (async)
timerFunc @ vue.runtime.esm.js:3106
nextTick @ vue.runtime.esm.js:3171
queueWatcher @ vue.runtime.esm.js:4227
Watcher.update @ vue.runtime.esm.js:3529
Dep.notify @ vue.runtime.esm.js:732
reactiveSetter @ vue.runtime.esm.js:967
proxySetter @ vue.runtime.esm.js:5376
click @ Users.vue:193
invokeWithErrorHandling @ vue.runtime.esm.js:3033
invoker @ vue.runtime.esm.js:1832
invokeWithErrorHandling @ vue.runtime.esm.js:3033
Vue.$emit @ vue.runtime.esm.js:3732
onClick @ NcAppNavigationItem-A1yAdDNN.mjs:418
click @ NcAppNavigationItem-A1yAdDNN.mjs:480
invokeWithErrorHandling @ vue.runtime.esm.js:3033
invoker @ vue.runtime.esm.js:1832
original_1._wrapper @ vue.runtime.esm.js:7500
12:10:02.154 
```

skjnldsv:
Fix #5296 

> This is causes by `@vueuse/component` being updated to `1.8.0` (you updated that on server?).
1.8.0 has a regression with Vue 2, they already fixed that but it is still unreleased:
> https://github.com/vueuse/vueuse/pull/3802


ShGKme (approved):

package.json
```diff
@@ -77,8 +77,8 @@
     "@nextcloud/logger": "^2.2.1",
     "@nextcloud/router": "^3.0.0",
     "@nextcloud/vue-select": "^3.25.0",
-    "@vueuse/components": "^10.0.2",
-    "@vueuse/core": "^10.1.2",
+    "@vueuse/components": "^10.7.2",
```
ShGKme, line 80: Doesn't it include the bug, because the fix was not released?

susnux: yes it needs to be:
`"@vueuse/components": "~10.7.2 || ^10.8.1"`

ShGKme: `npm ci` doesn't use lock from the dependencies, only the root.

skjnldsv: I'm not sure about this!
That's why doing `npm i`  or deleting the lockfile on server generates so many changes.
I just tested it, and installing this PR does in fact instal the proper `10.7.x` version on server too :shrug:

ShGKme: I am sure about it.

Lock files of the dependencies are not used, only on the root.

It installs `10.7.x` on server after `npm ci` if the server has `10.7.x` in the lock file.

But having `10.7.x` in the lock file of `@nextcloud/vue` does nothing when using the package.
Updating dependencies on server will install `10.8.x`.

Otherwise we would always have tons of packages
```
root
- A
  - package -- lib:^1.0.0
  - package-lock -- lib:1.2.3
- B
  - package -- lib:^1.0.0
  - package-lock -- lib:1.2.4
- C
  - package -- lib:^1.0.0
  - package-lock -- lib:1.3.0
```

The root above is supposed to have only one version of `lib`, any above `1.0.0`, not 3 versions based on 3 lock files.

> I just tested it

Didn't you test with `npm link`? Asking, because `npm link` doesn't simulate dependency resolving. It uses what you have installed in your local `@nextcloud/vue`.

susnux: The lockfile is only for development and testing, it is not even shipped with packages.
Meaning NPM will only consider `package.json` for when calculating the dependency tree.

Thus we need to specify the versions we support in  the `package.json` so npm can decide to duplicate a dependency or not.

susnux: Moreover this would not even be possible, imagine your dependency is using yarn or something other, they use different lock file formats. Only the `package.json` is standardized and considered for dependency lookup.

skjnldsv: We can also drop that PR, I fixed server manually :)

ShGKme: But fixing server doesn't fix apps. We can change `^10.7.2` to `~10.7.2` to avoid `10.8` and update it when the fix will be released, or use the string from `@susnux`, both will guarantee that everything works for the library users.

susnux: I would prefer to include the `1.8.1+` versions because otherwise users are bound to outdated library versions while there is no good reason to do so

skjnldsv: Feel free to edit this PR :)

susnux (approved):

PR #5302, opened 2024-02-24
feat: Add NcAppNavigationList

Pytal:
### ☑️ Resolves

- There are cases where having multiple `ul`s inside NcAppNavigation makes sense i.e. using NcAppNavigationCaption as a label for the next `ul`
- We add a new NcAppNavigationList component to be used in the default slot of NcAppNavigation
- Allow using NcAppNavigationCaption as a heading with a prop

### 🏁 Checklist


src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -66,7 +66,7 @@ emit('toggle-navigation', {
 			@keydown.esc="handleEsc">
 			<slot />
 			<!-- List for Navigation li-items -->
-			<ul class="app-navigation__list">
+			<ul v-if="hasListSlotItems" class="app-navigation__list">
```
ShGKme, line 69: Is `ul.app-navigation__list` different from `NcAppNavigationList`?

src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -132,6 +132,12 @@ export default {
 		}
 	},
 
+	computed: {
+		hasListSlotItems() {
+			return this.$scopedSlots.list?.()?.length > 0
+		},
+	},
```
ShGKme, line 139: `$scopedSlots` is not reactive and render function should only be used in rendering (another render function or template). Thought this `computed` is only used in template, it's safer to directly check for a list in the template with `v-if="$scopedSlots.list"`.

Is there a case there we really need to check the content of the slot, not only the slot presence?

ShGKme (approved):

JuliaKirschenheuter (approved):
great refactoring work!

src/components/NcAppNavigation/NcAppNavigation.vue
susnux: Also stumbled over this one, could we not just use `NcAppNavigationList` here instead?

src/components/NcAppNavigation/NcAppNavigation.vue
Pytal: > Is there a case there we really need to check the content of the slot, not only the slot presence?

Nope, fixed!

src/components/NcAppNavigation/NcAppNavigation.vue
Pytal: > Is `ul.app-navigation__list` different from `NcAppNavigationList`?

Yes slightly. The difference is `ul.app_navigation__list` always fills the height that is not already taken up by its siblings inside `nav.app-navigation__content` but when we have separate lists inside the nav we want don't want that e.g.

image

so we adjust the styles in NcAppNavigationList to keep the styling we want:

image

The `.app_navigation__list` styles also target `.app_navigation__content > ul` which we should keep to not break anything, so we also increase the specificity to have the styles apply correctly

PR #5304, opened 2024-02-26
fix(NcPopover): remove invalid aria-describedby

ShGKme:
## Resolves

When the popover is shown, `floating-vue` mutates the root elements of the trigger adding `data-popper-shown` and `aria-describedby` attributes.

`aria-describedby` is not correct here. All the popover content should not describe the trigger button.

See: https://github.com/Akryum/floating-vue/blob/8d4f7125aae0e3ea00ba4093d6d2001ab15058f1/packages/floating-vue/src/components/Popper.ts#L734

Unfortunately, because `floating-vue` directly mutates the button via DOM API, it is not possible to remove the attribute in a Vue way without the same DOM mutation...

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist


src/components/NcPopover/NcPopover.vue
```diff
@@ -368,14 +388,16 @@ export default {
...
+
 			this.$nextTick(() => {
+				/**
+				 * Triggered after the tooltip was visually displayed.
+				 *
+				 * This is different from the 'show' and 'apply-show' which
+				 * run earlier than this where there is no guarantee that the
+				 * tooltip is already visible and in the DOM.
+				 */
 				this.$emit('after-show')
```
ShGKme, line 401: Moved the comment so it works as event documentation for styleguidist.

src/components/NcPopover/NcPopover.vue
```diff
@@ -279,23 +279,43 @@ export default {
...
+		 * Remove incorrect aria-describedby attribute from the trigger.
+		 * @see https://github.com/Akryum/floating-vue/blob/8d4f7125aae0e3ea00ba4093d6d2001ab15058f1/packages/floating-vue/src/components/Popper.ts#L734
+		 */
+		removeFloatingVueAriaDescribedBy() {
+			// When the popover is shown, floating-vue mutates the root elements of the trigger adding data-popper-shown and incorrect aria-describedby attributes.
+			const triggerContainer = this.getPopoverTriggerContainerElement()
+			const triggerElements = triggerContainer.querySelectorAll('[data-popper-shown]')
+			for (const el of triggerElements) {
+				el.removeAttribute('aria-describedby')
+			}
```
ShGKme, line 300: In Vue 2 `triggerElements` is always only one root element, but in Vue 3 there could be multiple nodes. Using `querySelectorAll` for Vue 3 compatibility.

JuliaKirschenheuter (approved):
Great solution which is a bit hacky but prevents a new fork.

susnux (approved):
seems to work fine

emoral435 (approved):

PR #5305, opened 2024-02-26
fix(NcActions): put in order tab and arrow navigation

ShGKme:
### ☑️ Resolves

1. <kbd>Arrow</kbd> and <kbd>Tab</kbd> keyboard navigation should not be used at the same time
   - Either menu is an atomic UI element, and <kbd>Arrow</kbd> select menu items, <kbd>Tab</kbd> moves to the next UI element
   - Or the menu is an expanded list of UI elements
2. When `actions` are used for navigation, it should work as just an expanded list with classic "native-like" <kbd>Tab</kbd> navigation without focus trap
   - <kbd>Tab</kbd> on the last element closes the list and moves focus to the trigger
   - <kbd>SHIFT + Tab</kbd> on the first element - the same

Also, I've put all the config that changes the behavior of the component in one place, so it's clearer what can be different and in which case.

### 🖼️ Screenshots

Before Navigation worked as a Menu:
- Arrows - navigate between items
- Tab - navigate to the next UI element on the page

After Navigation works as an expanded block with native like behavior
- No arrow navigation
- Tab and Shift+Tab navigate back and forward
- No focus trap, before the first and after the last element actions are closed, focus is moved to the trigger button

nc-actions-navigation-tab

### 🏁 Checklist

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -571,7 +571,8 @@ export default {
 <template>
 	<div>
 		<h2>Application menu</h2>
-		Uses buttons, button groups, links and router links, separators, text. May have checkboxes and radio buttons.
+		<p>Has buttons, button groups, links and router links, separators, texts. May have checkboxes and radio buttons. Separator can be used to make groups of radio buttons as well.</p>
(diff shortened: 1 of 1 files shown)


JuliaKirschenheuter (approved):
great work!

PR #5306, opened 2024-02-26
fix(NcPopover): make popupRole non-required

ShGKme:
### ☑️ Resolves

- A small regression from https://github.com/nextcloud-libraries/nextcloud-vue/pull/5300
- `popupRole` was made non-required in `NcPopover` but then it should also be non-required in `NcPopoverTriggerProvider`

image

- And because `aria-haspopup` is not required now, it should not be checked when checking for required attrs

image

### 🏁 Checklist


JuliaKirschenheuter (approved):

ShGKme:
Found one more issue

susnux (approved):

PR #5313, opened 2024-02-26
fix(NcActions): focus first checked item on open

ShGKme:
### ☑️ Resolves

- Brings back https://github.com/nextcloud-libraries/nextcloud-vue/pull/4678 after changes in the component
- `aria-selected` should be `aria-checked`
- `aria-checked` should not only present, but be `'true'`
- `aria-checked` must be on the focusable element itself (otherwise there is an a11y issue)
- New possibility — check only radio buttons

IMO, this should happen only when ALL the buttons are one radio button group. But it is required even for heading in Text, where we have "Show outline" button together with heading radio buttons.

The solution is still not reliable, but I don't see any other simple solution... 

### 🖼️ Screenshots

focus-first-action

### 🏁 Checklist

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -1504,11 +1504,13 @@ export default {
 		focusFirstAction(event) {
 			if (this.opened) {
 				this.preventIfEvent(event)
-				// In case a button is considered aria-selected we will use this one as a initial focus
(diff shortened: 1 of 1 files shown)


Pytal (approved):

emoral435 (approved):
Nice! :)

PR #5316, opened 2024-02-27
feat: List team resources in related resources panel

juliushaertl:
### ☑️ Resolves

- Resolve https://github.com/nextcloud/related_resources/issues/360

### 🖼️ Screenshots
Screenshot 2024-03-01 at 15 09 11

Previous screen recording

### 🏁 Checklist

### DIFF
--- a/src/components/NcRelatedResourcesPanel/NcTeamResources.vue
+++ b/src/components/NcRelatedResourcesPanel/NcTeamResources.vue
@@ -1,0 +1,254 @@
+<template>
+	<div v-if="appEnabled && isVisible" class="team-resources">
+		<h5 class="team-resources__header">
+			{{ t('Related team resources') }}
+		</h5>
+		<details v-for="team in teamResources"
+			:key="team.teamId"
+			name="Team resources"
+			class="related-team"
+			:open="open(team.teamId)"
+			@toggle="(event) => toggleOpen(team.teamId, event.target.open)">
+			<summary class="related-team__header">
+				<h5 class="related-team__name">
+					<AccountGroup :size="20" />
+					{{ team.displayName }}
+				</h5>
+				<NcButton type="tertiary"
+					:href="team.link"
+					:aria-label="t('View team')"
+					:title="t('View team')">
+					<template #icon>
+						<OpenInNew :size="20" />
+					</template>
+				</NcButton>
+
+				<ChevronUp v-if="open(team.teamId)"
+					:size="20" />
+				<ChevronDown v-else
+					:size="20" />
+			</summary>
+
+			<div>
+				<div v-for="provider in teamProviders(team.teamId)"
+					:key="provider.id"
+					class="related-team-provider">
+					<h6 v-if="provider.resources.length > 0">
+						{{ provider.name }}
+					</h6>
+					<ul>
+						<li v-for="resource in provider.resources" :key="resource.url" class="related-team-resource">
+							<a :href="resource.url" class="related-team-resource__link">
+								<span v-if="resource.iconEmoji" class="resource__icon">
+									{{ resource.iconEmoji }}
+								</span>
+								<NcIconSvgWrapper v-else-if="resource.iconSvg"
+									class="resource__icon"
+									:svg="resource.iconSvg"
+									:size="20" />
+								<span v-else-if="resource.iconUrl" class="resource__icon">
+									<img :src="resource.iconURL" alt="">
+								</span>
+								<span class="resource__name">
+									{{ resource.label }}
+								</span>
+							</a>
+						</li>
+					</ul>
+				</div>
+			</div>
+		</details>
+	</div>
+</template>
+
+<script>
+import axios from '@nextcloud/axios'
+import { generateOcsUrl } from '@nextcloud/router'
+import AccountGroup from 'vue-material-design-icons/AccountGroup.vue'
+import ChevronDown from 'vue-material-design-icons/ChevronDown.vue'
+import ChevronUp from 'vue-material-design-icons/ChevronUp.vue'
+import OpenInNew from 'vue-material-design-icons/OpenInNew.vue'
+import NcButton from '../NcButton/NcButton.vue'
+import NcIconSvgWrapper from '../NcIconSvgWrapper/NcIconSvgWrapper.vue'
+
+import { t } from '../../l10n.js'
+
+export default {
+	name: 'NcTeamResources',
+
+	components: {
+		AccountGroup,
+		ChevronDown,
+		ChevronUp,
+		OpenInNew,
+		NcButton,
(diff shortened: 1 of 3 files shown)


GretaD (approved):
code looks good

src/components/NcRelatedResourcesPanel/NcTeamResources.vue
```diff
@@ -0,0 +1,293 @@
...
+	props: {
+		providerId: {
+			type: String,
+			default: null,
+		},
+		itemId: {
+			type: [String, Number],
+			default: null,
+		},
+		resourceType: {
```
mejo-, line 99: All props starting from here (`resourceType`, `limit`, `fileInfo`, `header`, `headerOther`, `description`, `primary`) seem unused, so probably leftovers from copying.

src/components/NcRelatedResourcesPanel/NcTeamResources.vue
```diff
@@ -0,0 +1,293 @@
...
+		primary: {
+			type: Boolean,
+			default: false,
+		},
+	},
+
+	data() {
+		return {
+			appEnabled: OC?.appswebroots?.circles !== undefined && (OC.config.version.split('.')[0] ?? 0) >= 29,
+			loading: false,
```
mejo-, line 146: `this.loading` seems to be never set to `true`. It's initiated with `false` and will be set to `false` at the end of `fetchTeamResources()`, but it doesn't seem to be set to `true` in between. It's only used in computed `isVisible`, which seems superfluous given that evaluating `this.error` and `this.teamResources` should be sufficient.

src/components/NcRelatedResourcesPanel/NcTeamResources.vue
```diff
@@ -0,0 +1,293 @@
...
+						<li v-for="resource in provider.resources" :key="resource.url" class="related-team-resource">
+							<a :href="resource.url" class="related-team-resource__link">
+								<span v-if="resource.iconEmoji" class="resource__icon">
+									{{ resource.iconEmoji }}
+								</span>
+								<NcIconSvgWrapper v-else-if="resource.iconSvg"
+									class="resource__icon"
+									:svg="resource.iconSvg"
+									:size="20" />
+								<span v-if="resource.iconUrl" class="resource__icon">
```
mejo-, line 47: I'm not exactly sure I get the logic here: If `resource.iconEmoji` is set, it will be displayed, otherwise `resource.iconSvg` will be displayed if set. But `resource.iconUrl` will always be displayed. Shouldn't the latter be a `v-else-if` as well?

src/components/NcRelatedResourcesPanel/NcTeamResources.vue
```diff
@@ -0,0 +1,293 @@
...
+			loading: false,
+			error: null,
+			teamResources: null,
+			teamOpen: null,
+		}
+	},
+
+	computed: {
+		isVisible() {
+			return !this.loading && this.error === null && this.teamResources?.length > 0
```
mejo-, line 155: Why do we need to check for `this.loading` and `this.error` here? If `this.teamResources` is set and has elements, they should be displayed, no?

Do we expect dynamic changing of `this.teamResources` (e.g. changing contexts)? In this case, hiding the resources while loading might make sense to not display outdated resources. And in this case evaluating `this.error` might make sense as well, to make sure that outdated don't get displayed. But I'd rather say we should reset `this.teamResources = null` in case of errors (line 208).

src/components/NcRelatedResourcesPanel/NcTeamResources.vue
```diff
@@ -0,0 +1,293 @@
...
+	methods: {
+		t,
+		async fetchTeamResources() {
+			try {
+				const response = await axios.get(generateOcsUrl(`/teams/resources/${this.providerId}/${this.itemId}`))
+				this.error = null
+				this.teamResources = response.data.ocs.data.teams
+				this.teamOpen = this.teamResources[0]?.teamId
+			} catch (e) {
+				this.error = e
```
mejo-, line 208: Maybe rather reset `this.teamResources = null` here and get rid of `this.error`? It's not displayed anywhere in UI anyway.

mejo- (changes requested):
Works well already, but I have some comments :blush:

src/components/NcRelatedResourcesPanel/NcTeamResources.vue
```diff
@@ -0,0 +1,293 @@
...
+		</div>
+		<div v-for="team in teamResources"
+			:key="team.teamId"
+			class="related-team"
+			:class="{ 'related-team__open': open(team.teamId) }">
+			<h5 class="related-team__header">
+				<a class="related-team__link" @click="toggleOpen(team.teamId)">
+					<AccountGroup :size="20" />
+					{{ team.displayName }}
+				</a>
```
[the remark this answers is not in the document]
juliushaertl: Thanks, I adjusted that accordingly. Makes a lot of sense to use that instead.

src/components/NcRelatedResourcesPanel/NcTeamResources.vue
```diff
@@ -0,0 +1,293 @@
...
+				<NcActions>
+					<NcActionLink :href="team.link" :label="t('View team')">
+						<template #icon>
+							<OpenInNew :size="20" />
+						</template>
+					</NcActionLink>
+				</NcActions>
```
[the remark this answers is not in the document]
juliushaertl: We were talking about having further actions in the future, but moved to an NcButton for now

src/components/NcRelatedResourcesPanel/NcTeamResources.vue
juliushaertl: Yes, that should indeed be a v-else-if, fixed

src/components/NcRelatedResourcesPanel/NcTeamResources.vue
juliushaertl: Right, I kept the loading flag and removed the error one as we just hide the component then anyways

src/components/NcRelatedResourcesPanel/NcTeamResources.vue
```diff
@@ -0,0 +1,293 @@
...
+		},
+	},
+}
+</script>
+
+<style lang="scss" scoped>
+h5 {
+	font-weight: bold;
+	margin-bottom: 6px;
+}
```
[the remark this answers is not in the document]
juliushaertl: Thanks, wasn't aware of that :+1:

mejo- (approved):

skjnldsv:
@juliushaertl  @mejo- is it intended that the component is exported?
The documentation is non-existent :(

image

PR #5342, opened 2024-03-04
fix(NcListItem): open _blank link by Enter and allow prevent

pulsejet:
I don't understand what that line is intended to do.
1. It prevents `target=_blank` above from working at all
2. You can't call `preventDefault` on the event anymore

~~This just makes zero sense to me.~~

My guess is, this is for the keyboard event? (even then, enter always triggers click if `href` is present, which should always be)

~Regardless, switched that to a explicit trigger.~

### DIFF
--- a/src/components/NcListItem/NcListItem.vue
+++ b/src/components/NcListItem/NcListItem.vue
@@ -345,7 +345,6 @@
 					@focus="showActions"
 					@focusout="handleBlur"
 					@click="onClick($event, navigate, routerLinkHref)"
-					@keydown.enter="onClick($event, navigate, routerLinkHref)"
 					@keydown.esc="hideActions">
 					<!-- @slot This slot is used for the NcAvatar or icon, the content of this slot must not be interactive -->
 					<slot name="icon" />
@@ -630,8 +629,6 @@ export default {
 			if (routerLinkHref) {
 				navigate?.(event)
 				event.preventDefault()
-			} else {
-				window.location = this.href
 			}
 		},
 


pulsejet:
Hello?

src/components/NcListItem/NcListItem.vue
```diff
@@ -345,7 +345,7 @@
 					@focus="showActions"
 					@focusout="handleBlur"
 					@click="onClick($event, navigate, routerLinkHref)"
-					@keydown.enter="onClick($event, navigate, routerLinkHref)"
+					@keydown.enter.prevent="$event.target.click()"
```
ShGKme, line 348: Makes sense, but programmatically clicking on a link is blocked by some browsers.

pulsejet: Like?

ShGKme: Enter (and click) should work natively for non-router links

ShGKme: > Like?

Firefox

pulsejet: > Firefox

Are you sure? This works fine for me on ff: https://codepen.io/pulsejet/pen/MWRKKoK?editors=1111

pulsejet: > > Like?
> 
> Firefox

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/click#browser_compatibility

ShGKme: > Are you sure? This works fine for me on ff: https://codepen.io/pulsejet/pen/MWRKKoK?editors=1111

Sorry for the inaccuracy. Not click itself, but opening a page (pop-up) is blocked.

In your example, there is no actual link — it doesn't have non-empty `href` and doesn't open any page. Add href, `target="_blank"`, remove `event.preventDefault()`. It is supposed to open a page by Enter like on click, but it won't.

pulsejet: Okay, then we throw out that handler. Just a conlick should be fine
https://codepen.io/pulsejet/pen/KKYVMob?editors=1111

pulsejet: > target="_blank"

Just a note, I don't believe this is working right now anyway since it sets location directly.

skjnldsv:
> Hello?

We're in the rush of the 29 release. Things are a bit busy at the moment, sorry 😔

ShGKme (approved):
Thanks!

Works fine with a router link and a native link, with `_blank` and `_self`, by click and by enter/space, preventable with `preventDefault()`.

src/components/NcListItem/NcListItem.vue
ShGKme: > Just a note, I don't believe this is working right now anyway since it sets location directly.

For me, it doesn't work completely correctly. It makes both - navigation on the current page (via window.location) and opening a new tab.

pulsejet: Huh? Where the window.location (after this patch)

ShGKme: I mean, in `master`. In this PR, everything seems to work in every case.

susnux (approved):

pulsejet:
> @pulsejet Please report whether #42891 could have solved the issue.

I've no idea since I can't reproduce the bug anyway. But that seems like a reasonable fix.

PR #5347, opened 2024-03-05
fix(NcAppNavigation): Wrap app navigation default slot with scrollable container

Pytal:
- Fix https://github.com/nextcloud/server/pull/43804#issuecomment-1977005782

### ☑️ Resolves

- To fix the unscrollable content in NcAppNavigationList we wrap the entire default slot with a container
- The `div.app-navigation__body` now fills the height with `height: 100%`
  - Remove the redundant `&:nth-last-of-type(2)` fill styles
- NcAppNavigationList now has `height: fit-content` in all contexts and is itself not scrollable now that scrolling is delegated to the container similar to the `ul.app-navigation__list` list slot behaviour

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppNavigation/NcAppNavigation.vue
+++ b/src/components/NcAppNavigation/NcAppNavigation.vue
@@ -64,7 +64,11 @@ emit('toggle-navigation', {
 			class="app-navigation__content"
 			:inert="!open || undefined"
 			@keydown.esc="handleEsc">
-			<slot />
+			<div v-if="$scopedSlots.default" class="app-navigation__body">
+				<!-- @slot Content within the nav, do NOT add NcModal/NcDialog inside this slot -->
+				<slot />
+			</div>
+
 			<!-- List for Navigation li-items -->
 			<ul v-if="$scopedSlots.list" class="app-navigation__list">
 				<slot name="list" />
@@ -253,6 +257,17 @@ export default {
 	&--close {
 		transform: translateX(-100%);
 		position: absolute;
+	}
+
+	&__body {
+		position: relative;
+		height: 100%;
+		width: 100%;
+		overflow-x: hidden;
+		overflow-y: auto;
+		box-sizing: border-box;
+		display: flex;
+		flex-direction: column;
 	}
 
 	&__content > ul,
--- a/src/components/NcAppNavigationList/NcAppNavigationList.vue
+++ b/src/components/NcAppNavigationList/NcAppNavigationList.vue
@@ -63,17 +63,11 @@ ul.app-navigation-list { // Increase specificity over NcAppNavigation styles
 	position: relative;
 	height: fit-content;
 	width: 100%;
-	overflow: unset;
+	overflow: visible;
 	box-sizing: border-box;
 	display: flex;
 	flex-direction: column;
 	gap: var(--default-grid-baseline, 4px);
 	padding: var(--app-navigation-padding);
-
-	&:nth-last-of-type(2) {
-		// Fill remaining space before NcAppNavigation footer
-		height: 100%;
-		overflow: auto;
-	}
 }
 </style>


src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -64,7 +64,10 @@ emit('toggle-navigation', {
 			class="app-navigation__content"
 			:inert="!open || undefined"
 			@keydown.esc="handleEsc">
-			<slot />
+			<div v-if="$scopedSlots.default" class="app-navigation__body">
+				<slot />
+			</div>
+
```
ShGKme, line 70: With the default slot, it is always more complicated. If not `<template #default>` is used, then the slot is passed. Even it is actually empty with empty strings from the template.

image

ShGKme: `$slots` will work here in Vue 2 as well as double-render with `isSlotPopupated()` in Vue 3

ShGKme: And because of flex, it is stretched by default:

image

Pytal: > With the default slot, it is always more complicated. If short syntax (not `<template #default>`) is used, then the slot is passed. Even it is actually empty with empty strings from the template.

In this case it seems to not be an empty string but rather a modal https://github.com/nextcloud/server/blob/c18ffe0cad13dc5903cbb08ffbdb787f01db5e35/apps/files/src/views/Navigation.vue#L71-L73 that is causing this 🤔

Not sure how we could resolve this on nc/vue side, any ideas?

> And because of flex, it is stretched by default:

How can I repro this?

ShGKme: > In this case it seems to not be an empty string but rather a modal https://github.com/nextcloud/server/blob/c18ffe0cad13dc5903cbb08ffbdb787f01db5e35/apps/files/src/views/Navigation.vue#L71-L73 that is causing this 🤔

Yes, but the same issue would be in other cases. Anything in components's content, even if it actually is not rendered, cause this issue. With the default slot it's always more annoying...

> Not sure how we could resolve this on nc/vue side, any ideas?

Let me have a look on the issue...

> How can I repro this?

I though that was on this branch + your branch on server for the same issue. Be let me double check.

ShGKme: > > And because of flex, it is stretched by default:
> 
> How can I repro this?

Ok, it seems to work fine now on your branch. But I'm not sure if it is not a regression for other usages.

JuliaKirschenheuter (approved):
seems to good ;) thank you!

Pytal:
Pushed a fixup to document NOT putting any NcModal/NcDialog inside the NcAppNavigation default slot

Only one place https://github.com/nextcloud/server/blob/c18ffe0cad13dc5903cbb08ffbdb787f01db5e35/apps/files/src/views/Navigation.vue#L71-L73 would need to be adjusted

susnux (approved):

ShGKme:
Should be also tested with Talk

ShGKme:
It breaks Talk
image

ShGKme:
And mail. So, any app that has both default and list slots.

image

ShGKme:
> I think then we should add a new slot for this specific use-case instead of changing default slot behaviour, what do you think @ShGKme?

I'd not add a new slot here. A new slot means — +4 cases to support in the component (combination with other slots) and a new interface — so no revert without a breaking change.



> 💀💀 we'll need to find a solution without using default slot then

I think, using the default slot is fine. The problem is probably mixing some styles between `#list` slot in `<NcAppNavigation>` and `<NcAppNavigationList>` which can be used in both the default slot and list.

We can try to keep `#list` slot styles as it is, removing them from the component itself, and then fix  layout on the User settings page because it has it's own layout.

PR #5353, opened 2024-03-06
fix(NcRichContenteditable): capture and stop `Esc` keydown event when closing tribute

### LINKED ISSUE #5352: [NcRichContenteditable] Pressing ESC not only closes the link picker but also unfocuses the input
A part of https://github.com/nextcloud-libraries/nextcloud-vue/issues/5351

Antreesy:
### ☑️ Resolves

- Fix #5352

### 🖼️ Screenshots
Tested against Talk (we handle Esc keyup depending whether tribute state is open or not 

Before (event isn't emitted -> blur input)

[Screencast from 06.03.2024 12:57:48.webm](https://github.com/nextcloud-libraries/nextcloud-vue/assets/93392545/b388d058-63a8-4931-a902-509675c8fe31)

After (Esc doesn't blur the input)

[Screencast from 06.03.2024 12:53:54.webm](https://github.com/nextcloud-libraries/nextcloud-vue/assets/93392545/7bff85ed-83e2-4571-af53-28398e72829d)

### 🏁 Checklist

### DIFF
--- a/src/components/NcRichContenteditable/NcRichContenteditable.vue
+++ b/src/components/NcRichContenteditable/NcRichContenteditable.vue
@@ -242,6 +242,7 @@ export default {
 			@compositionstart="isComposing = true"
 			@compositionend="isComposing = false"
 			@keydown.delete="onDelete"
+			@keydown.esc.capture="onKeyEsc"
 			@keydown.enter.exact="onEnter"
 			@keydown.ctrl.enter.exact.stop.prevent="onCtrlEnter"
 			@paste="onPaste"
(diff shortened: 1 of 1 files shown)


ShGKme (commented):
I'd not add tribute-related things to the component interface. If we want to add the interface to handle autocomplete state, it could be something like `autocomplite-toggle`, so we don't rely on what library we use internally.

And though it fixes the issue in Talk, it doesn't fix the initial issue. 

These events were needed only to handle `Escape` from `NcRichContenteditable` itself ignoring it from autocomplete. A better alternative is to not trigger `keydown` for `Escape` from open autocomplite. Then it works in Talk without the manual handling the autocomplite state.

Antreesy:
> A better alternative is to not trigger `keydown` for `Escape` from open autocomplite. Then it works in Talk without the manual handling the autocomplite state.

Seems to work with 'capturing' a keydown event and check tribute state, then close it manually

ShGKme (approved):

nickvergessen (approved):

PR #5359, opened 2024-03-06
fix(richEditor): reference server-scoped regular expression to match complex user mentions + fix(richText): when parse HTML content collapse consecutive spaces

Antreesy:
### ☑️ Resolves

- Ref https://github.com/nextcloud/server/blob/master/lib/private/Comments/Comment.php#L223
  - use the same regex as backend code uses for parsing comments / messages
- Fix https://github.com/nextcloud-libraries/nextcloud-vue/issues/4127
  - In `parseContent` when HTML is transformed into text, content collapses whitespace

### 🖼️ Screenshots

Before | After
---|---
image | image

### 🚧 Tasks

- [ ] Code review

### 🏁 Checklist

### DIFF
--- a/src/mixins/richEditor/index.js
+++ b/src/mixins/richEditor/index.js
@@ -27,12 +27,18 @@ import escapeHtml from 'escape-html'
 import stripTags from 'striptags'
 import Vue from 'vue'
 
-// Beginning or whitespace. Non-capturing group
-const MENTION_START = '(?:^|\\s)'
-// Anything that is not text or end-of-line. Non-capturing group
-const MENTION_END = '(?:[^a-z]|$)'
-export const USERID_REGEX = new RegExp(`${MENTION_START}(@[a-zA-Z0-9_.@\\-']+)(${MENTION_END})`, 'gi')
-export const USERID_REGEX_WITH_SPACE = new RegExp(`${MENTION_START}(@&quot;[a-zA-Z0-9 _/.@\\-']+&quot;)(${MENTION_END})`, 'gi')
+// Referenced from public function getMentions(): https://github.com/nextcloud/server/blob/master/lib/private/Comments/Comment.php
+// Beginning or whitespace. Non-capturing group within word boundary
+const MENTION_START = /\B(?<![^a-z0-9_\-@.'\s])/.source
+// Capturing groups like: @user-id, @"guest/abc16def", @"federated_user/user-id", @"user-id with space"
+const MENTION_SIMPLE = /(@[a-z0-9_\-@.']+)/.source
+const MENTION_GUEST = /@&quot;guest\/[a-f0-9]+&quot;/.source
+const MENTION_PREFIXED = /@&quot;(?:federated_)?(?:group|team|user){1}\/[a-z0-9_\-@.' /:]+&quot;/.source
+const MENTION_WITH_SPACE = /@&quot;[a-z0-9_\-@.' ]+&quot;/.source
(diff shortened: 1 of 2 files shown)


src/mixins/richEditor/index.js
```diff
@@ -145,9 +151,10 @@ export default {
 			wrapper.appendChild(mount)
 			document.body.appendChild(wrapper)
 
-			// Mount and get raw html
+			// Mount and get raw HTML. As HTML replaces new lines, empty lines and consecutive whitespaces
+			// with a single whitespace, which parseContent() can't trim, we do it manually
 			Item.$mount(mount)
-			const renderedHtml = wrapper.innerHTML
+			const renderedHtml = wrapper.innerHTML.replace(/>\s+</g, '><')
```
ShGKme, line 157: It seems that the problem is not in the component rendering, but in `parseContent` where HTML is transformed to text (though it probably doesn't hurt, for component rendering to HTML, it is valid to have these spaces).

juliushaertl (approved):

src/mixins/richEditor/index.js
Antreesy: When HTML is transformed, `stripTags` omits everything except for `@userId` and these spaces laying between tags. As user id could also contain spaces, I'm not sure if it would be better to use `replace` here, or match `parseContent` result with a regex again

ShGKme: > When HTML is transformed, `stripTags` omits everything except for `@userId` and these spaces laying between tags.

Yes, removing repeating spaces between tags should be before stripping tags. Same as here — for HTML. The problem in `NcRichContenteditable` is that then it takes value from content, it takes it from HTML content, and getting text content from HTML is invalid, 

`parseContent` works incorrectly — it takes not correct text content of HTML, not what native `textContent` would take.

While `renderComponent` works correctly, the returned HTML content is valid as Vue rendering is valid.

For this issue, it works both ways. But we cannot guarantee that we will never have HTML in contenteditable not from `renderComponent` or that spaces in `renderComponent` would be needed in some other place.

> As user id could also contain spaces, I'm not sure if it would be better to use `replace` here

But it cannot contain `/>\s+</`. Otherwise, changing them here wouldn't be safe as well.

Antreesy: I couldn't find an easy solution for this atm, so will drop this commit in favor of merging user regex fix

ShGKme: What I'm proposing is to just move your solution from from `renderComponent` to `parseContent`.

In the beginning of `parseContent`:
```diff
- let text = content.replace(/<br>/gmi, '\n')
- text = text.replace(/&nbsp;/gmi, ' ')
+ let text = content.replace(/>\s+</g, '><')
+ text = text.replace(/<br>/gmi, '\n')
+ text = text.replace(/&nbsp;/gmi, ' ')
```

Because this is `parseContent` - the method that doesn't parse HTML correctly, not `renderComponent` that renders HTML incorrectly.

ShGKme: image

Antreesy: There's also a check in NcRichcontenteditable
`this.value.trim() === this.parsedContent(htmlOrText)`. I'm not sure, whether it should be also trimmed by that moment or not

ShGKme: I have only found check `this.value.trim() !== this.parseContent(html).trim()` in watcher for value. It updates internal value and content based on value update. But I don't see how it is related to rendering/parsing. Both value and content from html either have spaces in between or not. When it is supposed to be equal, value is got from the same `this.parseContent` in `updateValue`.

susnux (approved):

PR #5389, opened 2024-03-11
fix(NcAppSidebar): animation glitch on toggle

RayBB:
Signed-off-by: RayBB <RayBB@users.noreply.github.com>

### ☑️ Resolves

- Fix https://github.com/nextcloud/server/issues/43969

~~Apologies, I cannot test this outside of the browser devtools because of https://github.com/nextcloud-libraries/nextcloud-vue/issues/5339 but it is a simple enough change I hope someone else can help test/verify until I can get a local instance working. Thanks!~~

Edit: Finally I was able to test in a local instance and it worked great, looks the same as the video I put below.

### 🖼️ Screenshots

The "after" video shows before/after

🏚️ Before | 🏡 After
---|---
image | https://github.com/nextcloud/server/assets/921217/0303ae9c-a791-4ed1-98c1-0755f81aa259

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppNavigation/NcAppNavigation.vue
+++ b/src/components/NcAppNavigation/NcAppNavigation.vue
@@ -237,7 +237,8 @@ export default {
 	transition: transform var(--animation-quick), margin var(--animation-quick);
 	width: $navigation-width;
 	// Left toggle button padding + toggle button + right padding from NcAppContent
-	max-width: calc(100vw - (var(--app-navigation-padding) + var(--default-clickable-area) + var(--default-grid-baseline)));
+	--app-navigation-max-width: calc(100vw - (var(--app-navigation-padding) + var(--default-clickable-area) + var(--default-grid-baseline)));
+	max-width: var(--app-navigation-max-width);
 	position: relative;
 	top: 0;
 	left: 0;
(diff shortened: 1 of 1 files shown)


ShGKme (approved):
Very cool, thanks!

Slowed down to better see the difference: 
Before | After
---|---
Before | after

This could be a little bit slower than before in performance, because it causes animation is width change of the main content, triggering resize observers. But should be fine, `NcAppSidebar` does similar change.

If this accurse bad for performance, we can make it absolute during animation:

Alternative

skjnldsv:
Please also check this works on very narrow screens like mobile at 320/400px

src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -257,8 +257,7 @@ export default {
 	backdrop-filter: var(--filter-background-blur, none);
 
 	&--close {
-		transform: translateX(-100%);
-		position: absolute;
+		margin-left: -$navigation-width;
```
ShGKme, line 260: On mobile width can be smaller, for example, ~260px on 320px window size.

So margin here should be set the same as on navigation width - using `max` between 300px and `100vw - (var(--app-navigation-padding) ...` (see line 240).

ShGKme (changes requested):

RayBB:
@ShGKme I updated to fix this for the mobile case.

<details>
  <summary>What master looks like now on small screens</summary>
  
  

https://github.com/nextcloud-libraries/nextcloud-vue/assets/921217/1aff36f7-26f4-4e79-9e04-3814f469ea7b


  
</details> 


<details>
  <summary>And here is what it looks like desktop/mobile with my current changes.</summary>
  
  

https://github.com/nextcloud-libraries/nextcloud-vue/assets/921217/dd2b0f22-a842-42c9-a571-296ff921b3a7


  
</details> 

Please let me know if I should change anything else. Cleaning up commits now.

ShGKme (approved):
Great

susnux (approved):

PR #5394, opened 2024-03-11
fix(NcReferenceWidget): Only use composables in `setup` context

### LINKED ISSUE #5393: NcDialog, NcReferenceWidget generate Vue warnings for the vueuse/core methods
Warning:
image

UPD: Looking as this also happened with NcReferenceWidget, I'd blame `useResizeObserver` => `useMounted` hook:
https://github.com/vueuse/vueuse/blob/9f10ad06a4fd2b4e6325dc72a3d1e36a62b6c254/packages/core/useResizeObserver/index.ts
https://github.com/vueuse/vueuse/blob/9f10ad06a4fd2b4e6325dc72a3d1e36a62b6c254/packages/core/useSupported/index.ts
https://github.com/vueuse/vueuse/blob/9f10ad06a4fd2b4e6325dc72a3d1e36a62b6c254/packages/core/useMounted/index.ts#L12-L17

But tracked it back to our library: https://github.com/nextcloud-libraries/nextcloud-vue/pull/5298

susnux:
### ☑️ Resolves

- Fix #5393 

### 🏁 Checklist


Antreesy (approved):
Good addition, tried the same an hour ago. Unfortunately, it doesn't help to get rid of warnings in the console, as NcDialog is using it in the `setup` already

Not sure if that's an upstream issue with vueuse and Vue 2, or just the way we use it in Talk...

Antreesy:
Could you also add a fix for the destroying method?
image

Should be this line:
https://github.com/nextcloud-libraries/nextcloud-vue/blob/f72035078469e18c44f57c09bd8803b63efa3b54/src/functions/reference/widgets.ts#L90

```js
window._vue_richtext_widgets[richObjectType].onDestroy?.(el)
```

ShGKme (approved):

ShGKme:
The source issue is probably in the upstream, but I'd prefer to always use composables only in setup.

At least `onMounted` doesn't work if it is called when the component was already mounted (it doesn't show any error, because there us setup context in lifecycle hooks, but the hook itself is not called when it happened already).

PR #5407, opened 2024-03-14
fix(NcAvatar): handle regex error on initials match

kyteinsky:
### ☑️ Resolves

- Fix #…

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | Screenshot from 2024-03-14 23-22-32

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/NcAvatar/NcAvatar.vue
+++ b/src/components/NcAvatar/NcAvatar.vue
@@ -472,7 +472,7 @@ export default {
 			if (this.showInitials) {
 				const user = this.userIdentifier.trim()
 				if (user === '') {
-					return '?'
+					return initials
 				}
 
 				/**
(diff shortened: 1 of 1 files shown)


kyteinsky:
hey, would it make sense to allow a few special symbols like `#` and `&`? They could be useful for some cases. One example could be the Slack integration that used to use a `#` symbol in the display name to create an icon for channels.

ShGKme (approved):

susnux (approved):

susnux:
would it make sense to allow a few special symbols like # and &?

I think they might bring back issues like https://github.com/nextcloud-libraries/nextcloud-vue/issues/2044
Maybe those integrations can use emojis like talk or similar?

PR #5418, opened 2024-03-17
feat: add v-model support for input-like components

kyteinsky:
Can we get this in nc/vue 8.x or does it not matter since it was added in vue9 recently (#4994)? Would be a nice QOL improvement still.

### 🏁 Checklist

### DIFF
--- a/src/components/NcTextArea/NcTextArea.vue
+++ b/src/components/NcTextArea/NcTextArea.vue
@@ -123,6 +123,11 @@ export default {
 
 	inheritAttrs: false,
 
+	model: {
+		prop: 'value',
+		event: 'update:value',
+	},
+
 	props: {
 		/**
 		 * The value of the input field
--- a/src/components/NcTextArea/NcTextArea.vue
+++ b/src/components/NcTextArea/NcTextArea.vue
@@ -71,7 +71,6 @@ It extends and styles an HTMLTextAreaElement.
 				:id="computedId"
 				ref="input"
 				class="textarea__input"
-				:type="type"
 				:disabled="disabled"
 				:placeholder="computedPlaceholder"
 				:aria-describedby="ariaDescribedby"
--- a/src/components/NcTextArea/NcTextArea.vue
+++ b/src/components/NcTextArea/NcTextArea.vue
@@ -13,34 +13,57 @@ It extends and styles an HTMLTextAreaElement.
 <template>
 	<div class="wrapper">
 		<NcTextArea label="Text area"
+			:value.sync="text1"
 			placeholder="Placeholders are possible"
 			helper-text="This is a regular helper text." >
 		</NcTextArea>
 		<NcTextArea label="Success state"
+			v-model="text2"
 			:success="true">
 		</NcTextArea>
 		<NcTextArea label="Error state"
+			:value.sync="text3"
 			placeholder="Enter something valid"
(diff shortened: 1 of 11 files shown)


ShGKme (commented):
I think, that would be good to add `model` to every component that has `modelValue` in the `next` version.

Just to make further migration to v9 easier for library users.

All these components from the Breaking Changes list where it doesn't conflict with the current implementation. 

https://github.com/nextcloud-libraries/nextcloud-vue/blob/next/CHANGELOG.md

ShGKme:
Hey @kyteinsky!

 What do you think about adding `v-model` to all these components?

kyteinsky:
Hello @ShGKme , sorry I didn't find time for it recently. Yeah, looks like a sound reasoning.

Is it alright if I do that the next week?

kyteinsky:
Some components did not need modification and already work with v-model:
1. `NcColorPicker`
2. `NcDateTimePicker`
3. `NcDateTimePickerNative`
4. `NcSelect`
5. `NcSelectTags`
6. `NcSettingsSelectGroup`
7. `NcTimezonePicker`

`NcActionRadio` does not make use of v-model in any useful way so I left it out. Enlighten me if I'm wrong here.

ShGKme:
> Some components did not need modification and already work with v-model:

Confirm, all of them have the default model with `:value + @input`

> `NcActionRadio` does not make use of v-model in any useful way so I left it out. Enlighten me if I'm wrong here.

It is the same as `NcActionCheckbox`, it has `checked` with `update:checked` event. Let's add event here as well.

src/components/NcTextField/NcTextField.vue
```diff
@@ -51,8 +51,10 @@ and `minlength`.
 			@trailing-button-click="clearText">
 		</NcTextField>
 		<NcTextField label="Disabled"
+			value="''"
```
susnux, line 54: Is the content of the literal string `''` expected, or did you mean:
```suggestion
			:value="''"
```

Antreesy: What you wrote, I made a typo 😬

src/components/NcTextField/NcTextField.vue
```diff
@@ -51,8 +51,10 @@ and `minlength`.
...
 			:disabled="true" />
 		<NcTextField label="Disabled + Success"
+			value="''"
```
susnux, line 57: Same here

src/components/NcTextField/NcTextField.vue
Antreesy: replaced with existing data attrs

susnux (approved):

src/components/NcTextArea/NcTextArea.vue
```diff
@@ -71,7 +71,6 @@ It extends and styles an HTMLTextAreaElement.
 				:id="computedId"
 				ref="input"
 				class="textarea__input"
-				:type="type"
```
ShGKme, line 74: Unrelated to PR

ShGKme (approved):

src/components/NcTextArea/NcTextArea.vue
Antreesy: Know, just was clearing up all the Vue warnings while adding docs

PR #5440, opened 2024-03-25
add fallBackFocus to NcPopover

grnd-alt:
The Focus-Trap library will throw an error if a trap does not have any elements it can focus. Therefore we need something to set focus on if there's nothing else to use. In the documentation it is recommended to give the container a tab-index of -1 to make it focusable but not appear in the tab flow, and then set that as the fallBack to focus on when theres nothing else in the container.

### DIFF
--- a/src/components/NcPopover/NcPopover.vue
+++ b/src/components/NcPopover/NcPopover.vue
@@ -327,6 +327,7 @@ export default {
 			}
 
 			const el = this.getPopoverContentElement()
+			el.tabIndex = -1
 
 			if (!el) {
 				return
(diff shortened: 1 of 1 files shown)


src/components/NcPopover/NcPopover.vue
```diff
@@ -327,6 +327,7 @@ export default {
 			}
 
 			const el = this.getPopoverContentElement()
+			el.tabIndex = -1
```
juliushaertl, line 330: Should we only set it if there is no existing tab index?

juliushaertl (approved):
Makes sense looking at the focus-trap documentation for fallBackFocus, but would appreciate some additional feedback from @ShGKme or @susnux here :)

susnux (approved):
makes sense also from documentation of focus trap

PR #5446, opened 2024-03-29
Merge master into next

susnux:
Merge master into next to prepare for a new alpha version.


src/components/NcActions/NcActions.vue
```diff
@@ -716,8 +716,8 @@ export default {
 			</NcActions>
 		</p>
 
-		<h2>Tooltip</h2>
-		Has only text and not interactive elements
+		<h2>Toolip</h2>
```
Antreesy, line 719: Is it on master as well?

Antreesy (approved):

src/components/NcActions/NcActions.vue
susnux: Did not change this - this is from the merge. So yes

susnux: Ah noticed the typo :see_no_evil:
That should not be backported

susnux, line 719:
```suggestion
		<h2>Tooltip</h2>
```

PR #5448, opened 2024-03-29
fix(NcActions): Show last action entry only partial to make it discoverable

susnux:
### ☑️ Resolves

- Resolves https://github.com/nextcloud/server/issues/44463

1. Fix wrong event name - the event is named `apply-show` and not `after-show`
2. If the popover content is too large and will be capped (max-height) make sure the last element is shown only partially, so users can discover there is more to scroll.

### 🖼️ Screenshots

See that before you can not see any hint of the overflow:

🏚️ Before | 🏡 After
---|---
Screenshot_20240329_220236|Screenshot_20240329_220309

### 🏁 Checklist

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -1441,10 +1441,42 @@ export default {
 			}
 		},
 
-		onOpen(event) {
+		/**
+		 * Called when popover is shown after the show delay
+		 */
+		onOpen() {
 			this.$nextTick(() => {
-				this.focusFirstAction(event)
+				this.focusFirstAction(null)
+				this.resizePopover()
 			})
+		},
+
+		/**
+		 * Hanle resizing the popover to make sure users can discover there is more to scroll
+		 */
+		resizePopover() {
+			// Get the inner v-popper element that defines the popover height (from floating-vue)
+			const inner = this.$refs.menu.closest('.v-popper__inner')
+			const maxHeight = Number.parseFloat(window.getComputedStyle(inner).maxHeight)
+			const height = this.$refs.menu.clientHeight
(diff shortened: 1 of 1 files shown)


susnux:

szaimen (approved):
Screenshots look good to me :)
(but didnt test and didnt review the code)

Pytal (approved):
Tested with different viewport heights and works nicely!

jancborchardt:
@susnux we also need to keep in mind this one though: https://github.com/nextcloud-libraries/nextcloud-vue/issues/3920
Yes long menus should have the last entry halved, but in general the menus can take up more vertical space.

susnux:
> but in general the menus can take up more vertical space.

Yes but that is a different story and should be handled as its own. This only fixes menus overflowing our "maximum" to show the last one half.

skjnldsv:
> * Fix wrong event name - the event is named `apply-show` and not `after-show`

Actually not sure, we do change the event name by catching it in the `NcPopover`
https://github.com/nextcloud/nextcloud-vue/blob/dd7ba318cc651cc2c32f69cab2a05d561a53122a/src/components/NcPopover/NcPopover.vue#L168-L169
https://github.com/nextcloud/nextcloud-vue/blob/dd7ba318cc651cc2c32f69cab2a05d561a53122a/src/components/NcPopover/NcPopover.vue#L401
https://github.com/nextcloud/nextcloud-vue/blob/dd7ba318cc651cc2c32f69cab2a05d561a53122a/src/components/NcPopover/NcPopover.vue#L410

susnux:
> Actually not sure, we do change the event name by catching it in the NcPopover

We also forward all listeners so we should be good either way. But seems you are right...
A bit weird to have two events with exactly the same meaning :/

PR #5451, opened 2024-04-04
fix(sidebar): propagate keydown unless mobile

pulsejet:
Someone else might want to handle that event

https://github.com/pulsejet/memories/issues/1115

### DIFF
--- a/src/components/NcAppSidebar/NcAppSidebar.vue
+++ b/src/components/NcAppSidebar/NcAppSidebar.vue
@@ -379,7 +379,7 @@ A working alternative would be using an icon together with an `aria-label`:
 			ref="sidebar"
 			class="app-sidebar"
 			:aria-labelledby="`app-sidebar-vue-${uid}__header`"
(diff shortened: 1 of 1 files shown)


skjnldsv:
Remember to sign your commits @pulsejet :)

skjnldsv (approved):

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -379,7 +379,7 @@ A working alternative would be using an icon together with an `aria-label`:
 			ref="sidebar"
 			class="app-sidebar"
 			:aria-labelledby="`app-sidebar-vue-${uid}__header`"
-			@keydown.esc.stop="isMobile && closeSidebar()">
+			@keydown.esc="isMobile && ($event.stopPropagation() | closeSidebar())">
```
ShGKme, line 382: Is it intended to use "bitwise Or" here?

If it is to call to independent actions, let's move an entire handler to a method

ShGKme (approved):

PR #5455, opened 2024-04-05
fix(NcRichText): include all label items

DorraJaouad:
### Breaking change:
event `'interact:todo'` now returns a string (id of the affected checkbox) instead of an object.

- Fix https://github.com/nextcloud/spreed/issues/12009

* Checkbox includes different elements in its label
* Added explicit indentation for nested lists
* Interactivity: used the DOM to calculate the position of the element and then use it in the string

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

image

Interactivity: 

https://github.com/nextcloud-libraries/nextcloud-vue/assets/84044328/130eee3e-5532-4b1b-92b3-8efa1a702a72

### 🏁 Checklist

### DIFF
--- a/src/components/NcRichText/NcRichText.vue
+++ b/src/components/NcRichText/NcRichText.vue
@@ -115,13 +115,29 @@ Table row | value A | value B
 			}
 		},
 		methods: {
-			handleInteraction(event) {
-				const uncheckedItem = '- [ ] ' + event.label + '\n'
-				const checkedItem = '- [x] ' + event.label + '\n'
-
-				this.text = event.value
-					? this.text.replace(uncheckedItem, checkedItem)
-					: this.text.replace(checkedItem, uncheckedItem)
+			handleInteraction(id) {
+				const parentId = id.split('-markdown-input-')[0]
+				const index = Array.from(document.querySelectorAll(`span[id^="${parentId}-markdown-input-"]`)).findIndex((el) => el.id.includes(id))
+				if (index === -1 ) {
+					return
+				}
+				let checkBoxIndex = 0
+				lines = this.text.split('\n')
+				for (let i = 0; i < lines.length; i++) {
+					if (lines[i].includes('[ ]') || lines[i].includes('[x]')) {
+						if (checkBoxIndex === index) {
+							const isChecked = lines[i].includes('[x]')
+							if (isChecked) {
+								lines[i] = lines[i].replace('[x]', '[ ]')
+							} else {
+								lines[i] = lines[i].replace('[ ]', '[x]')
+							}
+							break
+						}
+						checkBoxIndex++
+					}
+				}
+				this.text = lines.join('\n')
 			},
 		},
 	}
@@ -150,11 +166,13 @@ See [NcRichContenteditable](#/Components/NcRichContenteditable) documentation fo
 
 		<NcCheckboxRadioSwitch :checked.sync="autolink" type="checkbox">Autolink</NcCheckboxRadioSwitch>
 		<NcCheckboxRadioSwitch :checked.sync="useMarkdown" type="checkbox">Use Markdown</NcCheckboxRadioSwitch>
+		<NcCheckboxRadioSwitch :checked.sync="useExtendedMarkdown" type="checkbox">Use extended Markdown</NcCheckboxRadioSwitch>
 
 		<NcRichText :text="text"
 			:autolink="autolink"
 			:arguments="userMentions"
-			:use-markdown="useMarkdown" />
+			:use-markdown="useMarkdown"
+			:use-extended-markdown="useExtendedMarkdown" />
 	</div>
 </template>
 <script>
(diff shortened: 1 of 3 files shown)


src/components/NcRichText/NcRichText.vue
```diff
@@ -462,10 +462,10 @@ export default {
 										},
 										on: {
 											'update:checked': (value) => {
-												this.$emit('interact:todo', { id, label, value })
+												this.$emit('interact:todo', { id, label: labelParts.join(''), value })
 											},
 										},
-									}, [label])
+									}, [labelParts])
```
ShGKme, line 468: As `labelParts` is an array already:

```suggestion
									}, labelParts)
```

ShGKme (commented):
Works much better than before, but sub-lists are still a little bit broken

Before | After
---|---
image | image

```
## Test

List:
- List1
  - Sub item 1
- List 2
  - Sub item 2

Tasks:
- [ ] Task 1
- [ ] @"Test Offline" **open** [web-page](https://nextcloud.com) 
  - [ ] One
  - [ ] Two
```

src/components/NcRichText/NcRichText.vue
```diff
@@ -462,10 +462,10 @@ export default {
 										},
 										on: {
 											'update:checked': (value) => {
-												this.$emit('interact:todo', { id, label, value })
+												this.$emit('interact:todo', { id, label: labelParts.join(''), value })
```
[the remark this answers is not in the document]
DorraJaouad: You mean distinguish between `- [] Line 1` in position 1 and `- [] Line 1` in position 3 ? there is an id that makes unique regardless of label.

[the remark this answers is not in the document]
Antreesy: Yes, but I haven't found a proper way to convert id in DOM into position in String (which is needed for substring replacement)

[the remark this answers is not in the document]
DorraJaouad: I could use id but one tiny issue: when it's nested, my assumption relying on the order in `$children` is not correct :((((

src/components/NcRichText/NcRichText.vue
```diff
@@ -449,24 +471,37 @@ export default {
...
 										attrs: {
 											...inputNode.data.attrs,
 											id,
 											disabled: !this.interactive,
 										},
 										on: {
-											'update:checked': (value) => {
-												this.$emit('interact:todo', { id, label, value })
+											'update:checked': () => {
+												this.$emit('interact:todo', id)
```
ShGKme, line 498: Formally, this is a breaking change. Public event payload has changed.

Though it could be fine, it is not a widely used feature.

src/components/NcRichText/NcRichText.vue
```diff
@@ -449,24 +471,37 @@ export default {
...
 							if (this.useExtendedMarkdown) {
+								let nestedList
 								if (tag === 'li' && Array.isArray(children)
 									&& children[0].tag === 'input'
 									&& children[0].data.attrs.type === 'checkbox') {
-									const [inputNode, , label] = children
-									const id = 'markdown-input-' + GenRandomId(5)
+									const [inputNode, ...labelParts] = children
+
+									const nestedListIndex = labelParts.findIndex((child) => child.tag === 'ul' || child.tag === 'li')
```
Antreesy, line 480: also should be some other block-displayed elements:

```suggestion
									const nestedListIndex = labelParts.findIndex((child) => ['ul', 'ol', 'li', 'blockquote', 'pre'].includes(child.tag))
```

src/components/NcRichText/NcRichText.vue
```diff
@@ -449,24 +471,37 @@ export default {
...
-												this.$emit('interact:todo', { id, label, value })
+											'update:checked': () => {
+												this.$emit('interact:todo', id)
 											},
 										},
-									}, [label])
-									return h(tag, attrs, [inputComponent])
+									}, labelParts)
+
+									return h(tag, attrs, [inputComponent, nestedListIndex ? [nestedList] : []])
```
Antreesy, line 503: `-1` would be truthy value, if I'm not wrong. 
initially `nestedList` is undefined, so why don't we use it directly?

```suggestion
									return h(tag, attrs, [inputComponent, nestedList])
```

src/components/NcRichText/NcRichText.vue
```diff
@@ -449,24 +471,37 @@ export default {
 
 						if (!tag.startsWith('#')) {
 							if (this.useExtendedMarkdown) {
+								let nestedList
```
Antreesy, line 474: As it could be not only list:

```suggestion
								let nestedNode = null
```

Antreesy (commented):
Left a couple of comments. Good workaround with handling the correct line!

Antreesy (approved):

ShGKme (approved):

PR #5458, opened 2024-04-08
feat(app-sidebar): allow to define inline for secondary buttons

szaimen:
- For https://github.com/nextcloud/calendar/pull/5907

| Before | After |
|---|---|
| image | image |


jancborchardt (changes requested):
Also here, relevant comment at https://github.com/nextcloud/calendar/pull/5906#pullrequestreview-1986633149
→ Showing so many icons without labels makes things quite difficult to understand for normal people, and not all of these actions are very important. Thus there is a reason some are hidden. :)

PR #5466, opened 2024-04-09
chore: Add CI workflow to check for backport request

susnux:
### ☑️ Resolves

This adds a CI workflow to remember people to request a backport for our Vue3 migration.

### 🏁 Checklist

### DIFF
--- a/.github/workflows/ensure-backported.yml
+++ b/.github/workflows/ensure-backported.yml
@@ -1,0 +1,29 @@
+
+name: Ensure changes are migrated to Vue3
+
+on:
+  pull_request:
+    types: [opened, ready_for_review, reopened, synchronize, labeled]
+    branches:
+      - main
+      - master
+
+permissions:
+  contents: read
+
+jobs:
+  commit-message-check:
+    if: github.event.pull_request.draft == false
+
+    name: Require Vue3 migration backport
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Run check
+      run: |
+        if ! ${{ contains(github.event.pull_request.labels.*.name, 'backport-request') }}; then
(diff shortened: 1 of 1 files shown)


Pytal (approved):

.github/workflows/ensure-backported.yml
```diff
@@ -0,0 +1,29 @@
+
+name: Ensure changes are migrated to Vue3
+
+on:
+  pull_request:
+    types: [opened, ready_for_review, reopened, synchronize, labeled]
```
ShGKme, line 6: Maybe it should be on merged?

.github/workflows/ensure-backported.yml
```diff
@@ -0,0 +1,29 @@
...
+    branches:
+      - main
+      - master
+
+permissions:
+  contents: read
+
+jobs:
+  commit-message-check:
+    if: github.event.pull_request.draft == false
```
ShGKme, line 16: Should we also check, that it is not a PR from a bot?

.github/workflows/ensure-backported.yml
Antreesy: Nobody cares after it's merged, so should be done beforehand.

But I'm not sure if that's a good flow:
- open PR
- get red CI
- trigger bot
- rerun CI again

susnux: As soon as the bot adds the label the `labeled` trigger will cause the CI to be green - no manual interaction needed.

.github/workflows/ensure-backported.yml
susnux: I would say we can exclude dependabot - because that is useless for that branch.
All other bots need manual review anyways and probably should also be ported.

susnux:
Too close now

PR #5485, opened 2024-04-11
fix(NcReferenceWidget): do not try to render widget when there is no widget

ShGKme:
### ☑️ Resolves

`NcReferenceWidget` always tries to render the widget into `ref="customWidget"`, but `ref="customWidget"` exists only when there is a custom widget (`reference && hasCustomWidget`), for example, if the widget is not registered yet there is no div to render the widget.

It doesn't look like a bug for the end user, but spams with errors on the console.

image


juliushaertl (approved):

Antreesy (approved):

Antreesy:
Rebased fot Cypress fix

PR #5487, opened 2024-04-11
fix(autolink): correctly handle relative links without webroot

ShGKme:
### ☑️ Resolves

- Continue of https://github.com/nextcloud-libraries/nextcloud-vue/pull/5419

I just realized that relative links should be considered relative to the Nextcloud instance, not origins.

For example, for the page `http://cloud.ltd/nextcloud/index.php/apps/files/favorites`, the relative link should be `/apps/files/favorites`, not `/nextcloud/index.php/apps/files/favorites`.

- Adjusted relative links generation tests
- Added one specific test for this case
- Changed `getRoute`

Tests for all the cases looks a bit messy, but I don't know how to make it better.

### 🏁 Checklist

### DIFF
--- a/src/components/NcRichText/autolink.js
+++ b/src/components/NcRichText/autolink.js
@@ -114,16 +114,19 @@ export const parseUrl = (text) => {
  *
  * @param {import('vue-router').default} router - VueRouter instance of the router link
  * @param {string} url - absolute URL to parse
- * @return {string|null} a path that can be useed in the router link or null if this URL doesn't match this router config
- * @example http://cloud.ltd/nextcloud/index.php/app/files/favorites?fileid=2#fragment => /files/favorites?fileid=2#fragment
+ * @return {string|null} a path that can be used in the router link or null if this URL doesn't match this router config
+ * @example http://cloud.ltd/nextcloud/index.php/apps/files/favorites?fileid=2#fragment => /files/favorites?fileid=2#fragment
  */
 export const getRoute = (router, url) => {
 	/**
-	 * http://cloud.ltd /nextcloud /index.php/app/files /favorites?fileid=2#fragment
-	 * |_____origin____|__________router-base__________|_________router-path________|
-	 * |__________base____________|
-	 *                 |___root___|
+	 * http://cloud.ltd /nextcloud /index.php /apps/files /favorites?fileid=2#fragment
+	 * |_____origin____|___________router-base____________|                           |
+	 * |__________base____________|          |______________relative-url______________|
+	 * |               |___root___|_optional_|__app-base__|_________router-path_______|
 	 */
+
+	const removePrefix = (str, prefix) => str.startsWith(prefix) ? str.slice(prefix.length) : str
+	const removePrefixes = (str, ...prefixes) => prefixes.reduce((acc, prefix) => removePrefix(acc, prefix), str)
 
 	// Router is not defined in the app => not an app route
 	if (!router) {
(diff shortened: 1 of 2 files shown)


tests/unit/components/NcRichText/autoLink.spec.js
```diff
@@ -152,7 +169,9 @@ describe('autoLink', () => {
 					],
 				})
 
-				expect(getRoute(routerTalkDesktop, link)).toBe(null)
+				expect(getRoute(routerTalkDesktop, link)).toBe(expectedRoute)
+				expect(getRoute(routerTalkDesktop, link)).toBe(expectedRoute)
+				expect(getRoute(routerTalkDesktop, link)).toBe(expectedRoute)
```
Antreesy, line 174: Three times to be sure?

tests/unit/components/NcRichText/autoLink.spec.js
```diff
@@ -129,17 +129,34 @@ describe('autoLink', () => {
...
 			it.each([
-				['https://cloud.ltd/nextcloud/index.php/apps/spreed?callTo=alice'],
-				['https://cloud.ltd/nextcloud/index.php/call/abc123ef'],
-				['https://cloud.ltd/nextcloud/index.php/apps/files'],
-				['https://cloud.ltd/nextcloud/'],
-			])('should not get route for %s', (link) => {
-				// On Talk Desktop both Base and Root URL returns an absolute path because there is no location
+				['https://cloud.ltd/nextcloud/index.php/apps/spreed?callTo=alice', '/apps/spreed?callTo=alice'],
+				['https://cloud.ltd/nextcloud/index.php/call/abc123ef', '/call/abc123ef'],
+				['https://cloud.ltd/nextcloud/index.php/apps/files', null],
```
Antreesy, line 151: I don't get, why links like this or `https://cloud.ltd/apps/spreed` shouldn't work, if that's an app root (empty screen for Talk, home directory for Files)? 

Even for Talk Desktop `"file://.. ../talk_window/index.html#/apps/spreed"` is possible to open

P.S. Looking at the VueRouter setup in the test, maybe I've said some stupid things :see_no_evil: 
But in that case a plain test like `['/apps/spreed', '/apps/spreed']` (without queries) would be helpful to understand it

src/components/NcRichText/autolink.js
```diff
@@ -115,16 +115,19 @@ export const parseUrl = (text) => {
  * @param {import('vue-router').default} router - VueRouter instance of the router link
  * @param {string} url - absolute URL to parse
  * @return {string|null} a path that can be useed in the router link or null if this URL doesn't match this router config
```
Antreesy, line 117:
```suggestion
 * @return {string|null} a path that can be used in the router link or null if this URL doesn't match this router config
```

src/components/NcRichText/autolink.js
```diff
@@ -133,29 +136,21 @@ export const getRoute = (router, url) => {
 	const isAbsoluteURL = /^https?:\/\//.test(url)
 
 	// URL is not a link to this Nextcloud server instance => not an app route
-	if ((isAbsoluteURL && !url.startsWith(getBaseUrl())) || (!isAbsoluteURL && !url.startsWith(getRootUrl()))) {
+	if (isAbsoluteURL && !url.startsWith(getBaseUrl())) {
 		return null
 	}
 
-	// Vue 3: router.options.history.base
```
Antreesy, line 140: Don't loose this commit

tests/unit/components/NcRichText/autoLink.spec.js
ShGKme: Misclick 🙈

src/components/NcRichText/autolink.js
ShGKme: Installed Spell Checker extension to VSCode as well 🙈

tests/unit/components/NcRichText/autoLink.spec.js
ShGKme: `getRoute` returns client route - VueRouter's route for a relative or absolute URL. The path that use can use as `RouteLocationRaw`, as `<RouterLink :to=` value.

ShGKme: > I don't get, why links like this ... shouldn't work

`https://cloud.ltd/nextcloud/index.php/apps/files` should not work for talk router, because `/apps/files` doesn't match any 

> or `https://cloud.ltd/apps/spreed` shouldn't work

This should work, there is no test where `https://cloud.ltd/apps/spreed` results with `null` for Talk / Talk Desktop

src/components/NcRichText/autolink.js
ShGKme: It' already backported, so the comment didn't look useful anymore for me

Antreesy (approved):

DorraJaouad (approved):

PR #5493, opened 2024-04-12
fix(NcDateTime): When relative time is disabled return a string and not the reference to it

susnux:
### ☑️ Resolves

The `.value` call was missing in case of disabled relative time. This causes the string to be quoted (because of stringification of the computed value). To prevent similar issues the file is now migrated to Typescript, also fixed some other issues discovered by type checking.

### 🏁 Checklist


emoral435 (approved):
Great changes :)

Antreesy (approved):
LGTM

PR #5516, opened 2024-04-22
fix(NcCheckboxRadioSwitch): Do not use HTML attribute name `id` as property name

susnux:
### ☑️ Resolves

We tried already in the past to not use native attribute names for custom props to prevent future issues, so lets use a custom name here.

### 🏁 Checklist


Antreesy (approved):
Shouldn't be a breaking change, as it's an internal component, right?

src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
```diff
@@ -87,7 +87,7 @@ export default {
 		/**
 		 * Unique id attribute of the input to label
 		 */
-		id: {
+		inputId: {
```
ShGKme, line 90: This `id` is only used in `aria-labelledby` in `NcCheckboxRadioSwitch`. And for this component it makes no sense what is `inputId`: it is not directly connected to any input. It sets an id based on prop so that `NcCheckboxRadioSwitch` can using the same pattern build the same id and use it `aria-labelledby`.

Can we instead keep it `id` (because it actually sets the id of the root element) and just pass `${id}-label` from `NcCheckboxRadioSwitch`.

Then there also will be only one component that build string `${id}-label`, not two.

ShGKme:
> Shouldn't be a breaking change, as it's an internal component, right?

Yes, this is an internal component

src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
susnux: Makes sense, adjusted the code!

ShGKme (approved):
I forgot to approve back then 🙈

PR #5527, opened 2024-04-25
fix(NcRichText): adjust display so long texts in checkboxes can be shown correctly

DorraJaouad:
### ☑️ Resolves

At least for checkboxes in `RichText`, it makes sense to have inline display. 

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image| image

### 🏁 Checklist

### DIFF
--- a/src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
+++ b/src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
@@ -224,8 +224,6 @@ export default {
 
 	&__text {
 		flex: 1 0;
-		display: flex;
-		align-items: center;
 
 		&:empty {
 			// hide text if empty to ensure checkbox outline is a circle instead of oval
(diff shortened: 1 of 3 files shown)


Antreesy:
Maybe it should be fixed directly in the component, rather than using 'deep'?

ShGKme:
> Maybe it should be fixed directly in the component, rather than using 'deep'?

I wonder why checkbox text is a flex container in first place...

DorraJaouad:
> Maybe it should be fixed directly in the component, rather than using 'deep'?

Well, I thought it should remain as a flex container for some reason . It was added to center text vertically (https://github.com/nextcloud-libraries/nextcloud-vue/pull/5151)

I moved it to the specific checkbox type that needs it.

Antreesy (approved):
Tested, compared with `https://nextcloud-vue-components.netlify.app`.

Looks like a decent solution for now

ShGKme (approved):

PR #5546, opened 2024-05-04
feat: Add prettier for checking and fixing stylistic rules

susnux:
### ☑️ Resolves

* Added CI workflow
* Adjusted ESLint preset to not check stylistic rules
* Run prettier to fix formatting

### 🏁 Checklist

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,19 +1,19 @@
 ### ☑️ Resolves
 
-- Fix #…
+-   Fix #…
 
 ### 🖼️ Screenshots
 
-🏚️ Before | 🏡 After
----|---
-B | A
+| 🏚️ Before | 🏡 After |
+| --------- | -------- |
+| B         | A        |
 
 ### 🚧 Tasks
 
-- [ ] ...
+-   [ ] ...
 
 ### 🏁 Checklist
 
-- [ ] ⛑️ Tests are included or are not applicable
-- [ ] 📘 Component documentation has been extended, updated or is not applicable
-- [ ] 3️⃣ Backport to `next` requested with a Vue 3 upgrade
+-   [ ] ⛑️ Tests are included or are not applicable
+-   [ ] 📘 Component documentation has been extended, updated or is not applicable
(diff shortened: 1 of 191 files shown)


susnux:
@ShGKme you asked about this.
I think it looks pretty decent. The only thing I do not like is that arrays are not enforced to have one element per line, this would make git diff smaller when some elements are added or removed in the future.

But the good thing about prettier: There is no discussion as the options are pretty limited (print width discussion incomming :sweat_smile: )

.github/pull_request_template.md
```diff
@@ -1,19 +1,19 @@
...
+-   [ ] ...
 
 ### 🏁 Checklist
 
-- [ ] ⛑️ Tests are included or are not applicable
-- [ ] 📘 Component documentation has been extended, updated or is not applicable
-- [ ] 3️⃣ Backport to `next` requested with a Vue 3 upgrade
+-   [ ] ⛑️ Tests are included or are not applicable
+-   [ ] 📘 Component documentation has been extended, updated or is not applicable
+-   [ ] 3️⃣ Backport to `next` requested with a Vue 3 upgrade
```
ShGKme, line 19: Seems weird. Does it put three spaces to be in line with a header?

ShGKme:
@susnux I'm fine with and without Prettier understanding its advantages and drawbacks. It's ugly sometimes, but at least it is the same style everywhere and in personal projects I prefer to use it.

I'm only not fine with inconsistency, e.g. having our stylelint config saying "use Prettier for formatting" and then not using Prettier anywhere. Or using it rarely unlike the current usage of ESLint.

To be honest, I completely missed the discussion about migrating to Prettier in Nextcloud codebase and the config discussions...

> print width discussion incomming

<details>

😶‍🌫️ 120+, if we don't use punch cards 😶‍🌫️

</details>

.github/pull_request_template.md
susnux: No it aligns it with the tab width for this:
```md
-   First point
    Also the same point
-   Second point
```

susnux:
We can reopen if we decide to use it

PR #5551, opened 2024-05-06
style(NcSelect): restrict max-height to prevent overflow by component

Antreesy:
### ☑️ Resolves

- Fix cases then component could overflow page content (in sidebar, for example)
  - Restrict max-height of 'dropdown-toggle' element to 200px (5 visible lines of tags + 1 half-visible to scroll)
  - Make action button sticky to follow position

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image
before-select | after-select

### 🚧 Tasks

- [ ] Check other use-cases
- [ ] Discuss restriction size 

### 🏁 Checklist


ShGKme (approved):
Nice. I'd do it even smaller by default, e.g. 2.5 lines.

ShGKme:
Probably we need a different component to handle it. Something like GitHub's label dialog where you can search.

jancborchardt (commented):
Great idea! Agree with @ShGKme that a height of 2.5 lines would be nicer.

Antreesy:
Upd: reduce max-height to 100px (2.5 - 2.6 visible lines)
Github-styled labels is a good idea, but it's more a feature-level change (restyle of NcSelectTags wouldn't be enough, IMO), so I'd fix existing component first

susnux (approved):

jancborchardt (commented):
Thanks @Antreesy! :)

PR #5561, opened 2024-05-07
fix(NcReferenceWidget): use requestAnimationFrame in observers

max-nextcloud:
Without `requestAnimationFrame` resize observer causes cypress to crash.

Also use a ref for the width and a computed for the resulting number of lines.

### ☑️ Resolves

- Fixes cypress failures such as https://github.com/nextcloud/text/actions/runs/8976130504/job/24652414376?pr=5772

### 🖼️ Screenshots

grafik

### 🏁 Checklist

- [x] ⛑️ Tests are not applicable
- [x] 📘 Component documentation is not applicable

### DIFF
--- a/src/components/NcRichText/NcReferenceWidget.vue
+++ b/src/components/NcRichText/NcReferenceWidget.vue
@@ -62,7 +62,7 @@ export default {
 	},
 
 	setup() {
-		const width = ref(700)
+		const width = ref(0)
 		const isVisible = ref(false)
 		// This is the widget root node
 		const widgetRoot = ref()
--- a/src/components/NcRichText/NcReferenceWidget.vue
+++ b/src/components/NcRichText/NcReferenceWidget.vue
@@ -30,7 +30,7 @@
 </template>
 <script>
 import { useIntersectionObserver, useResizeObserver } from '@vueuse/core'
-import { ref } from 'vue'
+import { nextTick, ref } from 'vue'
 import { RouterLink } from 'vue-router'
 
 import { t } from '../../l10n.js'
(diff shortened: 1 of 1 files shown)


max-nextcloud:
Cypress fails like this:

grafik

No idea what is causing this.

src/components/NcRichText/NcReferenceWidget.vue
```diff
@@ -62,29 +62,25 @@ export default {
 	},
 
 	setup() {
-		const compact = ref(3)
+		const width = ref(700)
```
ShGKme, line 65: Feels a bit magic number defined as "some number that is more than 650".

ShGKme (approved):

src/components/NcRichText/NcReferenceWidget.vue
```diff
@@ -62,29 +62,25 @@ export default {
...
-			} else if (entries[0].contentRect.width < 550) {
-				compact.value = 1
-			} else if (entries[0].contentRect.width < 650) {
-				compact.value = 2
-			} else {
-				compact.value = 3
-			}
+			window.requestAnimationFrame(() => {
+				width.value = entries[0].contentRect.width
+			})
```
ShGKme, line 79: If that is only for Cypress, I'd add a comment. Otherwise it is very non-obious for my why there is a need for RAF

susnux (approved):

src/components/NcRichText/NcReferenceWidget.vue
max-nextcloud: Frankly I am not sure. In another case I looked into in text I was seeing a
`"ResizeObserver loop completed with undelivered notifications"`
error right before the crash.

I did look into that some and it seems to indicate that there is a problem with the callback. Some cause can be an infinite loop where the resize triggers the observer which triggers a callback that triggers another resize.

https://blog.flowpoint.ai/post/v2/solved-resizeobserver-loop-completed-with-undelivered-notifications/

We could also use a debounce but that seemed even more involved.

juliushaertl (approved):

src/components/NcRichText/NcReferenceWidget.vue
max-nextcloud: A `nextTick` also does the trick and seems more light weight.

PR #5563, opened 2024-05-07
Use variables for input border width

marcoambrosini:
https://github.com/nextcloud-libraries/nextcloud-vue/assets/26852655/a2c1e554-3488-4d99-8b82-a3c377f768ab

### DIFF
--- a/src/components/NcInputField/NcInputField.vue
+++ b/src/components/NcInputField/NcInputField.vue
@@ -371,7 +371,7 @@ export default {
 
 		background-color: var(--color-main-background);
 		color: var(--color-main-text);
-		border: 2px solid var(--color-border-maxcontrast);
+		border: var(--border-width-input, 2px) solid var(--color-border-maxcontrast);
 		border-radius: var(--border-radius-large);
 
 		cursor: pointer;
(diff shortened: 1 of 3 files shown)


ShGKme:
> Waiting for server `border-width` variable to be in before merging this

Can you link the PR? I cannot find it...

marcoambrosini:
@ShGKme it doesn't exist yet :)

jancborchardt (changes requested):
The input field looks good, but for the textarea, it jumps around vertically on focus, a result of the 1px/2px change on focus.

(And yes, waiting for server for the --border-width variable.)

marcoambrosini:
variable here https://github.com/nextcloud/server/pull/45247/commits/43510be0343aaedce183950e7a4a555951ec00aa

src/components/NcInputField/NcInputField.vue
```diff
@@ -371,7 +371,7 @@ export default {
 
 		background-color: var(--color-main-background);
 		color: var(--color-main-text);
-		border: 2px solid var(--color-border-maxcontrast);
+		border: var(--border-width-input) solid var(--color-border-maxcontrast);
```
ShGKme, line 374:
```suggestion
		border: var(--border-width-input, 2px) solid var(--color-border-maxcontrast);
```

ShGKme: And the same in other places

ShGKme (commented):
For compatibility with older server, we probably should use 2px as a fallback

marcoambrosini:
@ShGKme I've switched back to borders, but there's still a problem with the text jumping inside on hover and focus. I've tried to change the paddings but the problem is still there. Could you please have a look?

susnux (approved):

szaimen (approved):
🐘

ShGKme:
> but there's still a problem with the text jumping inside on hover and focus

Was it resolved?

PR #5565, opened 2024-05-08
feat(NcAppNavigationCaption): Add `heading-id` prop to allow setting the ID on the caption itself

susnux:
### ☑️ Resolves

Add a prop to set caption id to the caption, needed for linking list and caption.

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppNavigationCaption/NcAppNavigationCaption.vue
+++ b/src/components/NcAppNavigationCaption/NcAppNavigationCaption.vue
@@ -1,4 +1,6 @@
 <docs>
+### Basic usage
+
 ```vue
 	<template>
 		<ul class="nav">
@@ -95,6 +97,23 @@
 	</style>
 ```
 
+### Element used as a heading
+```vue
+	<template>
+		<!-- e.g. NcAppNavigation-->
+		<div style="display: flex; flex-direction: column;">
+			<NcAppNavigationCaption heading-id="mylist-heading"
+				is-heading
+				name="My navigation list" />
+			<NcAppNavigationList aria-labelledby="mylist-heading">
+				<NcAppNavigationItem name="First" />
+				<NcAppNavigationItem name="Second" />
+				<NcAppNavigationItem name="Third" />
+			</NcAppNavigationList>
+		</div>
+	</template>
+```
+
 </docs>
 
 <template>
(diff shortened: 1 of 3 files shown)


Pytal (approved):
🚀

susnux:
Alternatives:
- Add `caption-id` prop that is always set on the name component.
- Add `id` as prop and bind to name component, and add `actions-id` and bind to the actions component

ShGKme:
> * Add `caption-id` prop that is always set on the name component.

I'd go this way:
- We only need to set ID on the header (at least I don't see other cases)
- Seems very non-obvious behavior what attributes are proxied either on header or on actions menu trigger

And what if we need ID even when there are actions?

src/components/NcAppNavigationCaption/NcAppNavigationCaption.vue
```diff
@@ -148,6 +167,14 @@ export default {
...
 		},
 
+		/**
+		 * If `isHeading` is set, this defines the heading level that should be used
+		 */
+		headingLevel: {
+			type: Number,
+			default: 2,
+		},
+
```
ShGKme, line 177: A new feature then?

susnux: Probably. I need to fix a place that looks like this:
```
h2
  h2 < this should be h3 but is not possible
  ul < this is a appnavigation caption
```

susnux:
> I'd go this way [...]

I agree, but I would additionally change the props handling, because I think we have a problem with Vue3 here as `class` and `style` now are assigned to the actions instead of the wrapper.

Also I do not think we need additional attributes on the actions but should keep them on either the wrapper or the inner name.

src/components/NcAppNavigationCaption/NcAppNavigationCaption.vue
susnux: Splitted the PR.

ShGKme (approved):

src/components/NcAppNavigationCaption/NcAppNavigationCaption.vue
```diff
@@ -139,6 +160,15 @@ export default {
...
 		},
 
+		/**
+		 * `id` to set on the inner caption
+		 * Can be used for connecting the `NcActionCaption` with `NcActionList` using `aria-labelledby`.
+		 */
+		captionId: {
+			type: String,
+			default: null,
+		},
```
ShGKme, line 170: `headingId` for consistency with `isHeading` and `headingLevel`

susnux: I agree

Antreesy (approved):

susnux:
> /backport to next

Did it manually: https://github.com/nextcloud-libraries/nextcloud-vue/pull/5580

PR #5569, opened 2024-05-08
[next] fix: Adjust tsconfig and eslintrc for tests

susnux:
### ☑️ Resolves

* Fixes `rootdir` in invalid place in tsconfig.
* Fixes wrong location of eslintrc

### 🏁 Checklist


tsconfig.json
```diff
@@ -12,11 +12,13 @@
     "strict": true,
     "noImplicitAny": false,
     "outDir": "./dist",
+    "rootDir": "./src",
```
ShGKme, line 15: Maybe make it just the root to keep structure?
```suggestion
    "rootDir": ".",
```

ShGKme (commented):
Is it needed only for `next`?

tsconfig.json
susnux: The the output will be `dist/src/...` instead of `dist/...`

susnux:
> Is it needed only for next?

Yes only on next this is an issue

ShGKme (approved):

Antreesy (approved):

PR #5582, opened 2024-05-11
enh(useElementSize): track the size of an element

max-nextcloud:
### ☑️ Resolves

- Fix size computation broken in #5561.
- Add a composable for observing the size of an element.

#5561 did not just add the `nextTick` but also changed the resize handler to
```
				width.value = widgetRoot.value?.contentRect.width ?? 0
```

Where `widgetRoot` is the `ref` to the root node.
Turns out `contentRect` is only defined on the entries handed to the resizeObserver.
But widgetRoot is the ref of an HTML Element.

This PR fixes this,
adds tests,
and exposes the functionalilty of observing an elements size
which we are also using in text.

### 🏁 Checklist

### DIFF
--- a/cypress/component/composables/useElementSize.cy.ts
+++ b/cypress/component/composables/useElementSize.cy.ts
@@ -1,0 +1,58 @@
+/**
+ * SPDX-FileCopyrightText: 2024 Max <max@nextcloud.com>
+ * SPDX-License-Identifier: @license AGPL-3.0-or-later
+ *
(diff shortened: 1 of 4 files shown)


susnux:
Why not simply using: https://vueuse.org/core/useElementSize/#useelementsize

max-nextcloud:
Funny... I had looked for something like that but failed to see it. I guess ctrl-f would have helped.

PR #5584, opened 2024-05-11
feat(NcAppSidebar): Allow to set `open` state to prevent focus trap issues on mobile

susnux:
* Resolves https://github.com/nextcloud/forms/issues/2139
* Resolves https://github.com/nextcloud-libraries/nextcloud-vue/issues/5335

### ☑️ Resolves

If the current screen size is "mobile" (<= 512px) and the app sidebar is not removed from dom but only hidden (`v-show` instead of `v-if`) then currently a user can not interact with the page anymore.
This happens because a focus trap is registered but not removed when the element is hidden.

So this adds a comment about using `v-if` but for performance (e.g the sidebar is hidden and shown often) the sidebar also now allows to set the `open` prop for hiding it.

### 🏁 Checklist


susnux:
BTW the styleguide still does not work because multiple sidebars are opened that always set a focus trap.
So the only solution to fix the styleguide I can currently think of is hiding all examples by default.

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -737,7 +823,7 @@ export default {
 		this.toggleFocusTrap()
 	},
 
-	beforeDestroy() {
+	beforeUnmount() {
```
ShGKme, line 826: Have you checked this is working? I'm 99.99% sure new names are not backported to Vue 2.7...

ShGKme:
> BTW the styleguide still does not work because multiple sidebars are opened that always set a focus trap. So the only solution to fix the styleguide I can currently think of is hiding all examples by default.

You mean, if  styleguidist is open on mobile?

susnux:
> You mean, if styleguidist is open on mobile?

Yes, because currently one focus trap for each example is created and you can not interact with the page except for the first app sidebar example.

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -811,8 +897,14 @@ export default {
...
 			 *
 			 * @type {HTMLElement}
+			 * @deprecated Listen to `update:open` instead
 			 */
 			this.$emit('opening', element)
+			/**
+			 * Current open state emitted after the transitions are finished
+			 * @type {boolean}
+			 */
+			this.$emit('update:open', true)
```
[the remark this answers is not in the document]
susnux: removed

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -827,8 +919,10 @@ export default {
 			 * The sidebar is closing and the transition is in progress
 			 *
 			 * @type {HTMLElement}
+			 * @deprecated Listen to `update:open` instead
 			 */
 			this.$emit('closing', element)
+			this.$emit('update:open', false)
```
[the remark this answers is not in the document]
susnux: done

susnux:
@nextcloud-libraries/designers about the toggle icon, we have this discussion here:
https://github.com/nextcloud-libraries/nextcloud-vue/issues/2804

So I used the icon recommended by @nimishavijay but got some feedback that it does not self descriptive for the sidebar.

Screenshot_20240513_161641

Maybe instead use `menu-open`? https://pictogrammers.com/library/mdi/icon/menu-open/

jancborchardt (changes requested):
To make sure that it’s obvious and also different from the left navigation which uses menu-open, let’s use dock-right: https://pictogrammers.com/library/mdi/icon/dock-right/
dock-right

susnux:
@jancborchardt implemented :heavy_check_mark:

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -366,155 +370,237 @@ A working alternative would be using an icon together with an `aria-label`:
...
 ```
-</docs>
 
+### Conditionally show the sidebar
+
+If the sidebar should be shown conditionally (e.g. using a button)
+and the users are expected to open and close the sidebar multiple times,
+then using `v-if` might result in bad performance.
+So instead use the `open` property:
+
```
ShGKme, line 380: Can you also mention that when `open` prop is used, there is a built-in toggle button, you users don't have to implement it manually?

susnux: Done :white_check_mark:

ShGKme:
*Checking attrs inheritance*

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -366,155 +370,237 @@ A working alternative would be using an icon together with an `aria-label`:
...
+			:aria-label="t('Open sidebar')"
+			class="app-sidebar__toggle"
+			:class="toggleClasses"
+			type="tertiary"
+			@click="$emit('update:open', true)">
+			<template #icon>
+				<IconDockRight :size="20" />
+			</template>
+		</NcButton>
+		<transition appear
```
ShGKme, line 456: This is a small breaking change
- Before `<aside>` through `<Transition>` was actually a root element receiving all passed attrs such as classes, styles or `data-cy`
- After it is a fragment, no attributes are passed.

So we need to bind `$attrs` and `$listeners`, and add `inheritAttrs: false`

ShGKme (changes requested):

jancborchardt (approved):

ShGKme (approved):

PR #5604, opened 2024-05-17
fix: Revert #5514 and only force width limit on Nextcloud 30+

susnux:
### ☑️ Resolves

#5514 is breaking and should **not** have been merged for this major version, because it also breaks server on Nextcloud 28 and 29 when updating the library!

Instead deprecate the prop and force overwrite on Nextcloud 30+ so you can fix your app or server on Nextcloud 30+, but keep working on Nextcloud 29 and lower.

### 🏁 Checklist

### DIFF
--- a/src/components/NcSettingsSection/NcSettingsSection.vue
+++ b/src/components/NcSettingsSection/NcSettingsSection.vue
@@ -33,7 +33,8 @@ This component is to be used in the settings section of nextcloud.
 	<NcSettingsSection
 		name="Two-Factor Authentication"
 		description="Two-factor authentication can be enforced for all users and specific groups."
-		doc-url="https://docs.nextcloud.com/server/19/go.php?to=admin-2fa">
+		doc-url="https://docs.nextcloud.com/server/19/go.php?to=admin-2fa"
+		:limit-width="true">
 		<p>Your settings here</p>
 	</NcSettingsSection>
 </template>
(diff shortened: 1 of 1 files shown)


susnux:
>  3️⃣ Backport to next requested with a Vue 3 upgrade

No needed we can do the breaking changes there

src/components/NcSettingsSection/NcSettingsSection.vue
```diff
@@ -33,15 +33,16 @@ This component is to be used in the settings section of nextcloud.
 	<NcSettingsSection
 		name="Two-Factor Authentication"
 		description="Two-factor authentication can be enforced for all users and specific groups."
-		doc-url="https://docs.nextcloud.com/server/19/go.php?to=admin-2fa">
+		doc-url="https://docs.nextcloud.com/server/19/go.php?to=admin-2fa"
+		:limit-width="true">
```
ShGKme, line 37: Why do we need it with the default value?

ShGKme (approved):
Very nice fix, tested:

`master` | `stable28`
---|---
image | image

src/components/NcSettingsSection/NcSettingsSection.vue
susnux: I think in the past (v7 or so) it was false by default.

But I have not changed it, just revert

ShGKme: Maybe remove it from the example to not show the deprecated prop?

max-nextcloud (approved):

PR #5608, opened 2024-05-17
fix(NcAppSidebar): make closing animation less glitchy

ShGKme:
### ☑️ Resolves

- Fixes a number of issues with animations:
  1. Sidebar doesn't slide right, it becomes "narrow", actually **moving content** inside (especially with multiline title)
  1. There is a moment when there is **close** and **open** at **the same position**, because the close button is not moving
  1. It was not `width` but `max-width` animated. It means that if `max-width: 500px` and the current width is `27vw = 400px` - there is no animation for the first `100px` **making a delay** 
     - @marcoambrosini this also probably fixes your PR
  1. New `--app-sidebar-offset` changes too immediate, making **a jump effect**

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
before-quick | after-quick
before-slow | after-slow
mobile-before-quick | mobile-after-quick
mobile-before-slow | mobile-after-slow

### 🚧 Tasks

- [x] Use transition on `--app-sidebar-offset`
- [x] Position **Close button** relative to the sidebar
- [x] Slide with `margin` instead of `max-width`
- [x] Test on mobile

### 🏁 Checklist


Antreesy:
Wouldn't `transform: translate` more appropriate here to use for animation?

ShGKme:
> Wouldn't `transform: translate` be more appropriate here to use for animation?

No, it would keep a large empty place when the sidebar is closing and then immediately resize page content.

An alternative could be to use absolute position together with translate, so sidebar would open on top of page content without an empty space. However, it would also make content resize immediately.

So I used the same approach as we are using now on NcAppNavigation. We could go with an alternative solution if there are performance issues.

ShGKme:
@Antreesy see also gifs here: https://github.com/nextcloud-libraries/nextcloud-vue/pull/5389#pullrequestreview-1932803802

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -1116,11 +1116,21 @@ export default {
 </script>
 
 <style lang="scss">
+// Allows to use transition over a custom CSS property (CSS Variable)
+// Ignored on old browsers resulting in slightly noticeable jump
+@property --app-sidebar-offset {
```
susnux, line 1121: fancy!

ShGKme: Doesn't work in Firefox, except nightly build. But when it doesn't work, it's just a little more jumping

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -1148,6 +1155,9 @@ $top-buttons-spacing: 6px;
 	app-content will be shrinked properly
 */
 .app-sidebar {
+	--app-sidebar-width: clamp(300px, 27vw, 500px);
```
susnux, line 1158: Theoretically we support iOS 12.x (from our browserslist, from clients its 15.x only). So `clamp` does not work for iOS < 13.3

If we still support it, then:
```css
	--app-sidebar-width: 300px;
	--app-sidebar-width: clamp(300px, 27vw, 500px);
```

or
```css
	--app-sidebar-width: min(max(300px, 27vw), 500px);

susnux: nevermind is 15.5 nowaday

susnux (approved):
Looks smooth when testing

ShGKme:
Rebased onto `master`, re-checked, no changes, ready to review

ShGKme:
Snapshots should be updated. Previously they had no close button because it has `absolute` position without `relative` on `.app-sidebar`. It worked on the web-page, but not on cypress render.

I'll upload new snapshots when I have a stable internet connection.

marcoambrosini (approved):

marcoambrosini:
It worked for me in files, but now testing in talk it seems that the sidebar does not close anymore @ShGKme 

https://github.com/nextcloud-libraries/nextcloud-vue/assets/26852655/e08b388e-92e8-4eb0-866a-3b31a3785c03

PR #5610, opened 2024-05-17
[next] refactor!: Drop CJS entry points, only provide ESM

### LINKED ISSUE #4899: [RFC] Drop CommonJS entry point
We currently provide CommonJS and ESM entry points.
While ESM is able to import CommonJS packages, the opposite does only work with dynamic `import()` and not with `require()`.

This causes issues as some of our dependencies are pure ESM dependencies, meaning our CJS output is theoretically invalid, because it contains `require('pure-esm-module')`.
This causes issues with tools like jest, as the code is not valid.

*(Notice that for app users this is not a problem, if they use Vite it will load the ESM entry and if they use webpack it will transpile this library and all dependencies to UMD)*

But I see no benefit of providing both types, as CommonJS users will still have to transpile this library because of the dependencies.

---

My reasons for going pure ESM:
* Less error prone distribution (we had problems with one of both types being invalid with both webpack and vite)
* ESM is the standard (and hopefully someday being it also in the Node.js universe)
* Simplify tooling & build process
* Smaller distribution size
* Needs to be transpiled anyway so should be no change for users.

I am happy to hear your thoughts, especially if you know reasons for keeping both :smiley:

susnux:
### ☑️ Resolves

* Resolves #4899

There is very little need for the CJS built anymore as any modern bundler can handle ESM (webpack, vite).
This not only reduces the package size and prevents tree-shaking issues if the CJS entry is used, but also reduces potential pitfalls with maintaining two module formats.

### 🏁 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,37 +31,30 @@
     "cypress:gui": "TZ=UTC cypress open --component",
     "cypress:update-snapshots": "TZ=UTC cypress run --component --spec \"cypress/visual/**/*.cy.{t,j}s\" --env type=base --config screenshotsFolder=cypress/snapshots/base"
   },
-  "main": "dist/index.cjs",
   "exports": {
     ".": {
       "types": "./dist/index.d.ts",
-      "import": "./dist/index.mjs",
-      "require": "./dist/index.cjs"
+      "import": "./dist/index.mjs"
     },
     "./dist/Components/*.js": {
       "types": "./dist/components/*/index.d.ts",
-      "import": "./dist/Components/*.mjs",
-      "require": "./dist/Components/*.cjs"
+      "import": "./dist/Components/*.mjs"
     },
     "./dist/Directives/*.js": {
       "types": "./dist/directives/*/index.d.ts",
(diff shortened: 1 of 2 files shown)


ShGKme (approved):
> reduces the package size

It is 2.4 MB for a 28 MB package folder in an almost 1 GB `node_modules` in server (<10% for a package, ~0.24% for server development) 👀 😶‍🌫️

> prevents tree-shaking issues if the CJS entry is used

But everything that can handle ESM - uses ESM on bundling (both Webpack and Vite). Are there such issues?

> reduces potential pitfalls with maintaining two module formats

Agree

Also, reduces bundling time.

susnux:
> Are there such issues?

Well indirectly, there are packages which depend on the CJS entry and causing two different versions of the same component to be imported :see_no_evil: 
You can test this when only building on entry point (e.g. ESM) and then link in server and try to build.
(Also the other way around).

So doing this right now in alpha stage allows to prevent this situation with Vue 3 packages of our libraries.

Pytal (approved):

PR #5612, opened 2024-05-17
fix: Adjust license in `package.json`

susnux:
### ☑️ Resolves

We are using AGPL-3.0 or later, so adjust the package.json to reflect this

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,7 +7,7 @@
     "": {
       "name": "@nextcloud/vue",
       "version": "8.12.0",
-      "license": "AGPL-3.0",
+      "license": "AGPL-3.0-or-later",
       "dependencies": {
         "@floating-ui/dom": "^1.1.0",
         "@linusborg/vue-simple-portal": "^0.1.5",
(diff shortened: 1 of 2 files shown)


ShGKme (commented):
I might be wrong, but I though it is not valid to change 3.0 to 3-or-later... cc @nickvergessen

ShGKme (commented):
I might be wrong, but I though it is not valid to change 3.0 to 3-or-later... cc @nickvergessen

susnux:
@ShGKme relicensing is not possible without asking every contributor (if they are from community). But this is not re-licensing the library but fixing it. Every file in the source is AGPL-3.0-or-later, there is no AGPL-3.0 (or correctly AGPL-3.0-only) file.

ShGKme:
> Every file in the source is AGPL-3.0-or-later

Indeed. Except the `LICENSE` root file

ShGKme (approved):

ChristophWurst (approved):

skjnldsv (approved):

skjnldsv:
> Every file in the source is AGPL-3.0-or-later

Except the `package(-lock).json` :see_no_evil:

ShGKme:
> Except the `package(-lock).json` 🙈

Does it actually have its own license? Not the `license` property which determines the package license (and where `AGPL-3.0` also means `AGPL-3.0-or-later`)

skjnldsv:
> Does it actually have its own license?

I would (maybe wrongly) assume that every file have its own license anyway :thinking:

PR #5627, opened 2024-05-21
fix(NcAppSidebar): make sidebar a single node to allow using v-show

ShGKme:
### ☑️ Resolves

- Fix https://github.com/nextcloud-libraries/nextcloud-vue/issues/5623

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist


susnux:
I vote for not backporting this to next and instead require to use `open` prop. Maybe introduce `no-toggle` prop to hide the toggle button.

susnux (commented):
Do you prefer this one? Because I think it adds a lot of complexity.
If yes I will free some time for a review :sweat_smile:

ShGKme:
> Do you prefer this one?

I dislike both.

I think this is much simpler (check diff with hide whitespaces and without looking into docs — it's literally just removing `<Fragment>` and adding `<Portal>`).

But this could be more risky because we need to rely on `#content-vue` or another selector and because Teleport to a node inside Virtual DOM is always risky.

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -451,177 +448,265 @@ export default {
...
+			class="app-sidebar"
+			:aria-labelledby="`app-sidebar-vue-${uid}__header`"
+			@keydown.esc="onKeydownEsc">
+			<!--
+				We cannot render toggle button inside sidebar (aside#app-sidebar-vue), because it is hidden then the toggle is needed.
+				But we also need transition with the sidebar to be the root of this component to use it as a single UI element, allowing to use `v-show`.
+				So we cannot render the toggle button directly in this component.
+				As a simple solution - render it in the content to keep correct position.
+			-->
+			<Teleport selector="#content-vue">
```
susnux, line 553: We should add `no-toggle` prop to allow hiding the toggle button.

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -1151,7 +1235,6 @@ $top-buttons-spacing: $app-navigation-padding; // align with app navigation
 */
 .app-sidebar {
 	--app-sidebar-width: clamp(300px, 27vw, 500px);
-	--app-sidebar-padding: #{$app-navigation-padding};
```
susnux, line 1154: Dont we need to style the toggle?

src/components/NcAppSidebar/NcAppSidebar.vue
susnux: but unrelated, just to note.

src/components/NcAppSidebar/NcAppSidebar.vue
susnux, line 1154: Why removing this?

src/components/NcAppSidebar/NcAppSidebar.vue
ShGKme: I can restore a PR with a feature of placing `NcAppSidebarToggle` button anywhere and `no-toggle` prop.

ShGKme: Only `no-toggle`: https://github.com/nextcloud-libraries/nextcloud-vue/pull/5702

susnux (approved):
works.

I am fine with both solutions :shrug:

ShGKme:
- Rebased onto `master` to resolve conflicts
- Added a small `fixup`:
  - Removed `inheritAttrs: false` (not needed anymore)
  - Moved `v-if` from `<NcButton>` to `<teleport>` to not have unneeded teleport without button.

Antreesy (approved):
I'd personnaly prefer this solution (not a fan of several nodes under root template, now it's more predictable).
Tested, works fine as a middleground before moving to `open.sync`

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -1060,6 +1156,19 @@ export default {
 			this.$refs.tabs.focusActiveTabContent()
 		},
 
+		/**
+		 * Check if the toggle button container is available
+		 */
+		checkToggleButtonContainerAvailability() {
+			// Toggle button must be rendered, but there is no element to teleport it to
+			if (this.open === false && !this.noToggle && !this.ncContentSelector) {
+				console.warn(
```
susnux, line 1165: why not Vue.warn ?

susnux (approved):

src/components/NcAppSidebar/NcAppSidebar.vue
ShGKme: I'd keep it to have warning also in the production build of an app.

But we should definitely depside what we should use in which case for warnings in frontend.

ShGKme:
Rebased onto master and squashed

susnux:
@ShGKme backport?

PR #5632, opened 2024-05-22
fix(NcAppNavigationItem): Fix popper boundary element

ChristophWurst:
### ☑️ Resolves

- For https://github.com/nextcloud/mail/issues/8474

The query selector in the computed prop ran too early, before the element existed. There is no reactivity. Now the boundary element is set very late at mount, when the boundary element does exist in the DOM.

There are a few edge cases, but the menu positions better than before when the boundary fell back to the `body`.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
+++ b/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
@@ -629,6 +629,10 @@ export default {
 		}
 	},
 
+	mounted() {
+		this.actionsBoundariesElement = document.querySelector('#content-vue') || undefined
+	},
+
 	data() {
 		return {
 			editingValue: '',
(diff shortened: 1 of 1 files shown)


GretaD:
I end up in the same scenario as you. I can reproduce the issue on master, then i link and the issue is gone. I do npm ci, kill the link and same as you i cannot reproduce it anymore.

ChristophWurst:
Let's do this. It sets the element more reliably. I don't see a downside.

susnux (approved):
Makes sense because on mounted the parent should exist which is the content.

GretaD (approved):

PR #5639, opened 2024-05-27
Update temporary border radius variable

marcoambrosini:
Ref: https://github.com/nextcloud/server/pull/45295

### DIFF
--- a/src/assets/NcAppNavigationItem.scss
+++ b/src/assets/NcAppNavigationItem.scss
@@ -8,7 +8,7 @@
 	min-height: $clickable-area;
 	transition: background-color var(--animation-quick) ease-in-out;
 	transition: background-color 200ms ease-in-out;
-	border-radius: var(--border-radius-new-design, var(--border-radius-pill));
+	border-radius: var(--border-radius-element, var(--border-radius-pill));
 
 	&-wrapper {
 		position: relative;
@@ -117,7 +117,7 @@
 		&:focus-visible {
 			box-shadow: 0 0 0 4px var(--color-main-background);
 			outline: 2px solid var(--color-main-text);
-			border-radius: var(--border-radius-new-design, var(--border-radius-pill));
+			border-radius: var(--border-radius-element, var(--border-radius-pill));
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


szaimen (approved):

ShGKme (commented):
Shouldn't it also be used in input elements?

marcoambrosini:
> Shouldn't it also be used in input elements?

@ShGKme they use --large, my understanding it that all --element would go back to --large eventually

Anyway can we do this in a follow up it in the end it's needed? I need these changes in asap

susnux (approved):

ShGKme:
> @ShGKme they use --large, my understanding it that all --element would go back to --large eventually

@marcoambrosini Even if they have the same value doesn't mean they have the same semantics and should become a single variable in future.

So from the engineering part of theming and for customization, it is important to determine, is `--element` about border radius for all interactive elements and buttons + list items + inputs should have the same border radius (as it is said currently in the linked PR), or it is a different type of borders and we want to have separate configs for input and for buttons/list-items.

If they are different, we should also adjust the definition in https://github.com/nextcloud/server/pull/45295

All the variables are not only a part of internal implementation but also a part of the global public interface of Nextcloud, meaning that:
- It affects many apps and versions of Nextcloud and apps
- It is used for custom theming, useful for branded instances and helps developers to adjust theme
- Cannot be easily undone after the release without making a breaking change

marcoambrosini:
This goes in Nextcloud 30, as all the other design changes. Should these changes only be in next? 

> @marcoambrosini Even if they have the same value doesn't mean they have the same semantics and should become a single variable in future.

But didn't we say that we're going to remove this variable eventually and keep only `--large` and this new one was only temporary? If that's not the case, then it makes sense to use it with inputs as well yes.

susnux:
> This goes in Nextcloud 30, as all the other design changes. Should these changes only be in next?

No it needs to be also in v8.x, so just the milestone needs to be 8.x not 9.0.0 :)

> didn't we say that we're going to remove this variable eventually

We can only do so in roughly 2 years and then need refactor this library and possibly apps.
So not sure if it is not better to just keep the variable.

PR #5664, opened 2024-06-04
fix(NcVNodes): add re-export in root export and add docs

ShGKme:
### ☑️ Resolves

- For https://github.com/nextcloud-libraries/nextcloud-vue/issues/5662

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

A util component to render a VNode or an array of VNodes from the prop or the default slot.


susnux (approved):

raimund-schluessler:
Why not. Although we don't even export it in https://github.com/nextcloud-libraries/nextcloud-vue/blob/master/src/components/index.js

raimund-schluessler (approved):

ShGKme:
> we don't even export

Oh... 🥲

ShGKme:
> Although we don't even export it in

Fixed here as well.

In general component has been exported in `/dist/Components`.

ShGKme:
Commit renamed to `fix` with a new export.

PR #5665, opened 2024-06-04
chore(directives): add docs

ShGKme:
### ☑️ Resolves

- For: https://github.com/nextcloud-libraries/nextcloud-vue/issues/5662

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image
image | image

#### Directives usage

image

#### Focus

image

#### Linkify

image

### 🚧 Tasks

- [x] Add sections in directives
- [x] Add general Directives docs about registration
- [x] Update Tooltip directive doc page
- [x] Add Focus directive doc page
- [x] Add Linkify directive doc page

### DIFF
--- a/docs/directives/tooltip.md
+++ b/docs/directives/tooltip.md
@@ -1,0 +1,55 @@
+<!--
+ - SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+```js static
+import Tooltip from '@nextcloud/vue/dist/Directives/Tooltip.js'
+```
+
+Tooltip directive based on [v-tooltip](https://github.com/Akryum/v-tooltip).
+
+Note: it's preferred to use [native HTML title attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) instead for accessibility.
+
+## Usage
+
+In a component register the directive and use the `v-tooltip` in the template with static or dynamic value:
+
+```vue
+<template>
+    <button v-tooltip="'A tooltip text'">Hover me</button>
+</template>
+```
+
+You can specify the tooltip position as a modifier:
+
+```vue
+<button v-tooltip.bottom="'You have new messages.'">Hover me</button>
+```
+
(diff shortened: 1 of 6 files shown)


docs/directives/tooltip.md
```diff
@@ -0,0 +1,50 @@
+```js static
+import Tooltip from '@nextcloud/vue/dist/Directives/Tooltip.js'
+```
+
+Tooltip directive based on [v-tooltip](https://github.com/Akryum/v-tooltip).
+
+Note: it's preferred to use [native HTML title attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) instead for accessibility.
```
susnux, line 7: should we deprecate it?

ShGKme: I dunno, to be honest ⚆_⚆

In theory - yes, because we migrated everywhere in nextcloud-vue already to `title`. But apps are still using it also because of visual result.

Maybe we should re-check this rule.

ShGKme: But we can do it after mergin docs for already created and not yet marked deprecated directives 👀

susnux: Yes sure :)

susnux (approved):
Needs a rebase though

ShGKme:
Rebased onto `master`

styleguide.config.js
```diff
@@ -92,6 +92,21 @@ module.exports = async () => {
...
 				name: 'Directives',
 				content: 'docs/directives.md',
+				sectionDepth: 1,
+				sections: [
+					{
+						name: 'Focus',
+						content: 'docs/directives/focus.md',
+					},
+					{
+						name: 'Linkify',
```
Antreesy, line 102: I'd love to see component documentations in separate files as well, to keep SFC clean, but something for later maybe

ShGKme:
Fixed SPDX comments in new files

Antreesy (approved):
Looks legit

styleguide.config.js
ShGKme: Or at least move it to the end of the file, not the top 🥲

PR #5674, opened 2024-06-05
feat(NcRichContenteditable): programmatically show tributes

DorraJaouad:
### ☑️ Resolves

- Required for #https://github.com/nextcloud/spreed/issues/12250

- The added methods will be used to programmatically show slash commands by clicking on a button. 

### 🖼️ Screenshots

slash-command

### 🏁 Checklist

### DIFF
--- a/src/components/NcRichContenteditable/NcRichContenteditable.vue
+++ b/src/components/NcRichContenteditable/NcRichContenteditable.vue
@@ -210,6 +210,48 @@ export default {
 </style>
 ```
 
+### Using public methods
+
+```vue
+<template>
+	<div>
+		<div class="buttons-wrapper">
+			<NcButton class="show-slash-button" @click="showSlashCommands">Slash commands</NcButton>
+			<NcButton class="focus-button" @click="focus">Focus on input</NcButton>
+		</div>
+		<NcRichContenteditable
+			ref="contenteditable"
+			label="Write a comment"
+			:value.sync="message"
+			:maxlength="100"/>
+	</div>
+</template>
+<script>
+export default {
+	data() {
+		return {
+			message: '**Lorem ipsum** dolor sit amet. ',
+		}
+	},
+	methods: {
+		showSlashCommands() {
+			this.$refs.contenteditable.showTribute('/')
+		},
+		focus() {
+			this.$refs.contenteditable.focus()
+		},
+	},
+}
+</script>
+<style lang="scss" scoped>
+	.buttons-wrapper {
+		display: flex;
+		gap: 10px;
+		margin-bottom: 20px;
+	}
+</style>
+```
+
 </docs>
 
 <template>
@@ -922,6 +964,8 @@ export default {
 				this.getTributeContainer().setAttribute('class', this.tribute.current.collection.containerClass || this.$style['tribute-container'])
 
 				this.setupTributeIntegration()
+				// Remove the event handler if any
+				document.removeEventListener('click', this.hideTribute, true)
 			} else {
 				// Cancel loading data for autocomplete
 				// Otherwise it could be received when another autocomplete is already opened
@@ -1000,6 +1044,28 @@ export default {
 			} else {
 				this.getTributeContainer().classList.remove(this.$style['tribute-container--focus-visible'])
 			}
+		},
+
+		/**
+		 * Show tribute menu programmatically.
+		 * @param {string} trigger - trigger character, can be '/', '@', or ':'
+		 *
+		 * @public
+		 */
+		showTribute(trigger) {
+			this.focus()
+			const index = this.tribute.collection.findIndex(collection => collection.trigger === trigger)
+			this.tribute.showMenuForCollection(this.$refs.contenteditable, index)
+			document.addEventListener('click', this.hideTribute, true)
+		},
+
+		/**
+		 * Hide tribute menu programmatically
+		 *
+		 */
+		hideTribute() {
+			this.tribute.hideMenu()
+			document.removeEventListener('click', this.hideTribute, true)
 		},
 	},
 }


src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -1001,6 +1015,23 @@ export default {
...
+			let index = 0 // default to first collection '@'
+			if (trigger === ':') {
+				index += this.linkAutocomplete
+			} else if (trigger === '/') {
+				index += this.linkAutocomplete + this.emojiAutocomplete
+			}
+			this.tribute.showMenuForCollection(this.$refs.contenteditable, index)
+		},
+
+		hideTributeManully() {
```
susnux, line 1031:
```suggestion
		hideTributeManually() {
```

ShGKme, line 1031: There is no alternative for non-manual hiding, calling a function is always a manual call 😀

```suggestion
		hideTribute() {
```

src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -1001,6 +1015,23 @@ export default {
 				this.getTributeContainer().classList.remove(this.$style['tribute-container--focus-visible'])
 			}
 		},
+
+		showTributeManually(trigger) {
```
ShGKme, line 1019:
```suggestion
		showTribute(trigger) {
```

src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -194,7 +196,19 @@ export default {
 		},
 		onSubmit() {
 			alert(this.message)
-		}
+		},
+		showSlashCommands() {
+			this.$refs.contenteditable.showTributeManually('/')
```
Antreesy, line 201:
```suggestion
		showSlashCommands(trigger) {
			this.$refs.contenteditable.showTributeManually(trigger)
```

src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -194,7 +196,19 @@ export default {
...
+			// Add an event listener to the document
+			document.addEventListener('click', this.hideTribute)
+		},
+		hideTribute(event) {
+			if (!event.target.closest('.show-slash-button')) {
```
Antreesy, line 206: We need a click outside of tribute container here, right?
```suggestion
			if (!event.target.closest('#' + this.$refs.contenteditable.tributeId)) {
```

src/components/NcRichContenteditable/NcRichContenteditable.vue
ShGKme, line 1019: If this is a public method, please, add documentation in JSDoc with a description and `@public` tag

ShGKme: Same for hide

src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -1001,6 +1015,23 @@ export default {
...
+
+		showTributeManually(trigger) {
+			// Programmatic show of the tribute menu
+			this.$refs.contenteditable.focus()
+			let index = 0 // default to first collection '@'
+			if (trigger === ':') {
+				index += this.linkAutocomplete
+			} else if (trigger === '/') {
+				index += this.linkAutocomplete + this.emojiAutocomplete
+			}
```
Antreesy, line 1027:
```suggestion
			const index = this.tribute.collection.findIndex(collection => collection.trigger === trigger)
```

ShGKme, line 1027: This works fine, but a bit hacky. If we change collections order, or init process, or make `@` optional, it's easy to lose this place where a sum of booleans is used to find the index.

Let's find it explicitly instead:
```suggestion
			const index = this.tribute.collection.findIndex((collection) => collection.trigger === trigger)
```

ShGKme (changes requested):
Feature makes sense

src/components/NcRichContenteditable/NcRichContenteditable.vue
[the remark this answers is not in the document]
ShGKme: > should also be in the component:

Note, this is not required and it will hide all the other properties, including public ones

ShGKme:
Remember to add `enhancement` or `bug` label for proper CHANGELOG generation :P

src/components/NcRichContenteditable/NcRichContenteditable.vue
DorraJaouad: No, the tribute should close when clicking inside the tribute too, the only click that would be ignored is the click on the slash command button

Antreesy: If you pick something via tribute, it should automatically update text field with selection and close itself, so shouldn't be a problem. I only noticed, that if you won't stop a click event and put a listener, it will be propagated and immediately close the tribute you just opened. So it's not only about ignoring the slash button

src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -194,7 +196,19 @@ export default {
...
+			// Add an event listener to the document
+			document.addEventListener('click', this.hideTribute)
+		},
+		hideTribute(event) {
+			if (!event.target.closest('.show-slash-button')) {
+				this.$refs.contenteditable.hideTributeManully()
+				document.removeEventListener('click', this.hideTribute)
+			}
+		},
+
```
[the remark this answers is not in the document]
DorraJaouad: 1. Good point, It wouldn't be removed if the user deleted the slash or typed the text and then space. One solution I can think of is to add handler addition and removal in `tribute-active-true/false` event handlers but that also would touch tributes triggered by key down events too :/ (they behave the same)
2. Indeed, it is sthg to deal with always, we can just pass the button class/id to `showTribute` public method

src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -1001,6 +1008,28 @@ export default {
...
 		},
+
+		/**
+		 * Show tribute menu programmatically. If using a click event, stop propagation to prevent closing the menu
+		 * @param {string} trigger - trigger character, can be '/', '@', or ':'
+		 *
+		 * @public
+		 */
+		showTribute(trigger) {
+			this.$refs.contenteditable.focus()
```
ShGKme, line 1019:
```suggestion
			this.focus()
```

src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -32,7 +32,9 @@ Try mentioning user @Test01 or inserting emoji :smile
 ```vue
 <template>
 	<div>
+		<NcButton class="show-slash-button" @click.stop="showSlashCommands('/')">Slash commands</NcButton>
```
ShGKme, line 35: Can we move it to a new example to not start component docs with it?

For example, to have a separate example with all the public methods.

ShGKme (approved):

src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -210,6 +210,48 @@ export default {
 </style>
 ```
 
+### Using methods
```
Antreesy, line 213: Also sound good, and maybe more explicit?

```suggestion
### Using public methods
```

src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -1001,6 +1045,28 @@ export default {
...
+		 * Show tribute menu programmatically. If using a click event, stop propagation to prevent closing the menu
+		 * @param {string} trigger - trigger character, can be '/', '@', or ':'
+		 *
+		 * @public
+		 */
+		showTribute(trigger) {
+			this.focus()
+			const index = this.tribute.collection.findIndex(collection => collection.trigger === trigger)
+			this.tribute.showMenuForCollection(this.$refs.contenteditable, index)
+			document.addEventListener('click', this.hideTribute)
```
Antreesy, line 1059: What about using a capturing event here? Then we don't need a `@click.stop`:

```suggestion
document.addEventListener('click', this.hideTribute, true)
```

src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -1001,6 +1045,28 @@ export default {
...
+		},
+
+		/**
+		 * Hide tribute menu programmatically
+		 *
+		 */
+		hideTribute() {
+			this.tribute.hideMenu()
+			document.removeEventListener('click', this.hideTribute)
```
Antreesy, line 1068: and this one should have `true` as well to match added listener

src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -1001,6 +1045,28 @@ export default {
 				this.getTributeContainer().classList.remove(this.$style['tribute-container--focus-visible'])
 			}
 		},
+
+		/**
+		 * Show tribute menu programmatically. If using a click event, stop propagation to prevent closing the menu
```
Antreesy, line 1050:
```suggestion
		 * Show tribute menu programmatically.
```

src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -922,6 +964,8 @@ export default {
 				this.getTributeContainer().setAttribute('class', this.tribute.current.collection.containerClass || this.$style['tribute-container'])
 
 				this.setupTributeIntegration()
+				// Remove the event handler if any
+				document.removeEventListener('click', this.hideTribute)
```
Antreesy, line 968: true

src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -210,6 +210,48 @@ export default {
 </style>
 ```
 
+### Using methods
+
+```vue
+<template>
+	<div>
+		<div class="buttons-wrapper">
```
Antreesy, line 218: lint

Antreesy (approved):

ShGKme (approved):

PR #5688, opened 2024-06-11
fix(NcAppSidebar): restore attrs and classes inheritance

ShGKme:
### ☑️ Resolves

Makes `<NcAppSIdebar>` with `Fragment` work like there is no `Fragment`

- Regression from: https://github.com/nextcloud-libraries/nextcloud-vue/pull/5584
- Has alternative: https://github.com/nextcloud-libraries/nextcloud-vue/pull/5627/
- `<Fragment>` breaks:
  - Passing `attributes` and `classes`, because in Vue 2 they are not a part of `$attrs` and even cannot be defined as `$props`
  - scoped styles from parent's `data-v-{scopeid}` on the root element
  - using `v-show` **but it is not fixed in this PR**
- `<Transition>` does not pass attributes with disabled inheritance (no idea why, but in vue 2 this component was not so pure)

This PR restores it in **quite a dirty way**. I don't see any better solution without a breaking change.

Example of broken apps - `Files`:
- No `cy-data-sidebar` attributes for `e2e`
- No `app-sidebar--full` class
- No `data-v-{scopeId}`  attr required for `app-sidebar--full`

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 Tasks

- [x] Pass `attrs` and `listeners` on `<aside>` directly
- [x] Manually take and bind **classes** and **styles** from VNode
- [x] Manually set `data-v-{scopeid}` from `$parent`

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppSidebar/NcAppSidebar.vue
+++ b/src/components/NcAppSidebar/NcAppSidebar.vue
@@ -455,8 +455,6 @@ export default {
 		</NcButton>
 		<transition appear
 			name="slide-right"
-			v-bind="$attrs"
-			v-on="$listeners"
 			@before-enter="onBeforeEnter"
 			@after-enter="onAfterEnter"
 			@before-leave="onBeforeLeave"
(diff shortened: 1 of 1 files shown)


susnux (approved):
This one is quite nice, small and works :+1:

AndyScherzinger (approved):

ShGKme:
Closing in flavor of https://github.com/nextcloud-libraries/nextcloud-vue/pull/5627

PR #5698, opened 2024-06-13
feat: add isDarkTheme functions and composables

ShGKme:
### ☑️ Resolves

A copy from what we use in Talk

- Sometimes we need to check if the dark theme is enabled
- Currently, apps do it in different ways, and sometimes it's incorrect:
  - You should not use `window.matchMedia.('(prefers-color-scheme: dark)')`. It checks for the user's system theme, but Nextcloud Dark theme could be enabled even on the light system theme.
  - You should not use `[data-themes*=dark]` or `[data-theme-dark]` attributes on the body. It checks for explicitly set dark theme, but a user may use the system theme.
- Add utils and composables to check the Nextcloud dark theme based on `--background-invert-if-dark`

### 🖼️ Screenshots

image

image

### 🏁 Checklist


susnux:
> ⚠️ Requires dark theme support in docs

So you might want this a review ;)
https://github.com/nextcloud-libraries/nextcloud-vue/pull/5656

juliusknorr (approved):

docs/composables/useIsDarkTheme.md
```diff
@@ -0,0 +1,72 @@
...
+    </div>
+</template>
+
+<script>
+export default {
+    setup() {
+        const isDarkTheme = useIsDarkTheme()
+
+        // For documentation only. Do not use in production.
+        function switchTheme {
```
Antreesy, line 53:
```suggestion
        function switchTheme() {
```

Antreesy (approved):
Compared with Talk, works in Styleguids =)

docs/composables/useIsDarkTheme.md
ShGKme: Oops

src/composables/useIsDarkTheme/index.ts
```diff
@@ -0,0 +1,40 @@
...
+import type { DeepReadonly, Ref } from 'vue'
+import { ref, readonly, watch } from 'vue'
+import { createSharedComposable, usePreferredDark, useMutationObserver } from '@vueuse/core'
+import { checkIfDarkTheme } from '../../functions/isDarkTheme/index.ts'
+
+/**
+ * Check whether the dark theme is enabled on a specific element.
+ * If you need to check an entire page, use `useIsDarkTheme` instead for better performance.
+ * Reacts on element attributes change and system theme change.
+ * @param el - The element to check for the dark theme enabled on
```
susnux, line 15:
```suggestion
 * @param el - The element to check for the dark theme enabled on (defaults to `body` element)
```

tsconfig.json
```diff
@@ -3,6 +3,7 @@
   "include": ["./src/**/*.ts"],
   "exclude": ["./src/**/*.cy.ts"],
   "compilerOptions": {
+    "allowImportingTsExtensions": true,
```
susnux, line 6: This does not work, just do not import with `.ts` extension in Typescript files (in JS files it work).

```suggestion
```

src/composables/useIsDarkTheme/index.ts
```diff
@@ -0,0 +1,40 @@
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import type { DeepReadonly, Ref } from 'vue'
+import { ref, readonly, watch } from 'vue'
+import { createSharedComposable, usePreferredDark, useMutationObserver } from '@vueuse/core'
+import { checkIfDarkTheme } from '../../functions/isDarkTheme/index.ts'
```
susnux, line 9:
```suggestion
import { checkIfDarkTheme } from '../../functions/isDarkTheme/index'
```

susnux (commented):
Should not import with `ts` extension in Typescript files, while this works with bundlers, it does not for Typescript itself.

ShGKme:
Fixed TS config for styleguidist

src/composables/useIsDarkTheme/index.ts
ShGKme: Why?

tsconfig.json
ShGKme: This works, I missed that styleguidist uses a different config

ShGKme: We have 46 places with `.ts` imports in this repo

ShGKme:
Force-pushed docs fixes according to @susnux suggestions

tsconfig.json
ShGKme: > Importing with `.ts` extension in Typescript is invalid when emitting is enabled, as the output will be invalid.
> So as long as we use a bundler everything works, but e.g. `tsc` would not work, because Typescript does **not** rewrite import extensions, so you either have to import Typescript files with `.js` or without any extension.

We almost never use `tsc` as a compiler (except `notify_push`) and in general, in modern TypeScript usage it's rarely used as a compiler/language and not as a static analyzer (type checker).

Specifically, in this repo, we already had `@babel/typescript-preset` to work with TS as a static analyzer, and only the webpack config was not correct.

Antreesy:
backport?

skjnldsv:
> * Currently, apps do it in different ways, and sometimes it's incorrect:
>   
>   * You should not use `window.matchMedia.('(prefers-color-scheme: dark)')`. It checks for the user's system theme, but Nextcloud Dark theme could be enabled even on the light system theme.
>   * You should not use `[data-themes*=dark]` or `[data-theme-dark]` attributes on the body. It checks for explicitly set dark theme, but a user may use the system theme.

Somehow I missed that PR.
Those two sentences are incorrect. The way I designed the dark theme to behave is exactly as it's described above:

1. Dark mode user settings have the priority.
2. Then the default system/browser settings.
3. Finally, the default is false

This should do the trick just fine:
```ts
export function isDarkModeEnabled() {
	const darkModePreference = window?.matchMedia?.('(prefers-color-scheme: dark)')?.matches
	const darkModeSetting = document.body.getAttribute('data-themes')?.includes('dark')
	return darkModeSetting || darkModePreference || false
}
```

skjnldsv:
Small addition, considering the method takes an HtmlElement, we also use the `[data-theme-dark]` attribute to scope down some element and enforce dark theme.
So specifying an element should also add a check to climb up the dom tree up to the body and look for this attribute. If not, then we check for `darkModeSetting`, then `darkModePreference`, and finally: `false`

skjnldsv:
> You must always check both.

No, user setting will overrule matchMedia.

ShGKme:
> Small addition, considering the method takes an HtmlElement, we also use the `[data-theme-dark]` attribute to scope down some element and enforce dark theme.
> So specifying an element should also add a check to climb up the dom tree up to the body and look for this attribute. If not, then we check for `darkModeSetting`, then `darkModePreference`, and finally: `false`

It is covered by `--background-invert-if-dark`

skjnldsv:
> It is covered by `--background-invert-if-dark`

But feels incredibly hackish. I'd like to use proper logic instead of relying on custom css queries. Especially considering `--background-invert-if-dark` have been designed to be removed in the future

ShGKme:
> > You must always check both.
> 
> No, user setting will overrule matchMedia. https://github.com/nextcloud/server/blob/381077028adf388a7081cf42026570c6be47b198/apps/theming/lib/Service/ThemeInjectionService.php#L34-L61

Then this is not correct?

> This should do the trick just fine:
> 
> ```ts
> export function isDarkModeEnabled() {
> 	const darkModePreference = window?.matchMedia?.('(prefers-color-scheme: dark)')?.matches
> 	const darkModeSetting = document.body.getAttribute('data-themes')?.includes('dark')
> 	return darkModeSetting || darkModePreference || false
> }
> ```

ShGKme:
> But feels incredibly hackish. I'd like to use proper logic instead of relying on custom css queries. Especially considering `--background-invert-if-dark` have been designed to be removed in the future

Yes, it is hackish. And it was already the most popular way to check for the dark theme in Nextcloud apps. This is why I though it is important to have it here in the library in one specific place. To fix it only once if we change it, not in all the places in many apps.

We can explicitly go up the DOM tree to check attributes in every node up to the body. But I don't think this explicit is more important than performance here.

As an alternative, I can propose explicitly setting a custom CSS property for that like. Aka:
```css
[data-theme-dark] {
  --nextcloud-theme: dark;
  --nextcloud-theme-dark: 1;
}
```

ShGKme:
> Well, we return the user setting first, then check the system settings ? This is exactly what I wrote ? 🤔

In one comment - yes, you propose to `return darkModeSetting || darkModePreference || false`.

In another, you said "user setting will overrule matchMedia" which sounds like we don't need `darkModeSetting ` when and `darkModePreference` is enough.

So I'm a bit confused here

skjnldsv:
> We can explicitly go up the DOM tree to check attributes in every node up to the body. But I don't think this explicit is more important than performance here.

`const isDarkTheme = !!document.body.closest('[data-theme-dark]');` should be efficient enough :grin: 

> As an alternative, I can propose explicitly setting a custom CSS property for that like. Aka:

But I like that option better! :+1: 



> In another, you said "user setting will overrule matchMedia" which sounds like we don't need `darkModeSetting ` when and `darkModePreference` is enough.

Maybe my var names are confusing, `darkModeSetting` is the theming app user config. `darkModePreference` is the system preference.

ShGKme:
> should be efficient enough 😁

Ok, maybe you are right here. Maybe `el.closest()` is even more performant.

But it would still require for additional `matchMedia` check, otherwise the system theme is lost.

ShGKme:
> Ok, maybe you are right here. Probably `el.closest()` is even more performant.

Checked. `.closest() + matchMedia()` and `getComputedStyle(el).getPropertyValue()` is the same in performance.

PR #5708, opened 2024-06-14
feat(NcUserBubble): add RouterLink support

Antreesy:
### ☑️ Resolves

- Show interactive cursor `pointer` if an element has click event

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/NcUserBubble/NcUserBubble.vue
+++ b/src/components/NcUserBubble/NcUserBubble.vue
@@ -70,6 +70,7 @@ export default {
 			<component :is="isLinkComponent"
 				class="user-bubble__content"
 				:style="styles.content"
+				:to="to"
 				:href="hasUrl ? url : null"
 				:class="{ 'user-bubble__content--primary': primary }"
 				v-bind="attrs"
@@ -108,6 +109,7 @@ import NcUserBubbleDiv from './NcUserBubbleDiv.vue'
 import NcAvatar from '../NcAvatar/index.js'
 import NcPopover from '../NcPopover/index.js'
 import Vue from 'vue'
+import { RouterLink } from 'vue-router'
 
 export default {
 	name: 'NcUserBubble',
@@ -159,6 +161,13 @@ export default {
 					return false
 				}
 			},
+		},
+		/**
+		 * Use bubble as a router-link for in-app navigation
+		 */
+		to: {
+			type: [String, Object],
+			default: undefined,
 		},
 		/**
 		 * Default popover state. Requires the UserBubble
@@ -240,7 +249,9 @@ export default {
 		},
 
 		isLinkComponent() {
-			return this.hasUrl ? 'a' : 'div'
+			return this.hasUrl
+				? (this.to ? RouterLink : 'a')
+				: 'div'
 		},
 
 		popoverEmpty() {
@@ -329,4 +340,7 @@ export default {
 	}
 }
 
+a.user-bubble__content {
+	cursor: pointer;
+}
 </style>


src/components/NcUserBubble/NcUserBubble.vue
```diff
@@ -250,6 +273,10 @@ export default {
 			return true
 		},
 
+		isClickable() {
+			return !!this.$listeners?.click || !this.popoverEmpty
+		},
+
```
ShGKme, line 279: This is not portable to Vue 3. It may work if we remove `click` from component's events, but by design component is not aware of its event subscribers.

It also should not be clickable when it is just a `div`, otherwise there is an a11y issue. So we can check for `popoverEmpty` and `isLinkComponent`.

ShGKme (changes requested):
This PR is for: https://github.com/nextcloud/spreed/pull/12510/

There `NcUserBubble` is used as a clickable `div` to implement manual router navigation. It is a UX and a11y issue (how to open it in a new tab or via keyboard?).

Instead, it's better to add `RouterLink` support to this component. Then it will be actually an interactive link element.

Antreesy:
> here `NcUserBubble` is used as a clickable `div`

Without additional context (spreed PR), clickable div itself should be already an issue. Maybe we should extend it to `<a>`/`<button>`/`<div>` at least?

ShGKme:
> > here `NcUserBubble` is used as a clickable `div`
> 
> Without additional context (spreed PR), clickable div itself should be already an issue. Maybe we should extend it to `<a>`/`<button>`/`<div>` at least?

Component already supports `<a>` with `href` prop. We can also add `to` prop to support `<RouterLink>`.

With Popover it is still clickable `<div>`, but let's call it a separate issue

src/components/NcUserBubble/NcUserBubble.vue
```diff
@@ -239,8 +263,23 @@ export default {
 			return this.url && this.url.trim() !== ''
 		},
 
+		route() {
+			return getRoute(this.$router, this.url)
+		},
```
ShGKme, line 268: Let's add explicit `to` param similar to other components, so we can provide route location, not only URL string.

Do we have any place where we cannot explicitly set `url`/`to` and we need auto router integration?

src/components/NcUserBubble/NcUserBubble.vue
ShGKme: This is still an issue. For links interactive styles may be applied when it's `<a>`.

Antreesy: I thought I dropped this commit

src/components/NcUserBubble/NcUserBubble.vue
susnux: I would also prefer the explicit `to` prop instead of auto route.

src/components/NcUserBubble/NcUserBubble.vue
```diff
@@ -239,8 +263,23 @@ export default {
...
+
 		isLinkComponent() {
```
susnux, line 270: (just a note: Why was this called like this in the first place `isX` always sounds like a boolean condition :pensive: )

src/components/NcUserBubble/NcUserBubble.vue
```diff
@@ -250,6 +289,10 @@ export default {
 			return true
 		},
 
+		isClickable() {
+			return !!this.$listeners?.click || !this.popoverEmpty
```
susnux, line 293:
```suggestion
			return this.isLinkComponent === 'a'
```

ShGKme: This skips `RouterLink`. but we can use `a` as a selector directly in CSS

susnux: > This skips RouterLink. but we can use a as a selector directly in CSS

True, probably even easier. (BTW just adjusted the suggestion: Invert the condition ;) )

src/components/NcUserBubble/NcUserBubble.vue
```diff
@@ -240,7 +249,9 @@ export default {
 		},
 
 		isLinkComponent() {
-			return this.hasUrl ? 'a' : 'div'
+			return this.hasUrl
+				? (this.to ? RouterLink : 'a')
+				: 'div'
```
Antreesy, line 254: Can we address this in follow-up, maybe? or `<a role="button">` would suffice?
```suggestion
				: (this.popoverEmpty ? 'div' : 'button')
```

ShGKme: Yes, let's do it in a follow up. It is not a part of the original issue. TBH I even don't remember, where we use use bubble with popover...

ShGKme (approved):

DorraJaouad (approved):
LGTM

src/components/NcUserBubble/NcUserBubble.vue
raimund-schluessler, line 254: Isn't this logic incorrect? If it's supposed to be a `RouterLink` it only has `to`, but not `url` set. But here both would be required to become a `RouterLink`.
I think this would be the correct solution, if we want to give precedence to `url` over `to`:
```suggestion
			return this.hasUrl
				? 'a'
				: (this.to ? RouterLink : 'div')
```
Or did I get something wrong here?

src/components/NcUserBubble/NcUserBubble.vue
```diff
@@ -329,4 +340,7 @@ export default {
 	}
 }
 
+a.user-bubble__content {
```
raimund-schluessler, line 343: `a` already has `cursor: pointer;` applied by default, no?

src/components/NcUserBubble/NcUserBubble.vue
Antreesy: No, seems I mislooked it

src/components/NcUserBubble/NcUserBubble.vue
Antreesy: There are two rules (from server) applied, with different value. But maybe it's redundant now, since we don't redefine cursor for div

PR #5730, opened 2024-06-21
Feat: emit height difference of NcReferenceList after loading.

DorraJaouad:
### ☑️ Resolves

Needed for https://github.com/nextcloud/spreed/issues/10627. Link previews are adding extra height which causes jumping in the chat scroll. 

### 🖼️ Screenshots

No visual changes

### 🚧 Tasks

- [ ] event `loaded` now has height difference as a payload [CHANGED BEHAVIOR]

### 🏁 Checklist

### DIFF
--- a/src/components/NcRichText/NcReferenceList.vue
+++ b/src/components/NcRichText/NcReferenceList.vue
@@ -106,14 +106,15 @@ export default {
 				return
 			}
 
+			// get the height of the component before the references are loaded
+			const height = this.$el.offsetHeight
 			this.resolve().then((response) => {
 				this.references = response.data.ocs.data.references
-				this.loading = false
-				this.$emit('loaded')
 			}).catch((error) => {
 				console.error('Failed to extract references', error)
+			}).finally(() => {
 				this.loading = false
-				this.$emit('loaded')
+				this.$emit('loaded', height)
 			})
 		},
 		resolve() {
--- a/src/components/NcRichText/NcRichText.vue
+++ b/src/components/NcRichText/NcRichText.vue
@@ -378,7 +378,7 @@ export default {
 			default: true,
 		},
 	},
-	emits: ['interact:todo'],
+	emits: ['interact:todo', 'reference-list-loaded'],
 
 	data() {
 		return {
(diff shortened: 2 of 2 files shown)


src/components/NcRichText/NcReferenceList.vue
```diff
@@ -106,14 +106,15 @@ export default {
...
 			this.resolve().then((response) => {
 				this.references = response.data.ocs.data.references
-				this.loading = false
-				this.$emit('loaded')
 			}).catch((error) => {
 				console.error('Failed to extract references', error)
+			}).finally(() => {
 				this.loading = false
-				this.$emit('loaded')
+				this.$emit('loaded', height)
```
susnux, line 117: I would say reporting the height is not really a expected payload of the event, no?
Because you do not really need it (see below).

src/components/NcRichText/NcRichText.vue
```diff
@@ -387,6 +387,13 @@ export default {
 	},
 
 	methods: {
+		handleReferenceListLoaded(oldHeight) {
+			this.$nextTick(() => {
+				// emit height difference to parent component
+				const heightDiff = this.$refs.referenceList.$el.offsetHeight - oldHeight
+				this.$emit('reference-list-loaded', heightDiff)
+			})
+		},
```
susnux, line 396:
```suggestion
		handleReferenceListLoaded() {
			const oldHeight = this.$refs.referenceList.$el.offsetHeight
			this.$nextTick(() => {
				// emit height difference to parent component
				const heightDiff = this.$refs.referenceList.$el.offsetHeight - oldHeight
				this.$emit('reference-list-loaded', heightDiff)
			})
		},
```

susnux: But in general would it not just work if you emit any event here and in Talk react on that element by scrolling the element into view?
Then you do not need any height calculation here

DorraJaouad: calculating the old height here may be already updated with the reference data ? 

> in Talk react on that element by scrolling the element into view?

The goal not to scroll to the element but to calculate the extra height added (it will put the view above its previous position) and then we scroll back to the right position.

susnux: > calculating the old height here may be already updated with the reference data added height ?

I am not sure about the timing of Vue's event handling, sadly I could not find good documentation about this. Maybe @ShGKme knows more?

> The goal not to scroll to the element but to calculate the extra height added (it will scroll the view above its previous position) and then we scroll back to the right position.

I do not know the richtext part good enough (and the talk code) so your solution might be the best.
But then I still would not pass the height as magic number to the event but use a context object with proper property naming like

`this.$emit('reference-list-loaded', { heightDiff })`

PR #5740, opened 2024-06-27
feat(NcAppSidebar): force show navigation for a single tab

Antreesy:
### ☑️ Resolves

- Allows to show a tab navigation if there is only one tab available
- Should be non-interactive, act as a visual header

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 Tasks

- [ ] Discuss on disabling:
  - [ ] keydown events
  - [ ] mouse events
  - [ ] aria-attributes

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppSidebar/NcAppSidebar.vue
+++ b/src/components/NcAppSidebar/NcAppSidebar.vue
@@ -69,8 +69,10 @@ Single tab is rendered without navigation.
 ```vue
 <template>
 	<div>
+		<NcCheckboxRadioSwitch :checked.sync="forceTabs">Force tab navigation</NcCheckboxRadioSwitch>
 		<NcAppSidebar
 			name="cat-picture.jpg"
+			:force-tabs="forceTabs"
 			subname="last edited 3 weeks ago">
 			<NcAppSidebarTab name="Settings" id="settings-tab">
 				<template #icon>
@@ -87,6 +89,11 @@ import Cog from 'vue-material-design-icons/Cog.vue'
 export default {
 	components: {
 		Cog,
+	},
+	data() {
+		return {
+			forceTabs: false,
+		}
 	},
 }
 </script>
@@ -677,6 +684,7 @@ export default {
 			<NcAppSidebarTabs v-show="!loading"
 				ref="tabs"
 				:active="active"
+				:force-tabs="forceTabs"
 				@update:active="onUpdateActive">
 				<slot />
 			</NcAppSidebarTabs>
@@ -833,6 +841,13 @@ export default {
 		 * Force the actions to display in a three dot menu
 		 */
 		forceMenu: {
+			type: Boolean,
+			default: false,
+		},
+		/**
+		 * Force the tab navigation to display even if there is only one tab
+		 */
+		forceTabs: {
 			type: Boolean,
 			default: false,
 		},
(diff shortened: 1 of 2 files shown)


src/components/NcAppSidebar/NcAppSidebarTabs.vue
```diff
@@ -26,7 +26,7 @@
 				:button-variant="true"
 				:checked="activeTab === tab.id"
 				:wrapper-id="`tab-button-${tab.id}`"
-				:tabindex="activeTab === tab.id ? 0 : -1"
+				:tabindex="activeTab === tab.id && !showForSingleTab ? 0 : -1"
```
ShGKme, line 29: Is it needed? Can we have a single tab with a different active tab?

src/components/NcAppSidebar/NcAppSidebarTabs.vue
```diff
@@ -84,6 +84,13 @@ export default {
 			type: String,
 			default: '',
 		},
+		/**
+		 * Force the tab navigation to display even if there is only one tab
+		 */
+		forceNavigation: {
+			type: Boolean,
+			default: false,
+		},
```
ShGKme, line 93: I'd call it `forceTabs` or `showSingleTab`, because there is nothing e called "navigation" in this component including docs.

ShGKme:
> Discuss on disabling:
> * [ ]  keydown events
> * [ ]  mouse events
> * [ ]  aria-attributes

I'd say, it is fine. Yes, it doesn't make much sense to have tab navigation with one tab, but from a11y perspective it is still "active tab one of one"

src/components/NcAppSidebar/NcAppSidebarTabs.vue
Antreesy: The point was to skip a navigation (as part of event disabling). But if we'll decide to keep it, then I'll simply revert this change

ShGKme: Technically it is still an active interactive element

ShGKme (approved):

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -836,6 +844,13 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		/**
+		 * Force the tab navigation to display even if there is only one tab
+		 */
+		forceTabs: {
```
susnux, line 850: Maybe `force-navigation` would be more clear?

susnux (approved):
Makes sense,  just one comment about the prop naming but no blocker as I am fine with both if you agree.

src/components/NcAppSidebar/NcAppSidebar.vue
ShGKme: https://github.com/nextcloud-libraries/nextcloud-vue/pull/5740#discussion_r1656805220

👀

susnux: oh makes sense I was confused by the doc of the prop :sweat_smile:

ShGKme: 🏓 🏐 🏓

PR #5757, opened 2024-07-02
feat: migrate `NcCollectionList` component from `nextcloud-vue-collections`

Antreesy:
### ☑️ Resolves

- Move component from `nextcloud-vue-collections` to `@nextcloud/vue` library
- Refactor code:
	- migrate service from Class to simple functions
	- migrate store to composable
	- make NcCollectionListItem a visual component
	- update avatar styles
	- use RouterLink for relative paths in the same app
	- update message.pot translations
	- add docs

### 🖼️ Screenshots

image

image

### 🏁 Checklist


src/components/NcCollectionList/index.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * SPDX-FileCopyrightText: 2019 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+export { default as NcCollectionListItem } from './NcCollectionListItem.vue'
```
susnux, line 5: Could you split this into two separate components?
Meaning `src/components/NcCollectionList` and `src/components/NcCollectionListItem`

This would match our other components

src/components/NcCollectionList/service.ts
```diff
@@ -0,0 +1,90 @@
+/**
+ * SPDX-FileCopyrightText: 2019 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import axios, { type AxiosResponse } from '@nextcloud/axios'
+import { generateOcsUrl } from '@nextcloud/router'
+
+type OCSResponse = {
```
susnux, line 9: Could be used form `@nextcloud/typings/ocs`

src/components/NcCollectionList/index.js
Antreesy: Sorry, didn't get you there. I used structure from NcRichContenteditable as an example, is that not the same?

Anyway, `ListItem` is internal component, and shouldn't be used separately from `List`, so maybe we don't need to re-export it (like `NcUserBubbleDiv`)?

susnux: It depends I would say, if it is private do not export it then it could be like it is right now.

susnux (approved):

DorraJaouad (approved):

PR #5777, opened 2024-07-04
fix(NcCheckboxRadioSwitch): only bind aria attributes to the input

skjnldsv:
Fix regression from https://github.com/nextcloud-libraries/nextcloud-vue/pull/5507

Needed for:
- https://github.com/nextcloud/server/pull/46239
- https://github.com/nextcloud/server/pull/46237
- https://github.com/nextcloud/server/pull/46283
- https://github.com/nextcloud/server/pull/46292

### DIFF
--- a/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
+++ b/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
@@ -258,7 +258,7 @@ export default {
 		class="checkbox-radio-switch"
 		:style="cssVars"
 		:type="isButtonType ? 'button' : null"
-		v-bind="isButtonType ? $attrs : {}"
+		v-bind="isButtonType ? $attrs : dataAttrs"
 		v-on="isButtonType ? listeners : null">
 		<input v-if="!isButtonType"
 			:id="id"
@@ -272,7 +272,7 @@ export default {
 			:indeterminate.prop="hasIndeterminate ? indeterminate : null"
 			:required="required"
 			:name="name"
-			v-bind="$attrs"
+			v-bind="nonDataAttrs"
 			v-on="listeners">
 		<NcCheckboxContent :id="id"
 			class="checkbox-radio-switch__content"
@@ -450,6 +450,18 @@ export default {
 	emits: ['update:checked'],
 
 	computed: {
+		dataAttrs() {
+			// filter all data attributes
+			return Object.fromEntries(Object.entries(this.$attrs)
+				.filter(([key]) => key.startsWith('data-')))
+		},
+
+		nonDataAttrs() {
+			// filter all non-data attributes
+			return Object.fromEntries(Object.entries(this.$attrs)
+				.filter(([key]) => !key.startsWith('data-')))
+		},
+
 		isButtonType() {
 			return this.type === TYPE_BUTTON
 		},
--- a/.reuse/dep5
+++ b/.reuse/dep5
@@ -7,7 +7,7 @@ Files: package.json package-lock.json .github/pull_request_template.md
 Copyright: Nextcloud GmbH and Nextcloud contributors
 License: AGPL-3.0-or-later
 
-Files: styleguide/assets/icons.css styleguide/assets/server.css cypress/snapshots/* tests/unit/functions/usernameToColor/__snapshots__/usernameToColor.spec.js.snap
+Files: styleguide/assets/icons.css cypress/snapshots/* tests/unit/functions/usernameToColor/__snapshots__/usernameToColor.spec.js.snap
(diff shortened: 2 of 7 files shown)


Antreesy (approved):

src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
```diff
@@ -272,7 +272,7 @@ export default {
 			:indeterminate.prop="hasIndeterminate ? indeterminate : null"
 			:required="required"
 			:name="name"
-			v-bind="$attrs"
+			v-bind="ariaAttrs"
```
susnux, line 275: I think this is very unexpected and might again break in the future as the structure of the component is private API and you should never do assumptions about it

src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
```diff
@@ -450,6 +450,18 @@ export default {
...
+			return Object.fromEntries(Object.entries(this.$attrs)
+				.filter(([key]) => key.startsWith('aria-')))
+		},
+
+		nonAriaAttrs() {
+			// filter all non-aria attributes
+			return Object.fromEntries(Object.entries(this.$attrs)
+				.filter(([key]) => !key.startsWith('aria-')))
+		},
+
```
susnux, line 464: What I think we could do here, especially because you now can not pass native attributes anymore, is for a "quick fix" do not use `aria-` here but `data-`.

Pass everything to the input except `data-` as data is non functional attributes so we are save not to forward them.

skjnldsv: That would be the safest solution, I like that idea. But it's still an oversight. 
I'm pretty sure there are tons of other html attributes that would make sense on the outer wrapper :shrug:

skjnldsv:
> Instead I think the tests on server are broken by design, they assume some internal component layout that is not guaranteed. It is not guaranteed that `outer-element-selector input` does exist.
> We could even have `<div class="the checkbox"><span role="checkbox"></div>` as the component structure.

The tests itself could have been written differently, but this is not really the issue here. The tests failure just made us notice this change. We are touching a different dilemma about what we expect. 
I personally think dropping all attributes to the wrapper is a dangerous and counter-intuitive pattern to use.

That's why in the past we always declared aria variable as props, so that we declare the attributes that are forwarded to the background,

skjnldsv:
> If I assign any attributes then those are expected to be on the checkbox not on any random unknown element.

Even if that was the case, this change should not have been a minor nor a patch. Whatever decision taken prior, even if less than ideal, should stay consistent :)

Anyway, let's move forward, while I think and would like us to be more careful in the future, I think we can indeed adjust and only keep data attributes on the outer layer. I'm too deep into other areas to notice potential missed pitfalls. As this is your code @susnux, i'll let you take the decision.

skjnldsv:
Server fixed with this PR
image

susnux:
> Even if that was the case, this change should not have been a minor nor a patch. Whatever decision taken prior, even if less than ideal, should stay consistent :)

It is because that it internal, for components the structure is private like for classes private methods or properties.
This change here is more like forcing people to only use the public API, like if you change your JS class from

```js
class Foo {
_private: 'foo'
}
```
to
```js
class Foo {
#private: 'foo'
}
```

Same problem we have with class names because we do not use two way scoping, people use internal class names of components which is dangerous as that is nothing public. Meaning even accessing `.button-vue` might work there is no good reason for us not to change it to something different anytime.
Of course we do not do it because we would bit our own tail as we then have to fix a lot of stuff.

To sum it up, what I want to say: If you have the possibility to make tests less fragile, we should do it.
But in this case I agree if it affects a lot of places, then we should do this here for DX! (BTW I already work with low priority on more robust tests for account management)

susnux (approved):

skjnldsv:
> To sum it up, what I want to say: If you have the possibility to make tests less fragile, we should do it.
> But in this case I agree if it affects a lot of places, then we should do this here for DX! (BTW I already work with low priority on more robust tests for account management)

I agree with any internal structure, it's subject to change and it's fine.
This is not what it is about here. We're preventing any dev applying standard html attributes to a component now. We're not talking about anything private or public, we're literally stripping away the very core of a component rendering, controlling the outer layer.
What if I want to add some styling to the component, will it be applied to the input now?
That would be problematic.

susnux:
> EDIT: answering my own question, `<NcCheckboxRadioSwitch style="whatever" />` is applied to the input and not the component root. That is not acceptable to me 😕

@skjnldsv no it is not for Vue2 `style` and `class` are considered special attributes and always bound to parent.

PR #5806, opened 2024-07-15
fix(NcActions): Use full window height

susnux:
### ☑️ Resolves

- Fix https://github.com/nextcloud-libraries/nextcloud-vue/issues/3920

Calculate the space on the window based on the trigger location.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot 2024-07-15 at 15-18-35 Nextcloud Vue Style Guide|Screenshot 2024-07-15 at 15-18-00 Nextcloud Vue Style Guide

### 🏁 Checklist

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -956,7 +956,7 @@ import GenRandomId from '../../utils/GenRandomId.js'
 import { t } from '../../l10n.js'
 import { getTrapStack } from '../../utils/focusTrap.js'
 import { useElementBounding, useWindowSize } from '@vueuse/core'
-import Vue, { ref, computed } from 'vue'
+import Vue, { ref, computed, toRef } from 'vue'
 
 import IconDotsHorizontal from 'vue-material-design-icons/DotsHorizontal.vue'
 
@@ -1131,7 +1131,7 @@ export default {
 		 */
 		boundariesElement: {
 			type: Element,
-			default: () => document.querySelector('body'),
+			default: () => document.querySelector('#app-content-vue') ?? document.querySelector('body'),
 		},
 
 		/**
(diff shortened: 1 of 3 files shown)


ShGKme (approved):

src/components/NcActions/NcActions.vue
Antreesy, line 1: Since we're touching it, could you also consider a restricted window for height caclucation?

On screenshot:
Talk chat (boundaries element / red lines) => NcActions with a lot of items
Falling out because of style.height being more than needed to end at the border
image

jancborchardt (commented):
Awesome, thanks a bunch! :)

src/components/NcActions/NcActions.vue
susnux: What container is that? I was thinking about replacing the arbitrary header height with the app container size, but I do not like messing with ("optional") parent components .

Antreesy: "scroller messages-list__scroller", if that's what you mean

susnux: So completely custom element. Maybe allow to pass a selector for it?

Antreesy: It is passed down to the external Dropdown component, or I'm talking about something different?
image
image

susnux: Yes but we need to do it manually anyways for the 1/2 element height ob the scroller.

but I pushed a commit for respecting the boundaries element :)

Antreesy (approved):
Looks better (if change `container` prop in addition, but that's a todo for me now)
image

src/components/NcActions/NcActions.vue
```diff
@@ -1130,7 +1131,7 @@ export default {
 		 */
 		boundariesElement: {
 			type: Element,
-			default: () => document.querySelector('body'),
+			default: () => document.querySelector('#app-content-vue') ?? document.querySelector('body'),
```
Antreesy, line 1134: Whoops, we didn't check, and that's breaking NcActions in the NcAppNavigation and NcAppSidebar.Shouldn't it be `#content-vue`, as in NcContent?

a | b
-- | --
imageimage

PR #5807, opened 2024-07-16
Change action longtext max-width

GVodyanov:
### ☑️ Resolves

- Fix https://github.com/nextcloud/calendar/issues/6142

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot from 2024-07-16 10-18-44 | Screenshot from 2024-07-16 10-16-50

### DIFF
--- a/src/assets/action.scss
+++ b/src/assets/action.scss
@@ -80,7 +80,7 @@
 		// long text area
 		&__longtext-wrapper,
 		&__longtext {
-			max-width: 220px;
+			max-width: calc(var(--default-clickable-area) * 5);
 			line-height: 1.6em;
 
 			padding: calc((var(--default-clickable-area) - 1.6em) / 2) 0;


marcoambrosini (approved):

susnux:
I do not agree here, this - as is - will look weird in a lot of places and only might look good on that specific menu.
What is the reason to shrink it in width? This will lead to many lines (keep other languages in mind!).

The reason is if you have text and input actions in the menu you now make the text much less wide which then looks bad next to each other:
Bildschirmfoto_20240716_104657

See how the text is now not even as wide as the label of the next entry (and the white space on the right compared to the input element). There is plenty horizontal space, so no need to make the menu hard to read by forcing new lines.

So this brings me back to the initial question: What is the reason? Because this might make sense in English where while you break the text it still keeps on 2 lines, but with other languages it is now 3 and could event 4 lines of text which is much worse to read.
If you really want to force the menu to be smaller, then I would add a new `menuClasses` prop, and add a `max-width` to that class just on mail.

marcoambrosini:
In my opinion these menus should only contain very simple actions, no input fields, and I see this as a good step towards that. That said, I'm also fine with the classes approach you suggest @susnux

susnux:
It depends on what is the goal. Thats why I asked that in the first place.

1. If the menus should generally made smaller, then we can already do so by checking the content for any input and reduce the width if no inputs but only buttons
2. If this is only about this very specific position in mail, then fix this in mail by allowing to define a class for the popover and set a mail specific width.

so cc @nimishavijay for this question.

---

In my opinion **1.** would be not a good idea, the current "refresh" makes a lot sense UX wise as information is more dense and you do not have e.g. to scroll a lot for all information.
But this change here is (from my point) the opposite:
It makes reading the menu much harder as easy single line menus are now forced to be two lines, keep in mind that there are some other languages that are even longer than English(!)
Especially because I see no reason, we have so many dead horizontal space on Nextcloud apps but our main target audience for web UI are desktop users, where 16:9 is common and nowadays even UHD displays are not that uncommon¹

Some example with this changes:

current files | with this limit
---|---
Screen Shot 2024-07-17 at 02 26 19|Screen Shot 2024-07-17 at 02 26 27
Screen Shot 2024-07-17 at 02 24 42|Screen Shot 2024-07-17 at 02 24 51
Screen Shot 2024-07-17 at 02 23 52|Screen Shot 2024-07-17 at 02 24 05
Screen Shot 2024-07-17 at 02 22 31|Screen Shot 2024-07-17 at 02 22 46
Screen Shot 2024-07-17 at 02 21 49|Screen Shot 2024-07-17 at 02 22 03
Screenshot 2024-07-17 at 02-19-12 Dateien - Nextcloud Daily|Screenshot 2024-07-17 at 02-18-56 Dateien - Nextcloud Daily


So if you take a full screen shot of files, the size of the popover does not matter compared to the overall horizontal space:
Screenshot 2024-07-17 at 02-35-40 Fichiers - Nextcloud Daily

GVodyanov:
@nimishavijay @marcoambrosini Do you agree with @susnux? I'm a little torn on what to do here

GVodyanov:
@jancborchardt Thanks for the clarification! Makes sense, closing

PR #5815, opened 2024-07-17
style(NcAppNavigationCaption): remove margin-bottom

GVodyanov:
### ☑️ Resolves

- Fix https://github.com/nextcloud/mail/issues/9835

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image| image

### DIFF
--- a/src/components/NcAppNavigationCaption/NcAppNavigationCaption.vue
+++ b/src/components/NcAppNavigationCaption/NcAppNavigationCaption.vue
@@ -242,7 +242,7 @@ export default {
 		// padding to align the name with the icon of app navigation items
 		padding: 0 calc(var(--default-grid-baseline, 4px) * 2) 0 calc(var(--default-grid-baseline, 4px) * 2);
 		margin-top: 0px;
-		margin-bottom: 12px;
+		margin-bottom: var(--default-grid-baseline);
 	}
 
 	&__actions {


src/components/NcAppNavigationCaption/NcAppNavigationCaption.vue
```diff
@@ -242,7 +242,6 @@ export default {
 		// padding to align the name with the icon of app navigation items
 		padding: 0 calc(var(--default-grid-baseline, 4px) * 2) 0 calc(var(--default-grid-baseline, 4px) * 2);
 		margin-top: 0px;
-		margin-bottom: 12px;
```
susnux, line 245: Does not work on Nextcloud 30 as we have default styles for headings, there should be no margin at all then use

```css
    margin-bottom: 0px;
```

But I would rather go with:
```css
    margin-bottom: var(--default-grid-baseline);
```

Because otherwise the text of the heading end directly on the outline of the next list entry.

GVodyanov: Done

GVodyanov:
image

Now we have this with `margin-bottom: var(--default-grid-baseline);`

marcoambrosini (approved):

st3iny (approved):
LGTM! Please squash your commits.

susnux:
Please always backport to next (there is a todo list in the PR template ;) ) and assign correct labels (`bug`, `enhancement` etc) for changelog generation :)

PR #5819, opened 2024-07-17
NcAppSettingsSection: make titles normal h3 size

GVodyanov:
### ☑️ Resolves

- Fix https://github.com/nextcloud/mail/pull/9879

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot from 2024-07-17 17-52-36 | image

Not sure why their size was limited in the first place

### DIFF
--- a/src/components/NcAppSettingsSection/NcAppSettingsSection.vue
+++ b/src/components/NcAppSettingsSection/NcAppSettingsSection.vue
@@ -66,7 +66,7 @@ export default {
 .app-settings-section {
 	margin-bottom: 80px;
 	&__name {
-		font-size: 20px;
+		font-size: 1.6em;
 		margin: 0;
 		padding: 20px 0;
 		font-weight: bold;


src/components/NcAppSettingsSection/NcAppSettingsSection.vue
```diff
@@ -66,7 +66,6 @@ export default {
 .app-settings-section {
 	margin-bottom: 80px;
 	&__name {
-		font-size: 20px;
```
susnux, line 69: We do not have font sizes on Nextcloud 28 so we need a meaningful value also for 28, otherwise it looks like text.

ChristophWurst:
@GVodyanov see https://github.com/nextcloud/server/issues/43393. Until recently, headings in Nextcloud were a bit random

src/components/NcAppSettingsSection/NcAppSettingsSection.vue
ChristophWurst: Do you know any trick to achieve this?

GVodyanov:
> @GVodyanov see https://github.com/nextcloud/server/issues/43393. Until recently, headings in Nextcloud were a bit random

So is this PR still relevant or should we wait for someone to come along and standardize this everywhere? @ChristophWurst

ChristophWurst:
I'd say it is still relevant. The tricky part is to find a solution that works with 28/29 and 30. This lib is compatible with both.

src/components/NcAppSettingsSection/NcAppSettingsSection.vue
susnux: Just enforce the size we use in Nextcloud 30 :sweat_smile:

GVodyanov: Should be `26px` on Nextcloud 30, so I guess we can hard code it here

src/components/NcAppSettingsSection/NcAppSettingsSection.vue
```diff
@@ -66,7 +66,7 @@ export default {
...
+		font-size: 26px;
```
susnux, line 69:
```suggestion
		font-size: 1.6em;
```

Sorry :see_no_evil: 
https://github.com/nextcloud/server/pull/46587

susnux (approved):
please adjust the size, otherwise fine :)

ChristophWurst (approved):

PR #5829, opened 2024-07-18
NcSelect: make same height and color as NcTextField

GVodyanov:
### ☑️ Resolves

- Fix https://github.com/nextcloud/calendar/issues/6155

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### DIFF
--- a/src/components/NcSelect/NcSelect.vue
+++ b/src/components/NcSelect/NcSelect.vue
@@ -1177,7 +1177,7 @@ body {
 	/* Override default vue-select styles */
 	min-height: var(--default-clickable-area);
 	min-width: 260px;
-	margin: 0;
+	margin: 0 0 var(--default-grid-baseline);
 
 	.select__label {
 		display: block;
@@ -1185,7 +1185,8 @@ body {
 	}
 
 	.vs__selected {
-		height: 32px;
+		height: calc(var(--default-clickable-area) - var(--default-grid-baseline));
+		margin: calc(var(--default-grid-baseline) / 2);
 		padding: 0 8px 0 12px;
 		border-radius: 18px !important;
 		background: var(--color-primary-element-light);
@@ -1194,10 +1195,15 @@ body {
 
 	.vs__search {
 		text-overflow: ellipsis;
+		color: var(--color-main-text);
+
+		&::placeholder {
+			color: var(--color-text-maxcontrast);
+		}
 	}
 
 	.vs__search, .vs__search:focus {
-		margin: 2px 0 0;
+		margin: 0;
 	}
 
 	.vs__dropdown-toggle {
@@ -1263,12 +1269,13 @@ body {
 
 	.vs__selected-options {
 		// If search is hidden, ensure that the height of the search is the same
-		min-height: 40px; // 36px search height + 4px search margin
+		min-height: var(--default-clickable-area);
 
 		// Hide search from dom if unused to prevent unneeded flex wrap
 		.vs__selected ~ .vs__search[readonly] {
 			position: absolute;
 		}
+		padding: 0 5px;
 	}
 
 	&.vs--single {
@@ -1332,6 +1339,6 @@ body {
 
 // Selected users require slightly different padding
 .user-select .vs__selected {
-	padding: 0 2px !important;
+	padding: 0 5px !important;
 }
 </style>
--- a/src/components/NcInputField/NcInputField.vue
+++ b/src/components/NcInputField/NcInputField.vue
@@ -384,6 +384,10 @@ export default {
 		padding-inline: calc(var(--input-padding-start) + var(--input-border-width-offset)) calc(var(--input-padding-end) + var(--input-border-width-offset));
 		padding-block: var(--input-border-width-offset);
 
+		&::placeholder {
+			color: var(--color-text-maxcontrast);
+		}
+
 		&:active:not([disabled]),
 		&:hover:not([disabled]),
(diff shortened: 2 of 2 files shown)


marcoambrosini (changes requested):
Hi @GVodyanov, the height of these elements needs to be --default-clickable-area, the input fields will also shrink a bit soon

GVodyanov:
> Hi @GVodyanov, the height of these elements needs to be --default-clickable-area, the input fields will also shrink a bit soon

That's what it already should be set to actually

src/components/NcSelect/NcSelect.vue
```diff
@@ -1263,12 +1265,14 @@ body {
 
 	.vs__selected-options {
 		// If search is hidden, ensure that the height of the search is the same
-		min-height: 40px; // 36px search height + 4px search margin
+		min-height: auto; // 36px search height + 4px search margin
```
marcoambrosini, line 1268: So this min-height is unrelated?

src/components/NcSelect/NcSelect.vue
```diff
@@ -1186,6 +1186,7 @@ body {
 
 	.vs__selected {
 		height: 32px;
+		margin: 0;
```
susnux, line 1189: This causes the selected elements to directly "dock" onto the top border of the select.

susnux (changes requested):
The color changes should only be applied to the placeholder, not the input element as the input text must stay main text color.

src/components/NcSelect/NcSelect.vue
GVodyanov: Yeah, that's the min-height of the dropdown with the options

src/components/NcSelect/NcSelect.vue
GVodyanov: Made it the same height so it doesn't matter now

GVodyanov:
I made the height of everything `--default-clickable-area` all the way through, does it seem ok?

marcoambrosini (approved):

src/components/NcSelect/NcSelect.vue
```diff
@@ -1194,15 +1195,19 @@ body {
...
 	}
 
 	.vs__search, .vs__search:focus {
-		margin: 2px 0 0;
+		margin: 0;
 	}
 
 	.vs__dropdown-toggle {
 		position: relative;
-		max-height: 100px;
```
susnux, line 1205: Why this change? This can lead to really long element instead of scrolling. (like if you select many tags or something)

src/components/NcSelect/NcSelect.vue
```diff
@@ -1262,13 +1267,14 @@ body {
 	}
 
 	.vs__selected-options {
-		// If search is hidden, ensure that the height of the search is the same
```
susnux, line 1265: Why removing the comment?

GVodyanov: Typo

src/components/NcSelect/NcSelect.vue
GVodyanov: Ah ok

susnux: This is changing the behavior a lot:

before | after
---|---
Screenshot 2024-07-22 at 15-25-07 Nextcloud Vue Style Guide|Screenshot 2024-07-22 at 15-24-58 Nextcloud Vue Style Guide

src/components/NcSelect/NcSelect.vue
susnux: It does now every selected element has no gap to the others:

before | after
---|---
Screenshot 2024-07-22 at 15-23-35 Nextcloud Vue Style Guide|Screenshot 2024-07-22 at 15-23-31 Nextcloud Vue Style Guide
Screenshot 2024-07-22 at 15-23-06 Nextcloud Vue Style Guide|Screenshot 2024-07-22 at 15-22-59 Nextcloud Vue Style Guide

susnux: So easiest: Just add back like 2px margin and make the height `--default-clickable-area - 2 * margin-size`.
For example:
```css
height: calc(var(--default-clickable-area) - var(--default-grid-baseline));
margin: calc(var(--default-grid-baseline) / 2);
```

GVodyanov: Yep, added it!

susnux (approved):

susnux:
Please remember to add correct labels:
1. State of the PR (e.g. `3. to review`)
2. But more important the type of the PR: `bug`, `enhancement` or `refactor` this is needed for changelog generation.

Also always remember to backport to *next* (there is a TODO section in the PR template for this :) ).

src/components/NcSelect/NcSelect.vue
```diff
@@ -1263,12 +1269,13 @@ body {
 
 	.vs__selected-options {
 		// If search is hidden, ensure that the height of the search is the same
-		min-height: 40px; // 36px search height + 4px search margin
+		min-height: var(--default-clickable-area);
```
Antreesy, line 1272: Selector specificity either isn't big enough or scoped enough here (see leaked from server and viewer):
image

Even with that applied, NcSelect height is 36px, not 34px

susnux: There is #5883

PR #5830, opened 2024-07-18
NcListItem: remove max width

GVodyanov:
### ☑️ Resolves

- Helps https://github.com/nextcloud/calendar/pull/6169#issuecomment-2236754615

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image| image

@miaulalala

### DIFF
--- a/src/components/NcListItem/NcListItem.vue
+++ b/src/components/NcListItem/NcListItem.vue
@@ -765,7 +765,6 @@ export default {
 
 .list-item-content__name {
 	min-width: 100px;
-	max-width: 300px;
 	flex: 1 1 10%;
 	font-weight: 500;
 }
@@ -831,6 +830,7 @@ export default {
 			justify-content: start;
 			gap: 12px;
 			min-width: 0;
+			max-width: 300px;
 		}
 		.list-item-content__details {
 			flex-direction: row;


GretaD:
this will break mail on list layout, most probably.

susnux (approved):
While the usage in the other PR does not make sense (there is no list so no list item shall be used), I think this makes sense.

I am not 100% sure about the 1-line design as now the name will be much wider but I think its ok(?)

susnux:
> this will break mail on list layout, most probably.

If the name width is important than we should adjust it here by adding the `max-width: 300px` only for the 1-line design :)

GVodyanov:
> this will break mail on list layout, most probably.

@GretaD I did some testing and to me it looks fine actually, worst case scenario if you have a really long title this happens:
image

But as @susnux said we can add the `max-width` there if needed, makes more sense to have it only there than everywhere I think.

GretaD:
> But as @susnux said we can add the `max-width` there if needed, makes more sense to have it only there than everywhere I think.

then lets do that please, add the max-width: 300 for the one line, and remove it for the rest.

GVodyanov:
> > But as @susnux said we can add the `max-width` there if needed, makes more sense to have it only there than everywhere I think.
> 
> then lets do that please, add the max-width: 300 for the one line, and remove it for the rest.

Done @GretaD :)

src/components/NcListItem/NcListItem.vue
```diff
@@ -381,7 +381,7 @@
 					<!-- Main content -->
 					<div class="list-item-content">
 						<div class="list-item-content__main">
-							<div class="list-item-content__name">
+							<div class="list-item-content__name" :class="{ 'limit-width': oneLine }">
```
susnux, line 384: 1. Please use BEM (so it would be `list-item-content__name--one-line`)
2. But this is not needed as the modifier is already set on the parent class  (see comment below)

src/components/NcListItem/NcListItem.vue
```diff
@@ -761,11 +761,14 @@ export default {
 
 .list-item-content__name {
 	min-width: 100px;
-	max-width: 300px;
 	flex: 1 1 10%;
 	font-weight: 500;
 }
 
+.limit-width {
```
susnux, line 768: Some thing like this:
```suggestion
.list-item--one-line {
	.list-item-content__name {
		max-width: 300px;
	}
}
```

GretaD (approved):

PR #5831, opened 2024-07-18
feat(NcAppNavigation): Provide consistent in-app search

susnux:
### ☑️ Resolves

`NcAppNaviation` now provides an optional in-app search when `show-search` is set. This allows apps which have in app filtering / search to use a consistent layout.

### 🖼️ Screenshots

[vokoscreenNG-2024-07-18_17-02-03.webm](https://github.com/user-attachments/assets/fe00d617-27ca-4ed5-9e61-8296492e2e66)

### 🏁 Checklist


src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -47,6 +144,16 @@ emit('toggle-navigation', {
...
 			<div class="app-navigation__body" :class="{ 'app-navigation__body--no-list': !$scopedSlots.list }">
+				<NcAppNavigationSearch v-if="showSearch"
+					:label="searchLabel || undefined"
+					:value="search"
+					@update:value="$emit('update:search', $event)">
+					<template #actions>
+						<!-- @slot Optional NcAction* to show next to the search input -->
+						<slot name="search-actions" />
+					</template>
+				</NcAppNavigationSearch>
```
ShGKme, line 155: Instead of adding an optional but concrete `NcAppNavigationSearch` here, can we provide a slot, so users can directly use the component, having more control on it?

This content could be the default

src/components/NcAppNavigation/NcAppNavigationSearch.vue
```diff
@@ -0,0 +1,110 @@
...
+			<template #trailing-button-icon>
+				<IconClose :size="20" />
+			</template>
+		</NcInputField>
+		<Transition name="slide-fade">
+			<NcActions v-if="hasActions"
+				v-show="showActions"
+				ref="actions"
+				class="app-navigation-search__actions"
+				:inline="1">
```
ShGKme, line 31: Maybe make it configurable too?

src/components/NcAppNavigation/NcAppNavigation.vue
susnux: The idea was to not give control there as it should always look the same, or do you mean something different?

Antreesy: We are now using two different NcActions components, which means two different popovers. Don't see how it can be implemented from inline NcActionButton

susnux: Ah I see, well then lets remove the actions and just provide the slot for optional actions?

susnux: done

ShGKme: > The idea was to not give control there as it should always look the same, or do you mean something different?

Having `NcAppNavigationSearch` exposed you still have the control but can also control `NcAppNavigationSearch`, work with its events, props, attributes.

susnux: I am still not sure of the use case, all props and events are proxied.
And attaching any attributes make no sense as you can only attach to the wrapper element, so only `data-`, `class` or `style` would make sense.
But the idea was to not have custom styling, so having no way to assign class or style is what is intended?

ShGKme:
search-focus

One problem we have fixed in Talk: search buttons are not directly accessible via keyboard.

susnux:
> One problem we have fixed in Talk: search buttons are not directly accessible via keyboard.

Fixed now

src/components/NcAppNavigation/NcAppNavigation.vue
susnux: I agree here and adjusted this in the latest commit :)

src/components/NcAppNavigationSearch/NcAppNavigationSearch.vue
```diff
@@ -0,0 +1,252 @@
...
+import NcInputField from '../NcInputField/NcInputField.vue'
+
+defineProps({
+	/**
+	 * Current search input
+	 */
+	value: {
+		type: String,
+		default: '',
+	},
```
ShGKme, line 168: can we call it `modelValue`? To have no changes with `next` here.

src/components/NcAppNavigationSearch/NcAppNavigationSearch.vue
```diff
@@ -0,0 +1,252 @@
...
+	/**
+	 * Text used to label the search input
+	 */
+	label: {
+		type: String,
+		default: t('Search in app…'),
+	},
```
ShGKme, line 175: What about?
```suggestion
	label: {
		type: String,
		default: t('Search…'),
	},
```

AndyScherzinger (approved):

src/components/NcAppNavigationSearch/NcAppNavigationSearch.vue
susnux: Hm, I just wanted to allow Vue 2 to also use `v-model`

src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -47,16 +47,19 @@ emit('toggle-navigation', {
 			:inert="!open || undefined"
 			@keydown.esc="handleEsc">
 			<div class="app-navigation__body" :class="{ 'app-navigation__body--no-list': !$scopedSlots.list }">
-				<!-- The main content of the navigation. If no list is passed to the #list slot, stretched vertically. -->
+				<!-- @slot For in-app search you can pass a `NcAppNavigationSearch` component as the slot content. -->
+				<slot name="search" />
+
+				<!-- @slot The main content of the navigation. If no list is passed to the #list slot, stretched vertically. -->
 				<slot />
```
ShGKme, line 54: Both slots are in a common flex container, and 

> If no list is passed to the #list slot, stretched vertically

having this, is it expected behavior, or we expect users to not use body together with the search?

scroll

ShGKme:
Is it supposed to be without the list in the docs?

image

src/components/NcAppNavigationSearch/NcAppNavigationSearch.vue
```diff
@@ -0,0 +1,252 @@
...
+						<template #icon>
+							<IconStar :size="20" />
+						</template>
+					</NcAppNavigationItem>
+					<NcAppNavigationItem name="Second navigation entry">
+						<template #icon>
+							<IconStar :size="20" />
+						</template>
+					</NcAppNavigationItem>
+				</template>
```
ShGKme, line 52: Is it supposed to be without the list in the docs?

image

susnux: I thought having the list there would make it more clearer?

ShGKme: I mean, there is the list in the example's code, but not rendered...

src/components/NcAppNavigationSearch/NcAppNavigationSearch.vue
```diff
@@ -0,0 +1,252 @@
...
+								</NcActionButton>
+							</NcActions>
+							<NcButton aria-label="Search globally" type="tertiary">
+								<template #icon>
+									<IconSearchGlobal :size="20" />
+								</template>
+							</NcButton>
+						</template>
+					</NcAppNavigationSearch>
+				<template #list>
```
ShGKme, line 41: the list is put to the search 👀

```suggestion
					</NcAppNavigationSearch>
				</template>
				<template #list>
```

src/components/NcAppNavigationSearch/NcAppNavigationSearch.vue
ShGKme: See new comment

src/components/NcAppNavigation/NcAppNavigation.vue
susnux: Fixed this is not expected.

src/components/NcAppNavigationSearch/NcAppNavigationSearch.vue
susnux: :eyes:

src/components/NcAppNavigationSearch/NcAppNavigationSearch.vue
susnux: Already fixed :sweat_smile:

src/components/NcAppNavigationSearch/NcAppNavigationSearch.vue
susnux: done

src/components/NcAppNavigationSearch/NcAppNavigationSearch.vue
```diff
@@ -0,0 +1,272 @@
...
+
+/**
+ * Handle close the search
+ */
+function onCloseSearch() {
+	emit('update:modelValue', '')
+	if (hasActions()) {
+		showActions.value = true
+		nextTick(() => actionsContainer.value.$el.querySelector('button')?.focus())
+	}
```
ShGKme, line 227: This adds a new focus but doesn't prevent the previous one.

So, when you press <kbd>Tab</kbd>, focus first moves to real next element and only then after `100ms` makes a new move to actions. But it still counts as 2 independent moves for a web-browser.

Not sure it will be a problem in apps. In docs you can see it as a page scroll.

ShGKme (approved):
Thanks! 💙

src/components/NcAppNavigationSearch/NcAppNavigationSearch.vue
susnux: But this is only called on button click, not on tab.

src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -243,6 +247,14 @@ export default {
 		margin-left: calc(-1 * min($navigation-width, var(--app-navigation-max-width)));
 	}
 
+	&__search {
+		width: 100%;
+	}
+
+	&__body {
+		overflow-y: scroll;
```
ChristophWurst, line 255: What's the idea behind this? It causes scrolling in Mail: 
image

Antreesy: replace `<template #default>` with `<template #search>`

susnux: > replace `<template #default>` with `<template #search>`

Only if for the search component

PR #5833, opened 2024-07-18
fix(NcAppSidebar): manage focus only after transition has finished

ShGKme:
### ☑️ Resolves

- Fix https://github.com/nextcloud/spreed/issues/12747

When NcAppSidebar appears, two things happen:
1. Transition animation with margin
2. Focus is moved to the sidebar

Moving focus scrolls container (`NcContent`) to the sidebar which is out of the container (in the process of moving from the right side). It results in weird animation when the margin changes the width + scroll of the container instead of the position of the sidebar.

P.S. `overflow: hidden` only hides the scrollbar and content, but the container is still scrollable.
`overflow: clip` is used to actually remove scrolling.

Slowed animation 1000 times to see the effect: sidebar moves left and right by 1 pixel.

slow

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
before | after

### 🚧 Tasks

- [x] Move focus management to transition hooks
  - Both return focus and focus trap

### Alternative solution

Use `overflow: clip` on `NcContent` to remove scrolling.

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppSidebar/NcAppSidebar.vue
+++ b/src/components/NcAppSidebar/NcAppSidebar.vue
@@ -962,8 +962,6 @@ export default {
 		},
 
 		open() {
-			this.toggleFocusTrap()
-
 			this.checkToggleButtonContainerAvailability()
 		},
 	},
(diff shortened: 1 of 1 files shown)


Antreesy (approved):
Tested

susnux (approved):

PR #5837, opened 2024-07-19
Add ripple effect to buttons

marcoambrosini:


marcoambrosini:
pausing for a few days to focus on things that are actually visible in screenshots

PR #5859, opened 2024-07-23
fix(NcChip): Ensure chips without icon do not reserve empty space for icon

susnux:
### ☑️ Resolves

If there is no icon then there should be no empty space.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot 2024-07-23 at 19-22-53 Nextcloud Vue Style Guide|Screenshot 2024-07-23 at 19-19-17 Nextcloud Vue Style Guide

### 🏁 Checklist

### DIFF
--- a/src/components/NcChip/NcChip.vue
+++ b/src/components/NcChip/NcChip.vue
@@ -8,7 +8,8 @@
 ```vue
 <template>
 	<div style="display: flex; gap: 8px; flex-wrap: wrap;">
-		<NcChip text="Notes.txt" :icon-path="mdiFile" />
+		<NcChip text="Notes.txt" />
+		<NcChip text="Files" :icon-path="mdiFile" />
 		<NcChip text="Color" type="tertiary" :icon-path="mdiPalette" />
 		<NcChip text="Current time" type="primary" :icon-path="mdiClock" no-close />
 	</div>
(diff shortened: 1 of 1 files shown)


src/components/NcChip/NcChip.vue
```diff
@@ -70,7 +71,7 @@ export default {
 
 <template>
 	<div class="nc-chip" :class="{ [`nc-chip--${type}`]: true, 'nc-chip--no-actions': noClose && !hasActions() }">
-		<span class="nc-chip__icon">
+		<span class="nc-chip__icon" :class="{ 'nc-chip__icon--empty': !hasIcon() }">
```
ShGKme, line 74: Why not `v-if="hasIcon()"` to not render it at all?

src/components/NcChip/NcChip.vue
```diff
@@ -232,6 +234,13 @@ const onClose = () => {
 		overflow: hidden;
 		height: var(--chip-size);
 		width: var(--chip-size);
+
+		&--empty {
+			flex: 1;
+			width: 0;
+			// Add some more space to the border
+			margin-inline-end: calc(2 * var(--default-grid-baseline));
```
ShGKme, line 242: Hmm, because of this, I guess.

But then I'd add the class on the chip itself to add the margin. Otherwise there is always `__icon` in the layout even when there is no actual icon but just an element for margin.

src/components/NcChip/NcChip.vue
susnux: well that seems to be cleaner, adjusted

ShGKme (approved):

Antreesy (approved):
LGTM

PR #5861, opened 2024-07-24
chore(NcAppSidebar): correctly define and document exposed CSS variables

ShGKme:
### ☑️ Resolves

- Although AppSidebar's and AppNavigation's toggle button paddings have the same value, they don't have a single source
- Replaced using navigation SCSS variable in the sidebar with the direct value
- Fixed docs for `--app-sidebar-offset`
- Added "Exposed CSS Variables" docs section

### 🖼️ Screenshots

image

### 🏁 Checklist


susnux:
Thank you this is a very nice thing :)

ShGKme:
Force pushed just a commit message typo fix

susnux:
@ShGKme you need to update cypress snapshots :see_no_evil:

ShGKme:
> @ShGKme you need to update cypress snapshots 🙈

I changed nothing 🥲

ShGKme:
Will do so later today

susnux:
> I changed nothing 🥲

Maybe a padding or something else moved like 1px?

ShGKme:
Ok, actually, it is relevant. Variables are defined on an element outside the `AppSidebar` which doesn't work for Cypress.

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -1250,7 +1250,7 @@ export default {
   inherits: true;
 }
 
-.content {
```
susnux, line 1246: If you use root than themening can not applied here!
Theming is only available from body upwards

ShGKme: 🙈

susnux: Had to learn this when developing *forms* and settings styles on root, but on root only the default theme variables are available.
If you have styles from other themes they are set on body, so the variables are not defined :pensive:

susnux:
Rebased onto `main`

PR #5863, opened 2024-07-24
feat(NcCounterBubble): add count and limit props instead of slot

ShGKme:
### ☑️ Resolves

- Fix https://github.com/nextcloud-libraries/nextcloud-vue/issues/5486

Currently, the maximal number is limited via ... ellipsis. It doesn't make sense for numbers. And it is broken on Nextcloud 30 styles.

image

In general, having a slot doesn't make much sense for me. As the component is supposed to show a count, it is better to work with it as a number, not any content.

Instead, add an explicit `count` humanize output using:

- Added `count` prop for the value
- Humanize output with localization by default
- Even humanized output from the default slot content, if a plain number is passed
- Mark default slot as deprecated in flavor of `count` or `NcChip`
- Added `raw` prop to disable this behavior
- Improved docs
- Added tests

#### With new `count` prop

English | Chinese
---|---
image | image

#### With old slot

image

#### Docs for styling

image

### 🏁 Checklist

### DIFF
--- a/src/components/NcCounterBubble/NcCounterBubble.vue
+++ b/src/components/NcCounterBubble/NcCounterBubble.vue
@@ -3,36 +3,186 @@
   - SPDX-License-Identifier: AGPL-3.0-or-later
 -->
 
- <docs>
+<docs>
 
-### Normal Counter
+### Default usage
 
-```
-<NcCounterBubble>314+</NcCounterBubble>
-```
+NcCounterBubble displays a number from the `count` prop in a bubble.
 
-### Outlined Counter (e.g team mentions)
+By default, the number is **humanized** according to Nextcloud user's locale setting. Humanization can be disabled via `raw` prop.
 
-```
-<NcCounterBubble type="outlined">314+</NcCounterBubble>
-```
-
-### Highlighted Counter (e.g direct mentions)
-
-```
-<NcCounterBubble type="highlighted">314+</NcCounterBubble>
-```
-
-</docs>
-
+```vue
 <template>
-	<div :class="counterClassObject"
-		class="counter-bubble__counter">
-		<slot />
-	</div>
+	<table>
+		<tr>
+			<th>count</th>
+			<th>default</th>
+			<th>raw</th>
+		</tr>
+		<tr v-for="num in numbers" :key="num">
+			<td>{{ num }}</td>
+			<td>
+				<NcCounterBubble :count="num" />
+			</td>
+			<td>
+				<NcCounterBubble :count="num" raw />
+			</td>
+		</tr>
+	</table>
 </template>
 
 <script>
+export default {
+	setup() {
+		return {
+			numbers: [1, 9, 75, 450, 1042, 1750, 1999, 14567, 14567890, 2000000008],
+		}
+	},
+}
+</script>
+
+<style scoped>
+table {
+	border-collapse: collapse;
+}
+
+th,
+td {
+	border: 1px solid var(--color-border);
+	padding: var(--default-grid-baseline) calc(var(--default-grid-baseline) * 2);
+}
+
+th {
+	color: var(--color-text-maxcontrast);
+}
+</style>
+```
+
+### Styles
+
+Use different styles for different types of counters.
+
+```
+<template>
+	<table>
+		<tr>
+			<th>type</th>
+			<th>counter</th>
+			<th>Usage example</th>
+		</tr>
+		<tr>
+			<td>'' (default)</td>
+			<td>
+				<NcCounterBubble :count="3" />
+			</td>
+			<td></td>
+		</tr>
+		<tr>
+			<td>outlined</td>
+			<td><NcCounterBubble :count="3" type="outlined" /></td>
+			<td>Team/group mentions</td>
+		</tr>
+		<tr>
+			<td>highlighted</td>
+			<td>
+				<NcCounterBubble :count="3" type="highlighted" />
+			</td>
+			<td>Direct mentions</td>
+		</tr>
+		<tr>
+			<td>outlined active</td>
+			<td class="active-like">
+				<NcCounterBubble :count="3" type="outlined" active />
+			</td>
+			<td>Same as "outlined", but in an "active" container</td>
+		</tr>
+		<tr>
+			<td>highlighted active</td>
+			<td class="active-like">
+				<NcCounterBubble :count="3" type="highlighted" active />
+			</td>
+			<td>Same as "highlighted", but in an "active" container</td>
+		</tr>
+	</table>
+</template>
+
+<style scoped>
+table {
+	border-collapse: collapse;
+}
+
+th,
+td {
+	border: 1px solid var(--color-border);
+	padding: var(--default-grid-baseline) calc(var(--default-grid-baseline) * 2);
+
+	&.active-like {
+		background-color: var(--color-primary-element);
+	}
+}
+
+th {
+	color: var(--color-text-maxcontrast);
+}
+</style>
+```
+
+### Custom content (deprecated)
+
+You can use the default slot to pass any custom content. If you pass a plain number to the default slot, without raw prop it will be humanized like via `count` prop.
+
+**DEPRECATED:** passing count via slot content is **deprecated** and will be removed in the v9. Prefer using `count` prop for numbers or [NcChip](#/Components/NcChip) component for a custom content.
+
+```vue
+<template>
+	<table>
+		<tr>
+			<th>content</th>
+			<th>default</th>
+			<th>raw</th>
+		</tr>
+		<tr v-for="num in numbers" :key="num">
+			<td>{{ num }}</td>
+			<td>
+				<NcCounterBubble>{{ num }}</NcCounterBubble>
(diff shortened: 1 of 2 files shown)


jancborchardt (changes requested):
Very nice! :)

Is it possible to have this be default and not a choice/prop? That is, if a counter bubble is used anywhere, it should never go up further than 1000.

> What about using 1K+ instead of 999+?

I'd prefer 999+ as that's more universally localized and a bit clearer that it's "a ton".

ShGKme:
> Is it possible to have this be default and not a choice/prop? That is, if a counter bubble is used anywhere, it should never go up further than 1000.

The limit is `999` by default. However, it will work only if you pass the number as a count value, not as a custom HTML content. 

Changing the behavior of the current custom content rendering would be more breaking and also dirty implementation-wise.

susnux:
> Is it possible to have this be default and not a choice/prop?

It is 999 by default, which in most cases makes sense, but sometime even 99 would be ok, so we should not hard code this but just have a meaningful default, no?

src/components/NcCounterBubble/NcCounterBubble.vue
```diff
@@ -8,27 +8,127 @@
...
+````
+
 </docs>
 
 <template>
 	<div :class="counterClassObject"
 		class="counter-bubble__counter">
-		<slot />
+		<!--
+			@slot The content with the count number.
```
susnux, line 126:
```suggestion
			@slot (deprecated) The content with the count number.
```

Because styleguide output is not that clear I think

susnux (approved):

src/components/NcCounterBubble/NcCounterBubble.vue
```diff
@@ -55,6 +155,29 @@ export default {
...
+
+		/**
+		 * The maximal count number to be display. If the count is higher then "max+" is displayed.
+		 * For example, with limit 999, 1000 will be displayed as "999+".
+		 * Only works if the count is set via the `count` prop and not via the default slot.
+		 * String value is parsed to a number.
+		 * Set 0 to disable the limit.
+		 */
+		limit: {
+			type: [Number, String],
```
susnux, line 177: why allow string? I mean I am fine with that as it easily allows `limit="99"` but I am curious if there is more behind that decision.

ShGKme: To make it easier to pass constants. But I guess, as it is always passed from data, it should be number anyway. So, it can be removed.

src/components/NcCounterBubble/NcCounterBubble.vue
ShGKme: It has strikethrough

susnux: > It has strikethrough

I saw this, but it was not clear that this is deprecated or what there is a strike through

src/components/NcCounterBubble/NcCounterBubble.vue
ShGKme: Removed

src/components/NcCounterBubble/NcCounterBubble.vue
```diff
@@ -75,9 +205,7 @@ export default {
 	font-size: calc(var(--default-font-size) * .8);
 	overflow: hidden;
 	width: fit-content;
-	max-width: var(--default-clickable-area);
```
Antreesy, line 78: I remember we discussed a `min-width` for single-digit values, to make it round

Antreesy (approved):
Rest is good!

src/components/NcCounterBubble/NcCounterBubble.vue
```diff
@@ -75,9 +205,7 @@ export default {
 	font-size: calc(var(--default-font-size) * .8);
```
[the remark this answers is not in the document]
susnux: (this is just a note for possible other unrelated TODOs)

marcelklehr:
I would like to stress that numbers well above 1000 are a regular occurence. I would like to keep support for strings to allow consumers to pass things like "12.5k" if they so choose

st3iny:
As far as I can see, you can still use the slot. It was only deprecated for now. Removing it is a breaking change and implies a major bump.

ShGKme:
@marcelklehr do you have an example, where `NcCounterBubble` is used in such a way?

marcelklehr:
I use it in the bookmarks app to show bookmarks counts for folders and menu items

jancborchardt:
Ok, then let’s do it as you suggested (and also as e.g. Twitter does for favs and retweets):
- "1K" from 1000 on
- "1M" from 1 million on
- Maximum 1 decimal point

Then it fulfills multiple requirements:
- Still keeps an actual count (what @marcelklehr was saying)
- Shows there are a ton of things (what the "999+" was communicating)
- Doesn’t change small digits in large numbers

Sounds good @ShGKme @marcelklehr @susnux?

susnux:
Sounds good for me :)

(While I personally would prefer a small `k` but a capital `M` to align with unit prefixes)

jancborchardt:
@susnux small "k" and capital "M" sounds good to me too, it seems like the standard. Not sure what the specific reason to Twitter using a capital K is.

ShGKme:
@jancborchardt so we remove this feature with limit and add switch to show number in this way instead, correct?

Anyway, I'm going to do it as a new feature with a new prop, not as a slot parser, is that ok? So developers can enable it by switching from slot to component, but the old slot is still available to not break anything.

ShGKme:
> @susnux small "k" and capital "M" sounds good to me too, it seems like the standard. Not sure what the specific reason to Twitter using a capital K is.

I'll leave it on localization. It depends on language, and, for example, in Chinese its 万 for 10_000.

en | zh-CN
---|---
image | image

jancborchardt:
@ShGKme yup, using localization is best. And the example looks great! :100:

ShGKme:
> The counts will then still be short due to the magnitude indicators, and no need for settings there.

I'd say, we still need to support old usage. First, some apps already pass `99+` or `1k` which I'd not like to parse these numbers to new format.

Second, some apps have a custom format, for example, Tables in Smart Picker:

image

jancborchardt:
@ShGKme ok, in that case if anything that’s not a number gets passed, it will just be displayed as is? And numbers are subjected to the shortening. Would that work?

ShGKme:
> Would that work?

Yes, with my tears =D
(A bit complex implementation)

ShGKme:
Before I polish it, @susnux What do you think about this solution?

src/components/NcCounterBubble/NcCounterBubble.vue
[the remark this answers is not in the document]
ShGKme: Added usage of new variable

ShGKme:
- Rebased onto `master`
- Changed behavior:
  - Instead of `+` and `limit` -- humanize using Intl compact format
  - Also humanize plain number from slot content

ShGKme:
@jancborchardt Updated, see screenshots

src/components/NcCounterBubble/NcCounterBubble.vue
```diff
@@ -5,34 +5,184 @@
...
+}
+</style>
 ```
 
-</docs>
+### Custom content
+
+You can use slot to pass any custom content.
 
+Note, if you pass a plain number to the default slot, without raw prop it will be humanized like via `count` prop. This feature is **deprecated** and will be removed in the future. Prefer using `count` prop for numbers with humanization.
```
susnux, line 134: Shall we deprecate the whole default slot?
As the counter function is done by the prop, using the slot without the counter handling feels like you want to use NcChip or NcUserBubble.

ShGKme: Done
image

jancborchardt (approved):
Awesome @ShGKme, sorry for the tears :D

src/components/NcCounterBubble/NcCounterBubble.vue
susnux: BTW just for my reasoning:
From my point of view it makes no sense that we have so many components supporting the same use case.
This one is specialized for *counting* / showing a *count*, so we should stick with that behavior.

If there are complains about this, we still could add a `unit` prop where you can put `n('row', 'rows')` and the result would be `1 row`, `99 rows`, `1k rows`. But lets see if this is needed.

susnux (approved):
Super cool!

Maybe it makes sense to provide `humanizeCount` as part of `@nextcloud/l10n`?

susnux:
(just needs to be rebased)

ShGKme:
> Maybe it makes sense to provide `humanizeCount` as part of `@nextcloud/l10n`?

I'm not sure. It's basically a browser feature, just a default compact format + getting locale from `l10n`.

ShGKme:
Rebased onto `master` and squashed

ShGKme:
> Maybe it makes sense to provide `humanizeCount` as part of `@nextcloud/l10n`?

Maybe makes sense to make it easier for developers

PR #5866, opened 2024-07-24
feat: add NcAppNavigationSettingsButton

ShGKme:
### ☑️ Resolves

- Add a common component to be used as a Settings button in apps for consistency
- Looks like `NcAppNavigationItem`, has padding.
- Has no additional logic

More advanced component is coming soon ✨

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image
image | image
image | image
image | image
image | image

### 🚧 Tasks

- [x] Mark old `NcAppNavigationSettings` as deprecated
- [x] Add new `NcAPpNavigationSettingsButton`

### 🏁 Checklist


src/components/NcAppNavigationSettingsButton/NcAppNavigationSettingsButton.vue
```diff
@@ -0,0 +1,55 @@
...
+const emit = defineEmits(['click'])
+</script>
+
+<style lang="scss" scoped>
+.app-navigation-settings-button {
+	padding: var(--default-grid-baseline);
+
+	&__text {
+		// Make the settings button less presentative, like app navigation items
+		font-weight: normal;
```
susnux, line 52: This feels every different to all other elements and is different to all current implementations, I would like to ask @nextcloud-libraries/designers first

src/components/NcAppNavigationSettings/NcAppNavigationSettings.vue
```diff
@@ -3,6 +3,12 @@
   - SPDX-License-Identifier: AGPL-3.0-or-later
 -->
 
+<docs>
+⚠️ This component is deprecated and will be removed in v9.
+
+Use `<NcAppNavigationSettingsButton>` with `<NcAppSettingsButton>` instead.
```
susnux, line 9: `NcAppSettingsButton` ? You men `NcAppSettingsDialog`?

ShGKme: Yes

src/components/NcAppNavigationSettingsButton/NcAppNavigationSettingsButton.vue
ShGKme: > This feels every different to all other elements and is different to all current implementations

I actually agree, and that's exactly what we were discussing yesterday.

But from yesterday's design review call:
- It should not be bold or secondary, otherwise it is too presentive, more than other items in the menu.
- It should be the same in all navigation menus, no matter if `NcAppNavigationItem` or `NcListItem` is used
- It should be aligned with navigation items (`NcAppNavigationItem`)

cc @jancborchardt

jancborchardt (commented):
Design-wise this looks very nice and consistent! Just some details:
- [ ] In Files it looked great previously, with same spacing between the 3 elements on the bottom (quota, deleted files and settings), now Settings moved a bit down, leaving a gap instead of being equidistant.
- [ ] The placement does not look exactly the same in all of the apps, if you compare all the "After" screenshots. In Files, Photos and Talk it is further down than in Calendar and Contacts (is this cause they don’t use the new 34px yet?)

src/components/NcAppNavigationSettingsButton/NcAppNavigationSettingsButton.vue
jancborchardt: To confirm – what @ShGKme listed is what I said the requirements are. The design in this PR fulfills all those requirements and looks most consistent of the possible solutions.

src/components/NcAppNavigationSettings/NcAppNavigationSettings.vue
susnux:
```suggestion
Use `<NcAppNavigationSettingsButton>` with `<NcAppSettingsDialog>` instead.
```

PR #5870, opened 2024-07-25
feat(NcListItem): provide additional slots for NcActions / extra actions customization

Antreesy:
### ☑️ Resolves

- Ref https://github.com/nextcloud/spreed/issues/12800
- Allows more specific customization of NcActions component's icon 
- Allows additional elements inside NcListItem

### 🖼️ Screenshots

image

### 🏁 Checklist

### DIFF
--- a/src/components/NcListItem/NcListItem.vue
+++ b/src/components/NcListItem/NcListItem.vue
@@ -268,6 +268,94 @@
 	</template>
 </NcListItem>
 ```
+### NcListItem with extra actions slot and customized icon
+
+The `extra-actions` slot can be used to add other interactive components.
+
+The `actions-icon` slot can be used to pass icon to the inner NcActions component.
+
+```vue
+<template>
+	<div>
+		<NcListItem
+			name="This is an active element with extra actions slot"
+			:bold="true"
+			:active="true"
+			:force-display-actions="true">
+			<template #icon>
+				<NcAvatar disable-menu :size="44" user="janedoe" display-name="Jane Doe" />
+			</template>
+			<template #extra-actions>
+				<NcButton type="primary">
+					<template #icon>
+						<IconCog :size="20" />
+					</template>
+				</NcButton>
+			</template>
+			<template #actions-icon>
+				<IconNoteText :size="20" />
+			</template>
+			<template #actions>
+				<NcActionButton>
+					Button one
+				</NcActionButton>
+				<NcActionButton>
+					Button two
+				</NcActionButton>
+				<NcActionButton>
+					Button three
+				</NcActionButton>
+			</template>
+		</NcListItem>
+		<NcListItem
+			name="This is an element with extra actions slot"
+			:bold="true">
+			<template #icon>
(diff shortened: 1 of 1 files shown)


susnux (commented):
Not really sure, either we:

1. allow any actions, then there should be only one slot and we deprecate the `actions` slot
2. Only allow actions like currently, but then probably we need to allow at least an icon slot

Antreesy:
> 2\. we need to allow at least an icon slot

And something like an `actionProps`, like inline e.t.c

susnux:
Thats why I would go with *1* for the future, no?

ShGKme:
> 1. allow any actions, then there should be only one slot and we deprecate the `actions` slot

Then it is more complex to use because we need to define some props and events of actions

ShGKme:
@susnux What about another way:
1. Add `actions-icon` slot just for icon
2. Add new `extra-actions` slot for additional buttons.

So any additional buttons are added not via inline options but via a new slot as a normal buttons?

susnux:
> @susnux What about another way:
> 
>     1. Add `actions-icon` slot just for icon

This I think makes really sense!

>     2. Add new `extra-actions` slot for additional buttons.

Do you have a use case for this on Talk? Otherwise I think it is not a good idea to have multiple buttons here.
But would be also ok with this.

ShGKme:
> Do you have a use case for this on Talk?

We have almost exactly the same component re-implemented for participants in Talk. We have additional buttons there in a lobby and with SIP:

image

src/components/NcListItem/NcListItem.vue
```diff
@@ -421,13 +514,17 @@
 						:primary="isActive || active"
 						:aria-label="computedActionsAriaLabel"
 						@update:open="handleActionsUpdateOpen">
+						<template #icon>
+							<!-- @slot Provide the custom icon for the right side quick menu -->
+							<slot name="actions-icon" />
```
ShGKme, line 519: To not pass anything to `NcActions` when there is nothing.

```suggestion
						<template v-if="$slots['actions-icon']" #icon>
							<!-- @slot Provide the custom icon for the right side quick menu -->
							<slot name="actions-icon" />
```

src/components/NcListItem/NcListItem.vue
```diff
@@ -898,12 +995,19 @@ export default {
...
 			flex: 0 0 auto;
 			align-self: center;
 			justify-content: center;
 			margin-left: 4px;
 		}
+
+		&__extra-actions {
+			display: flex;
+			align-items: center;
+			gap: 4px;
```
ShGKme, line 1009:
```suggestion
			gap: var(--default-grid-baseline);
```

ShGKme (approved):

susnux:
> We have almost exactly the same component re-implemented for participants in Talk. We have additional buttons there in a lobby and with SIP:

Ah ok makes sense.

(OT: I really do not like this component it is so complex to fit every use case even if it would be simpler to just create a just custom component for it, while most often use case is just"a simple listitem").

susnux (approved):
Please apply the comment by @ShGKme (template v-if) otherwise fine

PR #5871, opened 2024-07-25
NcListItem: add tags line and make title ellipsise later

GVodyanov:
### ☑️ Resolves

- Fix https://github.com/nextcloud/mail/issues/9838
- Fix https://github.com/nextcloud/mail/issues/9334

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | Screenshot from 2024-07-26 17-34-13

### DIFF
--- a/src/components/NcListItem/NcListItem.vue
+++ b/src/components/NcListItem/NcListItem.vue
@@ -467,28 +467,26 @@
 							</div>
 
 							<div class="list-item-inner-content__details">
-								<div>
-									<div v-if="showDetails" class="list-item-details__details">
-										<!-- @slot This slot is used for some details in form of icon (prop `details` as a fallback) -->
-										<slot name="details">{{ details }}</slot>
-									</div>
+								<div v-if="showDetails" class="list-item-inner-content__details__details">
+									<!-- @slot This slot is used for some details in form of icon (prop `details` as a fallback) -->
+									<slot name="details">{{ details }}</slot>
+								</div>
 
-									<!-- Counter and indicator -->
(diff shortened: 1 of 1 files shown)


susnux:
I am really not sure if we should extend this component again. It is already a Jack of all trades, so many features for single use cases are merged into it so this component has a very high complexity.
Maybe it would be just better for Mail to use a custom component, that could easily implement the features Mail needs, but no other app.

(writing this as I currently try to fix the component and its about 1k lines for a simple list entry component :eyes: )

ChristophWurst:
If we do a custom component in Mail we have to redo look and feel of the standard list item (like the actions menu that only shows on hover). I'd like to avoid that.

susnux:
> If we do a custom component in Mail we have to redo look and feel of the standard list item (like the actions menu that only shows on hover). I'd like to avoid that.

Well copy paste it and adjust from there? You can drop a lot of stuff, as well as the library variant could drop a lot of stuff.
Currently there are many things in this component only for mail that do not make any / much sense for other apps, like the one line layout or having that many slots for everything you could think of.
In most cases you just need a list entry component which should look consistent, but having a component that can do everything makes maintenance harder.

I mean my comments here are no blockers, it looks well done, but this is more a personal opinion on usability of components :)

GVodyanov:
Other fixes:
| After | Before |
| - | - |
| Screenshot from 2024-07-30 19-14-18 | Screenshot from 2024-07-30 19-15-28 |

GVodyanov:
No longer needed, see https://github.com/nextcloud/mail/pull/9928#issuecomment-2277724312

PR #5883, opened 2024-07-29
fix(NcSelect): Ensure component height is same as input element

susnux:
### ☑️ Resolves

Currently the element was 36px instead of 34px which looks weird if using multiple input elements side by side (caused by search input being 34px + 2px border).
Also there was jumping because there are two different border width, so we need to adjust the input height based on the outer border width and the margin of the selected options.

### 🏁 Checklist

### DIFF
--- a/src/components/NcListItemIcon/NcListItemIcon.vue
+++ b/src/components/NcListItemIcon/NcListItemIcon.vue
@@ -111,6 +111,7 @@ It might be used for list rendering or within the multiselect for example
 <template>
 	<span :id="id"
 		class="option"
+		:class="{ 'option--compact': avatarSize < defaultSize }"
 		:style="cssVars"
 		v-on="$listeners">
 		<NcAvatar v-bind="$attrs"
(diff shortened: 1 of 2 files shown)


GVodyanov:
Is it supposed to be `44px` though? Because that is what I'm seeing. Why that number?

susnux:
> Is it supposed to be `44px` though? Because that is what I'm seeing. Why that number?

Where do you see that number? It should be always `--default-clickable-area`, so:
On Nextcloud 28 and 29 it should be `44px` and on Nextcloud 30 it should be `34px`.

GVodyanov:
> > Is it supposed to be `44px` though? Because that is what I'm seeing. Why that number?
> 
> Where do you see that number? It should be always `--default-clickable-area`, so: On Nextcloud 28 and 29 it should be `44px` and on Nextcloud 30 it should be `34px`.

Ok yeah my bad, tested it on 30 and looks good. But we will have to update some styling in calendar @ChristophWurst as it breaks

GVodyanov (approved):

marcoambrosini (approved):

ShGKme:
Seem now to become an issue on 28/29

PR #5896, opened 2024-08-02
feat(NcPasswordField): add as-text prop to remove autocomplete

ShGKme:
### ☑️ Resolves

- Sometimes we need to show secret but editable value, for example, a secret token
- To hide we should use password field
- But it is not a password. Neither creating a new password, nor authentication. It should not be saved, generated or autocompleted.
- The problem - browsers have save/generate prompts for password fields. This behavior cannot be disabled with `autocomplete="off"`, browsers ignore it.

Proposal: add a new `as-text` prop that forces using `type="text"` and emulates password field view.

Note: `-webkit-text-security` is not a standard feature, but well supported.

### 🖼️ Screenshots

image

### 🚧 Tasks

- [x] Add `as-text` prop to force text type and prevent autocomplete and 

### 🏁 Checklist

### DIFF
--- a/src/components/NcPasswordField/NcPasswordField.vue
+++ b/src/components/NcPasswordField/NcPasswordField.vue
@@ -40,6 +40,10 @@ General purpose password field component.
 		<NcPasswordField :value.sync="text5"
 			:disabled="true"
 			label="Disabled" />
+
+		<NcPasswordField :value.sync="text6"
+			label="Secret token"
+			as-text />
 	</div>
 </template>
 <script>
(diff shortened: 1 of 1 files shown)


Antreesy (approved):
Didn't test, but looks good by the code changes and caniuse info

susnux:
image


I do not like this, but yes support seems to be good and I do not see any non overcomplex solution (e.g. hidden input and fake ui). Because other solutions like replacing the shown context with `***` is not a solution as in form submit it would be the wrong value.

susnux (approved):

ShGKme:
Yeah, I didn't find a better solution :(

PR #5899, opened 2024-08-02
feat(useHotKey): add composable for keyboard shortcuts

Antreesy:
### ☑️ Resolves

Moved from [nextcloud/spreed#12698](https://github.com/nextcloud/spreed/pull/12698/commits/91ab5f7204bc670aab2cc6169ce1bc2295bdbdbc)

Add a new `useHotKey` composable with respect Nextcloud Accessibility setting
Available modifiers:
- ctrl: Ctrl key is required
- alt: Alt key is required
- shift: Shift key is required
- prevent: Prevent default action
- stop: Stop propagation
- push: Trigger on keydown and keyup

### 🖼️ Screenshots

image
image

### 🚧 Tasks

- [ ] Define list of `interactive` elements

### 🏁 Checklist

### DIFF
--- a/docs/composables/useHotKey.md
+++ b/docs/composables/useHotKey.md
@@ -1,0 +1,133 @@
+<!--
+ - SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+This composable allows to use keyboard shortcuts in the application.
+It respects Nextcloud's value of ```OCP.Accessibility.disableKeyboardShortcuts``` parameter.
+
+### Usage
+```js static
+import { useHotKey } from '@nextcloud/vue/dist/Composables/useHotKey/index.js'
+
+const stopCallback = useHotKey(key, callback, options)
+```
+where:
+- `key`: string representing the keyboard key to listen to
+
+	See [KeyboardEvent.key Value column](https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values) for possible values
+- `callback`: a function to be called when the key is pressed. Before called, it will be checked whether keyboard shortcuts are disabled, or interactive element is currently focused, or whether options should be applied
+- `options`: options to be applied to the shortcut:
+  - `push`: whether the event should be triggered on both keydown and keyup
+  - `prevent`: prevents the default action of the event
+  - `stop`: prevents propagation of the event in the capturing and bubbling phases
+  - `ctrl`: whether the Ctrl key should be pressed
+  - `alt`: whether the Alt key should be pressed
+  - `shift`: whether the Shift key should be pressed
+- `stopCallback`: a callback to stop listening to the event
+
+### Playground
+
+```vue
+
+<template>
+	<div class="container">
+		<p class="description">Press <kbd>W</kbd> <kbd>S</kbd> <kbd>A</kbd> <kbd>D</kbd> keys to move the ball</p>
+		<div class="square">
+			<div class="circle" :style="{ left: `${circleX}px`, top: `${circleY}px` }"></div>
+		</div>
+		<p class="description">
+			Push <kbd>M</kbd> to highlight the red area
+			<button @click="stop">Stop listen to this event</button>
+		</p>
+		<div ref="push"
+			 class="push-square"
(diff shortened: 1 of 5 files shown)


src/directives/KeyStroke/index.js
```diff
@@ -0,0 +1,76 @@
...
+			target: disableKeyboardShortcuts ? null : document,
+		})
+		if (binding.modifiers.push) {
+			onKeyStroke(keys, eventHandler(binding.value, binding.modifiers), {
+				eventName: 'keyup',
+				target: disableKeyboardShortcuts ? null : document,
+			})
+		}
+	},
+}
```
ShGKme, line 74: I'd be against implementing this as a directive, because it is technically not a directive.

By design, directive defines node's behavior. Like `v-show` defines its styles, `v-model` it's value property and updating event and etc.

This directive is bind to an element, but has no relation with the element. It doesn't use the `el` argument. When it's used in a component, moving directive from one node to another change nothing.

It even doesn't have `unbind`, so when the element with the directive is destroyed, the directive is still working until the component is unmounted.

IMO, it either can be implemented as a composable (because it is a wrapper around `onKeyStroke` composable). Or it can be actually bind to an element, for example, if short key means "click on this element" and not "call any handler".

Antreesy: Had it rewritten to a simple library function wrapper (with custom modifiers basically).

Tested with `script setup`, and options API hooks. Works correctly, adds and removes listeners correctly

src/composables/useKeyStroke/index.js
```diff
@@ -0,0 +1,80 @@
...
+		dedupe: !modifiers.press,
+		target: modifiers.target ?? document,
+	})
+
+	if (modifiers.push) {
+		onKeyStroke(keys, eventHandler(callback, modifiers), {
+			eventName: 'keyup',
+			target: modifiers.target ?? document,
+		})
+	}
```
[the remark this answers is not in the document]
Antreesy: Good catch. Added a callback and example

docs/composables/usekeystroke.md
```diff
@@ -0,0 +1,137 @@
...
+- `callback`: a function to be called when the key is pressed. Before called, it will be checked whether keyboard shortcuts are disabled, or interactive element is currently focused, or whether modifiers should be applied
+- `modifiers`: modifiers to be applied to the shortcut:
+  - `target`: the HTML element (or Vue ref) to listen to the event (`document` by default)
+  - `push`: whether the event should be triggered on both keydown and keyup
+  - `press`: whether the event should be triggered, continuously generating events
+  - `ctrl`: whether the Ctrl key should be pressed
+  - `alt`: whether the Alt key should be pressed
+  - `shift`: whether the Shift key should be pressed
+  - `stop`: prevents propagation of the event in the capturing and bubbling phases
+  - `prevent`: prevents the default action of the event
```
ShGKme, line 30: Although `target`, `press` are present in Vueuse, I'd not add support for it until its required to not increate API we want to support.

ShGKme (commented):
What about making it less a `vueuse` function wrapper, but more a specifically "hot keys" utility so that:
- explicitly mark it as an utility for hotkeys
- do not rely on vueuse and change implementation if needed
- do not promise to support interface we actually don't need in hot keys

(discussed in the office)

ShGKme:
Very cool docs!

docs/composables/useHotKey.md
```diff
@@ -0,0 +1,135 @@
+<!--
+ - SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+Extended version of [VueUse onKeyStroke](https://vueuse.org/core/onKeyStroke/)
+
+This composable allows to use keyboard shortcuts in the application.
+It respects Nextcloud's value of ```OCP.Accessibility.disableKeyboardShortcuts``` parameter.
```
ShGKme, line 9:
```suggestion
This composable allows to use keyboard shortcuts in the application.
It respects Nextcloud's value of ```OCP.Accessibility.disableKeyboardShortcuts``` parameter.
```

ShGKme (approved):
Thanks! Works fine, great examples in docs

susnux (approved):

PR #5922, opened 2024-08-05
fix(NcListItem): use font-size-small for subname and details

marcoambrosini:
### ☑️ Resolves

- Fix #…

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
 Screenshot 2024-08-05 at 17 36 52 | Screenshot 2024-08-05 at 17 36 28

### DIFF
--- a/src/components/NcListItem/NcListItem.vue
+++ b/src/components/NcListItem/NcListItem.vue
@@ -877,6 +877,7 @@ export default {
 }
 
 .list-item-content__subname {
+	font-size: var(--font-size-small);
 	flex: 1 0;
 	min-width: 0;
 	color: var(--color-text-maxcontrast);
@@ -1012,6 +1013,7 @@ export default {
 
 	&-details {
 		&__details {
+			font-size: var(--font-size-small);
 			color: var(--color-text-maxcontrast);
 			margin: 0 9px !important;
 			font-weight: normal;
--- a/src/components/NcListItem/NcListItem.vue
+++ b/src/components/NcListItem/NcListItem.vue
@@ -18,7 +18,7 @@
 			:counter-number="123"
 			counterType="highlighted">
 			<template #icon>
-				<NcAvatar disable-menu :size="44" user="janedoe" display-name="Jane Doe" />
+				<NcAvatar disable-menu :size="42" user="janedoe" display-name="Jane Doe" />
 			</template>
 			<template #subname>
 				In this slot you can put both text and other components such as icons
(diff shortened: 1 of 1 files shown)


src/components/NcListItem/NcListItem.vue
```diff
@@ -1012,6 +1013,7 @@ export default {
 
 	&-details {
 		&__details {
+			font-size: var(--font-size-small);
 			color: var(--color-text-maxcontrast);
 			margin: 0 9px !important;
 			font-weight: normal;
```
susnux, line 1019: Can we group the font related rules?
```suggestion
			color: var(--color-text-maxcontrast);
			font-size: var(--font-size-small);
			font-weight: normal;
			margin: 0 9px !important;
```

ShGKme:
I'm not sure about this. `NcListItem`s are not limited in height like the top bar.

Could you also add examples with Talk and Mail to the PR?

marcoambrosini:
Mail will have its own component @ShGKme. Talk will look kinda like this 
Screenshot 2024-08-05 at 18 23 45

susnux:
> I'm not sure about this. `NcListItem`s are not limited in height like the top bar.

I think it looks nice to have the subline smaller, but I also think we should make this consistent with similar components like ActionLink

jancborchardt (commented):
I also like this, gives it some nice contrast.

Also very related is the style we had a while ago where the main line of e.g. Talk conversations or Mails were bold. Maybe bold is a bit extreme there, but we could go for semibold.

src/components/NcListItem/NcListItem.vue
```diff
@@ -877,6 +877,7 @@ export default {
 }
 
 .list-item-content__subname {
+	font-size: var(--font-size-small);
```
susnux, line 880:
```suggestion
	font-size: var(--font-size-small, 13px);
```

src/components/NcListItem/NcListItem.vue
```diff
@@ -1013,8 +1014,9 @@ export default {
 	&-details {
 		&__details {
 			color: var(--color-text-maxcontrast);
-			margin: 0 9px !important;
+			font-size: var(--font-size-small);
```
susnux, line 1017:
```suggestion
			font-size: var(--font-size-small, 13px);
```

susnux (commented):
I think this makes sense and I like the look. Just the fallback values are missing.

marcoambrosini:
outdated

PR #5924, opened 2024-08-05
fix(NcChip): adjust border radius to new design

marcoambrosini:
## 🏚️ Before
Screenshot 2024-08-05 at 18 32 48

## 🏡 After
Screenshot 2024-08-05 at 18 33 11

### DIFF
--- a/src/components/NcChip/NcChip.vue
+++ b/src/components/NcChip/NcChip.vue
@@ -190,7 +190,7 @@ const onClose = () => {
 <style scoped lang="scss">
 .nc-chip {
 	--chip-size: 24px;
-	--chip-radius: calc(var(--chip-size) / 2);
+	--chip-radius: var(--border-radius-small);
 	// Setup size of wrapper
 	height: var(--chip-size);
 	max-width: fit-content;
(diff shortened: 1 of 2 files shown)


susnux:
Should we really do this? The avatar is the main focus of the userbubble, at least from my point of view having it not rounded looks weird because of the now visible corners around the avatar

src/components/NcChip/NcChip.vue
```diff
@@ -190,7 +190,7 @@ const onClose = () => {
 <style scoped lang="scss">
 .nc-chip {
 	--chip-size: 24px;
-	--chip-radius: calc(var(--chip-size) / 2);
+	--chip-radius: var(--border-radius-small);
```
susnux, line 193: Does not work as this does not exist on Nextcloud 28 and 29, so the fallback to current design is needed.

marcoambrosini:
Good point @susnux, then let's keep userbubble as is and only change chip

ShGKme:
TBH I also like chips in pills layout more. They are small and often have rounded avatar as user bubble, for example, when they are used to select users.

marcoambrosini:
They don't look ok when mixed up with other elements. 
Screenshot 2024-08-12 at 15 50 38
Anything else preventing this prom getting in?

susnux:
> TBH I also like chips in pills layout more. They are small and often have rounded avatar as user bubble, for example, when they are used to select users.

Here on github it is pretty the same, GitHub is not rounded, maybe even less than our new design. Yet they use round chips for e.g. issue labels:
image

Here you can see un-rounded UI elements next to the chips:
image

marcoambrosini:
All right, let's close

PR #5933, opened 2024-08-06
fix(NcActions): Role `menu` needs a label assigned so label by trigger

susnux:
### ☑️ Resolves

See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role

The `menu` needs a label set, so lets add it similar to dialog role. In general we could consider always add it to the `ul` as this is pretty valid and has good browser support.

### 🏁 Checklist


Antreesy (approved):
:seal:

PR #5940, opened 2024-08-08
feat(NcAppNavigationItem): Add prop to show collapse icon

Pytal:
### ☑️ Resolves

Adds opt-in showCollapse to override internal collapsible state

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
+++ b/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
@@ -356,7 +356,7 @@ Just set the `pinned` prop.
 						<slot name="actions" />
 					</NcActions>
 				</div>
-				<NcAppNavigationIconCollapsible v-if="collapsible" :open="opened" @click.prevent.stop="toggleCollapse" />
+				<NcAppNavigationIconCollapsible v-if="showCollapse || collapsible" :open="opened" @click.prevent.stop="toggleCollapse" />
 
 				<!-- @slot Slot for anything (virtual) that should be mounted in the component, like a related modal -->
 				<slot name="extra" />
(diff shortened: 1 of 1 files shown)


skjnldsv (approved):

susnux:
What is the use case? So when do you want collapsible behavior but not the toggle?
Because if you want only children you can remove the `allow-collapse` prop.

Having a collapsible item without toggle seems very inconsistent to me and should probably go through a design decision so we do not end up with inconsistent app navigation design.

sorbaugh (approved):

jancborchardt (commented):
Fully agree with @susnux. What’s the use-case for this? Otherwise it seems like an unnecessary prop.

susnux:
To me it seems the use case should be:
if a files view is not yet loaded, do not show the sub views in the navigation.

I think is quite an edge case and adding a prop for that will lead to inconsistency, because people will misuse it.

But there are some other options:
1. Let the component respect `open=false` even when `allow-collapse` is set to `false` -> will be the same behavior as with this prop here.
2. Or simply do not render the list while the view is loading (see my comment here https://github.com/nextcloud/server/pull/47122/files?diff=unified&w=0#r1709647015 )

Pytal:
The use case is showing the collapse icon even when there are no children in the default slot yet so the user can click expand which triggers loading of the children

> 1. Let the component respect `open=false` even when `allow-collapse` is set to `false` -> will be the same behavior as with this prop here.

I think this sounds like the same as showCollapse as this would also need a new prop to force this and without a prop it would break existing design and show the collapse icon everywhere by default

susnux:
> The use case is showing the collapse icon even when there are no children in the default slot yet so the user can click expand which triggers loading of the children

But this is already possible? You do not need a prop for this, you just need to make sure the slot is available, like:
```vue
<NcAppNavigationItem ...>
    <template>
        <li style="display: none"></li>
    </template>
</NcAppNavigationItem>
```

PR #5948, opened 2024-08-09
fix(NcUserBubble): make it perfectly aligned circle

ShGKme:
### ☑️ Resolves

- With 1 digit in `NcUserBubble` the digit is not exactly in the middle
- The problem is that `1em` line height is bigger than actual digit height
- Usually it is not a problem, but with a circle it looks better than it's perfectly in the middle
- Using `cap` fixes the problem
- `cap` is not well supported, so using `1em` as a fallback

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image
image | image
Top padding is bigger than the bottom | Padding is almost equal

### 🏁 Checklist

### DIFF
--- a/src/components/NcCounterBubble/NcCounterBubble.vue
+++ b/src/components/NcCounterBubble/NcCounterBubble.vue
@@ -72,14 +72,15 @@ export default {
 
 <style lang="scss" scoped>
 .counter-bubble__counter {
+	--counter-bubble-line-height: 1em;
 	font-size: calc(var(--default-font-size) * .8);
 	overflow: hidden;
 	width: fit-content;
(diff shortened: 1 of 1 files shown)


susnux:
Just to note: `1cap` is the height of the capital letters. But digits often have an interleave (they are slightly bigger).

Can we not just make it min-width min-height 1em and center it? Then it is always perfect?

susnux (approved):
works, but see comment :)

ShGKme:
> and center it?

How?

ShGKme:
To center it, you need to know the size and the position of the character relative to the container

marcoambrosini (approved):

PR #5976, opened 2024-08-19
fix(NcRichText): modify MENTION_START regex to work on older MobileSafari versions

### LINKED ISSUE #5589: RichText causes errors on WebKit | Invalid regex
The regex for mentions is using lookbehind which is not supported on all iOS versions / Safari versions we support.
It only works on 16.4+ but we support 15+.

The problem is within the regular expression:
https://github.com/nextcloud-libraries/nextcloud-vue/blob/a2854e9cb99d0c14108180cbcee67f6240dfed70/src/mixins/richEditor/index.js#L32

arthurlockman:
Fixes RichText causes errors on WebKit | Invalid regex #5589

This regex performs the same function as the one used before, but does not make use of negative lookbehinds and therefore should work on older MobileSafari versions.

To test, I looked first at the original mentioned regex from nextcloud server to get an idea of what the intended valid and invalid captures were. I tested both the original and modified regexes for both the server one and the one used in this component to verify that the matching groups were the same. The overall match 1 (blue highlight) is slightly different however the capture group 1 (green highlight) has the same content across both so this should be compatible.

### ☑️ Resolves

- Fix #5589 
- Fix https://github.com/nextcloud/forms/issues/2163

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
original | modified original
original concatenated userid | modified concatenated userid

### 🏁 Checklist

### DIFF
--- a/src/mixins/richEditor/index.js
+++ b/src/mixins/richEditor/index.js
@@ -11,8 +11,8 @@ import stripTags from 'striptags'
 import Vue from 'vue'
 
 // Referenced from public function getMentions(): https://github.com/nextcloud/server/blob/master/lib/private/Comments/Comment.php
-// Beginning or whitespace. Non-capturing group within word boundary
-const MENTION_START = /\B(?<![^a-z0-9_\-@.'\s])/.source
+// Beginning or whitespace. Uses positive lookahead (to work on MobileSafari <16.4)
+const MENTION_START = /(?=[a-z0-9_\-@.'])\B/.source
 // Capturing groups like: @user-id, @"guest/abc16def", @"federated_user/user-id", @"user-id with space"
 const MENTION_SIMPLE = /(@[a-z0-9_\-@.']+)/.source
 const MENTION_GUEST = /@&quot;guest\/[a-f0-9]+&quot;/.source


arthurlockman:
To add: I manually edited the minified JS on one of the instances I administer to the new regex pattern and the forms issue I linked is resolved!

Antreesy:
Failed tests show that a whitespace (which is now a part of matching group) is trimmed

arthurlockman:
Roger roger, will fix.

arthurlockman:
Tests now pass with this new iteration. It's using positive lookaheads which have been supported in MobileSafari since the beginning. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Lookahead_assertion

src/mixins/richEditor/index.js
```diff
@@ -11,8 +11,9 @@ import stripTags from 'striptags'
 import Vue from 'vue'
 
 // Referenced from public function getMentions(): https://github.com/nextcloud/server/blob/master/lib/private/Comments/Comment.php
-// Beginning or whitespace. Non-capturing group within word boundary
-const MENTION_START = /\B(?<![^a-z0-9_\-@.'\s])/.source
+// Modified to work on MobileSafari <16.4 (see issue #5589: https://github.com/nextcloud-libraries/nextcloud-vue/issues/5589)
+// Beginning or whitespace.
```
Antreesy, line 15:
```suggestion
// Beginning or whitespace. Uses positive lookahead (to work on MobileSafari <16.4)
```
Since issue will be closed, we don't need to mention it (still would be trackable by Git blame / commit history)

Antreesy (approved):
Checked test examples with both regex rules, seems to return the same matching groups now. Thanks for the contribution!

Could you try to pull the latest master and rebase your PR on it? Codecov shouldn't be triggered here

ShGKme (approved):

src/mixins/richEditor/index.js
susnux: Well I think it makes sense to keep this to answer the question when reading the code: Why not negative lookbehind?

susnux (approved):

src/mixins/richEditor/index.js
Antreesy: I meant to keep the suggested line, not the link to resolved issue =) Thanks for rebasing

arthurlockman:
Thanks for rebasing! Glad I could help sort this out. It was affecting forms users on one of the instances I manage.

Iwios:
Hello can I manually modify files?
It's really blocking

susnux:
It is already fixed since a couple of versions, where do you experience this issue?

PR #6006, opened 2024-08-21
feat(NcDialog): Allow to catch `reset` event

susnux:
* Depends on https://github.com/nextcloud-libraries/nextcloud-vue/pull/6005

### ☑️ Resolves

Sometimes it is useful to also have a reset button, we already support the native type `reset` so this allows to catch the `reset` event.

### 🏁 Checklist

### DIFF
--- a/src/components/NcDialog/NcDialog.vue
+++ b/src/components/NcDialog/NcDialog.vue
@@ -95,6 +95,7 @@ Note that this is not possible if the dialog contains a navigation!
 			name="Choose a name"
 			:open.sync="showDialog"
 			@submit="currentName = newName"
+			@reset="newName = ''"
 			@closing="newName = ''">
 			<NcTextField label="New name"
 				placeholder="Min. 6 characters"
@@ -115,6 +116,10 @@ export default {
 			newName: '',
 			currentName: 'none yet.',
 			buttons: [
+				{
+					label: 'Reset',
+					nativeType: 'reset',
+				},
 				{
 					label: 'Submit',
 					type: 'primary',
@@ -244,7 +249,7 @@ export default {
 					<NcDialogButton v-for="(button, idx) in buttons"
 						:key="idx"
 						v-bind="button"
-						@click="handleButtonClose" />
+						@click="(_, result) => handleButtonClose(button, result)" />
 				</slot>
 			</div>
 		</component>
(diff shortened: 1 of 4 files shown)


src/components/NcDialog/NcDialog.vue
```diff
@@ -114,8 +115,14 @@ export default {
...
 			showDialog: false,
 			newName: '',
-			currentName: 'none yet.',
+			currentName: 'non yet.',
 			buttons: [
+				{
+					label: 'Reset',
+					nativeType: 'reset',
+					// returning `false` to prevent the dialog from closing
+					callback: () => false,
```
ShGKme, line 124: Should it be the default behavior for reset?

susnux: this makes sense!

src/components/NcDialog/NcDialog.vue
```diff
@@ -113,8 +114,12 @@ export default {
 		return {
 			showDialog: false,
 			newName: '',
-			currentName: 'none yet.',
+			currentName: 'non yet.',
```
ShGKme, line 117: non?

ShGKme (approved):

src/components/NcDialog/NcDialog.vue
susnux: no idea wanted to fix this right now :sweat_smile: 

```suggestion
			currentName: 'none yet.',
```

Antreesy (approved):

tests/unit/components/NcDialogButton/NcDialogButton.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+				nativeType: 'reset',
+				callback: () => true,
+			},
+		})
+		await wrapper.find('button').trigger('click')
+		expect(wrapper.emitted('click')).toHaveLength(1)
+	})
+
+	it('has a loading state while the callback is awaited', async () => {
+		const { promise, resolve } = Promise.withResolvers<void>()
```
Antreesy, line 97: jsdom probably would need to have it mocked

susnux: ... this happens if you use a too new Node version ...

I use Node 22 where this works :pensive:

ShGKme: So we just increase `devEngines` 😎

PR #6013, opened 2024-08-23
fix(NcCheckBoxRadioSwitch): do not toggle on a click on a link inside

DorraJaouad:
… a link

### ☑️ Resolves

In rich text > tasks list, check boxes can have links

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
link1| link-2

### 🏁 Checklist

### DIFF
--- a/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
+++ b/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
@@ -575,8 +575,8 @@ export default {
 		t,
 		n,
 
-		onToggle() {
-			if (this.disabled) {
+		onToggle(event) {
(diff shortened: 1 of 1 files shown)


ShGKme (approved):

ShGKme:
But generally speaking, it is not valid to put an interactive element into another interactive element, and it fails a11y check

DorraJaouad:
>  it is not valid to put an interactive element into another interactive element, and it fails a11y check

Agree 🙈 , maybe we need to alter component to be interactive on checkbox level only (at least needed for `NcRichText`)

Antreesy (approved):
🙈

susnux:
> But generally speaking, it is not valid to put an interactive element into another interactive element, and it fails a11y check

I was thinking the same. Task lists should probably not put the content inside the checkbox (if not only text)

PR #6021, opened 2024-08-28
fix(useHotKey): respect press of MacOS Cmd key as Ctrl key

Antreesy:
### ☑️ Resolves

- Fix https://github.com/nextcloud/spreed/issues/13155
- Regression from #5899

### 🖼️ Screenshots

No visual changes

### 🚧 Tasks

- [x] Test on MacOS: Should allow to copy text with Cmd+C, if hotkey for 'C' is assigned

### 🏁 Checklist

### DIFF
--- a/src/composables/useHotKey/index.js
+++ b/src/composables/useHotKey/index.js
@@ -5,6 +5,7 @@
 import { onKeyStroke } from '@vueuse/core'
 
 const disableKeyboardShortcuts = window.OCP?.Accessibility?.disableKeyboardShortcuts?.()
+const isMac = /mac|ipad|iphone|darwin/i.test(navigator.userAgent)
 
 /**
  * Check if event target (active element) is editable (allows input from keyboard) or NcModal is open
@@ -26,7 +27,8 @@ function shouldIgnoreEvent(event) {
 }
 
 const eventHandler = (callback, options) => (event) => {
-	if (!!options.ctrl !== event.ctrlKey) {
+	const ctrlKeyPressed = isMac ? event.metaKey : event.ctrlKey
+	if (ctrlKeyPressed !== Boolean(options.ctrl)) {
 		// Ctrl is required and not pressed, or the opposite
 		return
 	} else if (!!options.alt !== event.altKey) {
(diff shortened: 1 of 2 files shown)


src/composables/useHotKey/index.js
```diff
@@ -26,7 +27,7 @@ function shouldIgnoreEvent(event) {
 }
 
 const eventHandler = (callback, options) => (event) => {
-	if (!!options.ctrl !== event.ctrlKey) {
+	if (isMac ? (!!options.ctrl !== event.metaKey) : (!!options.ctrl !== event.ctrlKey)) {
```
ShGKme, line 30: Having a ternary operator with a condition inside a condition feels hard to read.

```suggestion
	const ctrlKeyPressed = isMac ? event.metaKey : event.ctrlKey
	if (!!options.ctrl !== ctrlKeyPressed) {
```

ShGKme: Or we can say that `meta` = `ctrl alias on Mac`, to allow both keys, like search does:
```suggestion
	if (!!options.ctrl !== (event.metaKey || event.ctrlKey)) {
```

Antreesy: Maybe the first idea then? on Windows, Win key counts as metaKey, so Win+C would (unexpectedly) work

ShGKme: > so Win+C would (unexpectedly) work

I'm shocked

src/composables/useHotKey/index.js
```diff
@@ -26,7 +27,8 @@ function shouldIgnoreEvent(event) {
 }
 
 const eventHandler = (callback, options) => (event) => {
-	if (!!options.ctrl !== event.ctrlKey) {
+	const ctrlKeyPressed = isMac ? event.metaKey : event.ctrlKey
+	if (!!options.ctrl !== ctrlKeyPressed) {
```
susnux, line 31: I do not like double negation for casting, what do you think about:

```suggestion
	if (ctrlKeyPressed !== Boolean(options.ctrl)) {
```

Antreesy: Fine with that)

susnux (approved):

PR #6054, opened 2024-09-05
fix(NcAppNavigationItem): align utils with actions and other components

GVodyanov:
### ☑️ Resolves

- Calendar sidebar alignment, no issue

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot from 2024-09-05 13-58-56 | Screenshot from 2024-09-05 13-58-07

### DIFF
--- a/src/assets/NcAppNavigationItem.scss
+++ b/src/assets/NcAppNavigationItem.scss
@@ -170,7 +170,7 @@
 	/* counter */
 	.app-navigation-entry__counter-wrapper {
 		// Add slightly more space to the right of the counter
-		margin-right: calc(var(--default-grid-baseline) * 3);
+		margin-right: calc(var(--default-grid-baseline) * 2);
 		display: flex;
 		align-items: center;
 		flex: 0 1 auto;


src/assets/NcAppNavigationItem.scss
```diff
@@ -170,7 +170,7 @@
 	/* counter */
 	.app-navigation-entry__counter-wrapper {
 		// Add slightly more space to the right of the counter
```
GretaD, line 172: this will change the counter space, did you test it with the counter? if it looks off?

susnux: Bildschirmfoto_20240905_160729

2 and 4 have a counter, 3 has actions

ShGKme:
Don't forget to add `bug`/`enhancement` tags and follow conventional commit style. It is used in the changelog generation

src/assets/NcAppNavigationItem.scss
GVodyanov: Screenshot from 2024-09-05 16-39-13

I would argue it looks better now though

GVodyanov: image

If they are moved more to the right, the actions become misaligned with `NcAppNavigationCaption`, so unless you have any better ideas @susnux I would keep it the way it is currently

susnux: That was no critic, just wanted to add screenshots for Gretas comment, I am fine with any of that :)

susnux (approved):

GretaD (approved):

PR #6083, opened 2024-09-15
fix(NcModal): make dark header configurable

mejo-:
Required to fix regression with Text in Viewer in Nextcloud 30, where we still want a white viewer header.

### 🏁 Checklist

### DIFF
--- a/src/components/NcModal/NcModal.vue
+++ b/src/components/NcModal/NcModal.vue
@@ -191,7 +191,10 @@ export default {
 		<div v-show="showModal"
 			ref="mask"
 			class="modal-mask"
-			:class="{ 'modal-mask--dark': dark || !closeButtonContained || hasPrevious || hasNext }"
+			:class="{
+				'modal-mask--opaque': dark || !closeButtonContained || hasPrevious || hasNext,
+				'modal-mask--light': lightBackdrop,
+			}"
 			:style="cssVariables"
 			role="dialog"
 			aria-modal="true"
@@ -200,7 +203,9 @@ export default {
 			tabindex="-1">
 			<!-- Header -->
 			<transition name="fade-visibility" appear>
-				<div class="modal-header" data-theme-dark>
+				<div class="modal-header"
+					:data-theme-light="lightBackdrop"
+					:data-theme-dark="!lightBackdrop">
 					<h2 v-if="name.trim() !== ''"
 						:id="'modal-name-' + randId"
 						class="modal-header__name">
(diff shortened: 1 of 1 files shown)


mejo-:
https://github.com/nextcloud/text/pull/6404 contains screenshots of the regression this fixes.

src/components/NcModal/NcModal.vue
```diff
@@ -455,6 +455,14 @@ export default {
 			default: false,
 		},
 
+		/**
+		 * Make the modal header appear dark.
+		 */
+		darkHeader: {
```
susnux, line 461: This should not only change the header but the overall backdrop.
Meaning in this case also the backdrop should be bright.

So probably nicer to have: `brightBackdrop` prop with default `false`

src/components/NcModal/NcModal.vue
```diff
@@ -568,6 +576,11 @@ export default {
 		nextButtonAriaLabel() {
 			return t('Next')
 		},
+		dataThemeDarkAttr() {
+			return this.darkHeader
+				? { 'data-theme-dark': true }
```
susnux, line 581: This also needs enforce `light` theme if set, because otherwise you end up with mismatching button colors again

src/components/NcModal/NcModal.vue
```diff
@@ -200,7 +200,7 @@ export default {
 			tabindex="-1">
 			<!-- Header -->
 			<transition name="fade-visibility" appear>
-				<div class="modal-header" data-theme-dark>
+				<div class="modal-header" v-bind="dataThemeDarkAttr">
```
susnux, line 203: For me it would make the component better readable to move this not into a computed but keep it in the template:

```suggestion
				<div class="modal-header"
					:data-theme-light="brightBackdrop"
					:data-theme-dark="!brightBackdrop">
```

skjnldsv:
Not sure this is the right approach though.
Wouldn't it be easier to have a dark/light Viewer entirely? Focusing on th eheader only seems like a confusing solution :thinking:

mejo-:
>  Wouldn't it be easier to have a dark/light Viewer entirely? Focusing on th eheader only seems like a confusing solution 🤔

I get your point. But we have the situation that in light mode we want dark header per default but light headers when opening Text files :upside_down_face:

src/components/NcModal/NcModal.vue
mejo-: Implemented now with @susnux

susnux (approved):
@skjnldsv it does allow to have a light backdrop, no only the header.
We tested this with text and pdf viewer which will correctly adapt based on the color scheme.

But also the image viewer stays always dark as expected:)

skjnldsv (approved):

PR #6095, opened 2024-09-16
fix(NcDateTimePicker): change styling to look more like the native picker

GretaD:
### ☑️ Resolves

- Fix # https://github.com/nextcloud/calendar/pull/6322

### 🖼️ Screenshots

🏚️ Before | 🏡 After
after
Screenshot from 2024-09-30 11-52-42

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/NcDateTimePicker/NcDateTimePicker.vue
+++ b/src/components/NcDateTimePicker/NcDateTimePicker.vue
@@ -311,7 +311,7 @@ export default {
 					// 0 = sunday, 1 = monday
 					firstDayOfWeek: getFirstDay(),
 				},
-				monthFormat: 'MMM',
+				monthFormat: 'MMMM',
 			}
 		},
 
@@ -478,6 +478,10 @@ export default {
 	}
 }
 
+.v-select.select {
+	min-width: 246px;
+	left: -8px !important;
+}
 // TODO: This should be scoped or targeted by a specific selector, but the NcSelect component does not allow this yet.
 .vs__dropdown-menu--floating {
 	// Higher z-index than the popover in which the NcSelect is located.
--- a/src/components/NcDateTimePicker/index.scss
+++ b/src/components/NcDateTimePicker/index.scss
@@ -2,8 +2,9 @@
  * SPDX-FileCopyrightText: 2020 Nextcloud GmbH and Nextcloud contributors
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
- 
+
  $cell_height: 32px;
+ $opacity_full: 1;
 
 @import 'vue2-datepicker/scss/index';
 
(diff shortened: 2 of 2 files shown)


src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -477,10 +477,53 @@ export default {
...
+.mx-datepicker-main .mx-table .cell.disabled {
+	border-radius: var(--border-radius-small);
+	background-color: var(--color-background-dark);
+	color: var(--color-text-maxcontrast);
+	opacity: 1;
+}
+
+.mx-datepicker-main .mx-table .cell {
+	border-radius: var(--border-radius-small);
+	min-height: 30px !important;
```
susnux, line 499: This should be probably either default clickable or clickable-small right?

GretaD: yes, and also, i need to fix the classes to be shown as BEM.

susnux: But that are internal classes, so no renaming to BEM possible, no?

GretaD: i meant:

```
main {
.cell {
.today {
.other.. {
}
 }
  }
   }
```

nimishavijay (changes requested):
Looks really good! Very big improvement when we compare it to the current design :) 
- [ ] Is it possible to hide the first row in this case? The topmost week that is shown should be the week containing the 1st of the month (eg. we are on September but the last week of August is also shown)
- [x] Is it possible to hide the double arrows which skip to the same month next year (`«` and `»`) since I imagine that functionality is not used very often. This would also provide more horizontal space for the month name :) 
- [x] As @susnux mentioned the size should be a variable (`--clickable-area-small` should work here)
- [x] Nice-to-have: Show full month name

GretaD:
> Looks really good! Very big improvement when we compare it to the current design :)
> 
> * [ ]  Is it possible to hide the first row in this case? The topmost week that is shown should be the week containing the 1st of the month (eg. we are on September but the last week of August is also shown)

Unfortunately, i cannot find a way, but i removed the background color, so it doesnt look too crowded.

The rest is done. I updated to screenshot

susnux (changes requested):
Looks nice, **but** it is no longer possible to select a year that is more in the past / future as the buttons are removed:

before | now
---|---
image|image


Also when selecting the month the year should be centered like before:

before | now
---|---
image|image

GretaD:
> Looks nice, **but** it is no longer possible to select a year that is more in the past / future as the buttons are removed:


ahhh thanks for testing it, @nimishavijay i guess we need the buttons :D

GretaD:
> Please also have a look at #6125. I missed that you were also working on the date picker design 🙈
> 
> Ref [nextcloud/calendar#6324](https://github.com/nextcloud/calendar/issues/6324)

we can merge them, no problem, the change is the same, with small difference, like the hover button state. Do you want to include mine at yours, or should i add yours here?

susnux (approved):

ChristophWurst (approved):

Issue #125, opened 2018-11-26
Automatically figure out contacts menu position


georgehrke:
![20348e10-2911-4464-b57a-f6f055228357](https://user-images.githubusercontent.com/1250540/49018629-2c8db500-f18c-11e8-9122-ead5adb32e5f.jpg)

You can't use the contactsmenu on avatars on the left side, because the popover menu always opens with the arrow in the upper right corner

PR #6128, opened 2024-10-02
fix: Don't require name prop in NcListItem

artonge:
As it is only required if the slot is missing.


artonge:
> One problem, it is used in `computedActionsAriaLabel`

So let's announce it as primarily used for accessibility, and also as fallback for the `name` slot?

ShGKme:
> > One problem, it is used in `computedActionsAriaLabel`
> 
> So let's announce it as primarily used for accessibility, and also as fallback for the `name` slot? And keep it mandatory?

I'd say, we should just simplify `computedActionsAriaLabel`. We don't need to duplicate location there, especially when it's long. It's a menu in a specific list item and that enough.

artonge:
> I'd say, we should just simplify `computedActionsAriaLabel`. We don't need to duplicate location there, especially when it's long. It's a menu in a specific list item and that enough.

So dropping `computedActionsAriaLabel`, using `actionsAriaLabel` directly, and making it required? Not sure what is the state regarding enforcing accessibility properties.

ShGKme (approved):

src/components/NcListItem/NcListItem.vue
```diff
@@ -641,7 +640,7 @@ export default {
 		 */
 		actionsAriaLabel: {
 			type: String,
-			default: '',
+			required: true,
```
susnux, line 643: I would consider this more or less a breaking change :eyes:  As this is more strict.
Maybe just add a note that this will be required in v9 and instead of making it required add:
```suggestion
			default: t('Actions'),
```

artonge: True, I would be fine if we merge that in v9 only. I don't think a default aria label would be a good thing, this would mess up accessibility.

Is v9 the `next` branch?

susnux: leaving it unset will trigger the default `NcActions` label which is also just `Actions` so this would also be fine

susnux: > Is v9 the `next` branch?

yes

artonge: Alright, changed branch to target next.

ShGKme: @artonge As @susnux mentioned, we don't need to require aria label here. `NcActions` has the default label already, and it works fine by default. So there is no need to make users always create a name for the menu when using `NcListItem` and make it a breaking change.

artonge: But having a default "Actions" label for accessibility is not really accessible, no?

ShGKme: > But having a default "Actions" label for accessibility is not really accessible, no?

Why not?

ShGKme: > have x elements with the same description on the same page

It is not a problem, they have different contexts. It is not an "Actions" button somewhere on the page. It is a list item, and withing the list item there is a menu button with actions called "Actions". In a general case it's enough.

When a person uses only a screen-reader with no screen, they go to this button throw the page and list items. If a person uses a screen together with assistive technologies, then they know with which part of the UI they interact.

Like if we have a normal button to delete a list item in the list item with a normal text content, we don't write text "Delete user X" or "Delete list item #125". We just call it a "Delete" button.

susnux: Best rule I got told: Both should have the same information (visual vs accessible).
So we have visual information: Actions button.
So accessible information "Actions" would be ok. The context is already provided by the wrapping element.

src/components/NcListItem/NcListItem.vue
```diff
@@ -632,7 +632,7 @@ export default {
 		 */
 		actionsAriaLabel: {
 			type: String,
-			default: '',
+			default: t('Actions'),
```
ShGKme, line 635: Doesn't change behavior, but to reduce duplication and reuse `NcActions`'s default value
```suggestion
			default: undefined,
```

artonge:
Targeting master again as there is no breaking change anymore

skjnldsv (approved):

PR #6133, opened 2024-10-07
[next] Add Storybook support

ShGKme:
WIP

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,41 +1,44 @@
 {
-	"globals": {
-		"EMOJIS": true,
-		"PRODUCTION": true,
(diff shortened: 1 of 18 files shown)


susnux:
- Adjusted for upcoming Storybook v9 breaking changes
  - storysource is deprecated and will be removed - replaced with docs add-on
  - (also webpack support will be removed but we already use vite)
- Fixed vite config for Storybook -> it needs to built stand alone assets -> we can not use a library config
- Fixed some issues with arg table not being rendered correctly
- Added some examples for NcChip

susnux:
Migrated to storybook v9

PR #6138, opened 2024-10-10
Create new NcDatePicker component

GVodyanov:
This would make a new component ONLY for selecting date, based on PrimeVue's https://primevue.org/datepicker/

### ☑️ Resolves

- Fix https://github.com/nextcloud/calendar/issues/6385


GVodyanov:
This isn't needed any more after the decision to use the native one

PR #6195, opened 2024-11-06
feat: Add `spawnDialog` function

susnux:
### ☑️ Resolves

This allows to spawn a Vue dialog without the need of mounting the component.
So it is possible to spawn a dialog directly from any JS code outside of components (e.g. callback functions).

This method was originally taken from `@nextcloud/dialogs`, but I think it should belong to this Vue components library.

### 🖼️ Screenshots

grafik

### 🏁 Checklist

### DIFF
--- a/src/functions/dialog/index.ts
+++ b/src/functions/dialog/index.ts
@@ -1,0 +1,50 @@
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+import type { AsyncComponent, Component } from 'vue'
+
+import Vue, { toRaw } from 'vue'
+
+interface DialogProps {
+	[index: string]: unknown
+	container?: string
+}
+
+/**
+ * Helper to spawn a Vue dialog without having to mount it from a component
+ *
+ * @param dialog The dialog component to spawn
+ * @param props Properties to pass to the dialog
+ * @param props.container Optionally pass a query selector for the dialog container element
+ * @param onClose Callback when the dialog is closed
+ */
+export function spawnDialog(
+	dialog: Component | AsyncComponent,
+	props?: DialogProps,
+	onClose: (...rest: unknown[]) => void = () => {},
+): Vue {
+	const el = document.createElement('div')
+
+	const container: HTMLElement = typeof props?.container === 'string'
+		? (document.querySelector(props.container) || document.body)
+		: document.body
+	container.appendChild(el)
+
+	const vm = new Vue({
+		el,
+		name: 'VueDialogHelper',
+		render: (h) =>
+			h(dialog, {
+				props,
+				on: {
+					close: (...rest: unknown[]) => {
+						onClose(...rest.map(v => toRaw(v)))
+						vm.$destroy()
+						el.remove()
+					},
+				},
+			}),
+	})
+	return vm
+}
--- a/src/functions/index.ts
+++ b/src/functions/index.ts
@@ -4,6 +4,7 @@
  */
 
 export * from './a11y/index.ts'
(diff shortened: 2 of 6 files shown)


src/functions/dialog/index.ts
```diff
@@ -0,0 +1,37 @@
...
+
+	const vue = new Vue({
+		el,
+		name: 'VueDialogHelper',
+		render: (h) =>
+			h(dialog, {
+				props,
+				on: {
+					close: (...rest: unknown[]) => {
+						onClose(...rest.map(v => toRaw(v)))
```
ShGKme, line 30: Could you explain this line?

src/functions/index.ts
```diff
@@ -3,7 +3,8 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-export * from './a11y/index.ts'
-export * from './emoji/index.ts'
+export * from './a11y/index'
+export * from './dialog/index'
+export * from './emoji/index'
```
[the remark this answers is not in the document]
susnux: Still do not like non standard TS with extensions in imports / exports, but :shrug:

src/functions/dialog/index.ts
susnux: Originally this method was from the file picker, there passing `Node` as the payload caused passing the reactive object instead of the raw node instance.

src/functions/index.ts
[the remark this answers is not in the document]
ShGKme: > non standard TS with extensions

But there is no "standard" in TS saying to omit an extension. This is only a limitation when TS is used `tsc build` as a compiler.

I'd call it an old-school way when TS was used as a language. Since some time (starting `@babel/typescript-preset` to `esbuild` (incl. Vite), `tsx` and TS v5 in general) TS has been moving to be a static code analyze tool for typechecking almost like a linter.

When we use TS only for typecheck/declarations setting corresponding settings in tsconfig, it works fine.

On the other hand, there is a standard actually saying to keep the extension - ECMAScript standard. This is a standard, and it enforces ESM imports to have extensions. This is a requirement for web-browsers. Same for native TS support in runtimes, for example:
- Nodejs: https://nodejs.org/api/typescript.html#determining-module-system
- Deno: https://docs.deno.com/runtime/fundamentals/modules/#importing-modules

ShGKme (approved):

docs/functions/dialog.md
```diff
@@ -0,0 +1,109 @@
...
+
+```vue
+<template>
+	<div>
+		<NcButton @click="onSpawnDialog">🖱 click to spawn a dialog</NcButton>
+	</div>
+</template>
+
+<script>
+import { spawnDialog } from '@nextcloud/vue/dist/Functions/dialog.js'
```
ShGKme, line 42: Current module imports don't work in styleguidist (or we need to configure it somehow). 

This is why we added components/functions to globals like this

https://github.com/nextcloud-libraries/nextcloud-vue/blob/b7eeee8966b33e8f963b30d665b48ada2c2c6d48/styleguide/global.requires.js#L162-L171

st3iny (approved):

docs/functions/dialog.md
susnux: Hm in local styleguide it works?

ShGKme: It shouldn't...

Antreesy: I would prefer to stick to globals for now, even if it works another way. Thus, there will be less confusion when migrating to Storybook

ShGKme: I would prefer to have explicit imports to have complete examples. 

But they don't work and require additional setup. CI error is relevant.

susnux: Should be fixed

ShGKme (approved):

PR #6197, opened 2024-11-07
fix(NcCheckboxRadioSwitch): spacing to match NcButton + rtl support

skjnldsv:
Adjust the new 30 design with the ChecboxRadioSwitch.
- Also align the style to the NcButton so they look alike and can be used together
- Fixed rtl support for the component

## before/after
| Before | After | RTL after |
|:---------:|:------:|:------:|
||||

## Alignment
image


skjnldsv:
> I don't know whether it's desired, but now it's not aligned with the input field label and text anymore.

Considering the button isn't either, I assume we need to fix the input then :grin:

marcoambrosini (commented):
For this component we should just remove padding left and the input element should be in line with the beginning of the container's available space. 

A couple of examples of how this should look:
Screenshot 2024-11-08 at 10 30 48
Screenshot 2024-11-08 at 10 12 37

> Considering the button isn't either, I assume we need to fix the input then

I really think we should stop focusing on these micro alignments. Different elements have different indentations because they do different things and need different spacing. Besides, some irregularity in indentation helps with making it easier to tell elements apart and quickly scan the interface.

skjnldsv:
@marcoambrosini this isn't about the input.
This is about making sure the NcButton and the NcCheckboxRadioSwitch are aligned. This is not a "micro alignment", they are used together, and the misaligning stands out :)

Please see top screenshot

marcoambrosini:
I understand, but I think we should disregard this, because fixing it in this particular combination of components will inevitably push it out of place elsewhere, and I trying to fix all possible combinations is going to be very time consuming and very likely to fail.

Left aligning everything already provides some sense of order with minimal time investment. 

> Please see top screenshot

Yes, what I'm saying is that it should look like:

Screenshot 2024-11-08 at 12 07 23

---
In other words, this approach would avoid this kind of alignment issues:
Screenshot 2024-11-08 at 12 10 51

And ensure that things are tidy elsewhere too:
Screenshot 2024-11-08 at 12 10 04

marcoambrosini:
As another example, this is a form from google meet settings. Despite the components being very diverse and icons/text alignment being off, the left alignment of the components themselves provide a nice sense of order, with zero effort to keep track of where the icon sits or where the text sits in each component. 

Screenshot 2024-11-08 at 12 15 10

marcoambrosini:
@skjnldsv but what about the rtl part??

PR #6200, opened 2024-11-08
fix(NcActionButton): RTL support

### LINKED ISSUE #6199: [Bug] NcActionButton incorrect padding in RTL mode
Hello everyone,

This is an issue in the action buttons (NcActionButton) component. 

## Steps to reproduce:
Set interface language to RTL language such as Arabic.

Comparison between correct LTR mode padding and RTL. (In RTL padding stays on the right when it's supposed to go to the left):

✅  vs ❌ 

Padding should be on the left in RTL mode:
image

falghamdi125:
Hello
This PR fixes padding in RTL mode for NcActionButton.

### ☑️ Resolves

- Fix #6199 

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
 |  image
 |


src/assets/action.scss
```diff
@@ -75,6 +75,10 @@
 			.material-design-icon__svg {
 				vertical-align: middle;
 			}
+			body[dir=rtl] &.arrow-left-icon,
+			body[dir=rtl] &.chevron-right-icon {
+				transform: scaleX(-1);
+			}
```
ShGKme, line 81: Not sure about this part. It mirrors all the arrow-like icons in NcAction button, link, text elements.

ShGKme (approved):

src/assets/action.scss
falghamdi125: These two arrow icons are used to indicate a direction within the interface.

One way to resolve this issue is to modify the .vue file that sets the icons’ direction based on the interface mode. Alternatively, we can flip these icons horizontally using CSS, which is the approach I took.

susnux: I agree with @ShGKme 

Also I am not sure this should be handled by the library, because there might be use cases for left and right like the physical direction. It should probably be handled by the app instead.
Maybe we should either provide something like `NcIconArrowStart` and `NcIconArrowEnd` or a directive to flip if is RTL like `<IconArrayLeft v-rtl-icon />`.

hamza221 (approved):
I agree that the icon dir fix shouldn't be applied directly to body,
everything else looks good ✅

src/assets/action.scss
falghamdi125: What if we use NcPopover and NcActionButton to create something like NcContextMenu which handles submenus and draws arrows if menu item has children.

src/components/NcActionButton/NcActionButton.vue
```diff
@@ -547,7 +547,7 @@ export default {
 @include action-item('button');
 
 .action-button__pressed-icon {
-	margin-left: auto;
-	margin-right: calc($icon-margin * -1);
+	margin-inline-start: auto;
+	margin-inline-end: calc($icon-margin * -1);
```
susnux, line 551:
```suggestion
	margin-inline: auto calc($icon-margin * -1);
```

falghamdi125: Done

susnux (approved):

falghamdi125:
Hi,

I believe we need to add the left arrow in RTL mode using the isRTL function. I implemented this, and it successfully updates the icon based on the direction status.

image

@ShGKme  @hamza221  @susnux

falghamdi125:
Arrow direction change for RTL/LTR has been added to the PR! @ShGKme

PR #6209, opened 2024-11-12
feat(NcActions): submenus

ShGKme:
### ☑️ Resolves

- Fix: https://github.com/nextcloud-libraries/nextcloud-vue/issues/6066

Add submenu feature:
- `NcActionButton`'s `is-menu` is now not only for an icon, but actually to handle a specific menu
- `#submenu:NAME` slot is uses to pass a specific submenu

Takes into account:
- Multi-level menu stack
- Adding "Back" button to return from a submenu
- <kbd>Escape</kbd> closes submenu instead on an entire menu
- After opening a menu the first element is on focus
- After closing submenu the focus returns to the menu button

To discus:
- `submenu` vs `menu` name for slot

#### Docs

image

#### In action

NcActions-Submenu

### 🏁 Checklist

### DIFF
--- a/src/components/NcActionButton/NcActionButton.vue
+++ b/src/components/NcActionButton/NcActionButton.vue
@@ -387,6 +387,10 @@ export default {
 			from: 'NcActions:isSemanticMenu',
 			default: false,
 		},
+		openSubmenu: {
+			from: 'NcActions:openSubmenu',
+			default: () => () => {},
+		},
 	},
 
 	props: {
@@ -409,11 +413,12 @@ export default {
 		},
 
 		/**
-		 * If this is a menu, a chevron icon will
-		 * be added at the end of the line
+		 * Weather the button opens a submenu
+		 * - Boolean value makes button looks like a submenu opener
+		 * - String value can be used to open a specific submenu slot in NcActions
 		 */
 		isMenu: {
-			type: Boolean,
+			type: [Boolean, String],
 			default: false,
 		},
 
@@ -518,6 +523,11 @@ export default {
 		 */
 		handleClick(event) {
 			this.onClick(event)
+
+			if (typeof this.isMenu === 'string') {
+				this.openSubmenu(this.isMenu)
+			}
+
 			// If modelValue or type is set (so modelValue might be null for tri-state) we need to update it
 			if (this.modelValue !== null || this.type !== 'button') {
 				if (this.type === 'radio') {
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -788,6 +788,39 @@ p {
 </style>
 ```
 
+## Submenus
+
+To create multi-level menus:
+- Add `<NcActionButton is-menu="NAME" />` where NAME is the name of the submenu
+- Pass submenu to a `submenu:NAME` slot
+
+```vue
+<template>
+	<NcActions>
+		<template #default>
+			<NcActionButton>Option 1</NcActionButton>
+			<NcActionButton is-menu="two">Option 2</NcActionButton>
+			<NcActionButton is-menu="three">Option 3</NcActionButton>
+			<NcActionButton>Option 4</NcActionButton>
+		</template>
+		<template #submenu:two>
+			<NcActionButton>Option 2.1</NcActionButton>
+			<NcActionButton is-menu="two-two">Option 2.2</NcActionButton>
+			<NcActionButton>Option 2.3</NcActionButton>
+		</template>
+		<template #submenu:two-two>
+			<NcActionButton>Option 2.2.1</NcActionButton>
+			<NcActionButton>Option 2.2.2</NcActionButton>
+		</template>
+		<template #submenu:three>
+			<NcActionButton>Option 3.1</NcActionButton>
+			<NcActionButton>Option 3.2</NcActionButton>
+			<NcActionButton>Option 3.3</NcActionButton>
+		</template>
+	</NcActions>
+</template>
+```
+
 ## NcActions children limitations
 
 `<NcActions>` is supposed to be used with direct `<NcAction*>` children.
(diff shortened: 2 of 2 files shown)


skjnldsv:
Design team told me some time ago that "back" should not be used, but rather the name of the parent option :)

ShGKme:
> Design team told me some time ago that "back" should not be used, but rather the name of the parent option :)

Files | Talk
---|---
image | image

Indeed. But looks weird for me. We are in a "Setting reminder menu", and the first menu button is "Set reminder". Click on "Ser reminder" sets nothing and goes back to file actions. This is also a11y issue.

src/components/NcActions/NcActions.vue
```diff
@@ -1407,6 +1445,22 @@ export default {
 			}
 		},
 
+		pushSubmenu(submenu) {
```
Antreesy, line 1448: Need a check here if slot provided. Otherwise that would be a breaking, if a developer accidentally (🙈) passed a String instead of boolean, like:
```vue
<NcActionButton :is-menu="hasSubmenu && activeMenu">Option 2</NcActionButton>
```
```js
const hasSubmenu = true
const activeMenu = 'option2'
```

2024-11-12_11h37_17

src/components/NcActions/NcActions.vue
```diff
@@ -1769,6 +1839,20 @@ export default {
 		 */
 		const menuActions = actions.filter(action => !inlineActions.includes(action))
 
+		if (this.submenuStack.length) {
+			const backButton = h(NcActionButton, {
+				on: {
+					click: () => {
+						this.popSubmenu()
+					},
```
Antreesy, line 1847: Since we have an arrow navigation ⬆️ ⬇️ , would it make sense to add ⬅️ listener to 'Back' button, and ➡️ listener to 'isMenu' buttons?

DorraJaouad: Yeah, I would agree to use arrows rather escape (I would expect it to close the menu itself and not just submenus)

src/components/NcActionButton/NcActionButton.vue
```diff
@@ -409,11 +413,12 @@ export default {
 		},
 
 		/**
-		 * If this is a menu, a chevron icon will
-		 * be added at the end of the line
+		 * Weather the button opens a submenu
```
DorraJaouad, line 416:
```suggestion
		 * Wether the button opens a submenu
```
:3

ShGKme: 🌦️

src/components/NcActions/NcActions.vue
DorraJaouad: Maybe add `has-menu` instead of reusing `is-menu` ?

src/components/NcActionButton/NcActionButton.vue
DorraJaouad: merde, that's how you know I was using my phone :p

skjnldsv:
> Indeed. But looks weird for me. We are in a "Setting reminder menu", and the first menu button is "Set reminder". Click on "Ser reminder" sets nothing and goes back to file actions. This is also a11y issue.

This was a direct request from @jancborchardt. Make sure we get his approval before changing that behaviour please :)

src/components/NcActions/NcActions.vue
ShGKme: I agree to add arrows, but I'd like to keep <kbd>Escape</kbd>. Unlike a classic multi-level menu which literally opens side by side, Nextcloud's submenus act more like a new menu. 

For example, in this state when "Next week" is under focus, it's quite non-obvious to me that "Arrow left" can close the menu. There is nothing "on the left", no arrows on the button. 

image

src/components/NcActionButton/NcActionButton.vue
ShGKme: Fixed

src/components/NcActions/NcActions.vue
Antreesy: A bit unusual to have 'left' keydown only on 'Back button' (works in Windows - but they open submenu next to menu, not replacing it), but more comfortable anyway

Antreesy (approved):
Tested, waiting for design decision

marcoambrosini:
Nice job standardising this! 

Agree that the word "back" works best in this case. "set reminder" sounds like another action, while all the button is doing is literally bringing us back. I agree with @ShGKme point

That said, I think this UI should only be a mobile fallback and we should have proper cascading menus like everybody else. It's way more intuitive to hover or tab through options than having to dive in and dive back out.

Examples

**Microsoft 365**
Screenshot 2024-11-13 at 10 46 41

**Google drive**
Screenshot 2024-11-13 at 10 46 11

**Slack**
Screenshot 2024-11-13 at 10 55 06

**Zoom**
Screenshot 2024-11-13 at 10 55 28

ShGKme:
Converting to draft during design discussions in the issue

marcoambrosini:
> Converting to draft during design discussions in the issue

@ShGKme I opened a separate issue for the cascading proposal, I think what you've done is still useful for mobile. Don't you think?

ShGKme:
> @ShGKme I opened a separate issue for the cascading proposal, I think what you've done is still useful for mobile. Don't you think?
> And even for desktop it improves the current state, so I wouldn't want to block it

I'm not sure, because if we go with the cascade, implementation should be different from both HTML (a11y) side and vue component interface.

Something like that
```vue
<NcActions>
  <NcActionButton>Option 1<NcActionButton>
  <NcActionButton>Option 2<NcActionButton>
  <NcActionSubmenu label="Option 3">
    <NcActionButton>Option 3.1<NcActionButton>
    <NcActionButton>Option 3.2<NcActionButton>
  <NcActionSubmenu>
</NcActions>
```

This would be cool to support both views (e.g. desktop/mobile switch). but we'd need to think about the implementation, especially with our quite complex `NcActions` component.

PR #6212, opened 2024-11-12
feat(NcRichText): highlight code syntax if language provided

Antreesy:
### ☑️ Resolves

- Fix https://github.com/nextcloud/spreed/issues/13687
- install rehype-highlight lib v7.0.1
- use when render markdown
- apply colors from github light/dark theming

Supported languages: https://github.com/wooorm/lowlight/blob/main/lib/common.js

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image
image | image
image | image

### 🚧 Tasks

- [x] Introduce new prop?
  - We don't support language syntax before, so `use-extended-markdown` without providing lang prefix would work as before
- [ ] Color themes?
  - I haven't found a better way to do it (see PR). Suggestions are welcome

### 🏁 Checklist

### DIFF
--- a/src/components/NcRichText/NcRichText.vue
+++ b/src/components/NcRichText/NcRichText.vue
@@ -65,7 +65,7 @@ textarea {
 ### Flavored Markdown
 
 This component can support [Github Flavored Markdown](https://github.github.com/gfm/).
-It adds such elements, as tables, task lists, strikethrough, and supports autolinks by default
+It adds such elements, as tables, task lists, strikethrough, and supports code syntax highlighting and autolinks by default
 
 It is also possible to make a rendered content interactive and listen for events
 
@@ -94,6 +94,17 @@ It is also possible to make a rendered content interactive and listen for events
 Table header | Column A | Column B
 -- | -- | --
 Table row | value A | value B
+
+---
+
+\`\`\`js
+const GenRandomId = (length) => {
+\treturn Math.random()
+\t\t.toString(36)
+\t\t.replace(/[^a-z]+/g, '')
+\t\t.slice(0, length || 5)
+}
+\`\`\`
 `,
 			}
 		},
@@ -105,7 +116,7 @@ Table row | value A | value B
 					return
 				}
 				let checkBoxIndex = 0
-				lines = this.text.split('\n')
+				let lines = this.text.split('\n')
 				for (let i = 0; i < lines.length; i++) {
 					if (lines[i].includes('[ ]') || lines[i].includes('[x]')) {
 						if (checkBoxIndex === index) {
@@ -291,6 +302,7 @@ See [NcRichContenteditable](#/Components/NcRichContenteditable) documentation fo
 </docs>
 
 <script>
+import { ref } from 'vue'
 import NcReferenceList from './NcReferenceList.vue'
 import NcCheckboxRadioSwitch from '../NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue'
 import { getRoute, remarkAutolink } from './autolink.js'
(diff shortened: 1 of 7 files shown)


Antreesy:
cc @juliusknorr for possible affect on Office apps

src/components/NcRichText/highlight.scss
```diff
@@ -0,0 +1,195 @@
...
+* Colors and class selectors are extracted from source code of:
+* - library: highlight.js (v11.10.0)
+* - light theme: highlight.js/styles/github.css
+* - dark theme: highlight.js/styles/github-dark.css
+* and reworked to use with Nextcloud dark and light theme
+*/
+
+@mixin highlight-light-colors {
+	--hljs-color: #24292e;
+	--hljs-background-color: #ffffff;
```
susnux, line 16: I think this should always be background dark, so we are consistent with text etc.

```suggestion
	--hljs-background-color: var(--color-background-dark);
```

Antreesy: I kept original colors as a fallback (see L83:L84)
Though it's never be used, could be also set directly here

src/components/NcRichText/highlight.scss
susnux, line 1: Can we scope all those classes to `.rich-text--wrapper` ? So it does not conflict with other apps classes, as highlight js is pretty common.
e.g. `:root` -> `.rich-text--wrapper`
`.hljs-doctag` -> `.rich-text--wrapper .hljs-doctag`
etc

(BTW crying seeing the broken BEM for `rich-text--wrapper` class name :cry: )

src/components/NcRichText/highlight.scss
susnux: Ah yes saw that now, so I am fine with both

juliusknorr:
> cc @juliusknorr for possible affect on Office apps

Nice one, all good from my side. Only relevant usage is comment rendering in deck, where this works as expected 👍

ShGKme (commented):
Nice!

But for `NcRichText` it costs about 6.6 kB css + **329.53 kB js** (+100%) in production build.

ShGKme:
We can load it async on the first code block render, but I don't know if it worth it.

@susnux What do you think?

susnux:
> We can load it async on the first code block render, but I don't know if it worth it.

Yes I think it makes sense, sadly that plugin forces to bundle languages.
For `text` we load the languages dynamically which reduces the size a lot.

ShGKme:
> > We can load it async on the first code block render, but I don't know if it worth it.
> 
> Yes I think it makes sense, sadly that plugin forces to bundle languages. For `text` we load the languages dynamically which reduces the size a lot.

Proposal:

1. Add module-scope reactive property aka `highlightAvailable`
2. Use `rehypeHighlight` if `highlightAvailable === true`
3. Add a function to load `rehypeHighlight` with dynamic `import()`
4. Inside `rehype2react` check for `code` block and init loading `rehypeHighlight` from p.3

In theory when it is loaded p.1 changes and triggers rerender in p.2.

Antreesy:
Added, see screenshot from styleguide network tab.

image

DorraJaouad (approved):
Tested, looks good

src/components/NcRichText/NcRichText.vue
```diff
@@ -399,6 +412,7 @@ export default {
 	data() {
 		return {
 			parentId: GenRandomId(5),
+			rehypeHighlightLoaded: false,
```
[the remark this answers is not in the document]
ShGKme: If works with the local state, but then each new `NcRichText` thinks there is no `rehypeHighlight` and init it's "loading", which immediately resolves and then requires a second rendering to render it with highlight.

src/components/NcRichText/NcRichText.vue
```diff
@@ -306,6 +318,21 @@ import rehype2react from 'rehype-react'
...
+const rehypeHighlightLoaded = ref(false)
+/**
+ * @param {Function} callback - callback function to execute after library is loaded
+ */
+function importRehypeLibrary(callback) {
+	import('rehype-highlight').then((module) => {
+		rehypeHighlight = module.default
+		callback()
+	})
+}
```
ShGKme, line 334: You can just return the result and use Promise API, incl. async/await without manually providing callback

ShGKme (approved):

Antreesy:
Squashed and added a unit test, rewrote Promise syntax to async/await  - no other changes

jclsn:
How would I test this?

ShGKme:
> Thanks, I'll just wait until it's shipped then.

If you want to test it in a specific app, you can also link it with `npm link`

jclsn:
I am using the docker installation of Nextcloud. There is no hurry for me.

PR #6224, opened 2024-11-16
fix(NcRichTest): Allow markdown parenthesis in URL_PATTERN

provokateurin:
The URL_PATTERN_AUTOLINK regex already allowed parenthesis from markdown, so the links were clickable, but there was no preview shown due to this regex not allowing them.

### ☑️ Resolves

- Fix https://github.com/nextcloud/spreed/issues/13756

### 🏁 Checklist


provokateurin:
I'm not sure which milestones and branches are used in this repo, so feel free to backport however you see fit.

nickvergessen (changes requested):
This needs extensive tests added to not break cases like:
```
Please check the website (https://nextcloud.com/)
```
```
Please check the website https://nextcloud.com/wikipedia_(has-braces)
```
And the likes. It was a long fight with usability of most users. So we really prefer broken previous over broken UX of normal users posting behaviour

susnux:
The problem here is also that something like this would wrongly be parsed:
`some text https://nextcloud.com#some)`

provokateurin:
TBH I don't understand why it would be a problem, since URLs must be encoded and therefore can't contain plain braces.

susnux:
> TBH I don't understand why it would be a problem, since URLs must be encoded and therefore can't contain plain braces.

Depends on the moment of parsing, no? Because if you copy paste an URL from the browser address bar it will contain the `)` as is.

nickvergessen:
> TBH I don't understand why it would be a problem, since URLs must be encoded and therefore can't contain plain braces.

Because day-to-day usage shows that they are not pasted encoded, users don't encode and we receive multiple reports in the week after a Talk release when someone changes this :see_no_evil: 

The problem is even a step further, there are URLs where the brackets are expected (most famous is wikipedia)
- `We are aware of the problem (https://github.com/nextcloud/talk-desktop/issues/1421)`
- `We are aware of the problem (already reported in https://github.com/nextcloud/talk-desktop/issues/1421)`
- `My city https://en.wikipedia.org/wiki/Stuttgart_(disambiguation)`

PR #6259, opened 2024-11-22
fix(NcRichText): async import remark-gfm library

### LINKED ISSUE #6246: [Enhancement] NcRichText: `remark-gfm` could be async loaded to reduce initial size
rehype-highligh is now async loaded, which reduces the initial component size by ~110kB

image

Did a quick test, remark-gfm is 40kB and could be extracted as well. Shall we do it?

_Originally posted by @Antreesy in https://github.com/nextcloud-libraries/nextcloud-vue/issues/6212#issuecomment-2479118408_

Antreesy:
### ☑️ Resolves

- Fix #6246
- `remark-gfm` is 39.5 kB in dev build, which is 30% of rehype-highlight library size (111kB which we're also importing async)
- `rehype-highlight` won't be imported for code blocks without language provided/recognised
- ⚠️ if all conditions are met, that would cause two re-renders of the first component

### 🖼️ Screenshots

No visual changes

### 🏁 Checklist

Signed-off-by: Maksim Sukharev <antreesy.web@gmail.com>

### DIFF
--- a/src/components/NcRichText/NcRichText.vue
+++ b/src/components/NcRichText/NcRichText.vue
@@ -305,13 +305,13 @@ See [NcRichContenteditable](#/Components/NcRichContenteditable) documentation fo
 import { ref } from 'vue'
 import NcReferenceList from './NcReferenceList.vue'
 import NcCheckboxRadioSwitch from '../NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue'
+import NcLoadingIcon from '../NcLoadingIcon/NcLoadingIcon.vue'
 import { getRoute, remarkAutolink } from './autolink.js'
 import { remarkPlaceholder, prepareTextNode } from './placeholder.js'
 import GenRandomId from '../../utils/GenRandomId.js'
 
 import { unified } from 'unified'
 import remarkParse from 'remark-parse'
-import remarkGfm from 'remark-gfm'
 import breaks from 'remark-breaks'
 import remark2rehype from 'remark-rehype'
 import rehype2react from 'rehype-react'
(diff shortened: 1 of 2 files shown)


src/components/NcRichText/NcRichText.vue
```diff
@@ -410,6 +419,12 @@ export default {
 	},
 	emits: ['interact:todo'],
 
+	setup(props) {
+		if (props.useExtendedMarkdown && !remarkGfmLoaded.value) {
+			importRemarkGfmLibrary()
+		}
+	},
```
ShGKme, line 426: This loads the lib only if `useExtendedMarkdown` was true initially. It doesn't support prop update.

src/components/NcRichText/NcRichText.vue
```diff
@@ -489,9 +504,11 @@ export default {
...
 
 						if (!tag.startsWith('#')) {
-							if (this.useExtendedMarkdown) {
-								if (tag === 'code' && !rehypeHighlightLoaded.value) {
-									importRehypeLibrary()
+							if (this.useExtendedMarkdown && remarkGfmLoaded.value) {
+								if (tag === 'code'
+									&& attrs?.attrs?.class?.includes('language')
+									&& !rehypeHighlightLoaded.value) {
+									importRehypeHighlightLibrary()
```
ShGKme, line 511: makes sense 👍

src/components/NcRichText/NcRichText.vue
```diff
@@ -321,12 +320,22 @@ import { RouterLink } from 'vue-router'
...
+let remarkGfm
+const remarkGfmLoaded = ref(false)
+/**
+ * Load 'remark-gfm' library when prop `useExtendedMarkdown` is truthy
+ */
+async function importRemarkGfmLibrary() {
+	const module = await import('remark-gfm')
+	remarkGfm = module.default
+	remarkGfmLoaded.value = true
+}
```
ShGKme, line 332: Just realized, we only need one variable here.

```suggestion
const remarkGfm = ref(null)
/**
 * Load 'remark-gfm' library when prop `useExtendedMarkdown` is truthy
 */
async function importRemarkGfmLibrary() {
	const module = await import('remark-gfm')
	remarkGfm.value = module.default
}
```

ShGKme (commented):
I'm not sure this is a good idea with `remark-gfm`.

If the connection is slow - it first renders the raw text value, and then it flashes to rendered Markdown.

Unlike `code => code with highlights` update, this is a huge visible change. Imagine a message with headings and lists, that is first a plain text and then formatted.

I'm not sure we need a dynamic module here. But if we do it, we should render content ONLY when `remark-gfm` is loaded, and render nothing during loading.

juliusknorr (approved):

PR #6359, opened 2025-01-08
fix(NcListItem): Fix padding in RTL mode.

### LINKED ISSUE #6358: [Bug] NcListItem padding in RTL mode
Hello everyone,

This is a UI RTL issue in the NcListItem component. 

## Steps to reproduce:
- Set UI language to arabic to activate RTL mode
- Click on the user avatar in the nav bar
- You can see that the needed padding is missing before the username (in RTL only)

## Screenshots

In RTL  ❌
image
in LTR ✅
image

## Version details
nextcloud version: master
nextcloud-vue version: master

falghamdi125:
### ☑️ Resolves

- Fix #6358 

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image |  image

### DIFF
--- a/src/components/NcListItem/NcListItem.vue
+++ b/src/components/NcListItem/NcListItem.vue
@@ -872,7 +872,9 @@ export default {
 	.list-item-content__details,
 	.list-item-details__details {
 		white-space: nowrap;
-		margin: 0 auto 0 0;
+		margin-top: 0;
+		margin-bottom: 0;
+		margin-inline:0 auto;
 		overflow: hidden;
 		text-overflow: ellipsis;
 	}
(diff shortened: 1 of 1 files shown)


falghamdi125:
> Thanks!
> 
> This fixes the problem from the screen, but the component still has some problems on RTL, for example, with details.
> 
> Could you also fix other places where `margin/padding` sets different `left/right` values in this component, so it has complete RTL support?

All set!

falghamdi125:
> Changing specificity changes the result.
> 
> 
> 
> Elements still has non-logical properties, and after mixing with new logical properties the results changes.
> 
> 
> 
> All the properties should be logical to avoid mix caveats.

I think it's resolved now.

susnux (approved):

PR #6372, opened 2025-01-13
fix(NcRichContenteditable): remove value linkify

ShGKme:
### ☑️ Resolves

NcRichContenteditable has a hidden feature: when a content is set outside (not inputted ⚠️), it's linkified.

- Origin is https://github.com/nextcloud-libraries/nextcloud-vue/pull/1669/
- But it doesn't specify motivation

@juliusknorr @mejo- Could you confirm you don't rely on this behavior in text/collectives?

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

### DIFF
--- a/src/mixins/richEditor/index.js
+++ b/src/mixins/richEditor/index.js
@@ -4,7 +4,6 @@
  */
 
 import NcMentionBubble from '../../components/NcRichContenteditable/NcMentionBubble.vue'
-import Linkify from '../../utils/Linkify.js'
 
 import escapeHtml from 'escape-html'
 import stripTags from 'striptags'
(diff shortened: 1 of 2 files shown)


juliusknorr:
We're not using that component in text/collextives

nickvergessen (approved):

skjnldsv:
That seemed like a good feature, why are we removing this now?

susnux:
> That seemed like a good feature, why are we removing this now?

Probably because there is no usecase and not documented?
Also my problem is more that the UX is different if inputted / content from outside, so consistent behavior (remove or both) or feature toggle would be good I think.

Antreesy:
We're actually using this mixin to render description:
image
image

Maybe make it optional, so that NcRichContentEditable doesn't render links inside of editable content, but presenting components do?

Antreesy (approved):
Non-blocking for Talk anymore

juliusknorr (approved):

PR #6396, opened 2025-01-18
feat(NcBlurHash): Add a blur hash component

susnux:
### ☑️ Resolves

This allows to include image previews / placeholders in apps.

Currently something similar is used for the files and for the photos app. So this can be used to de-duplicate code.

### 🖼️ Screenshots

🏚️ Image | 🏡 BlurHash
---|---
Screenshot 2025-01-18 at 22-34-54 Nextcloud Vue Style Guide|Screenshot 2025-01-18 at 22-35-04 Nextcloud Vue Style Guide

### 🏁 Checklist


src/components/NcBlurHash/NcBlurHash.vue
```diff
@@ -0,0 +1,115 @@
...
+	},
+
+	/**
+	 * This is normally not needed, but if this blur hash is not only intended
+	 * for decorative purpose, descriptive text should be passed for accessibility.
+	 */
+	label: {
+		type: String,
+		default: ''
+	}
```
Antreesy, line 69: Lint warnings here
```suggestion
		default: '',
	},
```

src/components/NcBlurHash/NcBlurHash.vue
```diff
@@ -0,0 +1,115 @@
...
+})
+
+const canvas = ref()
+
+// Draw initial version on mounted
+onMounted(drawBlurHash)
+
+// On resize we redraw the BlurHash
+useResizeObserver(canvas, drawBlurHash)
```
Antreesy, line 78: I'm not sure if we need to recalculate blurhash on resize. Isn't it scaling fine (with CSS properties)?

Antreesy (approved):
Looks good, can't tell a difference from current in-Talk solution:
image

src/components/NcBlurHash/NcBlurHash.vue
```diff
@@ -0,0 +1,115 @@
...
+	},
+})
+
+const canvas = ref()
+
+// Draw initial version on mounted
+onMounted(drawBlurHash)
+
+// On resize we redraw the BlurHash
+useResizeObserver(canvas, drawBlurHash)
```
susnux, line 78:
```suggestion
import { decode } from 'blurhash'
import { onMounted, ref, watch } from 'vue'
import { logger } from '../../utils/logger.ts'

const props = defineProps({
	/**
	 * The blur hash value to use.
	 */
	hash: {
		required: true,
		type: String,
	},

	/**
	 * This is normally not needed, but if this blur hash is not only intended
	 * for decorative purpose, descriptive text should be passed for accessibility.
	 */
	label: {
		type: String,
		default: '',
	},
})

const canvas = ref()

// Draw initial version on mounted
onMounted(drawBlurHash)
```

src/components/NcBlurHash/NcBlurHash.vue
susnux: done

src/components/NcBlurHash/NcBlurHash.vue
```diff
@@ -0,0 +1,111 @@
...
+		logger.error('Invalid BlurHash value')
+		return
+	}
+	if (canvas.value === undefined) {
+		// Should never happen but better safe than sorry
+		logger.error('BlurHash canvas not available')
+		return
+	}
+
+	const { height, width } = canvas.value
```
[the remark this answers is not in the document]
Antreesy: As I remember it, with different dimensions a blurred thumbnail was quite distorted and different from original image, than 32*32 canvas. At least that's what upstream lib is using in demo:
https://blurha.sh/
https://github.com/nextcloud/spreed/commit/0fe0f018f2d26f6ad74ad28c1dcb30728a600c53

susnux:
>Having all that it seems to me, a developer still has to implement manually even more, than this component does.

Not sure as this can also lead to overengineering.
I have updated the docs with a proper example on how to use it.

The user could have different use cases (maybe its not a placeholder but just the preview?), so what I could imagine:
Add `source` or `href` prop that is used to preload an image and if ready replace the canvas with an `img` tag with that `href`

susnux:
@ShGKme I pushed a commit implementing that feature

[Bildschirmaufnahme_20250120_222940.webm](https://github.com/user-attachments/assets/2091229f-def2-4c9b-8d16-740c0ee7e4ee)

susnux:
For all that want to check the preview netlify:
To make assets (the images work) the output dir has to be different than the input, so we cannot build into `styleguide`.
We need to fix this.

In the meantime you can just append `/build` on the URL:
https://deploy-preview-6396--nextcloud-vue-components.netlify.app/build/

src/components/NcBlurHash/NcBlurHash.vue
```diff
@@ -0,0 +1,260 @@
...
+	},
+
+	/**
+	 * Optional an image source to load, during the load the blur hash is shown.
+	 * As soon as it is loaded the image will be shown instead.
+	 */
+	src: {
+		type: String,
+		default: '',
+	},
```
artonge, line 166: Shouldn't this be handled by the parent component? The usage in Files and Talk is quite straightforward, but in Photos, we have a more involved rendering logic and this won't work.

src/components/NcBlurHash/NcBlurHash.vue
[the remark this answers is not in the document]
artonge: I think using the canvas size is the way to go, as we do not always render square images. As long as the original image ratio is preserved, then it is fine.

src/components/NcBlurHash/NcBlurHash.vue
```diff
@@ -0,0 +1,260 @@
...
+
+	const ctx = canvas.value.getContext('2d')
+	if (ctx === null) {
+		logger.error('Cannot create context for BlurHash canvas')
+		return
+	}
+
+	const imageData = ctx.createImageData(width, height)
+	imageData.data.set(pixels)
+	ctx.clearRect(0, 0, width, height)
```
artonge, line 232: Is it needed? I would have assumed that the cleared context will be overridden by the image data.

src/components/NcBlurHash/NcBlurHash.vue
susnux: But this could remove the need for that logic in Files and Talk.
Photos can still have their own logic using the "manual" step as show in the "manual usage" example

src/components/NcBlurHash/NcBlurHash.vue
susnux: Not if we use the canvas size

```suggestion
```

artonge:
> @artonge but it is not the canvas or image, but blurhash render size

Blurhash render size should keep the same ratio as the original image size, no?

susnux:
> but it is not the canvas or image, but blurhash render size

It is, because the blur hash is just (more or less) an encoded gradient.
You need to render it in the correct aspect and one thing that might be browser dependent is how scaling of such data happens. Because if you only render 32px with a gradient and use no interpolation while scaling it will look horrible.

So I think the correct way to go is to render the hash on the canvas size.

susnux:
BTW Wolt (the author of that algorithm) also does it like this in their react implementation:
https://github.com/woltapp/react-blurhash/blob/master/src/BlurhashCanvas.tsx

src/functions/preloadImage/index.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import PQueue from 'p-queue'
+
+const queue = new PQueue({ concurrency: 5 })
```
ShGKme, line 8: Why? 🤔

PR #6425, opened 2025-01-23
refactor(utils): Replace `GenRandomId` with `getRandomId`

susnux:
### ☑️ Resolves

- Use Typescript for correct type annotations
- Fix length problem (the old implementation sometimes did not yield the expected length, e.g. if a "short" random number was generated)¹
- Add unit test for the util

¹ Did some benchmarks: Just the method is now ~10% slower. But the old method delivered too short values for ~25% of all cases.

### 🏁 Checklist


src/components/NcRichText/NcRichText.vue
```diff
@@ -556,7 +556,7 @@ export default {
 							labelParts.splice(nestedNodeIndex)
 						}
 
-						const id = this.parentId + '-markdown-input-' + GenRandomId(5)
+						const id = this.parentId + '-markdown-input-' + getRandomId(5)
```
susnux, line 559: This is the default 
```suggestion
						const id = this.parentId + '-markdown-input-' + getRandomId()
```

src/components/NcTimezonePicker/NcTimezonePicker.vue
```diff
@@ -73,7 +73,7 @@ export default {
 		 */
 		uid: {
 			type: [String, Number],
-			default: () => `tz-${GenRandomId(5)}`,
+			default: () => `tz-${getRandomId(5)}`,
```
susnux, line 76: The default

```suggestion
			default: () => `tz-${getRandomId()}`,
```

src/utils/getRandomId.ts
```diff
@@ -0,0 +1,20 @@
...
+export function getRandomId(length = 5) {
+	length = Math.max(1, length)
+	let id = ''
+	do {
+		id += (Math.PI - Math.random())
+			.toString(36)
+			.replace(/^\d\./g, '')
+		id = id.replace(/^[^a-z]+/, '')
+	} while (id.length < length)
+	return id.slice(0, length)
```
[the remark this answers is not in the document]
susnux: > Math.random().toString(36).slice(2, len - 2)

No this is not working. The length of `Math.random().toString(36)` is randomly between 1 and something longer.
Meaing your example would yield `'id-'`. or `'id-1'` or similar.

As said my implementation here is even faster when generating valid ids with length (its like 3M ops/s).
So I do not see a performance issue.
I mean we can "optimize":

```suggestion
			.replace('.', '')
	} while (id.length < length)
	return 'nc-' + id.slice(0, length)
```

susnux:
> In the next branch I'd prefer to use useId: https://vuejs.org/api/composition-api-helpers.html#useid
> It is Vue "native" way, and it's stable, unlinke Math.random().

For apps yes but for the library we can not be sure that the user has setup the id prefix correctly.

One alternative would be to provide a custom `useId` (for internal use!) that outputs:
```
const sanitizedAppName = String(appName).replaceAll(/[^0-9a-z-_]/ig, '')
return `${sanitizedAppName}-${useId()}`
```

This could be used in the library while apps of course would just use `useId` from Vue.

src/utils/getRandomId.ts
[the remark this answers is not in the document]
susnux: Still there is a tiny change that Math.random() returns the decimals of PI then this will fail :rofl: 
(And it outputs a too short id in 1:40000000 tries (tried that) :wink:  )

src/utils/getElementId.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * SPDX-FileCopyrightText: 2018 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+import { useId } from 'vue'
+
+const sanitizedAppName = String(appName).replaceAll(/[^a-z0-9-_]/gi, '') || 'nc'
```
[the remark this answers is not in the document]
susnux: Better safe than sorry, apps are not published on NPM so there could be everything and I saw a lot.
So always sanitize values you do not control!

src/utils/getElementId.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Get the next element id for use with HTML / CSS.
+ * We can not be sure that the app has a prefix setup so we prefix ourself
+ */
+export function getElementId() {
+	return `${sanitizedAppName}-${useId()}`
+}
```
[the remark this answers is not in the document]
susnux: True. So we can not use it.

I think I got a working solution that solves the points:
- consistent valid ids
- guaranteed unique ids (random is not necessary unique)
  - Handle different nc-vue versions in same app (e.g. because of libraries)
- Fast (only increment an integer)

Antreesy:
Since you've merged it, please don't forget to replace GenRandomId: https://github.com/nextcloud-libraries/nextcloud-vue/pull/6454/files#diff-c5595ba3afc4a147c5e8b442fe46ed5f41af898aad982d3b732103fe0fedb9f3R149

susnux:
Adjusted @Antreesy should be good now.

Antreesy:
CI wouldn't agree with you 😆 

I changed NcDateTimePickerNative inbetween, and someone else touched NavigationSettings:
image

src/utils/getElementId.ts
```diff
@@ -0,0 +1,17 @@
...
+const instanceId = md5(`${appName}@${SCOPE_VERSION}`).slice(0, 5)
+let id = 0
+
+/**
+ * Get the next element id for use with HTML / CSS.
+ * We can not be sure that the app has a prefix setup so we prefix ourself
+ */
+export function getElementId() {
+	return `nc${instanceId}-${id++}`
+}
```
ShGKme, line 17: Hmm, ok, in this case it makes sense not to use `Math.random()` for a stable result.

But still I'm not fun of `md5` hash.

Can we sanitize the app name instead of hashing it? No dependency and more readable in the layout.

Antreesy: `window.crypto` `getRandomValues() or `randomUUID()` could be applied here instead, maybe?

ShGKme: The idea of `md5` here is to hash a constant value (and have a stable ID sequence). For a random value we can simply use `Math.random()`.

Crypto API has methods to generate hashes as well, but they are async.

ShGKme: I was thinking about something like
```js
appName
  .replace(/\s+/g, '-')
  .replace(/[^\w-]/g, '')
```

susnux: Still a problem if there is a library using a different nc-vue version and also providing components like nc-upload or nc-password confirmation.

I think I found a solution by using a global id?

ShGKme: > I think I found a solution by using a global id?

Ah, so simple and so elegant

src/utils/getElementId.ts
```diff
@@ -0,0 +1,14 @@
...
+// make sure the id is unique
+window._nc_vue_element_id = window._nc_vue_element_id ?? 0
+
+/**
+ * Get the next element id for use with HTML / CSS.
+ * We can not be sure that the app has a prefix setup so we prefix ourself
+ */
+export function getElementId() {
+	return `nc-vue-${window._nc_vue_element_id!++}`
+}
```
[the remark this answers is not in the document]
susnux: > Or useless, and we should just always use getElementId as IDRef directly?

I would say this.

ShGKme, line 14: Maybe it's just me, but `get*` feels like it only gets some defined value, not generates/calcualtes with a side effect.

What about `generateElementId`/`createElementId` (though, a little bit long)?

ShGKme (approved):

Antreesy (approved):
Fine with any name

src/utils/getElementId.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * SPDX-FileCopyrightText: 2018 Nextcloud GmbH and Nextcloud contributors
```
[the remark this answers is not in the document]
susnux:
```suggestion
 * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
```

src/utils/getElementId.ts
susnux: TBH: I do not care :sweat_smile: 
So will go with your suggestion.

ShGKme:
Backport to master?
It is not a breaking change

Issue #6384, opened 2025-01-14
Last call for `@nextcloud/vue v9` breaking changes


ShGKme:
Before releasing the final v9.0.0 and making it the `main` branch we should make sure that all required breaking changes are applied.

When possible, migration should be simple:
- Removed API must be marked deprecated in v8
- Changed API should have a compatible API in v8

## Current breaking changes

- Vue 2 -> Vue 3
- CJS dropped
  - ✅ Forward-compatible via ESM
- `dist` and `.js` extension dropped from import path
  - :white_check_mark: Forward-compatible by just using the new paths like `@nextcloud/vue/components/NcButton`.
- `NcSettingsSection`: `limitWidth` prop was removed
  - ✅ Deprecated in v8
- `NcAppSidebar`: `closing` and `opening` events were removed
  - ✅ Forward-compatible via `update:open`
- `v-model` related changes
  - ✅ Forward-compatible via:
    - https://github.com/nextcloud-libraries/nextcloud-vue/pull/6172
    - https://github.com/nextcloud-libraries/nextcloud-vue/pull/5418
- `#default` slot changed to `#icon` for the leading icon in `Nc*Field`
  - ✅ Forward-compatible via:
    - https://github.com/nextcloud-libraries/nextcloud-vue/pull/6399
- `isFullscreen` and `isMobile` are removed
  - ✅ Corresponding composables are available on both versions
- `exact` prop removed from VueRouter-related components
  - ✅ Comes from Vue Router migration

## Proposals with breaking changes

- https://github.com/nextcloud-libraries/nextcloud-vue/pull/6385
  - ✅ Deprecated in v8 and forward-compatible
- https://github.com/nextcloud-libraries/nextcloud-vue/pull/6349
  - ✅ Deprecated in v8 and forward-compatible
- https://github.com/nextcloud-libraries/nextcloud-vue/pull/6390
  - ✅ Can be manually applied in v8, forward-compatible
- https://github.com/nextcloud-libraries/nextcloud-vue/pull/6385
  - ✅ Deprecated in v8 and forward-compatible
- https://github.com/nextcloud-libraries/nextcloud-vue/pull/6452
  - ✅ Deprecated in v8 and forward-compatible
- https://github.com/nextcloud-libraries/nextcloud-vue/pull/6429
  - ✅ Deprecated in v8 (not used)
- https://github.com/nextcloud-libraries/nextcloud-vue/pull/6472
[...]

PR #6429, opened 2025-01-23
refactor!: Drop `clickOutsideOptions` mixin and refactor using components

susnux:
* For #6384 

### ☑️ Resolves

- Drop `clickOutsideOptions` mixin and refactor using components.
  - Refactored to use `<script setup>` with proper Typescript types and slot definitions.
  - Following components which used the mixin were refactored to not need it anymore:
    - `NcHeaderMenu`
    - `NcAppNavigationSettings`

### 🏁 Checklist


src/components/NcHeaderMenu/NcHeaderMenu.vue
```diff
@@ -90,244 +87,197 @@ export default {
...
+	 * Additional visually hidden description text for the menu
+	 * open button
+	 */
+	description?: string
+
+	/**
+	 * A query-selector or an array of query-selectors
+	 * to be ignored when clicking outside an element
+	 */
+	excludeClickOutsideSelectors?: string | string[]
```
susnux, line 153:
```suggestion
	excludeClickOutsideSelectors?: string | string[]
```

susnux:
> Is it required to remove the mixin? If not, could we separate PR with mixing removing and component rewriting?

No but the refactor is needed before removal. So after that it is not needed anymore, so we can do it directly.

st3iny (approved):
Looks good

ShGKme (approved):

PR #6430, opened 2025-01-23
refactor(useHotKey): migrate code to Typescript

susnux:
### ☑️ Resolves

- Also a fix if you need to listen to a key and really do not care if a modifier is pressed.

### 🏁 Checklist


src/composables/useHotKey/index.ts
```diff
@@ -57,15 +91,20 @@ const eventHandler = (callback, options) => (event) => {
...
- * @param {Function} callback - callback function
- * @param {object} options - composable options
+ * @param keysOrFilter - keyboard key or keys to listen to, or filter function
+ * @param callback - callback function
+ * @param options - composable options
  * @see docs/composables/usekeystroke.md
  */
-export function useHotKey(keysOrFilter, callback = () => {}, options = {}) {
+export function useHotKey(
+	keysOrFilter: string | string[] | ((e: KeyboardEvent) => boolean),
```
Antreesy, line 104: do not skip `true`, it's a specific value to listen to all kbd events
```suggestion
	keysOrFilter: true | string | string[] | ((e: KeyboardEvent) => boolean),
```

Antreesy (approved):
Otherwise looks solid)

src/composables/useHotKey/index.ts
```diff
@@ -0,0 +1,164 @@
...
+	/**
+	 * If set the callback is only executed if the alt key is (not-)pressed
+	 */
+	alt?: boolean
+}
+
+/**
+ * Check if event target (active element) is editable (allows input from keyboard) or NcModal is open
+ * If true, a hot key should not trigger the callback
+ *
```
susnux, line 42:
```suggestion
 *
```

src/composables/useHotKey/index.ts
```diff
@@ -0,0 +1,164 @@
...
+}
+
+/**
+ * @param keysOrFilter - keyboard key or keys to listen to, or filter function
+ * @param callback - callback function
+ * @param options - composable options
+ * @see docs/composables/usekeystroke.md
+ */
+export function useHotKey(
+	keysOrFilter: string | string[] | ((e: KeyboardEvent) => boolean),
```
susnux, line 104:
```suggestion
/**
 * Composable to use keyboard shortcuts in the application.
 * It respects the users accessibility configuration (opt-out shortcuts).
 *
 * @param keysOrFilter - keyboard key(s) to listen to, or filter function or pass `true` for listening to all keys
 * @param callback - callback function
 * @param options - composable options
 * @see docs/composables/usekeystroke.md
 */
export function useHotKey(
	keysOrFilter: true | string | string[] | ((e: KeyboardEvent) => boolean),
```

susnux:
> Breaking change

I drop this part but it means I cannot use this as any key requires a hard "with OR without" meta instead of "with AND / OR without". For me it makes no sense to have a tri-state (true / false / undefined) and then hard force one of both states making it impossible to dont care about that.
Use case: Listen on escape but dont care about any other meta key.

susnux:
Dropped the commit and clarified the docs about different behavior for `undefined` of `ctrl` and `shift` option.

susnux:
> Then it's easier to review on GitHub.

But only github other tools correctly show it as renamed (e.g. `gitk`)

susnux:
> In which case do you need to allow users to press CTRL/Alt/Option/Meta + Esc?

Its more about not forbidding to press any other key, in that case only the escape key matters.

ShGKme:
> @ShGKme any reason to backport refactor?

- Reduce the complexity of backporting fixes and features until most of the apps are migrated, especially with migration to `<script setup>`
- There is nothing blocking backport, most of the migration to TS refactorings are mostly v8 (Vue 2) compatible
- PR description says:
  > Also a fix

susnux:
Makes sense, was just curious about the reason here.

> Any reason to not backport

Except from work no ;)

ShGKme:
> here

If you mean this PR specifically, it's the first in the list sorted by activity 👀

Nothing special here specifically

susnux:
> If you mean this PR specifically, it's the first in the list sorted by activity 👀

Yes I mean this PR (I do not plan to backport any Typescript PR to v8)

PR #6452, opened 2025-01-23
refactor: Use boolean props with default value false

susnux:
### ☑️ Resolves

Provide `disableSwipe` instead of `allowSwipeNavigation` (`NcAppContent`) and `enableSwipe` (`NcModal`).

Boolean prop should always - when possible - have a default value of `false` to allow shorthand prop assignment in the template. Similar to native HTML boolean attributes:

```html
<!-- native HTML -->
<button disabled>
<!-- boolean Vue prop with default false -->
<NcModal disable-swipe>
```

### 🏁 Checklist

### DIFF
--- a/src/components/NcModal/NcModal.vue
+++ b/src/components/NcModal/NcModal.vue
@@ -409,13 +409,23 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+
 		/**
 		 * Enable swipe between slides
+		 * @deprecated Will be removed in next version - use `disableSwipe` instead
 		 */
 		enableSwipe: {
 			type: Boolean,
 			default: true,
 		},
+		/**
+		 * Disable swipe between slides
+		 */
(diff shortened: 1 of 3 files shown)


GretaD (approved):

src/components/NcModal/NcModal.vue
```diff
@@ -409,13 +409,23 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+
 		/**
 		 * Enable swipe between slides
+		 * @deprecated Will be removed in next version - use `disableSwipe` instead
```
Antreesy, line 415: Not blocking, but don't forget to replace it for `next`
https://github.com/nextcloud-libraries/nextcloud-vue/blob/28860a17bb5024ec2edd2d8b1f9f24ff03b891dc/src/components/NcDialog/NcDialog.vue#L212

PR #6455, opened 2025-01-24
fix(NcAppNavigation + NcUserBubble + NcRichContenteditable): RTL support

DorraJaouad:
### ☑️ Resolves

- Many other components will need to be adjusted as well later.

Another issue to keep in mind: `NcRichContenteditable` is used with dir="auto" in Talk and is attached to the div with `"_input"` class. While we need to make a space for the emoji picker, the spacing is not adjusted to the RTL direction and it remains on the left side. We probably need to take the padding logic (and others) out of the `_input` class and put it in the wrapper class `rich-contenteditable` to follow the implemented language direction.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppNavigationCaption/NcAppNavigationCaption.vue
+++ b/src/components/NcAppNavigationCaption/NcAppNavigationCaption.vue
@@ -267,8 +267,8 @@ export default {
 		box-shadow: none !important;
 		flex-shrink: 1;
 		// padding to align the name with the icon of app navigation items
-		padding: 0 calc(var(--default-grid-baseline, 4px) * 2) 0 calc(var(--default-grid-baseline, 4px) * 2);
-		padding-right: 0;
+		padding-block: 0;
+		padding-inline: calc(var(--default-grid-baseline, 4px) * 2) 0;
 		margin-top: 0px;
 		margin-bottom: var(--default-grid-baseline);
 	}
--- a/src/assets/NcAppNavigationItem.scss
+++ b/src/assets/NcAppNavigationItem.scss
@@ -75,7 +75,7 @@
 
 	&:not(.app-navigation-entry--editing) {
 		.app-navigation-entry-link, .app-navigation-entry-button {
-			padding-right: $icon-margin;
+			padding-inline-end: $icon-margin;
 		}
 	}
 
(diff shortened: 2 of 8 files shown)


Antreesy:
> Another issue to keep in mind: `NcRichContenteditable` is used with dir="auto" in Talk and is attached to the div with `"_input"` class. While we need to make a space for the emoji picker, the spacing is not adjusted to the RTL direction and it remains on the left side.

But... I... https://github.com/nextcloud-libraries/nextcloud-vue/commit/1e5eadd45f3be107b705110b39fcbe60760b03de

> We probably need to take the padding logic (and others) out of the `_input` class and put it in the wrapper class `rich-contenteditable` to follow the implemented language direction.

Not sure that's wise. `_input` is rich element, with hover effects. wrapper is just to position optional label

Antreesy:
I'd also appreciate, if you move these changes to src/components/NcRichText/richtext.scss, so library styles are bidi-compliant, and not only Talk app:
 https://github.com/nextcloud/spreed/pull/14163/commits/b4e26ef03bbed01c586752b7a9d3479bcd06e653#diff-67dd2141faf250d624ec5cc487e5aef3ad38c666b339bf1e64dc76b49f87b340

DorraJaouad:
> dir="auto"

It's better to be kept, as it's user input and it will strange to write in the opposite direction. Also, other apps also have auto in the messages input.

Antreesy:
Then it should be handled upstream only if NcRichContenteditable will have a slot for leading / trailing icon.
Currently it's hacked in Talk, and therefore should be handled by Talk. As was discussed privately, we can work around paddings with CSS

src/components/NcAppNavigationCaption/NcAppNavigationCaption.vue
```diff
@@ -268,7 +268,7 @@ export default {
 		flex-shrink: 1;
 		// padding to align the name with the icon of app navigation items
 		padding: 0 calc(var(--default-grid-baseline, 4px) * 2) 0 calc(var(--default-grid-baseline, 4px) * 2);
-		padding-right: 0;
+		padding-inline-end: 0;
```
Antreesy, line 271:
```suggestion
		padding-block: 0;
		padding-inline: calc(var(--default-grid-baseline, 4px) * 2) 0;
```

Antreesy (approved):

PR #6466, opened 2025-01-27
fix(NcEmojiPicker): arrow navigation

ShGKme:
### ☑️ Resolves

- Part 1: https://github.com/nextcloud-libraries/nextcloud-vue/pull/6465
- Fix: https://github.com/nextcloud-libraries/nextcloud-vue/issues/1360
- `emoji-mart-vue-fast` has arrow navigation, but:
  - It doesn't work for a custom search field
  - It has a bug: scrolling to wrong positions and breaking on switch between sections

This PR:
- Adds arrow navigation using `emoji-mart-vue-fast` handlers with a custom fix
- Makes sure TAB navigation doesn't go over 1000 emojis

Note: selecting the category tab only scrolls and doesn't select the first emoji in the category. Also not implemented by upstream.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
arrow-before | arrow-after

### 🏁 Checklist

### DIFF
--- a/src/components/NcEmojiPicker/NcEmojiPicker.vue
+++ b/src/components/NcEmojiPicker/NcEmojiPicker.vue
@@ -139,7 +139,7 @@ This component allows the user to pick an emoji.
 			v-bind="$attrs"
 			@keydown.native.tab.prevent="handleTabNavigationSkippingEmojis"
 			@select="select">
-			<template #searchTemplate="slotProps">
+			<template #searchTemplate="{ onSearch }">
 				<div class="search__wrapper">
 					<NcTextField ref="search"
 						class="search"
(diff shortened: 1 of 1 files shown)


src/components/NcEmojiPicker/NcEmojiPicker.vue
```diff
@@ -391,47 +395,55 @@ export default {
...
-
-			// set focus on input search field
-			const input = this.$refs.search?.$refs.inputField?.$refs.input
-			if (input) {
-				input.focus()
-			}
+			this.$refs.search.focus()
 		},
 
 		afterHide() {
```
Antreesy, line 401: Can we focus whatever in trigger slot on close, to imitate focus-trap?

Antreesy (approved):
Nice and clean

src/components/NcEmojiPicker/NcEmojiPicker.vue
Antreesy: But it's not working in current state, so should be fine to follow-up separately

Antreesy (approved):
Tested

PR #6470, opened 2025-01-28
fix(NcActionRadio): follow-up fixes, promote migration to v-model

Antreesy:
### ☑️ Resolves

- Follow-up to #6264
  - Revert Boolean support and default value for `modelValue`:
    - ⚠️ This is not suppose to work as expected
    - ⚠️ Component emitted only `true` before, now it's value of prop `value`
    - `@change` emit works as before
  - Add Vue warning to push developers to migrate to supported types
  - Drop second emit and ref `input`
  - Typo fixes

### 🏁 Checklist

- [ ] 3️⃣ ~Backport to `next` requested with a Vue 3 upgrade~

### DIFF
--- a/src/components/NcActionRadio/NcActionRadio.vue
+++ b/src/components/NcActionRadio/NcActionRadio.vue
@@ -30,10 +30,10 @@ So that only one of each name set can be selected at the same time.
 		data() {
 			return {
 				radioOptions: [
-					{ value: 'first', label: 'First choise', disabled: false },
-					{ value: 'second', label: 'Second choise', disabled: false },
-					{ value: 'third', label: 'Third choise', disabled: false },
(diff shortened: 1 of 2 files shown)


src/components/NcActionRadio/NcActionRadio.vue
```diff
@@ -157,7 +158,11 @@ export default {
 		'change',
 	],
 
-	setup() {
+	setup(props) {
+		if (typeof props.modelValue === 'boolean') {
+			Vue.util.warn('[NcActionRadio] Boolean type of `modelValue` is deprecated and will be removed in next versions')
```
susnux, line 163: This will be only shown if set initially.
Meaning if the prop is mutated to a boolean it is not shown again

susnux (approved):
nevertheless good

PR #6033, opened 2024-08-30
refactor: Migrate `NcButton` from render function to template

susnux:
### ☑️ Resolves

1. Make it a template component instead of a render function
2. Add a `text` prop to allow passing text content as prop
3. Add `noIconAriaHidden` to allow making the icon appear on the accessible tree

### 🏁 Checklist

- [x] ⛑️ Tests are included or are not applicable
- [x] 📘 Component documentation has been extended, updated or is not applicable
- [x] 3️⃣ Backport to `next` requested with a Vue 3 upgrade

PR #6472, opened 2025-01-29
feat: Add consistent `variant` prop for design variant of buttons / chips

susnux:
* For https://github.com/nextcloud-libraries/nextcloud-vue/issues/6384#issuecomment-2601643307

### ☑️ Resolves

Currently we need to wrap the native `type` prop of a button into `nativeType` of `NcButton`.
This is confusing and we should never use names for props that are already attributes of the native HTML element.

The solution is to rename `type` to `variant`.
To not make it breaking following is applied:
1. `variant` prop is added replacing `type`
2. `nativeType` is deprecated in favor of `type`
3. `type` allows both, the native button type and the color variant
  - If a color variant is passed then the legacy behavior is applied
  - Otherwise it behaves like `nativeType`

The same is also applied for `NcDialogButton` and `NcActions`.
`NcChip` also referred to the `NcButton` for the `type` so this is also migrated the same way to make the wording `variant` consistent.

In a last step all components are refactored to use the new props instead.

### 🖼️ Screenshots

Nothing visually changed - if it does that's a regression :fearful: 

### 🏁 Checklist


src/components/NcActions/NcActions.vue
```diff
@@ -1158,6 +1160,22 @@ export default {
...
+		 *
+		 * Accepted values: primary, secondary, tertiary, tertiary-no-background, tertiary-on-primary, error, warning, success.
+		 * If left empty, the default button style will be applied.
+		 *
+		 * @since 8.23.0
+		 */
+		variant: {
+			type: String,
+			validator(value) {
+				return ['primary', 'secondary', 'tertiary', 'tertiary-no-background', 'tertiary-on-primary', 'error', 'warning', 'success'].indexOf(value) !== -1
```
Antreesy, line 1175: I'd prefer using `.includes(value)` here.
Can we also keep available variants in one place (as const, or util like `ValidateSlot`) to not duplicate it everywhere? In case it's not a limited list of only primary, secondary, tertiary

src/components/NcDialogButton/NcDialogButton.vue
```diff
@@ -91,6 +95,19 @@ const props = defineProps({
 		type: Boolean,
 		default: false,
 	},
+
+	/**
+	 * The button variant, see NcButton.
+	 *
+	 * @type {'primary'|'secondary'|'error'|'warning'|'success'}
```
Antreesy, line 102: 'tertiary' is skipped here and above

src/components/NcActions/NcActions.vue
susnux: > I'd prefer using .includes(value) here.

yes me too :) 

>  Can we also keep available variants in one place (as const, or util like ValidateSlot) to not duplicate it everywhere? In case it's not a limited list of only primary, secondary, tertiary

I will do so on v9 with #6033 as then we can share the types directly.

src/components/NcDialogButton/NcDialogButton.vue
susnux: done

st3iny (approved):

src/components/NcButton/NcButton.vue
```diff
@@ -572,6 +581,22 @@ export default {
...
 			default: null,
 		},
+
+		/**
+		 * Specifies the button variant.
+		 *
+		 * Accepted values: primary, secondary, tertiary, tertiary-no-background, tertiary-on-primary, error, warning, success.
+		 *
+		 * @default 'secondary'
+		 * @since 8.23.0
```
skjnldsv, line 591: I was wondering why that wasn't working :see_no_evil: 
This is in 8.24.0 :grin:

skjnldsv: @susnux :warning:

PR #6474, opened 2025-01-29
fix(NcRich*): style fixes and RTL support

Antreesy:
### ☑️ Resolves

- drop unused prop `markdownCssClasses` and styles for it
 - we did not support custom classes since component was merged, so should not be breaking
- apply logical rules to Rich components
- decrease size of 'tab' symbol (from 8 to 4 spaces)
- increase padding for list items (from 15px to 4ch)
  - closer to native ~40px from user agent

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image
image | image
image | image

### 🏁 Checklist


src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -1117,7 +1118,8 @@ export default {
 	overflow: auto;
 	// Hide container root element while initializing
 	position: absolute;
-	left: -10000px;
+	/* stylelint-disable csstools/use-logical */ /* upstream logic */
+	left: -100vw;
```
susnux, line 1122: why is this needed? Cant you just use `inset-inline-start`?

susnux (approved):
Looks good to me

src/components/NcRichContenteditable/NcRichContenteditable.vue
Antreesy: `.tribute-container` applies computed styles to position itself with `left`.
With RTL it's going to be `left: <calculated>px; right: -100vw;`, so it will be out of screen

Issue #5991, opened 2024-08-20
NcActionInput: Input looses focus when moving mouse


dartcafe:
The NcActionInput shows a confusing and annoing behavior. 

Hovering the mouse over the component gives it input focus.
When moving the mouse out of the way (even after clicking into the field) it again looses the focus. 

This means, it is necessary to keep the mouse inside the component borders to be able to edit the content.

Example is from a public poll with polls 7.2.0 and the current release of nextcloud-vue@8.16.0

https://github.com/user-attachments/assets/c212550c-29d9-403f-b1bb-a5ac3df37d18

ShGKme:
I think, we should remove focusing inputs on mouse move as discussed before. Will have a look at it this or next week

jaylinski:
This behavior (auto-focus on hover) is especially annoying when using a datepicker.

@ShGKme I'm very much in favor of removing this feature.

raimund-schluessler:
Will be fixed by #6475.

PR #6475, opened 2025-01-30
fix(NcActions): Remove on `mousemove` listener for auto focus elements

### LINKED ISSUE #6386: NcDateTimePickerNative: Dialog Appearance Issue in Safari
* Original report: https://github.com/nextcloud/server/issues/50065

### Bug description

In Safari, the DatePicker dialog disappears when the mouse pointer moves outside its bounds. However, this issue does not occur in other browsers such as Firefox or Chrome

https://github.com/user-attachments/assets/ba4d89f4-4a1f-4b9f-ac3f-569ae882346d

### Expected behavior

DatePicker dialog should not disappear and must be selectable

susnux:
* Resolves #6386 

### ☑️ Resolves

1. This causes issues with the native datetime picker on Safari as the mouse move is also considered outside when hovering a picker popover that overlays other actions.
2. In general this is very unexpected behavior, e.g. click on a text input, type something, move the mouse and keep typing. You expect to write to the same input but the focus is changed to a different while typing.

### 🏁 Checklist

### DIFF
--- a/src/assets/action.scss
+++ b/src/assets/action.scss
@@ -5,10 +5,14 @@
 
 @mixin action-active {
 	li.action {
+		&:hover,
 		&.active {
-			background-color: var(--color-background-hover);
 			border-radius: 6px;
 			padding: 0;
+		}
+
+		&:hover {
+			background-color: var(--color-background-hover);
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


susnux:
> This completely breaks mouse focus on elements.

Fixed that

raimund-schluessler (approved):
Great! Thanks a lot!

Supersedes #5014 and also closes #5991.

PR #6506, opened 2025-02-11
revert: import remark-gfm library in NcRichText as module, not async

Antreesy:
### ☑️ Resolves

- This reverts commit 49ce673235268a63e765859caeabec6800dba136. (#6259)

After some testing, I'm not happy with the results. async load of ~40kB comes with side effect of resizing of components, which doesn't look good in long lists (Talk)

### 🖼️ Screenshots

🏚️ Before

https://github.com/user-attachments/assets/13d6f094-169e-4be4-b34b-f8adafa82ad9

🏡 After

https://github.com/user-attachments/assets/8816a340-d674-41f4-a513-808cf86fc2c3

### 🏁 Checklist

### DIFF
--- a/src/components/NcRichText/NcRichText.vue
+++ b/src/components/NcRichText/NcRichText.vue
@@ -305,7 +305,6 @@ See [NcRichContenteditable](#/Components/NcRichContenteditable) documentation fo
 import { ref } from 'vue'
 import NcReferenceList from './NcReferenceList.vue'
 import NcCheckboxRadioSwitch from '../NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue'
-import NcLoadingIcon from '../NcLoadingIcon/NcLoadingIcon.vue'
 import { getRoute, remarkAutolink } from './autolink.js'
 import { remarkPlaceholder, prepareTextNode } from './placeholder.js'
 import { remarkUnescape } from './remarkUnescape.js'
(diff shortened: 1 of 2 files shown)


susnux:
Would it not be possible to just keep showing the skeletons until loading is finished?
That would fix the visual issue and also keep bundle size smaller (also performance of parsing).

Antreesy:
> Would it not be possible to just keep showing the skeletons until loading is finished?

If you're asking whether it's possible to do on app side, we then need to trigger async import while it's loading (e.g. place NcRichText with markdown somewhere hidden to trigger it on initial load), and show skeleton until internal component state is updated? Not sure that's a good idea..

susnux:
It was more like:
emit some `loaded` event when rendering is finished so the app can replace their loading (talk skeletons) with the rendered component.

> We can separate it into two components:

Do we use the non-markdown versions somewhere? Because if not then it makes not much sense.
For me the problem is the amount and size of dependencies that are synchronously imported by this component.
So for personal projects I do not use it but custom components with markdown rendering saving quite some size.

susnux (approved):
anyways makes sense here

Antreesy:
> Do we use the non-markdown versions somewhere?

Found some usages in Talk, Activity, integration apps, could be somewhere else. So some may benefit from it, especially if discussed `strip-markdown` option will be added

PR #6512, opened 2025-02-12
fix(NcPopover): blurry content on scaled page in Chromium

ShGKme:
### ☑️ Resolves

- Fix: https://github.com/nextcloud-libraries/nextcloud-vue/issues/6040

1. `floating-vue` uses `translate3d` (with `translateZ`) to trigger GPU acceleration
   - The same does perspective with rotation, for example
   - It also triggers a new composition layer
   - It also changes how Chromium calculates coordinates
2. Then we apply `filter: drop-shadow`
3. `filter: drop-shadow` requires previous layers of content on rendering
   - The same does anything using GPU acceleration with content behind, like `backdrop-filter` or `mix-blend-mode`
4. Somehow it makes Chromium first render and rasterize the popover before scaling
   - So Chromium scales raster image on render instead of scaling fonts/sizes and then render
   - Or it uses the wrong rasterization method on fraction scaling
   - Anyway, it results in blurry images

Replacing `filter: drop-shadow` with a simple `box-shadow` fixes the issue.

- Note: there still is a blur on highDPI screens with scaling
- It requires a fix in upstream: https://github.com/Akryum/floating-vue/issues/932

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image
image | image
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/NcPopover/NcPopover.vue
+++ b/src/components/NcPopover/NcPopover.vue
@@ -434,6 +434,9 @@ export default {
 }
 
 $arrow-width: 10px;
+// Move the arrow just slightly inside the popover
+// To prevent a visual gap on page scaling
+$arrow-position: $arrow-width - 1px;
 
 .v-popper--theme-dropdown {
 	&.v-popper__popper {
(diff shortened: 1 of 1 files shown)


src/components/NcPopover/NcPopover.vue
```diff
@@ -442,7 +445,18 @@ $arrow-width: 10px;
 		left: 0;
 		display: block !important;
 
-		filter: drop-shadow(0 1px 10px var(--color-box-shadow));
+		.v-popper__wrapper {
+			/*
+			 * In theory, "filter: drop-shadow" would look better here with arrow shadow.
+			 * In fact, in results in a blurry popover in Chromium on scaling.
```
susnux, line 451: Typical Blink engine issue, their hardware acceleration is good - sometimes - and really really bad in most cases. Gecko is buggy but at least consistent as they wrote the renderer themself.

susnux (approved):

szaimen (approved):
LGTM from the screenshots

PR #6523, opened 2025-02-13
fix(RTL): initialize language direction check once in utils

Antreesy:
### ☑️ Resolves

- Follow-up to #6507
- I thought it's better to keep const name shorter, but feel free to suggest another way

### 🖼️ Screenshots

image

### 🏁 Checklist

### DIFF
--- a/src/components/NcActionButton/NcActionButton.vue
+++ b/src/components/NcActionButton/NcActionButton.vue
@@ -355,8 +355,8 @@ export default {
 			<span v-else class="action-button__text">{{ text }}</span>
 
 			<!-- right(in LTR) or left(in RTL) arrow icon when there is a sub-menu -->
-			<ChevronRightIcon v-if="isMenu && !isRTL" :size="20" class="action-button__menu-icon" />
-			<ChevronLeftIcon v-else-if="isMenu && isRTL" :size="20" class="action-button__menu-icon" />
+			<ChevronRightIcon v-if="isMenu && !isRtl" :size="20" class="action-button__menu-icon" />
+			<ChevronLeftIcon v-else-if="isMenu && isRtl" :size="20" class="action-button__menu-icon" />
 			<CheckIcon v-else-if="isChecked === true" :size="20" class="action-button__pressed-icon" />
 			<span v-else-if="isChecked === false" class="action-button__pressed-icon material-design-icon" />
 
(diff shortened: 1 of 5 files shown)


src/components/NcActionButton/NcActionButton.vue
```diff
@@ -371,7 +371,9 @@ import CheckIcon from 'vue-material-design-icons/Check.vue'
 import ChevronRightIcon from 'vue-material-design-icons/ChevronRight.vue'
 import ChevronLeftIcon from 'vue-material-design-icons/ChevronLeft.vue'
 import ActionTextMixin from '../../mixins/actionText.js'
-import { isRTL } from '@nextcloud/l10n'
+import { isRTL as isRTLFunction } from '@nextcloud/l10n'
+
+const isRTL = isRTLFunction()
```
susnux, line 376: just one line we save here but how about a composable `useIsRtl`?
This could cache it across all components and reduces duplicate logic (even if only one line)

susnux: Depends on the definition it provides a state (isRTL) and you can inject it in setup context to the component.

It could also be just a module level variable like:
```ts
// utils/rtl.ts
export const isRtl = isRTL()
```

and then you can just use it like
```ts
import { isRtl } from '../../utils/rtl.ts'

// ...
setup() {
    return {
        // ...
        isRtl,
    }
}
// ...
```

ShGKme: > ```ts
> import { isRtl } from '../../utils/rtl.ts'
> 
> // ...
> setup() {
>     return {
>         // ...
>         isRtl,
>     }
> }
> // ...
> ```

Agree with this 👍

Wouldn't agree witha composable

```js
const isRtl = l10nIsRtl()

/** Composable to have isRtl */
export function useIsRtl() {
  return isRtl
}

// ----

import { useIsRtl } from '../../composables/rtl.ts'

// ...
setup() {
    return {
        // ...
        isRtl: useIsRtl(),
    }
}
// ...
```

Antreesy: Sounds good, let me rearrange files

susnux: > Agree with this 👍

:+1:

ShGKme (approved):

susnux (approved):

PR #6543, opened 2025-02-18
feat(functions): Provide `formatDateTime` function

susnux:
### ☑️ Resolves

Split the `useFormatDateTime` composable into a general function without reactivity and the composable that uses it.
This allows to also use this logic in other places.

### 🖼️ Screenshots

No visual changes.

### 🏁 Checklist

### DIFF
--- a/src/composables/useFormatDateTime/index.ts
+++ b/src/composables/useFormatDateTime/index.ts
@@ -5,7 +5,7 @@
 
 import { getCanonicalLocale, getLanguage } from '@nextcloud/l10n'
 import { computed, onUnmounted, ref, onMounted, watch, unref, type Ref } from 'vue'
-import { t } from '../l10n.js'
+import { t } from '../../l10n.js'
 
 const FEW_SECONDS_AGO = {
 	long: t('a few seconds ago'),
--- a/src/composables/useFormatDateTime/index.ts
+++ b/src/composables/useFormatDateTime/index.ts
@@ -3,29 +3,31 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-import { getCanonicalLocale, getLanguage } from '@nextcloud/l10n'
-import { computed, onUnmounted, ref, onMounted, watch, unref, type Ref } from 'vue'
-import { t } from '../../l10n.js'
+import type { FormatDateOptions } from '../../functions/formatDateTime/index.ts'
+import type { Ref } from 'vue'
 
-const FEW_SECONDS_AGO = {
-	long: t('a few seconds ago'),
-	short: t('seconds ago'), // FOR TRANSLATORS: Shorter version of 'a few seconds ago'
-	narrow: t('sec. ago'), // FOR TRANSLATORS: If possible in your language an even shorter version of 'a few seconds ago'
-}
+import { getCanonicalLocale } from '@nextcloud/l10n'
+import { computed, onUnmounted, ref, onMounted, watch, unref } from 'vue'
+import { formatDateTime } from '../../functions/formatDateTime/index.ts'
 
-interface FormatDateOptions {
+interface FormatDateComposableOptions {
 	/**
 	 * The format used for displaying, or if relative time is used the format used for the title
+	 * @default  '{ timeStyle: "medium", dateStyle: "short" }'
 	 */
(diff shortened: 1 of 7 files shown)


artonge (approved):

ShGKme:
Interface seems confusing to me.

- Function name says `formatDateTime`, so I'd expect it to format timestamp or duration like **01 Jan 2025** or **12 minutes**
- Documentation says `Format as relative time`, like `Intl.RelativeTimeFormat`
- Composable returns
  - `formattedTime` which sounds like `DateTimeFormat` but actually `RelativeTimeFormat`
  - `formattedFullTime` which sounds like it is a full form (and there is a short form?) but actually "format (date) time" using `DateTimeFormat`

What about:
1. Keeping naming close to native naming and popular libraries:
   - `formatTime`/`formatDateTime` to format as date or duration like `DateTimeFormat.format()`
   - `formatRelativeTime` to format as relative time like `RelativeTime.format()`
2. Doing the same in the composable result
   - `formattedTime` -> `formattedRelativeTime`
   - `formattedFullTime` -> `formattedTime`/`formattedDateTime`
3. Separating a single universal function/composable into 2 with different roles:
   - To keep the interface simpler (less params)
   - To avoid unneeded actions
   - To also have a function for formatted time (that sets the locale)
   - It seems a very rare case when you need both formatted for one date
   - In this case composables can directly return the result without wrapping into an object and be simpler to use

```js
/**
 * Current
 */

// Complex renaming
const { formattedFullTime: formattedBirthday } = useFormatDateTime(birthdayDate)
// Confusing "fulltime" for a duration
const { formattedFullTime: formattedEventDuraction } = useFormatDateTime(eventDuration)
// Not obious that formatted time is relative
const { formattedTime: clearAtLabel } = useFormatDateTime(userStatus.clearAt)

/**
 * Proposal
 */
const formattedBirthday = useFormatDateTime(birthdayDate)
const formattedEventDuraction = useFormatDateTime(eventDuration)
const clearAtLabel = useRelativeTime(userStatus.clearAt)
```

ShGKme:
> But for the composable I am not sure as one of the use cases is to have relative vs "plain" time as a configurable options, so with your way you need to include two variables and switch between in a computed?

Do you mean, to have a switch from relative to absolute/duration value?

Do you have an example for a use case? It seems to me, relative and general time are always different parts of UI

src/composables/useFormatDateTime/index.ts
```diff
@@ -0,0 +1,114 @@
...
+			intervalId = window.setInterval(() => { currentTime.value = Date.now() }, updateInterval.value)
+		}
+	})
+
+	// Start the interval for setting the current time if relative time is enabled
+	onMounted(() => {
+		if (wrappedOptions.value.relativeTime !== false) {
+			intervalId = window.setInterval(() => { currentTime.value = Date.now() }, updateInterval.value)
+		}
+	})
```
ShGKme, line 102: It doesn't require component to be in the DOM (or even to be a component at all)
```suggestion
	// Start the interval for setting the current time if relative time is enabled
	if (wrappedOptions.value.relativeTime !== false) {
		intervalId = window.setInterval(() => { currentTime.value = Date.now() }, updateInterval.value)
	}
```

src/composables/useFormatDateTime/index.ts
```diff
@@ -0,0 +1,114 @@
...
+
+	// ensure interval is cleared
+	onUnmounted(() => {
+		window.clearInterval(intervalId)
+	})
```
ShGKme, line 107:
```suggestion
	// ensure interval is cleared
	onScopeDispose(() => {
		if (intervalId !== undefined) {
			window.clearInterval(intervalId)
		}
	})
```

susnux:
> Do you have an example for a use case? It seems to me, relative and general time are always different parts of UI

Logreader if the user selects relative time in the settings or "normal time".

src/composables/useFormatDateTime/index.ts
ShGKme: Oh, it is not available in Vue 2

susnux:
https://github.com/nextcloud-libraries/nextcloud-l10n/pull/921

PR #6608, opened 2025-03-08
Fix useModelMigration in async components

### LINKED ISSUE #6607: useModelMigration broken in async components
All components that use `useModelMigration` are broken when used inside an asynchronously loaded component. I've attached a patch you can apply to https://github.com/nextcloud/app_template/commit/5b507c644128661c863d81cf40216b00079d479d to reproduce.

<details><summary>minimal patch</summary>

```diff
diff --git a/src/App.vue b/src/App.vue
index e579eba..860ff4b 100644
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,18 +1,22 @@
 <template>
 	<NcAppContent>
 		<div id="app_template">
-			<h1>Hello world!</h1>
+			<Repro />
 		</div>
 	</NcAppContent>
 </template>
 
 <script>
 import NcAppContent from '@nextcloud/vue/dist/Components/NcAppContent.js'
+import { defineAsyncComponent } from 'vue'
+// import Repro from './Repro.vue'
 
 export default {
 	name: 'App',
 	components: {
 		NcAppContent,
+		// Repro,
+		Repro: defineAsyncComponent(() => import('./Repro.vue')),
 	},
 }
 </script>
diff --git a/src/Repro.vue b/src/Repro.vue
new file mode 100644
index 0000000..5b5f8bb
--- /dev/null
+++ b/src/Repro.vue
@@ -0,0 +1,36 @@
+<template>
+	<div>
+		<NcActions>
+			<NcActionRadio v-for="option in radioOptions"
+				:key="option.value"
+				:value="option.value"
+				:disabled="option.disabled"
+				name="uniqueId"
+				v-model="radioValue">
+				{{ option.label }}
+			</NcActionRadio>
+		</NcActions>
+		<span>Selected value: {{ radioValue }}</span>
+	</div>
+</template>
+
+<script>
+import NcActions from '@nextcloud/vue/components/NcActions'
+import NcActionRadio from '@nextcloud/vue/components/NcActionRadio'
+
+export default {
+	components: {
+		NcActions,
+		NcActionRadio,
+	},
+	data: () => ({
+		radioOptions: [
+			{ value: 'first', label: 'First choice', disabled: false },
+			{ value: 'second', label: 'Second choice', disabled: false },
+			{ value: 'third', label: 'Third choice', disabled: false },
+			{ value: 'fourth', label: 'Fourth choice (disabled)', disabled: true },
+		],
+		radioValue: 'first',
+	}),
+}
+</script>
```

</details>

baltitenger:
It was using getCurrentInstance() which, as I understand it, isn't part of vue's public API anymore and isn't guaranteed to work in every situation.

Makes call site a bit uglier, but I don't think there's a better way to do it.

### ☑️ Resolves

- Fix #6607

### 🏁 Checklist

### DIFF
--- a/src/components/NcActionCheckbox/NcActionCheckbox.vue
+++ b/src/components/NcActionCheckbox/NcActionCheckbox.vue
@@ -139,8 +139,8 @@ export default {
 		'update:model-value',
 	],
 
-	setup() {
-		const model = useModelMigration('checked', 'update:checked')
+	setup(props, {emit}) {
+		const model = useModelMigration(props, emit, 'checked', 'update:checked')
 		return {
 			model,
 		}
--- a/src/components/NcActionInput/NcActionInput.vue
+++ b/src/components/NcActionInput/NcActionInput.vue
@@ -416,8 +416,8 @@ export default {
 		'update:model-value',
 	],
 
-	setup() {
-		const model = useModelMigration('value', 'update:value')
+	setup(props, {emit}) {
+		const model = useModelMigration(props, emit, 'value', 'update:value')
 		return {
 			model,
 		}
--- a/src/components/NcActionRadio/NcActionRadio.vue
+++ b/src/components/NcActionRadio/NcActionRadio.vue
@@ -158,12 +158,12 @@ export default {
 		'change',
 	],
 
-	setup(props) {
+	setup(props, {emit}) {
 		if (typeof props.modelValue === 'boolean') {
 			Vue.util.warn('[NcActionRadio] Boolean type of `modelValue` is deprecated and will be removed in next versions')
 		}
 
-		const model = useModelMigration('checked', 'update:checked')
+		const model = useModelMigration(props, emit, 'checked', 'update:checked')
 		return {
 			model,
 		}
--- a/src/components/NcActionTextEditable/NcActionTextEditable.vue
+++ b/src/components/NcActionTextEditable/NcActionTextEditable.vue
@@ -166,8 +166,8 @@ export default {
 		'submit',
 	],
 
-	setup() {
-		const model = useModelMigration('value', 'update:value')
+	setup(props, {emit}) {
+		const model = useModelMigration(props, emit, 'value', 'update:value')
 		return {
 			model,
 			isRtl,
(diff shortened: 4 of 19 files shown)


ShGKme:
> It was using getCurrentInstance() which, as I understand it, isn't part of vue's public API anymore 

It is a part of exposed Advanced API. It was called internal in the past indeed. Because it returns `ComponentInternalInstance` which is internal API and can be changed. However, Vue 2.7 is EOL so the result will never change. Also, Vue 2 didn't have a separation to `ComponentInternalInstance` and `ComponentPublicInstance` - so the result is a public instance.

> and isn't guaranteed to work in every situation.

Could you eliminate on a case when it isn't supposed to work?

baltitenger:
> Could you clarify on a case when it isn't supposed to work?

I'm not actually sure, I think I read somewhere that it's only guaranteed to work from lifecycle events but setup is one, so I don't know why it wouldn't work in this situation. But I did observe it returning null so clearly it can happen (see the mentioned issue).

Another thing, I just noticed that I forgot to update the jsdoc comment for useModelMigration, but I'm not sure how to type these two new parameters. (plus `typescript-language-server` seems to ignore typing information in jsdoc, I guess in typescript you're supposed to have it inline)

ShGKme:
> I'm not actually sure, I think I read somewhere that it's only guaranteed to work from lifecycle events but setup is one, so I don't know why it wouldn't work in this situation

`getCurrentInstance` only supposed to work when there is the current instance - in the **setup context**. Same as all the other Composition API function and composables.

And the setup function has the setup context. Otherwise not only this function wouldn't work here, but any composable and even reactive effect wouldn't be cleared after component is unmounted and etc.

> But I did observe it returning null so clearly it can happen (see the mentioned issue).

I'd guess that you test not with installed `@nextcloud/vue`, but with `npm link` to local `@nextcloud/vue`. In this case there are 2 copies of Vue in different `node_modules`. And Vue doesn't support it in general.

You need to either use `npm pack` instead of `npm link` to have a complete reproduction of installed package, or resolve `vue` as alias to app's `node_modules`.

ShGKme:
Closing as the justification for the change is not correct, `getCurrentInstance` must be called in `setup`.

PR #6636, opened 2025-03-21
fix(NcHeaderMenu): allow to disable focus-trap

Antreesy:
### ☑️ Resolves

- Allows to fix focus-trap issue, when there is no focusable content
  - introduces new prop `focusTrap: Boolean`
- Aligned with NcPopover logic
- Needed for notifications app

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist


src/components/NcHeaderMenu/NcHeaderMenu.vue
```diff
@@ -186,7 +217,7 @@ export default {
 
 	data() {
 		return {
-			focusTrap: null,
+			$focusTrap: null,
```
susnux, line 220: Not a fan of local variables prefixed with `$`, always feels like this is a variable injected by the app (like `$router`).

susnux (approved):
Seems to work, but:

```diff
- fix(NcHeaderMenu): allow to disable focus-trap #6636 
+ feat(NcHeaderMenu): allow to disable focus-trap #6636 
```

As it adds new props

src/components/NcHeaderMenu/NcHeaderMenu.vue
Antreesy: => `focusTrapInstance`?

src/components/NcHeaderMenu/NcHeaderMenu.vue
```diff
@@ -173,6 +196,14 @@ export default {
...
 			default: null,
 		},
+
+		/**
+		 * Enable content focus trap
+		 */
+		focusTrap: {
+			type: Boolean,
+			default: true,
+		},
```
ShGKme, line 206: We just made a breaking change in v9 to get rid of boolean props with default true value, as they conflict with the HTML template style.

Let's call it `disableFocusTrap` or `noFocusTrap`.

ShGKme: Or we can check if there are any focusable elements and enable/disable it automatically.

ShGKme: Ref: https://github.com/nextcloud-libraries/nextcloud-vue/pull/6452

Antreesy:
Let's start over, I should have used debugger in the first place

ShGKme (approved):

susnux (approved):

PR #6648, opened 2025-03-25
[next] feat: use `nextcloud-theme-dark` var for dark theme computation

skjnldsv:
Require https://github.com/nextcloud/server/pull/51702

~~On next only so we're compatible with 32 and above? :crossed_fingers:~~

### DIFF
--- a/src/functions/isDarkTheme/index.ts
+++ b/src/functions/isDarkTheme/index.ts
@@ -20,10 +20,19 @@ export function checkIfDarkTheme(el: HTMLElement = document.body): boolean {
 	// 2. second priority is the data-themes attribute on the body, aka the user setting
 	// 3. third priority is the system preference
 	// 4. lastly is the default light theme
-	const darkModeSystemPreference = window?.matchMedia?.('(prefers-color-scheme: dark)')?.matches
-	const darkModeAccountSetting = document.body.getAttribute('data-themes')?.includes('dark')
-	const darkModeElementOverride = el.closest('[data-theme-dark]') !== null
-	return darkModeElementOverride || darkModeAccountSetting || darkModeSystemPreference || false
+	if (el.closest('[data-theme-dark]') !== null) {
+		return true
+	}
+
+	if (document.body.getAttribute('data-themes')?.includes('dark')) {
+		return true
+	}
+
+	if (window.matchMedia('(prefers-color-scheme: dark)')?.matches) {
+		return true
+	}
+
+	return false
 }
 
 /**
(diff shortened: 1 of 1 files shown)


ShGKme:
> On next only so we're compatible with 32 and above? 🤞

Some apps are using it already as this is the only current Vue supporting version

skjnldsv:
> Some apps are using it already as this is the only current Vue supporting version

Ah, so I need to stay compatible with the old ways :+1: 
i'll adjust the PR

skjnldsv:
Done @ShGKme

src/functions/isDarkTheme/index.ts
```diff
@@ -9,18 +9,21 @@
...
-	return false
+	// Else we check the old ways
+	// 1. first priority is the data-theme-dark attribute on a parent element
+	// 2. second priority is the data-themes attribute on the body, aka the user setting
+	// 3. third priority is the system preference
+	// 4. lastly is the default light theme
+	const darkModeSystemPreference = window?.matchMedia?.('(prefers-color-scheme: dark)')?.matches
+	const darkModeAccountSetting = document.body.getAttribute('data-themes')?.includes('dark')
+	const darkModeElementOverride = el.closest('[data-theme-dark]') !== null
+	return darkModeElementOverride || darkModeAccountSetting || darkModeSystemPreference || false
```
ShGKme, line 26: To do lazy checks and not perform x3 DOM API calls

```suggestion
	if (el.closest('[data-theme-dark]') !== null) {
		return true
	}

	if (document.body.getAttribute('data-themes')?.includes('dark')) {
		return true
	}

	if (window.matchMedia('(prefers-color-scheme: dark)')?.matches) {
		return true
	}

	return false
```

ShGKme: Or
```suggestion
	const darkModeSystemPreference = () => window.matchMedia('(prefers-color-scheme: dark)')?.matches
	const darkModeAccountSetting = () => document.body.getAttribute('data-themes')?.includes('dark')
	const darkModeElementOverride = () => el.closest('[data-theme-dark]') !== null
	return darkModeElementOverride() || darkModeAccountSetting() || darkModeSystemPreference() || false
```

ShGKme: > 	// 1. first priority is the data-theme-dark attribute on a parent element
> 	// 2. second priority is the data-themes attribute on the body, aka the user setting

Could you provide with the example where we need 2 and 1 was not enough?

skjnldsv: @ShGKme future proof. The Theming engine have been initially designed so other themes could be added or even provided on the appstore.
The `[data-theme-dark]` wasn't even properly discussed :see_no_evil:

PR #6652, opened 2025-03-27
fix(NcEmojiPicker): return focus steals actual focus

ShGKme:
### ☑️ Resolves

Like in a `focus-trap`, we should only return focus on close if the closing action didn't add the focus already, for example, when the emoji picker was closed by clicking on an input field.

### 🏁 Checklist


Antreesy (approved):
Works

DorraJaouad (approved):
Saw it working

PR #6657, opened 2025-03-27
refactor: drop `md5` dependency for `usernameToColor` and migrate to TS

susnux:
### ☑️ Resolves

- Migrate to Typescript
- Simplify
- Remove useless dependency on md5

### 🏁 Checklist


ShGKme (approved):

skjnldsv (approved):

max-nextcloud:
I think the color returned by this function used to match the one computed by the server as a background color for the avatar: https://github.com/nextcloud/server/blob/776a689e09aea18dd455b885bdb575849b525d3e/lib/private/Avatar/Avatar.php#L256

I think it would be nice to maintain that property - so you can compute the color of he default avatar on the client side. I'd like to use this in text for the author indicator so it matches the user avatar if the user did not change it.

So what's the best way forward here? Reimplementing the simple hash in php in the server function?

**update:** just noticed we don't need it in text as colors are computed on the server anyway.

susnux:
Yes basically this is "never" used as there is always the initials avatar returned by server. Its only about is-no-user of the avatar component

PR #6666, opened 2025-03-28
feat(NcAppSidebar): add `info` slot

Antreesy:
### ☑️ Resolves

- Allows to put custom content with `info` slot
- Expose `var(--app-sidebar-close-button-offset)` for custom content positioning (respecting the space taken by close button)
- Introduce `NcAppSidebarHeader` for a11y navigation

See https://github.com/nextcloud-libraries/nextcloud-vue/pull/6666/files?w=1 for non-whitespace changes
See https://github.com/nextcloud-libraries/nextcloud-vue/tree/feat/noid/appsidebar-header-empty-slot-vue2 for Vue2 branch to test with apps

### 🖼️ Screenshots

Quick prototype for Talk:

Clean | WIth explanation
-- | --
image | image

https://github.com/user-attachments/assets/8815ee43-7b35-4453-a1ef-852bb7cac181

### 🏁 Checklist

- [x] 3️⃣ Backport to `stable8` requested

### DIFF
--- a/src/components/NcAppSidebar/NcAppSidebar.vue
+++ b/src/components/NcAppSidebar/NcAppSidebar.vue
@@ -550,7 +550,8 @@ export default {
 				As a simple solution - render it in the content to keep correct position.
 			-->
 			<Teleport v-if="ncContentSelector && !open && !noToggle" :to="ncContentSelector">
-				<NcButton :aria-label="t('Open sidebar')"
+				<NcButton ref="toggle"
+					:aria-label="t('Open sidebar')"
 					class="app-sidebar__toggle"
 					:class="toggleClasses"
 					variant="tertiary"
@@ -618,17 +619,13 @@ export default {
 						<div class="app-sidebar-header__name-container">
 							<div class="app-sidebar-header__mainname-container">
 								<!-- main name -->
-								<h2 v-show="!nameEditable"
-									:id="`app-sidebar-vue-${uid}__header`"
-									ref="header"
-									v-linkify="{text: name, linkify: linkifyName}"
-									:aria-label="title"
-									:title="title"
+								<NcAppSidebarHeader v-show="!nameEditable"
 									class="app-sidebar-header__mainname"
+									:name
+									:linkify="linkifyName"
+									:title
 									:tabindex="nameEditable ? 0 : -1"
-									@click.self="editName">
-									{{ name }}
-								</h2>
+									@click.self="editName" />
 								<template v-if="nameEditable">
 									<form v-click-outside="() => onSubmitName()"
 										class="app-sidebar-header__mainname-form"
@@ -669,6 +666,11 @@ export default {
 						</div>
 					</div>
 				</div>
+				<!-- a11y fallback for empty content -->
+				<NcAppSidebarHeader v-if="empty"
+					class="app-sidebar-header__mainname--hidden"
+					:name
+					tabindex="-1" />
 
 				<NcButton ref="closeButton"
 					:aria-label="closeTranslated"
@@ -706,13 +708,14 @@ export default {
 <script>
 import NcAppSidebarTabs from './NcAppSidebarTabs.vue'
 import NcActions from '../NcActions/index.js'
+import NcAppSidebarHeader from '../NcAppSidebarHeader/index.ts'
 import NcButton from '../NcButton/index.ts'
 import NcEmptyContent from '../NcEmptyContent/index.js'
 import NcLoadingIcon from '../NcLoadingIcon/index.js'
 import Focus from '../../directives/Focus/index.js'
-import Linkify from '../../directives/Linkify/index.ts'
 import { vOnClickOutside as ClickOutside } from '@vueuse/components'
 import { createFocusTrap } from 'focus-trap'
+import { provide, ref, warn } from 'vue'
 import { useIsSmallMobile } from '../../composables/useIsMobile/index.js'
 import { createElementId } from '../../utils/createElementId.ts'
 import { getTrapStack } from '../../utils/focusTrap.ts'
@@ -730,6 +733,7 @@ export default {
 
 	components: {
 		NcActions,
+		NcAppSidebarHeader,
 		NcAppSidebarTabs,
 		NcButton,
 		NcLoadingIcon,
@@ -743,7 +747,6 @@ export default {
 
 	directives: {
 		Focus,
-		Linkify,
 		ClickOutside,
 	},
 
@@ -922,9 +925,13 @@ export default {
 	],
 
 	setup() {
+		const headerRef = ref(null)
+		provide('NcAppSidebar:header:ref', headerRef)
+
 		return {
 			uid: createElementId(),
 			isMobile: useIsSmallMobile(),
+			headerRef,
 		}
 	},
 
@@ -1135,7 +1142,16 @@ export default {
 		 * @public
 		 */
 		focus() {
-			(this.$refs.header ?? this.$refs.toggle)?.focus()
+			if (!this.open && !this.noToggle) {
+				this.$refs.toggle.$el.focus()
+				return
+			}
+
+			try {
+				this.headerRef.focus()
+			} catch {
+				warn('NcAppSidebar should have focusable header for accessibility reasons. Use NcAppSidebarHeader component.')
+			}
 		},
 
 		/**
@@ -1495,6 +1511,17 @@ $top-buttons-spacing: 6px;
 					}
 				}
 			}
+		}
+
+		// Hidden a11y fallback
+		.app-sidebar-header__mainname--hidden {
+			position: absolute;
+			top: 0;
+			inset-inline-start: 0;
+			margin: 0;
+			width: 1px;
+			height: 1px;
+			overflow: hidden;
 		}
 
 		// sidebar description slot
--- a/src/components/NcAppSidebar/NcAppSidebar.vue
+++ b/src/components/NcAppSidebar/NcAppSidebar.vue
@@ -1287,6 +1287,9 @@ $top-buttons-spacing: 6px;
 	}
 
 	.app-sidebar-header {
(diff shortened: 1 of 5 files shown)


src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -664,8 +663,13 @@ export default {
 						</div>
 					</div>
 				</div>
+				<div v-else-if="$slots['content']" class="app-sidebar-header__info">
+					<!-- @slot Alternative to the default header content: use 'empty' prop for bare AppSidebar with custom inner content and tabs -->
+					<slot name="content" />
```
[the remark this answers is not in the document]
Antreesy: Valid point. I was trying to implement it with minimal diff

susnux (approved):
technically makes sense - not sure about design.
But in general I like the idea of splitting it into a layout component and the specialized components, so :+1:

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -565,107 +565,111 @@ export default {
...
+								</p>
+							</div>
 						</div>
 					</div>
-				</div>
+				</slot>
 
-				<NcButton ref="closeButton"
+				<NcButton v-if="!noClose"
+					ref="closeButton"
```
ShGKme, line 672: When there is no button
```
TypeError: Cannot read properties of undefined (reading 'focus')
```

Antreesy: Caused by `empty` prop, when there is no `$refs.header` rendered

DorraJaouad (approved):
Makes sense

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -1155,7 +1166,7 @@ export default {
 		 * @public
 		 */
 		focus() {
-			this.$refs.header.focus()
+			this.$refs.header?.focus()
```
ShGKme, line 1169: Sidebar must be focused on open, we cannot just hide the error here... I'll check tomorrow with a screen reader.

Antreesy: what about:
```js
const element = tabbable(this.$refs.sidebar)[0]
if (element) {
	element.focus()
}
```
Then it either will fall back either to:
- focusable/tabbable elements provided by developers
- close button
- <here we can throw an warning, I guess>

`this.$refs.header` is either non-visible or has tabindex `-1`, it should not be in focus in both cases

susnux: For screen readers you could also focus non interactive elements to update the position of the "viewer"

susnux (changes requested):
Paddings seem to be off, also close button will not be placed consistent.
I think we need to polish the implementation a bit before pushing this to public API.

src/components/NcAppSidebar/NcAppSidebar.vue
Antreesy: checking main branch now:
- `ref="toggle"` was lost in 6e13c042a65b21c859810d8a423a542b35e26281
- `focus()` method has it: `(this.$refs.header ?? this.$refs.toggle?.$el)?.focus()`
- should be `this.$refs.toggle?.$el`, or NcButton should implement and expose `focus()`

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -743,10 +746,15 @@ export default {
 
 	directives: {
 		Focus,
-		Linkify,
 		ClickOutside,
 	},
 
+	provide() {
+		return {
+			'NcAppSidebar:header:id': computed(() => this.headerId),
```
ShGKme, line 754: It is a static value

```suggestion
			'NcAppSidebar:header:id': this.headerId,
```

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -1008,7 +1020,7 @@ export default {
 				document.querySelector('#header'),
 			], {
 				allowOutsideClick: true,
-				fallbackFocus: this.$refs.closeButton.$el,
+				fallbackFocus: this.$refs.closeButton?.$el ?? this.$refs.sidebar,
```
ShGKme, line 1023: It is always available 

```suggestion
				fallbackFocus: this.$refs.closeButton.$el,
```

src/components/NcAppSidebar/NcAppSidebarHeader.vue
```diff
@@ -0,0 +1,26 @@
...
+const id = inject('NcAppSidebar:header:id', 'app-sidebar-vue__header')
+</script>
+
+<template>
+	<h2 :id="id"
+		v-linkify="{ text: name, linkify: linkifyName }"
+		:aria-label="title"
+		:title="title">
+		<slot>{{ name }}</slot>
+	</h2>
```
ShGKme, line 25: (⌐■_■)

```suggestion
	<h2 :id
		v-linkify="{ text: name, linkify: linkifyName }"
		:aria-label="title"
		:title>
		<slot>{{ name }}</slot>
	</h2>
```

src/components/NcAppSidebar/index.js
```diff
@@ -4,3 +4,4 @@
  */
 
 export { default } from './NcAppSidebar.vue'
+export { default as NcAppSidebarHeader } from './NcAppSidebarHeader.vue'
```
ShGKme, line 7: As it is a public component, let's move it into a new `components/NcAppSidebarHeader` directory.

Otherwise this component needs to me imported with named import.

src/components/NcAppSidebar/NcAppSidebarHeader.vue
```diff
@@ -0,0 +1,26 @@
...
+-->
+
+<script setup lang="ts">
+import { inject } from 'vue'
+import vLinkify from '../../directives/Linkify/index.ts'
+
+defineProps<{
+	name: string,
+	title?: string,
+	linkifyName?: boolean,
```
ShGKme, line 13: For simplicity

```suggestion
	linkify?: boolean,
```

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -1135,7 +1147,16 @@ export default {
...
 		focus() {
-			(this.$refs.header ?? this.$refs.toggle)?.focus()
+			if (!this.open && !this.noToggle) {
+				this.$refs.toggle.$el.focus()
+				return
+			}
+
+			try {
+				document.getElementById(this.headerId).focus()
+			} catch (e) {
```
ShGKme, line 1157: Note: alternative for providing ID is to provide a (template) ref.

ShGKme: @Antreesy By "provide" I meant Vue provide to the NcAppSidebarHeader. Then the header can inject it and bind as a template ref.

ID works, this component is single use and has ID anyway, but ref would be slightly more vue way

Antreesy: Oh.

susnux (approved):

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -570,105 +581,104 @@ export default {
...
-									@click.prevent="toggleStarred">
-									<template #icon>
-										<NcLoadingIcon v-if="starLoading" />
-										<IconStar v-else-if="isStarred" :size="20" />
-										<IconStarOutline v-else :size="20" />
-									</template>
-								</NcButton>
-							</slot>
+				<!-- @slot Alternative to the default header content: use for bare NcAppSidebar with tabs. NcAppSidebarHeader would be required to use for accessibility reasons -->
+				<slot name="content">
```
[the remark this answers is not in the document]
ShGKme: From my understanding of the `empty` prop, sidebar with `empty` state should render no header, including custom header (new slot).

ShGKme:
Proposals:
- Rename a new slot to `#info` or something else to not mix with the entire sidebar content
- Hide new slot in `empty` state
- Add a hidden header for `empty` state
- For research - check if we need to replace `__info` block or define custom `__info` content
- Follow-up: fallback `<NcAppSidebarHeader>` name and title to `<NcAppSidebar>`'s corresponding values

src/components/NcAppSidebar/NcAppSidebar.vue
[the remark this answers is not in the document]
Antreesy: That is controversial to use both together. But I'll mention it in slot docs

src/components/NcAppSidebarHeader/NcAppSidebarHeader.vue
```diff
@@ -0,0 +1,37 @@
...
+	linkify?: boolean,
+}>()
+
+const headerRef = inject('NcAppSidebar:header:ref')
+</script>
+
+<template>
+	<h2 ref="headerRef"
+		v-linkify="{ text: name, linkify }"
+		:tabindex="-1"
```
ShGKme, line 33:
```suggestion
		tabindex="-1"
```

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -600,6 +593,11 @@ export default {
 							<slot class="app-sidebar-header__background" name="header" />
 						</div>
 
+						<!-- a11y fallback for empty content -->
+						<NcAppSidebarHeader v-if="empty"
+							class="hidden-visually"
+							:name
+							:tabindex="-1" />
```
ShGKme: It should be outside the info block, as the entire info is supposed to be hidden in an empty state

ShGKme (approved):
I like it

<details>

image


</details>

susnux (approved):

PR #6683, opened 2025-04-02
feat(NcActions): fix listening to wrong event names and add `opened` event / [NcPopover] animation wait

skjnldsv:
## NcAcions

Following https://github.com/nextcloud-libraries/nextcloud-vue/pull/6065

The evets are `after-hide` and `after-show`, not `apply-hide` and `apply-show`
https://github.com/nextcloud/nextcloud-vue/blob/ae77137aee5097dcbe8cef6c53eb1ce3bc81c7ee/src/components/NcPopover/NcPopover.vue#L230-L237

## NcPopover
The vue-floating lib does not properly compute how long it needs to wait to know the popup have been shown/hidden.
I investigated and it's [only using `requestAnimationFrame`](https://github.com/Akryum/floating-vue/blob/19857764c4f73dea7ed44a7d970adb968ee7ad90/packages/floating-vue/src/components/Popper.ts#L728-L782) which in that case _isn't_ enough to ensure its completion.
Because we have our own custom animation variables, we need to manually check for it ourselves.

Tested on Server with log entries to see the duration ([we use `--animation-quick`](https://github.com/nextcloud/nextcloud-vue/blob/ed4b48d50d219b5e52fc5bd96eb72329ecba3959/src/components/NcPopover/NcPopover.vue#L511-L521) which is 100ms for us)
| Before | After |
|:---------:|:------:|
|image||

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -1179,12 +1179,15 @@ export default {
 	},
 
 	emits: [
-		'open',
-		'update:open',
-		'close',
-		'focus',
-		'blur',
 		'click',
+		'blur',
+		'focus',
+
+		'close',
+		'closed',
+		'open',
+		'opened',
+		'update:open',
 	],
 
 	setup(props) {
@@ -1471,10 +1474,17 @@ export default {
 		/**
 		 * Called when popover is shown after the show delay
 		 */
-		onOpen() {
+		onOpened() {
 			this.$nextTick(() => {
 				this.focusFirstAction(null)
 				this.resizePopover()
+
+				/**
+				 * Event emitted when the popover menu is opened.
+				 *
+				 * This event is emitted after `update:open` was emitted and the opening transition finished.
+				 */
+				this.$emit('opened')
 			})
 		},
 
@@ -1901,9 +1911,9 @@ export default {
 					},
 					on: {
 						show: this.openMenu,
-						'apply-show': this.onOpen,
+						'after-show': this.onOpened,
 						hide: this.closeMenu,
-						'apply-hide': this.onClosed,
+						'after-hide': this.onClosed,
 					},
 				},
 				[
--- a/src/components/NcPopover/NcPopover.vue
+++ b/src/components/NcPopover/NcPopover.vue
@@ -174,7 +174,6 @@ import NcPopoverTriggerProvider from './NcPopoverTriggerProvider.vue'
  * @typedef {import('focus-trap').FocusTargetValueOrFalse} FocusTargetValueOrFalse
  * @typedef {FocusTargetValueOrFalse|() => FocusTargetValueOrFalse} SetReturnFocus
  */
-
 export default {
 	name: 'NcPopover',
 
@@ -239,6 +238,7 @@ export default {
 	data() {
 		return {
 			internalShown: this.shown,
+			animationDuration: 100,
 		}
 	},
 
@@ -254,6 +254,7 @@ export default {
 
 	mounted() {
 		this.checkTriggerA11y()
+		this.animationDuration = parseInt(getComputedStyle(this.$el).getPropertyValue('--animation-quick')) || 100
 	},
 
 	beforeDestroy() {
@@ -385,6 +386,7 @@ export default {
 			await this.useFocusTrap()
 			this.addEscapeStopPropagation()
 
+			setTimeout(() => {
 			/**
 			 * Triggered after the tooltip was visually displayed.
 			 *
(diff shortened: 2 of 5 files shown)


src/components/NcActions/NcActions.vue
```diff
@@ -1471,10 +1481,17 @@ export default {
 		/**
 		 * Called when popover is shown after the show delay
 		 */
-		onOpen() {
+		onOpened() {
```
skjnldsv, line 1484: Only renamed for consistency with the `onClosed` method name

src/components/NcPopover/NcPopover.vue
```diff
@@ -385,22 +386,27 @@ export default {
 			await this.useFocusTrap()
 			this.addEscapeStopPropagation()
 
+			setTimeout(() => {
```
ShGKme, line 389: Instead of reading a CSS variable and adding a new task with timeout, can we use a native event here?

```js
this.getPopoverContentElement().addEventListener('transitionend', () => {
	this.$emit('after-show')
}, { once: true })
```

src/components/NcPopover/NcPopover.vue
```diff
@@ -385,22 +386,27 @@ export default {
...
 			 * This is different from the 'show' and 'apply-show' which
 			 * run earlier than this where there is no guarantee that the
 			 * tooltip is already visible and in the DOM.
 			 */
-			this.$emit('after-show')
+				this.$emit('after-show')
+			}, this.animationDuration)
 		},
-		afterHide() {
+		async afterHide() {
```
ShGKme, line 400: Why async?

ShGKme (approved):

src/components/NcPopover/NcPopover.vue
skjnldsv: leftover code :)

src/components/NcPopover/NcPopover.vue
skjnldsv: Love it yeah! Didn't realized `transitionend` was an option

src/components/NcActions/NcActions.vue
```diff
@@ -1471,10 +1474,17 @@ export default {
...
 			this.$nextTick(() => {
 				this.focusFirstAction(null)
 				this.resizePopover()
+
+				/**
+				 * Event emitted when the popover menu is opened.
+				 *
+				 * This event is emitted after `update:open` was emitted and the opening transition finished.
+				 */
+				this.$emit('opened')
```
susnux, line 1487: Feature not fix (new event)

ShGKme: Fix for NcPopover, feat for NcActions

src/components/NcActions/NcActions.vue
```diff
@@ -1901,9 +1911,9 @@ export default {
 					},
 					on: {
 						show: this.openMenu,
-						'apply-show': this.onOpen,
+						'after-show': this.onOpened,
```
susnux, line 1914: This will now run quite late.
`after-show` will already await the next tick and the focus trap, so the additional tick in `onOpened` will cause a 1 tick delay which seems to not be needed

src/components/NcActions/NcActions.vue
susnux: but the fix in ncpopover is unrelated to the feature for NcActions.
So this is semantically a feature PR ;)

susnux (approved):
Not sure about the additional tick we now have, but in general makes sense

src/components/NcActions/NcActions.vue
skjnldsv: I think it's best to stay consistent. I'd like to avoid to have three different timings to watch for. 
Two are perfectly fine:
1. we start closing
2. we are sure everything is finished and closed/hidden

src/components/NcActions/NcActions.vue
skjnldsv: Yeah, was missing :shrug:

src/components/NcPopover/NcPopover.vue
skjnldsv: > Instead of reading a CSS variable and adding a new task with timeout, can we use a native event here?
> 
> ```js
> this.getPopoverContentElement().addEventListener('transitionend', () => {
> 	this.$emit('after-show')
> }, { once: true })
> ```

Was testing this, let me push a followup if it works :hourglass:

skjnldsv:
> Instead of reading a CSS variable and adding a new task with timeout, can we use a native event here?
> 
> ```js
> this.getPopoverContentElement().addEventListener('transitionend', () => {
> 	this.$emit('after-show')
> }, { once: true })
> ```

Was still testing this, let me push a followup if it works :hourglass: 
I should have disabled auto merge, my bad

ShGKme:
A bit renamed for change log generation

skjnldsv:
@ShGKme tested `addEventListener('transitionend' a bit, and it is slower than a simple `setTimeout`. Weirdly waiting much longer and

| `setTimeout` | `transitionend` |
|:---------:|:------:|
||image|

skjnldsv:
Doesn't matter, much cleaner anyway
https://github.com/nextcloud-libraries/nextcloud-vue/pull/6687

ShGKme:
@skjnldsv On my test:
```
afterHide | 10967
transitionstart | 10992
afterHide:setTimeout | 11072
transitionend | 11085
```

So it's 13ms longer, than timeout, but the animation might also have started later, than we think in the timeout

ShGKme:
@skjnldsv On my test:
```
afterHide | 10967
transitionstart | 10992
afterHide:setTimeout | 11072
transitionend | 11085
```

So it's 13ms longer, than timeout, but the animation might also have started later, than we think in the timeout

PR #6684, opened 2025-04-02
refactor: migrate color utils to Typescript

susnux:
### ☑️ Resolves

- for #841 
- also added some tests

### 🏁 Checklist


src/utils/colors.ts
```diff
@@ -118,20 +125,19 @@ export const defaultPalette = [
...
-export function GenColors(steps) {
-	if (!steps) {
+export function generatePalette(steps: number): Color[] {
+	if (steps <= 0) {
 		steps = 6
 	}
 
 	if (steps === 4) {
-		return defaultPalette
+		return [...defaultPalette]
```
ShGKme, line 140: Why?
If to make it immutable, we need to clone Colors as well

tests/unit/utils/colors.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+			expect(defaultPalette).toHaveLength(12)
+		})
+
+		it('contains Nextcloud blue', () => {
+			const nextcloudBlue = defaultPalette.find((color) => color.color === '#0082c9')
+			expect(nextcloudBlue, 'Nextcloud blue not found').not.toBeUndefined()
+			expect(nextcloudBlue!.name).toBe('Nextcloud blue')
+		})
+
+		it('all colors are named', () => {
```
ShGKme, line 22:
```suggestion
		it('has all colors named', () => {
```

ShGKme (approved):

src/utils/colors.ts
susnux: I do not think its a problem of immutable colors but of the wrapping array as this is on module scope.
But good point.

ShGKme: > but of the wrapping array as this is on module scope.

Why is it a problem?

ShGKme: Colors inside the array are also in the module-scope

susnux: it should in general immutable IMHO

PR #6689, opened 2025-04-02
fix(NcAppContent): split size in range

susnux:
WIP


ShGKme:
🤔

ShGKme:
@GretaD Maybe you remember if it was needed?

PR #6725, opened 2025-04-05
chore: adjust package engines versions

susnux:
### ☑️ Resolves

This package does not even target Node.JS so we could even drop the `node` dependency for `engines.
But lets at least allow the current LTS version (22) along side with LTS 20.

### 🏁 Checklist

### DIFF
--- a/.github/workflows/command-l10n-update.yml
+++ b/.github/workflows/command-l10n-update.yml
@@ -39,7 +39,7 @@ jobs:
 
     steps:
       - name: Checkout ${{ needs.init.outputs.head_ref }}
-        uses: actions/checkout@v4
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           token: ${{ secrets.COMMAND_BOT_PAT }}
           fetch-depth: 0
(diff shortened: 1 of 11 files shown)


.github/workflows/command-l10n-update.yml
```diff
@@ -50,21 +50,10 @@ jobs:
...
-
-      - name: Set up node ${{ steps.package-engines-versions.outputs.nodeVersion }}
-        uses: actions/setup-node@v4
-        with:
-          node-version: ${{ steps.package-engines-versions.outputs.nodeVersion }}
-          cache: npm
-
-      - name: Set up npm ${{ steps.package-engines-versions.outputs.npmVersion }}
-        run: npm i -g npm@"${{ steps.package-engines-versions.outputs.npmVersion }}"
+          node-version-file: package.json
```
ShGKme, line 56: But it doesn't support `devEngines` yet, right?

package.json
```diff
@@ -160,8 +160,20 @@
...
+    "packageManager": {
+      "name": "npm",
+      "version": "^10",
+      "onFail": "error"
+    },
+    "runtime": {
+      "name": "node",
+      "version": "^22",
+      "onFail": "error"
+    }
```
ShGKme, line 176: I recall having a script to update all libs/apps

ShGKme (approved):

.github/workflows/command-l10n-update.yml
susnux: https://github.com/actions/setup-node/issues/1255

susnux: So no but works as long as devEngines are in the upper bound of the engines.
IMHO browser libraries like this do not even need any `engine` as they only target the browser not Node (in contrast to e.g. vite-config which targets Node).

skjnldsv (approved):

PR #6727, opened 2025-04-05
docs(NcDateTimePicker): add prop documentation for all props

susnux:
### ☑️ Resolves

- no functional changes
- document every prop we provide
- add some developer docs on what some computed properties do
- also resort the props in consistent order

### 🏁 Checklist

### DIFF
--- a/src/components/NcDateTimePicker/NcDateTimePicker.vue
+++ b/src/components/NcDateTimePicker/NcDateTimePicker.vue
@@ -142,7 +142,7 @@ picked time zone, but you will have to convert it yourself when necessary.
 		<NcDateTimePicker
 			v-model="time"
 			type="datetime"
-			:show-timezone-select="true"
+			show-timezone-select
 			v-model:timezone-id="tz" /><br>
 		{{ time }} | {{ tz }}
 	</span>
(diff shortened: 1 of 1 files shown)


ShGKme (approved):

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -207,26 +217,28 @@ const props = withDefaults(defineProps<{
...
 
-	/**
-	 * Default increment step for minutes in the time picker.
-	 * @default 10
-	 */
-	minuteStep?: number
-
 	/**
 	 * Preview format for the picker input field.
+	 * Can either be a string of Unicode tokens or a function that takes a date,
```
Antreesy, line 222: I would specify here 
```suggestion
	 * Can either be a string of Unicode tokens or a function that takes a Date object
```

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -207,26 +217,28 @@ const props = withDefaults(defineProps<{
...
+	 * or for range picker an array of two dates, and returns the formatted date as string.
 	 *
 	 * @default Intl.DateTimeFormat is used to format dates and times
 	 * @see https://www.unicode.org/reports/tr35/tr35-dates.html#Date_Field_Symbol_Table
 	 */
-	format?: string
+	format?: string | ((date: Date) => string) | ((dates: Date[]) => string)
```
Antreesy, line 228: Tuples aren't a bad practice, as Enums, or I'm missing something? 
```suggestion
	format?: string | ((date: Date) => string) | ((dates: [Date, Date]) => string)
```

Antreesy (approved):

src/components/NcDateTimePicker/NcDateTimePicker.vue
ShGKme: No, tuple is a type and enum is not (not only)

src/components/NcDateTimePicker/NcDateTimePicker.vue
susnux: makes sense, thank you!

src/components/NcDateTimePicker/NcDateTimePicker.vue
susnux: Tuple are just more restrictive array types, in this case it is exactly two Date objects.
(Which is not true without the outher PR as for partial range it could have emitted `[Date, null]` but we fixed that :heavy_check_mark: )

PR #6736, opened 2025-04-06
chore: update pull_request_template.md

ShGKme:
### ☑️ Resolves

Otherwise "unchecked" might mean both "it is not needed" and "checklist is not complete"

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -16,4 +16,4 @@ B | A
 
 - [ ] ⛑️ Tests are included or are not applicable
 - [ ] 📘 Component documentation has been extended, updated or is not applicable
-- [ ] 2️⃣ Backport bugfixes to `stable8` for maintained Vue 2 version.
+- [ ] 2️⃣ Backport to `stable8` for maintained Vue 2 version or not applicable


.github/pull_request_template.md
```diff
@@ -16,4 +16,4 @@ B | A
 
 - [ ] ⛑️ Tests are included or are not applicable
 - [ ] 📘 Component documentation has been extended, updated or is not applicable
-- [ ] 2️⃣ Backport bugfixes to `stable8` for maintained Vue 2 version.
```
ShGKme, line 19: `bugfixes` removed, as we need to have `stable8` forward compatible as much as possible until most of the apps are migrated

susnux: How is this related? We can have v9 exclusive features as well v8 is then still forward compatible.
Only breaking changes need to have forward replacements.

susnux (approved):

.github/pull_request_template.md
ShGKme: Sorry, I meant, feature parity (where possible).

Antreesy (approved):

.github/pull_request_template.md
susnux: V8 featured should be in v9, but other way for now yes later only if needed by specific apps due to complexity of backporting (later = in a year or so)

PR #6737, opened 2025-04-06
chore: type check project on CI

susnux:
- built on top of https://github.com/nextcloud-libraries/nextcloud-vue/pull/6735

### ☑️ Resolves

1. enforce project to be typescript compliant
2. fix found issues
  - Missing type import in NcButton
  - Wrong types in NcDateTimePicker (less strict types only)
  - Missing event in NcDialog 

### 🏁 Checklist

### DIFF
--- a/.github/workflows/lint-typescript.yml
+++ b/.github/workflows/lint-typescript.yml
@@ -1,0 +1,39 @@
+# SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+# SPDX-License-Identifier: MIT
+
+name: Type checking
+
+on: pull_request
+
+permissions:
+  contents: read
+
+concurrency:
+  group: lint-typescript-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    name: Type checking
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+
+      - name: Set up node
+        uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v3
+        with:
+          node-version-file: package.json
+
+      - name: Install dependencies
+        env:
+          CYPRESS_INSTALL_BINARY: 0
+          PUPPETEER_SKIP_DOWNLOAD: true
+        run: npm ci
+
+      - name: Check types
+        run: |
+          npm run --if-present check-types
+          npm run --if-present ts:check
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
     "test:component": "playwright test -c playwright.config.ts",
     "test:component:gui": "playwright test --ui -c playwright.config.ts",
     "test:coverage": "TZ=UTC vitest run --coverage",
+    "ts:check": "vue-tsc --noEmit",
     "update:snapshots": "npm run test:component -- --grep @visual --update-snapshots",
     "watch": "npm run dev:watch"
   },
@@ -154,6 +155,7 @@
     "vue-eslint-parser": "^10.1.1",
     "vue-material-design-icons": "^5.3.1",
     "vue-styleguidist": "^4.72.4",
+    "vue-tsc": "^2.2.8",
     "webpack": "^5.98.0",
     "webpack-merge": "^6.0.1"
   },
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -60,6 +60,7 @@
     "styleguide:build": "vue-styleguidist --config styleguide.config.cjs build",
     "stylelint": "stylelint \"src/**/*.vue\" \"src/**/*.scss\" \"src/**/*.css\"",
     "stylelint:fix": "stylelint \"src/**/*.vue\" \"src/**/*.scss\" \"src/**/*.css\" --fix",
+    "check-types": "vue-tsc --noEmit",
```
ShGKme, line 63: I seems that this script is usually called `typecheck` ot `type-check` (at least in Vute/Vue ecosystem). For example:
- [create-vue/template - `type-check`](https://github.com/vuejs/create-vue/blob/3498eb178d3f191b52593235dd0ca8691708c020/template/config/typescript/package.json#L5):
- [vite codebase - `typecheck`](https://github.com/vitejs/vite/blob/a0e1a0402648e0df60fb928ffd97b0230999990d/package.json#L25)

I'd go with `typecheck` to not intersect with existing package 👀

.github/workflows/lint-typescript.yml
```diff
@@ -0,0 +1,37 @@
...
+name: Type checking
+
+on: pull_request
+
+permissions:
+  contents: read
+
+concurrency:
+  group: lint-eslint-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
```
ShGKme, line 13: A typo or intentionally `eslint`?

package.json
ShGKme: Or `ts:check` (personal preference)

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -418,15 +425,15 @@ const pickerType = computed(() => ({
...
+function onUpdateModelValue(value: LibraryModelValueType): void {
 	let date = value as Date | [Date, Date]
 	if (props.type === 'month') {
 		const data = value as { month: number, year: number }
 		date = new Date(data.year, data.month, 1)
 	} else if (props.type === 'year') {
 		date = new Date(value as number, 0)
 	} else if (props.type === 'week') {
-		date = value[0]
+		date = value?.[0]
```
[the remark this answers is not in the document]
susnux: no but the value can be `null` if it was cleared.

If there is a value it will be `[Date, Date]` (start of week , end of week).
If the value was cleared (`clearable=true`) then it will be just `null`.

.github/workflows/lint-typescript.yml
susnux:
```suggestion
concurrency:
  group: lint-typescript-${{ github.head_ref || github.run_id }}
  cancel-in-progress: true
```

package.json
susnux: In our libraries / apps:

- `type-check`
  - notify_push-client
  - nextcloud-moment
  - nextcloud-logger
  - nextcloud-browser-storage
- `test-types`
  - nextcloud-axios

Antreesy: Should we can go with something consistent here, which can be later moved to workflows for all other repos?

susnux: I would either stick with `check-types` as we do everywhere or prefer something like `lint:types` or `test:types` .
But I do not have a strong opinion on the naming and would accept any reasonable suggestion on how to proceed.

src/components/NcDateTimePicker/NcDateTimePicker.vue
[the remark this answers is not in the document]
ShGKme: Ah, I missed that `LibraryModelValueType` is different from the previous type

package.json
ShGKme: > In our libraries / apps:

Updated :P

ShGKme: `test` without a test runner seems wierd to me

susnux: @ShGKme `spreed` uses `typescript:check` not `typecheck`: https://github.com/nextcloud/spreed/blob/d215b474519e6b386cde00df1a989ffd63712305/package.json#L12

susnux: But yes as said I am fine with every name as long as it makes sense.
Currently I feel like `check-types` is the consistent name.
Alternatively `type-check` to follow Vite and Vue ecosystem.

> test without a test runner seems weird to me

well its a *test* that checks *types* so `test:types` run by tsc ;)
No but `test:...` also feels wrong, maybe `lint:types` but also weird. Probably best to either stick with one of the first mentioned.

skjnldsv: Asking the important questions I see :grin: 
I personally prefer `typecheck` or `typescript:check`

ShGKme: Or `ts:check` to have fewer letters to type 🦥

ShGKme (approved):
Approving the current state. We can change the script name after the discussion anytime

package.json
susnux: Renamed.

PR #6749, opened 2025-04-07
perf(utils): ensure only 64px or 512px avatars are loaded

susnux:
### ☑️ Resolves

The backend only supports those values,
so loading any other size will result in one of them but will be cached under a different URL by the browser. This should reduce the avatar queries of the browser to only max. 2 per user.

### 🏁 Checklist

- [x] 2️⃣ Backport bugfixes to `stable8` for maintained Vue 2 version.

### DIFF
--- a/src/composables/useIsDarkTheme/index.ts
+++ b/src/composables/useIsDarkTheme/index.ts
@@ -3,10 +3,10 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-import type { DeepReadonly, Ref } from 'vue'
+import type { DeepReadonly, MaybeRefOrGetter, Ref } from 'vue'
 
 import { createSharedComposable, usePreferredDark, useMutationObserver } from '@vueuse/core'
-import { ref, readonly, watch, inject, computed } from 'vue'
+import { computed, inject, ref, readonly, toValue, watch } from 'vue'
 import { checkIfDarkTheme } from '../../functions/isDarkTheme/index.ts'
 import { INJECTION_KEY_THEME } from './constants.ts'
 
@@ -17,17 +17,21 @@ import { INJECTION_KEY_THEME } from './constants.ts'
  * @param el - The element to check for the dark theme enabled on (default is `document.body`)
  * @return {DeepReadonly<Ref<boolean>>} - computed boolean whether the dark theme is enabled
  */
-export function useIsDarkThemeElement(el: HTMLElement = document.body): DeepReadonly<Ref<boolean>> {
-	const isDarkTheme = ref(checkIfDarkTheme(el))
+export function useIsDarkThemeElement(el?: MaybeRefOrGetter<HTMLElement>): DeepReadonly<Ref<boolean>> {
+	const element = computed(() => toValue(el) ?? document.body)
+
+	const isDarkTheme = ref(checkIfDarkTheme(element.value))
 	const isDarkSystemTheme = usePreferredDark()
 
 	/** Update the isDarkTheme */
 	function updateIsDarkTheme() {
-		isDarkTheme.value = checkIfDarkTheme(el)
+		isDarkTheme.value = checkIfDarkTheme(element.value)
 	}
 
 	// Watch for element change to handle data-theme* attributes change
-	useMutationObserver(el, updateIsDarkTheme, { attributes: true })
+	useMutationObserver(element, updateIsDarkTheme, { attributes: true })
+	// Watch for ref changes
+	watch(element, updateIsDarkTheme)
 	// Watch for system theme change for the default theme
 	watch(isDarkSystemTheme, updateIsDarkTheme, { immediate: true })
 


src/components/NcAvatar/NcAvatar.vue
```diff
@@ -417,6 +420,16 @@ export default {
 			default: 'body',
 		},
 	},
+
+	setup() {
+		const root = useTemplateRef('main')
+		const isDarkTheme = useIsDarkThemeElement(root)
```
Antreesy, line 426: Is there a need to check specifically `main` and not `body` and thus use shared composable?

If so, I would at least go with `props.menuContainer`, so it inherits it from container

src/components/NcDateTimePickerNative/index.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * SPDX-FileCopyrightText: 2022 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+export type * from './NcDateTimePickerNative.vue'
```
Antreesy, line 6: 👀

src/utils/getAvatarUrl.ts
```diff
@@ -4,12 +4,47 @@
...
+	/**
+	 * Is the user a guest user.
+	 */
+	isGuest?: boolean
+
+	/**
+	 * Size of the avatar.
+	 * @default 64
+	 */
+	size?: number
```
Antreesy, line 24:
```suggestion
	size?: 64 | 512
```

but then your .spec.ts are going to complain, right?)

src/components/NcDateTimePickerNative/index.ts
susnux: wrong file committed :see_no_evil:

src/components/NcAvatar/NcAvatar.vue
ShGKme: > Is there a need to check specifically `main` and not `body` and thus use shared composable?

To avoid misunderstanding, `main` here is template ref key, not `<main>` query selector

Antreesy: > how does talk handle this currently?

`<div class="top-bar__wrapper" :data-theme-dark="isInCall">` 🙈

src/composables/useIsDarkTheme/index.ts
```diff
@@ -15,19 +15,22 @@ import { checkIfDarkTheme } from '../../functions/isDarkTheme/index.ts'
...
+		isDarkTheme.value = checkIfDarkTheme(element.value)
 	}
 
 	// Watch for element change to handle data-theme* attributes change
-	useMutationObserver(el, updateIsDarkTheme, { attributes: true })
+	useMutationObserver(element, updateIsDarkTheme, { attributes: true })
 	// Watch for system theme change for the default theme
 	watch(isDarkSystemTheme, updateIsDarkTheme, { immediate: true })
+	// Watch for element changes
+	watch(element, updateIsDarkTheme)
```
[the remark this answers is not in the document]
susnux: Element updated?

src/components/NcRichContenteditable/NcAutoCompleteResult.vue
```diff
@@ -81,25 +83,30 @@ export default {
 			default: () => ({}),
 		},
 	},
+
+	setup() {
+		const root = useTemplateRef('root')
+		const isDarkTheme = useIsDarkThemeElement(root)
```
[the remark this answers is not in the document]
ShGKme: Or in the component provider, both are fine for me.

```html
<!-- before -->
<div data-theme-dark>
  ...content...
</div>

<!-- after -->
<NcThemeProvider dark>
  ...content...
</NcThemeProvider>
```

Only for the cases when we override the theme.

I worry about it because overriding the theme is quite a rare case, but to support it this PR adds observer for every avatar instance, and we may have a lot.

src/utils/getAvatarUrl.ts
```diff
@@ -4,12 +4,47 @@
...
+		: 512
+
+	const guestUrl = options?.isGuest
+		? '/guest'
+		: ''
+	const themeUrl = options?.isDarkTheme ?? checkIfDarkTheme(document.body)
+		? '/dark'
+		: ''
+
+	return generateUrl(`/avatar${guestUrl}/{user}/{size}${themeUrl}`, {
```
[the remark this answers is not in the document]
susnux: Maybe :grey_question: 
Could do so in a follow up.

ShGKme (approved):

src/components/NcRichContenteditable/NcMentionBubble.vue
```diff
@@ -57,14 +57,23 @@ export default {
...
+
 	computed: {
 		avatarUrl() {
 			if (this.iconUrl) {
 				return this.iconUrl
 			}
 
 			return this.id && this.source === 'users'
-				? this.getAvatarUrl(this.id, 44)
+				? this.getAvatarUrl(this.id, { isDarkTheme: this.isDarkTheme })
```
Antreesy: missing import?

```suggestion
				? getAvatarUrl(this.id, { isDarkTheme: this.isDarkTheme })
```

src/components/NcAvatar/NcAvatar.vue
```diff
@@ -420,6 +420,16 @@ export default {
...
```
ShGKme, line 426: I am not sure it is a performance improvement if every single avatar now adds mutation observer to support live theme change...

src/components/NcRichContenteditable/NcAutoCompleteResult.vue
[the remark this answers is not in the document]
ShGKme: cc @susnux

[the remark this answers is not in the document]
ShGKme: This issue is not resolved

src/components/NcAvatar/NcAvatar.vue
ShGKme: Avatars are actively used in lists, mentions and etc.
This component should be lightweight.

ShGKme: Is it related to the "ensure only 64px or 512px avatars are loaded"?

PR #6751, opened 2025-04-07
chore(deps): remove unused dependencies

ShGKme:
- `sanitize-filename` was used in Cypress
- `vue-eslint-parser` is used internally in `@nextcloud/eslint-config`


susnux (approved):

package.json
```diff
@@ -153,7 +152,6 @@
     "url-loader": "^4.1.1",
     "vite": "^6.2.4",
     "vitest": "^3.0.3",
-    "vue-eslint-parser": "^10.1.1",
```
susnux, line 156: We need this as we explicitly use it in the config. It does not work without flat config (the v9 version of the config).

ShGKme: 🙁

susnux: But at least `sanitize-filename` is fine.

susnux (approved):

Antreesy (approved):

PR #6752, opened 2025-04-07
fix(dialogs/spawnDialog)!: support vue-devtool but loose appContext

ShGKme:
### ☑️ Resolves

- Doing small changes from: https://github.com/nextcloud-libraries/nextcloud-vue/issues/6731
- Allows debugging dialog content with `vue-devtools`
- Also, `getCurrentInstance` only works when the dialog is spawned in the setup context, so doesn't even work in a click handler

This looses `appContext`.

Workarounds:
- Always use locally imported components
- When possible - provide data via props
- For `provide/inject` - provide directly from the root dialog component
- For Pinia - should work via `activePinia`, or via provide in the root dialog component

Alternative:
- Pass app/appContext via options and then do hacks - re-provide every provider

### 🏁 Checklist

### DIFF
--- a/docs/functions/spawnDialog.md
+++ b/docs/functions/spawnDialog.md
@@ -17,7 +17,6 @@ import {
  * @param dialog The dialog component to spawn - the component must emit the 'close' event whenever it is closed
  * @param props Properties to pass to the dialog
  * @param props.container Optionally pass a query selector for the dialog container element
- * @param props.appContext Optionally the app context to use (e.g. for registered components)
  * @param onClose Callback when the dialog is closed (parameters of the 'close' event of the dialog)
  */
 function spawnDialog(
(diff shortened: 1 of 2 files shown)


susnux (approved):
makes sense and is cleaner, but quite some negative aspects for just some DX.

ShGKme:
> makes sense and is cleaner, but quite some negative aspects for just some DX.

I tried many options and found no way to have both app context and independent rendering.

We can either hack with re-registering all global component, re-providing everything from providers.

Or manually add `vnode` to the parents `children` (IF there is a parent, and it's not mounted outside Vue apps).

ShGKme:
And as far as I know, there is no case there we need it at the moment.

susnux:
> We can either hack with re-registering all global component, re-providing everything from providers.

I think this is quite complex solution for a rare condition (when there is even one). So yes lets stick with this PR.

Antreesy (approved):
LGTM. Let's see how it works in app, when we get to it

PR #6756, opened 2025-04-08
feat(dialogs/spawnDialog): separate spawning options from dialog props and allow Element as a container

ShGKme:
### ☑️ Resolves

- One more small improvement from a queue
- Do not mix dialog options and 
- Coming next:
  - Fixing container usage
  - Return Promise
  - Typing

### 🏁 Checklist


docs/functions/spawnDialog.md
```diff
@@ -11,19 +11,41 @@ import {
...
+	props?: object,
+	onClose?: (...rest: unknown[]) => void,
+): void
+
+export function spawnDialog(
+	dialog: Component,
+	props?: object,
+	options?: SpawnDialogOptions,
+	onClose?: (...rest: unknown[]) => void,
+): void
```
susnux, line 33: Especially with this overload I would think that a promise is nicer (too many parameters IMHO)

```suggestion
export function spawnDialog(
	dialog: Component,
	props?: object,
	options?: SpawnDialogOptions,
): Promise<unknown[]>
```

susnux (approved):
non breaking change as such I approve. But I would really favorite to move the `onClose` callback to a promise based return.

docs/functions/spawnDialog.md
ShGKme: Removing `onClose` is a breaking change

(But it's fine to me)

ShGKme: https://github.com/nextcloud-libraries/nextcloud-vue/pull/6759

susnux (approved):

ShGKme:
> But I don't get what's the reason for moving container out? it should be delegated to Vue on where to mount component instead of 'appendChild'?

Do you mean why we create a new element **inside** the container and not mount directly **into** the container?

ShGKme:
If yes - because it allows using container for many dialogs and have other elements there.

For example, if you pass `container: 'body'`, you probably expect that the dialog will be mounted somewhere at the end of the body. And you don't expect that entire `body` content will be replaced with the dialog only.

src/functions/dialog/index.ts
```diff
@@ -5,38 +5,68 @@
...
-	container.appendChild(el)
+	if (typeof optionsOrOnClose === 'function') {
+		onClose = optionsOrOnClose
+		optionsOrOnClose = {}
+	}
+
+	let { container } = optionsOrOnClose
+
+	// For backwards compatibility try to use container from props
+	if ('container' in props && typeof props.container === 'string') {
```
susnux, line 51:
```suggestion
	if (container&& 'container' in props && typeof props.container === 'string') {
```

Do not override if already in opts

ShGKme: Did
```
if ('container' in props && typeof props.container === 'string') {
	container ??= props.container
}
```
instead.

(Seems slightly more readable to me to not think about brackets in `!container && 'container' in props`)

ShGKme: `!container && 'container' in props` vs `!container && ('container' in props)`

PR #6758, opened 2025-04-08
feat(dialogs/spawnDialog): return promise with the result

ShGKme:
### ☑️ Resolves

- Non braking alternative for: https://github.com/nextcloud-libraries/nextcloud-vue/pull/6759
- Not only call the `onClose` callback, but also return a promise

### 🏁 Checklist

### DIFF
--- a/src/functions/dialog/index.ts
+++ b/src/functions/dialog/index.ts
@@ -17,14 +17,14 @@ export function spawnDialog(
 	dialog: Component,
 	props?: object,
 	onClose?: (...rest: unknown[]) => void,
-): void
+): Promise<unknown>
 
 export function spawnDialog(
 	dialog: Component,
 	props?: object,
 	options?: SpawnDialogOptions,
 	onClose?: (...rest: unknown[]) => void,
-): void
+): Promise<unknown>
 
 /**
  * Spawn a single-use Vue dialog instance to get the result when it is closed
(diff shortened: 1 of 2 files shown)


src/functions/dialog/index.ts
```diff
@@ -5,38 +5,80 @@
...
+ * Spawn a single-use Vue dialog instance to get the result when it is closed
  *
- * @param dialog The dialog component to spawn
- * @param props Properties to pass to the dialog
- * @param props.container Optionally pass a query selector for the dialog container element
- * @param onClose Callback when the dialog is closed
+ * @param dialog - Dialog component to spawn
+ * @param props - Props to pass to the dialog instance
+ * @param optionsOrOnClose - Spawning options or a callback when the dialog is closed
+ * @param onClose - Callback when the dialog is closed
```
susnux, line 35:
```suggestion
 * @param onClose - Callback when the dialog is closed (*deprecated* use the returned promise instead)
```

src/functions/dialog/index.ts
```diff
@@ -5,38 +5,80 @@
...
+			...props,
+			// If dialog has no `container` prop passing a falsy value does nothing
+			// Otherwise it is expected that `null` disables teleport and mounts dialog in place like NcDialog/NcModal
+			container: null,
+			onClose(...rest: unknown[]) {
+				app.unmount()
+				element.remove()
+				const result = rest.map(v => toRaw(v))
+				onClose?.(...result)
+				resolve(result.length > 1 ? result : result[0])
```
susnux, line 73: So you have to emit array or object your self?

ShGKme: Yes. And in general, it's a common practice in Vue to always emit with one parameter in the payload. Then it's HTML-like and supports `$event` in template

ShGKme: But here, if you emiteg many values - you got an array.

And if you emitted 1 value - you got it directly

src/functions/dialog/index.ts
ShGKme: So deprecated or breaking change? 
- https://github.com/nextcloud-libraries/nextcloud-vue/pull/6759

src/functions/dialog/index.ts
susnux: but you only resolve the first argument

ShGKme: No, I only resolve the first argument when there are no arguments after the first

susnux: I see :see_no_evil:

PR #6759, opened 2025-04-08
feat(dialogs/spawnDialog)!: replace onClose callback with Promise

ShGKme:
### ☑️ Resolves

- Breaking alternative for https://github.com/nextcloud-libraries/nextcloud-vue/pull/6758
- Return promise **instead of** `onClose` callback

### 🏁 Checklist

### DIFF
--- a/src/functions/dialog/index.ts
+++ b/src/functions/dialog/index.ts
@@ -13,40 +13,20 @@ type SpawnDialogOptions = {
 	container?: Element | string
 }
 
-export function spawnDialog(
-	dialog: Component,
-	props?: object,
-	onClose?: (...rest: unknown[]) => void,
-): Promise<unknown>
-
-export function spawnDialog(
-	dialog: Component,
-	props?: object,
-	options?: SpawnDialogOptions,
-	onClose?: (...rest: unknown[]) => void,
-): Promise<unknown>
-
 /**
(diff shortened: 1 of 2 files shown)


src/functions/dialog/index.ts
```diff
@@ -5,38 +5,59 @@
...
+			onClose(...rest: unknown[]) {
+				app.unmount()
+				element.remove()
+				const result = rest.map(v => toRaw(v))
+				resolve(result.length > 1 ? result : result[0])
+			},
+			'onVue:unmounted'() {
+				app.unmount()
+				element.remove()
+				reject(new Error('Dialog was unmounted without close event'))
```
susnux, line 57: Maybe it makes sense to have a custom error to distinguish this, like:

```ts
class DialogUnmounted extends Error {}
```

But not sure if needed

susnux (commented):
looks good to me

src/functions/dialog/index.ts
ShGKme: I don't think so because this is an actual exception, not an application error thrown as an exception. Function users never need to distinguish this error to handle it differently. It only says that the function is used wrong.

susnux (approved):
I prefer this one for v9

Antreesy (approved):

PR #6761, opened 2025-04-08
refactor: deprecate `Tooltip` directive for accessibility

susnux:
### ☑️ Resolves

- port all usage within this library to native `title` attribute
- deprecate `Tooltip` directive, its recommended to use the browser native tooltips using `title`.

### 🏁 Checklist

### DIFF
--- a/src/directives/Tooltip/index.js
+++ b/src/directives/Tooltip/index.js
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-import { VTooltip, options } from 'floating-vue'
+import { vTooltip, options } from 'floating-vue'
 import './index.scss'
 
 options.themes.tooltip.html = false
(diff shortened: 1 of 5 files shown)


ShGKme (approved):

src/directives/Tooltip/index.js
```diff
@@ -3,13 +3,18 @@
...
 options.themes.tooltip.html = false
 options.themes.tooltip.delay = { show: 500, hide: 200 }
 options.themes.tooltip.distance = 10
 options.themes.tooltip['arrow-padding'] = 3
 
-export default VTooltip
-export { options }
+export {
+	/**
+	 * @deprecated Use the native `title` attribute instead.
```
Antreesy, line 16: In some cases NcPopover would be a good alternative, in case tooltip has some complex logic or styling inside ( so I'd mention it here)

Antreesy (approved):

src/directives/Tooltip/index.js
Antreesy: Probably because it was only us until 29 releases, maybe that's why =D

directive basically [allows same props](https://floating-vue.starpad.dev/api/#directive-options) as VDropdown component, so for heavy customization someone might consider it (we used `delay` back then, for example)

susnux: added documentation

Antreesy: Moving here from the follow-up PR: `NcRichContenteditable` is exactly the case where another solution should be used, and not the title visible on hover.

If maxlength exceeded, it's not shown anymore to the end-user, thus hiding error visibility

B | A
-- | --
image | image

PR #6770, opened 2025-04-11
fix(dialogs/spawnDialog): improve typing

ShGKme:
### ☑️ Resolves

```ts
// Props are checked to be valid MyDialog props
// result type is inferred
const result = await spawnDialog<typeof MyDialog>(MyDialog, { prop: 'value' })

// But doesn't work this way :(
const result = await spawnDialog(MyDialog, { prop: 'value' })
```

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/functions/dialog/index.ts
+++ b/src/functions/dialog/index.ts
@@ -5,6 +5,32 @@
 import type { Component } from 'vue'
 import { createApp } from 'vue'
 
+/**
+ * Utility type to extract props from Vue component
+ */
+type ComponentProps<T> =
+	T extends new (...args: unknown[]) => { $props: infer P; } ? NonNullable<P> :
+	T extends (props: infer P, ...args: unknown[]) => unknown ? P :
+	Record<string, never>
+
+/**
+ * Utility type to transform event name to corresponding prop name
+ */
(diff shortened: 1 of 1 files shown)


src/functions/dialog/index.ts
```diff
@@ -21,11 +47,15 @@ type SpawnDialogOptions = {
...
  */
-export function spawnDialog(
-	dialog: Component,
-	props: object = {},
+export function spawnDialog<
+	C extends Component,
+	Props extends Record<EventPropName<'close'>, (...payload: unknown[]) => unknown> = ComponentProps<C>, // eslint-disable-line space-before-function-paren
+	Payload extends unknown[] = EventPayload<NonNullable<Props[EventPropName<'close'>]>>,
+>(
+	dialog: C,
```
ShGKme, line 55: I don't understand why it works, we `C` is specified as `typeof MyDialog`, but it cannot infer it from `MyDialog` passed to the `dialog` param...

ShGKme:
- Closing in favor of https://github.com/nextcloud-libraries/nextcloud-vue/pull/6781

PR #6781, opened 2025-04-15
fix(dialogs/spawnDialog): enhance `spawnDialog` types

mattersj:
### ☑️ Resolves

- A next iteration of https://github.com/nextcloud-libraries/nextcloud-vue/pull/6770 that improves `spawnDialog` types:
- Now only components with a declared `@close` event can be passed to `spawnDialog`, otherwise an error is thrown on a type level suggesting what went wrong
- Autocompletion for props
- Return type is automatically inferred based on a `@close` event's payload

The implementation passes all the test cases from https://github.com/nextcloud-libraries/nextcloud-vue/pull/6770#issuecomment-2804453176

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/functions/dialog/index.ts
+++ b/src/functions/dialog/index.ts
@@ -2,8 +2,29 @@
  * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
-import type { Component } from 'vue'
-import { createApp } from 'vue'
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+import { createApp, type Component } from 'vue'
+
+type ComponentProps<T extends Component> = T extends new (...args: any) => { $props: infer P }
+	? P extends Record<string, any>
+		? P
+		: never
+	: never
+
+type DialogComponent<T extends Component> = 'onClose' extends keyof ComponentProps<T>
+	? T
+	: 'Please provide a Dialog Component that supports `@close` event'
+
+type NormalizedPayload<T> = T extends []
+	? void
+	: T extends [infer F]
+		? F
+		: T
+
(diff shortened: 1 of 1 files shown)


src/functions/dialog/index.ts
```diff
@@ -2,8 +2,29 @@
  * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
-import type { Component } from 'vue'
-import { createApp } from 'vue'
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+import { createApp, type Component } from 'vue'
```
ShGKme, line 7: We prefer to always separate type and non-type imports (not included in the old eslint config, using in this repo).

```suggestion
import { type Component } from 'vue'
import { createApp } from 'vue'
```

src/functions/dialog/index.ts
```diff
@@ -2,8 +2,29 @@
...
+
+type ComponentProps<T extends Component> = T extends new (...args: any) => { $props: infer P }
+	? P extends Record<string, any>
+		? P
+		: never
+	: never
+
+type DialogComponent<T extends Component> = 'onClose' extends keyof ComponentProps<T>
+	? T
+	: 'Please provide a Dialog Component that supports `@close` event'
```
ShGKme, line 17: That's the best `never` I've ever seen

ShGKme (approved):
Awesome 💙

src/functions/dialog/index.ts
mattersj: Sure, fixed.

susnux (approved):
:rocket:

PR #6792, opened 2025-04-16
refactor(NcChip): use `defineSlots` for proper slot definition

susnux:
### ☑️ Resolves

This resolves 2 errors reported by Typescript when building the package (missing "private" type exports during the declaration creation).

### 🏁 Checklist

### DIFF
--- a/src/components/NcChip/NcChip.vue
+++ b/src/components/NcChip/NcChip.vue
@@ -77,7 +77,6 @@ export default {
 			'nc-chip--no-icon': !hasIcon(),
 		}">
 		<span v-if="hasIcon()" class="nc-chip__icon">
-			<!-- @slot The icon slot can be used to set the chip icon. Make sure that the icon is not exceeding a height of `24px`. For round icons a exact size of `24px` is recommended. -->
 			<slot name="icon">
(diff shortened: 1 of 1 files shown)


src/components/NcChip/NcChip.vue
```diff
@@ -162,8 +162,18 @@ const props = withDefaults(defineProps<{
...
+	/**
+	 * Emitted when the close button is clicked
+	 */
+	close: []
+}>()
+
+const slots = defineSlots<{
+	actions(): unknown
+	default(): unknown
+	icon(): unknown
```
ShGKme, line 175:
```ts
import type { Slot } from 'vue'
```

+

```suggestion
	actions?: Slot
	default?: Slot
	icon?: Slot
```

ShGKme: I guess, that they don't want to make it "too complex", as in 90% cases you don't need it.

And because they might add new level magic to check slot content like for specific nodes.

susnux:
Partly already done - otherwise replaced by https://github.com/nextcloud-libraries/nextcloud-vue/pull/6903

PR #6799, opened 2025-04-17
fix(NcRichText): `references` prop type

julien-nc:
The references prop is passed to `NcReferenceList` and is processed as a list, not an object. This prevents Vue warnings and fixes the docs.

### 🏁 Checklist

### DIFF
--- a/src/components/NcRichText/NcReferenceList.vue
+++ b/src/components/NcRichText/NcReferenceList.vue
@@ -32,7 +32,7 @@ export default {
 			default: '',
 		},
 		referenceData: {
-			type: Object,
+			type: Array,
 			default: null,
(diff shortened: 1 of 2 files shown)


ShGKme (approved):

src/components/NcRichText/NcReferenceList.vue
```diff
@@ -32,7 +32,7 @@ export default {
 			default: '',
 		},
 		referenceData: {
-			type: Object,
+			type: Array,
 			default: null,
```
susnux, line 36: How can it be null but type `Array`? (null is object)
Should the default not be rather `() => []` (empty array?) or `undefined`?

PR #6811, opened 2025-04-20
fix(NcSelectUsers): add missing `search` event

susnux:
### ☑️ Resolves

The `search` event is commonly used to search for more users instead of preload all.
So we need to add this event as well.

Reported by @Chartman123 

### 🏁 Checklist

### DIFF
--- a/src/components/NcSelectUsers/NcSelectUsers.vue
+++ b/src/components/NcSelectUsers/NcSelectUsers.vue
@@ -165,7 +165,7 @@ export default {
 </docs>
 
 <script setup lang="ts">
-import { ref } from 'vue'
+import { ref, watch } from 'vue'
 
 import NcListItemIcon from '../NcListItemIcon/index.js'
 import NcSelect from '../NcSelect/index.js'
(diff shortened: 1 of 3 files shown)


src/components/NcSelectUsers/NcSelectUsers.vue
```diff
@@ -303,7 +303,17 @@ defineProps<{
  * The `v-model` directive may be used for two-way data binding.
  */
 defineModel<IUserData>('modelValue')
+
+const emit = defineEmits<{
+	/**
+	 * Emitted when the user enters some query.
+	 * This can be used to asynchronously fetch more options.
+	 */
+	search: [string]
```
ShGKme, line 312: Shall it be a model then?

ShGKme (approved):

src/components/NcSelectUsers/NcSelectUsers.vue
susnux: I also thought about that. Not sure if it is useful to pass the search as prop.
But would be fine with it.

Antreesy (approved):

src/components/NcSelectUsers/NcSelectUsers.vue
susnux: (also we cannot forward the search value to the VueSelect)

PR #6831, opened 2025-04-23
fix(NcAvatar): do not load avatar from server, if iconClass was provided

Antreesy:
### ☑️ Resolves

- Fix loading of faulty avatars (for non-users) from NcUserBubble

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 Tasks

- [ ] do we need explicit `is-no-user` prop at NcUserBubble for such cases?

### 🏁 Checklist

### DIFF
--- a/tests/unit/components/NcAvatar/NcAvatar.spec.ts
+++ b/tests/unit/components/NcAvatar/NcAvatar.spec.ts
@@ -4,8 +4,8 @@
  */
 
 import { mount, shallowMount } from '@vue/test-utils'
-import { describe, expect, it } from 'vitest'
-import { nextTick } from 'vue'
+import { beforeAll, afterAll, describe, expect, it, vi } from 'vitest'
+import { h, nextTick } from 'vue'
 import NcAvatar from '../../../../src/components/NcAvatar/NcAvatar.vue'
 
 describe('NcAvatar.vue', () => {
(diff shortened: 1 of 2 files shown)


tests/unit/components/NcAvatar/NcAvatar.spec.ts
```diff
@@ -115,4 +115,90 @@ describe('NcAvatar.vue', () => {
...
+		it('should render image with avatar url pointing to a user', async () => {
+			const wrapper = mount(NcAvatar, {
+				props: {
+					displayName: 'Guest',
+					user: 'user1',
+				},
+			})
+			/* Mock img.onload callback */
+			wrapper.vm.isAvatarLoaded = true
+			wrapper.vm.avatarUrlLoaded = avatarUrl
```
ShGKme, line 131: Test should never manually change the internal state of the component.

ShGKme (changes requested):

tests/unit/components/NcAvatar/NcAvatar.spec.ts
Antreesy: How to proceed then? Maybe mock an Image() constructor for these tests to emit load event?

susnux (approved):
makes sense

DorraJaouad (approved):

PR #6861, opened 2025-04-28
fix(NcAppNavigationItem): multi level padding

GretaD:
### ☑️ Resolves

- Fix https://github.com/nextcloud/mail/issues/11037

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot from 2025-04-28 12-18-14| Screenshot from 2025-04-28 13-46-58

### 🚧 Tasks

- [ ] Files have custom styling. https://github.com/nextcloud/server/blob/master/apps/files/src/components/FilesNavigationItem.vue#L100
 Check with them so they are aware this is fixed here.

### 🏁 Checklist

- [X] 2️⃣ Backport to `stable8` for maintained Vue 2 version or not applicable

### DIFF
--- a/src/assets/NcAppNavigationItem.scss
+++ b/src/assets/NcAppNavigationItem.scss
@@ -126,17 +126,21 @@
 }
 /* Second level nesting for lists */
 .app-navigation-entry__children {
+	--app-navigation-item-child-offset: 10px;
 	position: relative;
 	display: flex;
 	flex: 0 1 auto;
 	flex-direction: column;
 	width: 100%;
 	gap: var(--default-grid-baseline, 4px);
+	padding-inline-start: var(--app-navigation-item-child-offset);
 
 	.app-navigation-entry {
 		display: inline-flex;
 		flex-wrap: wrap;
-		padding-inline-start: $icon-size;
+	}
+	& & & & & & & {
+		--app-navigation-item-child-offset: 0;
 	}
 }
 


ShGKme (commented):
The level should be limited (or add scrolling), otherwise there is no place for deep nested items.

src/assets/NcAppNavigationItem.scss
```diff
@@ -132,11 +132,12 @@
 	flex-direction: column;
 	width: 100%;
 	gap: var(--default-grid-baseline, 4px);
+	padding-inline-start: 10px;
 
 	.app-navigation-entry {
 		display: inline-flex;
 		flex-wrap: wrap;
-		padding-inline-start: $icon-size;
+		padding-inline-start: 10px;
```
GretaD, line 140: i changed this to 10px, instead of 16px (icon-size), because the indentation is unnecessarily large.

susnux:
> The level should be limited (or add scrolling), otherwise there is no place for deep nested items.

I do not remember where but I think we had discussed this for files someday and designers mentioned a cap of 3rd level.

ShGKme:
Proposals for nesting limit:
1. CSS solution - use CSS variable for the padding and reset it on a specific level
```css
.app-navigation-entry__children {
  & & & & & & & {
    --app-navigation-item-child-offset: 0;
  }
}
```
2. Vue solution - provide depth and set class depending on the depth:
```js
const depth = inject('NcAppNavigationItem:depth', 0) + 1
provide('NcAppNavigationItem:depth', depth)
```

src/assets/NcAppNavigationItem.scss
```diff
@@ -126,17 +126,23 @@
...
 .app-navigation-entry__children {
+	--app-navigation-item-child-offset: 10px;
+
 	position: relative;
 	display: flex;
 	flex: 0 1 auto;
 	flex-direction: column;
 	width: 100%;
 	gap: var(--default-grid-baseline, 4px);
+	padding-left: var(--app-navigation-item-child-offset);
```
ShGKme, line 137:
```suggestion
	padding-inline-start: var(--app-navigation-item-child-offset);
```

ShGKme:
There are 2 places with `padding`, and it stops at different moments. So the last item with the padding has more padding than the following (Level 7 on the screenshot).

image

There is no problem if we either don't have the offset for the item at all, or we have it fixed, not changed depending on the level.

ShGKme:
I'd say there are 3 approaches:
1. Add padding to the container = every nested button is actually nested
   - Example: custom files
   - image
2. Add incremental padding for the button content = no padding for buttons, but LOOKS nested
   - Example: Windows
   - image
3. Mixed - nested padding for container + 1 level internal for being inline with the parent
   - Example: Files
   - image

> right, so what would you do?

1. Simple just remove the padding from the entry
2. More complex, requires JS (provide/inject approach)
3. Also not complex, needs a small adjustment. But I don't like it :D

I'd leave it for designers :P

PR #6869, opened 2025-05-02
fix(NcIconSvgWrapper): center svg span wrapper

skjnldsv:
Parent is flex, but not the intermediate span
See the folder icon:

### DIFF
--- a/src/components/NcIconSvgWrapper/NcIconSvgWrapper.vue
+++ b/src/components/NcIconSvgWrapper/NcIconSvgWrapper.vue
@@ -287,9 +287,7 @@ export default {
 
 	// Icon svg wrapper
 	span {
-		display: flex;
-		justify-content: center;
-		align-items: center;
+		line-height: 0;
 	}
 
 	&:deep(svg) {
--- a/src/components/NcIconSvgWrapper/NcIconSvgWrapper.vue
+++ b/src/components/NcIconSvgWrapper/NcIconSvgWrapper.vue
@@ -285,6 +285,13 @@ export default {
 		vertical-align: text-bottom;
 	}
 
+	// Icon svg wrapper
+	span {
+		display: flex;
+		justify-content: center;
+		align-items: center;
+	}
+
 	&:deep(svg) {
 		fill: currentColor;
 		width: v-bind('iconSize');


src/components/NcIconSvgWrapper/NcIconSvgWrapper.vue
```diff
@@ -285,6 +285,13 @@ export default {
 		vertical-align: text-bottom;
 	}
 
+	// Icon svg wrapper
+	span {
+		display: flex;
+		justify-content: center;
+		align-items: center;
```
susnux, line 292: No other flex needed, the span normally should just wrap the svg. But as we enforce a line height its at least that height. So just removing it will solve the problem.

```suggestion
		line-height: 0;
```

skjnldsv: Who are you with all your wizardry! :fearful:

src/components/NcIconSvgWrapper/NcIconSvgWrapper.vue
```diff
@@ -285,6 +285,11 @@ export default {
 		vertical-align: text-bottom;
 	}
 
+	// Icon svg wrapper
+	span {
+		line-height: 0;
+	}
```
ShGKme, line 291: I'd say that the root issue is that this span is a text (`inline`) element, which applies the `line-height`.

Instead of overriding text height to 0, I'd propose to make it a block or inline-block element.

I am unsure `line-height` won't be overridden by some a11y extension or have other browser-specific issues.

So I'd consider it a more risky solution than solving the root issue.

susnux: Block does not help as it still considers the line height.
Either set line height like here or use either `grid` (makes no sense) or `flex` which will  fit content.

ShGKme: > Block does not help as it still considers the line height.

Hmm, indeed

> `flex` which will fit content.

Yep, `flex`/`inline-flex`/`grid` works fine.

Or `display: contents`

> makes no sense

IMO, it makes more sense, as it sets a more relevant box type, than changing line-height of non-text 👀

ShGKme (approved):
Non blocking

PR #6871, opened 2025-05-02
fix(NcButton): re-export types

ShGKme:
### ☑️ Resolves

- Fix https://github.com/nextcloud-libraries/nextcloud-vue/issues/6870

### 🏁 Checklist


skjnldsv (approved):

susnux (approved):
Affects much more components.
Either fix all manually or we deprecate using the main import in general and simply do not export types there so you only import

```ts
import type { ButtonVariant } from '@nextcloud/vue/components/NcButton'
```

ShGKme:
> Affects much more components

This is the only component with type export

skjnldsv:
I stopped using the type, cannot tell 🙈

susnux:
converted to draft until we decided if this is the way we want it to be, ref:
https://github.com/nextcloud-libraries/nextcloud-vue/issues/6870#issuecomment-2909748056

susnux:
Reopen if we decide the other way around but otherwise lets keep types in component entry points

PR #6888, opened 2025-05-07
chore: move from `dev:watch` to `watch`

skjnldsv:
So, we use `watch` everywhere, literally for over 7 years.
I see some repos that are slowly usig this new `dev:watch` syntax. Can we please stick to the one we used for years ? :see_no_evil: 

- https://github.com/nextcloud-libraries/nextcloud-vue/blob/f27d675871a16916e944cfbb5e1af196fd26524b/package.json#L17
- https://github.com/nextcloud/server/blob/2e7cbc5e1b5caed31cef9fc39c3dbc838478e03a/package.json#L13

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
   "scripts": {
     "build": "vite build --mode production ",
     "dev": "vite build --mode development",
-    "dev:watch": "vite build --mode development --watch",
+    "watch": "vite build --mode development --watch",
     "l10n:extract": "node build/extract-l10n.mjs",
     "lint": "eslint src",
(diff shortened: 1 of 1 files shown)


ShGKme (approved):

susnux (approved):
I do not care, I thought dev:watch was the standard like having dev and dev:watch (similar test and test:coverage).

Should be documented somewhere.

skjnldsv:
https://docs.nextcloud.com/server/stable/developer_manual/digging_deeper/npm.html#npm-run-dev-npm-run-watch

PR #6894, opened 2025-05-11
[stable8] Fix npm audit

nextcloud-command:
# Audit report
No fixable problems found (4 unfixable, 39 only fixable manually using --force)


ShGKme (approved):
`@nextcloud/vite-config` updated from `1.5.4` to `1.5.6`

PR #6910, opened 2025-05-13
refactor(NcAppContentDetails): migrate component to Typescript

susnux:
### ☑️ Resolves

* for #841 

### 🏁 Checklist


src/components/NcAppContentDetails/NcAppContentDetails.vue
```diff
@@ -5,13 +5,18 @@
...
-}
+<script setup lang="ts">
+import type { Slot } from 'vue'
+
+defineSlots<{
+	/**
+	 * Provide content to the details view.
+	 * Passing content to this slot is required!
+	 */
+	default: Slot
```
ShGKme, line 20:
```suggestion
	default?: Slot
```

susnux: But this time its required

ShGKme: Vue does not support required slots, and this type doesn't define it. It is a type representing `context.slots` object.

susnux: > Vue does not support required slots, and this type doesn't define it. It is a type representing context.slots object.

But mostly this is describing it (as the object then needs to be available

ShGKme (approved):

Antreesy (approved):
Let's move `script` above `template`?

PR #6923, opened 2025-05-15
fix(docs): NcContent with skip-actions crashes styleguidist

ShGKme:
### ☑️ Resolves

- Fix https://github.com/nextcloud-libraries/nextcloud-vue/issues/6922
- Because `#skip-actions` was rendered by a Vue component in one of the examples, it had an unexpected crash on unmounted - there where no element to unmount
- Solution - render `#skip-actions` on the document level, not inside a specific Vue component example


ShGKme:
This PR also fixes the issue by not using NcContent in the docs:
- https://github.com/nextcloud-libraries/nextcloud-vue/pull/6913/files

kesselb (approved):
Thank you :pray:

susnux (approved):

PR #6932, opened 2025-05-19
feat(NcActionButton): introduce `description` prop

Antreesy:
### ☑️ Resolves

- There might be a need to use a button as 'action name' + 'description'
- Current solution: `name` prop + `default` slot is a bit bulky
- Alternative with this PR:  `default` slot + `subline` prop

### 🖼️ Screenshots

Documentation screenshot:
image

Same content comparison:
image

### 🚧 Tasks

- [ ] First commit: check if it's non-breaking

### 🏁 Checklist

### DIFF
--- a/src/assets/action.scss
+++ b/src/assets/action.scss
@@ -110,7 +110,7 @@
 			overflow: hidden;
 			white-space: nowrap;
 			max-width: 100%;
-			display: inline-block;
+			display: block;
 		}
 
 		&__menu-icon {
--- a/src/assets/action.scss
+++ b/src/assets/action.scss
@@ -113,6 +113,15 @@
 			display: block;
 		}
 
+		&__description {
+			display: block;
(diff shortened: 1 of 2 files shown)


nimishavijay:
Looks good! 

I'm wondering if we should align the icon with the title instead of centering it in the whole action item element. I remember @marcoambrosini had some thoughts about this?

marcoambrosini:
Yes, to tidy up things visually, the icon should be aligned with the title of the actionbutton

ShGKme:
> I'm wondering if we should align the icon with the title instead of centering it in the whole action item element. I remember @marcoambrosini had some thoughts about this?

It's unrelated, applied to all action items
- https://github.com/nextcloud-libraries/nextcloud-vue/issues/5284

susnux (approved):

PR #6936, opened 2025-05-20
fix(NcEmptyContent): description styling for mobile

Plesin:
Signed-off-by: Radek Pleskac <info@radekpleskac.com>

### ☑️ Resolves

- Fix https://github.com/nextcloud/server/issues/39398
- NcEmptyContent is a generic component used in multiple places, would like to verify more such places and also a couple diff languages, short and long ones, hence draft PR initially.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot 2025-05-20 at 11 27 13 | Screenshot 2025-05-20 at 11 20 34

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/NcEmptyContent/NcEmptyContent.vue
+++ b/src/components/NcEmptyContent/NcEmptyContent.vue
@@ -232,6 +232,8 @@ defineSlots<{
 
 	&__description {
 		color: var(--color-text-maxcontrast);
+		text-align: center;
+		text-wrap-style: balance;
 	}
 
 	&__action {


Antreesy:
Shall we set default padding (e.g. `var(--default-grid-baseline)` for both text nodes?
`balance` is a neat feature, but heading or description still can hit the borders

Plesin:
> Shall we set default padding (e.g. `var(--default-grid-baseline)` for both text nodes, or .empty-content itself? `balance` is a neat feature, but heading or description still can hit the borders

Sure, adding it to the `.empty-content` makes the possible edge cases safer. Will push in a bit.
Screenshot 2025-05-20 at 14 46 19

Antreesy (approved):
That was quick =) Awesome, thanks for the contribution!

Plesin:
> That was quick =) Awesome, thanks for the contribution!
Small css fix. Thanks for review and suggestion!
I was happy to try out the local dev setup.

ShGKme:
> Shall we set default padding (e.g. `var(--default-grid-baseline)` for both text nodes, or .empty-content itself?

This is a small, but breaking change. If the component is used in a places with a specific (correct) padding (let's say, 3 * base, inline with the content around, or common for content and no-content), then with this change it became 4 * base fo no-content.

nimishavijay (commented):
Design-wise this looks great, approving. No strong feelings about including or not including the padding because the contents shouldn't be that close to the edge anyway

marcoambrosini (approved):

Antreesy:
> This is a small, but breaking change

@ShGKme would be enough to mention it in changelog. so devs can check the styling in own apps?

ShGKme:
Fine by me, as this is only a problem with a long (multi-line) description, which is a rare case, and fine for center-aligned text (it usually has offset anyway).

But in general I'd avoid adding non-configurable default container padding.

ShGKme (approved):

PR #6966, opened 2025-05-27
fix(NcRichText): filter links based on protocol

max-nextcloud:
### ☑️ Resolves

Only allow links to specific protocols.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
grafik | grafik

### 🏁 Checklist

- [x] ⛑️ Tests are included
- [x] 📘 Component documentation is not applicable

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,6 +41,7 @@
         "remark-breaks": "^4.0.0",
         "remark-parse": "^11.0.0",
         "remark-rehype": "^11.1.1",
+        "remark-unlink-protocols": "^1.0.0",
         "splitpanes": "^4.0.3",
         "striptags": "^3.2.0",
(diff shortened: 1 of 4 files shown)


ShGKme:
> There are many potential use cases depending on your app.

It is not broken if it is not a link. Users still can copy the link manually.

For something app-specific we can add a prop for additional protocols (on request)

> * allow `nc://` to allow opening files in the desktop client

I don't think this would be a valid approach. What if there is no client? The link in the text should work by default, and opening via desktop should only be an additional feature, that providers can handle, like file link preview.

max-nextcloud:
Since you both approved and adding further protocols or a prop would be easy if we determine it's needed I'll move on with this.

PR #6968, opened 2025-05-27
refactor(NcDialog): migrate to script-setup and use Typescript for module

susnux:
### ☑️ Resolves

- for #841 

1. Migrate to script-setup
2. Use Typescript for entry points (index.ts)
3. Fix import extensions (.ts instead of .js) (not a bug as the bundler fixes it anyways but its nicer to have the real)

### 🏁 Checklist


src/utils/VueTypes.ts
```diff
@@ -9,3 +9,6 @@ import type { AllowedComponentProps, Component, VNodeProps } from 'vue'
 export type ComponentProps<C extends Component> = C extends new (...args: any) => any
 	? Omit<InstanceType<C>['$props'], keyof VNodeProps | keyof AllowedComponentProps | `on${string}`>
 	: never;
+
+type ClassType = string | Record<string, boolean | undefined>
+export type VueClassType = ClassType | ClassType[] | VueClassType[]
```
Antreesy: Recursion? Or it's fine here?

ShGKme (approved):

src/utils/VueTypes.ts
ShGKme: It is fine, because Vue class can be recursive, e.g.:
`:class="['a', ['b', 'c', { d: true }]]"`

susnux: Rebased as this is already in main branch ;)

PR #6993, opened 2025-06-02
fix(NcTextField): make trailing button icon for confirmation also work on RTL

susnux:
### ☑️ Resolves

- Fix https://github.com/nextcloud-libraries/nextcloud-vue/pull/6951#discussion_r2120928708

Add new value `arrowEnd` and deprecate `arrowRight` to be backported to v8.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🏁 Checklist


ShGKme (approved):

src/components/NcTextField/NcTextField.vue
```diff
@@ -192,11 +193,42 @@ defineSlots<{
...
+const isDirectionalIcon = computed(() => props.trailingButtonIcon === 'arrowEnd' || props.trailingButtonIcon === 'arrowRight')
+
+/**
+ * The MDI icon path to use for the specified icon.
+ */
+const trailingButtonIconPath = computed(() => {
+	switch (props.trailingButtonIcon) {
+	case 'arrowEnd':
+	case 'arrowRight':
+		return mdiArrowRight
```
Antreesy: Wouldn't be more readable that way? And then you would need to remove one line to deprecate
```suggestion
const trailingButtonIconComponent = computed(() => {
	switch (props.trailingButtonIcon) {
	case 'arrowEnd':
	case 'arrowRight':
		return {
			path: mdiArrowRight,
			label: t('Save changes'),
			directional: true,
		}
```

ShGKme: For readability I'd get read of switch case labels and just go with a mapping for the icon and just 2 cases in template instead of a universal node.

susnux: Why so many people hate `switch-case` :pensive:

susnux: Simplified

susnux:
/backport 9c7aea56c4892d1e40b391cc740e44170b09e485 to stable8

src/components/NcTextField/NcTextField.vue
ShGKme: > Why so many people hate `switch-case` 😔

Because it's kind of a goto with the only case for labels, usually has weird spacing, with a risk of lost `break`, longer and imperative.

Unlike FP langs with pattern matching and langs like Java where switch checks that you didn't forget any value case, JS old-school switch-case doesn't bring anything in comparison to declarative and short key-value map, or simple if-else 👀

susnux: Benefit: Tooling warns you if you forget a value, for if-else-if this is not really feasible.
For this reason I would like to have a `match` like PHP has, but JS alternative of something like:
```js
const result = {
foo: 'a',
bar: 'b',
}[key]
```

looks a bit ugly and again requires explicit typing so that tooling can catch invalid or missing keys.

PR #7010, opened 2025-06-05
fix(NcAppContent): incorrect page title from a different `core.apps` format in Nextcloud 30

julien-nc:
Main issue: The `core -> apps` initial state is treated like an object in NcAppContent but [it is an array](https://github.com/nextcloud/server/blob/master/lib/private/TemplateLayout.php#L73).

So in the page title, instead of the current app's name, the `core -> active-app` initial state is used. In the case of the Files app, this value is not what we want, it's "Nextcloud" which leads to page titles like
`CURRENT_DIR - All files - Nextcloud - THEMING_INSTANCE_NAME`
instead of the expected
`CURRENT_DIR - All files - Files - THEMING_INSTANCE_NAME`

### 🏁 Checklist


susnux:
> /backport to stable8

This is not 1:1 possible as on stable29 its an object (its an PHP array in all cases but serialized to array only on 30+).
So needs some more advanced logic on the backport.

julien-nc:
Ouch. Can we just use `Array.isArray()`?

PR #7031, opened 2025-06-10
[stable8] fix(NcDialog): deprecated nativeType usage check

skjnldsv:
`nativeType` is deprecated

### DIFF
--- a/src/components/NcDialog/NcDialog.vue
+++ b/src/components/NcDialog/NcDialog.vue
@@ -551,7 +551,7 @@ export default defineComponent({
 		 */
 		function handleButtonClose(button, result) {
 			// Skip close on submit if invalid dialog
-			if ((button.type ?? button.nativeType) === 'submit'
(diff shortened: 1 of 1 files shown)


src/components/NcDialog/NcDialog.vue
```diff
@@ -551,7 +551,7 @@ export default defineComponent({
 		 */
 		function handleButtonClose(button, result) {
 			// Skip close on submit if invalid dialog
-			if (button.nativeType === 'submit'
+			if ((button.type ?? button.nativeType) === 'submit'
```
susnux:
```suggestion
			if ((button.type === 'submit' || button.nativeType === 'submit)
```

I think we need to check both as `type` could be 'secondary' when using the legacy options.

Alternative:
```suggestion
			if ((button.nativeType ?? button.type) === 'submit'
```

susnux (approved):

PR #7036, opened 2025-06-11
Remove modal header for smart picker dialogs

marcoambrosini:
The modal header is unused in smartpicker dialogs and prevents using the closing button.
Required for https://github.com/nextcloud/assistant/pull/257

### DIFF
--- a/src/components/NcRichText/NcReferencePicker/NcReferencePickerModal.vue
+++ b/src/components/NcRichText/NcReferencePicker/NcReferencePickerModal.vue
@@ -186,4 +186,13 @@ export default {
 		}
 	}
 }
+// Hide the empty modal header which prevents from accessing the close button
+// Cannot use display:none because there's a `display: flex !important` in the
+// Ncmodal component which would override it. This works because the
+// `.modal-header` element has `position: absolute`.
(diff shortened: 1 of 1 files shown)


julien-nc:
I tried linking with Talk, it works fine :+1:

src/components/NcRichText/NcReferencePicker/NcReferencePickerModal.vue
```diff
@@ -186,4 +186,13 @@ export default {
 		}
 	}
 }
+// Hide the empty modal header which prevents from accessing the close button
+// Cannot use display:none because there's a `display: flex !important` in the
+// Ncmodal component which would override it. This works because the
+// `.modal-header` element has `position: absolute`.
+// TODO: Add prop to NcModal to hide the header
```
susnux, line 193: If this is a dialog then it should use `NcDialog` and not `NcModal` which also would fix this issue.

susnux:
The commit message needs to be adjusted to follow conventional commits.
Also its probably best to go through `main` branch first and backport - in this case here you have to forward port to `main` afterwards.

marcoambrosini:
closing as its been already removed

PR #7051, opened 2025-06-16
fix(NcDateTimePicker): correctly import library CSS

raimund-schluessler:
### ☑️ Resolves

This adjusts the CSS import of the `NcDateTimePicker` vendor styles to not be scoped. I don't know if this is the proper way to handle this issue, so plead advise if not.

- Fixes https://github.com/nextcloud-libraries/nextcloud-vue/issues/7050#issuecomment-2978166946.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Image | grafik

### 🏁 Checklist

### DIFF
--- a/src/components/NcDateTimePicker/NcDateTimePicker.vue
+++ b/src/components/NcDateTimePicker/NcDateTimePicker.vue
@@ -319,6 +319,7 @@ const props = withDefaults(defineProps<{
 const timezoneId = defineModel<string>('timezoneId', { default: 'UTC' })
 
 const target = useTemplateRef('target')
+const picker = useTemplateRef('picker')
 
 const emit = defineEmits<{
 	/**
@@ -555,11 +556,28 @@ const ariaLabels = computed(() => ({
 	monthPicker: (overlay: boolean) => overlay ? t('Month picker overlay') : t('Month picker'),
 	yearPicker: (overlay: boolean) => overlay ? t('Year picker overlay') : t('Year picker'),
 }))
+
+/**
+ * Select the current value.
+ * This is used by the confirmation button if `confirmation` was set.
+ */
+function selectDate() {
+	picker.value!.selectDate()
+}
+
+/**
+ * Cancel the current selection by closing the overlay.
+ * This is used by the confirmation button if `confirmation` was set.
+ */
+function cancelSelection() {
+	picker.value!.closeMenu()
+}
 </script>
 
 <template>
 	<div class="vue-date-time-picker__wrapper">
-		<VueDatePicker :aria-labels
+		<VueDatePicker ref="picker"
+			:aria-labels
 			:auto-apply="!confirm"
 			class="vue-date-time-picker"
 			:cancel-text="t('Cancel')"
@@ -580,6 +598,14 @@ const ariaLabels = computed(() => ({
 			:week-start
 			v-bind="pickerType"
 			@update:model-value="onUpdateModelValue">
+			<template #action-buttons>
+				<NcButton size="small" variant="tertiary" @click="cancelSelection">
+					{{ t('Cancel') }}
+				</NcButton>
+				<NcButton size="small" variant="primary" @click="selectDate">
+					{{ t('Pick') }}
+				</NcButton>
+			</template>
 			<template #clear-icon="{ clear }">
 				<NcButton :aria-label="t('Clear value')"
 					variant="tertiary-no-background"
--- a/src/components/NcDateTimePicker/NcDateTimePicker.vue
+++ b/src/components/NcDateTimePicker/NcDateTimePicker.vue
@@ -651,26 +651,26 @@ function cancelSelection() {
 
 .vue-date-time-picker__wrapper {
 	// This is under :root in @vuepic/vue-datepicker/dist/main.css, so importing it scoped won't work
-	--dp-common-transition: all 0.1s ease-in;
+	--dp-common-transition: all var(--animation-quick) ease-in;
 	--dp-menu-padding: 6px 8px;
-	--dp-animation-duration: 0.1s;
+	--dp-animation-duration: var(--animation-quick);
 	--dp-menu-appear-transition-timing: cubic-bezier(.4, 0, 1, 1);
 	--dp-transition-timing: ease-out;
 	--dp-action-row-transtion: all 0.2s ease-in;
-	--dp-font-family: -apple-system, blinkmacsystemfont, "Segoe UI", roboto, oxygen, ubuntu, cantarell, "Open Sans", "Helvetica Neue", sans-serif;
-	// --dp-border-radius: 4px;
-	--dp-cell-border-radius: 4px;
+	--dp-font-family: var(--font-face);
+	--dp-border-radius: var(--border-radius-element);
+	--dp-cell-border-radius: var(--border-radius-small);
 	--dp-transition-length: 22px;
-	--dp-transition-timing-general: 0.1s;
-	--dp-button-height: 35px;
-	--dp-month-year-row-height: 35px;
-	--dp-month-year-row-button-size: 25px;
+	--dp-transition-timing-general: var(--animation-quick);
+	--dp-button-height: var(--default-clickable-area);
+	--dp-month-year-row-height: var(--default-clickable-area);
+	--dp-month-year-row-button-size: var(--clickable-area-small);
 	--dp-button-icon-height: 20px;
 	--dp-calendar-wrap-padding: 0 5px;
-	--dp-cell-size: 35px;
+	--dp-cell-size: var(--default-clickable-area);
 	--dp-cell-padding: 5px;
 	--dp-common-padding: 10px;
-	// --dp-input-icon-padding: 35px;
+	--dp-input-icon-padding: var(--default-clickable-area);
 	--dp-input-padding: 6px 30px 6px 12px;
 	--dp-menu-min-width: 260px;
 	--dp-action-buttons-padding: 1px 6px;
@@ -678,11 +678,11 @@ function cancelSelection() {
 	--dp-calendar-header-cell-padding: 0.5rem;
 	--dp-multi-calendars-spacing: 10px;
 	--dp-overlay-col-padding: 3px;
-	--dp-time-inc-dec-button-size: 32px;
+	--dp-time-inc-dec-button-size: var(--default-clickable-area);
 	--dp-font-size: 1rem;
-	--dp-preview-font-size: 0.8rem;
+	--dp-preview-font-size: var(--font-size-small);
 	--dp-time-font-size: 2rem;
-	--dp-action-button-height: 22px;
+	--dp-action-button-height: var(--clickable-area-small);
 	--dp-action-row-padding: 8px;
 	--dp-direction: ltr;
 
@@ -692,18 +692,13 @@ function cancelSelection() {
 		@include meta.load-css('@vuepic/vue-datepicker/dist/main.css');
 	}
 
-	/*
-	 * These are our customizations to the date picker.
-	 */
-	--dp-border-radius: var(--border-radius-element);
-	--dp-input-icon-padding: var(--default-clickable-area);
-
 	.vue-date-time-picker__timezone {
 		min-width: unset;
 		width: 100%;
 	}
 
 	:deep(.icon-vue) {
+		// we enforce full opacity to not create a11y issues with contrast
 		opacity: 1 !important;
 	}
 
@@ -735,46 +730,18 @@ function cancelSelection() {
 	:deep(input) {
 		padding-inline-start: var(--dp-input-icon-padding) !important;
 	}
-	:deep(.dp__input) {
-		margin: 3px;
-		margin-inline-start: 0;
-		padding: 0 12px;
-		font-size: var(--default-font-size);
-		background-color: var(--color-main-background);
-		color: var(--color-main-text);
-		border: 2px solid var(--color-border-maxcontrast);
-		outline: none;
-		border-radius: var(--border-radius-large);
-		text-overflow: ellipsis;
-		cursor: pointer;
-	}
-	:deep(.dp__btn),
-	:deep(.dp--time-overlay-btn),
-	:deep(.dp__action_button) {
+	:deep(.dp__btn) {
 		margin: 0;
-		font-weight: bold;
-		border-radius: var(--border-radius-element);
-		padding: calc((var(--default-clickable-area) - 1lh) / 2) calc(3 * var(--default-grid-baseline));
-		font-size: var(--default-font-size);
-		width: auto;
-		min-height: var(--default-clickable-area);
-		cursor: pointer;
-		box-sizing: border-box;
-		color: var(--color-primary-element-light-text);
-		background-color: var(--color-primary-element-light);
-		border: none;
-	}
-	:deep(.dp--time-overlay-btn),
-	:deep(.dp__action_button) {
-		margin: 3px;
-		margin-inline-start: 0;
-	}
-	:deep(.dp__inc_dec_button) {
-		padding: calc((var(--default-clickable-area) - 20px) / 2);
 	}
 	:deep(.dp__inner_nav) {
 		height: fit-content;
 		width: fit-content;
+	}
+
+	// make the bottom page toggle stand out better
+	:deep(.dp__btn.dp__button.dp__button_bottom) {
+		color: var(--color-primary-element-light);
+		background-color: var(--color-primary-element-light);
 	}
 
 	// Fix server styles causing buttons to be primary colored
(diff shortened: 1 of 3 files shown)


raimund-schluessler:
@susnux @ShGKme Could you have a look at this, please? This problem really prevents using `NcDateTimePicker` at all.

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -606,9 +606,11 @@ const ariaLabels = computed(() => ({
 	</VueDatePicker>
 </template>
 
-<style scoped lang="scss">
+<style lang="css">
 @import '@vuepic/vue-datepicker/dist/main.css';
+</style>
```
ShGKme: This removes scoping from the external library completely. This works fine only until there is an update in the library changing styling/layout.

I'm checking the library's source to see if there is any workaround.

If there is no, then we need to go with this solution and be cautious with the library updates.

susnux: @raimund-schluessler can you check this works for you?
https://github.com/nextcloud-libraries/nextcloud-vue/pull/7066/files

Then we can keep the scoping

ShGKme: We can teleport it not to the body but to a specific manually created element with a specific class.
Then we can apply styles to both, the picker itself and the teleport target.

raimund-schluessler: > We can teleport it no to body, but to a specific manually created element with a specific class. Then we can apply styles to both, the picker itself and the teleport target.

That would be complicated, because this element needs to be positioned, if I understand it correctly. We could of course ask  `@vuepic/vue-datepicker` to add a  `teleport-class` prop that adds a specific class to the picker. But that would require getting the PR merged plus a new release of the lib.

I would propose to stay pragmatic here. We should use global styles for now, simply to get `@nextcloud/vue` working at all (and probably well in most cases) and additionally enhance the datepicker lib. We can then adjust nc/vue to be scoped once the datepicker lib got updated.

susnux: If I see it correctly we can do something like:

```
<template>
<span class="our-wrapper-class">
    <VueDatePicker class="vue-date-picker" />
</span>
<component :is="appendToBody ? 'teleport' : 'div'">
    <div ref="target" class="our-wrapper-class" />
</component>
</template>
```

Then we always pass `refs.target` to the `teleport` prop of `VueDatePicker` and scope the styles to `our-wrapper-class`. This would duplicate some styles but should not be a problem and in theory work.

ShGKme: > That would be complicated, because this element needs to be positioned,

This is already supported by the component. We can teleport it to any selector, not only `body`.

> We could of course ask `@vuepic/vue-datepicker` to add a `teleport-class` prop that adds a specific class to the picker

There is such a prop already. The problems are:
1. We don't have separated styles for the menu, so we'd need to duplicate all styles for the menu
2. We cannot apply style for the root element of the menu with our custom class

We need a class to the menu wrapper.

So we need to manually wrap the entire datepicker and create a wrapper to be used as teleport target.

Unfortunately, styles will be duplicated, I don't see a better solution.

> I would propose to stay pragmatic here. We should use global styles for now, simply to get `@nextcloud/vue` working at all (and probably well in most cases) and additionally enhance the date picker lib.

It is an unrevertable change.

Once we add it, any app using this `@nextcloud/vue` version may break the component for other apps in the future, and we won't be able to fix it.

We can prevent the library from polluting global scoped.

But we cannot save the library from already polluted global scope.

raimund-schluessler: And how do you import the CSS now without it being scoped? I don't seem to get it working.

susnux: As shown in my pr you can import the styles into another selector

raimund-schluessler: The append to body is not even the problem, but importing the CSS without the scope is what I don't get working.

raimund-schluessler:
@susnux Is this (roughly) what you had in mind? It works in the Tasks app, but there are still issues:
- [x] `appendToBody` still broken
- [x] Nextcloud styles not correctly applied yet when `appendToBody=true`
- [x] styleguide does not build, since the CSS import fails vs. npm run build fails
- [x] `@vuepic/vue-datepicker/dist/main.css` defines CSS variables under `:root`. Importing that nested in a class is not valid, it seems. Is there a better solution than manually duplicating the variables (which seems terrible, because minor changes in the lib might break the CSS).

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -555,59 +555,109 @@ const ariaLabels = computed(() => ({
...
+				<NcTimezonePicker v-model="timezoneId"
+					class="vue-date-time-picker__timezone"
+					:append-to-body="false"
+					:input-label="t('Timezone')" />
+			</template>
+		</VueDatePicker>
+	</span>
+	<Teleport to="body" :disabled="!appendToBody">
+		<div ref="target" class="date-time-picker-scope" />
+	</Teleport>
```
ShGKme: Let's move it inside the span. Otherwise this component has no root element, which is unexpected for a UI component.

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -555,59 +555,109 @@ const ariaLabels = computed(() => ({
...
+	--dp-time-font-size: 2rem;
+	--dp-action-button-height: 22px;
+	--dp-action-row-padding: 8px;
+	--dp-direction: ltr;
+
+	// we need to import the vuepic styles but at least scoped to our class and scope
+	// plain @import does not work as this will scope all styles imported.
+	:deep() {
+		// Importing like this does not work for webpack, it seems?
+		@include meta.load-css('@vuepic/vue-datepicker/dist/main.css');
```
ShGKme, line 692: In webpack it is not considered an npm module import without a trailing `~` which breaks Vite.

To fix styleguide, it requires adding `path.resolve(__dirname, 'node_modules')` to `sassOptions.includePaths[]`

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -189,7 +189,7 @@ import {
 	getDayNamesMin,
 	getCanonicalLocale,
 } from '@nextcloud/l10n'
-import { computed } from 'vue'
+import { computed, Teleport, useTemplateRef } from 'vue'
```
ShGKme: Teleport is a built-in component, it works without import.

ShGKme:
> * styleguide does not build, since the CSS import fails vs. npm run build fails

Fixable by adding `path.resolve(__dirname, 'node_modules')` to `sassOptions.includePaths[]` in the webpack config.

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -555,59 +557,109 @@ const ariaLabels = computed(() => ({
...
+	--dp-time-font-size: 2rem;
+	--dp-action-button-height: 22px;
+	--dp-action-row-padding: 8px;
+	--dp-direction: ltr;
+
+	// we need to import the vuepic styles but at least scoped to our class and scope
+	// plain @import does not work as this will scope all styles imported.
+	:deep() {
+		// Importing like this does not work for webpack, it seems?
+		@include meta.load-css('node_modules/@vuepic/vue-datepicker/dist/main.css');
```
susnux: Should be 
```suggestion
		@include meta.load-css('@vuepic/vue-datepicker/dist/main.css');
```

For webpack (styleguide) only the sass config needs to be adjusted

ShGKme:
> * Is there a better solution than manually duplicating the variables (which seems terrible, because minor changes in the lib might break the CSS).

I'd consider changing CSS variables here a breaking change for the library as it is a public API for theming.

And anyway, it is less dangerous than overriding many other styles in this component.

raimund-schluessler:
Ok, I think this now passes as a PoC. I will try to fix the remaining issues over the weekend if I have time.

ShGKme:
> * Nextcloud styles not correctly applied yet when `appendToBody=true`

Could you elaborate, what you mean?

raimund-schluessler:
> > * Nextcloud styles not correctly applied yet when `appendToBody=true`
> 
> 
> 
> Could you elaborate, what you mean?

The style adjustments we do are not applied yet, if the picker is appended to body, because the css classes are not correct yet. You can see this at the buttons at the top or the highlight color.
Just needs some css class adjustments.

raimund-schluessler:
The styles are adjusted now and work for `appendToBody` as well. But the time-picker elements are differently styled now than before and I couldn't yet figure out what is missing.

raimund-schluessler:
> The styles are adjusted now and work for `appendToBody` as well. But the time-picker elements are differently styled now than before and I couldn't yet figure out what is missing.

It seems the button server styles are not applied anymore, because they are overwritten by vuepic styles with higher priority now.

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -555,61 +557,111 @@ const ariaLabels = computed(() => ({
...
-			<NcIconSvgWrapper inline :path="mdiChevronUp" :size="20" />
-		</template>
-		<template v-if="showTimezoneSelect" #action-extra>
-			<NcTimezonePicker v-model="timezoneId"
-				class="vue-date-time-picker__timezone"
-				:append-to-body="false"
-				:input-label="t('Timezone')" />
-		</template>
-	</VueDatePicker>
+	<span class="vue-date-time-picker-wrapper">
```
susnux:
```suggestion
	<span class="vue-date-time-picker__wrapper">
```

keep it BEM

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -555,61 +557,111 @@ const ariaLabels = computed(() => ({
...
+	--dp-calendar-header-cell-padding: 0.5rem;
+	--dp-multi-calendars-spacing: 10px;
+	--dp-overlay-col-padding: 3px;
+	--dp-time-inc-dec-button-size: 32px;
+	--dp-font-size: 1rem;
+	--dp-preview-font-size: 0.8rem;
+	--dp-time-font-size: 2rem;
+	--dp-action-button-height: 22px;
+	--dp-action-row-padding: 8px;
+	--dp-direction: ltr;
```
susnux: TBH I do not like this as this is error prone if there are new styles added (non breaking library change).
But I see that there is little we can do about this.

Only alternative I see would be patching the imported CSS to replace the `:root` selector.

julien-nc:
The component can now be used. Picking a date works fine. But the clear button cannot be clicked.

image

src/components/NcDateTimePicker/NcDateTimePicker.vue
ShGKme: > Only alternative I see would be patching the imported CSS to replace the `:root` selector.

Or, as the component is not a transparent wrapper anymore, change the wrapped library to something wrapping friendly and styleless 👀

susnux: Did not find anything better maintained than this library.

raimund-schluessler: We can ask the library authors to accept a PR adjusting this.

raimund-schluessler:
The styles are now in line with server and I hope to have fixed everything. Please review again.

raimund-schluessler:
> The component can now be used. Picking a date works fine. But the clear button cannot be clicked.
> 
> image

Fix is in https://github.com/nextcloud-libraries/nextcloud-vue/pull/7103.

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -657,9 +709,42 @@ const ariaLabels = computed(() => ({
...
+		background-color: var(--color-primary-element-light);
+		border: none;
+	}
+	:deep(.dp--time-overlay-btn),
+	:deep(.dp__action_button) {
+		margin: 3px;
+		margin-inline-start: 0;
+	}
+	:deep(.dp__inc_dec_button) {
+		padding: calc((var(--default-clickable-area) - 20px) / 2);
```
ShGKme: Is it needed only after scoping or a new change?

raimund-schluessler: This might not necessarily be related to scoping. I think it depends in which order the server and library styles are included, which might be arbitrary. If server styles come after the library styles, it works fine. But if not, the library styles overwrite the server styles (mainly the ones for `button`) and the style looks broken.

raimund-schluessler:
Can we merge this for the next RC, please? It's really an annoying issue for apps.

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -556,67 +558,117 @@ const ariaLabels = computed(() => ({
...
+				<NcTimezonePicker v-model="timezoneId"
+					class="vue-date-time-picker__timezone"
+					:append-to-body="false"
+					:input-label="t('Timezone')" />
+			</template>
+		</VueDatePicker>
+		<Teleport to="body" :disabled="!appendToBody">
+			<div ref="target" class="vue-date-time-picker__wrapper" />
+		</Teleport>
+	</span>
```
ShGKme: It is `div` inside `span` which is not valid HTML. Also, `.vue-date-time-picker` is a div. Let's make a wrapper also a `div`.

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -556,67 +558,117 @@ const ariaLabels = computed(() => ({
...
-			<NcIconSvgWrapper inline :path="mdiChevronUp" :size="20" />
-		</template>
-		<template v-if="showTimezoneSelect" #action-extra>
-			<NcTimezonePicker v-model="timezoneId"
-				class="vue-date-time-picker__timezone"
-				:append-to-body="false"
-				:input-label="t('Timezone')" />
-		</template>
-	</VueDatePicker>
+	<span class="vue-date-time-picker__wrapper">
```
ShGKme: Nitpick, but it is not valid BEM structure - element cannot be a parent of its own block.

src/components/NcDateTimePicker/NcDateTimePicker.vue
susnux: But it should not be needed, at least many of this styles are already configured by overwriting the variable of the library.
For example the `font-size` is already configured by `--dp-font-size` and should not be needed to again configure for all this classes here.

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -556,67 +558,117 @@ const ariaLabels = computed(() => ({
...
+
+.vue-date-time-picker__wrapper {
+	// This is under :root in @vuepic/vue-datepicker/dist/main.css, so importing it scoped won't work
+	--dp-common-transition: all 0.1s ease-in;
+	--dp-menu-padding: 6px 8px;
+	--dp-animation-duration: 0.1s;
+	--dp-menu-appear-transition-timing: cubic-bezier(.4, 0, 1, 1);
+	--dp-transition-timing: ease-out;
+	--dp-action-row-transtion: all 0.2s ease-in;
+	--dp-font-family: -apple-system, blinkmacsystemfont, "Segoe UI", roboto, oxygen, ubuntu, cantarell, "Open Sans", "Helvetica Neue", sans-serif;
```
susnux: we should probably use our styles here

src/components/NcDateTimePicker/NcDateTimePicker.vue
raimund-schluessler: Well, if it's not done, then the style is off. But I am fine with any other solution you have.
I just want this issue to be solved and you see my solution here. Please, feel free to adjust it to your liking.

ShGKme (approved):

src/components/NcDateTimePicker/NcDateTimePicker.vue
susnux: We do this also in other places :sweat_smile:

src/components/NcDateTimePicker/NcDateTimePicker.vue
susnux: done

susnux:
1. rebased
2. squashed some commits
3. migrate span to div

Antreesy (approved):
Tested. Not sure if related to this PR, but some format options are no longer valid

PR #7058, opened 2025-06-18
refactor: deprecate events not comply with Vue event naming rules

susnux:
### ☑️ Resolves

Deprecate events which not comply with camelCase naming.
Only those two places have an invalid naming as normally only the `update:MODEL` event is scoped and other events just use camel case names like `selectSomething` (aka `select-something`) instead of `select:something`.

### 🏁 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -317,6 +317,16 @@ The `richEditing` mixin can be replaced by just using the `NcRichText` component
 The value `arrowRight` of the property `trailingButtonIcon` is deprecated and will be removed in a future version.
 It is replaced by `arrowEnd` which reflects that the directions depends on the text directions (LTR vs RTL).
 
+#### Event names
+Custom events now have a consistent naming without custom scoping of the events.
+Thus following events are deprecated in favor of a new consistent event name:
+
+     Component |       Old event |       New event
+---------------|-----------------|----------------
+`NcAppContent` |   `resize:list` |   `resize-list`
+  `NcRichText` | `interact:todo` | `interact-todo`
+
+
 ## [v8.27.0](https://github.com/nextcloud-libraries/nextcloud-vue/tree/v8.27.0) (2025-05-28)
 
 ### 📝 Notes
--- a/src/components/NcRichText/NcRichText.vue
+++ b/src/components/NcRichText/NcRichText.vue
@@ -77,7 +77,7 @@ It is also possible to make a rendered content interactive and listen for events
 		<NcRichText :text="text"
 			:use-extended-markdown="true"
 			:interactive="true"
-			@interact:todo="handleInteraction"/>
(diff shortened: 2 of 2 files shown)


julien-nc (approved):
Nice of you to keep the deprecated ones for a while :+1:

CHANGELOG.md
```diff
@@ -317,6 +317,16 @@ The `richEditing` mixin can be replaced by just using the `NcRichText` component
 The value `arrowRight` of the property `trailingButtonIcon` is deprecated and will be removed in a future version.
 It is replaced by `arrowEnd` which reflects that the directions depends on the text directions (LTR vs RTL).
 
+#### Event names
+With Vue 3 it is recommended to use camel case event names and only scope events for models like `update:modelValue` or `update:shown`.
```
ShGKme: Same as: https://github.com/nextcloud-libraries/nextcloud-vue/pull/7059#discussion_r2154485389

susnux: You might be right and I misinterpreted the docs.
But still we only have this two places, `resize:list` also makes not much sense it should rather be consistent with all other events we have (`resize-list`).

For `interact:todo` it might make sense if there are any other events for interaction, but its the only one so not sure.

ShGKme: > You might be right and I misinterpreted the docs.

This might also be from the `eslint-plugin-vue` member's comments, where `foo:barBaz` is not considered `camelCase`, and `:` is considered a special case for `update:*` events (that are "never used without `v-model`/`.sync` directive").

IMO, this is only a personal opinion, and Vue documentation never says it directly.

I saw namespaced event names in many apps and good libs, which is a great thing for complex components.

There was added an additional option for ESLint rule for this case:
https://github.com/vuejs/eslint-plugin-vue/pull/1321

ShGKme: > But still we only have this two places, `resize:list` also makes not much sense it should rather be consistent with all other events we have (`resize-list`).

Agree as `resize:` is not a namespace.

> For `interact:todo` it might make sense if there are any other events for interaction, but its the only one so not sure.

Yes, but future-proof.

ShGKme: So I'm fine with renaming in general.

But disagree with "With Vue 3 it is recommended ... only scope events for models" part

DorraJaouad: > For interact:todo it might make sense if there are any other events for interaction

I am fine with renaming it bc now it conveys it as big feature whereas it's just one simple interaction and we don't have any plans to expand interaction.

susnux: Agree! Adjusted the changelog.

About the scoping: I do not have a problem with `interact:todo` but it seems its the only one where this is needed so we could also just use the plain event. But if you like then I revert that @ShGKme and add it as an exception to the eslint config.

PR #7066, opened 2025-06-19
fix(NcDateTimePicker): do not scope all imported styles

susnux:
* alternative for https://github.com/nextcloud-libraries/nextcloud-vue/pull/7051

### DIFF
--- a/src/components/NcDateTimePicker/NcDateTimePicker.vue
+++ b/src/components/NcDateTimePicker/NcDateTimePicker.vue
@@ -607,11 +607,17 @@ const ariaLabels = computed(() => ({
 </template>
 
 <style scoped lang="scss">
-@import '@vuepic/vue-datepicker/dist/main.css';
+@use "sass:meta";
 
 .vue-date-time-picker {
 	--dp-border-radius: var(--border-radius-element);
 	--dp-input-icon-padding: var(--default-clickable-area);
+
+	// we need to import the vuepic styles but at least scoped to our class and scope
+	// plain @import does not work as this will scope all styles imported.
+	:deep() {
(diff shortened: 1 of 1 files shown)


src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -607,12 +607,18 @@ const ariaLabels = computed(() => ({
...
 .vue-date-time-picker {
 	--dp-border-radius: var(--border-radius-element);
 	--dp-input-icon-padding: var(--default-clickable-area);
 
+	// we need to import the vuepic styles but at least scoped to our class and scope
+	// plain @import does not work as this will scope all styles imported.
+	:deep() {
+		@include meta.load-css('@vuepic/vue-datepicker/dist/main.css');
+	}
+
```
ShGKme, line 621: Didn't test yet, but I don't understand how it might work in theory...
`.vue-date-time-picker` is rendered in place, while the picker itself (`dp__menu` dialog) is in `body` and not in `.vue-date-time-picker`.

ShGKme: And it doesn't apply styles on the `.vue-date-time-picker` element itself, does it?

ShGKme: Does it work for you?

Seems broken to me even without a teleport now (with fixed import for Webpack, maybe I fixed it wrong)
image

susnux: You are right, it would only be possible by adding a wrapper around.
For the body mounted version one could add an event listener on `open` and so something like:
`pickerRef.value!.getDpWrapMenuRef().value.classList.add('custom-scope')`

PR #7067, opened 2025-06-19
feat: add `useFormatRelativeTime` composable

susnux:
### ☑️ Resolves

* alternative to #6543 

This adds new composables as discussed in the PR mentioned above using the `@nextcloud/l10n` method.

### 🚧 Tasks

- Write documentation if accepted

### 🏁 Checklist

### DIFF
--- a/src/composables/useFormatDateTime.ts
+++ b/src/composables/useFormatDateTime.ts
@@ -3,18 +3,42 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-import type { MaybeRef } from 'vue'
-import { getCanonicalLocale, getLanguage } from '@nextcloud/l10n'
-import { computed, onUnmounted, ref, toValue, watchEffect } from 'vue'
+import type { FormatDateOptions } from '@nextcloud/l10n'
+import type { MaybeRefOrGetter, Ref } from 'vue'
+
+import { formatRelativeTime, getCanonicalLocale } from '@nextcloud/l10n'
+import { computed, onUnmounted, readonly, ref, toValue, watchEffect } from 'vue'
 import { t } from '../l10n.js'
 
-const FEW_SECONDS_AGO = {
-	long: t('a few seconds ago'),
-	short: t('seconds ago'), // FOR TRANSLATORS: Shorter version of 'a few seconds ago'
-	narrow: t('sec. ago'), // FOR TRANSLATORS: If possible in your language an even shorter version of 'a few seconds ago'
+interface FormatRelativeTimeOptions extends Partial<Omit<FormatDateOptions, 'ignoreSeconds'>> {
+	ignoreSeconds?: boolean
+
+	/**
+	 * If set to false the relative time will not be updated anymore.
+	 * @default true - Meaning the relative time will be updated if needed
+	 */
+	update?: boolean
 }
 
-interface FormatDateOptions {
+interface FormatTimeOptions {
+	/**
+	 * Locale to use for formatting.
+	 * @default current locale
+	 */
+	locale?: string
+
+	/**
+	 * The format used for displaying.
+	 *
+	 * @default { timeStyle: 'medium', dateStyle: 'short' }
+	 */
+	format?: Intl.DateTimeFormatOptions
+}
+
+/**
+ * @deprecated
+ */
+interface LegacyFormatDateTimeOptions {
 	/**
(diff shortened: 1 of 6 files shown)


tests/unit/composables/useFormatDateTime.spec.js
```diff
@@ -3,15 +3,89 @@
...
+
+	it('works with computed timestamp', () => {
+		const formatted = useFormatRelativeTime(computed(() => time))
+		expect(formatted.value).toBe('1 minute ago')
+	})
+
+	it('works with a date object', () => {
+		const formatted = useFormatRelativeTime(time)
+		expect(formatted.value).toBe('1 minute ago')
+	})
```
ShGKme: What is the difference between the 1st and the 3rd tests?

tests/unit/composables/useFormatDateTime.spec.js
```diff
@@ -3,15 +3,89 @@
...
 
-import { afterAll, beforeAll, describe, expect, it, vi } from 'vitest'
+describe('useFormatRelativeTime composable', () => {
+	const time = Date.parse('2025-01-01T00:00:00Z')
+
+	beforeAll(() => vi.useFakeTimers())
+	beforeEach(() => vi.setSystemTime(time + 60000))
+
+	it('works with timestamp', () => {
+		const formatted = useFormatRelativeTime(time)
```
susnux, line 17:
```suggestion
		const formatted = useFormatRelativeTime(new Date(time))
```

tests/unit/composables/useFormatDateTime.spec.js
susnux:
```suggestion
	it('works with timestamp', () => {
		const formatted = useFormatRelativeTime(time)
		expect(formatted.value).toBe('1 minute ago')
	})

	it('works with computed timestamp', () => {
		const formatted = useFormatRelativeTime(computed(() => time))
		expect(formatted.value).toBe('1 minute ago')
	})

	it('works with a date object', () => {
		const formatted = useFormatRelativeTime(new Date(time))
		expect(formatted.value).toBe('1 minute ago')
	})
```

tests/unit/composables/useFormatDateTime.spec.js
```diff
@@ -3,15 +3,89 @@
...
+		expect(formatted.value).toBe('1 minute ago')
+	})
+
+	it('works with a date object', () => {
+		const formatted = useFormatRelativeTime(time)
+		expect(formatted.value).toBe('1 minute ago')
+	})
+
+	it('works with computed date object', () => {
+		const formatted = useFormatRelativeTime(computed(() => time))
```
susnux:
```suggestion
		const formatted = useFormatRelativeTime(computed(() => new Date(time)))
```

src/composables/useFormatDateTime.ts
```diff
@@ -3,125 +3,173 @@
...
+
+	/**
+	 * Update the relative time string.
+	 * This is the callback for the interval.
+	 */
+	function updateRelativeTime() {
+		relativeTime.value = formatRelativeTime(date.value, options.value)
+
+		if (toValue(opts).update !== false) {
+			timeoutId = window.setTimeout(updateRelativeTime, options.value.ignoreSeconds ? 30000 : 1000)
```
ShGKme: Maybe make the interval more adaptive? E.g. if the time is far in the past already, nothing changes in hours and there is no need to check every 30 seconds.

ShGKme (approved):

src/composables/useFormatDateTime.ts
```diff
@@ -3,125 +3,177 @@
...
+interface LegacyFormatDateTimeOptions {
 	/**
 	 * The format used for displaying, or if relative time is used the format used for the title
 	 */
 	format?: Intl.DateTimeFormatOptions
 	/**
 	 * Ignore seconds when displaying the relative time and just show `a few seconds ago`
 	 */
-	ignoreSeconds?: boolean
+	ignoreSeconds?: true
```
Antreesy: Wouldn't it break, if I want it conditional?

Antreesy (approved):
Otherwise looks fine

src/composables/useFormatDateTime.ts
susnux: I am fine with that:

```suggestion
	ignoreSeconds?: boolean
```

PR #7074, opened 2025-06-23
fix(NcChip): adjust default prop declaration

Antreesy:
### ☑️ Resolves

- Fix TS errors in typed components
  - NcChip: `variant` is optional
  - NcDateTimePicker: `modelValue`, `min` and `max` are nullable

So far noticed only these
### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/NcDateTimePickerNative/NcDateTimePickerNative.vue
+++ b/src/components/NcDateTimePickerNative/NcDateTimePickerNative.vue
@@ -124,7 +124,7 @@ export default defineComponent({
 		 * Pass null to clear the input field.
 		 */
 		modelValue: {
-			type: Date,
+			type: [Date, null],
 			default: null,
 		},
 
@@ -159,7 +159,7 @@ export default defineComponent({
 		 * default type: null
 		 */
 		min: {
-			type: [Date, Boolean],
+			type: [Date, Boolean, null],
 			default: null,
 		},
 		/**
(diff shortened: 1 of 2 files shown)


src/components/NcDateTimePickerNative/NcDateTimePickerNative.vue
```diff
@@ -159,15 +159,15 @@ export default defineComponent({
 		 * default type: null
 		 */
 		min: {
-			type: [Date, Boolean],
+			type: [Date, Boolean, null],
```
susnux, line 162: We really should get rid of all that `null` values where instead `undefined` is meant

susnux (approved):

src/components/NcDateTimePickerNative/NcDateTimePickerNative.vue
ShGKme: It means we need additional mappings in apps, as serializable DTOs can only have `null`.

ShGKme (approved):

src/components/NcDateTimePickerNative/NcDateTimePickerNative.vue
susnux: Why? Undefined can be serialized as well. Simply not defined aka do not set the key on the object.

Null is a really ugly thing in JS, in 99,9% of the places you mean undefined not null.
Because you mean an undefined value and not the defined null value which means the variable has a value of null instead of the variable is unset.
(especially that null is type of object is :scream: )

PR #7077, opened 2025-06-23
fix(NcPopover): component crash when unmounted shown

Antreesy:
### ☑️ Resolves

`getPopoverContentElement()` can be undefined, so it should not fail when trying to add a listener

### 🏁 Checklist

### DIFF
--- a/src/components/NcPopover/NcPopover.vue
+++ b/src/components/NcPopover/NcPopover.vue
@@ -565,7 +565,16 @@ export default {
 			this.addEscapeStopPropagation()
 		},
 		afterHide() {
-			this.getPopoverContentElement().addEventListener('transitionend', () => {
+			/**
+			 * On component unmounting Vue:
+			 * 1. Resets the template ref to null
+			 * 2. Triggers <Dropdown> `beforeUnmount` - it emits `apply-hide` event
+			 *   2.1. At that moment this.$refs.popover is null already, and we cannot use `this.getPopoverContentElement()`
+			 *   2.2. We can ignore emitting `after-hide` event (was also not the case on stable8)
+			 * 3. Actually removes <Dropdown> node
+			 * 4. Triggers <Dropdown> `unmounted`
+			 */
+			this.getPopoverContentElement()?.addEventListener('transitionend', () => {
 				/**
 				 * Triggered after the tooltip was visually hidden.
 				 *


ShGKme:
> `getPopoverContentElement()` can be undefined

Is it expected? Why?

Antreesy:
> Is it expected? Why?

I haven't checked, but probably cause it's a ref. Didn't you add it yourself? 👀

susnux:
> so it should not fail when trying to add a listener

when is this the case? We had the same discussion when I added the optional chaining operator:
https://github.com/nextcloud-libraries/nextcloud-vue/pull/6802#discussion_r2049288253

susnux:
And from the code I do not see where this can ever be undefined in those event listeners

Antreesy:
Currently happening, when there is an error  while closing popover. Maybe it's redundant, but nevertheless - we can put it in one place and skip in another two

susnux (approved):

ShGKme:
> I haven't checked, but probably cause it's a ref. Didn't you add it yourself? 👀

When it's `undefined`, `afterShow` and `afterHide` events are not triggered.

If `undefined` is expected here, then the component triggers no `afterShow`/`afterHide` event in a specific case. Then it's interesting, what is the case. E.g., this behavior should be documented.

If `undefined` is not expected here, then something goes wrong, and we need to fix it. By adding `?.` we hide an issue caused by another issue.

ShGKme:
Could you confirm, that `undefined` is expected here and it's fine to not emit the events?
Best with a comment in the code explaining the case (it seems non-obious to me)

Antreesy:
1. Adjust NcPopover rendering logic
	1.1 Click should cause `v-if` condition to render a popover to change, and it's unmounted before apply-hide
2. See result:
image


P.S. deferring unmount with nextTick also works, but I don't think it's the best way to control it

Antreesy:
See reproduction details cc @susnux @ShGKme 

<details>
<summary>Details</summary>

#### WIP error a `<NcButton>` as a trigger:

```vue
<template>
	<div style="display: flex">
		<NcButton @click="mountPopover = !mountPopover">Mount-unmount</NcButton>
		<NcPopover v-if="mountPopover" popup-role="dialog">
			<template #trigger>
				<NcButton>I am the trigger</NcButton>
			</template>
			<template #default>
				<form tabindex="0" role="dialog" aria-labelledby="popover-example-dialog-header-1" @submit.prevent>
					<h2 id="popover-example-dialog-header-1">this is some content</h2>
					<p>
						Lorem ipsum dolor sit amet, consectetur adipiscing elit. <br/>
						Vestibulum eget placerat velit.
					</p>
					<label>
						Label element
						<input type="text" placeholder="input element"/>
					</label>
				</form>
				<NcButton @click="mountPopover = !mountPopover">Mount-unmount</NcButton>
			</template>
		</NcPopover>
	</div>
</template>

<script>
	export default {
		data() {
			return {
				mountPopover: false,
			}
		},
	}
</script>
```

</details>

src/components/NcPopover/NcPopover.vue
```diff
@@ -565,7 +565,11 @@ export default {
 			this.addEscapeStopPropagation()
 		},
 		afterHide() {
-			this.getPopoverContentElement().addEventListener('transitionend', () => {
+			/**
+			 * Component might be unmounted before 'apply-hide' is emitted.
+			 * Optional chaining is added to prevent errors in this case.
+			 */
+			this.getPopoverContentElement()?.addEventListener('transitionend', () => {
```
ShGKme: > Component might be unmounted before 'apply-hide' is emitted 

This keeps it undetermined in which case it might happen and why it is expected and correct behavior.

And this is not possible. At the moment`<Dropdown>` emits `apply-hide` event, it is not unmounted yet. Component instance is triggering the event. After unmounting, there is no instance to trigger the event.

ShGKme: Ignoring with `?.` here also means `<NcPopover>` doesn't trigger `after-hide` on unmounting

ShGKme: If we don't expect `<NcPopover>` to trigger `after-hide` on unmounting:

```suggestion
			// On component unmounting Vue:
			// 1. Resets the template ref to null
			// 2. Triggers child's `beforeUnmount`
			//    - On this moment <Dropdown> instance still exists and emits the `afterHide` event
			// 3. Actually removes the child node
			// 4. Triggers child's `unmounted`
			// 
			// Thus when the last `apply-hide` is triggered during unmounting, the template ref is null already (even the child <Dropdown> is still mounted), and we cannot getPopoverContentElement
			// 
			// In this case we don't need to trigger `after-hide` and can ignore it
			this.getPopoverContentElement()?.addEventListener('transitionend', () => {
```

ShGKme: If we do expect `<NcPopover>` to trigger `after-hide` on unmounting:

```suggestion
			// On component unmounting Vue:
			// 1. Resets the template ref to null
			// 2. Triggers child's `beforeUnmount`
			//    - On this moment <Dropdown> instance still exists and emits the `afterHide` event
			// 3. Actually removes the child node
			// 4. Triggers child's `unmounted`
			// 
			// Thus when the last `apply-hide` is triggered during unmounting, the template ref is null already (even the child <Dropdown> is still mounted), and we cannot getPopoverContentElement
			
			/* handle it some other way */
```

ShGKme: In `stable8` we didn't trigget `after-hide` on unmounting, so lest ignore it in next as well

ShGKme: Update: in `stable8` there is no `after-hide` on unmounting, because `transitionend` with the event triggering happens after the component is unmounted already, so emit does nothing.

ShGKme: So let's ignore it when undefined, and if we will want the even later, let's add it in both versions.

ShGKme:
For the history: Vue 3 sets template ref to `null` before starting the unmounting, including `beforeUnmount` event.

https://github.com/vuejs/core/blob/v3.5.17/packages/runtime-core/src/renderer.ts#L2110

PR #7088, opened 2025-06-24
chore(NcDialog): export prop type to use in apps

Antreesy:
### ☑️ Resolves

- Allows to import types from vue-library without extra efforts
- E.g. can be used for typing a NcDialog wrapper in `spawnDialog` function

image

### 🚧 Tasks

- [ ] More generic way to do it for all components?

### 🏁 Checklist

### DIFF
--- a/src/components/NcDialog/NcDialog.vue
+++ b/src/components/NcDialog/NcDialog.vue
@@ -215,24 +215,22 @@ import type { Slot } from 'vue'
 import type { ComponentProps, VueClassType } from '../../utils/VueTypes.ts'
 
 import { useElementSize } from '@vueuse/core'
-import { computed, ref, useTemplateRef } from 'vue'
+import { computed, ref, useModel, useTemplateRef } from 'vue'
 import { createElementId } from '../../utils/createElementId.ts'
 import NcDialogButton from '../NcDialogButton/index.ts'
 import NcModal from '../NcModal/index.js'
 
 type NcDialogButtonProps = ComponentProps<typeof NcDialogButton>
 
-/**
- * Whether the dialog should be shown
- */
-const open = defineModel<boolean>('open', { default: true })
-
-const props = withDefaults(defineProps<{
+export type NcDialogProps = {
 	/** Name of the dialog (the heading) */
 	name: string
 
 	/** Text of the dialog */
 	message?: string
+
+	/** Whether the dialog should be shown */
+	open?: boolean
 
 	/** Additional elements to add to the focus trap */
 	additionalTrapElements?: Array<string | HTMLElement>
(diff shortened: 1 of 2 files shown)


src/components/NcDialog/NcDialog.vue
```diff
@@ -223,7 +223,7 @@ import { createElementId } from '../../utils/createElementId.ts'
 
 type NcDialogButtonProps = ComponentProps<typeof NcDialogButton>
 
-const props = withDefaults(defineProps<{
+export type NcDialogProps = {
```
ShGKme, line 225: When defining the props type for export, you cannot use `defineModel`. `open` is missing in the type.

Instead, you need to define `open` and `update:open` explicitly, and then replace `defineModel` compile macros with `useModel` vue function.

src/components/NcDialog/NcDialog.vue
```diff
@@ -328,9 +334,15 @@ const props = withDefaults(defineProps<{
 /**
  * Whether the dialog should be shown
  */
-const open = defineModel<boolean>('open', { default: true })
+const open = useModel(props, 'open')
 
 const emit = defineEmits<{
```
ShGKme: Might make sense to also export `NcDialogEmits` type

ShGKme (approved):

src/components/NcDialog/NcDialog.vue
```diff
@@ -328,9 +334,15 @@ const props = withDefaults(defineProps<{
 /**
  * Whether the dialog should be shown
  */
-const open = defineModel<boolean>('open', { default: true })
+const open = useModel(props, 'open')
 
-const emit = defineEmits<{
+export type NcDialogEmits = {
```
susnux, line 320: why is this needed? Also because its not exported from the `index`

Antreesy: I forgot 🙈 

May be needed to implement similar to Vue 2 `v-on="$listeners"` ? But I personally don't see the case, for manually picked events we can do it without NcDialogEmits

ShGKme: > why is this needed?

Because it is also a part of the component interface, and instance params are props + events.

For example, if you want to make a transparent wrapper over `NcDialog`, you need all 3 types:
1. props
2. emits
3. slots

Otherwise you wrapper has no typing.

src/components/NcDialog/index.ts
```diff
@@ -3,4 +3,6 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
+export type { NcDialogProps } from './NcDialog.vue'
```
susnux: In this case please also export here:

```suggestion
export type { NcDialogEmits, NcDialogProps } from './NcDialog.vue'
```

susnux:
What is the exact use case? Because for in-library usage this makes sense but limit us here to only use `useModel` vs `defineModel` just to be able to export types seems to be wrong.
In general you can also always just do:

```ts
type ComponentProps<T> = InstanceType<T>["$props"]

type NcDialogProps = ComponentProps<typeof NcDialog>
```

Antreesy:
I think it's there I was starting
image

Antreesy:
> ```ts
> type ComponentProps<T> = InstanceType<T>["$props"]
> 
> type NcDialogProps = ComponentProps<typeof NcDialog>
> ```

What if we do it in index.ts?
image

App seems to interpret it correctly

susnux:
> App seems to interpret it correctly

Yes I use this pattern in other apps as the types are already exported the library - just not as plain types but as the instance type :see_no_evil:

susnux:
That is why I asked, because normally you want to have the types of *your* dialog not of `NcDialog`.

Antreesy:
Then I guess I'll just copy-paste VueTypes file to the project for this use-case. Feels rather specific after discussion

PR #7090, opened 2025-06-24
fix(NcRichContentEditable): wait for input to focus so it can show tribute correctly

DorraJaouad:
### 🖼️ Screenshots

🏚️ Before

https://github.com/user-attachments/assets/84698410-8df3-43c2-b2f0-09faf8747eb2

🏡 After

https://github.com/user-attachments/assets/e7726744-5462-49a6-a128-c574833006ad

### 🏁 Checklist

### DIFF
--- a/src/components/NcRichContenteditable/NcRichContenteditable.vue
+++ b/src/components/NcRichContenteditable/NcRichContenteditable.vue
@@ -950,9 +950,14 @@ export default {
 		 */
 		showTribute(trigger) {
 			this.focus()
-			const index = this.tribute.collection.findIndex(collection => collection.trigger === trigger)
-			this.tribute.showMenuForCollection(this.$refs.contenteditable, index)
-			this.updateValue(this.$refs.contenteditable.innerHTML)
+			// For an unknown reason in Vue 3 calling `tribute.showMenuForCollection` immediately after `contenteditable.focus()` 
+			// results in the wrong positioning.
+			// Waiting one animation frame for focusing fixes the issue.
+			requestAnimationFrame(() => {
+				const index = this.tribute.collection.findIndex(collection => collection.trigger === trigger)
+				this.tribute.showMenuForCollection(this.$refs.contenteditable, index)
+				this.updateValue(this.$refs.contenteditable.innerHTML)
+			})
 			document.addEventListener('click', this.hideTribute, true)
 		},
 


src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -549,8 +549,8 @@ export default {
 		 *
 		 * @public
 		 */
-		focus() {
-			this.$refs.contenteditable.focus()
+		async focus() {
+			await this.$refs.contenteditable.focus()
```
ShGKme: `focus()` does not return Promise.

`await` here may still work, it is allowed in JS to await any value. 

But it works because any `await` adds a microtask, so two redundant `await`s adds to microtasks here and create waiting.

Instead, let explicitly either wait for `Promise.resolve()` (1 microtask) or use `raf`.

src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -948,8 +948,8 @@ export default {
 		 *
 		 * @public
 		 */
-		showTribute(trigger) {
-			this.focus()
+		async showTribute(trigger) {
+			await this.focus()
```
susnux: Why not just

```suggestion
		async showTribute(trigger) {
			this.focus()
			await nextTick()
```

ShGKme: When we tested, nextTick was not enough.

For me worked with at least RAF

ShGKme: Also, we don't understand why it is broken in Vue 3 and how it is related to Vue 3...

During debugging it works fine, even with going step by step, and is only reproducible without a debugger.

DorraJaouad: It's still not enough to wait for focus, it seems it takes longer then DOM update

src/components/NcRichContenteditable/NcRichContenteditable.vue
DorraJaouad: 1 microtask is not enough, either two or `raf` -> so going with the latter

src/components/NcRichContenteditable/NcRichContenteditable.vue
```diff
@@ -950,9 +950,13 @@ export default {
 		 */
 		showTribute(trigger) {
 			this.focus()
-			const index = this.tribute.collection.findIndex(collection => collection.trigger === trigger)
-			this.tribute.showMenuForCollection(this.$refs.contenteditable, index)
-			this.updateValue(this.$refs.contenteditable.innerHTML)
+			// Focus may be delayed, so we wait for the next frame to show the menu.
+			// Tribute requires caret position (this focus) to set its position
```
ShGKme: Maybe something like

```suggestion
			// For an unknown reason in Vue 3 calling `tribute.showMenuForCollection` immediately after `contenteditable.focus()` results in the wrong positioning.
			// Waiting one animation frame for focusing fixes the issue.
```

ShGKme (approved):

src/components/NcRichContenteditable/NcRichContenteditable.vue
ShGKme: Or we need to debug tribute 👀

susnux (approved):

ShGKme:
Checked one more time. The problem is the `focus-trap`. It doesn't allow leaving the `NcActions` menu

ShGKme:
The root of the issue is in broken `NcActions`

PR #7097, opened 2025-06-26
feat(NcFilePicker): add picker component to select local files

susnux:
- related to https://github.com/nextcloud-libraries/nextcloud-upload/issues/1809

### ☑️ Resolves

This can be used e.g. to upload files, for example in the forms app but also for the files or photos app.

### 🏁 Checklist


src/components/NcFilePicker/NcFilePicker.vue
```diff
@@ -0,0 +1,236 @@
...
+
+			<!-- App defined upload actions -->
+			<slot name="actions" />
+		</NcActions>
+
+		<!-- Hidden files picker input -->
+		<input ref="input"
+			:accept="accept?.join(', ')"
+			:multiple="multiple"
+			class="hidden-visually"
```
ShGKme, line 258: It should have a different class, to be visually hidden, but stays in the same position, like with the hidden input in `NcCheckboxRadioSwitch`

src/components/NcFilePicker/NcFilePicker.vue
```diff
@@ -0,0 +1,236 @@
...
+			</NcActionButton>
+
+			<!-- App defined upload actions -->
+			<slot name="actions" />
+		</NcActions>
+
+		<!-- Hidden files picker input -->
+		<input ref="input"
+			:accept="accept?.join(', ')"
+			:multiple="multiple"
```
ShGKme: :P
```suggestion
			:multiple
```

src/components/NcFilePicker/NcFilePicker.vue
```diff
@@ -0,0 +1,236 @@
...
+}
+
+/**
+ * Handle picking some local files
+ */
+function onPick() {
+	const files = inputElement.value?.files ? Array.from(inputElement.value.files) : []
+	console.error('ON PICK', files)
+	emit('pick', files)
+}
```
ShGKme, line 183: Didn't test yet, but we probably need to clean the `input` after we got the files (or in `triggerPickFiles`). Otherwise it is not possible to select the same files, as it doesn't trigger the `change` event

src/components/NcFilePicker/NcFilePicker.vue
```diff
@@ -0,0 +1,236 @@
...
+				close-after-click
+				@click="triggerPickFiles(true)">
+				<template #icon>
+					<IconFolderUpload style="color: var(--color-primary-element)" :size="20" />
+				</template>
+				{{ t('Upload folders') }}
+			</NcActionButton>
+
+			<!-- App defined upload actions -->
+			<slot name="actions" />
```
ShGKme: Maybe add an option to include default `'apps/files/api/v1/templates'` actions?

src/components/NcFilePicker/NcFilePicker.vue
susnux: why does it needs to stay in that position? Was not a problem in files and photos

src/components/NcFilePicker/NcFilePicker.vue
susnux: Where would this be used?
Files has its own way to register "new"-menu entries and it is not really related to an local-files picker?

src/components/NcFilePicker/NcFilePicker.vue
susnux: you are right, fixed.

src/components/NcFilePicker/NcFilePicker.vue
ShGKme: Everywhere where Filepicker is used to pick a file that is to share, and where it can also be a new file.

Very similar: Files, Talk

image

A bit similar (with a way to select from files, and files templates seem to be missing)

Notes (text):

image

Calendar:

image

src/components/NcFilePicker/NcFilePicker.vue
ShGKme: For the same reason as in `<NcCheckboxRadioSwitch>` - assistive technologies may focus on an actual `input`. Unlike visually hidden text, it is still a part of the tab sequence. And then, if the input is moved away, focusing it moves scrolling containers and users with assistive technologies far away.

But now, while writing it, I think it makes less sense for the file input because it is never used directly (like a checkbox does).

src/components/NcFilePicker/NcFilePicker.vue
ShGKme: > it is not really related to an local-files picker?

As files are stored in Nextcloud, it often makes sense to also be able to share a file from Nextcloud apart from uploading a new one.

For example, currently 4 different apps have it but name it differently.

ShGKme: So it makes sense to me to have a shared component for it.

susnux: > So it makes sense to me to have a shared component for it.

Yes but in this case I wanted to first have a file input component, which we could reuse for that.
Because e.g. my use case in forms does not need this. It only needs to pick files from local computer.


> Maybe add an option to include default 'apps/files/api/v1/templates' actions?

But here my question was more like that this is apps specific as the files UI does not use this API but instead uses the actions registered on the frontend:
https://nextcloud-libraries.github.io/nextcloud-files/functions/index.getNewFileMenuEntries.html

src/components/NcFilePicker/NcFilePicker.vue
```diff
@@ -0,0 +1,238 @@
...
+
+const formElement = useTemplateRef('form')
+const inputElement = useTemplateRef('input')
+
+/**
+ * Check whether the current browser supports uploading directories
+ * Hint: This does not check if the current connection supports this, as some browsers require a secure context!
+ */
+const canUploadFolders = computed(() => {
+	return props.allowFolders && 'webkitdirectory' in document.createElement('input')
```
ShGKme: To avoid DOM API calls in a `computed`

```suggestion
	return props.allowFolders && 'webkitdirectory' in HTMLInputElement.prototype
```

src/components/NcFilePicker/NcFilePicker.vue
```diff
@@ -0,0 +1,238 @@
...
+	// reset the form
+	reset()
+
+	// Setup directory picking if enabled
+	if (canUploadFolders.value) {
+		inputElement.value!.webkitdirectory = uploadFolders
+	}
+	// Trigger click on the input to open the file picker
+	nextTick(() => inputElement.value!.click())
+}
```
ShGKme, line 175: Nitpick:

All the comments here just duiplicate what the code says (resets resets, click clicks), but doesn't explain the purpose (e.g. why mutate `webkitdirectory` and not bind in the template, why wait for `nextTick`).

```suggestion
function triggerPickFiles(uploadFolders: boolean) {
	// Without reset selecting the same file doesn't trigger the change event
	reset()

	// Vue does not support setting "webkitdirectory" via template <why>
	if (canUploadFolders.value) {
		inputElement.value!.webkitdirectory = uploadFolders
	}

	// Wait for <reason to wait>
	nextTick(() => inputElement.value!.click())
}
```

src/components/NcFilePicker/NcFilePicker.vue
```diff
@@ -0,0 +1,259 @@
...
+/**
+ * Trigger file picker
+ *
+ * @param uploadFolders - Whether to upload folders or files
+ */
+function triggerPickFiles(uploadFolders: boolean) {
+	// Without reset selecting the same file doesn't trigger the change event
+	reset()
+
+	// Vue does not support setting "webkitdirectory" via template <why>
```
ShGKme: `<why>` 👀

src/components/NcFilePicker/NcFilePicker.vue
```diff
@@ -0,0 +1,259 @@
...
+	if (canUploadFolders.value) {
+		inputElement.value!.webkitdirectory = uploadFolders
+	}
+
+	// Wait for <reason to wait>
```
ShGKme: `<reason to wait>` 👀

susnux: You are too fast ;)
Reason was to wait for the DOM to be updated after the `<why>`

src/components/NcFilePicker/NcFilePicker.vue
susnux: because it would always trigger picking a directory even if "pick files" was selected.

src/components/NcFilePicker/NcFilePicker.vue
ShGKme: > You are too fast ;)

You force pushed and requested my review :P

> Reason was to wait for the DOM to be updated after the <why>

But you updated DOM via direct DOM manipulation, not via Vue rendering...

susnux: Well two things its the reset and the DOM - updated the comment

ShGKme:
@susnux is there anything you want to change here? I think you mentioned it at a meeting.

susnux:
> is there anything you want to change here? I think you mentioned it at a meeting.

Yes will work on it

susnux:
Just to clarify: This is **NOT** the uploader.
The idea here is that this is just a plain file picker (styled and enhanced `input[type=file]`.
The use case are apps which have their own way of handling uploads - like forms - or if you directly process files in the browser.

There will be a follow up with `NcUploader` or `NcUploadPicker` (?) which will be the migrated component from `@nextcloud/upload` and will be based on this component.

Current state of this PR is: I will add proper tests for this new component.
As soon as that is done I will set the state to "3. to review" and mark it as "ready for review".

src/components/NcFilePicker/NcFilePicker.vue
```diff
@@ -0,0 +1,288 @@
...
+				@click="triggerPickFiles(true)">
+				<template #icon>
+					<IconFolderUpload style="color: var(--color-primary-element)" :size="20" />
+				</template>
+				<!-- If this is not the only action in the NcActions this is a menu entry and we need a generic name - otherwise this will be a single button where we need to apply the label -->
+				{{ !directoryOnly || $slots.actions ? t('Upload folder') : currentLabel }}
+			</NcActionButton>
+
+			<!-- App defined upload actions -->
+			<slot name="actions" />
```
ShGKme: How much custom are the actions?

Would it be possible to replace the `#actions` slot with some props with menu lists:
- createNewOptions
- additionalOptions

where every option is a description:
- icon
- label
- callback

To avoid coupling on `NcActions`.

susnux: See #7882 

Do you mean like I did there?

ShGKme: Yes

susnux: @ShGKme I have adjusted the PR accordingly (as a new commit for easier review of changes).
Does this match what you had in mind?

src/components/NcFilePicker/NcFilePicker.vue
```diff
@@ -0,0 +1,331 @@
...
+import IconPlus from 'vue-material-design-icons/Plus.vue'
+import IconUpload from 'vue-material-design-icons/Upload.vue'
+import NcActionButton from '../NcActionButton/NcActionButton.vue'
+import NcActionCaption from '../NcActionCaption/NcActionCaption.vue'
+import NcActions from '../NcActions/NcActions.vue'
+import NcIconSvgWrapper from '../NcIconSvgWrapper/NcIconSvgWrapper.vue'
+import NcLoadingIcon from '../NcLoadingIcon/NcLoadingIcon.vue'
+import { t } from '../../l10n.js'
+
+export interface MenuAction {
```
ShGKme: Since it is a public API, can we use less generic name? Something like, `FilePickerMenuAction` or `FilePickerItem` to not bind to `NcActions`.

susnux: done, please re-review :)

ShGKme:
> [ ] 2️⃣ Backport to `stable8` for maintained Vue 2 version or not applicable

No backport?

susnux:
> No backport?

Of course, but did not check it yet as I did not request it.

PR #7100, opened 2025-06-27
fix(NcButton): render <a> with "to" and prevent wrong context crash

ShGKme:
## ☑️ Resolves

- Fix some regressions from https://github.com/nextcloud-libraries/nextcloud-vue/pull/6033
- Use a different way to work with `RouterLink`

### Problem 1: crash with `router = undefined` inside `useLink()` during `computed`

Previous approach uses `useLink` composable outside of the setup context:
- composables should not be used outside `setup`
- Everything inside `useLink` is a dependency of the `computed`
- Triggering it at the wrong moment results in computing without router in the context
  - Example: prod lib build with dev build of the lib and vise versa

Instead: use `<RouterLink>` component directly

### Problem 2: `<NcButton to="/route">` is rendered as a `<button>` instead of a link `<a>`

- Kind of a breaking change

Instead: render a router link as a link

## 🏁 Checklist


susnux:
I think this is even easier to be fixed:
There are two options in setup context:
- either `router` is already available
- or it is not

So we do not need it to be computed but just check if a router is available and use `useLink` or if not we do it like we did with the computed but we just move everything in the setup context directly.

Why should this work? Because the router needs to be already available when the active Vue is created (`createApp`) adding the router later will not work or at least it even did not work in the previous implementation.

ShGKme:
> I think this is even easier to be fixed

Yes, if we only fix 1/2 issues.

susnux:
> Yes, if we only fix 1/2 issues.

Well for the `<a>` all that is needed is to adjust the computed to also check `routerLinkProps.href` (`!!props.href || !!routerLinkProps?.href`)

raimund-schluessler:
Can I help to get this fixed? It breaks the `NcBreadcrumbs` and needs to be fixed before a final release of `9.0.0` for sure.

ShGKme:
> It breaks the `NcBreadcrumbs`

Likely `NcBreadcrumbs` issue is unrelated

src/components/NcButton/NcButton.vue
```diff
@@ -603,106 +602,74 @@ defineSlots<{
...
-// And if the app uses a router we need to make sure a `to` prop was passed to use to router
-const routerLink = computed(() => (
-	inject(routerKey, null) !== null && props.to
-		? useLink({ to: toRef(() => props.to!) })
-		: undefined
-))
+const isRouterLink = computed(() => !!props.to)
+const isNativeLink = computed(() => !!props.href)
+const isLink = computed(() => isNativeLink.value || isRouterLink.value)
+const isNativeButton = computed(() => !isLink.value)
```
susnux: Why removing all the comments? Yes descriptive names are good but having a comment if you are curious does not harm and helps understanding other peoples way of thinking.
Also especially doc block comments help to get the context again if you hover that variable somewhere used.

susnux: Seems to be useless to have a computed thats just the negated of a already existing - you can always just use `!isLink` or by the usage just revert the tertiary operator below.

src/components/NcButton/NcButton.vue
```diff
@@ -603,106 +602,74 @@ defineSlots<{
...
 		return 'secondary'
 	}
+
 	return props.variant
 })
 
-/**
- * The flexbox alignment of the button content
- */
 const flexAlignment = computed(() => props.alignment.split('-')[0])
```
susnux, line 636: e.g. what harm does the additional context that flex refers to `flex box` of the button content. (e.g. the context button context now is lost and you have to lookup the template).

src/components/NcButton/NcButton.vue
```diff
@@ -603,106 +602,74 @@ defineSlots<{
...
-
-/**
- * If the button content should be reversed (icon on the end)
- */
 const isReverseAligned = computed(() => props.alignment.includes('-'))
```
susnux, line 637: same here context what reverse aligned does is lost.

src/components/NcButton/NcButton.vue
```diff
@@ -603,106 +602,74 @@ defineSlots<{
...
-		v-bind="{
-			...ncPopoverTriggerAttrs,
-			...(isLink ? linkAttrs : buttonAttrs),
-		}"
+
+		:to
+		:active-class="isRouterLink ? 'active' : undefined"
+
+		:href="isNativeLink ? (href || '#') : undefined"
+		:target
```
susnux: breaks if not a link but target set (HTML validity)

src/components/NcButton/NcButton.vue
```diff
@@ -603,106 +602,74 @@ defineSlots<{
...
+		:download
```
susnux: breaks if not a link but target set (HTML validity)

src/components/NcButton/NcButton.vue
```diff
@@ -603,106 +602,74 @@ defineSlots<{
...
+		:rel="isNativeLink && 'nofollow noreferrer noopener'"
+
+		:aria-haspopup="ncPopoverTriggerAttrs['aria-haspopup']"
+		:aria-expanded="ncPopoverTriggerAttrs['aria-expanded']"
```
susnux: sounds like knowing internals of differently injected value.
In this case the injection type should be adjusted to exactly this type as otherwise it looks like you throwing away possible other popover trigger attributes.
(Meaning either it only injects those two, then those should be static as interface, or it injects variable others  - than it should be done as `v-bind`).

src/components/NcButton/NcButton.vue
```diff
@@ -603,106 +602,74 @@ defineSlots<{
...
 			},
 		]"
-		:disabled
 		:aria-label
-		v-bind="{
-			...ncPopoverTriggerAttrs,
-			...(isLink ? linkAttrs : buttonAttrs),
-		}"
+
+		:to
```
susnux: breaks as soon as this is a native link (HTML validity)

src/components/NcButton/NcButton.vue
```diff
@@ -603,106 +602,74 @@ defineSlots<{
...
 }>()
 
-// Make sure the component also works if the app does not use any router
-// And if the app uses a router we need to make sure a `to` prop was passed to use to router
-const routerLink = computed(() => (
-	inject(routerKey, null) !== null && props.to
-		? useLink({ to: toRef(() => props.to!) })
-		: undefined
-))
+const isRouterLink = computed(() => !!props.to)
```
susnux: Should this not checking if a router is available? Because otherwise the template hard fails without an router (see the unit tests I added in the other PR)

susnux (commented):
Tested this a bit: Another problem with this solution is that the resulting `a` element for a router link **does not** have the proper `href` attribute set.

src/components/NcButton/NcButton.vue
ShGKme: Yes, if a developer sets an invalid set of attributes like
```html
<NcButton to="/route" target="_blank" />
```

But this is initially invalid. We can add a check here, but should we?
Without `inheritAttrs: false` and explicitly defining every allowed attribute, a developer can also use the button like this, and we don't prevent it.

```html
<NcButton to="/route" methods="POST" min="5" wrong="attr" />
```

src/components/NcButton/NcButton.vue
ShGKme: Resolving conversation to have it in a single place

src/components/NcButton/NcButton.vue
ShGKme: It is knowing internals, but:
- This component is from the same codebase, not an external independent module
- Changing popover attributes is very unlikely (as they are defined by HTML requirements currently). As per the comment note below in the code, for one of the most widely used components, I'd prefer to have Vue 3 rendering optimizations (patch flags) over being safe from an unlikely change. The case we may prevent here: we need a new HTML attribute in popover, and when adding it, we didn't check how it is used in the same library.

src/components/NcButton/NcButton.vue
ShGKme: Resolving conversation to have it in a single place

src/components/NcButton/NcButton.vue
ShGKme: I am not sure.

If we don't check, it results in `<routerlink>` in HTML, which is not valid.
But only if a developer uses `<NcButton to="/route">` in a context without `vue-router` which seems unlikely because:
- This doesn't work as a link - likely to be caught on testing
- It seems unlikely that a developer will set route prop without router

We can prevent it with a simple check, but it results in adding `vue-router` in the bundle even if an app doesn't need router.

ShGKme: Though we may rely on the tree-shaking and believe, it will only bundle the injection key and nothing else from the router.

ShGKme:
> Tested this a bit: Another problem with this solution is that the resulting `a` element for a router link **does not** have the proper `href` attribute set.

Indeed, `:href="undefined"` overrides Child component root's `href` .·´¯`(>▂<)´¯`·.

src/components/NcButton/NcButton.vue
ShGKme: Pushed a check, will add tests later

ShGKme:
> Indeed, `:href="undefined"` overrides Child component root's `href` .·´¯(>▂<)¯`·.

I'll double-check, but likely this is not possible to have both performance optimization and now code duplication in this component (having RouterLink and a/button cases separately)

src/components/NcButton/NcButton.vue
ShGKme: The context of the variable is `NcButton` component itself. It is an internal variable in a small component that has alignment as a part of its interface.

What reversed alignment means for the component is already described in the component public API documentation. Having it explained in many places likely lead to inconsistent changes in the future, when it only changed in a single place.

If it is still unclear in the component source, what `isReverseAligned` mean because it may have different meaning, we may think about a better naming. But for this specific variable - I don't have any idea how it might be misunderstood or misinterpreted.

src/components/NcButton/NcButton.vue
ShGKme: In general - yes. but if we have a complex template with many checks for a specific component type case, I'd prefer to have as direct check as possible.

- `:type="isNativeButton ? type : undefined"` - direct
- `:type="!isLink ? type : undefined"` - less direct
- `:type="!isNativeLink && !isRouterLink ? type : undefined"` - even less direct
- `:type="!to && !href ? type : undefined"` - least direct

ShGKme: As described in another PR: to make code cleaner, shorter and more reliable in the future.

Comments are needed and good when:
- They describe public API, like component props
- They describe some non-obvious code, like a rare case, helping to understand why this code is needed here or what it does

But not when they describe exactly what code does directly.

Such comments:
- In old code often lead not to explaining something, but to a conflict with the actual code, because developers change the code without changing a comment all the time
- Increase code size (in the case of this specific code block - increase 5 times)
- Provoke to write explanation instead of improving code making it cleaner (e.g. via variable name or structure)
- Provoke to add redundant comments in new code because there are comments everywhere already

This leads to a habit to quickly add comments everywhere with quantity over quality.

ShGKme: > * Provoke to add redundant comments in new code because there are comments everywhere already

For example, if some code already has comments everywhere like this, developers want to add same comments to the new code (to be consistent)

```js
// Use ResizeObserver to observe size changes
useResizeObserver()

// fetch all data
fetchAll()

// recover file
recoverFile()
```

And then by having less value of every comment, the comments are less useful in places where they are really needed, moving to quantity over quality.

An old example from `NcFilePicker` where comments were added, explaining what code does, but not why.

```js
// reset the form
reset()

// Trigger click on the input to open the file picker
nextTick(() => inputElement.value!.click())
```

Example in this component: this comment explained that `is link` checks `if this is link` and then explains that `link is <a>`.

```js
/**
 * If this is a link (<a>) element
 */
const isLink = ...
```

On the one hand, the comment added nothing to the variable name (isLink checks for "is link")

On the other hand, it gives a wrong understanding because using a component as a link (component is a link) still results in `<button>` element.

src/components/NcButton/NcButton.vue
ShGKme: Explained in the first comment

ShGKme: If `flexAligment` name is not clear in the internal context of a small component, let's name it better to make it directly clear, not after reading documentation on hover in IDE.

src/components/NcButton/NcButton.vue
susnux: > :type="isNativeButton ? type : undefined" - direct
> :type="!isLink ? type : undefined" - less direct

Why not this so no negate needed if thats the problem:

`:type="isLink ? undefined : type"`

---

But this is not really important for me, its just one additional computed :sweat_smile:

src/components/NcButton/NcButton.vue
susnux: Yes that true, but IMHO at least we should sanitize *our* props from misuse

src/components/NcButton/NcButton.vue
susnux: Sure I do not say its bad to know this - its internal - all I say that in that case the type of `ncPopoverTriggerAttrs` should not be `Record<string, ...>` but static `{ 'aria-haspopup': ..., 'aria-expanded: ... }`.

src/components/NcButton/NcButton.vue
susnux: yes see also: https://github.com/nuxt/nuxt/blob/e832d74376fef300642f13794b9c6bbfe7453459/packages/nuxt/src/app/components/nuxt-link.ts#L183-L184

src/components/NcButton/NcButton.vue
ShGKme: > Why not this so no negate needed if thats the problem:

For the same reason. `type` is a button attribute, and we set it to some value because we have a button.

When we check for a link, it has a hidden condition that "a link is the opposite for a button".

---

But this is not really important because I need to revert this change due to `href` issue

src/components/NcButton/NcButton.vue
ShGKme: Not relevant anymore after reverting to `v-bind="attrs"` approach

src/components/NcButton/NcButton.vue
ShGKme: Not relevant anymore after reverting to `v-bind="attrs"` approach

src/components/NcButton/NcButton.vue
ShGKme: Would be useful if `useLink` is used

Also they use `useRouter` which warn in case of no router in the setup context.

ShGKme:
Unfortunately, passing `:attr="undefined"` to a component overrides its root attribute.

Thus it is not possible to pass different sets of attributes individually for patch flags optimization...

Going back to `v-bind="attrs"` approach.

ShGKme:
- Fixed `to` type
- Added tests

tests/unit/components/NcButton/NcButton.spec.ts
```diff
@@ -0,0 +1,163 @@
+/**
+ * SPDX-FileCopyrightText: 2023 Nextcloud GmbH and Nextcloud contributors
```
susnux:
```suggestion
/*
 * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
```

susnux (approved):
code wise looks good - still not agree with the comment topic but thats only a small part here.

src/components/NcButton/NcButton.vue
susnux: Yes but still it now requires you to spent time reading the template, instead of just get the full context of the variable by read itself (or hover it in your IDE).

> In old code often lead not to explaining something, but to a conflict with the actual code, because developers change the code without changing a comment all the time

Sounds then like a bad developer not updating the comment ;)

> Increase code size (in the case of this specific code block - increase 5 times)

Only lines but not real code, also from my point they help to separate - especially in script-setup - the variables visually giving it a bit of structure. On the otherhand they do not increase complexity but ideally reduce it by explaining whats going on without need of deep insights into the whole code - usefull if multiple developers work on it and especially only rarely every now and then.

> Provoke to add redundant comments in new code because there are comments everywhere already

*Well I know some developers that would argue that is good and want nearly 1 line comment per 1 line code (yes I had to work on such projects).* (I agree that useless comments like the `reset` example above should be removed).

src/components/NcButton/NcButton.vue
```diff
@@ -603,105 +603,93 @@ defineSlots<{
...
+	}
 
-/**
- * Attribute to pass to the <button> element
- */
-const buttonAttrs = computed(() => ({
-	'aria-pressed': hasPressedState.value ? String(props.pressed) : undefined,
-	type: props.type,
-}))
+	return undefined
```
susnux, line 664: I wonder that TS is not complaining about dead code here as this can never be reached (and it not needed because no return also is undefined)

ShGKme: Without it ESLint is complaining about missing return in computed

src/components/NcButton/NcButton.vue
```diff
@@ -603,105 +603,97 @@ defineSlots<{
...
 const isReverseAligned = computed(() => props.alignment.includes('-'))
 
 const getNcPopoverTriggerAttrs = inject<() => Record<string, string|undefined>>('NcPopover:trigger:attrs', () => ({}), false)
 const ncPopoverTriggerAttrs = computed(() => getNcPopoverTriggerAttrs())
 
-/**
- * Attributes to pass to the <a> element
- */
-const linkAttrs = computed(() => ({
-	role: 'button',
```
Antreesy, line 657: There's no harm to lose a role for a link?

ShGKme: Before this PR a router link was rendered as an HTML button.

In this PR router link is rendered as a native link and has a native link role.

ShGKme:
- Rebased onto main
- Squashed fixup commits
- Added comments to the `variantWithPressed`

susnux (changes requested):
> In this PR router link is rendered as a native link and has a native link role.

:warning: this is a breaking change!
In v8 (and before) it also always had the `button` role, this will break many e2e tests as the role now is `link`

https://github.com/nextcloud-libraries/nextcloud-vue/blob/c5dd42d739bc48817c7004c5f011001b9fbc231c/src/components/NcButton/NcButton.vue#L711

---

either revert or add this to the breaking changes section of the changelog

ShGKme:
> this will break many e2e tests as the role now is `link`

Do you have an example?

Why do they rely on a button role while clicking on something with a link role?

susnux:
> Why do they rely on a button role while clicking on something with a link role?

`cy.findByRole('button', ...)`

Because thats much less error prone as searching for a `button` element, as there could - like in this example - also be a `a` element with `button` role. So for e2e tests its recommended to do what a human would do: Look for a button like element with given label.

---

But also tests aside this is breaking and should either be marked as such (I am fine with that!) or re-add the role.

ShGKme:
> or re-add the role.

It is not valid semantically and from a11y

susnux:
> It is not valid semantically and from a11y

Well if it is a router link than it does not behave like a link but more like a button (it does not reload the page at least).
But as said I agree it makes no sense as this is just a button styles link. Though we should mention this in the changelog.

susnux (approved):

susnux:
rebased + docs

PR #7103, opened 2025-06-29
fix(NcDateTimePicker): make clearable work

raimund-schluessler:
### ☑️ Resolves

- Fixes `NcDateTimePicker` clear button, both function and style wise.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
grafik | grafik

### 🏁 Checklist


src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -579,8 +580,13 @@ const ariaLabels = computed(() => ({
 		<template #input-icon>
 			<NcIconSvgWrapper :path="mdiCalendarBlank" :size="20" />
 		</template>
-		<template #clear-icon>
-			<NcIconSvgWrapper inline :path="mdiClose" :size="20" />
+		<template #clear-icon="{ clear }">
+			<NcButton variant="tertiary-no-background"
```
susnux:
```suggestion
			<NcButton :aria-label="t('Clear value')"
				variant="tertiary-no-background"
```

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -261,7 +262,7 @@ const props = withDefaults(defineProps<{
 	 *
 	 * When using the range picker then an array containing the start and end date needs to be passed.
 	 */
-	modelValue?: Date | [Date, Date]
+	modelValue?: Date | [Date, Date] | null
```
susnux, line 265: Either all of those changes here in `<script>` or just adjust the listener to emit `undefined` instead of `null` then all this stays valid.

raimund-schluessler: I don't mind. I just thought it was the more correct thing to do, since the docs of `modelValue` state: `null is emitted if clearable is set to true and the value was cleared.`.

PR #7110, opened 2025-07-01
fix(useHotKey): allow to pass key codes ('a' as 'KeyA')

Antreesy:
### ☑️ Resolves

- Fix https://github.com/nextcloud/spreed/issues/15393
- Allows to use hotkeys on non-latin keyboards
- Since validator accepts string arrays,  `['w', 'KeyW']` is backward-compatible with `'w'`, but can also be replaced with `'KeyW'`

### 🖼️ Screenshots

en | ru | de
-- | -- | --
image | image | image

### 🚧 Tasks

- [ ] AFAIK we do not use keys that occure mixed in different layouts (e.g. QWERTZ or AZERTY). Would it be an issue?

### 🏁 Checklist

### DIFF
--- a/src/composables/useHotKey/index.ts
+++ b/src/composables/useHotKey/index.ts
@@ -135,6 +135,7 @@ export function useHotKey(
 			return event.key === key
 		}
 		return event.key.toLowerCase() === key.toLowerCase()
+			|| event.code === key
 	}
 
 	/**
--- a/docs/composables/useHotKey.md
+++ b/docs/composables/useHotKey.md
@@ -37,6 +37,7 @@ where:
 
 <template>
 	<div class="container">
+		<p class="description">{{ loggedKey }}</p>
 		<p class="description">Press <kbd>W</kbd> <kbd>S</kbd> <kbd>A</kbd> <kbd>D</kbd> keys to move the ball</p>
 		<div class="square">
 			<div class="circle" :style="{ left: `${circleX}px`, top: `${circleY}px` }"></div>
@@ -57,11 +58,35 @@ where:
 	import { ref } from 'vue'
 	import { useHotKey } from '../../src/composables/useHotKey/index.js'
 
+	const isMac = /mac|ipad|iphone|darwin/i.test(navigator.userAgent)
+
 	export default {
 		setup() {
+			const loggedKey = ref('Press any key')
 			const circleX = ref(20)
 			const circleY = ref(20)
 			const highlighted = ref(false)
+
+			const updateLoggedKey = (event) => {
+				if (['Ctrl', 'Meta', 'Alt', 'Shift'].includes(event.key)) {
+					return
+				}
+
+				loggedKey.value = `Key pressed: ${event.key} | `
+				if (event.ctrlKey) {
+					loggedKey.value += 'Ctrl + '
+				}
+				if (event.metaKey) {
+					loggedKey.value += isMac ? 'Cmd + ' : 'Meta + '
+				}
+				if (event.altKey) {
+					loggedKey.value += 'Alt + '
+				}
+				if (event.shiftKey) {
+					loggedKey.value += 'Shift + '
+				}
+				loggedKey.value += event.code
+			}
 
 			const moveUp = (event) => {
 				circleY.value = Math.max(0, circleY.value - 10)
@@ -79,13 +104,15 @@ where:
 				highlighted.value = !highlighted.value
 			}
 
-			useHotKey('w', moveUp)
-			useHotKey('s', moveDown)
-			useHotKey('a', moveLeft)
-			useHotKey('d', moveRight)
+			useHotKey(true, updateLoggedKey)
+			useHotKey(['w', 'KeyW'], moveUp)
+			useHotKey(['s', 'KeyS'], moveDown)
+			useHotKey(['a', 'KeyA'], moveLeft)
+			useHotKey(['d', 'KeyD'], moveRight)
 			const stop = useHotKey('m', toggleHighlighted, { push: true })
 
 			return {
+				loggedKey,
 				circleX,
 				circleY,
 				highlighted,


susnux (approved):

src/composables/useHotKey/index.ts
```diff
@@ -135,6 +135,7 @@ export function useHotKey(
 			return event.key === key
 		}
 		return event.key.toLowerCase() === key.toLowerCase()
+			|| event.code === key
```
ShGKme: Does it mean that we optionally allow using key code (`KeyW`) as the hotkey, but the current usage as well as simple usage with `w` value still only works with latin layout?

Antreesy: Yeah, I guess.
I don't see how we can work around it without adding a keymap 'latin' - 'code'.

ShGKme: > Yeah, I guess

Then it doesn't fix the issue, as all the current broken places will be still broken. Only add a feature to support a new case.

> I don't see how we can work around it without adding a keymap 'latin' - 'code'.

Cannot we start with transforming single character `<c>` into `Key<C>` code?

Antreesy: Same result (and same qwertz issue), but no changes from app side? I don't mind, but if only we explicitly allow to set Key and advice in documentation (not deprecate normal keys)

But to keep in mind, it will cover only [a-z] and not digits/numpad/system keys

susnux: Question:
Is this always on the same key? Or are those short cuts localized?
Because then we probably need some other mechanic to localize the shortcuts.

Antreesy: It's not e.g. for german layout (z - KeyY), if that is what you mean.
- now: to trigger 'z', you need to press where German 'z' key is located, and with Cyrillic layout you simply can't do it;
- with this PR: if adjust in app from 'z' to 'KeyZ', in German it will trigger from 'y' instead.
- with this PR + fixup: no adjustments in apps, in German it will trigger from both 'z' (match) and 'y' (match with KeyZ).

To localize - you thinking about showing it as keyboard shortcuts in settings? If so, then yeah, we would need a way to detect different layouts and keymap, like in https://developer.mozilla.org/en-US/docs/Web/API/KeyboardLayoutMap/get

susnux: > with this PR + fixup: no adjustments in apps, in German it will trigger from both 'z' (match) and 'y' (match with KeyZ).

This sounds quite bad. E.g. `ctrl + z` is undo and `ctrl + y` is redo, so both would end up on the same key.

Antreesy: Then we should drop the new changes =P

And let the app-dev decide, which buttons they expect, being aware, that combinations like `Ctrl+z` might be in different place or not available for others layouts

susnux: Or add a new option to only use either code or key.

Or as said we need a new method of localizing keys, not sure where you would expect this (locale or language).
As the Browser API is only available on chromium based browsers.

Antreesy: > Or add a new option to only use either code or key.

First-commit implementation allows exactly that, developer is free to pick key, code, or any combination of them. I wouldn't limit that with passing an explicit option, if I can already `useHotKey('KeyA')`

> we need a new method of localizing keys

That one would be the actual fix for current issue.. What if we reverse the process?
- we limit function use to latin symbols (a-zA-Z), digits 0-9 and lang-independent characters (?,!. and so on) with regex validation
- developer sets `useHotKey('z')`
- user clicks where 'z' is drawn on their keyboard (e.g. with Cyrillic that would be `{ key: 'я', code: 'KeyZ' }`
- key doesn't match any of expected characters, so we trying to interpret the key (KeyZ -> z -> match!)

As far as I know and see on screenshots, all non-latin languages have QWERTY as backup

ShGKme: This must be a common problem for multi-lang apps. Let's check what the common solutions for it are.

Antreesy: 1. Discord uses fork of this react lib : https://github.com/ianstormtaylor/is-hotkey/blob/master/src/index.js
  It's checking for event.key, and in case of no match - for deprecated event.which -> I guess that would be a common approach
2. From observed vue libraries, they do not support non-latin chars. Or like vueuse, provide a filter function to define on app level

susnux: > Let's check what the common solutions for it are.

I see those options used:
- Ignore this possibility (e.g. cyrillic keyboard) :x: 
- Use default English (Z) but let the user customize short cuts (for us I would say this only makes sense if we configure them globally for all apps)
- Use the users language (they often do not distinguish locale and language) - This is not really helpful as you can have e.g. language set to something other but use an English keyboard.
- Just have fallbacks like use `z` or `З` or `ؤ`

The last option probably is the best as its rarely that you have the English and a different letter on the keyboard, I also only know this problem for Cyrillic and Arabic as others just use keyboards similar to QWERTY and then have composing.

Antreesy: > * Just have fallbacks like use `z` or `З` or `ؤ`

Then there is Chinese/Greek/Armenian/Georgian/... You can't win this fight.
Then only common thing that they might have underneath is a second layout on a keyboard, corresponding to keycodes (and it's also not 100% reliable, as it's usually QWERTY, but @DorraJaouad replied she's got Arabic + French AZERTY)

susnux: > replied she's got Arabic + French AZERTY

And you cannot even trust that, a friend of mine uses Arabic + QWERTY :sweat_smile:

susnux: > Then there is Chinese/Greek/Armenian/Georgian/... You can't win this fight.

Why? We have translators that know what the correct key for their keyboard would be.
So we could always try e.g. `if (key === 'z' || key === t('z'))` (on app level?).

Antreesy: > We have translators that know what the correct key for their keyboard would be.

If it's about making a map key - code for each language, I wouldn't trust it to translators and compile it ourselves (e.g. exports Windows / macOS layouts and keep letters only).

But IMO trying to guess `я -> KeyZ - z` is the easiest way to get the most coverage. We shouldn't guess user keyboard layout from NC user language. 

There's a recent descriptive review on available libs, where nobody does that =)
https://blog.duvallj.pw/posts/2025-01-10-all-javascript-keyboard-shortcut-libraries-are-broken.html

susnux: True. Maybe:

1. Check if key matches
2. If not check if key is a roman letter or a number
3. If it is not then check code

E.g. if you press `z` it will react on it. If you press `я` then it will be non roman and you use the code which is `KeyZ`.

This should work nearly everywhere.

ShGKme: > * If not check if key is a roman letter or a number
> * If it is not then check code

And what if yes, and it is a Roman letter? E.g. Z for CTRL+Z pressed by CTRL+Y on German layout or the opposite.

As far as I know, on German layout undo is still the same key, CTRL+Y on German layout (keeping the muscle memory while switching layout).

susnux: Undo is `ctrl + z` also on German keyboard but the physical location is where on English keyboards the `y` is located. Meaning for undo on a German keyboard you press ctrl + z (which is the same as ctrl + KeyY).

ShGKme: > Undo is `ctrl + z` also on German keyboard but the physical location is where on English keyboards the `y` is located.

Which is exactly what I said :)

So for Z in undo we also need to check for `KeyZ`, like with `я`, don't we?

susnux: As far as I know the shortcuts are everywhere the same (like `ctrl + z`) only for keyboards with different alphabet it depends on the underlying roman keyboard layout.

So there would be 1 problem and thats when the underlying roman keyboard does not match this assumption.
E.g. normally undo on a Cyrillic keyboard should be CTRL + я but I know that there are keyboards with QWERTZ layout - especially famous for Russian speaking Germans where undo would be CTRL + Н

ShGKme: > 1. Check if key matches
> 2. If not check if key is a roman letter or a number
> 3. If it is not then check code

Maybe I understood this wrong or I misunderstood what `else` supposed to be.
But this algorithm seems to break CTRL+Z in German, doesn't it?

susnux: > So for Z in undo we also need to check for KeyZ, like with я, don't we?

No its Z the key Z not the code KeyZ.
Meaning Undo is in key: `CTRL + Z` which is in code: `ControlLeft + KeyY`.

So for roman keyboards its always the `key` not the `code`. Only for non-roman keyboards you have to check the `code`. And then ignore special cases like the mentioned one :sweat_smile:

Antreesy: yup, if you on German layout, Ctrl+z should work as expected, since z is in [a-z]
Otherwise it might be expected as other apps behave. e.g. for me in PHPStorm and QWERTY kbd - with english and russian layout Undo is ctrl+z, and with german layout is ctrl+y

So, sounds like we agreeing on solution like in the last commit now?)

src/composables/useHotKey/index.ts
```diff
@@ -134,7 +136,15 @@ export function useHotKey(
 		if (options.caseSensitive) {
 			return event.key === key
 		}
+
+		if (latinLettersRegex.test(key) && nonLatinLettersRegex.test(event.key)) {
+			// If received key is a single letter, but non-Latin, derive it from key code
+			const parsedKey = event.code.replace('Key', '').toLowerCase()
+			return parsedKey === key.toLowerCase()
+		}
+
```
susnux: We should have the same for numbers I think. At least I do not know if Arabic will use the Arabic numbers on `key` or use the Indian numbers they use (e.g. `2` vs `٢`)

src/composables/useHotKey/index.ts
```diff
@@ -134,7 +136,15 @@ export function useHotKey(
...
 		return event.key.toLowerCase() === key.toLowerCase()
+			|| event.code === key
```
susnux: Imho if we fix it with your current solution we should revert all of the "allow passing `code`" thing.
It opens the gates for more issues :sweat_smile:

src/composables/useHotKey/index.ts
```diff
@@ -134,6 +137,12 @@ export function useHotKey(
 		if (options.caseSensitive) {
 			return event.key === key
```
susnux: what if I make it case sensitive but user has a different keyboard layout?
Maybe we need:

```js

if (casesensitive) {
    if (key === event.key) { return true }
    const isLowerCase = key.toLowerCase() === key
    const isPressedKeyLowerCase = event.key.toLowerCase() === event.key
    return derivedKey === key.toUpperCase() && isLowerCase === isPressedKeyLowerCase
}
```

We cannot use modifiers as IIRC there is no capslock modifier in the event.

Antreesy: Got you covered. Thanks for the hints!

src/composables/useHotKey/index.ts
```diff
@@ -124,17 +126,34 @@ export function useHotKey(
...
+
+		const isKeyInLowerCase = key === key.toLowerCase()
+		const isEventKeyInLowerCase = event.key === event.key.toLowerCase()
+
+		// If key and event.key are in different cases, invalid with caseSensitive = true. Do not perform further checks
+		if (options.caseSensitive && isKeyInLowerCase !== isEventKeyInLowerCase) {
+			return false
+		} else if (event.key.toLowerCase() === key.toLowerCase()) {
+			return true
+		}
```
susnux:
```suggestion
		// If key exactly matches event.key, valid with any caseSensitive option. Do not perform further checks
		if (event.key === key) {
			return true
		}

		if (options.caseSensitive) {
			const isKeyInLowerCase = key === key.toLowerCase()
			const isEventKeyInLowerCase = event.key === event.key.toLowerCase()

			if (isKeyInLowerCase !== isEventKeyInLowerCase) {
				return false
			}
			return event.key.toLowerCase() === key.toLowerCase()
		}
```

Otherwise the `options.caseSensitive` is ignored if they do not match.
if `caseSensitive` is set this branch **must** return.

Antreesy: It's not ignored, and you missing deriving that way

If caseSensitive is true and cases do not match
  return false

Otherwise they match, sensitive option is fulfilled, and we continue checking

Maybe else-if is confusing, I can put it to newline

susnux: Then the whole else if can just be removed

Antreesy: It can not be removed, it's comparing `z` with `Z` 👀

Need some tests for it probably

tests/unit/composables/useHotKey.spec.js
```diff
@@ -0,0 +1,155 @@
...
+		it.each([
+			// Latin characters
+			['a', { key: 'a', code: 'KeyA' }, {}],
+			['a', { key: 'A', code: 'KeyA' }, {}],
+			// Latin characters with different layout
+			['a', { key: 'a', code: 'KeyQ' }, {}],
+			// Non-Latin characters
+			['a', { key: 'ф', code: 'KeyA' }, {}],
+			['a', { key: 'Ф', code: 'KeyA' }, {}],
+			['a', { key: 'ち', code: 'KeyA' }, {}],
```
susnux: I do not think this is a real life example, as CJK layouts would emit the plain letter first and then start composing, no?

tests/unit/composables/useHotKey.spec.js
```diff
@@ -0,0 +1,155 @@
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import { isRef, nextTick } from 'vue'
+import { mount } from '@vue/test-utils'
```
susnux:
```suggestion
```

unused

tests/unit/composables/useHotKey.spec.js
```diff
@@ -0,0 +1,155 @@
...
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import { useHotKey } from '../../../src/composables/useHotKey/index.ts'
+import { resizeWindowWidth } from '../testing-utils.ts'
```
susnux:
```suggestion
```

unused

tests/unit/composables/useHotKey.spec.js
```diff
@@ -0,0 +1,155 @@
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
```
susnux:
```suggestion
 * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
```

tests/unit/composables/useHotKey.spec.js
```diff
@@ -0,0 +1,155 @@
...
+	describe('options', () => {
+		it('should accept array of keys and invoke callback for all of them', () => {
+			const stop = useHotKey(['a', 'b'], mockCallback)
+			triggerKeyDown({ key: 'a', code: 'KeyA' })
+			triggerKeyDown({ key: 'b', code: 'KeyB' })
+			stop()
+
+			expect(mockCallback).toHaveBeenCalledTimes(2)
+		})
+
```
susnux, line 51:
```suggestion
```

src/composables/useHotKey/index.ts
```diff
@@ -124,16 +126,31 @@ export function useHotKey(
...
+		if (event.key === key) {
+			return true
+		}
+
+		const isKeyInLowerCase = key === key.toLowerCase()
+		const isEventKeyInLowerCase = event.key === event.key.toLowerCase()
+
+		// If key and event.key are in different cases, invalid with caseSensitive = true. Do not perform further checks
+		if (options.caseSensitive && isKeyInLowerCase !== isEventKeyInLowerCase) {
+			return false
```
susnux: maybe do not compute this all the time?

```suggestion
		// If key and event.key are in different cases, invalid with caseSensitive = true. Do not perform further checks
		if (options.caseSensitive) {
			const isKeyInLowerCase = key === key.toLowerCase()
			const isEventKeyInLowerCase = event.key === event.key.toLowerCase()
 			if (isKeyInLowerCase !== isEventKeyInLowerCase) {
				return false
			}
```

susnux (approved):

tests/unit/composables/useHotKey.spec.js
Antreesy: Replaced with Hebrew `ש` as a compromise

susnux (approved):

src/composables/useHotKey/index.ts
```diff
@@ -124,16 +126,32 @@ export function useHotKey(
...
+			const isEventKeyInLowerCase = event.key === event.key.toLowerCase()
+			if (isKeyInLowerCase !== isEventKeyInLowerCase) {
+				return false
+			}
 		}
+
+		// If received event.key is not a printable ASCII character code (character code 32-127),
+		// try to derive it from event.code and match with expected key
+		if (derivedKeysRegex.test(key) && nonAsciiPrintableRegex.test(event.key)) {
+			return event.code.replace(/(Key|Digit|Numpad)/, '') === key.toUpperCase()
```
ShGKme:
```suggestion
			return event.code.replace(/^(?:Key|Digit|Numpad)/, '') === key.toUpperCase()
```

PR #7116, opened 2025-07-02
fix(NcActionText): remove br between action name and action long text

GretaD:
### ☑️ Resolves

The `br` adds too much space between `action-text__name` and `action-text__longtext`. Talk app and Mail app use NcActionText and so far we could find only these apps where it breaks.

On [NcActionButton](https://github.com/nextcloud-libraries/nextcloud-vue/pull/6932) Maksim solved it by adding a description prop, but in our case the entry is fully informational and standalone([cn](https://github.com/nextcloud/mail/pull/11347#issuecomment-3027766033)).

### 🖼️ Screenshots

🏚️ Before | 🏡 After
Screenshot from 2025-07-02 13-50-44|Screenshot from 2025-07-02 16-37-12

Screenshot from 2025-07-02 12-35-21
Screenshot from 2025-07-02 12-35-41

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/NcActionText/NcActionText.vue
+++ b/src/components/NcActionText/NcActionText.vue
@@ -22,7 +22,6 @@
 				<strong class="action-text__name">
 					{{ name }}
 				</strong>
-				<br>
 				<!-- white space is shown on longtext, so we can't
 				put {{ text }} on a new line for code readability -->
 				<span class="action-text__longtext" v-text="text" />


Antreesy (approved):
I'll follow-up with 'description' prop PR later, still think it makes sense to have it aligned with other components

GretaD:
> I'll follow-up with 'description' prop PR later, still think it makes sense to have it aligned with other components

i can do that as well, if you think we must have it.

PR #7135, opened 2025-07-09
feat(NcPasswordField): allow to set visibility state

susnux:
### ☑️ Resolves

Allows to programmatically set the obfuscation state of the password.
This is useful e.g. to hide the password in some cases, like submitting the login form.

### 🏁 Checklist


src/components/NcPasswordField/NcPasswordField.vue
```diff
@@ -162,9 +139,14 @@ const props = withDefaults(defineProps<Omit<NcInputFieldProps, 'trailingButtonLa
...
 })
 
-const modelValue = defineModel<string>('modelValue', { default: '' })
+const modelValue = defineModel<string>({ default: '' })
 watch(modelValue, debounce(checkPassword, 500))
 
+/**
+ * The obfuscating state of the password.
+ */
+const isPasswordHidden = defineModel<boolean>('obfuscated', { default: true })
```
ShGKme: To not have one thing exist with to names in toe instance.
Currently in the template `obfuscated` and `isPasswordHidden` is the same thing.

```suggestion
const obfuscated = defineModel<boolean>('obfuscated', { default: true })
```

ShGKme: But I'd vote for a simpler name, something more commonly used for a password component like `visible` (Ant Design, Mantine).

susnux: Its the wording MDN uses to describe the feature :sweat_smile:

susnux: But for me any name would work

kesselb (approved):
Thank you for picking it up 🙏 

Using "v-model:obfuscated" is much nicer than modifying the state directly. ;)

susnux:
`obfuscated` ( :heart: ) vs `visible` ( :rocket: ) for the property name?

src/components/NcPasswordField/NcPasswordField.vue
ShGKme: `visible` follows simple language and is used by 2 good UI Kits: Mantine and Ant-Design

One other Vue UI Kit uses `show-pasword` which is also simple, but multiword.

I haven't found any other good UI kit with a dedicated password field and/or visibility toggle.

susnux: Well `obfuscated` is also a simple word - at least in technological context this probably also is in the basic vocabulary.

In this case its probably the opposite `obfuscated` as default `true` and `visible` with default `false`.

BTW in bigger projects like Material UI (react) its `showPassword` and for others you just manually switch the type from `password` to `text` (e.g. swiftui).

ShGKme: As mentioned, I'm also fine with `show-password`

susnux:
Lets go with visible, even if I think `obfuscated` is also a good name, its better to stick with boolean values that default to `false` - in this case it would be `non-obfuscated` or we inverse it and just use `visible`.

PR #7138, opened 2025-07-09
refactor(NcProgressbar): migrate component to Typescript and update docs

susnux:
### ☑️ Resolves

- for #841 

1. Migrate to Typescript and script setup
2. Updated docs to include examples for all use cases, helps to spot potential regressions.
3. Use variables instead of magic numbers

### 🖼️ Screenshots

Screenshot 2025-07-09 at 23-08-38 Nextcloud Vue Style Guide

### 🏁 Checklist


src/components/NcProgressBar/NcProgressBar.vue
```diff
@@ -31,31 +34,125 @@ This is a simple progress bar component.
...
+
+/**
+ * Current progress as a value between 0 and 1
+ */
+const progress = computed(() => Math.max(0, Math.min(100, props.value)) / 100)
+
+// design constants
+const gridBaseline = Number.parseInt(window.getComputedStyle(document.body).getPropertyValue('--default-grid-baseline'))
+const clickableArea = Number.parseInt(window.getComputedStyle(document.body).getPropertyValue('--default-clickable-area'))
+const clickableAreaSmall = Number.parseInt(window.getComputedStyle(document.body).getPropertyValue('--clickable-area-small'))
```
ShGKme: Every component instance will now make DOM API calls on creation, which makes this component less performant after `TS/script setup` migration.

Do we have a case where we need to support different theming for different component instances?

If there is no such case, let's calculate it only once.

src/components/NcProgressBar/NcProgressBar.vue
```diff
@@ -31,31 +34,125 @@ This is a simple progress bar component.
...
+
+/**
+ * The real radius of the progress circle which includes some padding against the wrapper
+ */
+const radiusNormalized = computed(() => radius.value - stroke.value)
+
+/**
+ * The circumfence which is used to calculate the prgress bar within the circle.
+ */
+const circumference = computed(() => radiusNormalized.value * 2 * Math.PI)
```
ShGKme: With many similar changes during TS/<script setup>, I have a fear that this will provoke bad habits for the code style by adding code comments everywhere.
- It provokes writing explanations in a native language instead of making the code clearer, understandable from the code itself with better naming or structure
- Having comments everywhere provokes adding meaningless comments to new code (because there are comments everywhere already)
  ```ts
  // Use ResizeObserver to observe size changes
  useResizeObserver()
  
  // fetch all data
  fetchAll()
  
  // recover file
  recoverFile()
  ```
- It makes simple code much longer (here 4 times longer)
- It often results in inconsistency between the code and the comment after some changes. Developers update code without changing comments all the time.

Let's only keep comments when:
- We cannot make it clear from naming or code structure
- There is some complex edge case which requires explanation
- It is a part of a public API like component props

For example:
- `cssHeight` may be named `heightPx` making it clear that it is the same height but in `px` and not some other height to use in CSS with a different value
- `radiusNormalized` may be named `innerRadius` or `innerCircleRadious` (otherwise there is a need for an explanation what "normalizing" means here)

susnux: I adjusted the names. Originally I only took over them without changing.
I agree they are not self descriptive.

src/components/NcProgressBar/NcProgressBar.vue
susnux: fixed

src/components/NcProgressBar/NcProgressBar.vue
ShGKme: @susnux You forgot to push?
Or I misunderstood what you agree with 🤔

susnux: I adjusted the variable names to match better their purpose

ShGKme (approved):

PR #7140, opened 2025-07-10
export everything from NcRichtext/index.js

grnd-alt:
in components/index.ts the default export of `NcRichText/index.js` was only exported as NcRichtext.
In deck we import NcReferenceList from `@nextcloud/vue` so we would need that export
(also it makes ci fail on my [deck vue3-pr](https://github.com/nextcloud/deck/actions/runs/16169657238/job/45639745008?pr=7003) :( )

- [ ] 2️⃣ Backport to `stable8` for maintained Vue 2 version or not applicable (idk if necessary)


src/components/index.ts
```diff
@@ -70,7 +70,7 @@ export { default as NcPopover } from './NcPopover/index.js'
 export { default as NcProgressBar } from './NcProgressBar/index.js'
 export { default as NcRelatedResourcesPanel } from './NcRelatedResourcesPanel/index.js'
 export { default as NcRichContenteditable, NcAutoCompleteResult, NcMentionBubble } from './NcRichContenteditable/index.js'
-export { default as NcRichText } from './NcRichText/index.js'
+export { NcRichText, NcReferenceList, NcReferenceWidget, NcReferencePicker, NcReferencePickerModal, NcSearch } from './NcRichText/index.js'
```
susnux, line 73: This sounds bad especially because of the names.
- `NcSearch` this is not a real search component but its for the reference picker and should probably be something like `NcReferenceSearch` or even `NcReferencePickerSearch` (or at least `NcRichtextSearch`)

grnd-alt: @susnux just The NcSearch one, or the exports in general? For my usecase in deck I only need NcReferenceList currently, but thought it'd be better to have them all exported. I think the ones called NcReference... are fine name wise?

ShGKme: Agree with renaming `NcSearch` to either `NcReferenceSearch` or `NcReferencePickerSearch` (if it is supposed to be used inside `NcReferencePicker`, following vue naming convention).

ShGKme: > For my usecase in deck I only need NcReferenceList currently, but thought it'd be better to have them all exported

There is one downside of exporting a component: it became a part of public API. This means that further renaming/removing such a component or changing its API is a breaking change.

So before making it public, I'd first make sure there is nothing we want to change and that there is a need.

susnux: > For my usecase in deck I only need NcReferenceList currently

Why then not import from `@nextcloud/vue/components/NcRichText`? There it is already exported?

PR #7151, opened 2025-07-15
fix(NcListItem): correctly handle unmounting

Antreesy:
### ☑️ Resolves

- Similar to #7077 
- When component is unmonted, `$refs.name` is null and therefore can no longer be used

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | A

### 🏁 Checklist

### DIFF
--- a/src/components/NcListItem/NcListItem.vue
+++ b/src/components/NcListItem/NcListItem.vue
@@ -776,7 +776,9 @@ export default {
 				return
 			}
 			// do not hide if focus is kept within
-			if (this.$refs['list-item'].contains(event.relatedTarget)) {
+			// On component unmounting Vue 3 resets the template ref to null,
+			// so we can ignore this check
+			if (this.$refs['list-item']?.contains(event.relatedTarget)) {
 				return
 			}
 			this.hideActions()


ShGKme (approved):

src/components/NcListItem/NcListItem.vue
```diff
@@ -776,7 +776,9 @@ export default {
 				return
 			}
 			// do not hide if focus is kept within
-			if (this.$refs['list-item'].contains(event.relatedTarget)) {
+      // On component unmounting Vue 3 resets the template ref to null,
+      // so we can ignore this check
+			if (this.$refs['list-item']?.contains(event.relatedTarget)) {
```
susnux:
```suggestion
			// On component unmounting Vue 3 resets the template ref to null,
			// so we can ignore this check
			if (this.$refs['list-item']?.contains(event.relatedTarget)) {
```

src/components/NcListItem/NcListItem.vue
```diff
@@ -776,7 +776,9 @@ export default {
 				return
 			}
 			// do not hide if focus is kept within
-			if (this.$refs['list-item'].contains(event.relatedTarget)) {
+      // On component unmounting Vue 3 resets the template ref to null,
```
susnux: is it null? thought undefined

Antreesy: Grigorii showed the place in vue code, but I don't have a link unfortunately, Just remember it is set with `= null`

ShGKme: @susnux https://github.com/vuejs/core/blob/v3.5.17/packages/runtime-core/src/renderer.ts#L2110

PR #7161, opened 2025-07-16
refactor(functions): Make sure only named exports are used (`usernameToColor`)

susnux:
### ☑️ Resolves

This one we somehow missed when looking for inconsistent export type.
Here the function was default exported - but then again named re-exported in the for the package exports.

Use consistent named exports for everything except components.

### 🏁 Checklist


Antreesy (approved):
I'm afraid it'll be lost in changelog (but also app will fail to build). Is there a need to backport it in compatible way with deprecation flag?

susnux:
> I'm afraid it'll be lost in changelog (but also app will fail to build). Is there a need to backport it in compatible way with deprecation flag?

Yes will export both in stable8 for forward compatibility. I think this only affects very few places (luckily).

PR #7177, opened 2025-07-17
refactor: replace deprecated CSS variables and legacy fallbacks

susnux:
### ☑️ Resolves

Those variables are deprecated since Nextcloud 30 and the fallbacks are not needed when using Nextcloud 30+.
As this version (v9) only targets Nextcloud 30+ at the moment we do not need them anymore.

### 🏁 Checklist

### DIFF
--- a/src/components/NcCounterBubble/NcCounterBubble.vue
+++ b/src/components/NcCounterBubble/NcCounterBubble.vue
@@ -212,7 +212,7 @@ const originalCountAsTitleIfNeeded = computed(() => {
 	text-align: center;
 	line-height: var(--counter-bubble-height); // Expand line-height to full height to center text vertically
 	padding: 0 calc(1.5 * var(--default-grid-baseline));
-	border-radius: var(--border-radius-pill);
+	border-radius: 0.5lh;
 	background-color: var(--color-primary-element-light);
 	font-weight: bold;
 	color: var(--color-primary-element-light-text);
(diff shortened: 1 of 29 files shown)


src/components/NcCounterBubble/NcCounterBubble.vue
```diff
@@ -212,7 +212,7 @@ const originalCountAsTitleIfNeeded = computed(() => {
 	text-align: center;
 	line-height: var(--counter-bubble-height); // Expand line-height to full height to center text vertically
 	padding: 0 calc(1.5 * var(--default-grid-baseline));
-	border-radius: var(--border-radius-pill);
+	border-radius: calc(var(--counter-bubble-height) / 2);
```
ShGKme:
```suggestion
	border-radius: var(--border-radius-pill);
```

ShGKme (approved):

src/components/NcCounterBubble/NcCounterBubble.vue
susnux: Pill is deprecated

ShGKme: What is the replacement for rounded (pill) elements?

susnux: Just use half of the height?

ShGKme: Why not use universal value without calculations?

ShGKme: 1lh also works here as a simple value, but I didn't get why remove universal pill variable

ShGKme: And half height is the same as just height here

susnux: Sure we could also just use
```suggestion
	border-radius: 0.5lh;
```

PR #7185, opened 2025-07-20
fix(useIsDarkTheme): provide default value to silence runtime warnings

susnux:
### ☑️ Resolves

- Fix https://github.com/nextcloud-libraries/nextcloud-vue/pull/6764#discussion_r2217310037

Otherwise if no `NcThemeProvider` is used a runtime warning is shown in the dev tools console.

### 🏁 Checklist

### DIFF
--- a/src/composables/useIsDarkTheme/index.ts
+++ b/src/composables/useIsDarkTheme/index.ts
@@ -54,7 +54,7 @@ const useInternalIsDarkTheme = createSharedComposable(() => useIsDarkThemeElemen
  */
 export function useIsDarkTheme(): DeepReadonly<Ref<boolean>> {
 	const isDarkTheme = useInternalIsDarkTheme()
-	const enforcedTheme = inject(INJECTION_KEY_THEME)
+	const enforcedTheme = inject(INJECTION_KEY_THEME, undefined)
 
 	return computed(() => {
 		if (enforcedTheme?.value) {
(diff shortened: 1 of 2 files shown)


src/composables/useIsDarkTheme/index.ts
```diff
@@ -54,10 +54,10 @@ const useInternalIsDarkTheme = createSharedComposable(() => useIsDarkThemeElemen
  */
 export function useIsDarkTheme(): DeepReadonly<Ref<boolean>> {
 	const isDarkTheme = useInternalIsDarkTheme()
-	const enforcedTheme = inject(INJECTION_KEY_THEME)
+	const enforcedTheme = inject(INJECTION_KEY_THEME, computed<''>(() => ''))
```
ShGKme: Why not `undefined`, to not create additional computed?
```suggestion
	const enforcedTheme = inject(INJECTION_KEY_THEME, undefined)
```

ShGKme (approved):

src/composables/useIsDarkTheme/index.ts
susnux: much better!

Antreesy: Could you tell if it's worth making a new release?
From what I see locally, this one thing is killing performance (when many avatars rendered, 136 in this example):

before | with `, undefined` added
-- | --
image | image

ShGKme: @Antreesy in production build?

Antreesy: So far noticed only while testing 9.0.0-rc.4 against this branch.
Maybe a result of #7157

PR #7192, opened 2025-07-21
fix(NcButton): ensure that the component also works without vue-router

susnux:
### ☑️ Resolves

- tested alternative to #7100 

1. Make the component properly work without vue-router (useLink must be used in setup context not in computed). We can always assume that a router does not exist if it is not available in setup context (it cannot be registered later).
2. Make sure a router link is also just an `a` element.
3. Ensure the behavior is the same as on v8: If a `to` is provided it has precedence over a `href` prop.

### 🏁 Checklist

### DIFF
--- a/src/components/NcButton/NcButton.vue
+++ b/src/components/NcButton/NcButton.vue
@@ -434,7 +434,7 @@ td.row-size {
 
 <script setup lang="ts">
 import type { Slot } from 'vue'
-import type { RouteLocation } from 'vue-router'
+import type { RouteLocationRaw } from 'vue-router'
 
 import { computed, inject, toRef } from 'vue'
 import { routerKey, useLink } from 'vue-router'
@@ -535,7 +535,7 @@ interface NcButtonProps {
 	 *
 	 * Note: This takes precedence over the href attribute.
 	 */
-	to?: string|RouteLocation
+	to?: string|RouteLocationRaw
 
 	/**
 	 * Specifies the button native type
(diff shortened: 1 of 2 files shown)


ShGKme (approved):

src/components/NcButton/NcButton.vue
```diff
@@ -605,16 +605,14 @@ defineSlots<{
 
 // Make sure the component also works if the app does not use any router
 // And if the app uses a router we need to make sure a `to` prop was passed to use to router
-const routerLink = computed(() => (
-	inject(routerKey, null) !== null && props.to
-		? useLink({ to: toRef(() => props.to!) })
-		: undefined
-))
+const routerLink = inject(routerKey, null) !== null
+	? useLink({ to: toRef(() => props.to!) })
```
ShGKme, line 609: This means that if there is vue-router in the context, but I did not pass `to`, I have `useLink({ to: undefined })` which is not a valid router link

ShGKme: So every native button and every native link instance still tries to resolve a route.

https://github.com/vuejs/router/blob/main/packages/router/src/RouterLink.ts#L142

susnux: Yes, but thats fine - it will just resolve to undefined route anyways.
It fixes the other issues - this one is only still an issue in `dev` mode.

But I am fine with closing this if you find a better solution :)

EDIT: I saw you pushed again on the other PR and made it reviewable so maybe lets move on there - only opened this as we needed to fix this asap and the other PR was in draft mode.

PR #7193, opened 2025-07-22
fix(NcRichText): correctly parse absolute links

Antreesy:
### ☑️ Resolves

- Fix rendering autolinks for cases:
  - protocol is missing - e.g. `www.nextcloud.com` is not resolved
  - markdown URL - e.g. `[www.nextcloud.com](www.nextcloud.com)` is resolved as relative link
    - having `//` is a requirement for an absolute link, can only prepend this and not full scheme
### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 Tasks

- [ ] adjust tests (for invalid links cases, and relative links cases)
- [ ] align with other apps (Collectives) on usage of URL_PATTERN / URL_PATTERN_AUTOLINK
- [ ] adjust both regexes to match links more correctly

### 🏁 Checklist


src/components/NcRichText/NcRichText.vue
```diff
@@ -577,6 +790,8 @@ export default {
 							...props,
 							to: route,
 						}, { default: () => children })
+					} else if (!/^https?:\/\//.test(props.href)) {
+						props.href = 'https://' + props.href
```
susnux, line 794: do we allow other protocols? Then this would be invalid (e.g. `tel:` or `mailto:`)

Antreesy: Internal regex doesn't allow it, but flavored markdown does. Maybe we can test against any schema present

PR #7200, opened 2025-07-24
refactor(NcAppNavigation): migrate component to Typescript and `script-setup`

susnux:
### ☑️ Resolves

- for #841 

### 🏁 Checklist


src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -134,10 +134,128 @@ emit('toggle-navigation', {
...
+	/**
+	 * aria-labelledby attribute to describe the navigation
+	 */
+	ariaLabelledby?: string
+}>()
+
+const setHasAppNavigation = inject<(v: boolean) => void>('NcContent:setHasAppNavigation', () => warn('NcAppNavigation is not mounted inside NcContent, this is probably an error.'), false)
+
+const appNavigationContainer = useTemplateRef('appNavigationContainer')
+const focusTrap = ref<FocusTrap>()
```
ShGKme: It is a general variable, not a reactive state

```suggestion
let focusTrap: FocusTrap
```

ShGKme: Or, if we really want a reactive variable just in case, let's not make entire trap object deep reactive.
```suggestion
const focusTrap = shalowRef<FocusTrap>()
```

susnux: TBH I did not wanted to touch implementation - it was in `data` thus reactive.
But I guess it does not need to be reactive here.

ShGKme (approved):

src/components/NcAppNavigation/NcAppNavigation.vue
ShGKme: It was, but in Options API there is no direct way to make something non-reactive or shallow reactive.

ShGKme: But I appreciate single-change refactoring

susnux: > It was, but in Options API there is no direct way to make something non-reactive or shallow reactive.

You could have put in on the instance in mounted without declaring it in the `data` - should always work (without type support).

Antreesy (approved):

PR #7202, opened 2025-07-24
fix(NcContent): flip skip action image in rtl mode

skjnldsv:
### ☑️ Resolves

- Fix https://github.com/nextcloud/server/issues/48848

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
 | 

### 🏁 Checklist

### DIFF
--- a/src/components/NcContent/NcContent.vue
+++ b/src/components/NcContent/NcContent.vue
@@ -219,6 +219,11 @@ function setAppNavigation(value: boolean): void {
 
 	&__image {
 		margin-top: 12px;
+
+		// Flip the image in RTL mode
+		&:dir(rtl) {
+			transform: rotateY(180deg);
+		}
 	}
 }
 


src/components/NcContent/NcContent.vue
```diff
@@ -188,6 +188,11 @@ function setAppNavigation(value: boolean): void {
 	height: 100vh;
 	padding: var(--body-container-margin)!important;
 	backdrop-filter: brightness(50%);
+
+	// Flip the image in RTL mode
+	.vue-skip-actions__image:dir(rtl) {
+		transform: rotateY(180deg);
+	}
```
susnux: Should be moved below to the `scoped` styles

skjnldsv: Does it matter? 🤭

ShGKme: It's a bit unclear why it is for :focus-within only

susnux: > Does it matter? 🤭

Yes to not have weird behavior if different library versions are used

skjnldsv: Fair enough !

skjnldsv: > It's a bit unclear why it is for :focus-within only 

Wondered the same. Tbh I put the rule here to avoid duplicating the focus-within part in case we change something in the future. Rules duplication is always error prone in the long run

ShGKme: We can avoid duplication With nesting (in both `<style>` and `<style scoped>`)

skjnldsv: > We can avoid duplication With nesting

Which is what I did ? :raised_eyebrow: 
I was referring to Ferdinand asking to move my code to the scoped section

susnux: Yes but that is not needed, you do not need the `:focus-within` for the image mirroring - it can be applied all the time

skjnldsv: ...

skjnldsv: I'm an idiot :see_no_evil:

skjnldsv: Done

skjnldsv:
Done!

susnux (approved):

ShGKme (approved):

PR #7208, opened 2025-07-28
fix(NcDateTimePicker): adjust input padding inline end

raimund-schluessler:
### ☑️ Resolves

- Adjusts the inline end padding for the `NcDateTimePicker` input field, so that less space is wasted if no clearable button is present.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
grafik | grafik
grafik | grafik

### 🏁 Checklist

### DIFF
--- a/src/components/NcDateTimePicker/NcDateTimePicker.vue
+++ b/src/components/NcDateTimePicker/NcDateTimePicker.vue
@@ -580,7 +580,7 @@ function cancelSelection() {
 			:aria-labels
 			:auto-apply="!confirm"
 			class="vue-date-time-picker"
-			:class="{ "vue-date-time-picker--clearable": clearable }"
+			:class="{ 'vue-date-time-picker--clearable': clearable }"
 			:cancel-text="t('Cancel')"
 			:clearable
 			:day-names


Antreesy (approved):
Looks good. Please squash commits and add a `clearable` prop in any example

ShGKme (approved):

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -692,6 +693,10 @@ function cancelSelection() {
 		@include meta.load-css('@vuepic/vue-datepicker/dist/main.css');
 	}
 
+	.vue-date-time-picker--clearable :deep(.dp__input) {
+		padding-inline-end: var(--default-clickable-area);
```
susnux, line 697: Should be not just update the `--dp-input-padding` variable here?

raimund-schluessler: You mean like this?
```suggestion
		--dp-input-padding: 6px var(--default-clickable-area) 6px 12px;
```
This should also work, I don't have a preference here, honestly.

Antreesy: but also if 6px/12px will change, we have to update in in both places. `-inline-end` is limiting changes

susnux: My reasoning was just that the variables are public API but the class names could break someday.

PR #7080, opened 2025-06-23
refactor(NcColorPicker): migrate component to Typescript

susnux:
### ☑️ Resolves

- for #841 

1. deprecate `update:open` its a duplicate of `close` and there is not even a `open` prop
2. migrate to Typescript and script-setup

### 🏁 Checklist

- [ ] ⛑️ Tests are included or are not applicable
- [x] 📘 Component documentation has been extended, updated or is not applicable
- [x] 2️⃣ Backport to `stable8` for maintained Vue 2 version or not applicable

PR #7212, opened 2025-07-28
feat(NcColorPicker): add proper open-state handling

susnux:
### ☑️ Resolves

With #7080 we dropped the `update:open` event, but the usage of the component is different - as you can already see in the docs of it. We in-officially used the attributes passing to the underlying `NcPopover` for custom open state handling.

So instead of this we should revert that change and instead just implement it correctly by providing the 2-way binding for `open` (consistent props naming) - then we need to reverse the event removal and remove `close` (which is covered by `update:open` and for fully closed (transitions done) the proper (consistent) named event `closed`).

I would backport the `feat` commits to v8 for forward compatibility.

### 🏁 Checklist


ShGKme (approved):

susnux:
/backport 292d4a7a7d8d111fec67acc0f7273d82a4e3f495 f4d8bc961e538f1a9e2ba17b0be717629d557a7c to stable8

PR #7213, opened 2025-07-28
chore: work around bug in `corepack` blocking dependency updates

susnux:
### ☑️ Resolves

dependabot uses `corepack` for all updates of node dependencies. The problem here is that `corepack` breaks if the
`devEngines.packageManager` does not define a **full semantic version** meaning it does not support semantic ranges.
We can work around this by just using the array syntax with only one entry.

For reference the *corepack* bug report: https://github.com/nodejs/corepack/issues/729

### 🏁 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,11 +67,11 @@
     "extends @nextcloud/browserslist-config"
   ],
   "overrides": {
-    "mdast-util-gfm": {
-      "mdast-util-gfm-autolink-literal": "2.0.0"
-    },
     "@playwright/experimental-ct-core": {
       "vite": "^6.3.4 || ^7.0.0"
+    },
+    "mdast-util-gfm": {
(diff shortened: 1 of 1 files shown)


ShGKme:
If I understand it correctly, corepack does not support array value for `packageManager` https://github.com/nodejs/corepack/blob/0b492c9c97e4b3d5e9a139eeedb931860127b470/sources/specUtils.ts#L89-L91

susnux:
@ShGKme yes and that is exactly the work around - we do not want `corepack` to "detect" that there is a value.
Because *corepack* then will throw an error as it only supports fixed versions and not version ranges.

So the work around it to make *corepack* ignore this - for a proper fix *corepack* needs to be fixed, see my PR on it.

ShGKme (approved):

PR #7218, opened 2025-07-29
feat(NcUserStatusIcon): update status icons

Antreesy:
### ☑️ Resolves

- Support to https://github.com/nextcloud/spreed/issues/15465
- Distinct supported 'Busy' status
- Refresh all other icons to circle-like

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

2025-07-29_14h29_29
2025-07-29_14h29_54

image
image

### 🚧 Tasks

- [ ] In-app screenshots

### 🏁 Checklist


src/components/NcUserStatusIcon/NcUserStatusIcon.vue
```diff
@@ -39,6 +50,7 @@ import { t } from '../../l10n.ts'
 
 import onlineSvg from '../../assets/status-icons/user-status-online.svg?raw'
 import awaySvg from '../../assets/status-icons/user-status-away.svg?raw'
+import busySvg from '../../assets/status-icons/user-status-busy.svg?raw'
 import dndSvg from '../../assets/status-icons/user-status-dnd.svg?raw'
 import invisibleSvg from '../../assets/status-icons/user-status-invisible.svg?raw'
```
susnux, line 55: what I ask myself always here is why not just import from mdi?

susnux (approved):
code looks good

ShGKme (approved):

src/components/NcUserStatusIcon/NcUserStatusIcon.vue
Antreesy: Fill color is currently forced inside of each svg.
We can do it inside of the component with mdi path, of course, but I already copied new svgs from Material symbols 👀

susnux: > Fill color is currently forced inside of each svg

Yes its more like expected that to happen in `NcUserStatusIcon` - but no worries thats completely unrelated to your PR ;)

Antreesy:
Another question - do we trying to scope it for v32?
Asking as it should end up on server, and thus would require backport to stable8 branch

jancborchardt (approved):
Looks good to me!

ShGKme:
> Another question - do we trying to scope it for v32?
> Asking as it should end up on server, and thus would require backport to stable8 branch

Until we say that nextcloud-vue@8 is not compatible with the current nextcloud/server, we should backport.

But it shouldn't be a problem, at it only adds support for a new value, so it doesn't do anythign with servers without Busy status support, do it?

Antreesy:
> so it doesn't do anything with servers without Busy status support

Nope, but general appearance of icons will change, as soon as someone bumps a library. Then users might be surprised to see new 'Online' and 'Away' icons

susnux:
> Nope, but general appearance of icons will change, as soon as someone bumps a library. Then users might be surprised to see new 'Online' and 'Away' icons

I do not think this is really a big issue - so I would just backport and use as is.

Rello:
@Antreesy where can I get the icons - with color?
we need to use the same for the clients. but the svgs here are all black. I want to ensure that everyone has the same icons

PR #7233, opened 2025-07-29
chore: migrate to ESLint v9

susnux:
### ☑️ Resolves

- TBD

### 🏁 Checklist


ShGKme:
Before PR is ready to review, could you split changes into smaller/simpler commits to make it easier to review? Especially if auto-fixed changes are mixed with manual changes. 👉👈

image

susnux:
> image

Mostly package lock

susnux:
@ShGKme fixed everything but one open question:
If there is `default` and `required: true`:
- remove `default`
- remove `required`

I did the first but that caused behavior changes, so I pushed a last commit reverting it for NcRichContenteditable to remove the `required` instead. What do you think about this?

l10n/messages.pot
```diff
@@ -295,7 +289,6 @@ msgstr ""
 msgid "offline"
 msgstr ""
 
-#. TRANSLATORS: A color name for RGB(165, 184, 114)
```
ShGKme: Is it an intentional removal?

ShGKme:
> @ShGKme fixed everything but one open question: If there is `default` and `required: true`:
> 
> * remove `default`
> * remove `required`
> 
> I did the first but that caused behavior changes, so I pushed a last commit reverting it for NcRichContenteditable to remove the `required` instead. What do you think about this?

I think that it is fine to remove `default` there. The prop was `required`, and the component doesn't make sense without having a model. Its only purpose is to input a value.

l10n/messages.pot
```diff
@@ -515,11 +507,10 @@ msgstr ""
 msgid "W"
 msgstr ""
 
-#. TRANSLATORS: A color name for RGB(211, 169, 103)
 msgid "Whiskey"
 msgstr ""
 
-msgid "Write a message …"
+msgid "Write a message …"
```
ShGKme, line 522: Are these changes necessary?
- It is a new translation string - requires updating all translations
- Does Transifex check that every translation also uses non-breaking space?

susnux: > Does Transifex check that every translation also uses non-breaking space?

Not sure if enforced on their side, but I know that this was discussed with the translators to always use nonbreaking space there if possible.

> It is a new translation string - requires updating all translations

Yes but in the meantime - as you can see - its already fixed here so its stays translated.

l10n/messages.pot
susnux: fixed

susnux:
Fixed the commits:
- f8fbc1defd2e2ba30593dda59aaf307ac8da3efe (remove default and adjust the then failing test case)
- 5efb9edc7a7e428887128989656bcc43e1c9d72f (correct indention of the colors to make the translators comments work)

susnux:
I will resolve the conflicts once reviewed, as otherwise this is a never ending fight with conflicts ;)

Antreesy (approved):
Should've split the auto-fixable changes as well, can't handle the diff =) But the rest seems good to go

ShGKme:
Also, it's useful to have the ESLint rule name in the commit message/description, it reverted it in the eslint config when migrating a large app. Then it is clear what rule does which large changes, and what changes are not eslint-related.

ShGKme:
> chore: adjust component attribute assignment

Could you describe the source of this change? Was it from some ESLint rule that is not auto-fixable?

src/utils/colors.ts
```diff
@@ -74,27 +74,56 @@ function mixPalette(steps: number, color1: Color, color2: Color): Color[] {
  * Like GenColor(4) but with labels
  */
 export const defaultPalette = [
-	new Color(182, 70, 157, t('Purple')),
-	new Color(191, 103, 139, t('Rosy brown'), // TRANSLATORS: A color name for RGB(191, 103, 139)
+	/* eslint-disable @stylistic/function-call-argument-newline -- Allow to put translators comments */
+	new Color(
+		182, 70, 157,
+		t('Purple'),
```
ShGKme: Why make it multiline?

tests/component/components/NcField/NcPassword.story.vue
```diff
@@ -17,8 +17,8 @@
 import NcPasswordField from '../../../../src/components/NcPasswordField/NcPasswordField.vue'
 
 withDefaults(defineProps<{
-	asText?: boolean
-	visible?: boolean
+	asText?: boolean | undefined
+	visible?: boolean | undefined
```
ShGKme, line 21: `asText?: boolean | undefined` is the same as `asText?: boolean`, isn't it?

tests/global-setup.js
```diff
@@ -7,5 +7,5 @@
  * Global test setup used to configure process environment.
  */
 export function setup() {
-	process.env.TZ = 'UTC'
+	globalThis.process.env.TZ = 'UTC'
```
ShGKme: Why?

l10n/messages.pot
```diff
@@ -14,7 +14,6 @@ msgstr ""
 msgid "a few seconds ago"
 msgstr ""
 
-#. TRANSLATORS: A color name for RGB(110, 166, 143)
```
ShGKme: Is it expected?

l10n/messages.pot
ShGKme: > Yes but in the meantime - as you can see - its already fixed here so its stays translated.

Does it update values in Transifex, if we change translations in the source?

tests/component/components/NcField/NcPassword.story.vue
Antreesy: `[key]?: type` and `[key]: type|undefined` should be different, as key might be present, but with undefined value?

ShGKme: > as key might be present, but with undefined value?

Component props always defined even when not passed (`undefined`).

> [key]?: type and [key]: type|undefined should be different

In TS - yes. But in vue compilation macros and `?:` defines non-required prop with `required: false`.

susnux: `asText?: boolean` -> The value is a boolean. Not set means false.
`asText?: boolean | undefined` -> The value can also be `undefined`.

In this case it is to forward the `undefined` meaning  the `NcPassword` default will be used.

susnux: For Vue it seems `prop?: Type` is the same as `prop: { type: Type, required: false }` and `prop?: Type|undefined` is the same as `prop: { type: Type|undefined, required: false }`

src/utils/colors.ts
susnux: This one? To make it consistent with the other.
Or all of them? To have the translators comment - its required to be in the same lineas the translation here to be extracted.
See previous - which is even worse as it had mixed newlines in function call

tests/global-setup.js
susnux: Silencing: 
grafik

Probably tests might be completely added as node files (thus they have node globals).

src/utils/colors.ts
susnux: It was multiline before to include the translators comment which is otherwise not possible.
With the previous one we would need to silence `@stylistic/function-paren-newline` but I think its a real issue how it looked before (only the closing parenthesis on a new line).
This one is cleaner IMHO.

So this was a compromise between the ugly hack before and the proper without any rules to disable, which would be:
```js
new Color(
	191,
	103,
	139,
	t('Rosy brown'), // TRANSLATORS: A color name for RGB(191, 103, 139)
),
```

Note that neither this:
```js
// TRANSLATORS: A color name for RGB(191, 103, 139)
new Color(191, 103, 139, t('Rosy brown'))
```

Not this works with l10n extraction:
```js
new Color(191, 103, 139, t('Rosy brown')) // TRANSLATORS: A color name for RGB(191, 103, 139)
```

l10n/messages.pot
susnux: Its just a problem of this comment see the ones after about the default color palette and my comment about the l10n extraction issue.
If I move this commit behind it works correctly - see pushed

susnux:
Only change in the push: moved 7b76ae987 behind all other commits which then makes it disappear within bbb08e4

l10n/messages.pot
susnux: You mean in the translations? No.
But it helps with keeping translations until the translators had time to adjust them - as otherwise all of those would be lost until they are re-translated just because of whitespace change

ShGKme:
Translations seems to be fixed now

ShGKme:
> > chore: adjust component attribute assignment
> 
> Could you describe the source of this change? Was it from some ESLint rule that is not auto-fixable?

Still interesting to know

l10n/messages.pot
ShGKme: Not only this:
image

tests/global-setup.js
ShGKme: > Silencing

As this is a configuration issue, not a problem in the code, we can silence it by `// eslint-ignore-next-line`

susnux:
- Rebased and resolved conflicts
- Use new eslint-config release where the only change is that we could drop some parts of our custom config here

tests/global-setup.js
susnux: Fixed with new release

tests/unit/components/NcTextArea/NcTextArea.spec.ts
```diff
@@ -53,6 +53,7 @@ describe('NcTextArea', () => {
 	it('should NOT have the label as the placeholder on Nextcloud 32', () => {
 		const wrapper = mount(NcTextArea, {
 			props: {
+				modelValue: '',
```
Antreesy, line 56: `chore: remove unused import` 👀

susnux: Damn wrong rebase it seems (BTW what this change does is adding the required props).
Should have been in 48de32e67fde11b7848375a692e5dc69aed17356

PR #7235, opened 2025-07-30
[stable8] fix: migrate from _oc_config to initial state

Antreesy:
### ☑️ Resolves

- Fix checking values on non-web clients (Talk Desktop)

### 🏁 Checklist

### DIFF
--- a/src/components/NcListItem/NcListItem.vue
+++ b/src/components/NcListItem/NcListItem.vue
@@ -535,9 +535,13 @@ The `actions-icon` slot can be used to pass icon to the inner NcActions componen
 </template>
 
 <script>
+import { loadState } from '@nextcloud/initial-state'
 import NcActions from '../NcActions/index.js'
 import NcCounterBubble from '../NcCounterBubble/index.js'
 import NcVNodes from '../NcVNodes/index.js'
+
+const [major] = loadState('core', 'config', { version: '30.0' }).version.split('.', 2) ?? []
+const isLegacy = major && Number.parseInt(major) < 30
 
 export default {
 	name: 'NcListItem',
@@ -694,9 +698,6 @@ export default {
 	],
 
 	setup() {
-		const [major] = window._oc_config?.version.split('.', 2) ?? []
-		const isLegacy = major && Number.parseInt(major) < 30
-
 		return {
 			isLegacy,
 		}
--- a/src/components/NcSettingsSection/NcSettingsSection.vue
+++ b/src/components/NcSettingsSection/NcSettingsSection.vue
@@ -63,9 +63,14 @@ This component is to be used in the settings section of nextcloud.
 </template>
 
 <script>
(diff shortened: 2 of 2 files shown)


Antreesy:
> to have a single shared `getNextcloudVersion` util

Sounds good, but I'm not sure it's worth to add it for 3 places in stable version, if it's not used on main branch

ShGKme:
> > to have a single shared `getNextcloudVersion` util
> 
> Sounds good, but I'm not sure it's worth to add it for 3 places in stable version, if it's not used on main branch

We wanted to add it at least for status icons, didn't we?

Antreesy:
Status icons and Vue3 are officially Nextcloud 32+ only, so there is no checks on main branch

ShGKme:
> Status icons and Vue3 are officially Nextcloud 32+ only, so there is no checks on main branch

But `main` is for Nextcloud 30+

susnux (commented):
I am not sure this is better.
`loadState` always adds a small amount of parsing overhead - in this case for every section and every list item.

Also I do not consider `loadState` a public API, its still a private API of the app that provides this (its rather a communication channel from backend to *your* frontend).

It would be better to have a *capability* for this, as capabilities are considered somewhat public API.

susnux:
> But main is for Nextcloud 30+

Probably 31+ as 30 is EOL when a first stable is released :sweat_smile: 

> [...] Vue3 [is] officially Nextcloud 32+ only

Why should it? There are already apps for 31 with Vue 3.
Just because there is no Nextcloud supported app it does not mean this library would only support Vue 3 on 32.

While I think it could make sense to make things much easier (no fallbacks for older designs),
there is no other reason to not support the currently active version of Nextcloud as long as it is not EOL.
(Thats why I mentioned the 31 thing - 30 will be EOL in September but 31 will be there for a while).

ShGKme:
> if it fixed a Talk Desktop issue.

For the Talk Desktop we can fix it on the Talk Desktop side by defining `_oc_config`

> 1. use capability over initial-state

There is no capability with the server version :(

ShGKme:
> in this case for every section and every list item.

We can:
- Check the version once in module scope (with lazy) instead of `setup`
- Add caching to `loadState`

susnux:
> There is no capability with the server version :(

yes thats why I said follow up - would be needed to create first.

> Check the version once in module scope (with lazy) instead of setup

I think this makes sense in the scope of this PR

Antreesy:
Moved out both from setup/computed to module level, should do the job for now.

This introduced a change for `NcSettingsSection` - if initialstate.core.config.version for some reason would be `undefined` or `' '`, this would be resolved as non-legacy now, limiting the width. Don't think it's breaking though

ShGKme:
If this is only for Talk Desktop - we can define `_oc_config` there

PR #7237, opened 2025-07-30
fix(NcBreadCrumb): correctly pass container for `NcActions`

susnux:
### ☑️ Resolves
I am not sure how this was supposed to work, as it set a random attribute on the `li` tag - that was sometimes not even a valid HTML attribute name.
We can instead just use the ref to the li element directly.

### 🏁 Checklist

### DIFF
--- a/src/components/NcBreadcrumb/NcBreadcrumb.vue
+++ b/src/components/NcBreadcrumb/NcBreadcrumb.vue
@@ -16,7 +16,7 @@ Renders a button element when given no redirection props, otherwise, renders <a/
 	<li ref="crumb"
 		class="vue-crumb"
 		:class="[{'vue-crumb--hovered': hovering}, $props.class]"
-		:[crumbId]="''"
+		:data-crumb-id="crumbId"
 		draggable="false"
 		@dragstart.prevent="() => {/** Prevent the breadcrumb from being draggable. */}"
 		@drop.prevent="dropped"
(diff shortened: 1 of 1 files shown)


src/components/NcBreadcrumb/NcBreadcrumb.vue
```diff
@@ -44,7 +43,7 @@ Renders a button element when given no redirection props, otherwise, renders <a/
 			:menu-name="name"
 			:title="title"
 			:force-name="true"
-			:container="`.vue-crumb[${crumbId}]`"
+			:container="$refs.crumb"
```
ShGKme: But at the moment of rendering `$refs` is empty...

susnux: Too expensive to re-render the `NcActions` in the second iteration when `refs` is populated?

Otherwise can just use a proper attribute for this, but the current state can never really work.

susnux: fixed

ShGKme (approved):

src/components/NcBreadcrumb/NcBreadcrumb.vue
```diff
@@ -39,12 +39,12 @@ Renders a button element when given no redirection props, otherwise, renders <a/
...
 			ref="actions"
+			:container="actionsContainer"
 			:force-menu="forceMenu"
+			force-name
 			:open="open"
 			:menu-name="name"
-			:title="title"
-			:force-name="true"
-			:container="`.vue-crumb[${crumbId}]`"
+			:title
```
skjnldsv, line 47: 🤔

ShGKme: @skjnldsv Could you elaborate on the problem? 🤔

susnux: Yes not 100% related but as I am on this part I just wanted to also apply some minor refactoring and [use property shorthand](https://vuejs.org/guide/essentials/template-syntax#same-name-shorthand).

skjnldsv: I almost thought GitHub mobile was wrongly rendering the diff.
Alright, did not know the shortand was a thing! Not sure if I like it 🫣

Good for me then

Ps: here's what it looks like on mobile lol
I think it looks very odd because it's the only property like this in this element block

skjnldsv (approved):

PR #7239, opened 2025-07-30
fix: `Set operation on key "value" failed: target is readonly`

ShGKme:
Vue warns in some edge cases when a ref in a component instance and template ref key string have the same name.

- Ref: https://github.com/vuejs/core/issues/12852

### ☑️ Resolves

- Fix https://github.com/nextcloud-libraries/nextcloud-vue/issues/7238

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppNavigation/NcAppNavigation.vue
+++ b/src/components/NcAppNavigation/NcAppNavigation.vue
@@ -182,7 +182,7 @@ defineSlots<{
 let focusTrap: FocusTrap
 const setHasAppNavigation = inject<(v: boolean) => void>('NcContent:setHasAppNavigation', () => warn('NcAppNavigation is not mounted inside NcContent, this is probably an error.'), false)
 
-const appNavigationContainer = useTemplateRef('appNavigationContainer')
+const appNavigationContainer = useTemplateRef('app-navigation-container-key')
 const isMobile = useIsMobile()
 const open = ref(!isMobile.value)
 
@@ -275,7 +275,7 @@ function handleEsc(): void {
 </script>
 
 <template>
-	<div ref="appNavigationContainer"
+	<div ref="app-navigation-container-key"
 		class="app-navigation"
 		:class="{'app-navigation--closed':!open }">
 		<nav id="app-navigation-vue"
--- a/src/components/NcInputField/NcInputField.vue
+++ b/src/components/NcInputField/NcInputField.vue
@@ -158,7 +158,7 @@ defineSlots<{
 
 const attrs = useAttrs()
 
-const input = useTemplateRef('input')
+const input = useTemplateRef('input-key')
 
 const hasTrailingIcon = computed(() => props.showTrailingButton || props.success)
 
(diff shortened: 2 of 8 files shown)


src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -275,7 +275,7 @@ function handleEsc(): void {
 </script>
 
 <template>
-	<div ref="appNavigationContainer"
+	<div ref="app-navigation-container-key')"
```
susnux:
```suggestion
	<div ref="app-navigation-container-key"
```

src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -182,7 +182,7 @@ defineSlots<{
 let focusTrap: FocusTrap
 const setHasAppNavigation = inject<(v: boolean) => void>('NcContent:setHasAppNavigation', () => warn('NcAppNavigation is not mounted inside NcContent, this is probably an error.'), false)
 
-const appNavigationContainer = useTemplateRef('appNavigationContainer')
+const appNavigationContainer = useTemplateRef('app-navigation-container-key')
```
susnux, line 185: shall we just kebab case until the vue fix is merged (its already "read-to-merge") to not introduce new naming convention with `-key` suffix?

susnux (commented):
changes make sense - but there is a typo

src/components/NcAppNavigation/NcAppNavigation.vue
ShGKme: It doesn't fix the problem with single-word keys.

src/components/NcAppNavigation/NcAppNavigation.vue
ShGKme: Oops

src/components/NcAppNavigation/NcAppNavigation.vue
ShGKme: I don't insist on this style specifically. We have different styles anyway.

API Reference, for example, adds `*Ref` suffix to the variable name. but I don't like to add types to variable names :D 

https://vuejs.org/guide/essentials/template-refs.html#accessing-the-refs

src/components/NcInputField/NcInputField.vue
```diff
@@ -158,7 +158,7 @@ defineSlots<{
 
 const attrs = useAttrs()
 
-const input = useTemplateRef('input')
+const input = useTemplateRef('input-key')
```
Antreesy, line 161: I think we might use some of these internal refs in apps
Although it's not correct, change should be mentioned somewhere?

ShGKme: I want to say "it is not a part of Public API, if it fails in an app - it's the app's problem".

but we can rename a variable instead of renaming a key.

susnux: Its not public api. If you need something nested then we should add probs or exposed methods. 

So IMHO its fine to rename the key and better readable than renaming vars

PR #7247, opened 2025-08-01
ci: sync stable branch translations

susnux:
### ☑️ Resolves

Only main branch will be localized by Transifex.
To also update the stable branch with translations we need to synchronize them.

This actions does following:
- clone both branches
- combine po files from both branches for each language
- check the stable branch po template which messages are needed
- copy needed messaged from the combined po file (use the main if the translation differ)

This works locally, but I am not sure if the msgcat and msgmerge commands are installed on the runner.

### 🏁 Checklist

### DIFF
--- a/.github/workflows/sync-l10n.yml
+++ b/.github/workflows/sync-l10n.yml
@@ -1,0 +1,60 @@
+# SPDX-FileCopyrightText: 2021-2024 Nextcloud GmbH and Nextcloud contributors
+# SPDX-License-Identifier: MIT
+
+name: Sync stable branch translations
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 5 * * *"
+
+permissions:
+  contents: read
+
+jobs:
+  server-styling-update:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout main
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          path: main
+
+      - name: Checkout stable branch
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          path: stable-branch
+          ref: stable8
+
+      - name: Merge translations
+        run: |
+          cd "$GITHUB_WORKSPACE"
+          # combine entries
+          mkdir merged-entries
+          for f in stable-branch/l10n/*.pot; do
+            [ "$f" = "messages.pot" ] && continue
+            msgcat -o "merged-entries/$(basename $f)" --use-first "main/l10n/$(basename $f)" "$f"
+          done
+          # update the stable one with all needed translations
+          for f in merged-entries/*.pot; do
+            msgmerge -N "$f" "stable-branch/l10n/messages.pot" > "stable-branch/l10n/$(basename $f)"
+          done
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@271a8d0340265f705b14b6d32b9829c1cb33d45e # v7.0.8
+        with:
+          path: stable-branch
+          token: ${{ secrets.COMMAND_BOT_PAT }}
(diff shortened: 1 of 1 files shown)


ShGKme:
> To also update the stable branch with translations we need to synchronize them

Isn't it supposed to be covered by the bot (`.tx/backports` file)?

susnux:
> Isn't it supposed to be covered by the bot (`.tx/backports` file)?

I did not find any information on this :thinking:

skjnldsv (approved):

ShGKme:
I mean this:
https://docs.nextcloud.com/server/latest/developer_manual/app_development/translation_setup.html#branch-selection-tx-backport

susnux:
> I mean this: https://docs.nextcloud.com/server/latest/developer_manual/app_development/translation_setup.html#branch-selection-tx-backport

But this is only for apps, we do not use the app translations but the native transifex integration

.github/workflows/sync-l10n.yml
```diff
@@ -0,0 +1,60 @@
...
+          for f in merged-entries/*.pot; do
+            msgmerge -N "$f" "stable-branch/l10n/messages.pot" > "stable-branch/l10n/$(basename $f)"
+          done
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@271a8d0340265f705b14b6d32b9829c1cb33d45e # v7.0.8
+        with:
+          path: stable-branch
+          token: ${{ secrets.COMMAND_BOT_PAT }}
+          commit-message: "chore(assets): Update styling from server"
```
nickvergessen: 🙈

nickvergessen (commented):
> But this is only for apps, we do not use the app translations but the native transifex integration

That's correct, and iirc, the sync simply has a branch list on the config, but i never tried it.

.github/workflows/sync-l10n.yml
susnux: you see where I copied the base :facepalm: 

```suggestion
          commit-message: "chore(l10n): sync translations with main branch"
```

susnux:
@nickvergessen yes you can add branches for Transifex, but if I understood that correctly it will create new resources for translators. That is not what we want - no?
Because in this specific case we want to sync main to stable8 as in theory they should have the same strings (stable 8 is just the Vue 2 version of main - feature wise and thus string wise (code wise there are more changes)).

ShGKme:
> But this is only for apps, we do not use the app translations but the native transifex integration

Does it mean we need it globally in `nextcloud-libraries/.github` as it is not `nextcloud/vue` specific proble?

nickvergessen:
We barely add and remove language strings here, so I think for now I will test if simply adding the branch works and does not create a new resource, we'll do that or manually copy the language files one time afterwards

susnux:
> Does it mean we need it globally in `nextcloud-libraries/.github` as it is not `nextcloud/vue` specific proble?

How many libraries use translations?
And then how many libraries use translations and have multiple supported branches?

ShGKme:
> How many libraries use translations?

At least 4

> And then how many libraries use translations and have multiple supported branches?

At least 2

ShGKme (approved):

nickvergessen (changes requested):
Unless there is immediate need to fix this due to customer demand, please let this sit until after my vacation, so we can first try if it works with transifex directly.

nickvergessen:
So after checking it seems the inline feature of Transifex is focusing on feature Pull requests, not on "stable branches". So copying them to/from main sounds like the best idea. Whether we need to merge them or not is another question. We can give translation folders instead of files, so each branch could have their own files:
https://help.transifex.com/en/articles/6265125-github-installation-and-configuration#h_b827747da8

So we could have a cron action that copies the PO files from stable to main and the POT from main to the stables?

susnux:
> So we could have a cron action that copies the PO files from stable to main and the POT from main to the stables?

But the translations are not necessarily the same. That is why I needed to merge them

susnux:
seems like this is not wanted

PR #7262, opened 2025-08-06
[stable8] chore: sync translations from master branch

susnux:
### ☑️ Resolves

```sh
mkdir merged-entries
for f in stable8/l10n/*.pot; do
    [ $(basename "$f") = "messages.pot" ] && continue
    msgcat -o "merged-entries/$(basename $f)" --use-first "main/l10n/$(basename $f)" "$f"
done

# update the stable one with all needed translations
for f in merged-entries/*.pot; do
    msgmerge -N "$f" "stable8/l10n/messages.pot" | sed '/^#~/d' | cat -s  > "stable8/l10n/$(basename $f)"
done
```

see also #7247 - not that many active translations thus best to check with German for success of this operation.

### DIFF
--- a/l10n/af.pot
+++ b/l10n/af.pot
@@ -1,13 +1,14 @@
-# 
+#
 # Translators:
 # Transifex Bot <>, 2023
-# 
+#
 msgid ""
 msgstr ""
 "Last-Translator: Transifex Bot <>, 2023\n"
-"Language-Team: Afrikaans (https://app.transifex.com/nextcloud/teams/64236/af/)\n"
-"Content-Type: text/plain; charset=UTF-8\n"
+"Language-Team: Afrikaans (https://app.transifex.com/nextcloud/teams/64236/"
+"af/)\n"
 "Language: af\n"
+"Content-Type: text/plain; charset=UTF-8\n"
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
 
 msgid "{tag} (invisible)"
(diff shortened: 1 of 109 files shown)


Antreesy:
I don't see it fixing 'Keyboard shortcuts' for me. msgid and msgstr pair are identical for ru.pot I'm checking, just sorted alphabetically:

image

Example copied from styleguide, so maybe there's another issue I don't see

UPD: Or it's the same issues we've seen before, some chunks in _l10n-494SYRqq.mjs are missing:
image

susnux:
> I don't see it fixing 'Keyboard shortcuts' for me. msgid and msgstr pair are identical for ru.pot I'm checking, just sorted alphabetically

No for keyboard shortcuts you PR should have already covered it.
But yours included some string not available in stable8 and also had the wrong space (see that in the diff this one adds the non breaking space before the ellipsis).
Best to check well translated languages like de or es.
E.g. check `Message limit of {count} characters reached` in the diff or `more items`

Antreesy:
Yes, they weren't replaced, only new ones were added to the end of file.
So I have both `Write a message ...` and `Write a message&nbsp;...`, and second one was working, as it matches the key; whereas first one was redundant

> Best to check well translated languages like de or es.

Pretty sure it's not the language coverage issue
image

susnux:
> I don't see it fixing 'Keyboard shortcuts' for me. msgid and msgstr pair are identical for ru.pot I'm checking, just sorted alphabetically:

Oh yes you are right, there was a bug in the l10n plugin :exploding_head: 
https://github.com/nextcloud-libraries/nextcloud-vue/pull/8037

ShGKme:
But I'd rerun the script

Antreesy:
> But I'd rerun the script

7 updated lines in fr.pot only

PR #7084, opened 2025-06-24
fix(NcModal): correctly handle when trying to activate non-existing focus-trap

Antreesy:
### ☑️ Resolves

- Fix handling of focus trap stack, if there are several
- Fix issue when trying to activate non-existing focus-trap
- Copied form NcEmojiPicker
- Tested in Talk

### 🖼️ Screenshots

Before

image

After

https://github.com/user-attachments/assets/8b5ab329-67ea-4ed3-b747-fa894f86e91d

### 🏁 Checklist

- [ ] ⛑️ Tests are included or are not applicable
- [ ] 📘 Component documentation has been extended, updated or is not applicable
- [x] 2️⃣ Backport to `stable8` for maintained Vue 2 version or not applicable

PR #7277, opened 2025-08-12
fix: auto-resolve focus trap stack, if was changed outside of controller

Antreesy:
### ☑️ Resolves

- Follow-up to #7084
  - NcModal is working fine with default trap stack, so there's no need for `useTrapStackControl`
- To be backported with #7085
  - Does not change behaviour for main / Vue 3
  - Fixes following scenario for Vue 2:
    - Talk app -> shrink to mobile -> NcAppNavigation creates trap 0 -> trap 0 is active
    - open NcActions next to search bar -> pause trap 0
    - select 'Create open conversation' -> starts to mount NcModal
    - NcModal mounted -> useFocusTrap() -> creates trap 1 -> trap 1 is active
    - NcActions with `close-after-click` is closed -> unpause pausedStack (it has only [trap 0] from first step) -> trap 0 unpauses
    - trap 0 activates - trap 1 is paused -> focus trap is now in NcAppNavigation instead of NcModal
  - If trap stack was modified outside of stackcontroller, it should be safe to assume, that user action lead to that, and it's currently correctly resolved

### 🖼️ Screenshots

🏚️ Before (Vue 2 at #7085)

https://github.com/user-attachments/assets/037b6a8d-f2f4-4cab-b48f-2d06882e965e

🏡 After (Vue 2)

https://github.com/user-attachments/assets/44e08243-a4d2-4ae9-a4cd-6b68bdc5d761

### 🏁 Checklist

### DIFF
--- a/src/utils/focusTrap.ts
+++ b/src/utils/focusTrap.ts
@@ -45,10 +45,14 @@ export function createTrapStackController() {
 		},
 		/**
 		 * Unpause the paused focus trap stack
+		 * If the actual stack is different from the paused one, there were changes
+		 * outside of this controller, so we assume it's self-regulated and do not unpause.
 		 */
 		unpause() {
-			for (const trap of pausedStack) {
-				trap.unpause()
+			if (pausedStack.length === getTrapStack().length) {
+				for (const trap of pausedStack) {
+					trap.unpause()
+				}
 			}
 			pausedStack = []
 		},
--- a/src/components/NcModal/NcModal.vue
+++ b/src/components/NcModal/NcModal.vue
@@ -345,7 +345,6 @@ import Timer from '../../utils/Timer.js'
 import Close from 'vue-material-design-icons/Close.vue'
 import Pause from 'vue-material-design-icons/Pause.vue'
 import Play from 'vue-material-design-icons/Play.vue'
-import { useTrapStackControl } from '../../composables/useTrapStackControl.js'
 
 export default {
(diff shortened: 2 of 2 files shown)


susnux (approved):
Code makes sense to me

susnux:
this one can be backported to fix the stable8 issue, right?

src/utils/focusTrap.ts
```diff
@@ -45,10 +45,14 @@ export function createTrapStackController() {
 		},
 		/**
 		 * Unpause the paused focus trap stack
+		 * If the actual stack is different from the paused one, there were changes
+		 * outside of this controller, so we assume it's self-regulated and do not unpause.
```
ShGKme: Can you add an explanation here? Why we only need to unpause if the length is the same

Antreesy: It's the opposite of what is already written, do we need additional explanation here? I currently don't see how it could be expanded, please, advise

susnux: Maybe just align with code?

```suggestion
		 * Only if the actual stack is the same as the paused one, there were no changes
		 * outside of this controller, so we can save assume it's not self-regulated and we need to unpause.
```

But I also think the current comment is already explaining it well

Antreesy: Will put PR description into composable comment, so devs can read and act from there at least. As it's not blocking, we should be able to follow-up with better explanation

PR #7280, opened 2025-08-13
fix(NcIconSvgWrapper): do not restrict icon minimal size

Antreesy:
### ☑️ Resolves

- Upstream to https://github.com/nextcloud/spreed/pull/15662

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/NcIconSvgWrapper/NcIconSvgWrapper.vue
+++ b/src/components/NcIconSvgWrapper/NcIconSvgWrapper.vue
@@ -254,8 +254,8 @@ const cleanSvg = computed(() => {
 	display: flex;
 	justify-content: center;
 	align-items: center;
-	min-width: var(--default-clickable-area);
-	min-height: var(--default-clickable-area);
+	min-width: v-bind('iconSize');
+	min-height: v-bind('iconSize');
 	opacity: 1;
 
 	&#{&}--inline {


DorraJaouad (approved):

Antreesy:
> the icon should be by default the default clickable area

But prop description states:

        /**
         * Set if the icon should be used as inline content e.g. within text.
         * By default the icon is made a block element for use inside `icon`-slots.
         */

Almost all the icon slots I can think of are present in elements, which size is already limited by 34px (buttons, input fields, notes, e.t.c.). So icon should not exceed this?

susnux:
Yes exactly if not set the element is a block element with the size of the icon slot which should always be the default clickable area. `inline` is to allow any arbitrary size (real size of the icon).

Antreesy:
Ok, it covers our specific case with non-standard icon wrapper. But IMO it's still confusing to ignore `size` prop, when it's passed.

susnux:
> Ok, it covers our specific case with non-standard icon wrapper. But IMO it's still confusing to ignore size prop, when it's passed.

You should just use `icon` prop - it was explicitly added for cases like you now have - no need for custom patches.

The icon size only controls the size of the icon itself - not of the wrapper. The wrapper for legacy reasons (or better for legacy reasons of all components that have a `icon` slot) needed to always have a fixed size of clickable area.
Thus we could not change that - instead we added the `inline` slot for this -> the icon no longer has a wrapper with padding but its just the specified icon size.

PR #7283, opened 2025-08-14
fix(NcInputField, NcTextareaField): remove placeholder on Nextcloud 32+

susnux:
### ☑️ Resolves

- Resolves https://github.com/nextcloud-libraries/nextcloud-vue/issues/7281

### 🏁 Checklist

### DIFF
--- a/src/components/NcInputField/NcInputField.vue
+++ b/src/components/NcInputField/NcInputField.vue
@@ -25,6 +25,7 @@ import { computed, useAttrs, useTemplateRef, warn } from 'vue'
 import NcButton from '../NcButton/index.ts'
 import NcIconSvgWrapper from '../NcIconSvgWrapper/index.ts'
 import { createElementId } from '../../utils/createElementId.ts'
+import { isLegacy } from '../../utils/legacy.ts'
 
 export interface NcInputFieldProps {
 	/**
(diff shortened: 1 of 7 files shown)


src/components/NcInputField/NcInputField.vue
```diff
@@ -162,6 +163,8 @@ const input = useTemplateRef('input-key')
 
 const hasTrailingIcon = computed(() => props.showTrailingButton || props.success)
 
+const internalPlaceholder = computed(() => props.placeholder || (isLegacy ? props.label : ''))
```
Antreesy: I feel it should be like this, to not have an empty attribute
```suggestion
const internalPlaceholder = computed(() => props.placeholder || (isLegacy ? props.label : undefined))
```

src/utils/legacy.ts
```diff
@@ -0,0 +1,6 @@
+/*
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+export const isLegacy = Number.parseInt(window.OC?.config.version.split('.')[0] ?? '0') < 32
```
Antreesy: That doesn't look future-proof, esp. if we'll make some other changes for Nextcloud 33 =)

Maybe something more specific?
```suggestion
function isLegacyFor(majorVersion: number): boolean {
	return Number.parseInt(window.OC?.config.version.split('.')[0] ?? '0') < majorVersion
}

export const isLegacy32 = isLegacyFor(32)
```

Antreesy (approved):
Otherwise good

src/utils/legacy.ts
ShGKme: As this is a completely internal function, I'd say it's fine to adjust it when we hav this problem.

src/components/NcInputField/NcInputField.vue
susnux: makes sense!

susnux:
/backport f2a85ca133ed673f436d29890114e921430a5e67 e4cc209262502196faf959ea50e3cc4ce5621f4a to stable8

PR #7287, opened 2025-08-15
docs: v9 will only provide visual compatibility with Nextcloud 32+

susnux:
### ☑️ Resolves

- Due to recent design changes this simplifies maintaining this new major version - especially as no stable of v9 was released yet.

### 🏁 Checklist


ShGKme:
So we keep existing Vue 3 app with different style on older servers, right?

susnux:
> So we keep existing Vue 3 app with different style on older servers, right?

Yes could be framed like this.
32 is the first version that officially supports Vue 3. Thus also official apps were merged for 32+.

PR #7300, opened 2025-08-18
refactor(NcButton, NcNoteCard): use status text color for icons and text

susnux:
### ☑️ Resolves

- for https://github.com/nextcloud/server/pull/54439


susnux:
> I'm not sure if changing those to be darker would be part of this PR? If yes though, then they should be.

No its defined in server - also its already fixed there, the variables are just outdated here.

susnux:
> the variables are just outdated here

Updated now, thus you can recheck if you like

PR #7311, opened 2025-08-19
fix: adjust color variable to work with Nextcloud 32+

susnux:
### ☑️ Resolves

Due to new secondary style of colors we need to use the new variables for colored text / borders instead which provide good contrast.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist


src/assets/status-icons/user-status-away.svg
```diff
@@ -4,6 +4,6 @@
 -->
 <svg viewBox="0 -960 960 960" width="24px" height="24px" xmlns="http://www.w3.org/2000/svg">
 	<path
-		fill="var(--color-warning)"
+		fill="var(--color-favorite)"
```
Antreesy, line 7: Maybe also leave a fallback here?
```suggestion
		fill="var(--color-favorite, var(--color-warning))"
```

kra-mo (approved):

skjnldsv (approved):

src/assets/status-icons/user-status-away.svg
susnux: It would not be used as it exists also for old versions.
The difference is that favorite always tries to look more yellow while warning can get somekind of brownish depending on the color schema used.

PR #7313, opened 2025-08-19
fix(utils): legacy version detection

skjnldsv:
See https://github.com/nextcloud/server/pull/54497#issuecomment-3200538375
"Regression" from https://github.com/nextcloud-libraries/nextcloud-vue/pull/7283

### DIFF
--- a/src/utils/legacy.ts
+++ b/src/utils/legacy.ts
@@ -3,4 +3,5 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-export const isLegacy = Number.parseInt(window.OC?.config.version.split('.')[0] ?? '32') < 32
+const version = window.OC?.config?.version?.split('.')[0] || '32.0.0'
+export const isLegacy = Number.parseInt(version) < 32


susnux:
This is fine - server just needs proper mocks in tests.

src/utils/legacy.ts
```diff
@@ -3,4 +3,5 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-export const isLegacy = Number.parseInt(window.OC?.config.version.split('.')[0] ?? '32') < 32
+const version = window.OC?.config?.version || '32.0.0'
```
susnux: Should not be needed - this is guaranteed to be available.
You just need proper global state mocks.

src/utils/legacy.ts
```diff
@@ -3,4 +3,5 @@
...
+export const isLegacy = Number.parseInt(version?.split('.')[0] ?? '32') < 32
```
susnux: But in case we want to be more graceful then keep the above and then we do not need the fallback here

```suggestion
export const isLegacy = Number.parseInt(version?.split('.')[0]!) < 32
```

susnux (approved):

ShGKme (approved):

src/utils/legacy.ts
skjnldsv: Assuming properties to be present isn't great, if you also really trusted it, you would not have added a "32" fallback in the first place :wink: 

I think better safe than sorry :+1:

src/utils/legacy.ts
skjnldsv: Not even needed.
It actually even complains: "Optional chain expressions can return undefined by design - using a non-null assertion is unsafe and wrong.eslint[@typescript-eslint/no-non-null-asserted-optional-chain](https://typescript-eslint.io/rules/no-non-null-asserted-optional-chain)"

src/utils/legacy.ts
```diff
@@ -3,4 +3,5 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-export const isLegacy = Number.parseInt(window.OC?.config.version.split('.')[0] ?? '32') < 32
+const version = window.OC?.config?.version || '32.0.0'
+export const isLegacy = Number.parseInt(version?.split('.')[0]) < 32
```
ShGKme:
```suggestion
export const isLegacy = Number.parseInt(version?.split('.')[0]!) < 32
```

src/utils/legacy.ts
skjnldsv: And now it complains again :grin: 
`error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'string'.`

Adding the fallback back

src/utils/legacy.ts
skjnldsv: See previous comment

src/utils/legacy.ts
```diff
@@ -3,4 +3,5 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-export const isLegacy = Number.parseInt(window.OC?.config.version.split('.')[0] ?? '32') < 32
+const version = window.OC?.config?.version?.split('.')[0] || '32.0.0'
```
ShGKme, line 6: Semantically, it looks a bit weird that we use a float value as a fallback for an integer value

ShGKme:
```suggestion
const version = window.OC?.config?.version?.split('.')[0] || '32'
```

susnux: Damn yes this should be applied either before the split or just without the dot versions

skjnldsv: yes https://github.com/nextcloud-libraries/nextcloud-vue/pull/7314

PR #7315, opened 2025-08-19
[stable8] fix(utils): legacy check syntax

skjnldsv:
Backport of https://github.com/nextcloud-libraries/nextcloud-vue/pull/7313

### DIFF
--- a/src/utils/legacy.ts
+++ b/src/utils/legacy.ts
@@ -1,6 +1,7 @@
-/*
+/**
  * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-export const isLegacy32 = Number.parseInt(window.OC?.config.version.split('.')[0] ?? '0') < 32
+const version = window.OC?.config?.version?.split('.')[0] || '32'
+export const isLegacy32 = Number.parseInt(version) < 32


src/utils/legacy.ts
```diff
@@ -1,6 +1,7 @@
-/*
+/**
```
susnux, line 1: Its not really a doc block so lately we did not use the second `*` to prevent issues with doc extraction - but :shrug:

susnux (approved):

src/utils/legacy.ts
skjnldsv: Ahaha, fair question!
It's old habit because of the `@copyright` we used to use.

BUT, if this wasn't a doc block, we would not need to prefix each row with `*`.

I don't know when some of us started removing the second wildcard, but it turned the header block into a half comment-block/half doc-block

image

skjnldsv: Valid multi-line comment block
```js
/*
SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
SPDX-License-Identifier: AGPL-3.0-or-later
*/
```

Valid doc block header
```js
/**
 */
```

Weird hybrid of the two
```js
/*
 */
```

susnux: For license like header

```
/*!
 */
 ```

Its somewhat official - I do not know its originally from webpacks `legalComments` config option.
But also used in `esbuild` and `terser`.

I really do not have a strong opinion here - especially in the library it does not matter that much, but in API libraries the license header often gets confused with the doc block for the whole module :see_no_evil:

Antreesy (approved):

PR #7329, opened 2025-08-20
fix: icon-collapse has wrong colour

GretaD:
### ☑️ Resolves

- Ref # 

### 🖼️ Screenshots

the problem: 

Screenshot from 2025-08-20 14-26-12

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist


src/components/NcAppNavigationItem/NcAppNavigationIconCollapsible.vue
```diff
@@ -62,10 +62,10 @@ export default {
 .button-vue.icon-collapse {
 	position: relative;
 	z-index: 105; // above a, under button
-	color: var(--color-main-text);
+	color: var(--color-primary-element-text);
 	inset-inline-end: 0;
 	&--open {
-		color: var(--color-main-text);
+		color: var(--color-primary-element-text);
```
susnux: But only if it is active?
Because otherwise the colors can be different - IMHO this needs the active state passed from the ncappnavigationitem parent or we do not set a color should also work if we set it in the parent.

src/components/NcAppNavigationItem/NcAppNavigationIconCollapsible.vue
```diff
@@ -62,10 +62,10 @@ export default {
...
 		&:hover{
 			color: var(--color-primary-element);
```
susnux: See also this color - it invisible then on hover?

src/components/NcAppNavigationItem/NcAppNavigationIconCollapsible.vue
GretaD: you're right

src/components/NcAppNavigationItem/NcAppNavigationIconCollapsible.vue
```diff
@@ -39,6 +40,10 @@ export default {
 			type: Boolean,
 			default: true,
 		},
+		active: {
+			type: Boolean,
+			default: false,
```
susnux: I think from the usage this should be rather:
```suggestion
			required: true,
```

src/components/NcAppNavigationItem/NcAppNavigationIconCollapsible.vue
```diff
@@ -66,10 +71,13 @@ export default {
 	inset-inline-end: 0;
 	&--open {
 		color: var(--color-main-text);
-		&:hover{
+		&:hover {
 			color: var(--color-primary-element);
 		}
 	}
+	&--active {
+		color: var(--color-primary-element-text);
```
susnux: have you checked if the hover color works as well?

src/components/NcAppNavigationItem/NcAppNavigationIconCollapsible.vue
```diff
@@ -6,7 +6,8 @@
 <template>
 	<!-- Button to expand or collapse children -->
 	<NcButton class="icon-collapse"
-		:class="{'icon-collapse--open':open}"
+		:class="{'icon-collapse--open':open,
+			'icon-collapse--active':active}"
```
susnux:
```suggestion
		:class="{
			'icon-collapse--active': active,
			'icon-collapse--open': open,
		}"
```

susnux (commented):
Looks good! Just had no time to test the case mentioned in the comment

src/components/NcAppNavigationItem/NcAppNavigationIconCollapsible.vue
GretaD: no, i will

ShGKme (approved):

PR #7330, opened 2025-08-20
fix(NcPopover): scope styles to avoid leak

ShGKme:
### ☑️ Resolves

- Missed styles scoping
  - The difference between v8 and v9 already breaks each other due to `box-sizing` change
  - Example of broken apps: text
- Also changing `theme` name to prevent influence of older library versions
  - Otherwise unscoped classes are still the same
- Using CSS Modules to scope styles on a custom class
  - `<style scoped>` is not available here due to floating-vue implementation

### 🏁 Checklist

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -1947,7 +1947,7 @@ export default {
 <style lang="scss">
 // We overwrote the popover base class, so we can style
 // the popover__inner for actions only.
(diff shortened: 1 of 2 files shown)


max-nextcloud (approved):
This removed the styles leaked from notifications to text.

I did not check if it does not break notifications.

ShGKme:
Added theme to the `options`

PR #7349, opened 2025-08-21
refactor: use properly types injection keys instead of plain names

susnux:
### ☑️ Resolves

somewhat for #841

### DIFF
--- a/src/components/NcContent/constants.ts
+++ b/src/components/NcContent/constants.ts
@@ -1,0 +1,16 @@
+/*!
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import type { InjectionKey } from 'vue'
+
+/**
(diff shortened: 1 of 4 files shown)


src/components/NcContent/constants.ts
```diff
@@ -0,0 +1,11 @@
+import type { InjectionKey } from 'vue'
+
+/**
+ * Injects a callback to register the app navigation to the content container.
+ */
+export const INJECTION_KEY_HAS_APP_NAVIGATION: InjectionKey<(hasAppNavigation: boolean) => void> = Symbol.for('NcContent:setHasAppNavigation')
```
ShGKme: More common name would be `HAS_APP_NAVIGATION_KEY`, to not duplicate the entire type name in the variable name.

ShGKme: Or using it via composables, e.g. `useNcContentProvider()` and `useNcContent()`

ShGKme (approved):

src/components/NcContent/constants.ts
susnux: done

PR #7351, opened 2025-08-21
feat: add new border design for NcInputField and NcTextArea

susnux:
### ☑️ Resolves

- resolves https://github.com/nextcloud-libraries/nextcloud-vue/issues/7221

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot 2025-08-21 at 19-01-55 Nextcloud Vue Style Guide|Screenshot 2025-08-21 at 19-01-16 Nextcloud Vue Style Guide
Screenshot 2025-08-21 at 19-01-49 Nextcloud Vue Style Guide|Screenshot 2025-08-21 at 19-01-28 Nextcloud Vue Style Guide

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist


kra-mo (changes requested):
Nice :)

But yeah, adjusting `border-top-color` doesn't work, because it creates this unnatural cutoff between the two colors (which is especially visible on HiDPI):

Image

Here, I used a `box-shadow` instead with `0 -1px var(--input-border-color),  0 0 0 1px color-mix(in srgb, var(--input-border-color), 66% transparent)` and that worked much better:

Image

Additionally, when hovered or active, only one color should be used all-around. If we use the stronger color, couldn't we also make it only 1px instead of 2?

Also, with the current approach, NcTextArea with success or error doesn't preserve this styling when active.

Also also, it would be nice if the two states could transition smoothly over maybe 200ms if possible, but that would just be a bonus :)

kra-mo:
Also, maybe we could do

```css
rgb(from var(--input-border-color) r g b / 35%)
```

now instead of

```css
color-mix(in srgb, var(--input-border-color), 65% transparent)
```

https://caniuse.com/css-relative-colors

or is support too low for what we're targeting?

susnux:
> But yeah, adjusting border-top-color doesn't work, because it creates this unnatural cutoff between the two colors (which is especially visible on HiDPI):

Well this was the easy way - which only needed simple changes. My thought was that 1px is rather not noticeable on high res displays (e.g. I do not use scaling with by 2.6k display, so 1px border is only 0.1mm :sweat_smile: )

But yes with scaling it might be noticeable, but then this will get complex - because the box shadow is already used for accessibility purposes on this components.

susnux:
> or is support too low for what we're targeting?

Today you only need to look at Safari - as the IE of our time - if it supports it with version 15 then its fine.
So in this case not.

susnux:
> Additionally, when hovered or active, only one color should be used all-around. If we use the stronger color, couldn't we also make it only 1px instead of 2?

No this was done for accessibility to have the proper active visible state

kra-mo:
> the box shadow is already used for accessibility purposes

I meannn, you can maybe use `filter: drop-shadow()`, although not exactly the proper solution lol.

kra-mo:
> No this was done for accessibility to have the proper active visible state

Yeah, I thought so. Is 2px with *such* high contrast really mandated for that? It is *really* visible lol.

susnux:
> Yeah, I thought so. Is 2px with such high contrast really mandated for that? It is really visible lol.

I guess that what it should be for accessibility: Really visible.
Lets try what is possible

susnux:
@kra-mo what do you think about this solution? (Input fields only not the textarea)

kra-mo:
> @kra-mo what do you think about this solution? (Input fields only not the textarea)

Really nice! :)

susnux:
> If this is not easily fixable, no problem. But if it is, the order used for fields surprisingly looks more natural and I'd adapt that for areas too.

Sure adjusted the textarea to the inputfield look :)

PR #7363, opened 2025-08-24
feat(NcColorPicker): allow to clear selected value

susnux:
### ☑️ Resolves

- Resolves https://github.com/nextcloud-libraries/nextcloud-vue/issues/4037

Allow to clear the selected color:
- palette can be unchecked
- custom color can be cleared

### 🏁 Checklist

### DIFF
--- a/src/components/NcColorPicker/NcColorPicker.vue
+++ b/src/components/NcColorPicker/NcColorPicker.vue
@@ -100,6 +100,46 @@ export default {
 </style>
 ```
 
+* Allowing to clear the selected color
+
+```vue
+<template>
+	<div class="container1">
+		<NcColorPicker clearable v-model="color">
+			<NcButton>Click Me</NcButton>
+		</NcColorPicker>
+		<div :style="{'background-color': color}" class="color-preview" :class="{'color-preview--none': !color}" />
+	</div>
+</template>
+<script>
+export default {
+	data() {
+		return {
+			color: '#0082c9',
+			open: false
+		}
+	}
+}
+</script>
+<style scoped>
+.container1 {
+	display: flex;
+	gap: 20px;
+}
+
+.color-preview {
+	width: 100px;
+	height: 34px;
+	border: 1px solid black;
+	border-radius: 6px;
+}
+
+.color-preview--none {
+	background: repeating-conic-gradient(#808080 0 25%, #0000 0 50%) 50% / 20px 20px;
+}
+</style>
+```
+
 * Using advanced fields including HEX, RGB, and HSL:
 
 ```vue
@@ -137,13 +177,14 @@ export default {
 </docs>
 
 <script setup lang="ts">
+import type { Payload } from '@ckpack/vue-color'
 import type { Color } from '../../utils/colors.ts'
 
 import { Chrome as VueChrome } from '@ckpack/vue-color'
-import { mdiArrowLeft, mdiCheck, mdiDotsHorizontal } from '@mdi/js'
+import { mdiArrowLeft, mdiCheck, mdiCloseCircleOutline, mdiDotsHorizontal } from '@mdi/js'
 import { computed, ref } from 'vue'
 import { t } from '../../l10n.ts'
-import { defaultPalette } from '../../utils/colors.ts'
+import { COLOR_BLACK, COLOR_WHITE, defaultPalette } from '../../utils/colors.ts'
 import { createElementId } from '../../utils/createElementId.ts'
 import logger from '../../utils/logger.ts'
 import NcButton from '../NcButton/index.ts'
@@ -151,9 +192,11 @@ import NcIconSvgWrapper from '../NcIconSvgWrapper/index.ts'
 import NcPopover from '../NcPopover/index.js'
 
 /**
- * A HEX color that represents the initial value of the picker
+ * A HEX color that represents the initial value of the picker.
+ *
+ * If the `clearable` prop is set then also `undefined` might be given or emitted for when the color gets cleared.
  */
-const currentColor = defineModel<string>({ required: true })
+const currentColor = defineModel<string | undefined>({ required: true })
 
 /**
  * The open state of the color picker.
@@ -165,6 +208,12 @@ const props = withDefaults(defineProps<{
 	 * Set to `true` to enable advanced fields including HEX, RGB, and HSL
 	 */
 	advancedFields?: boolean
+
+	/**
+	 * Allow to clear the current color.
+	 * When set the `update:modelValue` and `submit` event might emit also `undefined`.
+	 */
+	clearable?: boolean
 
 	/**
 	 * Selector for the popover container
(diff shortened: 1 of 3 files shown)


src/components/NcColorPicker/NcColorPicker.vue
```diff
@@ -166,6 +204,12 @@ const props = withDefaults(defineProps<{
 	 */
 	advancedFields?: boolean
 
+	/**
+	 * Allow to clear the current color.
+	 * When set the `update:modelValue` and `submit` event might emit also `undefined`.
+	 */
+	clear?: boolean
```
Antreesy: We have props like this in several components, I'd expect it to match
```suggestion
	clearable?: boolean
```

src/components/NcColorPicker/NcColorPicker.vue
```diff
@@ -229,18 +275,19 @@ const normalizedPalette = computed(() => {
...
-			: t('A color with a HEX value {hex}', { hex: item.color }),
+			: t('A color with a HEX value {hex}', { hex: typeof item === 'string' ? item : item.color }),
 	}))
 })
 
-const contrastColor = computed(() => {
-	const black = '#000000'
-	const white = '#FFFFFF'
-	return calculateLuma(currentColor.value) > 0.5 ? black : white
+const isCustomColor = computed(() => {
```
Antreesy: There's no option to clear selection from the default pallete, probably because it doesn't treated as custom color?

Image

Antreesy (commented):
Maybe we should cache initial value when popup is opened and compare with it?

src/components/NcColorPicker/NcColorPicker.vue
susnux: You can - just press the color again to unselect.
But this means its not obvious enough and we should always show the clear button.

src/components/NcColorPicker/NcColorPicker.vue
susnux: Yes makes sense! Fully support this!

susnux:
> Maybe we should cache initial value when popup is opened and compare with it?

I get the idea but this would not allow the user to clear it at all. It would just reset it to the previous.

IMHO this is a different feature - which I would like to have:
Add a `confirm` option similar to `NcDateTimePicker` which would enable the "Submit" button but also would allow to add a "Reset" button.

src/components/NcColorPicker/NcColorPicker.vue
```diff
@@ -100,6 +100,41 @@ export default {
...
 
+* Allowing to clear the selected color
+
+```vue
+<template>
+	<div class="container1">
+		<NcButton @click="open = !open"> Click Me </NcButton>
+		<NcColorPicker clearable v-model="color" v-model:open="open" v-slot="{ attrs }">
+			<div v-bind="attrs" :style="{'background-color': color}" class="color1" />
+		</NcColorPicker>
```
ShGKme: This is not valid, as `NcColorPicker`'s content as well as `NcPopover`'s content is supposed to tbe the trigger element, including the reciever of `attrs`.

ShGKme (approved):
Fince, except for the example

ShGKme:
image

Though, with the hardcoded width looks a bit not in line

ShGKme:
We can use somethiung like `mdiCircleOffOutline` or `mdiCloseCircleOutline` colro as the "no color" button.

image

image

susnux:
I adjusted it as suggested by @ShGKme it looks like this now which seems to scale much better:

[Bildschirmaufnahme_20250910_230713.webm](https://github.com/user-attachments/assets/4c963579-f191-4939-98ae-6382c282d05f)

what do you think @nextcloud-libraries/designers ?

Antreesy:
Can we make it 15 with black and white? just for the sake of row balance
image

susnux:
> Can we make it 15 with black and white? just for the sake of row balance

I think we can just increase the default palette when `clearable` is set?

susnux:
@Antreesy adjusted it, what do you think?

Screenshot 2025-09-11 at 13-02-37 Nextcloud Vue Style Guide

nimishavijay:
I was bit confused if it would set a different "clear" color (as in rgba(0,0,0,0) will be set) or reset to what was set before the color picker was opened, maybe because of the position which makes it look like another color. What do you think of:
- [ ] exchanging the positions of the x and the 3dot button, so it looks like a "More colours" button
- [ ] And use a [reset icon](https://fonts.google.com/icons?icon.set=Material+Symbols&icon.size=24&icon.color=%23222222&icon.query=reset&selected=Material+Symbols+Outlined:replay:FILL@0;wght@400;GRAD@0;opsz@24) instead of the circle-close icon?

That way we will have a list of colours and a "more colours" button, and on the next row there is a confirm and reset button, what do you think?

ShGKme:
As I understand, this is not to reset (to some predefined default value), but to set no value

susnux:
yes @nimishavijay it is not to reset to some value but to clear any selected value basically selecting no-color (the undefined state).

ShGKme:
Though, the linked issue says:

> Would be nice to have a 'Clear' button inside of NcPopover to be able to reset color, which has been set up before

ShGKme:
@Antreesy Do you remember the use-case?

susnux:
My use case is to have a proper way to have **no color** selected and reset the input completely.

As mentioned above I think the other one is a different use case, see here: https://github.com/nextcloud-libraries/nextcloud-vue/pull/7363#issuecomment-3219638000
Which I think should be solved with a `confirm` prop like we do in other pickers (see e.g. date time picker) where we can have "confirm" and "reset". And the value is only emitted on "confirm".

Antreesy:
> @Antreesy Do you remember the use-case?

I've no idea 🙈 Maybe it was just a spontaneous thought to be able to dismiss changes, but since it hasn't made it to live in 2 years - not relevant anymore

nimishavijay:
Yes I am having a hard time seeing a use case for this. I can see the use case to reset to the previous value (as otherwise there is not even a close button on the little popup) but the only time I can think of where "no color" could be useful is when setting a background for an emoji avatar for a Talk conversation. You may want just an emoji without any colour in the background, so then selecting "no color" makes sense (is this something similar to what you had in mind?). 
Either way then it can be a separate item on its own row with text that says "No color" and the `mdiCircleOffOutline` icon to avoid confusion, what do you think?

susnux:
> Yes I am having a hard time seeing a use case for this.

Same use case as clearing the date time picker or any other field: Because you want to reset the answer to "no answer" state. E.g. useful in the forms app.

src/components/NcColorPicker/NcColorPicker.vue
```diff
@@ -325,21 +401,34 @@ function hexToRGB(hex: string): [number, number, number] {
 								:aria-label="name"
 								:name="`color-picker-${id}`"
 								:checked="color === currentColor"
-								@click="pickColor(color)">
+								@click="toggleColor(color)">
+						</label>
+						<label class="color-picker__clear" :title="t('No color')">
```
Antreesy:
```suggestion
						<label v-if="clearable" class="color-picker__clear" :title="t('No color')">
```
Otherwise it always appears?

Image

src/components/NcColorPicker/NcColorPicker.vue
```diff
@@ -255,15 +307,36 @@ function handleConfirm(hideCallback: () => void) {
...
+
+/**
+ * Get the text color with the most constrast for a given background color.
+ *
+ * @param color - The background color
+ */
+function getContrastColor(color: string): string {
+	const black = '#000000'
+	const white = '#FFFFFF'
+	return calculateLuma(color) > 0.5 ? black : white
```
Antreesy: Can we do it here, since we have colors now?
```suggestion
	return calculateLuma(color) > 0.5 ? COLOR_BLACK.color : COLOR_WHITE.color
```

susnux: done :heavy_check_mark:

jancborchardt (approved):
Seems good, if necessary :)

PR #7367, opened 2025-08-24
fix(NcHeaderMenu): apply `ariaLabel` prop and add proper default values

susnux:
### ☑️ Resolves

1. The `ariaLabel` was not applied to the button.
2. Some defaults were only added implicit causing Vue warnings - so added explicit `undefined` as default.

### 🏁 Checklist

### DIFF
--- a/src/components/NcHeaderMenu/NcHeaderMenu.vue
+++ b/src/components/NcHeaderMenu/NcHeaderMenu.vue
@@ -69,6 +69,8 @@ import { getTrapStack } from '../../utils/focusTrap.js'
 import NcButton from '../NcButton/index.ts'
 
 const {
+	ariaLabel = undefined,
+	description = undefined,
 	excludeClickOutsideSelectors = [],
 	open = false,
 	isNav = false,
@@ -252,9 +254,10 @@ function clearFocusTrap() {
 		<NcButton
 			:id="isNav ? triggerId : null"
 			ref="trigger-button-key"
-			class="header-menu__trigger"
 			:aria-controls="`header-menu-${id}`"
 			:aria-expanded="isOpened.toString()"
+			:aria-label
+			class="header-menu__trigger"
 			size="large"
 			variant="tertiary-no-background"
 			@click.prevent="toggleMenu">


src/components/NcHeaderMenu/NcHeaderMenu.vue
```diff
@@ -69,6 +69,8 @@ import { getTrapStack } from '../../utils/focusTrap.js'
 import NcButton from '../NcButton/index.ts'
 
 const {
+	ariaLabel = undefined,
+	description = undefined,
```
ShGKme, line 73: Doesn't it have `undefined` value anyway when it is not set?

src/components/NcHeaderMenu/NcHeaderMenu.vue
```diff
@@ -252,9 +254,10 @@ function clearFocusTrap() {
 		<NcButton
 			:id="isNav ? triggerId : null"
 			ref="trigger-button-key"
-			class="header-menu__trigger"
 			:aria-controls="`header-menu-${id}`"
 			:aria-expanded="isOpened.toString()"
+			:aria-label
+			class="header-menu__trigger"
```
ShGKme, line 260: Do you sort HTML attributes alphabetically? 🤨

ShGKme (approved):

src/components/NcHeaderMenu/NcHeaderMenu.vue
susnux: Yes but this is a bit more explicit when reading and also satisfies https://eslint.vuejs.org/rules/require-default-prop

Yes we can ignore but 1 comment line to ignore or just add this default is no difference and I consider this here cleaner. (Also after script-setup compilation its no overhead).

src/components/NcHeaderMenu/NcHeaderMenu.vue
susnux: :abc: 

No but here I added the missing `aria-label`

src/components/NcHeaderMenu/NcHeaderMenu.vue
ShGKme: I forgot about the eslint rule

Issue #7120, opened 2025-07-02
Enhance `<NcCheckboxRadioSwitch>` with new layout and description


ShGKme:
- [x] Add `description` prop and slot (for the non-grouped view at least) (https://github.com/nextcloud-libraries/nextcloud-vue/pull/7378)
- [ ] Add a new layout with the switch on the end
- [ ] Optional additional breaking change: split grouped and non-grouped into 2 components

ShGKme:
- Closed by https://github.com/nextcloud-libraries/nextcloud-vue/pull/7781

PR #7378, opened 2025-08-25
feat(NcCheckboxRadioSwitch): Add support for a description field

CarlSchwan:
### ☑️ Resolves

- Related #7120

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
+++ b/src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
@@ -19,6 +19,7 @@
 			class="checkbox-content__icon"
 			:class="{
 				'checkbox-content__icon--checked': isChecked,
+				'checkbox-content__icon--has-description': !isButtonType && $slots.description,
 				[iconClass]: true,
 			}"
 			:aria-hidden="true"
@@ -39,9 +40,18 @@
 			</slot>
 		</span>
 
-		<span v-if="$slots.default" class="checkbox-content__text" :class="[textClass]">
-			<!-- @slot The checkbox/radio label -->
-			<slot />
+		<span class="checkbox-content__wrapper">
+			<span
+				v-if="$slots.default"
+				:id="labelId"
+				class="checkbox-content__text"
+				:class="textClass">
+				<!-- @slot The checkbox/radio label -->
+				<slot />
+			</span>
+			<span v-if="!isButtonType && $slots.description" :id="descriptionId" class="checkbox-content__description">
+				<slot name="description" />
+			</span>
 		</span>
 	</span>
 </template>
@@ -143,6 +153,22 @@ export default {
 			type: Number,
 			default: 24,
 		},
+
+		/**
+		 * Label id attribute
+		 */
+		labelId: {
+			type: String,
+			required: true,
+		},
+
+		/**
+		 * Description id attribute
+		 */
+		descriptionId: {
+			type: String,
+			required: true,
+		},
 	},
 
 	computed: {
@@ -199,8 +225,11 @@ export default {
 	// but restrict to content so plain checkboxes / radio switches do not expand
 	max-width: fit-content;
 
-	&__text {
+	&__wrapper {
 		flex: 1 0;
+	}
+
+	&__text {
 
 		&:empty {
 			// hide text if empty to ensure checkbox outline is a circle instead of oval
@@ -214,10 +243,24 @@ export default {
 		margin-block: calc((var(--default-clickable-area) - 2 * var(--default-grid-baseline) - var(--icon-height)) / 2) auto;
 	}
 
+	&-checkbox:not(&--button-variant) &__icon--has-description,
+	&-radio:not(&--button-variant) &__icon--has-description,
+	&-switch:not(&--button-variant) &__icon--has-description {
+		display: flex;
+		align-items: center;
+		margin-block-end: 0;
+		align-self: start;
+	}
+
 	&__icon > * {
 		width: var(--icon-size);
 		height: var(--icon-height);
 		color: var(--color-primary-element);
+	}
+
+	&__description {
+		display: block;
+		color: var(--color-text-maxcontrast);
 	}
 
 	&--button-variant {
--- a/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
+++ b/src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
@@ -22,6 +22,10 @@ Note: All generic attributes on the component, except `class` and `style`, are p
 		<NcCheckboxRadioSwitch v-model="sharingEnabled">
 			Enable sharing. This can contain a long multiline text, that will be wrapped in a second row. It is generally not advised to have such long text inside of an element
 		</NcCheckboxRadioSwitch>
+
+		<NcCheckboxRadioSwitch v-model="sharingEnabled" description="The description">
+			Enable sharing with description
+		</NcCheckboxRadioSwitch>
 		<br>
 		sharingEnabled: {{ sharingEnabled }}
 	</div>
@@ -233,6 +237,16 @@ export default {
 		<NcCheckboxRadioSwitch v-model="sharingEnabled" type="switch">
 			Enable sharing. This can contain a long multiline text, that will be wrapped in a second row. It is generally not advised to have such long text inside of an element
 		</NcCheckboxRadioSwitch>
+		<NcCheckboxRadioSwitch v-model="sharingEnabled" type="switch" description="Instead you can use a description as a prop which can also be a long multiline text, that will be wrapped in a second row.">
+			Enable sharing.
+		</NcCheckboxRadioSwitch>
+
+		<NcCheckboxRadioSwitch v-model="sharingEnabled" type="switch">
+			Enable sharing.
+			<template #description>
+				Or you can use a description as slot which can also be a <strong>long multiline text</strong>, that will be wrapped in a second row.
+			</template>
+		</NcCheckboxRadioSwitch>
 		<br>
 		sharingEnabled: {{ sharingEnabled }}
 	</div>
@@ -276,7 +290,8 @@ export default {
 		<input
 			v-if="!isButtonType"
 			:id="id"
-			:aria-labelledby="!isButtonType && !ariaLabel ? `${id}-label` : null"
+			:aria-labelledby="!isButtonType && !ariaLabel ? labelId : null"
+			:aria-describedby="!isButtonType && (description || $slots.description) ? descriptionId : null"
 			:aria-label="ariaLabel || undefined"
 			class="checkbox-radio-switch__input"
 			:disabled="disabled"
@@ -298,11 +313,20 @@ export default {
 			:button-variant="buttonVariant"
 			:is-checked="isChecked"
 			:loading="loading"
+			:description="description"
+			:label-id="labelId"
+			:description-id="descriptionId"
 			:icon-size
 			@click="onToggle">
 			<template #icon>
 				<!-- @slot The checkbox/radio icon, you can use it for adding an icon to the button variant -->
 				<slot name="icon" />
+			</template>
+			<template v-if="$slots.description || description" #description>
+				<!-- @slot The checkbox/radio/switch description, you can use it for adding a more complex description element as opposed to the description prop -->
+				<slot name="description">
+					{{ description }}
+				</slot>
 			</template>
 
 			<template v-if="!!$slots.default" #default>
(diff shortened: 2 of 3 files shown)


src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
```diff
@@ -143,13 +150,25 @@ export default {
...
 	},
 
 	computed: {
 		isButtonType() {
 			return this.type === TYPE_BUTTON
 		},
 
+		hasDescription() {
+			return !this.isButtonType && (this.$slots.description || this.props.description)
+		},
```
ShGKme: `$slots` is not reactive and cannot be used in a `computed`. It should be a method or a check in the template.

ShGKme: And it looks like it should have been `this.description` instead of `this.props.description`.

src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
```diff
@@ -39,10 +40,16 @@
...
-		</span>
+		<div class="checkbox-content__wrapper">
+			<span v-if="$slots.default" class="checkbox-content__text" :class="[textClass]">
+				<!-- @slot The checkbox/radio label -->
+				<slot />
+			</span>
+			<em v-if="hasDescription" class="checkbox-content__description">
+				{{ description }}
+				<slot name="description" />
+			</em>
```
ShGKme: For a11y the description should be connected to the `input` element via `aria-describedby`

src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
```diff
@@ -39,10 +40,20 @@
...
 
-		<span v-if="$slots.default" class="checkbox-content__text" :class="[textClass]">
-			<!-- @slot The checkbox/radio label -->
-			<slot />
-		</span>
+		<div class="checkbox-content__wrapper">
+			<span
+				v-if="$slots.default"
+				class="checkbox-content__text"
+				:class="[textClass]"
```
ShGKme: nitpik: 
```suggestion
				:class="textClass"
```

ShGKme: As we touch it anyway

src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
```diff
@@ -39,10 +40,20 @@
...
+				:aria-describedby="descriptionId">
+				<!-- @slot The checkbox/radio label -->
+				<slot />
+			</span>
+			<em v-if="!isButtonType && ($slots['description'] || description)" :id="descriptionId" class="checkbox-content__description">
+				{{ description }}
+				<slot name="description" />
```
ShGKme: We never need both the plain text prop and the slot content one after another.
We can use plain text as the default slot value.

```suggestion
				<slot name="description">
					{{ description }}
				</slot>
```

src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
```diff
@@ -298,12 +312,17 @@ export default {
...
 			:icon-size
 			@click="onToggle">
 			<template #icon>
 				<!-- @slot The checkbox/radio icon, you can use it for adding an icon to the button variant -->
 				<slot name="icon" />
 			</template>
+			<template #description>
+				<!-- @slot The checkbox/radio description, you can use it for adding an complex description element to the button variant as opposed to the description slot -->
+				<slot name="description" />
+			</template>
```
ShGKme: Otherwise we always pass the description slot content, even if there is no description.
```suggestion
			<template v-if="$slots.description" #description>
				<!-- @slot The checkbox/radio description, you can use it for adding an complex description element to the button variant as opposed to the description slot -->
				<slot name="description" />
			</template>
```

ShGKme:
image

When there is a description, alignment breaks and the icon is no longer aligned with the first line.

CarlSchwan:
> When there is a description, alignment breaks and the icon is no longer aligned with the first line.

should be fixed

image

src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
```diff
@@ -39,10 +40,21 @@
 			</slot>
 		</span>
 
-		<span v-if="$slots.default" class="checkbox-content__text" :class="[textClass]">
-			<!-- @slot The checkbox/radio label -->
-			<slot />
-		</span>
+		<div class="checkbox-content__wrapper">
```
ShGKme: I **AM** sorry for making many reviews, always missing something 🙈 
It must be `<span>`, otherwise it is a block element inside an inline element, which is not valid for HTML.

CarlSchwan: done, thanks for bearing with me and all my small mistakes :see_no_evil:

ShGKme (approved):

src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
```diff
@@ -19,6 +19,7 @@
 			class="checkbox-content__icon"
 			:class="{
 				'checkbox-content__icon--checked': isChecked,
+				'checkbox-content__icon--hasDescription': !isButtonType && ($slots['description'] || description),
```
susnux: We normally use kebab-case within templates:

```suggestion
				'checkbox-content__icon--has-description': !isButtonType && ($slots['description'] || description),
```

src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
```diff
@@ -39,9 +40,20 @@
 			</slot>
 		</span>
 
-		<span v-if="$slots.default" class="checkbox-content__text" :class="[textClass]">
-			<!-- @slot The checkbox/radio label -->
-			<slot />
+		<span class="checkbox-content__wrapper">
```
susnux, line 43: Would be better readable if this is a `div` because this is *not* an inline element but a block element.

src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
```diff
@@ -39,9 +40,20 @@
...
+			<span
+				v-if="$slots.default"
+				class="checkbox-content__text"
+				:class="textClass"
+				:aria-describedby="descriptionId">
```
susnux: This needs to be set on the `<input>`.
Also with a description the `id` set to `aria-labelledby` needs to be the id of this span.

So I would suggest to create the IDs on the parent component (the NcCRS) and pass it to this component as `label-id` and `description-id` then you can set both to the `input` and apply them here to their respective containers.

susnux (dismissed):
The aria labelling needs to be adjusted to have label and description on the proper text containers

ShGKme:
> When there is a description, alignment breaks and the icon is no longer aligned with the first line.

I just found that on the layout in the new design it was on the middle. cc @kra-mo What about keeping it aligned with the top line?

image

susnux:
> What about keeping it aligned with the top line?

IMHO this looks nicer if there are multiple switches

ShGKme:
> IMHO this looks nicer if there are multiple switches

This - like on screenshot (middle) or aligned with the top line? :D

susnux:
Aligned with the top line so it looks consistent and not random based on length of text.

src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
CarlSchwan: but then we have a div in a span and it's not valid. See https://github.com/nextcloud-libraries/nextcloud-vue/pull/7378#discussion_r2301200107

kra-mo:
> cc @kra-mo What about keeping it aligned with the top line?

Pf, if there was a chance that two of these items could be next to each other horizontally, I'd say top.

Otherwise, I'm dubious about how this would look with eg. an icon instead of a switch.

On the other hand, if you had a really long description, you'd always want it top-aligned, so if developers won't get the freedom to decide alignment, top is probably safer.

Sure, go for top, especially because of the wildly variable height.

src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
CarlSchwan: done

src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
susnux: Well yes basically all that `span` are "invalid` and should rather be `div` - but ok for the moment.

susnux: (as it is unrelated to the PR)

src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
```diff
@@ -39,9 +40,20 @@
...
+				:id="labelId"
+				class="checkbox-content__text"
+				:class="textClass">
+				<!-- @slot The checkbox/radio label -->
+				<slot />
+			</span>
+			<em v-if="!isButtonType && ($slots.description || description)" :id="descriptionId" class="checkbox-content__description">
+				<slot name="description">
+					{{ description }}
+				</slot>
```
susnux: I wonder if we should simplify this like

```suggestion
				<slot name="description" />
```

And also completely remove the description prop from this component

src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
```diff
@@ -298,12 +313,19 @@ export default {
...
 			:icon-size
 			@click="onToggle">
 			<template #icon>
 				<!-- @slot The checkbox/radio icon, you can use it for adding an icon to the button variant -->
 				<slot name="icon" />
 			</template>
+			<template v-if="$slots.description" #description>
+				<!-- @slot The checkbox/radio description, you can use it for adding an complex description element to the button variant as opposed to the description slot -->
+				<slot name="description" />
+			</template>
```
susnux: Then we could just pass it like this:

```suggestion
			<template v-if="description || $slots.description" #description>
				<!-- @slot The checkbox/radio description, you can use it for adding an complex description element to the button variant as opposed to the description slot -->
				<slot name="description">
					{{ description }}
				</slot>
			</template>
```

src/components/NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue
```diff
@@ -298,12 +313,19 @@ export default {
...
+			:label-id="labelId"
+			:description-id="descriptionId"
 			:icon-size
 			@click="onToggle">
 			<template #icon>
 				<!-- @slot The checkbox/radio icon, you can use it for adding an icon to the button variant -->
 				<slot name="icon" />
 			</template>
+			<template v-if="$slots.description" #description>
+				<!-- @slot The checkbox/radio description, you can use it for adding an complex description element to the button variant as opposed to the description slot -->
```
susnux: This description makes no sense, no? Because this is about the description slot.
Also for button variant it does not exist?

src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
```diff
@@ -39,9 +40,20 @@
...
+		<span class="checkbox-content__wrapper">
+			<span
+				v-if="$slots.default"
+				:id="labelId"
+				class="checkbox-content__text"
+				:class="textClass">
+				<!-- @slot The checkbox/radio label -->
+				<slot />
+			</span>
+			<em v-if="!isButtonType && ($slots.description || description)" :id="descriptionId" class="checkbox-content__description">
```
susnux: I am not sure `em` is a proper tag here as semantically it `represents stress emphasis of its contents` so I do not see how this should be semantically more emphasized than the label. would just use a `<span>` as well here.

susnux (commented):
a11y should be fine now :rocket:

src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
ShGKme: Why?

susnux: Why does the sub component also need to provide both? Only having the slot simplifies the sub-component by only care about slot not slot or prop.

PR #7398, opened 2025-08-27
feat(NcAssistant*): provide components for consistent integrations design

susnux:
* see also https://github.com/nextcloud/server/pull/54679

### 🖼️ Screenshots

Bildschirmfoto am 2025-08-27 um 17 55 16

### 🏁 Checklist


Antreesy:
Quickly checked on preview:
- top border should be present? IMO it doesn't look good at all for these buttons to have a border:
2025-08-27_18h52_35 2025-08-27_18h52_43

- can we apply saturation filter to make it more grey-ish, when disabled? This doesn't really look disabled:
2025-08-27_18h52_51

susnux:
> top border should be present? IMO it doesn't look good at all for these buttons to have a border:

From the screenshots shared by @jancborchardt yes they contain that border.

src/components/NcAssistantIcon/NcAssistantIcon.vue
```diff
@@ -0,0 +1,72 @@
...
+
+<style module>
+.svg {
+	width: v-bind('sizePx');
+	height: v-bind('sizePx');
+	max-width: v-bind('sizePx');
+	max-height: v-bind('sizePx');
+}
+
+.icon:not(.iconInline) {
```
ShGKme: Though CSS Modules allow having a great encapsulation between components, let’s still follow the BEM convention:
- To have a consistent naming convention
- To make searching source easier by class name

```suggestion
.icon-assistent:not(.icon-assistent--inline) {
```

ShGKme: Same for other components

ShGKme:
image

On the screenshot in the PR description, the icon position looks a bit not aligned with anything, not centered. Is it an intentional position?

src/components/NcAssistantIcon/NcAssistantIcon.vue
susnux: Its soo ugly to use non camelCase names in modules :cry: 
(compare `$styles.primaryButton` vs `$styles['primary-button']` especially as modules would not use this kind of constructs but would use `composes` instead).

Nevertheless will adjust this.

susnux:
> On the screenshot in the PR description, the icon position looks a bit not aligned with anything, not centered. Is it an intentional position?

Its just for showcasing the content wrapper (as that border is a bit more complex to create and thus a very simple container can help). I did not put too much into the styling :sweat_smile: 
I can rework the docs example.

src/components/NcAssistantIcon/NcAssistantIcon.vue
ShGKme: BEM doesn't forbid to use camelCase.

It can still be `primaryButton` for a block, `primaryButton__icon` for an element and `primaryButton_wide` for a modifier. All these names can be used with a dot notation.

https://en.bem.info/methodology/naming-convention/#camelcase-style

ShGKme: But the most important part of BEM is not the class names style. It's just a style.

BEM is a methodology of layout structure in the first place: blocks, elements, modifiers.

ShGKme:
I missed that this icon is not a part of the `NcAssistentContent`

src/components/NcAssistantContent/NcAssistantContent.vue
```diff
@@ -0,0 +1,91 @@
...
+	<div :class="$style.wrapper">
+		<div :class="[$style.content, contentClasses]">
+			<slot />
+		</div>
+	</div>
+</template>
+
+<style module lang="scss">
+.wrapper {
+	background-image: var(--color-border-assistant, linear-gradient(125deg, #7398FE 50%, #6104A4 125%));
```
ShGKme, line 77: Just curious, isn't it possible to apply linear-gradient for border-color?

susnux: No sadly that is not possible. That is why we needed a similar hack (box shadow) for the input elements.

Antreesy: What about https://developer.mozilla.org/en-US/docs/Web/CSS/border-image

susnux:
> can we apply saturation filter to make it more grey-ish, when disabled? This doesn't really look disabled:

@Antreesy resolved this issue

src/components/NcAssistantIcon/NcAssistantIcon.vue
susnux: Done

src/components/NcAssistantContent/NcAssistantContent.vue
susnux: @Antreesy 

> [border-radius](https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius) has no effect on the border [when using border-image]

Antreesy: Haven't tried this fiddle, but looks like it might work:

```css
.btn-gradient-2 {
  background: linear-gradient() padding-box,  linear-gradient() border-box;
  border-radius:... 
  border:... 
}
```

ShGKme (approved):

PR #7401, opened 2025-08-27
feat: add NcKbd component

ShGKme:
### ☑️ Resolves

- Ref: https://github.com/nextcloud-libraries/nextcloud-vue/issues/7124
- First component for keyboard shortcuts — replacement for `<kbd>`
  - Custom styling
  - Localized key names
  - Auto icons
  - macOS handling
  - Custom slot for customisation

**Note:** complete form will have an additional component for an entire keyboard shortcut list

### 🖼️ Screenshots

image

### 🏁 Checklist


src/components/NcKbd/NcKbd.vue
```diff
@@ -0,0 +1,274 @@
...
+	// TRANSLATORS Shift key on keyboard (Windows, Linux)
+	'shift': t('Shift'),
+	// TRANSLATORS Enter key on keyboard (Windows, Linux)
+	'enter': t('Enter'),
+	// TRANSLATORS Tab key on keyboard (Windows, Linux)
+	'tab': t('Tab'),
+	// TRANSLATORS Delete key on keyboard (Windows, Linux)
+	'delete': t('Delete'),
+	'esc': t('Escape'),
+
```
Antreesy: `'space': t('Space'),` isn't missing here intentionally?

ShGKme: Fixed

kra-mo (approved):

src/components/NcKbd/NcKbd.vue
```diff
@@ -0,0 +1,243 @@
...
+	Tab: !mac
+		? t('Tab') // TRANSLATORS: Tab key on keyboard
+		: '⇥',
+	Delete: !mac
+		? t('Delete') // TRANSLATORS: Delete key on keyboard
+		: '⌫',
+	Escape: !mac
+		? t('Escape') // TRANSLATORS: Escape key on keyboard
+		: '⎋',
+	Space: t('Space'), // TRANSLATORS: Space key on keyboard
```
susnux, line 65: if we sometimes use icons for them, maybe use this for space?
https://pictogrammers.com/library/mdi/icon/keyboard-space/

its quite common, no? cc @nextcloud-libraries/designers

src/components/NcKbd/NcKbd.vue
```diff
@@ -0,0 +1,243 @@
...
+	line-height: 1;
+	white-space: nowrap;
+
+	& + .kbd {
+		margin-inline-start: calc(1 * var(--default-grid-baseline));
+	}
+}
+</style>
+
+<docs>
```
susnux, line 101: docs on the bottom? (I like it but its new)

src/components/NcKbd/NcKbd.vue
```diff
@@ -0,0 +1,243 @@
...
+		</slot>
+	</kbd>
+</template>
+
+<style lang="scss" module>
+.kbd {
+	display: inline-flex;
+	align-items: center;
+	justify-content: center;
+	height: var(--default-clickable-area);
```
susnux, line 85: IMHO it looks weird that those "keys" have different width.
You can see it in your example on the `<-` and the `^` keys.
Maybe we should - for those single letter keys - add a min-width here, like:

```suggestion
	height: var(--default-clickable-area);
	min-width: var(--default-clickable-area);
```

src/components/NcKbd/NcKbd.vue
ShGKme: We never use icons (images). It was discussed with @kra-mo that symbols look fine on macOS.

Same for the space — even though there is a symbol, it was decided to use a word here.

image

src/components/NcKbd/NcKbd.vue
kra-mo: Oh right, min-width is a thing. Yeah, that would totally solve the issue here ^^

src/components/NcKbd/NcKbd.vue
Antreesy: Definitely would put it for mac, otherwise it stands out too much.
Also might be less confusion for translators. That's usually the only button that doesn't have a text on a physical keyboard, and also the one i would translate out of all other

UPD: maybe this? `␣`

src/components/NcKbd/NcKbd.vue
ShGKme: I can move it back to the top :D

It just always annoyed me that I need to scroll long docs to work with the component's code...

Also less diff after we switch from styleguidist ._.

src/components/NcKbd/NcKbd.vue
ShGKme: Good point, done.

src/components/NcKbd/NcKbd.vue
susnux: Yes it will look silly translated like in German I rarely see this like

`[Strg] + [Leertaste]`

I would expect the icon for the space key as it does not has any label on the keyboard.
But also for the other icons you added for Mac, are you sure they work for screen readers? If not maybe just use icons and add aria label with the real name.

src/components/NcKbd/NcKbd.vue
susnux: For me this is just much nicer!
Similar like moving template below - Most of the time you are interested in the script ;)

susnux (approved):
Space as icon or as text is a @nextcloud-libraries/designers decision, while I rarely saw a label used (especially translated) its probably ok.
So I let this up to the designers - code is fine!

src/components/NcKbd/NcKbd.vue
ShGKme: I didn't test these symbols specifically, but usually it works well with screen readers.
I'd prefer to use symbols to safe from translating it.

ShGKme: > Yes it will look silly translated like in German I rarely see this like
> 
> `[Strg] + [Leertaste]`

Could you show some example in an app on Linux?

ShGKme: I actually had MDI icons + aria-label at some point :D

kra-mo: I think this might be a language-by-language thing. I do know that for example in Hungarian, `Space` and `Enter` are common, even though there are localized equivalents…

…but then again, whether ⏎ is called `Enter` or `Return` is also locale-specific, so I think it probably is best to leave it to translators, with a note telling them to only localize it if it is common to do so in their language.

ShGKme: We can also add a comment for translator to translate it only if it is a common practice.

E.g., `CTRL` almost never translated in Russian.

kra-mo: In any case, I think the word `Space` is still more well-recognized internationally than the `␣` symbol, but I might be wrong.

src/utils/platform.ts
```diff
@@ -0,0 +1,6 @@
+/*
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+export const isMacos = /Mac|iPhone|iPod|iPad/i.test(navigator.userAgent)
```
Antreesy: we also used 'darwin' in useHotKey, would be good to check, if needed here, and then unify the usage

src/components/NcKbd/NcKbd.vue
susnux: > Could you show some example in an app on Linux?

E.g.:

grafik

src/components/NcKbd/NcKbd.vue
```diff
@@ -0,0 +1,244 @@
...
+ * - Windows/Linux uses localized names
+ * - In ternary expressions, // TRANSLATORS comments only works for the second operand, but not for `else`
+ */
+const labels = computed(() => ({
+	ArrowUp: '↑',
+	ArrowDown: '↓',
+	ArrowLeft: '←',
+	ArrowRight: '→',
+	Control: !mac
+		? t('Ctrl') // TRANSLATORS: Ctrl key on keyboard (Windows/Linux)
```
Antreesy, line 45: I think we noticed it last week in spreed repo: //TRANSLATORS comment should be above the t() function, otherwise it's not shown in Transifex.

See https://github.com/nextcloud/spreed/pull/15713/files

src/components/NcKbd/NcKbd.vue
susnux: At least everything is better than the firefox and thunderbird shortcuts help which uses really silly names...
If you both did not see the symbol that often then its probably fine, it just stands out quite visible for MacOS

src/components/NcKbd/NcKbd.vue
susnux: No it works here as you can see in the extracted strings.
The problem is if this is a new statement, e.g. this `t(...), // ...` does not work because of the comma.

ShGKme: > The problem is if this is a new statement, e.g. this `t(...), // ...` does not work because of the comma.

Actually, it does work with coma. BUT it doesn't work in 3rd operand in ternary operator here.

That's why I needed to invert `mac` to `!mac`

src/utils/platform.ts
ShGKme: Done

src/components/NcKbd/NcKbd.vue
susnux: > Actually, it does work with coma. BUT it doesn't work in 3rd operand in ternary operator here.

Ok then its when used in contructor like

```js
new Color(t(...), ...) // TRANSLATORS ...
```

this does not work :pensive:

susnux (approved):

ShGKme:
Rebased onto main, squashed

ShGKme:
If we decied to use an icon for the space, we can update it later without a breaking change.

src/components/NcKbd/NcKbd.vue
```diff
@@ -0,0 +1,244 @@
...
+
+const {
+	symbol = undefined,
+	mac = isMac,
+} = defineProps<{
+	/**
+	 * Key name or symbol to display.
+	 * For common special keys (CTRL, ALT, SHIFT, Arrow keys, etc.) it will display the symbol on macOS and the localized name on Windows/Linux.
+	 */
+	symbol?: 'ArrowUp' | 'ArrowDown' | 'ArrowLeft' | 'ArrowRight' | 'Control' | 'Alt' | 'Shift' | 'Enter' | 'Space' | 'Tab' | 'Delete' | 'Escape' | (string & {})
```
ShGKme, line 21: Just in case before merging.

@susnux Are you fine with this prop name?
(`key` cannot be used as it is a special attribute).
Alternative: `keyname`, `keycode` (but it has a different meaning in web api), `code`.

src/components/NcKbd/NcKbd.vue
kra-mo: > it just stands out quite visible for MacOS

I do agree that it does, but for macOS specifically, it's just about following platform conventions

src/components/NcKbd/NcKbd.vue
susnux: As you also write "Key name" as the doc I think `keyname` is pretty good.
But also `symbol` seems to fit here so you can merged if you like ;)

susnux: But proper name would be `keyName` which would be `<NcKbd key-name="..." />` maybe stick with symbol instead :sweat_smile:

ShGKme: @Antreesy Do I remember it correctly, that in `useHotKey` the preferred way to set a key is the keyCode (like `KeyF`, not `F`)?

susnux: No it takes `F` and then checks if `key` is equal (e.g. `F === F`) otherwise falls back to get the `keyCode` and removes the `Key` or `Digit` and then compares

susnux:
> backport to stable1

Quite old branch ;)

ShGKme:
> Quite old branch ;)

I'm tired 😴

src/components/NcKbd/NcKbd.vue
Antreesy: Just F would be preferable, as we can then derive the keyCode from it

PR #7418, opened 2025-08-31
[stable8] chore: migrate to ESLint 9 and @nextcloud/eslint-config 9

ShGKme:
### ☑️ Resolves

- Fix dramastic difference between `main` and `stable8` branches, making backporting complicated
- Fix issues with TS support and `import` eslint plugin
- Manual backport of https://github.com/nextcloud-libraries/nextcloud-vue/pull/7233
- See by commit:
  - They mostly apply a single change described in the message
  - Almost every commit has green test/eslint/build

### 🏁 Checklist

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -80,14 +80,4 @@ export default defineConfig([
 			},
 		},
 	},
-
-	// Temporarily disable rules during migration
-	{
-		name: '@nextcloud/vue/disabled-during-migration',
-		files: ['**/*.{js,ts,vue,mjs,mts}'],
-		rules: {
-			'vue/no-unused-properties': 'off',
-			'vue/no-unused-refs': 'off',
-		},
-	},
 ])


ShGKme:
image

susnux:
Looks like a massive amount of commits :eyes:
Weird when looking through git history - maybe squash them into auto fixable and rest by logic.

ShGKme:
> Looks like a massive amount of commits 👀

To make it easier to review, with a clear understanding of what was the source of what change and why.

A massive amount of commits for a massive amount of changes.

> Weird when looking through git history

Why is it a problem?

ShGKme:
I can squash some similar auto-fix commits (that fix related things in the same place as I already did).

But I don't want to squash everything:
- Very inconvenient for review — you have just a massive change (that isn't even properly loading in GitHub) of different changes mixed
- Hard to see what rule was the source of what change. It is especially useful when the change is weird.
- For manual change - also good to see what exactly the change is fixing

E.g., in the original PR the source of some changes is still unclear for me.

ShGKme:
Rebased onto `stable8`, resolved conflicts in `NcCheckboxContent`, no actual changes.

Antreesy:
> One review please

image

image

build/format-changelog.mjs
```diff
@@ -5,14 +5,16 @@
...
+
 console.info('🔎 checking format of CHANGELOG.md')
 
 const file = join(import.meta.dirname, '..', 'CHANGELOG.md')
 const content = await readFile(file, { encoding: 'utf-8' })
 
 const formatted = content.replaceAll(
 	/by @([^ ]+) in ((https:\/\/github.com\/)nextcloud-libraries\/nextcloud-vue\/pull\/(\d+))/g,
-	'[\#$4]($2) \([$1]($3$1)\)'
+	'[#$4]($2) ([$1]($3$1))',
```
susnux: Diverged here from main as it should include the escaping backslash:

```suggestion
	'[\\#$4]($2) \\([$1]($3$1)\\)',
```

cypress/visual/NcPasswordField/NcPasswordField.story.vue
```diff
@@ -15,13 +16,17 @@
 <script setup>
 import NcPasswordField from '../../../src/components/NcPasswordField/NcPasswordField.vue'
 
+/* eslint-disable vue/require-prop-comment */
```
susnux: can be removed see below

eslint.config.mjs
```diff
@@ -0,0 +1,82 @@
...
+		rules: {
+			// we are that library - we provide the deprecations...
+			'@nextcloud/vue/no-deprecated-props': 'off',
+		},
+	},
+
+	// Disable required property documentation for test components
+	{
+		name: '@nextcloud/vue/test-override',
+		files: ['tests/**/*.vue'],
```
susnux: For this branch it needs to be:
```suggestion
		files: [
			'cypress/**/*.vue',
			'tests/**/*.vue',
		],
```

eslint.config.mjs
```diff
@@ -0,0 +1,82 @@
...
+				it: 'readonly',
+				test: 'readonly',
+				expect: 'readonly',
+				beforeEach: 'readonly',
+				afterEach: 'readonly',
+				beforeAll: 'readonly',
+				afterAll: 'readonly',
+				jest: 'readonly',
+			},
+		},
```
susnux, line 65: Could also make explicit instead with `import { ... } from '@jest/globals'`
Which we already do in a bunch of test files for better compatibility with vitest (backports work better as in the diff only the import module name changed)

src/components/NcActionButton/NcActionButton.vue
```diff
@@ -324,8 +324,10 @@ export default {
...
 	<li class="action" :class="{ 'action--disabled': disabled }" :role="isInSemanticMenu && 'presentation'">
-		<button :aria-label="ariaLabel"
-			:class="['action-button button-vue', {
+		<button
+			:aria-label="ariaLabel"
+			class="action-button button-vue"
+			:class="[{
 				'action-button--active': isChecked,
 				focusable: isFocusable,
 			}]"
```
susnux:
```suggestion
			:class="{
				'action-button--active': isChecked,
				focusable: isFocusable,
			}"
```

eslint.config.mjs
ShGKme: > Could also make explicit instead with `import { ... } from '@jest/globals'`

Yes, but I didn't want to overload already large PR. This is not about ESLint migration itself and can be done in a follow-up.

At this moment this is configuration problem in the first place.

build/format-changelog.mjs
ShGKme: I don't understand, why we need a slash here.
But I'll add it to have the same as on the `main` branch.

susnux:
> Why is it a problem?

Its hard to see different changes, you have to scroll more than 60 commits in the history log to get to the next changes.

> A massive amount of commits for a massive amount of changes.

Yes but very related changes: *"Migrate to ESLint 9 and Nextcloud ESLint config v9"* :wink: 
For review it might make sense (as this is cosmetic only - I for example only review on the full changes as I only need to check if behavior is changed and or doc strings make sense).

---

Nevertheless changes look good, some few comments.

src/components/NcActionButton/NcActionButton.vue
ShGKme: Same in some other places, fixing all of them

ShGKme:
Added fixups according to the comments, except for import from jest globals. Will do it in a follow up as an a bit unrelated and a large change.

src/components/NcAppSidebar/NcAppSidebar.vue
```diff
@@ -1196,10 +1238,8 @@ export default {
 		checkToggleButtonContainerAvailability() {
 			// Toggle button must be rendered, but there is no element to teleport it to
 			if (this.open === false && !this.noToggle && !this.ncContentSelector) {
-				console.warn(
-					'[NcAppSidebar] It looks like you want to use NcAppSidebar with the built-in toggle button. '
-					+ 'This feature is only available when NcAppSidebar is used in NcContent.',
-				)
+				logger.warn('[NcAppSidebar] It looks like you want to use NcAppSidebar with the built-in toggle button. '
+					+ 'This feature is only available when NcAppSidebar is used in NcContent.')
```
DorraJaouad, line 1242: Isn't this a useless concat?

susnux: I think its done to make the line shorter^^

ShGKme: It can be replaced with
```
logger.warn('[NcAppSidebar] It looks like you want to use NcAppSidebar with the built-in toggle button. This feature is only available when NcAppSidebar is used in NcContent.')
```

But then it is a very long line.

So, I'd say - no. It allows making it multi-line in the code.

susnux: If its just visually in the editor then you also could just configure your IDE to break lines after X characters, saving one string concat instruction.
But both is fine for me here :)

ShGKme: Anyway, it is not a "useless concat" from the new ESLint rule's perspective 👀

susnux:
I think rebase and squash fixups (DCO) than this should be ready to :runner:

ShGKme:
> I think rebase and squash fixups (DCO) than this should be ready to 🏃

Doing it right now.

ShGKme:
> how did you upgrade?

I put the command in the commit description to make it clear and reproduceable :P

```sh
npm un eslint vue-eslint-parser @nextcloud/eslint-config eslint-plugin-cypress
npm i -D eslint@9 @nextcloud/eslint-config@next
```

I double-checked and I haven't found any problems in dependencies bump.

ShGKme:
- Rabased onto `stable8`
- Squashed `fixup!` commits
- Squashed *some* related and small commits decreasing the amount from 69 to 56. I hope this is a bit less scary now :)

No actual more changes.

ShGKme:
> Its hard to see different changes, you have to scroll more than 60 commits in the history log to get to the next changes.

For me the review simplicity, rebasing and backporting seems to be more important. (To be honest, I don't know in which case I rely on the scrolling length in git). Because it's about the time of other developers during review.

We also have tons of dependabot commits (one commit + one merge per dependency), many small l10n updates (sometimes several times per day).

ShGKme:
> Yes but very related changes: _"Migrate to ESLint 9 and Nextcloud ESLint config v9"_ 😉

This is too large change to work with it as a single change.

The PR to the `main` was complicated to review for me (that also why I postponed the review). And I saw no chance to backport it commit by commit also because of the size of the changes.

PR #7440, opened 2025-09-02
feat: add NcFormGroupButton - an interactive button/link element of NcFormGroup

ShGKme:
### ☑️ Resolves

- Fix: https://github.com/nextcloud-libraries/nextcloud-vue/issues/7123
- An interactive button-like element of a form. Can be used for consistent form items in a stack.
- ⚠️ Current questions: Component name?

### 🖼️ Screenshots

image

### 🏁 Checklist

### DIFF
--- a/src/components/NcFormGroupButton/NcFormGroupButton.vue
+++ b/src/components/NcFormGroupButton/NcFormGroupButton.vue
@@ -260,6 +260,23 @@ export default {
 	flex-direction: column;
 	gap: var(--default-grid-baseline);
 }
+
+/* WIP */
+/* Added for demonstration */
+.sample > * {
+	border-radius: var(--border-radius) !important;
+
+	&:first-child {
+		border-start-start-radius: var(--border-radius-element) !important;
+		border-start-end-radius: var(--border-radius-element) !important;
+	}
+
+	&:last-child {
+		border-end-start-radius: var(--border-radius-element) !important;
+		border-end-end-radius: var(--border-radius-element) !important;
+	}
+}
+/* WIP */
 </style>
 ```
 </docs>


src/components/NcFormItem/NcFormItem.vue
```diff
@@ -0,0 +1,242 @@
...
+	/**
+	 * Additional optional text (sub-line).
+	 * Note: it is not semantically a description. You can still use aria-description or aria-describedby when needed.
+	 */
+	description?: string
+	/**
+	 * RouterLink's to props and a trigger to use RouterLink.
+	 * Note: This takes precedence over the href attribute.
+	 */
+	to?: RouteLocationRaw
```
susnux: Do we need the complexity of adding router support? Otherwise this could have just been a simple `<a>` element

ShGKme: You by complexity you mean the `useButtonLink` - I'd also use it for `NcListItem` (instead of `<a href="#">` and `NcAppnavigationItem`.

I don't have a case for a router link in mind, but for me it seems just a link case in general in an SPA.

Like if you have extended settings in another route. Or a link to the "Note to self".

Otherwise people will use a button with `@click="router.push"`.

susnux: The only thing I do not really like is the name :see_no_evil: 
It is not a form element, it is neither an input nor a button, its an link or list element.

It is not even used in any form, its first usage will be in the app settings, so it would be a settings item.
Or something more generic, but not really any kind of form related.

> I'd also use it for NcListItem

It would be really nice to have it even without any interactive wrapper - but unrelated here.

ShGKme: > It is not even used in any form, its first usage will be in the app settings

I don't see why it cannot be used in any forms if it looks well there. Like in custom modal dialogs with forms.

We are going to extend the new design to other settings/forms later anyway.

So, I'd not stick to `NcAppSettingsDialog` or "settings", because this is not a setting, but a button-like item for a form.

ShGKme: However, I agree that the current name is not perfect.

But I didn't come up with a better naming :(

ShGKme: Ideas are welcome, but better not limit to `NcAppSettingsDialog` or `* settings`, IMO

susnux: For me as a developer "form" is a form not the settings. We explicitly named the settings "*Settings*", so "form" is a real input form like everything you put into `<form>`.

> but better not limit to NcAppSettingsDialog or * settings, IMO

IIRC the new design should be used for app settings and normal settings, so the question is if we want to make this very generic or just have it for that use case.
Because looking at `NcListItem` its horrible because its too generic, it has special flags for everything so we have a Frankenstein for usage in NcAppContent like mail does but also for very simple list items.

So to be honest I would have just called it `NcSettingsItem` just to make it clear it is designed for the settings and not for every use case you can come up with and then we are again in hell of styling.

ShGKme: > For me as a developer "form" is a form not the settings

Would you say that an input element is not a form element if it is used in settings instead of a form with a submit button?

ShGKme: Is there anything that would prevent using new design in any form?

Fieldset, radio groups. All the new components are very classic parts of a form.

Maybe except this one item specifically.

ShGKme:
- Adjusted according to the design review:
  - Icon size: 20px
  - Added border like on other buttons for Nextcloud 32
  - Adjusted paddings
  - Adjusted colours, but with new colours...
- Instead of `description` new props for optional text are `supline` (above the text) and `subline` (below)

ShGKme:
Renamed to `NcFormGroupButton`, where `NcFormGroup` is a visual group of items.
Yet not the final name 👀

ShGKme:
- Updated border color according to the design review

src/components/NcFormGroupButton/NcFormGroupButton.vue
```diff
@@ -0,0 +1,282 @@
+<!--
+  - SPDX-FileCopyrightText: 2021 Nextcloud GmbH and Nextcloud contributors
```
GVodyanov, line 2:
```suggestion
  - SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
```

Looking really cool!

kra-mo:
> ⚠️ Current questions: Component name?

Maybe something like `NcActionField`, or if we would prefer more specific ones later like copy, then `NcBaseField` so that we could have an `NcCopyField` or something?

But in general, I would prefer if we could call all of these form components fields.

susnux:
> Maybe something like NcActionField

This confuses with `NcActions` and all its `NcActionX` sub-components.

> if we would prefer more specific ones later like copy, then NcBaseField so that we could have an NcCopyField or something?

But its not a field like a text input its currently only a copy button or an external link.

> NcFormGroupButton

I think this is ok it follows the Vue naming rules where you suffix the component where it is used.
As this is not a stand-alone component like `NcButton` but must only be used within `NcFormGroup` the name `NcFormGroupButton` seems to be matching for the use case.

kra-mo:
> This confuses with NcActions and all its NcActionX sub-components.

Sure, I did think of that. But at the end of the day, action is such a generic term that I'm not sure anything else could cover all of this. Maybe it's fine to expand our use of the word as well?

But if we would prefer more specific descendants, then base would probably be better anyway.

> But its not a field like a text input its currently only a copy button or an external link.

Field in the context of forms just means an element. At least that's how I view it, and I think it would be a nice way to tie it all up. "A form has fields and the fields are grouped."

> Vue naming rules

Well alright, although if we do establish the convention that you use fields in forms, I feel like that would be clear enough and we would avoid long component names like `NcFormGroupCopyButton` in favor of `NcCopyField`.

You have the final say, but internally as designers I think we'll continue to call them fields so it would be nice to have consistency. At least from a design POV, this isn't really a "button", just a form field/item/row with information and an associated action.

But again, you have the final say. The naming is not that important for me, I just wanted to outline how I view the component :)

sorbaugh:
> > Maybe something like NcActionField
> 
> This confuses with `NcActions` and all its `NcActionX` sub-components.
> 
> > if we would prefer more specific ones later like copy, then NcBaseField so that we could have an NcCopyField or something?
> 
> But its not a field like a text input its currently only a copy button or an external link.
> 
> > NcFormGroupButton
> 
> I think this is ok it follows the Vue naming rules where you suffix the component where it is used. As this is not a stand-alone component like `NcButton` but must only be used within `NcFormGroup` the name `NcFormGroupButton` seems to be matching for the use case.

@ShGKme from a brief talk with @susnux , the PR is fine as it is and needs only to solve the merge conflicts -> review -> merge

ShGKme:
Since other components' names have been settled, the naming will be
- `NcFormBoxButton` - the button/link
  - image
- `NcFormBoxCopyButton` - button with built-in copying functionality
  - image
- `NcFormBoxSwitch` - the switch component for the form box
  - image

Later:
- `NcFormboxFilePicker` ?

ShGKme:
To have less confusion and since the component has been reworked significantly, recreated PR:
- https://github.com/nextcloud-libraries/nextcloud-vue/pull/7779

PR #7441, opened 2025-09-02
feat(NcRadioGroup): add component to group radio buttons

susnux:
### ☑️ Resolves
* resolves https://github.com/nextcloud-libraries/nextcloud-vue/issues/7121

This allows to only add the model value on the wrapper and the radio buttons automatically update it.
Also it has a nicer implementation of the button variant.

### 🖼️ Screenshots
Bildschirmfoto_20250902_153221

### 🚧 Tasks

- [ ] Proper styling (waiting for NcFieldset to reuse)

### 🏁 Checklist


susnux:
Adjusted the label a bit, I think this is a quite solid base for that feature.
As soon as we have `NcFieldset` we can plug it into the `NcRadioGroup` implementation for better / consistent styling.

But for now it should already work as intended.

kra-mo (approved):
I was gonna write a bit about how this would all fit together with the other component changes, but it's honestly not all that relevant to this specific PR.

As it stands, it looks nice from a design POV, approved :)

Still, could we have a conversation about the specific changes being made across components atm? Because I feel like I don't have a clear overview.

kra-mo:
Although there are two things which should probably be changed: The hover color doesn't transition and just snaps instantly and for the button style, the cursor does not change to the pointer like with other buttons.

susnux:
> Although there are two things which should probably be changed: The hover color doesn't transition and just snaps  instantly and for the button style, the cursor does not change to the pointer like with other buttons.

Both resolved.

src/components/NcRadioGroup/NcRadioGroup.vue
```diff
@@ -0,0 +1,198 @@
...
+import { INSIDE_RADIO_GROUP_KEY } from './useNcRadioGroup.ts'
+
+const modelValue = defineModel<string>({ required: false, default: '' })
+
+defineProps<{
+	/**
+	 * If the radio group should be shown as buttons.
+	 * When set `NcRadioGroupButton` must be used in the default slot.
+	 */
+	buttonVariant?: boolean
```
artonge: Can't it be detected based on the content of the slot?

src/components/NcRadioGroupButton/NcRadioGroupButton.vue
artonge, line 1: I got confused by the name, as I thought the component was supposed to be an equivalent to `NcRadioGroup` but for buttons, whereas it is a radio input with the button style. Maybe that is just me, but `NcRadioButton` feels more explicit. Or `NcRadio` with a property `style=button/default`.

artonge (approved):
Looks good

src/components/NcRadioGroup/NcRadioGroup.vue
susnux: we can do this yes - more by using inject / provide instead of parsing the slot content (we messed this already up in actions :sweat_smile: ).
What do you think @ShGKme ?

src/components/NcRadioGroupButton/NcRadioGroupButton.vue
susnux: I am following the vue naming convention here to point out that this component **must** only be used within a radio group: https://vuejs.org/style-guide/rules-strongly-recommended.html#tightly-coupled-component-names

It is not possible (or should not be done) to use this outside as a stand alone component

PR #7448, opened 2025-09-05
fix: revert renaming template refs resulting in a breaking change

ShGKme:
### ☑️ Resolves

- Regression from https://github.com/nextcloud-libraries/nextcloud-vue/pull/7239
  - PR was merged with unresolved active discussion
- Though we may call it not an explicit public API, it is used in apps and unneeded changes. The initial issue could be fixed by renaming internal variables instead.

### 🏁 Checklist


DorraJaouad (approved):
Alright

PR #7470, opened 2025-09-08
fix(usernameToColor): restore original color algorithm

Antreesy:
### ☑️ Resolves

- Revert changes from dropping md5 library in #6657

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
2025-09-08_12h11_23 | 2025-09-08_12h20_20

### 🏁 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,6 +93,7 @@
     "@vueuse/core": "^13.9.0",
     "blurhash": "^2.0.5",
     "clone": "^2.1.2",
+    "crypto-browserify": "^3.12.1",
     "debounce": "^2.2.0",
     "dompurify": "^3.2.6",
     "emoji-mart-vue-fast": "^15.0.5",
--- a/src/functions/usernameToColor/index.ts
+++ b/src/functions/usernameToColor/index.ts
@@ -5,6 +5,7 @@
 
 import type { Color } from '../../utils/colors.ts'
 
+import { createHash } from 'crypto-browserify'
 import { generatePalette } from '../../utils/colors.ts'
 
 /**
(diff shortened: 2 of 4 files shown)


src/functions/usernameToColor/index.ts
```diff
@@ -5,6 +5,7 @@
 
 import type { Color } from '../../utils/colors.ts'
 
+import md5 from 'crypto-js/md5.js'
```
ShGKme: Bundle size is 8.06 kB -> 3.13 kB (gzip)

ShGKme (approved):

src/functions/usernameToColor/index.ts
ShGKme: https://bundlejs.com/?q=crypto-js%2Fmd5.js&treeshake=%5B%7B+default+%7D%5D

Antreesy: Previous algorithm (from md5 lib):

> Bundle size is 6.09 kB -> 2.59 kB (gzip)

package.json
```diff
@@ -93,6 +93,7 @@
     "@vueuse/core": "^13.9.0",
     "blurhash": "^2.0.5",
     "clone": "^2.1.2",
+    "crypto-js": "^4.2.0",
```
susnux: That dependency is quite heavy and not good for tree shaking.
Best (even if it uses node buffers) is probably `md5.js`

susnux: Meaning `crypto-browserify` (and there used `md5.js`) is already used by quite some libraries we ship.
But also `md5` would work as its included by `webdav` package.

susnux (approved):

PR #7471, opened 2025-09-08
fix(NcCheckboxRadioSwitch): ensure label less radio has proper size

susnux:
### ☑️ Resolves

When there is no visual label the checkbox / radio should be square.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Bildschirmfoto_20250908_153835|Bildschirmfoto_20250908_153924

### 🏁 Checklist

### DIFF
--- a/src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
+++ b/src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
@@ -40,7 +40,7 @@
 			</slot>
 		</span>
 
-		<span class="checkbox-content__wrapper">
+		<span v-if="$slots.default || $slots.description" class="checkbox-content__wrapper">
 			<span
 				v-if="$slots.default"
 				:id="labelId"


src/components/NcCheckboxRadioSwitch/NcCheckboxContent.vue
```diff
@@ -228,6 +228,12 @@ export default {
 	&__wrapper {
 		flex: 1 0 0;
 		max-width: 100%;
+
+		&:empty {
+			// hide the container if there is no content
+			// otherwise the outer container will have the gap of the checkbox-content as additional width
+			display: none;
+		}
```
ShGKme: What about not rendering the wrapper at all when there is no content?

susnux: sure also possible

ShGKme:
Btw, where do we use it without the visual label?

susnux:
> Btw, where do we use it without the visual label?

In forms as we add a label on top of it for linear scale (it has a visual label but its external to not mess with the component styles...).
grafik

Antreesy (approved):

PR #7473, opened 2025-09-08
fix(NcDateTimePicker): use proper day names

susnux:
### ☑️ Resolves

- https://github.com/nextcloud-libraries/nextcloud-vue/issues/7469

* ref https://github.com/Vuepic/vue-datepicker/issues/1159

Either this is a bug in the upstream library or wrong documentation. It seems to require the day names in the same order as they would be shown, so first day name is always the `week-start`.

### 🏁 Checklist

### DIFF
--- a/src/components/NcDateTimePicker/NcDateTimePicker.vue
+++ b/src/components/NcDateTimePicker/NcDateTimePicker.vue
@@ -516,9 +516,13 @@ function formatLibraryTime(time: LibraryTimeObject): Date {
 
 // Localization
 
-const dayNames = getDayNamesMin()
-
 const weekStart = getFirstDay()
+
+const dayNames = [...getDayNamesMin()]
+// see https://github.com/Vuepic/vue-datepicker/issues/1159
+for (let i = 0; i < weekStart; i++) {
+	dayNames.push(dayNames.shift() as string)
+}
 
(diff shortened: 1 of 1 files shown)


src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -516,10 +516,14 @@ function formatLibraryTime(time: LibraryTimeObject): Date {
...
-const dayNames = getDayNamesMin()
-
 const weekStart = getFirstDay()
 
+const dayNames = [...getDayNamesMin()]
+// see https://github.com/Vuepic/vue-datepicker/issues/1159
+for (let i = 0; i < weekStart; i++) {
+	dayNames.push(dayNames.shift() as string)
+}
+
```
ShGKme, line 526: Does it mean, if the library got fixed, we need to revert it, otherwise we have an invalid value again?

susnux: I am not sure it is a bug or by design from the optional Function style prop I guess its by design :see_no_evil:

susnux: So yes will keep an eye on the issue

PR #7476, opened 2025-09-09
fix(NcAvatar): make min status size visually accessible

DorraJaouad:
### ☑️ Resolves

- With size below 14px, it becomes visually hard to see emoji. 

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image
image| image

### 🏁 Checklist

### DIFF
--- a/src/components/NcAvatar/NcAvatar.vue
+++ b/src/components/NcAvatar/NcAvatar.vue
@@ -976,16 +976,14 @@ export default {
 		// - ⏹️ Best-fit: the status icon is as large as possible without exceeding the avatar box
 		// See PR for math explanation: PR #6004
 		--avatar-status-size-orbital: calc(var(--avatar-size) * (1 - 1 / sqrt(2)));
-		// Limit the status icon size to the status of a small clickable avatar
+		// Limit the status icon size to minimum font size to keep it readable
 		// Ideally avatars with a smaller should not be used with the status icon at all
-		--avatar-status-size-min: calc(var(--default-clickable-area) * (1 - 1 / sqrt(2)));
+		--avatar-status-size-min: var(--font-size-small);
 		--avatar-status-size: max(var(--avatar-status-size-orbital), var(--avatar-status-size-min));
-		// Because the status icon size is limited, smaller avatar requires a position offset to keep the status icon orbital
-		--avatar-status-icon-position: min(0px, (var(--avatar-status-size-orbital) - var(--avatar-status-size)) / 2);
 		box-sizing: border-box;
 		position: absolute;
-		inset-inline-end: var(--avatar-status-icon-position);
-		inset-block-end: var(--avatar-status-icon-position);
+		inset-inline-end: 0;
+		inset-block-end: 0;
 		height: var(--avatar-status-size);
 		width: var(--avatar-status-size);
 		line-height: 1;


ShGKme (dismissed):

susnux:
So the status icon is hidden if the avatar is below 50px? Then this is a no go as this will break many places, it must at least work with default clickable area (34px) avatar size.

DorraJaouad:
> So the status icon is hidden if the avatar is below 50px?

No, it is just bigger from 11px -> around 14px

susnux:
Or maybe I read it wrong, it sets the min size?
So it will not longer be perfectly an orbit?

Then we need to fix this in a different way (the original issue) as this then breaks again some places where it was fixed by the original PR.

src/components/NcAvatar/NcAvatar.vue
```diff
@@ -976,9 +976,9 @@ export default {
 		// - ⏹️ Best-fit: the status icon is as large as possible without exceeding the avatar box
 		// See PR for math explanation: PR #6004
 		--avatar-status-size-orbital: calc(var(--avatar-size) * (1 - 1 / sqrt(2)));
-		// Limit the status icon size to the status of a small clickable avatar
+		// Limit the status icon size to the status of a 50px avatar
 		// Ideally avatars with a smaller should not be used with the status icon at all
```
susnux, line 980: This comment makes no sense anymore. Where do you use a 50px avatar?
I think most cases use default-clickable-area and thus it must work wit 34px

Antreesy:
Also it's cropped on 'After' screenshot 🙈 

IMO we should define a minimum size of an icon, and 'break' orbital rule the way, that icon is moving slightly inside the avatar, not going out of square frame

src/components/NcAvatar/NcAvatar.vue
```diff
@@ -976,9 +976,9 @@ export default {
...
-		--avatar-status-size-min: calc(var(--default-clickable-area) * (1 - 1 / sqrt(2)));
+		--avatar-status-size-min: calc(50px * (1 - 1 / sqrt(2)));
```
susnux: This is more than 14px, why a random number like 50px?
I think it makes more sense to use either clickable area large which is a value you might choose for the avatar when not using magic numbers:

```suggestion
		--avatar-status-size-min: calc(var(--clickable-area-large) * (1 - 1 / sqrt(2)));
```

Or because you want to limit the emoji to a readable size, then you do not need the avatar size here at all but just use the min. accessible size:

```suggestion
		--avatar-status-size-min: var(--font-size-small);
```

susnux:
> IMO we should define a minimum size of an icon, and 'break' orbital rule the way, that icon is moving slightly inside the avatar, not going out of square frame

Yes that exactly is my problem with this change here, because it will break server styles again.
Where we need the avatar to fit in a square with status icon when the avatar has a size of `--default-clickable-area`.

I think the best solution is to limit the icon size to `--font-size-small` which is the smallest size for icons and text.
And then calculate the icon position based on the icon size which can be simplified to just set the *insets* both to `0px` so it will always be on the correct location and for larger avatars the orbital behavior is preserved.

src/components/NcAvatar/NcAvatar.vue
DorraJaouad: font size seems to be optimal

ShGKme:
Then we probably need to only show the status when the avatar is not too small.

image

ShGKme:
Also, initials overlap the status

susnux (approved):
Seems to work fine now also for default clickable area

susnux:
> Also, initials overlap the status

But not related to this PR, the order of elements is wrong, the status icon element should be the last one.

ShGKme (approved):

ShGKme:
> But not related to this PR

Kind of related. It became more visible with the new position and size on the status icon.

Antreesy (approved):
Added lib screenshots. Fixing initials would be moving 9 lines around, so shouldn't be a problem

PR #7483, opened 2025-09-10
fix(NcRadioGroup): fieldsets always need a label for accessibility

susnux:
### ☑️ Resolves

- follow up on #7441 

### 🏁 Checklist


src/components/NcRadioGroup/NcRadioGroup.vue
```diff
@@ -43,6 +48,14 @@ provide(INSIDE_RADIO_GROUP_KEY, computed(() => ({
 	onUpdate,
 })))
 
+watchEffect(() => {
+	if (!props.ariaLabel && !props.label) {
+		warn('[NcRadioGroup] Either a visual "label" or an "ariaLabel" must be specified.')
+	} else if (props.ariaLabel !== undefined && props.label) {
+		warn('[NcRadioGroup] If a visual "label" is used then no "ariaLabel" should be specified, the accessible name is already inferred from the "label".')
```
Antreesy: Do we need to warn in this case? Or it should be only one of them?

susnux: Its a potential pitfall, as when there is a visual label there should be no aria-label

src/components/NcRadioGroup/NcRadioGroup.vue
```diff
@@ -6,13 +6,18 @@
...
 import { INSIDE_RADIO_GROUP_KEY } from './useNcRadioGroup.ts'
 
 const modelValue = defineModel<string>({ required: false, default: '' })
 
-defineProps<{
+const props = defineProps<{
+	/**
+	 * If no visual `label` is set, then at least this aria label of the radio group must be set for accessibility.
+	 */
+	ariaLabel?: string | undefined
```
ShGKme: Instead of `aria-label`, we can always render the required `legend`, but allow to make it hidden in an additional boolean prop like.

Then it simplifies checking for correct labelling and simplifies a general fieldset usage, for examplem, by providing IDRef for its heading.

susnux: Adjusted, as this is a nice suggestion :)

src/components/NcRadioGroup/NcRadioGroup.vue
susnux: adjusted

src/components/NcRadioGroup/NcRadioGroup.vue
```diff
@@ -43,6 +49,12 @@ provide(INSIDE_RADIO_GROUP_KEY, computed(() => ({
 	onUpdate,
 })))
 
+watchEffect(() => {
+	if (!props.label) {
+		warn('[NcRadioGroup] For accessibility reasons a proper "label" has to be set.')
+	}
+})
+
```
ShGKme: This `watch` duplicates Vue's built-in required prop check...

susnux: Vue only checks if a value is passed but it does not check if the value is truethy aka the string is not empty.

Antreesy: Can it be done just in setup() or a mounted() hook? should be enough for a dev to catch it

susnux: > Can it be done just in setup() or a mounted() hook? should be enough for a dev to catch it

Sure but there should be no overhead because if the label does not change the watch is not executed.
If the label changes the `onMounted` hook would not catch it.
We can also remove this completely because it would only trigger if developers "hack" this explicitly.

Antreesy: Sounds good, since it's required now

susnux: done

susnux:
squashed fixup commits.

PR #7501, opened 2025-09-13
fix: adjust input border styles for dark theme and NcSelect

susnux:
### ☑️ Resolves

- resolves https://github.com/nextcloud-libraries/nextcloud-vue/issues/7472
- partially (NcSelect) for https://github.com/nextcloud-libraries/nextcloud-vue/issues/7386

Move input border styles into a scss mixin to prevent input like elements get out-of-sync designwise.
Then apply correct dark theme changes as requested and for showcasing this new mixin also apply it for NcSelect.

### 🏁 Checklist


Antreesy:
height is off, according to new snapshots? Looks  larger than 34px, and trailing icon isn't aligned

susnux:
> height is off, according to new snapshots? Looks larger than 34px, and trailing icon isn't aligned

I looked into this in styleguide but there it is correct :eyes:

src/components/NcInputField/NcInputField.vue
```diff
@@ -161,6 +162,8 @@ const attrs = useAttrs()
 
 const inputElement = useTemplateRef('input')
 
+const isDarkTheme = useIsDarkTheme()
```
ShGKme: There is no way to style the border without setting class from JS?

src/assets/input-border.scss
```diff
@@ -0,0 +1,53 @@
+/*!
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+ 
+ @mixin boxShadow($borderColor) {
```
ShGKme:
```suggestion
@mixin boxShadow($borderColor) {
```

src/components/NcInputField/NcInputField.vue
susnux: How would we that work? The problem is we need to care about the browser preferred theme but also user override e.g. with theme provider. So I do not see a good solution without using the composable.

ShGKme: `@media` + `[data-theme-*]` on pure styles?

It can be not the entire style property, but only a custom property with the value. Or even only the variable part. Currently, If I didn't miss anything, only `1px` changes to `-1px` in a single place.

susnux: done

src/assets/input-border.scss
```diff
@@ -0,0 +1,58 @@
...
+
+	&:hover:not([disabled]) {
+		box-shadow: 0 0 0 1px $borderColor;
+	}
+
+	@at-root [data-theme-default] #{&} {
+		@media (prefers-color-scheme: dark) {
+			--input-border-box-shadow: var(--input-border-box-shadow-dark);
+		}
+	}
```
ShGKme: Is it guaranteed to have `data-theme-default` attribute?
Won't a plain media query also work, because explicit `[data-theme-*]` redefines the value?

```suggestion
	@media (prefers-color-scheme: dark) {
		--input-border-box-shadow: var(--input-border-box-shadow-dark);
	}
```

susnux: done

susnux:
@jancborchardt quickly mocked in the styleguide:

before | after
---|---
Screenshot 2025-11-29 at 13-16-04 Nextcloud Vue Style Guide|Screenshot 2025-11-29 at 13-14-53 Nextcloud Vue Style Guide

susnux:
(rebased to squash fixup)

PR #7514, opened 2025-09-15
refactor(NcModal): migrate to Typescript and script-setup + feat(useHotKey): add new option `allowInModal`

susnux:
### ☑️ Resolves

- Large part of #841 

This consists of multiple refactoring as you can review individually with each commit.
Basically this migrates the component to Typescript and then to script-setup with new consistent tag order.
By doing so also removed the last JS util (Timer) and just uses the `useIntervalFn` from vueuse.

### 🏁 Checklist


susnux:
(just a rebase for CI)

ShGKme:
I'll check on the morning

src/components/NcModal/NcModal.vue
```diff
@@ -1126,6 +1012,8 @@ $radius: 15;
 $pi: 3.14159265358979;
 
 .modal-mask .play-pause-icons {
+	--slideshow-duration: v-bind('cssSlideshowDelay');
+
```
ShGKme: Is this variable needed? Cannot we use `v-bind('cssSlideshowDelay')` directly where the variable is used? Otherwise it looks like creating a CSS variable just to define another CSS variable just to use it in a single place.

src/components/NcModal/NcModal.vue
```diff
@@ -3,885 +3,549 @@
...
+	actions?: Slot
+
+	/**
+	 * The modal content to show.
+	 */
+	default?: Slot
+}>()
+
+const modalId = createElementId()
+const maskElement = useTemplateRef<HTMLDivElement>('mask')
```
ShGKme:
```suggestion
const maskElement = useTemplateRef('mask')
```

src/components/NcModal/NcModal.vue
```diff
@@ -3,885 +3,549 @@
...
+
+const internalShow = ref(true)
```
ShGKme: It seems to me we can simplify the component by just using `defineModel`. Then we don't need the second state, and the rest around the value is simpler. Is there anything blocking it I'm missing?

src/components/NcModal/NcModal.vue
```diff
@@ -3,885 +3,549 @@
...
+const { stop: stopSwipe } = useSwipe(maskElement, {
+	onSwipeEnd: handleSwipe,
+})
+onUnmounted(stopSwipe)
+
+// Setup hotkeys (keyboard navigation)
+useHotKey('Escape', () => {
+	const trapStack = getTrapStack()
+	// Only close the most recent focus trap modal
+	if (trapStack.length === 0 || trapStack[trapStack.length - 1] === focusTrap) {
```
ShGKme:
```suggestion
	if (trapStack.at(-1) === focusTrap) {
```

src/components/NcModal/NcModal.vue
```diff
@@ -3,885 +3,549 @@
...
-		showModal() {
-			this.modal = true
-		},
-		closeModal() {
-			this.modal = false
-		}
+import { mdiChevronLeft, mdiChevronRight, mdiClose, mdiPause, mdiPlay } from '@mdi/js'
+import { useIntervalFn, useSwipe } from '@vueuse/core'
+import { createFocusTrap } from 'focus-trap'
+import { computed, nextTick, onMounted, onUnmounted, ref, toRef, useTemplateRef, warn as VueWarn, watch, watchEffect } from 'vue'
```
ShGKme: > warn as VueWarn

Seems like we can just use `warn` now

ShGKme (approved):
Nitpicks only.

Thanks for splitting into commits.

src/components/NcModal/NcModal.vue
```diff
@@ -3,885 +3,549 @@
...
+
+useHotKey(['ArrowLeft', 'ArrowRight'], (event) => {
+	// Ignore arrow navigation, if there is a current focus outside the modal.
+	// For example, when the focus is in Sidebar or NcActions' items,
+	// arrow navigation should not be intercepted by modal slider
+	if (document.activeElement && !maskElement.value!.contains(document.activeElement)) {
+		return
+	}
+
+	if (event.key === 'ArrowLeft' || isRtl) {
```
Antreesy: We want to flip for RTL, so it should be:

`A ^ B = (A || B) && !(A && B)` - also `A !== B`

A | B | A ^ B | Current (A \|\| B)
-- | -- | -- | --
1 | 1 | 0 | 1 ⚠️ 
0 | 1 | 1 | 1
1 | 0 | 1 | 1
0 | 0 | 0 | 0

src/components/NcModal/NcModal.vue
```diff
@@ -3,885 +3,549 @@
...
-			this.modal = false
+/**
+ * handle the swipe event
+ *
+ * @param e - The touch event
+ * @param direction - Swipe direction
+ */
+function handleSwipe(e: TouchEvent, direction: UseSwipeDirection) {
+	if (!props.disableSwipe) {
+		if (direction === 'left' || (direction === 'right' && isRtl)) {
```
Antreesy: Same A^B here

src/components/NcModal/NcModal.vue
```diff
@@ -1215,5 +871,160 @@ $pi: 3.14159265358979;
...
+### Usage of popover in modal
+
+* Set container property to .modal-mask to inject popover context of the modal:
+
+```vue
+<template>
+	<div>
+		<NcButton @click="showModal">Show Modal</NcButton>
+		<NcModal v-if="modal" @close="closeModal" size="small" class="emoji-modal">
+			<NcEmojiPicker container=".emoji-modal" @select="select">
```
Antreesy: Example is no longer working, caused by teleport? 

Image

src/components/NcModal/NcModal.vue
```diff
@@ -1215,5 +871,160 @@ $pi: 3.14159265358979;
...
+	},
+	computed: {
+		currentPage() {
+			return PAGES[this.page]
+		},
+	}
+}
+</script>
+<style scoped>
+.modal__content {
```
Antreesy, line 971: Styles are leaking to another examples

Antreesy (commented):
Otherwise nice work!

src/components/NcModal/NcModal.vue
susnux: Good catch!

```suggestion
	if ((event.key === 'ArrowLeft') ^ isRtl) {
```

src/components/NcModal/NcModal.vue
```diff
@@ -3,885 +3,549 @@
...
+ * handle the swipe event
+ *
+ * @param e - The touch event
+ * @param direction - Swipe direction
+ */
+function handleSwipe(e: TouchEvent, direction: UseSwipeDirection) {
+	if (!props.disableSwipe) {
+		if (direction === 'left' || (direction === 'right' && isRtl)) {
+			nextSlide(e)
+		} else if (direction === 'right') {
```
susnux:
```suggestion
		if (direction !== 'left' && direction !== 'right') {
			return
		}

		if ((direction === 'left') ^ isRtl) {
			nextSlide(e)
		} else {
```

src/components/NcModal/NcModal.vue
Antreesy: not a fan of bitwise, but yes, this should also work)
Eslint agrees with me, it seems)

```js
if ((event.key === 'ArrowLeft') !== isRtl) {
	previous
}
```

```js
if ((direction === 'left') !== isRtl) {
	next
}
```

susnux: Do you prefer `(event.key === 'ArrowLeft') !== isRtl` ?
I think here the `^` makes it a bit clearer, no?

Antreesy: well, here we're aware both sides are boolean, so bitwise operator should work correctly.
I'm also fine with suppressing the linter here, with a small comment

susnux:
Added commits fixing the review comments

src/components/NcModal/NcModal.vue
```diff
@@ -3,885 +3,550 @@
...
+// Setup hotkeys (keyboard navigation)
+useHotKey('Escape', () => {
+	const trapStack = getTrapStack()
+	// Only close the most recent focus trap modal
+	if (trapStack.at(-1) === focusTrap) {
+		close()
+	}
+})
+
+useHotKey(['ArrowLeft', 'ArrowRight'], (event) => {
```
Antreesy, line 238: i just realized, that useHotKey is disabled if any modal is opened 🙈 

https://github.com/nextcloud-libraries/nextcloud-vue/blob/70edb2234f8c0dcad876f003e1d008112140d73c/src/composables/useHotKey/index.ts#L62

This is to prevent hotkeys triggered from components, when modal is opened above them

Antreesy (approved):
Otherwise fine. Maybe we'll keep native eventListeners for now?

src/components/NcModal/NcModal.vue
susnux: Pushed a fix

susnux:
@ShGKme @Antreesy please check the new commit (latest)

Antreesy (approved):

susnux:
(rebased to check CI)

susnux:
/backport 0bdf7ec4e8aa58b47f201f93a850bf0725e5cad2 to stable8

PR #7515, opened 2025-09-15
refactor(NcAppSettingsDialog): migrate to Typescript and script-setup

susnux:
### ☑️ Resolves

* for #841 

Migrate component to Typescript and script-setup with consistent tag order.
Also prepare for migrating the section component by providing an internal composable for the injection of the register methods.

Best to review commit by commit.

### 🏁 Checklist


src/components/NcAppSettingsDialog/useAppSettingsDialog.ts
```diff
@@ -0,0 +1,38 @@
...
+ * This is to be used in the app settings sections.
+ */
+export function useAppSettingsDialog() {
+	const registerSection = inject(REGISTER_SECTION_KEY)
+	const unregisterSection = inject(UNREGISTER_SECTION_KEY)
+
+	return {
+		registerSection,
+		unregisterSection,
+	}
```
ShGKme: Do we need 2 injections for providing a single component's context?

src/components/NcAppSettingsDialog/useAppSettingsDialog.ts
```diff
@@ -0,0 +1,38 @@
...
+import { inject } from 'vue'
+
+/**
+ * Register a new section on the app settings dialog
+ *
+ * @param id - The section ID
+ * @param name - The section name
+ * @param icon - Optional icon component
+ */
+export const REGISTER_SECTION_KEY: InjectionKey<(id: string, name: string, icon?: VNode[]) => void> = Symbol.for('NcAppSettingsDialog.registerSection')
```
ShGKme: To follow the convention of other components:
```suggestion
export const REGISTER_SECTION_KEY: InjectionKey<(id: string, name: string, icon?: VNode[]) => void> = Symbol.for('NcAppSettingsDialog:registerSection')
```

src/components/NcAppSettingsDialog/useAppSettingsDialog.ts
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Unregistered from the app settings dialog to remove it from dialog
+ *
+ * @param id - The section ID
+ */
+export const UNREGISTER_SECTION_KEY: InjectionKey<(id: string) => void> = Symbol.for('NcAppSettingsDialog.unregisterSection')
```
ShGKme:
```suggestion
export const UNREGISTER_SECTION_KEY: InjectionKey<(id: string) => void> = Symbol.for('NcAppSettingsDialog:unregisterSection')
```

src/components/NcAppSettingsSection/NcAppSettingsSection.vue
```diff
@@ -15,9 +15,15 @@
...
+import { REGISTER_SECTION_KEY, UNREGISTER_SECTION_KEY } from '../NcAppSettingsDialog/useAppSettingsDialog.ts'
+
 export default {
 	name: 'NcAppSettingsSection',
-	inject: ['registerSection', 'unregisterSection'],
+
+	inject: {
+		registerSection: { from: REGISTER_SECTION_KEY },
+		unregisterSection: { from: UNREGISTER_SECTION_KEY },
+	},
```
ShGKme: Why not use the new composable?

src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
```diff
@@ -3,6 +3,302 @@
...
+
+provide(REGISTER_SECTION_KEY, registerSection)
+provide(UNREGISTER_SECTION_KEY, unregisterSection)
+
+const settingsScrollerElement = useTemplateRef('settingsScroller')
+watch(settingsScrollerElement, (ref) => {
+	if (ref) {
+		ref.onscroll = handleScroll
+	}
+})
```
ShGKme: Cannot we add an event listener via `v-on` directive?

src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
```diff
@@ -3,6 +3,302 @@
...
+	},
+
+	/**
+	 * Additional elements to add to the focus trap
+	 */
+	additionalTrapElements: {
+		type: Array as PropType<(string | HTMLElement)[]>,
+		default: () => [],
+	},
+})
```
ShGKme: On JS, not TS?

src/components/NcAppSettingsDialog/useAppSettingsDialog.ts
susnux: We can also inject only once context object - I just migrated it and did not change this logic :)

src/components/NcAppSettingsDialog/useAppSettingsDialog.ts
susnux: for context: My reasoning was `object.property` like JS. But yes consistency would be better!

src/components/NcAppSettingsDialog/useAppSettingsDialog.ts
susnux: done

src/components/NcAppSettingsSection/NcAppSettingsSection.vue
susnux: done!

src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
susnux: done!

src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
susnux: Sorry, what do you mean?

src/components/NcAppSettingsDialog/useAppSettingsDialog.ts
ShGKme: Adding a composable is already more, than a pure migration :)

src/components/NcAppSettingsDialog/useAppSettingsDialog.ts
ShGKme: But it is not a property, it is rather a namespace

src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
ShGKme: It has JS runtime props definition, not TS (type-based) definition:

```suggestion
const props = withDefaults(defineProps<{
	/**
	 * Shows the navigation on desktop if true
	 */
	showNavigation: boolean,

	/**
	 * Selector for the popover container
	 */
	container: string,

	/**
	 * Name of the settings
	 */
	name: string,

	/**
	 * Additional elements to add to the focus trap
	 */
	additionalTrapElements: (string | HTMLElement)[],
}>(), {
	showNavigation: false,
	container: 'body',
	name: '',
	additionalTrapElements: () => [],
})
```

ShGKme: Or with 3.5 props destruction

src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
```diff
@@ -3,6 +3,299 @@
   - SPDX-License-Identifier: AGPL-3.0-or-later
 -->
 
+<script setup lang="ts">
+import type { PropType, Slot, VNode } from 'vue'
+
+import debounce from 'debounce'
+import { computed, provide, ref, useTemplateRef, warn, watch } from 'vue'
```
ShGKme:
```suggestion
import { computed, provide, ref, useTemplateRef, warn } from 'vue'
```

src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
susnux: Yes see next commits

susnux: oh wait forgot to push

ShGKme (approved):

susnux:
rebased with `--autosquash`

susnux:
> backport to stable8

specific reason? We did not backport all the other Typescript refactoring and some of it does not work with Vue 2.
(first commit will work though).

ShGKme:
> specific reason?

Still the same as you asked the last time I requested refactoring backport :)
- Make developers' life easier by simplifying backporting of features/fixes, prevent the suffering from refactoring
- Improve DX on `stable8` which also wins from TS support

It doesn't look like we will migrate to Vue 3 completely very soon, so we need to have all the features and fixes on both branches. Having such divergent branches is painful for developers. Especially when there are more and more changes after a large refactoring.

> We did not backport all the other Typescript refactoring

True. They are missing. But it is a fixable problem. And we don't need to do it completely manually.

> some of it does not work with Vue 2

Same for the other PRs using 3.3..3.5 features.

susnux:
> It doesn't look like we will migrate to Vue 3 completely very soon

If everything went well 33 will be the last that (partly) ship Vue 2 (in server core) :)

ShGKme:
> If everything went well 33 will be the last that (partly) ship Vue 2 (in server core) :)

I am unsure if all the teams and apps will migrate until 33. If not, then it is at least half a year of supporting 2 branches with constant updates.

IMO, it is much easier and safer to adjust a migration to 2.7 compatible state (not so much difference) than to support two different versions of each component.

Especially for developers occasionally working on the library and contributors.

Refactoring should not bring more pain than it solves 👀

PR #7516, opened 2025-09-15
refactor(NcAppSidebarTabs): use custom components for tab buttons

susnux:
### ☑️ Resolves

NcCheckboxRadioSwitch with button styling is deprecated. The tabs also had overridden styles so instead of reusing the radio component and hacking styles we create a small custom component for this use case.
Also the wcag examples use buttons and not radio.

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppSidebar/NcAppSidebarTabs.vue
+++ b/src/components/NcAppSidebar/NcAppSidebarTabs.vue
@@ -42,6 +42,7 @@
 </template>
 
 <script>
+import { getCanonicalLocale } from '@nextcloud/l10n'
 import NcAppSidebarTabsButton from './NcAppSidebarTabsButton.vue'
 
 export default {
@@ -82,7 +83,7 @@ export default {
 
 	emits: ['update:active'],
 
-	data() {
+	data(props) {
 		return {
 			/**
 			 * Tab descriptions from the passed NcSidebarTab components' props to build the tab navbar from.
@@ -91,7 +92,7 @@ export default {
 			/**
 			 * Local active (open) tab's ID. It allows to use component without v-model:active
 			 */
-			activeTab: '',
+			activeTab: props.active,
 		}
 	},
 
@@ -115,6 +116,12 @@ export default {
 	},
 
 	watch: {
+		tabs() {
+			if (this.active) {
+				this.updateActive()
+			}
+		},
+
 		active(active) {
 			// Prevent running it twice
 			if (active !== this.activeTab) {
@@ -197,11 +204,9 @@ export default {
 		 * Update the current active tab
 		 */
 		updateActive() {
-			this.activeTab = (this.active && this.tabs.some((tab) => tab.id === this.active))
+			this.activeTab = (this.active && this.tabs.some(({ id }) => id === this.active))
 				? this.active
-				: this.tabs.length > 0
-					? this.tabs[0].id
-					: ''
+				: (this.tabs[0]?.id ?? '')
 		},
 
 		/**
@@ -213,7 +218,7 @@ export default {
 			this.tabs.push(tab)
 			this.tabs.sort((a, b) => {
 				if (a.order === b.order) {
-					return OC.Util.naturalSortCompare(a.name, b.name)
+					return a.name.localeCompare(b.name, [getCanonicalLocale()])
 				}
 				return a.order - b.order
 			})


src/components/NcAppSidebar/NcAppSidebarTabsButton.vue
```diff
@@ -0,0 +1,114 @@
...
+}>()
+</script>
+
+<template>
+	<button
+		:class="[$style.sidebarTabsButton, $style.specificityHelper, {
+			[$style.sidebarTabsButton_active]: active,
+		}]"
+		role="tab"
+		:aria-selected="active ? 'true' : 'false'"
```
ShGKme: Same as
```suggestion
		:aria-selected="active"
```

tests/unit/components/NcAppSidebar/NcAppSidebarTabs.spec.js
ShGKme, line 1: Why?

tests/component/components/NcAppSidebar/AppSidebarTabs.story.vue
ShGKme, line 1: `AppSidebarTabs.story.vue` -> `NcAppSidebarTabs.story.vue`

Same as with test files, having one thing in the same name everywhere simplifies searching for related things and files.

E.g. CTRL+P or Shift+Shift in IDE, type `NcAppSidebarT` and you immediately see the component, its tests and its stories.

src/components/NcAppSidebar/NcAppSidebarTabs.vue
```diff
@@ -20,30 +20,15 @@
...
-				</span>
-				<template #icon>
-					<NcVNodes :vnodes="tab.renderIcon()">
-						<span class="app-sidebar-tabs__tab-icon" :class="tab.icon" />
-					</NcVNodes>
-				</template>
-			</NcCheckboxRadioSwitch>
+				:active="activeTab === tab.id"
+				:aria-controls="`tab-${tab.id}`"
+				:tab="tab"
```
ShGKme: :P
```suggestion
				:tab
```

src/components/NcAppSidebar/NcAppSidebarTabsButton.vue
```diff
@@ -0,0 +1,107 @@
...
+
+import NcVNodes from '../NcVNodes/NcVNodes.vue'
+
+const active = defineModel<boolean>('active', { required: true })
+
+defineProps<{
+	/**
+	 * The sidebar tab this button controls
+	 */
+	tab: typeof NcAppSidebarTab
```
ShGKme:
```suggestion
	tab: InstanceType<typeof NcAppSidebarTab>
```

ShGKme (commented):
Nice

src/components/NcAppSidebar/NcAppSidebarTabsButton.vue
```diff
@@ -0,0 +1,107 @@
...
+			<NcVNodes :vnodes="tab.renderIcon()">
+				<span :class="[$style.sidebarTabsButton__legacyIcon, tab.icon]" />
+			</NcVNodes>
+		</span>
+	</button>
+</template>
+
+<style module lang="scss">
+// specificityHelper to increase CSS specificity to override server styles
+.specificityHelper {
```
ShGKme: If that was to remove the `<button>` styles, you can add `button-vue` class, which is used to signal that the button is not a plain native button but a Vue component.

Then we don't need to remove styles, because they are not applied in the first place.

src/components/NcAppSidebar/NcAppSidebarTabsButton.vue
```diff
@@ -0,0 +1,107 @@
...
+  - SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<script setup lang="ts">
+import type NcAppSidebarTab from '../NcAppSidebarTab/NcAppSidebarTab.vue'
+
+import NcVNodes from '../NcVNodes/NcVNodes.vue'
+
+const active = defineModel<boolean>('active', { required: true })
```
ShGKme: Not important for an internal component, but semantically it represents the `selected` button (tab), not an active.

src/components/NcAppSidebar/NcAppSidebarTabsButton.vue
susnux: This does work? Has Vue some magic handling here? Because its not a prop its a string attribute

tests/component/components/NcAppSidebar/AppSidebarTabs.story.vue
susnux: :+1:

src/components/NcAppSidebar/NcAppSidebarTabsButton.vue
susnux: it is. Thats so crappy - instead we should scope all elements to get server styles e.g. only style elements that have `nc` class or similar.

src/components/NcAppSidebar/NcAppSidebarTabsButton.vue
susnux: isnt that the same? Its the currently active tab which is the selected tab? For me this is equal?

src/components/NcAppSidebar/NcAppSidebarTabsButton.vue
ShGKme: See: https://v3-migration.vuejs.org/breaking-changes/attribute-coercion.html

src/components/NcAppSidebar/NcAppSidebarTabsButton.vue
susnux: I adjusted it anyways, but would be still interested in the difference

ShGKme: `active` is some sort our own interface name, while `selected` represents native meaning. For example, it sets `aria-selected` attribute to mark the tab as `selected`.

src/components/NcAppSidebar/NcAppSidebarTabsButton.vue
ShGKme: > Thats so crappy

Yes.

I'd think about it together with a general global styles overview and refactoring for the next release.

But requiring a class is breaking and makes all general buttons ugly native buttons.

I'd say, we can go the opposite way and introduce either a general "no-server-styles" class instead of `button-vue`, or a new `reset` class.

src/components/NcAppSidebar/NcAppSidebarTabsButton.vue
susnux: its more like "set tab as selected" and "set tab as active" is / sounds equivalent to me :eyes:

ShGKme: I mostly referred to this:
image

ShGKme (changes requested):
The design has changed
Before | After
--- | ---
image | image

I don't know if it was an intentional change, but it is not a refactoring 👀

kra-mo (dismissed):
From a design POV: This is a nice change overall, but there are some smaller things that could be improved and in general the styling could be tweaked a bit to avoid having to do that retroactively.

First, while the buttons are focused, the selected item stays highlighted, but when no longer focused, the highlight disappears.

The highlight is actually desired and consistent with the [list selection changes](https://github.com/nextcloud-libraries/nextcloud-vue/issues/7222), it would be nice if it was kept even after unfocusing.

Second, the bottom edge should get the same border radius as the top one.

Third, as a result, the blue line indicating highlight should be rounded and less wide.

Fourth, the active item changing to be bold is quite distracting to me. I feel like we could drop that behavior as we don't do it everywhere for selection either way.

And fifth, a nitpick:
When hovering, this thing that happens because of the border radius looks weird:

Image
 
There should be at least a 2px gap here to avoid this. The container itself could also be more rounded.

Here is how it should look like with these changes:

Image


Also, the focus ring is messed up in Safari:

Image

kra-mo:
Also, I don't want to ask too much, it would really just be a nice thing that we could do and definitely not necessary, but…

it would be cool if the blue line could do a little animation, growing in width when selected instead of just suddenly appearing, like so:

https://github.com/user-attachments/assets/5a178c9e-b126-4e9d-a132-236b09d65a94

jancborchardt (commented):
Very nice! Agree with @kra-mo 's assessment and also @ShGKme ’s remark about the text and icon switching places. The text should stay below the icon as we have elsewhere like in the app nav and grid view as well. :)

ShGKme:
@kra-mo This is an internal code refactoring, it does not suppose to introduce a new design or change anything for users. Only how the component is implemented inside.

ShGKme:
If we want to re-design tabs, let's create a new issue

kra-mo:
> @kra-mo This is an internal code refactoring, it does not suppose to introduce a new design or change anything for users. Only how the component is implemented inside.

Fair enough, I just thought that maybe the scope could be expanded if it is being reimplemented anyway so that we won't have to change things retroactively. I got my hopes up a bit when a review was requested from Design :)

But I wouldn't block this on any of these apart from making it consistent with the previous design.

Antreesy:
Speaking of legit nitpicks: buttons are now work differently with overflowing content
Nice addition (but also not a part of refactoring) would be to extend the active button to fit all content (see bottom screenshot

B | A
-- | --
image | image
image | ?

kra-mo:
> If we want to re-design tabs, let's create a new issue

Sure, opened #7520

susnux:
> The design has changed

Yes wrong order during rebases, this PR is not supposed to change any design.
It should only refactor internal component usage.

susnux:
Should now be visually the same as before.

ShGKme:
What was changed in the force-push? :)

src/components/NcAppSidebar/NcAppSidebarTabsButton.vue
```diff
@@ -0,0 +1,111 @@
...
+
+	* {
+		cursor: default;
+	}
+}
+
+.sidebarTabsButton__name {
+	font-weight: normal;
+	overflow: hidden;
+	text-overflow: ellipsis;
```
susnux, line 93: @ShGKme `--autosquash` and those two lines for the proper overflow. As well as name and icon switch position back to normal.

tests/unit/components/NcAppSidebar/NcAppSidebarTabs.spec.js
susnux: moved to component tests

src/components/NcAppSidebar/NcAppSidebarTabsButton.vue
```diff
@@ -0,0 +1,111 @@
...
+	</button>
+</template>
+
+<style module lang="scss">
+.sidebarTabsButton {
+	border: none;
+	border-bottom: var(--default-grid-baseline) solid transparent !important;
+	border-radius: var(--border-radius-small);
+	background-color: var(--color-main-background);
+	color: var(--color-main-text);
```
ShGKme, line 48: Setting `font-size` is missing.

```suggestion
	color: var(--color-main-text);
	font-size: var(--default-font-size);
```

ShGKme (changes requested):

src/components/NcAppSidebar/NcAppSidebarTabsButton.vue
susnux: To reset? Normally it should already be set by parent (as its set to `inherit` for all other elements by Nextcloud styles).

ShGKme: It was not set for a button, so it had `13.333px` font size.

ShGKme (approved):

src/components/NcAppSidebar/NcAppSidebarTabsButton.vue
susnux: For me it had `inherit` set and then it was set implicitly to 15 :eyes:

susnux:
rebase with autosquash.

ShGKme:
@jancborchardt This is an internal implementation refactoring. Not visual component update. There supposed to be no visual changes.

src/components/NcAppSidebar/NcAppSidebarTabsButton.vue
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	* {
+		cursor: default;
+	}
+}
+
+.sidebarTabsButton__name {
+	font-weight: normal;
+	overflow: hidden;
```
Antreesy, line 92: `white-space: nowrap;` seems to be missing
image

susnux: done, good catch!

susnux:
rebased

PR #7524, opened 2025-09-17
fix(NcPopover): revert disabling close on click outside by default

ShGKme:
### ☑️ Resolves

- Close NcPopover on click outside by default
- Regression from: https://github.com/nextcloud-libraries/nextcloud-vue/pull/6802/
- `closeOnClickOutside` setting `autohide` has `default: false`, although in the past `autoHide` has `default: true`

### 🏁 Checklist

### DIFF
--- a/src/components/NcPopover/NcPopover.vue
+++ b/src/components/NcPopover/NcPopover.vue
@@ -251,7 +251,8 @@ export default {
 		 */
 		closeOnClickOutside: {
 			type: Boolean,
-			default: false,
+			// eslint-disable-next-line vue/no-boolean-default
+			default: true,
 		},
 
 		/**


Antreesy (approved):

susnux:
> Regression from

Not really a regression it was for the breaking changes call and used `default: false` for boolean props.

ShGKme:
> > Regression from
> 
> Not really a regression it was for the breaking changes call and used `default: false` for boolean props.

It is a regression, since it changed the component behaviour, requiring now to provide a prop to have the previous behaviour. Having `NcPopover` not closed by click outside was not a part of any proposal. `default: false` was, but in this case the component should have had an inverted prop added.

PR #7525, opened 2025-09-17
feat(NcPopover): replace `closeOnClickOutside` with `noCloseOnClickOutside`

ShGKme:
### ☑️ Resolves

- On top of: https://github.com/nextcloud-libraries/nextcloud-vue/pull/7524
- Keeping boolean props `false` by default

### 🏁 Checklist

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -1788,7 +1788,7 @@ export default {
 					...this.manualOpen && {
 						triggers: [],
 					},
-					closeOnClickOutside: !this.manualOpen,
+					noCloseOnClickOutside: this.manualOpen,
 					popoverBaseClass: 'action-item__popper',
 					popupRole: this.config.popupRole,
 					setReturnFocus: this.config.withFocusTrap ? this.$refs.triggerButton?.$el : undefined,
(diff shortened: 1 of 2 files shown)


DorraJaouad (approved):

ShGKme:
> There's also the same prop in NcModal used, IIRC

Yes, but there it has the `false` default value as expected.

susnux (approved):
Ok, but it was in the v9 already with `default: false` due to the breaking changes.
So this now brings back a `default: true` and adds a new prop while it was already fixed in v9 breaking :sweat_smile:

ShGKme:
@susnux I don't understand, what PR are you referring to? What was the breaking change? Intentional?

susnux:
@ShGKme in v9 we removed the transparent wrapper around floating-vue thus only the exposed props are working (officially) so "closeOnClickOutside" with `default = false` was a breaking change in v9 which is now reverted again and a new prop `noCloseOnClickOutside` was added.

Meaning its again "re-breaking" by adding back the v8 behavior and new-deprecate the already removed behavior ;)

susnux:
rebased for ci

ShGKme:
> in v9 we removed the transparent wrapper around floating-vue thus only the exposed props are working (officially)

So I guess, you are referring to this PR, right?
It indeed removes the transparent wrapper, which was expected and described change.
- https://github.com/nextcloud-libraries/nextcloud-vue/pull/6802

> "closeOnClickOutside" with `default = false` was a breaking change in v9

- It is not mentioned in the PR description
- PR title/commit doesn't mark it as breaking (not `feat!`)
- There is no "breaking" label
- It is not mentioned in the changelog.

Are you saying, this was an intentional breaking change and not an unintended regression by the default value change?

susnux:
> Are you saying, this was an intentional breaking change and not an unintended regression by the default value change?

Not planned but I think from the changelog could be expected change. Still this approach here is cleaner! Thank you :)

PR #7572, opened 2025-09-27
[stable8] fix(NcAppContent): add reactivity to pane config key

GVodyanov:
### ☑️ Resolves

Previously if changing the pane config key the settings of the pane would not update automatically, now they do.


GVodyanov:
See this https://github.com/nextcloud-libraries/nextcloud-vue/pull/7573

PR #7587, opened 2025-10-01
feat(NcModal): expose `initialFocus` prop for focus-trap options

Antreesy:
### ☑️ Resolves

- Exposing `initialFocus` and `setReturnFocus` to NcModal and NcDialog consumers
	- See https://github.com/focus-trap/focus-trap#initialfocus
	- Allows to manipulate initial focus (e.g. in case first element in the trap is NcSelect and we don't want it to be open)

### 🏁 Checklist

### DIFF
--- a/src/components/NcDialog/NcDialog.vue
+++ b/src/components/NcDialog/NcDialog.vue
@@ -211,6 +211,7 @@ export default {
 </docs>
 
 <script setup lang="ts">
+import type { FocusTargetOrFalse } from 'focus-trap'
 import type { Slot } from 'vue'
 import type { ComponentProps, VueClassType } from '../../utils/VueTypes.ts'
 
@@ -236,6 +237,12 @@ const props = withDefaults(defineProps<{
 
 	/** Additional elements to add to the focus trap */
 	additionalTrapElements?: Array<string | HTMLElement>
+
+	/** Set element to return focus to after focus trap deactivation */
+	setReturnFocus?: FocusTargetOrFalse
+
+	/** Specify an element to receive initial focus after focus trap activation */
+	initialFocus?: FocusTargetOrFalse
 
 	/**
 	 * The element where to mount the dialog, if `null` is passed the dialog is mounted in place.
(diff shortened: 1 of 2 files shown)


Antreesy:
I forgot about that PR🪦 Shall we get it in?

DorraJaouad:
Was it for the translation dialog in chat?

Antreesy:
Yes, in general it fits for any dialog that does not need to have an initial focus on the child element

ShGKme:
> * (e.g. in case first element in the trap is NcSelect and we don't want it to be open)

With a small CSS fix, and by using available props and events we can make `NcSelect` open only manually but not by focus.

Would it also work for you?

Antreesy:
> we can make `NcSelect` open only manually but not by focus. Would it also work for you?

Depends on how it changes current default behaviour, and what app developers should do to implement it. I'd take a look =)

susnux:
Whats the current state of this PR?

Antreesy:
> Whats the current state of this PR?

Ready to be merged for two months already 👀

susnux:
What I am not sure about is whether we should focus the first element.
Depending on how to interpret the accessibility guidelines it even makes sense to focus the dialog itself.

src/components/NcDialog/NcDialog.vue
```diff
@@ -237,6 +238,12 @@ const props = withDefaults(defineProps<{
 	/** Additional elements to add to the focus trap */
 	additionalTrapElements?: Array<string | HTMLElement>
 
+	/** Set element to return focus to after focus trap deactivation */
```
susnux, line 241: Maybe better

```suggestion
	/** Set element to return focus to after dialog was closed */
```

src/components/NcDialog/NcDialog.vue
```diff
@@ -237,6 +238,12 @@ const props = withDefaults(defineProps<{
...
+	setReturnFocus?: FocusTargetOrFalse
+
+	/** Specify an element to receive initial focus after focus trap activation */
```
susnux, line 244:
```suggestion
	/** Specify an element to receive initial focus after dialog was opened */
```

susnux (commented):
IMHO this feature makes sense in some cases, for those I approve.
But if its really only about NcSelect then we can also fix that component

PR #7589, opened 2025-10-01
fix(NcCounterBubble): also shorten in German

ShGKme:
### ☑️ Resolves

- Browsers decided that in German numbers should not be shortened like `2048` should be `2048` instead of `2K` in English
- This PR uses English style instead of German to shorten the number in German
- ⚠️ Same problem in Italian. Shall we do the same?

Before | After
-- | --
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/NcCounterBubble/NcCounterBubble.vue
+++ b/src/components/NcCounterBubble/NcCounterBubble.vue
@@ -162,12 +162,20 @@ const props = withDefaults(defineProps<{
 	type: '',
 })
 
+// Special case: browsers do not shorten numbers in German
+// Fallback to English to have 2K instead of 2048
+// TODO: what about Italian with the same issue?
+let locale = getCanonicalLocale()
+if (locale === 'de' || locale === 'de-DE') {
+	locale = 'en'
+}
+
 const humanizedCount = computed(() => {
 	if (props.raw) {
 		return props.count.toString()
 	}
 
-	const formatter = new Intl.NumberFormat(getCanonicalLocale(), {
+	const formatter = new Intl.NumberFormat(locale, {
 		notation: 'compact',
 		compactDisplay: 'short',
 	})
--- a/tests/unit/components/NcCounterBubble/NcCounterBubble.spec.ts
+++ b/tests/unit/components/NcCounterBubble/NcCounterBubble.spec.ts
@@ -3,9 +3,22 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
+import type { MockedFunction } from 'vitest'
+
+import { getCanonicalLocale } from '@nextcloud/l10n'
 import { mount } from '@vue/test-utils'
-import { describe, expect, it } from 'vitest'
+import { afterEach, describe, expect, it, vi } from 'vitest'
 import NcCounterBubble from '../../../../src/components/NcCounterBubble/NcCounterBubble.vue'
+
+vi.mock('@nextcloud/l10n', async () => {
+	const actual = await vi.importActual('@nextcloud/l10n')
+	return {
+		...actual,
+		getCanonicalLocale: vi.fn(() => 'en'),
+	}
+})
+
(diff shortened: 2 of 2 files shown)


ShGKme:
@marcoambrosini For Italian we keep it in full form, correct?

szaimen:
Honestly, I am not sure about this change. Using 2K is very uncommon in german afaik...

kra-mo:
Isn't there a library for this with already localized short forms? I understand that it would be an extra dependency but I don't feel too great about doing this for every language manually, especially because the thousands are not the only thing that is usually shortened.

kra-mo:
I'm sure you've seen [this one](https://youtu.be/0j74jcxSunY) before :D

marcoambrosini:
Correct for Italian @ShGKme

ShGKme:
@kra-mo we don't do it manually for every language. We are using the native Intl formatter.

The problem is that the default formatter doesn't shorten thousands in German and Italian languages, unlike all the other languages. And as far as I know, it is on purpose.

So the proposal is to use English style shortening for German (and, maybe, Italian).

kra-mo:
> @kra-mo we don't do it manually for every language. We are using the native Intl formatter.
> 
> The problem is that the default formatter doesn't shorten thousands in German and Italian languages, unlike all the other languages. And as far as I know, it is on purpose.
> 
> So the proposal is to use English style shortening for German (and, maybe, Italian).

Ah, I see. Maybe ask upstream? Otherwise, my hands are off then :)

ShGKme:
> Maybe ask upstream?

Upstream is Google, Microsoft, Apple and Mozilla with their browsers, so, I'm unsure if it is an option :)

The question is: do we want to shorten numbers in German (and Italian) or not.

jancborchardt (changes requested):
To be correct in German (and in English as well?) it would have to be a lowercase "k", so "2k".

ShGKme:
> (and in English as well?) it would have to be a lowercase "k", so "2k"

@jancborchardt It uses native browsers formatter, not a custom one. So in English it is K uppercase as defined by the browser. Like in other places (likes/views on social media, YouTube and etc.).

As we are making a custom option for German anyway, I can make it lowercase, but could you provide complete rules in this case? For example, is it always lower case, or only for `k`.

ShGKme:
Note: counter in EN and DE on YouTube. Indeed, it is not shortened with K, as @szaimen noticed.

image

image

jancborchardt:
So in German it is not super common, but if we need a shortened one, "2k" is definitely possible.
A more common one would be "2 Tsd." but that doesn’t really shorten it. For million the common one is "2 Mio.", which would shorten it better.

The German page for info on the "thousand" shortening: https://de.wikipedia.org/wiki/Tausend#Gebr%C3%A4uchliche_Abk%C3%BCrzungen_und_Synonyme

jancborchardt:
What we could also do is then simply not shorten the thousands if we think it’s fine. What happens above certain limits like with 10k, 100k etc? At what point does it shorten at all?

ShGKme:
> What we could also do is then simply not shorten the thousands if we think it’s fine. What happens above certain limits like with 10k, 100k etc? At what point does it shorten at all?

Starting `1000000` formatted to `1 Mio.`

ShGKme:
> What we could also do is then simply not shorten the thousands if we think it’s fine.

That is the current state

susnux:
I think this is then probably fine in German.
While for me and likely many others "2 k" would be ok and understandable, the official shortenings should be "Tsd." which is basically not shorter and only useful when writing numbers as text.

susnux:
Shall we close this @ShGKme ?

jancborchardt:
Yeah, sounds good and more consistent to just use the default. It's a similar thing in Japanese, where the delimiter is not "thousand" but actually "ten thousand" as "万".

And yes, 1 million is "100万" / "百万", which is "100 x 10k". :)

PR #7590, opened 2025-10-01
fix(NcAppContent): avoid content re-mounting

ShGKme:
### ☑️ Resolves

- Checking slots in `mounted` + `updated` is not correct:
  - The first rendering (during mounting) has no `hasSlots` 
  - There is a need to first re-render the component, and then again after changing the slot existence
  - In general, slots should only be checked during the rendering
- It resulted in mounting the content twice

### 🏁 Checklist


max-nextcloud (approved):
Did not test yet. But code looks good.

PR #7620, opened 2025-10-06
chore: prepare release of v8.32.0

susnux:
### 🚀 Enhancements
* feat(NcColorPicker): allow to clear selected value [\#7531](https://github.com/nextcloud-libraries/nextcloud-vue/pull/7531)
* feat(NcModal): add new `closeButtonOutside` prop and deprecated `closeButtonContained` [\#7553](https://github.com/nextcloud-libraries/nextcloud-vue/pull/7553) \([Antreesy](https://github.com/Antreesy)\)

### 🐛 Fixed bugs
* fix(useHotKey): use correct callback type [\#7511](https://github.com/nextcloud-libraries/nextcloud-vue/pull/7511)
* fix(NcRadioGroupButton): adjust hight if needed [\#7554](https://github.com/nextcloud-libraries/nextcloud-vue/pull/7554)
* fix(NcAvatar): log error and URL on error [\#7561](https://github.com/nextcloud-libraries/nextcloud-vue/pull/7561) \([ChristophWurst](https://github.com/ChristophWurst)\)
* fix(console): fix log context object [\#7563](https://github.com/nextcloud-libraries/nextcloud-vue/pull/7563) \([ChristophWurst](https://github.com/ChristophWurst)\)
* fix(NcAppContent): avoid double content mount [\#7619](https://github.com/nextcloud-libraries/nextcloud-vue/pull/7619)
* fix(NcAppContent): add reactivity to pane config key  [\#7573](https://github.com/nextcloud-libraries/nextcloud-vue/pull/7573) \([GVodyanov](https://github.com/GVodyanov)\)

### Other Changes
* refactor(NcColorPicker): migrate component to Typescript and script-setup [\#7532](https://github.com/nextcloud-libraries/nextcloud-vue/pull/7532)
* refactor(NcAppSettingsDialog): migrate to Typescript and script-setup [\#7517](https://github.com/nextcloud-libraries/nextcloud-vue/pull/7517)

PR #7621, opened 2025-10-06
chore: prepare release of v9.0.1

susnux:
Align with #7620 to get the fixes also for v9

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@
 
 All notable changes to this project will be documented in this file.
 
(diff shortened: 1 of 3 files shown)


ShGKme (approved):

susnux:
@Antreesy its split off, basically v8 and v9 are separate branches so all changes in v8 are either only v8 related or already included in v9 versions. So it makes no sense to have new v8 releases in v9 changelog anymore.
(we also never do it in other libraries after the stable next was released)

PR #7623, opened 2025-10-06
chore: switch dependency for MD5 to `ts-md5`

### LINKED ISSUE #7603: [NcSelectUsers] security risks warning and library size
Nextcloud v32.0.0
"vite": "^7.1.9",
"@nextcloud/vue": "^9.0.0"
"vue": "^3.5.22",

Vite build shows this warning after adding ` import NcSelectUsers from '@nextcloud/vue/components/NcSelectUsers'` to a component:

`node_modules/vm-browserify/index.js (110:11): Use of eval in "node_modules/vm-browserify/index.js" is strongly discouraged as it poses security risks and may cause issues with minification.`

The bundle size grows from 
`./js/myapp-main.mjs  2,338.53 kB │ gzip: 472.66 kB │ map: 4,880.63 kB`

to 
` ./js/myapp-main.mjs  3,791.68 kB │ gzip: 745.32 kB │ map: 7,488.03 kB`

susnux:
### ☑️ Resolves

- Fix #7603 

This decreases bundle size when using `usernameToColor` by ~700kiB.

### 🖼️ Screenshots

Bundle when just using `NcSelectUsers` and nothing else.
🏚️ Before | 🏡 After
---|---
Bildschirmfoto_20251006_223709|Bildschirmfoto_20251007_000657

<details

With md5.js
This does not change any dependencies as `md5.js` is already used internally but it decreases the public API that we expose so only the md5 part will be included (crypto-browserify is really bad treeshakeable)

🏚️ Before | 🏡 md5.js | ts-md5
---|---|---
Bildschirmfoto_20251006_223709|Bildschirmfoto_20251006_223900|Bildschirmfoto_20251007_000657

</details>

### 🏁 Checklist


susnux:
> What about other libraries, such as blueimp-md5, js-md5, or just coping md5 function to the source, as it is quite small and doesn't need maintenance?

Mentioned ones are even worse, while they are pretty small at first sight they add dependencies instead of reusing.
In Nextcloud ecosystem you already nearly always end up with having parts of browserify bundled due to some internal Node module usage and then the polyfill.
So this fixes useless bundling of full `crypto-browserify` but still reuses a lot of packages.
Moreover blueimp cannot be used that package is dead and thus is not applicable for security policies.

As you can see in the after screenshot of the bundle analyzer there is not much left of the md5 dependency.

(Best solution is to change the algorithm in server to use sha256 then we can just use the browsers crypto module...)

susnux:
Pushed a fixup with using `ts-md5` which has good tree shaking and **0 dependencies** so its really small.

If we use `md5.js` we share its dependencies.
If we use `ts-md5` we have no new sub-dependencies, and a small md5 implementation once Node modules are not used anymore by other dependencies.

I am ok with both... But `ts-md5` is even smaller (another 150kiB saved) and has proper types.

package.json
```diff
@@ -141,6 +141,7 @@
     "@vue/tsconfig": "^0.8.1",
     "babel-loader-exclude-node-modules-except": "^1.2.1",
     "core-js": "^3.45.1",
+    "dependency-cruiser": "^17.0.2",
```
Antreesy: Intended? Do we need scripts for that to use maybe?

src/functions/usernameToColor/index.ts
```diff
@@ -18,7 +18,7 @@ function hashCode(str: string): number {
 
 	// Hash a given string, if it is not md5 hash already
 	if (str.match(/^([0-9a-f]{4}-?){8}$/) === null) {
-		hash = createHash('md5').update(str).digest('hex')
+		hash = new Md5().appendAsciiStr(str).end() as string
```
Antreesy:
```suggestion
		hash = Md5.hashAsciiStr(str)
```
By source code, should also work like this

ShGKme:
Tests are failing, colors have changed

ShGKme:
solution | usernameToColor impact | without translations
-- | -- | --
crypto-browserify | 733.77 kb | 693.99 kb
md5.js | 148.45 kb | 108.67 kb
ts-md5 | 47.30 kb | 7.52 kb
blueimp-md5 | 44.81 kb | 5.03 kb
md5 source | 41.94 kb | 2.16 kb

`ts-md5` is fine by me

package.json
susnux: Only for testing

src/functions/usernameToColor/index.ts
susnux: Turns out we need `hashString` not ascii string. Fixed.

ShGKme:
Great number of weekly downloads
image

ShGKme (approved):

Antreesy (approved):

susnux:
`git rebase --autosquash origin/main`

PR #7630, opened 2025-10-07
fix(NcInputField): use empty string to fix label position

susnux:
### ☑️ Resolves

If there is no placeholder we need still at least an empty string to make the label animation with `:placeholder-shown` work.

### 🏁 Checklist

### DIFF
--- a/src/components/NcInputField/NcInputField.vue
+++ b/src/components/NcInputField/NcInputField.vue
@@ -163,7 +163,17 @@ const inputElement = useTemplateRef('input')
 
 const hasTrailingIcon = computed(() => props.showTrailingButton || props.success)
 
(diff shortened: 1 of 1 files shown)


susnux:
Did not find a rule that forbids it, its also what others use to get this behavior so seems to be fine.
Also it seems that empty is seen just as non existing for screen readers so also there should work.

jancborchardt (approved):
Thanks! @arnowelzel this will also fix the label issue/regression you mentioned on the forum. :)

ShGKme:
> Is it valid for a11y and screen readers to have an empty placeholder instead of no placeholder?

Tested with NVDA

ShGKme (approved):

PR #7631, opened 2025-10-07
build: rolldown

ShGKme:
- 1.5 seconds (8%) faster on my machine (Linux)
- I don't know if it is worth it 🙈

🏚️ Before | 🏡 After
---|---
18 seconds | 16.5 seconds

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -88,7 +88,7 @@
         "ts-node": "^10.9.2",
         "typescript": "^5.9.3",
         "url-loader": "^4.1.1",
-        "vite": "^7.1.9",
+        "vite": "npm:rolldown-vite@^7.1.16",
         "vitest": "^3.2.4",
         "vue-eslint-parser": "^10.2.0",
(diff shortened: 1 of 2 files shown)


susnux:
For libraries it does not change much, with apps you have more control with `advancedChunks` option.

ShGKme:
> For libraries it does not change much

I expect it to be faster as this is one of the reasons to use rolldown (1/2).

Currently, this lib builds slower than the entire Talk with ~10 entrypoint and 140MB dist 👀
Even without translations.

susnux:
> I expect it to be faster as this is one of the reasons to use rolldown (1/2).

But the long running things here are neither the transpiler nor the bundler, so nothing rolldown can help with.
The two slow parts are the l10n plugin and the DTS plugin (try `DTSPluginOptions = false` it reduces build time for me from 10s to 2s).

susnux:
I think we then would need to switch to https://github.com/sxzz/rolldown-plugin-dts instead when rolldown is used

PR #7633, opened 2025-10-08
feat: add `NcAppSettingsSectionShortcuts`, `NcHotkeyList` and `NcHotkey`

ShGKme:
### ☑️ Resolves

- Ref: https://github.com/nextcloud-libraries/nextcloud-vue/issues/7124#issuecomment-3225287374
- `Hotkey` name can be changed to `Shortcuts`. But we have `useHotkey` already.
- `<NcHotkey>` - describes a single hotkey with a label and the hotkey
  - hotkey automatically resolves to `<NcKbd>` components
- `<NcHotkeyList>` - list of hotkeys with an optional heading
- `<NcAppSettingsSectionShortcuts>` - NcAppSettingsDialog for the hotkey lists with a gap between them and built-in fixed name "Keyboard Shortcuts".

### 🖼️ Screenshots

image

image

### 🏁 Checklist


ShGKme:
Force pushed:
- l10n fix (had `Keyboard shortcuts` and `Keyboard Shortcuts`)
- stylelint fix (an empty style)

ShGKme:
@kra-mo This is the last design you have approved, no layout changes.

ShGKme:
- Removed one unnecessary CSS line

src/components/NcAppSettingsSectionShortcuts/index.ts
```diff
@@ -0,0 +1,6 @@
+/*
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
```
susnux, line 2: so its now the 3rd way of adding SPDX in this repo? 😆

src/components/NcAppSettingsSectionShortcuts/NcAppSettingsSectionShortcuts.vue
```diff
@@ -0,0 +1,90 @@
...
+import { t } from '../../l10n.ts'
+
+defineSlots<{
+	/**
+	 * Shortcuts described via <NcHotkeyList> components
+	 */
+	default?: Slot
+}>()
+
+const name = t('Keyboard shortcuts')
```
susnux: not a blocker, but IMHO this belongs into the template

src/components/NcHotkey/NcHotkey.vue
```diff
@@ -0,0 +1,92 @@
...
+
+<template>
+	<li :class="$style.hotkey">
+		<span role="term" :class="$style.hotkey__keys">
+			<slot name="hotkey">
+				<NcKbd v-for="(symbol, index) in symbols" :key="index" :symbol />
+			</slot>
+		</span>
+		<span role="definition" :class="$style.hotkey__label">
+			<slot>
```
susnux, line 52: Do we need the slot? Meaning do we have a use case? Otherwise I would not add it until there is one and use the prop instead

src/components/NcHotkey/NcHotkey.vue
```diff
@@ -0,0 +1,92 @@
...
+const {
+	label = undefined,
+	hotkey = undefined,
+} = defineProps<{
+	/**
+	 * Label for the hotkey
+	 */
+	label?: string
+	/**
+	 * Hotkey combination string separated by space, for example, `Control Alt T` or `F2` (see `<NcKbd>` for supported key names)
```
susnux, line 21: Would it not be better to pass it as array?

src/components/NcAppSettingsSectionShortcuts/index.ts
ShGKme: It is the only way in JetBrains IDEs (the built-in copyright feature).

src/components/NcHotkey/NcHotkey.vue
ShGKme: IMO, it is in general a good practice in UI components to provide a slot for basic customisation. It doesn't add complexity or problems except in some specific cases.

Also, if the text is long (it is semantically a content for HTML), it sometimes looks bad in the HTML attribute while it looks native between tags.

src/components/NcAppSettingsSectionShortcuts/index.ts
susnux: JetBrains :broken_heart: webpack?

src/components/NcHotkey/NcHotkey.vue
ShGKme: For implementation simplicity — it probably will.

But for the usage, I like that you can use the component in a short string without any bindings.

And if you provide it wrong, you will see that something is wrong.

However, we can add some checks to the `NcKbd` component to make sure,  a user doesn't use `Ctrl` instead of `Control`. Or we automatically replace `Ctrl` with `Control` and check the case.

src/components/NcHotkey/NcHotkey.vue
susnux: from my perspective if you can add a simple data like a string by prop thats easier because you can work with it without needing the render function to be called.

But this is also not a blocker for me

src/components/NcAppSettingsSectionShortcuts/NcAppSettingsSectionShortcuts.vue
ShGKme: Pushed fixup

src/components/NcHotkey/NcHotkey.vue
susnux: probably right for manual usage, I was more thinking from automated generation of this, e.g. for files plugins

ShGKme: I don't think there is any difference between the automated generation of a string and of an array of strings. If you mean the `NcHotkeyList` generation.

src/components/NcHotkey/NcHotkey.vue
ShGKme: > without needing the render function to be called.

But it is better to call many small render functions than one big render function. That is also why slots in vue 3 are faster than when the default slot content was rendered in the parent in Vue 2

> a simple data like a string

But it can be a long string, describing the action.

src/components/NcHotkey/NcHotkey.vue
susnux: It just feels weird to convert it multiple times, from array to string and back to array. As the data type even is an array, but I get that for manual usage passing keys as array seems to be weird.

ShGKme: > from array to string and back to array

Do you have a Files API for hotkeys already? Does it use array? Not, for example, modifier + key?

ShGKme:
> 1. The group titles should be aligned with the section titles and labels, and

@kra-mo This PR only add the keyboard shortcuts components, it doesn't (yet) changes the entire settings design.

image

> 2. Padding on the right is not a lot, I would increase it by quite a bit, or rather, decrease the max-width of the whole thing

By "the whole thing" do you mean the shortcuts list or the settings dialog?
And just a bit, or significantly, to make labels close to the keys?

> since now the labels are quite far away from the keys

That what I didn't like initially with the new design in comparison with the current one, where keys where right after the labels.

image

src/components/NcHotkey/NcHotkey.vue
susnux: Not really but for me putting things into a string to de-serialize it again looks really ugly - its basically machine readable and then again humanized in between just for sake of nice template layout.
But as said I am fine with current solution, as for 90% of all use cases its better :)
Just that one use case would be nice to have an array like for classes where you can pass array or string.

Maybe that is too much, so I am fine with this solution :white_check_mark:

kra-mo:
> This PR only add the keyboard shortcuts components, it doesn't (yet) changes the entire settings design.

Where do the titles such as Actions, Selection or Navigation in the screenshot come from then? Are they not part of the component? Those were what I was thinking about because right now, they look inset.

> By "the whole thing" do you mean the shortcuts list or the settings dialog?

I meant just the list :)

> And just a bit, or significantly, to make labels close to the keys?

Significantly, as I don't think that labels will be too long here. It's nice to get related items closer together, as you said as well.

Eyeballing it roughly, I'd say 400px max width, which would look like this:

image

kra-mo:
> Those were what I was thinking about because right now, they look inset.

This seems to be it:

image

image

Removing it, it looks expected.

> `/* Align with form group */`

I see though. But the items themselves are not aligned like that? All headings, labels, text, etc. should be aligned the same. So it probably does make sense to keep it, but then:

1. it should also apply to the list of items, not just the group headings and
2. it should (later) apply to settings section titles, and everything else as well (which I'm guessing is already planned)

kra-mo:
Also, looking at it, it also would be nice to have more of a gap between entire groups, since the padding below and above group headings now seems identical. The heading should be closer to the group that it represents, so there should be more spacing above it.

src/components/NcHotkeyList/NcHotkeyList.vue
```diff
@@ -0,0 +1,102 @@
...
+		<ul :aria-labelledby="labelId">
+			<slot />
+		</ul>
+	</div>
+</template>
+
+<style module>
+.hotkeyList__heading {
+	font-weight: bold;
+	margin-inline: var(--border-radius-element); /* Align with form group */
```
susnux, line 50: I think this is the problem as level-0 headings should not have indention according to the mockup.

kra-mo: [Here](https://github.com/nextcloud-libraries/nextcloud-vue/issues/7124#issuecomment-3225287374) we did discuss and come to the conclusion that everything should be aligned.

For that, I think indentation *is* needed, we should just make sure that it is the same everywhere so the text is aligned nicely.

ShGKme:
> Eyeballing it roughly, I'd say 400px max width,

👍

> 1\. it should also apply to the list of items, not just the group headings

My bad, missed it.

> 2\. it should (later) apply to settings section titles, and everything else as well (which I'm guessing is already planned)

Yes

ShGKme:
- Increased twice the gap between groups
- Added the same internal padding to the list as the heading (as was planned before)
- Added `max-width: 400px`

src/components/NcHotkey/NcHotkey.vue
ShGKme: To not map things twice we need Files API to be the same as the component and then the `useHotkey` composable. So it must be separated to boolean modifiers and then array of strings like

```html
<NcHotkey :label control alt :keys="['s']" />
```

Is it what you wanted?

ShGKme:
hid comments resolved by the approved change above

src/components/NcHotkey/NcHotkey.vue
susnux: Not sure I like it. I would say lets keep it simple for now and use the string you implemented

src/components/NcHotkey/NcHotkey.vue
```diff
@@ -0,0 +1,93 @@
...
+	 */
+	default?: Slot
+	/**
+	 * Custom hotkey content for an advanced hotkey layout with <NcKbd> components or other custom content
+	 */
+	hotkey?: Slot
+}>()
+
+const symbols = computed(() => {
+	return hotkey?.split(/\s+/).map((s) => s.trim())
```
susnux, line 38: Should this not be sorted to ensure a consistent order?
Like always CTRL first, then Shift, then Alt, then number or letter?

src/components/NcHotkey/NcHotkey.vue
susnux: Thats valid, its just that often you have more control over the component usage by limit it.
Like limiting it to a prop enforces a string, while a slot also could be other components / formatting.

src/components/NcHotkey/NcHotkey.vue
ShGKme: Then I'd definitely go with a more programmatic API:D
```
<NcHotkey :label control alt :keys="['s']" />
```

ShGKme: ping @susnux as it looks like you are working on programmatic hotkeys generation in Files :)

I'd avoid creating a new component to deprecate its interface soon :D

ShGKme:
Example of the custom content from the new drafts:

image

src/components/NcHotkey/NcHotkey.vue
susnux: Its not really important, as I can also just create the string from any other input.
But I planned to define it just like `useHotKey` works.

ChristophWurst:
Backport to stable8?

ShGKme:
> Backport to stable8?

Checking first if previous backports on the related files where merged.

PR #7634, opened 2025-10-08
fix(NcEmptyContent): use name as label

susnux:
### ☑️ Resolves

> Naming a note is optional, but can help screen reader users understand its context and purpose

So its not required but it was a breaking change when we made the `name` not a heading anymore (while still it was an a11y fix). But now the note is not really identifiable when using a11y-api, so we need to use the name as the label.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🏁 Checklist


src/components/NcEmptyContent/NcEmptyContent.vue
```diff
@@ -165,18 +167,20 @@ defineSlots<{
...
 <template>
-	<div class="empty-content" role="note">
+	<div :aria-labelledby="nameId" class="empty-content" role="note">
 		<div v-if="$slots.icon" class="empty-content__icon" aria-hidden="true">
 			<slot name="icon" />
 		</div>
-		<slot name="name">
-			<span v-if="name !== ''" class="empty-content__name">
+		<div v-if="name !== '' || $slots.name" :id="nameId" class="empty-content__name">
+			<slot name="name">
```
ShGKme, line 180: This also means that this is not possible anymore to replace content with the default `.empty-content__name` styles

susnux: Does it make sense to do so?
The only thing that you cannot reset in the slot styles is the margin but I would expect that to be part of component styles anyways.

ShGKme: I did not check how the slot is used in the apps. I saw that PR describes an a11y fix, but the code also changes the design for the slot content. Thus I don't know if it was an intended change or just a side effect of fixing the a11y issue.

If this is an intended change and it doesn't break the current usage — fine by me.

susnux: Well I can also just add an unstyled wrapper around it in this PR with the id.
Thats also possible.

But I thought its more useful to wrap the name in general to have save default styling and then being able to adjust visual styling like color.

I also checked all apps in the nextcloud org and they only use the slot for passing the name as slot instead of prop, so should be fine.

PR #7636, opened 2025-10-08
fix(NcAppSettingsDialog): check slot children recursively

Antreesy:
### ☑️ Resolves

- Fix sections order in navigation
- `slots.default()` iterates only over the first layer of children, whereas some might be nested in fragments:

```html
<NcAppSettingsDialog>
	<NcAppSettingsSection />
	<template>
		<NcAppSettingsSection />
		<NcAppSettingsSection />
	</template>
</NcAppSettingsDialog>
```

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
2025-10-08_16h26_43 | 2025-10-08_16h27_06

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist


ShGKme (commented):
I'd avoid working with render functions here and add `order` prop like we have in the sidebar tabs. Especially because app settings are always managed by a single app, so we can always provide the correct order from the app side

Antreesy:
> and add `order` prop

We would still need a middle-ground solution, when not provided... Maybe don't sort at all? Or there was an issue with it before?

ShGKme:
> Maybe don't sort at all? Or there was an issue with it before?

Fine by me. Should work well with static sections.

ShGKme:
An idea:
1. Create a counter in the NcAppSettingsDialog
2. If a section is registered with an explicit order — use it and set as the last `counter` value
3. Otherwise increment the counter and use it as the order fallback

In a simple case - order is the mounting order, which should work fine with a static section list.

src/components/NcAppSettingsDialog/useAppSettingsDialog.ts
```diff
@@ -13,9 +13,10 @@ interface AppSettingsRegistrationContext {
 	 *
 	 * @param id - The section ID
 	 * @param name - The section name
+	 * @param order - Optional section order in the list
 	 * @param icon - Optional icon component
 	 */
-	registerSection(id: string, name: string, icon?: VNode[]): void
+	registerSection(id: string, name: string, order?: number, icon?: VNode[]): void
```
susnux, line 19: maybe we should at one point use a context object instead ^^

ShGKme: I'm doing it in a follow-up to https://github.com/nextcloud-libraries/nextcloud-vue/pull/7651

src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
```diff
@@ -145,12 +155,23 @@ function registerSection(id: string, name: string, icon?: VNode[]) {
...
-	const newSections = [...registeredSections.value, { id, name, icon }]
+	// Ensure tab order is set
+	if (order !== undefined) {
+		sectionsOrderMap.set(id, order)
+	} else if (sectionsOrderMap.has(id)) {
+		order = sectionsOrderMap.get(id)
+	} else {
+		order = sectionsOrderMap.size
+			? Math.max(...sectionsOrderMap.values()) + 1
+			: 1
```
ShGKme: nitpik:
```suggestion
		order = Math.max(...sectionsOrderMap.values()) + 1
```

src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
```diff
@@ -293,18 +314,24 @@ Just nest the `AppSettingSections` component into `NcAppSettingsDialog`,
 providing the section's name prop. You can put your settings within each
 `NcAppSettingsSection` component.
 
+In case of complex settings panel with conditional rendering it is better to set ``order`` prop for each section
+to ensure the correct order in the navigation list, when registering.
```
ShGKme: Since the render function approach is not fully compatible with Vue here, I'd ~~require~~ **highly recommend** setting the order if sections are dynamic.

Something like:

```suggestion
Sections order is defined during initial rendering. In case of dynamic/conditional sections rendering explicit `order` prop must be used for ordering.
```

Let's also make it an additional example, like with icons?

src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
Antreesy: it results to `-Infinity`, if there are no values in map

UPD: `Math.max(0, -Infinity)` is 0 =P

ShGKme: 👀
```suggestion
		order = Math.max(...sectionsOrderMap.values(), 0) + 1
```

ShGKme:
- Backport after https://github.com/nextcloud-libraries/nextcloud-vue/pull/7654

PR #7655, opened 2025-10-14
refactor: use named export for logger

ShGKme:
### ☑️ Resolves

- It is a bit annoying during backporting that `import logger` is different on `main` and `stable8`
  - Default on `main`
  - Named on `stable8`
- Using named on both branches to reduce the difference and prefer named exports

### 🏁 Checklist


susnux:
(so named exports should be used - during the beta phase I was sure to migrated all to named exports but maybe missed it here? Not sure...)

PR #7661, opened 2025-10-15
fix(NcRadioGroupButton): scope expensive CSS selectors

Antreesy:
### ☑️ Resolves

- Noticed then checking Performance tab in DevTools:
  - Styles are `module` for this component, therefore css selectors work differently
  - PR changes sounds valid for this exact case

### 🖼️ Screenshots

image

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/NcRadioGroupButton/NcRadioGroupButton.vue
+++ b/src/components/NcRadioGroupButton/NcRadioGroupButton.vue
@@ -102,11 +102,11 @@ function onUpdate() {
 		cursor: pointer;
 	}
 
-	:has(&__label) {
+	&:has(&__label) {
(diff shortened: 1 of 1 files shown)


Antreesy:
Although component isn't used in the app, styles are bundled with the nextcloud-vue css chunk.
Same could be said for `:has(splitpanes--dragging)` from NcAppContent (but there is a dist imported directly in the component). Maybe it's a part of a bigger problem, that non-used components shouldn't be bundled?

ShGKme:
> Maybe it's a part of a bigger problem, that non-used components shouldn't be bundled?

Not imported components are not supposed to be bundled...

PR #7663, opened 2025-10-15
[stable8] fix(NcAppNavigationSettings): alignment of button icon

mejo-:
### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppNavigationSettings/NcAppNavigationSettings.vue
+++ b/src/components/NcAppNavigationSettings/NcAppNavigationSettings.vue
@@ -134,6 +134,10 @@ export default {
 				white-space: nowrap;
 				text-overflow: ellipsis;
 			}
+
+			&__icon {
+				display: flex;
+			}
 		}
 	}
 


Antreesy:
Isn't it coming from these styles? See all of them are affecting c.nc

image
image

mejo-:
> Isn't it coming from these styles? See all of them are affecting c.nc

Indeed, that's the cause of it. But looking at it's source, it seems to come from `core/css/inputs.scss` in server and is there since 2017 (with `not(.button-vue)` since 2022).

Antreesy:
> not(.button-vue)

Then either `<button>` in the component should have this class; or should be migrated to NcButton. It's already done on `main` branch, simply backporting changes from there would be a better solution rather than patching some server CSS

mejo-:
> Then either <button> in the component should have this class; or should be migrated to NcButton. It's already done on main branch, simply backporting changes from there would be a better solution rather than patching some server CSS

Good point, will do that.

mejo-:
Superseeded by https://github.com/nextcloud-libraries/nextcloud-vue/pull/7700

PR #7352, opened 2025-08-21
feat(NcFieldset): new component to structure settings

susnux:
### ☑️ Resolves

- Resolves #7122 
- For #7124 

### 🖼️ Screenshots

Screenshot 2025-08-21 at 23-22-52 Nextcloud Vue Style Guide

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

- [ ] ⛑️ Tests are included or are not applicable
- [ ] 📘 Component documentation has been extended, updated or is not applicable
- [ ] 2️⃣ Backport to `stable8` for maintained Vue 2 version or not applicable

PR #7689, opened 2025-10-20
feat: add NcFormGroup

ShGKme:
### ☑️ Resolves

- Fix https://github.com/nextcloud-libraries/nextcloud-vue/issues/7122
- Closes #7352 
- Differences from the initial plan:
  - No special "nesting" support — it was decided to use specific section components for large form sections like `<NcAppSettingsSection>`
  - No content style changing — it will be handled by a special `NcFormBox` component

The current solution has some `TODOs`.

### 🖼️ Screenshots

image

### 🏁 Checklist


ShGKme:
@kra-mo Currently the heading (legend, description) has 12px inline padding as per design. To be inline with elements having 8px border radius. So we can call it "Radius + 1 grid baseline".

Is that correct and what is intended by the design?

In thi case, we'll also adjust the field label paddings.

susnux:
I think it would be better to make the prop called either name or label and not introduce a third variant for the label of the element here ("legend")

ShGKme:
`name` has a different meaning. We have it in many places after the change (going from `title`) in v8.0.0, and it was quite consistent. But later new components (such as `NcDialogButton`, `NcRadioGroup`) were introduced with a new prop already (`label`).

Unlike `name`, `label` may work here by the meaning (it labels an element with the group role).

But I wouldn't call the `legend` a "third variant". It represents the [`<legend>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/legend) element of the `<fieldset>`.

Open for others' opinions.

kra-mo:
Nice, I'm really happy to see this :)

---

> @kra-mo Currently the heading (legend, description) has 12px inline padding as per design. To be inline with elements having 8px border radius. So we can call it "Radius + 1 grid baseline".
> 
> Is that correct and what is intended by the design?
> 
> In thi case, we'll also adjust the field label paddings.

Do you mean to adjust the input fields' leading padding to match? Then yes, it should just be aligned.

| Mockup | Implementation |
|--------|--------|
| image | image |

---

> No special "nesting" support — it was decided to use specific section components for large form sections like <NcAppSettingsSection>

Yes, I thought about this too, it makes total sense :)

---

> No content style changing — it will be handled by a special NcFormGroup component

Just so I understand: where would the `FormGroup` sit in the tree relative to individual fields and the `Fieldset` itself?

---

PS: Not to encroach on your territory, but shouldn't the S in `FieldSet` be capital? Just because the XML tag comes from two words. But please discard me if you think lowercase is more in line with naming standards :)

src/components/NcFieldset/NcFieldset.vue
```diff
@@ -0,0 +1,174 @@
...
+
+<docs>
+### General
+
+Native `<fieldset>` element for grouping form elements with a legend.
+
+```vue
+<template>
+	<NcFieldset legend="Personal Information">
+		<NcTextField label="First Name" />
```
kra-mo:
```suggestion
		<NcTextField label="First name" />
```

Just a nitpick for these labels in the examples in general because of our writing guidelines.

ShGKme:
> Do you mean to adjust the input fields' leading padding to match? Then yes, it should just be aligned.

No, the question wasn't "should they be aligned" but is the margin supposed to be border radius + 1 grid baseline. I can see that it is 8 + 4 = 12px from the `.png` for the Nextcloud 33 theme, but I don't know that was the idea behind the value to adjust it to the previus versions.

kra-mo:
> No, the question wasn't "should they be aligned" but is the margin supposed to be border radius + 1 grid baseline. I can see that it is 8 + 4 = 12px from the `.png` for the Nextcloud 33 theme, but I don't know that was the idea behind the value to adjust it to the previus versions.

Okay, yes. 12px came from "it looked the best for fields both with and without a description" :)

Border radius + 1 grid baseline makes sense for me to describe it if you think it does as well.

ChristophWurst:
Backport to stable8?

kra-mo:
> Then yes, it should just be aligned.

And what I was mainly trying to say here was just that "it is not my territory, as long as it is aligned, I am happy" :)

susnux:
I think `name` is not really good as it confuses with technical `name` within forms.

But `label` is basically what this does, it labels the element in an accessible way (possible synonym would be 'caption').
Especially thinking for a11y its common to use the term `label` as this also will set the accessible name of the fieldset.

I mean yes its technically a `legend` but this is only down in the implementation, the user only sees this as the label / name of the element, similar to other elements.
So from the API point for me it makes sense to have consistent names for similar use cases, so if I - as a library user - use a component I did not use before I already know what kind of props I can expect :)

src/components/NcFieldset/NcFieldset.vue
```diff
@@ -0,0 +1,174 @@
...
+	 */
+	legend?: string
+	/**
+	 * Optional fieldset description. #description slot can be used for custom description content
+	 */
+	description?: string
+	/**
+	 * Hide the legend visually but keep it accessible for screen readers
+	 */
+	hideLegend?: boolean
```
susnux:
```suggestion
	label = undefined,
	description = undefined,
	hideLabel = false,
	hideDescription = false,
	noGap = false,
} = defineProps<{
	/**
	 * Label of the field set. `#label` slot can be used for custom formatted label content.
	 */
	label?: string
	/**
	 * Optional fieldset description. #description slot can be used for custom description content
	 */
	description?: string
	/**
	 * Hide the label visually but keep it accessible for screen readers
	 */
	hideLabel?: boolean
```

src/components/NcFormGroup/NcFormGroup.vue
```diff
@@ -0,0 +1,174 @@
...
+	 */
+	hideLegend?: boolean
+	/**
+	 * Hide the description visually but keep it accessible for screen readers
+	 */
+	hideDescription?: boolean
+	/**
+	 * Disable default fieldset content gap between content elements
+	 */
+	noGap?: boolean
```
susnux, line 37: Whats the use case?

src/components/NcFieldset/NcFieldset.vue
```diff
@@ -0,0 +1,174 @@
...
+
+const slots = defineSlots<{
+	/**
+	 * Fieldset content
+	 */
+	default?: Slot
+	/**
+	 * Custom legend content
+	 */
+	legend?: Slot
```
susnux: See other comment


```suggestion
	/**
	 * Custom label content
	 */
	label?: Slot
```

src/components/NcFieldset/NcFieldset.vue
```diff
@@ -0,0 +1,174 @@
...
+const getDescriptionId = () => hasDescription() ? descriptionId : undefined
+</script>
+
+<template>
+	<fieldset
+		:class="[$style.fieldset, { [$style.fieldset_noGap]: noGap }]"
+		:aria-describedby="getDescriptionId()">
+		<legend :id="labelId" :class="[$style.fieldset__legend, { 'hidden-visually': hideLegend }]">
+			<slot name="legend">
+				{{ legend || '⚠️ Missing legend' }}
```
susnux: See above


```suggestion
			<slot name="label">
				{{ label || '⚠️ Missing label' }}
```

src/components/NcFieldset/NcFieldset.vue
```diff
@@ -0,0 +1,174 @@
...
+
+const hasDescription = () => !!description || !!slots.description
+const getDescriptionId = () => hasDescription() ? descriptionId : undefined
+</script>
+
+<template>
+	<fieldset
+		:class="[$style.fieldset, { [$style.fieldset_noGap]: noGap }]"
+		:aria-describedby="getDescriptionId()">
+		<legend :id="labelId" :class="[$style.fieldset__legend, { 'hidden-visually': hideLegend }]">
```
susnux:
```suggestion
		<legend :id="labelId" :class="[$style.fieldset__legend, { 'hidden-visually': hideLabel }]">
```

ShGKme:
> PS: Not to encroach on your territory, but shouldn't the S in `FieldSet` be capital? Just because the XML tag comes from two words. But please discard me if you think lowercase is more in line with naming standards :)

In this case, it represents a native `fieldset` element, I'm keeping the native casing. Same as `RichContenteditable`.

ShGKme:
> I mean yes its technically a `legend` but this is only down in the implementation, the user only sees this as the label / name of the element, similar to other elements.

But the prop isn't for the end user, it is for the developers, and it represents `<legend>` element.

If we consider `<fieldset> + <legend>` only an internal detail that should not be reflected in the naming, do you also mean that we should rename the component itself? (Because `<fieldset>` is an internal implementation, and it could be just `<div role="group" aria-labelledby="ID">`)

kra-mo:
> Same as `RichContenteditable`.

Yes okay, fair :)

src/components/NcFieldset/NcFieldset.vue
ShGKme: Done

susnux:
> But the prop isn't for the end user, it is for the developers, and it represents <legend> element.

End user is the developer in that case and API wise it makes sense to have a consistent way of naming instead of forcing the developer to think about that component is used behind and what special cases it has.

We always use `label` for this where `name` is not appropriate, so I think we should stick with it instead of add a new one just because the underlying HTML tag uses a different name.
So at least our API is consistent when HTML is not.

ShGKme:
> > But the prop isn't for the end user, it is for the developers, and it represents  element.
> 
> End user is the developer in that case and API wise it makes sense to have a consistent way of naming instead of forcing the developer to think about that component is used behind and what special cases it has.
> 
> We always use `label` for this where `name` is not appropriate, so I think we should stick with it instead of add a new one just because the underlying HTML tag uses a different name. So at least our API is consistent when HTML is not.

What about the second paragraph of the quoted comment? :)

susnux:
> What about the second paragraph of the quoted comment? :)

You mean:

> (Because <fieldset> is an internal implementation, and it could be just <div role="group" aria-labelledby="ID">)

Do you have a better name? I think Fieldset is quite easy to understand as a set of fields.

src/components/NcFormGroup/NcFormGroup.vue
ShGKme: General good practice for UI components — if a component has not only its own styles but also its position/content (its margin, slots styles) - it should be optional.

It could be a list of checkboxes without a gap, or some custom set of elements.

Alternative - ask to wrap content in a div.

ShGKme:
> Do you have a better name?

I used `NcFieldset + :legend` as the names for `<fieldset> + <legend>`.

If we consider `:legend` a bad name because it is semantically a `label` internally implemented via `<legend>`, then a component can be called `<NcFormGroup>` because it is semantically a `group` internally implemented via `<fieldset>`.

susnux:
> If we consider :legend a bad name because it is semantically a label internally implemented via <legend>, then a component can be called <NcFormGroup> because it is semantically a group internally implemented via <fieldset>.


As said its more about consistency, as a developer I expect API to be consistent. If other naming-elements are called `label` I also expect `label` here and not have to think 'ah yes in HTMl they decided to go with a new legend element instead of label' just to know i have to use the `legend` prop just for this component.

The name of the component I do not really case as long as the use case is clear.

susnux:
> The name of the component I do not really case as long as the use case is clear.

also for #7690 : I am not sure the naming is good. Because its not really a group as a group would semantically require a name. I expected this component here (`NcFieldset`) to be called `NcFormGroup` as discussed earlier and the other component `NcFormBox` or similar.

I do not care about the actual name of the component, but they should be connected and represent their use case :)

ShGKme:
> As said its more about consistency, as a developer I expect API to be consistent

Me too. But "consistency" doesn't mean that the same world should be used everywhere. `name` is a great example, where we have the same "consistent" prop title when it means different things.

Here for consistency the prop was named `legend` for the `<legend>`'s value of the `<fieldset>` in the `NcFieldset` component.

ShGKme:
Rebased to have clean CI, no changes in the code

susnux:
> Here for consistency the prop was named legend for the <legend>'s value of the <fieldset> in the NcFieldset component.

But its basically a `label` there is no reason to have `legend` except from trying to keep it in sync with HTML which makes no sense because in contrast to `name` vs `title` there is no conflict with `label`.
The legend labels the fieldset so we in a *custom* component can use a prop name that is consistent with *our* API instead of this one special case in HTML.

ShGKme:
> is no reason to have `legend`

A reason — to not give one thing a name of another thing when having a component directly representing an existing element.

Otherwise let's not call it a fieldset to not confuse with an existing element.

susnux:
> Otherwise let's not call it a fieldset to not confuse with an existing element.

I have no problem with a different component name, not as this is a quite generic component what do you think?
Because I think that the best alternative name would be `NcFormGroup` but how to call the other component then?

ShGKme:
> not as this is a quite generic component what do you think

I didn't understand what you mean here...



> Because I think that the best alternative name would be `NcFormGroup` but how to call the other component then?

As was proposed - `NcFormBox`, because it is visually a "box", quite a common term in UI kits.

src/components/NcFormGroup/NcFormGroup.vue
```diff
@@ -0,0 +1,174 @@
...
+const getDescriptionId = () => hasDescription() ? descriptionId : undefined
+</script>
+
+<template>
+	<fieldset
+		:class="[$style.fieldset, { [$style.fieldset_noGap]: noGap }]"
+		:aria-describedby="getDescriptionId()">
+		<legend :id="labelId" :class="[$style.fieldset__legend, { 'hidden-visually': hideLabel }]">
+			<slot name="label">
+				{{ label || '⚠️ Missing label' }}
```
GVodyanov, line 69: Do we want to always have a label? I can imagine a use-case like having a Fieldset inside a Modal which already has a title. I feel something might eventually come up where you would want to just not have a label, and having it show "Missing label" would be uncomfortable.

If we want to always have a standard label as a design decision though, I could understand it.

kra-mo: I want to ask again:

> Just so I understand: where would the FormGroup sit in the tree relative to individual fields and the Fieldset itself?

But from what I understand right now, from a design POV, there is definitely  a case where there is no label, including in the mockups.

susnux: > Do we want to always have a label? 

If its a fieldset we must. Its obligatory for wcag. If its only visually then its the form box not a real group.

susnux:
> As was proposed - NcFormBox, because it is visually a "box", quite a common term in UI kits.

As said I am 💯 for that name. Meaning form group and form box.
But maybe even if its not 100% accurate also call it form group button, not form box button?

ShGKme:
- Component was renamed `NcFieldset` -> `NcFormGroup`
- Prop was renamed `legend` -> `label`, `hideLegend` -> `hideLabel`
- Auto-margin between `NcFormGroup` has been removed

PR #7690, opened 2025-10-20
feat: add `NcFormBox` and adjust `NcRadioGroup` to `NcFormbox` and `NcFormGroup`

ShGKme:
### ☑️ Resolves

- A part of: https://github.com/nextcloud-libraries/nextcloud-vue/issues/7122
- `NcFormBox*` components are coming in individual PRs

### 🖼️ Screenshots

image

Before | After
-- | --
image | image
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/NcRadioGroup/NcRadioGroup.vue
+++ b/src/components/NcRadioGroup/NcRadioGroup.vue
@@ -7,7 +7,8 @@
 import type { Slot } from 'vue'
 
 import { computed, provide, ref, warn } from 'vue'
-import { createElementId } from '../../utils/createElementId.ts'
+import NcFormBox from '../NcFormBox/NcFormBox.vue'
+import NcFormGroup from '../NcFormGroup/NcFormGroup.vue'
 import { INSIDE_RADIO_GROUP_KEY } from './useNcRadioGroup.ts'
 
 const modelValue = defineModel<string>({ required: false, default: '' })
@@ -40,7 +41,6 @@ defineSlots<{
 	default?: Slot
 }>()
 
-const descriptionId = createElementId()
 const buttonVariant = ref<boolean>()
 
 provide(INSIDE_RADIO_GROUP_KEY, computed(() => ({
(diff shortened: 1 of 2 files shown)


ChristophWurst:
Adjusting PR state to labels

ChristophWurst:
> * Backport to `stable8` for maintained Vue 2 version or not applicable

This is ticked. Does that mean the component is not ported for 8.x?

src/components/NcFormGroup/NcFormGroup.vue
susnux, line 1: As mentioned in #7689 originally this component was called `NcFormBox` which also is not the best name.
But it made it clear that this is not semantically a group like fieldset would be.
This is more for the visual adjustments within a group.

So maybe we should call the new components in a way that its clear how they are related and how they are intended to be used?
- NcFormGroup ↔️ #7689 
- NcFormGroupLayout ↔️ this layout component
- NcFormGroupButton ↔️ the button in #7440

ShGKme: I don't think that `NcFormGroupLayout` makes it clearer what this component does. It is not a form layout like as column of labels and inputs.

> how they are intended to be used

This component is intended to be used as a visual group. For components that then looks like a solid group even if it isn't semantical and has no label.

The `FormBox` option was an alternative, to avoid `group`, and because it is visually a box (also "box" is a common name for such things in UI kits).

> NcFormGroupButton 

If this component is `NcFormGroupLayout`, when the button for it should be `NcFormGroupLayoutButton`, and then the Copy button is something like `NcFormGroupLayoutCopyButton`.

ShGKme:
Rebased to have clean CI, no changes in the code

src/components/NcFormGroup/NcFormGroup.vue
susnux: >  It is not a form layout like as column of labels and inputs.

But its purely for layout not for any semantic usage.

> If this component is NcFormGroupLayout, when the button for it should be NcFormGroupLayoutButton, and then the Copy button is something like NcFormGroupLayoutCopyButton.

Why? It would be a button to be used in a form group, which also works in the layout only variant but generally in both.

For users the easiest would be to only have one component that changes behavior based on whether a `label` is passed or not. Because from the long going discussion I feel either its me or it will really confuse to have two "unrelated" components doing the same but one with label one not.

ShGKme: > But its purely for layout not for any semantic usage.

And not a layout in a general "layout" meaning. That's why another option was `FormBox`.

> Why? It would be a button to be used in a form group, which also works in the layout only variant but generally in both.

Following the naming convention, because this is a component to be used specifically inside another one.

> components doing the same

Why are they doing the same? They have nothing in common.

One component is a container with a heading and a description.

Another one is a group/box of special components forming a visually solid group.

susnux: So the one is a solid group with a label and description and the other is just a solid group?
This sounds confusing and likely we will see this wrongly used in some apps then.

That is why I am a bit picky here to make sure:
1. we really need to expose both components and cannot just use one internally
2. if we really need both then make clear that one is just layout and the other is semantic grouping.

susnux: So basically the question is:

- Do we need both or should we just have on `NcFormGroup` that changes semantic behavior if no `label` is set?

:heavy_plus_sign: very easy to use
:heavy_minus_sign: combine two things in one component

If not then ok, then we also can stick with how it is, but its better to think about it before as adding new stuff is always possible, getting rid of things is harder.

ShGKme: > So the one is a solid group with a label and description and the other is just a solid group?

No, it doesn't sound like the comment above.

One is a classic **fieldset**. A set of form elements with a heading and description. Directly representing an existing native element.

Another is a group (in a simple meaning for a general developer, not WCAG role="group" area meaning) of very specific group item components. That visually looks like closely spaced elements rounded together as an organic (solid) whole (a box).

Not a layout for any form element to put them horizontally/vertically like form layout components do.

If we consider `fieldset` and `group` confusing (For example, because developers don't use `<fieldset>` so often nowadays, or because a group sounds like `role="group"` in the context of a11y), we can also call it a group and a box.

> 1. we really need to expose both components and cannot just use one internally

We can implement it as a single component for both purposes. But then:
1. It has a complex implementation, indirectly to the user deciding what HTML structure, semantics, effect on elements, and gap between elements it has
2. It has unclear conditions to force providing a visual/hidden label. So we just have to force it every time, including a single-button group

Do we really need to implement a universal component for both things?

ShGKme: > If not then ok, then we also can stick with how it is, but its better to think about it before as adding new stuff is always possible, getting rid of things is harder.

That's what I always repeat when public API change is quickly merged with yolo 😉

susnux: > If we consider fieldset and group confusing (For example, because developers don't use <fieldset> so often nowadays, or because a group sounds like role="group" in the context of a11y), we can also call it a group and a box.

Thats what I meant and what I am in favor of :)

susnux: thanks for renaming :)

ShGKme:
- Rebased
- Fixed a minor issue with horizontal items stretching (no before/after difference now for the last example)
- Fixed a typo in re-export

AndyScherzinger:
@susnux can you have a look please 🙏 Thanks in advance!

susnux:
> @susnux can you have a look please 🙏 Thanks in advance!

Sure I am on it

PR #7748, opened 2025-10-29
fix(NcRichText): start heading from h4

ShGKme:
### ☑️ Resolves

- Fix https://github.com/nextcloud-libraries/nextcloud-vue/issues/7743
- Level down headings (-3 levels)
- Also, keep it small visually, so users don't feel they need to use low-level heading just because 1-3 levels are too large

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/NcRichText/NcRichText.vue
+++ b/src/components/NcRichText/NcRichText.vue
@@ -546,6 +546,13 @@ export default {
 			delete props.children
 
 			if (!String(type).startsWith('#')) {
+				// <h1>..<h3> headings are used on the page for semantic structure
+				// Using them for user content leads to accessibility issues
+				// Levelling down headings to start from <h4>
+				if (['h1', 'h2', 'h3', 'h4', 'h5', 'h6'].includes(String(type))) {
+					type = `h${Math.min(+String(type)[1] + 3, 6)}`
+				}
+
 				let nestedNode = null
 				if (this.useExtendedMarkdown) {
 					if (String(type) === 'code' && !rehypeHighlight.value
@@ -681,8 +688,16 @@ export default {
 		font-weight: bold;
 	}
 
-	h1 {
-		font-size: 30px;
+	h4 {
+		font-size: 20px;
+	}
+
+	h5 {
+		font-size: 18px;
+	}
+
+	h6 {
+		font-size: 15px;
 	}
 
 	ul, ol {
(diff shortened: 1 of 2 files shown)


ShGKme:
Adjusted tests

ShGKme:
> But not sure we could always guarantee no heading level is skipped?

We cannot. But skipping isn't a problem if it doesn't break the structure. Users can also start with `h6`.

We can also use `<section>` element, but it doesn't seem to be well supported in the context of headings...

And the worst idea I have — actually query for the heading to start from the next level. But I'd avoid it.

src/components/NcRichText/NcRichText.vue
```diff
@@ -546,6 +546,13 @@ export default {
 			delete props.children
 
 			if (!String(type).startsWith('#')) {
+				// <h1>..<h3> headings are used on the page for semantic structure
+				// Using them for user content leads to accessibility issues
+				// Levelling down headings to start from <h4>
+				if (['h1', 'h2', 'h3', 'h4', 'h5', 'h6'].includes(String(type))) {
+					type = `h${Math.min(+String(type)[1] + 3, 6)}`
```
susnux, line 553: Is this always valid? What if the component is used right under a h1?
Maybe we need some magic like `$el.closest` and then check whats the closest headline?

ShGKme: > What if the component is used right under a h1?

Answered in the comment above

susnux: can we maybe apply some classes to visually differentiate `#### heading` from `###### heading`?

ShGKme: IMO, distinguish `#####` from `######` must be already an edge case. It should mean that 4 heading levels were already used, and you still need 2 more levels. It's hard to imagine for me even in a dissertation, and this component is just for a comment/message.

And if users use h5-h6 without upper levels, then there is probably something wrong. For example, they may do it because h1-h3 are way too large. This is why I'm keeping them smaller here.

@kra-mo Maybe you have any opinion here? Should we have different styles for all h1..h6 and if we do, which style (keeping h1 small enough for a comment/message)?

jancborchardt (approved):

src/components/NcRichText/NcRichText.vue
kra-mo: > @kra-mo Maybe you have any opinion here?

Personally, I don't see the need to differentiate visually. Some apps make headings even smaller than body text or use all-caps to indicate the levels, but I think both are suboptimal.

ShGKme:
Merging with 2 designers approving the design with equal low-level heading styles.

PR #7778, opened 2025-11-05
fix(NcActions): handle expensive height computation by popover library

Antreesy:
### ☑️ Resolves

- Replaces #5806
  - `autoBoundaryMaxSize` lets floating vue resize the popper inner container to the available size
  - drawback - losing feature of showing 50% of last item cropped, if it doesn't fit the screen
  - benefits - no more jumping and second re-render (first by library, then manually by us), no memory leaks

### 🖼️ Screenshots

🏚️ Before

https://github.com/user-attachments/assets/dcfa199e-0612-4ac3-9c01-79fb868b02de

🏡 After

https://github.com/user-attachments/assets/2285e070-38a9-4d0c-9c75-993c95f3930c

### 🚧 Tasks

- [x] Check effect on Memory/Performance
- [ ] Check for side effects

### 🏁 Checklist

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -1342,7 +1342,6 @@ export default {
 		onOpened() {
 			this.$nextTick(() => {
 				this.focusFirstAction(null)
-				this.resizePopover()
 
 				/**
 				 * Event emitted when the popover menu is opened.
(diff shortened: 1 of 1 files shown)


susnux:
So its now 4ms slower and it crops the actions so that while scrolling they do no longer work? :see_no_evil:

Antreesy:
> So its now 4ms slower

`After` is general style recalculation - it isn't caused by NcActions. I am receiving the same results if component is completely removed

> and it crops the actions so that while scrolling they do no longer work?

I did not get this part. Actions work same way as before. Is it about previous 50% crop on last visible action?

susnux:
> I did not get this part. Actions work same way as before. Is it about previous 50% crop on last visible action?

I was talking about second 4 of the screencast:
grafik

But after watching both again, the same happens also before.

susnux:
So while it now takes full height, it looses the main reason for the manual resizing:
If the content is longer than the boundary then we want to have it cut off at 50% of the element so the user knows about that there is more to scroll.
With only the library it does no longer cut off at 50% of the next element.

So codewise I am fine, but just needs approval of @jancborchardt (added that feature).

DorraJaouad:
image


Effect is confirmed and rendering is faster although there might be still some bottleneck. Behavior wise, it is a bit unusual to have popover auto adjusting while scrolling

Antreesy:
> so the user knows about that there is more to scroll.

Drawback would be then only for browsers, who don't show the scrollbar by default, right?

src/components/NcActions/NcActions.vue
```diff
@@ -1109,38 +1108,13 @@ export default {
...
 
-	setup(props) {
+	setup() {
 		const randomId = createElementId()
-		const triggerRandomId = `trigger-${randomId}`
-
 		const triggerButton = ref()
 
-		const { top, bottom } = useElementBounding(triggerButton)
-		const { top: boundaryTop, bottom: boundaryBottom } = useElementBounding(toRef(() => props.boundariesElement))
```
susnux: I wonder if its not also possible to just improve this function and keep the features.

It seems like if we could share the observer on the boundary element we would win a lot.

Antreesy: We can cut some corners, agree (or even try `updateTiming: 'next-frame'` to delay a bit), but still for each mounted NcActions it is:
- 2 useElementBounding is 2 listeners and 2 observers each;
- useWindowSize is 2 listeners each;

I'm not sure it's all worth to keep a feature that targets maybe 4-5 actual tall menu list use cases in some browsers that hide scrollbar by default. Maybe better to force scrollbar?

ShGKme: `useWindowSize` definetely can be shared.

`useElementBounding` can be shared with a map by the boundary, but it only wins something with reusable boundaries.

Antreesy: Done in 7808, I will rebase this one later once it's merged

jancborchardt (approved):
Really nice improvement so I’d approve. :) We can check later if there is adjustment needed for cutting of at 1/2 of entries, or if that’s an issue at all (probably depending on system and scrollbar visibility).

susnux:
Isnt this solved by https://github.com/nextcloud-libraries/nextcloud-vue/pull/7808 ?

Antreesy:
> Isnt this solved?

Performance - mostly solved.
But it still triggers re-rendering of a popover after mounting, it looks jumpy and frozen, if trigger button is scrolled (up and down). So I would still like to opt in for native library solution

ShGKme (approved):

ShGKme:
@Antreesy backport?

PR #7779, opened 2025-11-05
feat: add NcFormBoxButton

ShGKme:
### ☑️ Resolves

- Fix: https://github.com/nextcloud-libraries/nextcloud-vue/issues/7123
- Reworked: https://github.com/nextcloud-libraries/nextcloud-vue/pull/7440

Described in the docs.

### 🖼️ Screenshots

image

image

image

image

### 🏁 Checklist


ShGKme:
- Rebased onto main
- Added a small comment in CSS, no actual changes

src/components/NcFormBox/NcFormBoxItem.vue
susnux, line 1: new folder structure? Why not put it inside `components/NcFormBox/` like we do with some other shared components?

ShGKme: To make it clear that it is not a public API and prevent "I forgot to exclude from the docs" situations.

But it is an internal thing, so we can refactor the structure any moment later.

ShGKme: Is there anything Public API/design/implementation related?

Since this is the only one comment in the review, I don't know, if it is the full review or you submitted with this comment only

susnux: > Is there anything Public API/design/implementation related?

No otherwise its fine!

susnux: > To make it clear that it is not a public API

But only what is exported from the `index.ts` is public API.
IMHO having a new folder structure is more confusing than helpful.

> and prevent "I forgot to exclude from the docs" situations.

True but thats due to non-explicitly defined styleguidist config but also not hard to add.

ShGKme: > But only what is exported from the `index.ts` is public API.

For the machine during building the NPM package - yes. But no part of the file structure shows it => a developer needs to check almost 100 lines to know it.

> True but thats due to non-explicitly defined styleguidist config but also not hard to add.

Or due to having no structural distinguish between publicly documented modules and internal files. Styleguidist here sees the same as the human developer.

> IMHO having a new folder structure is more confusing than helpful.

What part is confusing?
Isn't it helpful to avoid problems that we really had many times?
1. Private component is shown in public documentation
2. Private component is exported from the package by mistake
3. Public component is forgotten to be re-exported from the package by mistake  

---

If the `shared` name is the problem, or its position in the structure — we can discuss it.

For example:
- it can be `internal` or `private` instead of `shared`
- it can be in the `components` instead of split

If **not** having public/private component mixed is the problem — I disagree.

GVodyanov (approved):
Tested, works, and code looks good!

I wanted to mention though that it would be nice to have the "copy" action to be implemented inside of the component, seeing as in the current state it would require every app to create their own implementation. Seeing as this is time-critical though, I think it's totally fine to do that in another PR later.

ShGKme:
@GVodyanov NcFormBoxCopyButton is in local draft, but will be pushed later since it has lower priority over base components

src/components/NcFormBox/NcFormBoxItem.vue
```diff
@@ -0,0 +1,190 @@
...
+		descriptionId?: string
+	}>
+	/** Icon content */
+	icon?: Slot
+}>()
+
+const { formBoxItemClass } = useNcFormBox()
+
+const descriptionId = createElementId()
+const hasDescription = () => !!description || !!slots.description
```
Antreesy, line 61: Is it fine to use that as function and not reactive/computed?

src/components/NcFormBox/NcFormBoxItem.vue
```diff
@@ -0,0 +1,190 @@
...
+				[$style.formBoxItem_inverted]: invertedAccent && hasDescription(),
+				[$style.formBoxItem_legacy]: isLegacy,
+			},
+		]">
+		<span :class="$style.formBoxItem__content">
+			<component
+				:is="tag"
+				:class="[$style.formBoxItem__element, itemClasses]"
+				v-bind="$attrs"
+				@click="$emit('click', $event)">
```
Antreesy, line 80: isn't this click listener too deep (only on label)? Also structure-wise div > span > **div** by default IMO isn't fine, if we expect inline component, should be span|a|button?

src/components/NcFormBox/NcFormBoxItem.vue
```diff
@@ -0,0 +1,190 @@
...
+	transition-duration: var(--animation-quick);
+	transition-timing-function: linear;
+	user-select: none;
+	cursor: pointer;
+
+	* {
+		cursor: inherit;
+	}
+
+	&:has(:disabled) {
```
Antreesy, line 130: `:has` might add some extra scans here. Don't we know it's going to be `formBoxItem__element:disabled`?

src/components/NcFormBox/NcFormBoxItem.vue
ShGKme: slots isn't reactive, so it is not possible to use computed

Performance-wise it is fine because it is a tiny computation, even smaller than computed internals

src/components/NcFormBox/NcFormBoxItem.vue
ShGKme: I'm trying to avoid `:has()` as much as possible, but in this case I don't see a simple way to avoid it.

The only solution I see is always providing the `disabled` prop and set the class depending on the prop and not the element `:disabled` state. But it means:
- We don't pass `disabled` props for unsupported elements lika `<a>` links (or we manually disable the link) - some additional code I wanted to avoid
- We need to make sure prop and element states are consistent since they are in different components (should be fine if this component is internal only)

So I just hope this won't be a problem because this component isn't supposed to be used in a large content / popovers etc.

src/components/NcFormBox/NcFormBoxItem.vue
ShGKme: I am unsure if I get this one. The click listener is on the interactive element itself (`<button>`, `<label>`, `<a>`), so, it listens on the element that is supposed to be clicked.

> Also structure-wise div > span > div by default

True, removed the default `tag` value, made it required. So it is `div > span > button` like.

ShGKme: Pushed

src/components/NcFormBox/NcFormBoxItem.vue
susnux: > For the machine during building the NPM package - yes. But no part of the file structure shows it => a developer needs to check almost 100 lines to know it.

Well currently it is:

`components/NAME/NAME.vue` every other vue file in that folder is private, so no need to check anything

susnux: This now forces you to check 2+ folder trees for find components.
Before you just know 'its a FormBoxItem so either its a public component and in FormBoxItem folder or its private and in FormBox folder'.
Now you need to open a completely different folder tree for finding components, while being in another tree for using it.

Also its not really matter as a developer if the component is public or not, because you can use both.

But I agree that currently its not good to have to care if you e.g. put a composable into a subfolder (= exported) or not (=internal).
Maybe we should really group all of them in some 'public' (or 'internal') folders, but then I would prefer to do this with all in one PR and not mix things up.

susnux: So I would prefer moving it in this PR back to the normal location `/src/components/NcFormBox/NcFormBoxItem.vue` and do a follow up where we e.g. move all internal components and composables to consistent new locations. So we can just continue this one without mixing build related changes.

ShGKme: > components/NAME/NAME.vue

Even if we call `components/NAME/NAME.vue` public, it doesn't specify the position for the private component, does it? Works if it is a subcomponent of a specific public one, but what if it is shared?

- `components/Foo/Foo.vue` -> uses -> `Shared.vue`
- `components/Bar/Bar.vue` -> uses -> `Shared.vue`

Ok, `Foo.vue` and `Bar.vue` are public. Where `Shared.vue` is supposed to be?
If in Foo - why not Bar. If in Bar - why not Foo.

> every other vue file in that folder is private

The following components are "every other" and they are public:
- `components/NcRichContenteditable/NcMentionBubble.vue`
- `components/NcRichContenteditable/NcAutoCompleteResult.vue`

---

So:
1. No, not only such paths are `public`
2. No, having such paths doesn't specify the path for `private` even if p.1 were actual

ShGKme: > Also its not really matter as a developer if the component is public or not, because you can use both.

It does matter a lot. In a private component I can do everything, rewrite it, rename, change the interface, throw it away. In a public component I need to be careful and not break anything. It is easy to change it and hard to bring back after being released.

ShGKme: > a completely different folder

I also proposed 2 alternative names for the folder and 1 alternative location.

Again, I'm fine with a different location, folder name or both.

I'm not fine with just mixing public with private files together and only separate by including/excluding in re-export and the list on the docs.

ShGKme: @susnux I pushed a commit moving the component to the `components/NcFormBox/NcFormBoxItem.vue`. I don't think this is the best place for it, but it can be discussed later, outside the `NcFormBoxButton` PR.

ShGKme:
@susnux Since your comment is not marked as blocking, review is not on `Request changes` and the discussed part is purely internal and stylistic, I'm merging the PR.

ShGKme:
Rebased onto main and squashed, no other changes.

PR #7781, opened 2025-11-05
feat: add NcFormBoxSwitch

ShGKme:
### ☑️ Resolves

- A switch toggle for the button
- Based on: https://github.com/nextcloud-libraries/nextcloud-vue/pull/7779

#### ⚠️ Why not NcCheckboxRadioSwitch

- The design is quite different — I do not want to introduce the 3rd structure to a super component
- Much can be reused with `NcFormBoxButton`
- **Switch toggle** is a straightforward UI element, it doesn't need everything we have for checkboxes, radio, shift+click features as well as being in a group

### 🖼️ Screenshots

image

Accessible input position:

image

### 🏁 Checklist


ShGKme:
Rebased after merging the base component

src/components/NcFormBoxSwitch/NcFormBoxSwitch.vue
```diff
@@ -0,0 +1,155 @@
...
+	/** Disabled state */
+	disabled?: boolean
+}>()
+
+const emit = defineEmits<{
+	/** Switch is toggled ON */
+	enable: []
+	/** Switch is toggled OFF */
+	disable: []
+}>()
```
susnux, line 36: I do not really get the use case, its basically duplicates the model value.
All you can do with the events you can do with `update:modelValue` as well?

ShGKme: > All you can do with the events you can do with update:modelValue as well?

You can, with additional actions and less explicit.

With events:
```html
<template>
  <NcFormBoxCopyButton @enable="handleEnabling">
</template>
```

Without:
```html
<script setup>
/**
 * Handle mopdel value change
 *
 * @param newValue - new modelValue
 */
function handleChange(newValue: boolean) {
  if (newValue) {
    handleEnabling()
  }
}
</script>

<template>
  <NcFormBoxSwitch @update:modelValue="handleEnabling">
</template>
```

or

```html
<template>
  <NcFormBoxSwitch @update:modelValue="$event && handleEnabling()">
</template>
```

susnux: TBH: I do not like duplication like that, its already available in a universal way, so other events just bloats the component signature.
But if thats considers useful by anyone else :shrug:

kra-mo (changes requested):
Looks good, but quickly toggling it causes the text to be selected.

Could we somehow disable the double-click to select text behavior? Or more generally, disable text selection for FormBoxButton and co. maybe?

ShGKme:
@kra-mo user selection is disabled in both components. What web-browser you are testing with?

kra-mo:
> @kra-mo user selection is disabled in both components. What web-browser you are testing with?

Interesting because it does work for one but not the other. macOS Safari.

ShGKme:
I was just about to push removing the event you didn't want to have, split them into a separated PR =\

ShGKme:
I'll consider your merge as "I like the events" then 👀

PR #7787, opened 2025-11-05
feat: add NcFormBoxCopyButton

ShGKme:
### ☑️ Resolves

- Ref: https://github.com/nextcloud-libraries/nextcloud-vue/pull/7779#pullrequestreview-3423161191

### 🖼️ Screenshots

copy

image

### Fallback with no clipboard access

fallback

### 🏁 Checklist


ShGKme:
Rebased after merging the base component

nickvergessen (approved):

src/components/NcFormBoxCopyButton/useCopy.ts
```diff
@@ -0,0 +1,50 @@
...
+			await navigator.clipboard.writeText(value)
+
+			isCopied.value = true
+			setTimeout(() => {
+				isCopied.value = false
+			}, DELAY)
+		} catch (error) {
+			logger.error('Failed to copy to clipboard', { error })
+		}
+	}
```
susnux, line 40: This should rather have a fallback for non-secure contexts. We had a lot of complains in server and added that there always.

Something like:

```suggestion
		const value = toValue(content)
		if (!value) {
			logger.warn('No content to copy')
			return
		}
		try {
			await navigator.clipboard.writeText(value)
		} catch {
			window.prompt(t('Clipboard not available. Please copy the value manually.'), value)
		} finally {
			isCopied.value = true
			setTimeout(() => {
				isCopied.value = false
			}, DELAY)
		}
	}
```

ShGKme: - Ref: https://github.com/nextcloud-libraries/nextcloud-vue/issues/7790

ShGKme: > `window.prompt(t('Clipboard not available. Please copy the value manually.'), value)`

Feels quite ugly. But a nice workaround without complete re-design.

I'd avoid adding a long instruction string to the library, though.

And I'd keep the `isCopied` value changed only when we are sure, the value has been copied.

Pushed a fixup.

susnux: But this now means the copy event is not emitted, while it is either copied or the the user did not care.
This will very likely cause confusion due to different behavior and potential bugs later.

Antreesy: native browser promp is good enough, to not overengineer it?) consider that non-secure context is not something we encourage to use, and there is still a way to copy it. Confused people can click it once more to do `Ctrl+C`

ShGKme: Done. Also removed logging (seems unneeded) and check for an empty content.

susnux: > is not something we encourage to use

But a lot of users do it anyways :sweat_smile:
There was such a high demand on fixing that^^

susnux (approved):
looks good

AndyScherzinger (approved):

ShGKme:
- Squashed

PR #4924, opened 2023-12-06
feat(NcRichContenteditable): put caret in the end when focus input

Antreesy:
### ☑️ Resolves

* Put caret to the end of inputted text when focus it, instead of start

### 🖼️ Screenshots

Casted some tests (with mouse clicks, range selections, keyboard tabbing)

[Screencast - focus tests](https://github.com/nextcloud-libraries/nextcloud-vue/assets/93392545/88b89afe-b91c-4aeb-aa7b-29a144155bc3)

### 🚧 Tasks

- [ ] Manual tests:
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
- [ ] TODO
  - [ ] Check if code could be reused:

https://github.com/nextcloud-libraries/nextcloud-vue/blob/9dcd33b3810293a23e9dd964d60814f97639c8e2/src/components/NcRichContenteditable/NcRichContenteditable.vue#L700-L705

### 🏁 Checklist

- [ ] ⛑️ Tests are included or are not applicable
- [x] 📘 Component documentation has been extended, updated or is not applicable

PR #7791, opened 2025-11-06
fix(NcRichContenteditable): keep previous cursor position on focus

Antreesy:
### ☑️ Resolves

- Since #4924 either something was changed in browsers, or wasn't tested in different OS properly
- ATM on Linux, unfocusing page with focused NcRichContenteditable and the focus the page back trigger `@focus="moveCursorToEnd"` event, effectively setting cursor to the end
- Now, we check if element was previously focused/has a selection range, and keep it as-is
- on Windows, can be tested with `$0.blur(); $0.focus()`

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/NcRichContenteditable/NcRichContenteditable.vue
+++ b/src/components/NcRichContenteditable/NcRichContenteditable.vue
@@ -715,6 +715,12 @@ export default {
 			if (!document.createRange) {
 				return
 			}
+
+			if (window.getSelection().rangeCount > 0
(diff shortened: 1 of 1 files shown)


nickvergessen:
Confirmed to fix the problem in:
- Ubuntu + Chromium
- Ubuntu + Firefox

Antreesy:
Confirmed on macOS - Safari

ShGKme (approved):

PR #7807, opened 2025-11-07
fix(NcHotkeyList): align with form labels

ShGKme:
### ☑️ Resolves

- The offset was changed between NcHotkeyList and other components implementation
- Follow-up: share the CSS variable (somehow)

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/NcHotkey/NcHotkey.vue
+++ b/src/components/NcHotkey/NcHotkey.vue
@@ -58,10 +58,11 @@ const NO_LABEL_WARNING = '⚠️ NcHotKey must have a label or slot content'
 
 <style lang="scss" module>
 .hotkey {
+	--form-element-label-offset: calc(var(--border-radius-element) + var(--default-grid-baseline));
 	display: flex;
 	flex-direction: row-reverse;
 	align-items: stretch;
-	padding-inline: var(--border-radius-element); /* Align with form group */
+	padding-inline: var(--form-element-label-offset);
 
 	&:first-child {
 		.hotkey__label,
--- a/src/components/NcHotkeyList/NcHotkeyList.vue
+++ b/src/components/NcHotkeyList/NcHotkeyList.vue
@@ -41,13 +41,14 @@ const labelId = `NcHotkeyList_${createElementId()}`
 
 <style module>
 .hotkeyList {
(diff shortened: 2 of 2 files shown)


susnux:
unrelated but it looks a bit weird that the shortcuts are not fully expanded (width)

kra-mo:
> unrelated but it looks a bit weird that the shortcuts are not fully expanded (width)

I agree. If we're not gonna limit the width of the rest for now, maybe it's best to not limit the width of the shortcuts either.

ShGKme:
> unrelated but it looks a bit weird that the shortcuts are not fully expanded (width)

Comes from the design

> I agree

That was your request 🥲

susnux:
> I agree. If we're not gonna limit the width of the rest for now, maybe it's best to not limit the width of the shortcuts either.

IMHO limiting the width looks weird, the content should use the full width of the dialog.
If we want to have it tighter then we should decrease the dialog and not introduce empty space on the side, no?

ShGKme:
> If we want to have it tighter then we should decrease the dialog and not introduce empty space on the side, no?

We use NcDialog, which has a predefined size including navigation. And which then uses NcModal with predefined sizes.

To have a single dialog in a specific width, we need more than a size setting.

susnux:
> To have a single dialog in a specific width, we need more than a size setting.

Well basically thats where the presets come from: Usage.
If we decide the normal dialog needs to be tighter (this usage here) then the preset is basically to be adjusted.
Otherwise we end with a bunch of randomly sized dialogs.

ShGKme:
> If we decide the normal dialog needs to be tighter (this usage here) then the preset is basically to be adjusted.

And if we decide that every normal dialog is the settigns dialog

kra-mo:
> That was your request 🥲

Yeah, sorry, I do eventually want to limit the width, but definitely of the entire form together. Doing it as a one-off maybe isn't the best.

I mean it's not too bad as it stands, and it fixes the issue of the labels being too far apart from the keys (as here it does matter, since they are not clickable) but I do agree that it looks weird.

I think we need to discuss what to do exactly. I do think limiting the dialog width further would probably be the best though.

kra-mo:
For what it's worth, variable dialog widths are common across design systems as well. Not everything necessarily needs to adhere to a single size.

PR #7811, opened 2025-11-07
fix(NcHotkeyList): align with the section

ShGKme:
### ☑️ Resolves

- A part of https://github.com/nextcloud-libraries/nextcloud-vue/pull/7807

Before | After
-- | --
image | image


ShGKme:
> Maybe then we can discuss later if/how we want to limit dialog sizes in general. I'll open an issue for it.

- We can do it as a part of the dialog redesign: https://github.com/nextcloud-libraries/nextcloud-vue/issues/7641

ShGKme:
> Sorry for the miscommunication, I guess it should have been thought through more ahead of time.

No problem, removing it is just removing a line of code.

PR #7817, opened 2025-11-07
fix(NcListItem): do not mount NcAction until necessary

ShGKme:
### ☑️ Resolves

- Fix performance issues with active/massive `NcListItem`s rendering with (unforced) actions
- Only render `NcActions` when needed instead of hidding visually
- I checked the history and did not found any specific reason to use `v-show`, it looks like that's just how it used to be
- Works with keyboard, focus, mouse over etc. (not worse than in the past)

### 🏁 Checklist


ShGKme:
This change has a regression in the opposite way - now hovering `NcListItem`s is a bit slower. Noticed it on battery safe mode.

PR #7837, opened 2025-11-10
feat(NcAppSettingsDialog): add version on the bottom and `noVersion` prop to disable it

ShGKme:
### ☑️ Resolves

- Ref: https://github.com/nextcloud/server/issues/55667
- I missed that it was also a part of the design
- ⚠️ I've removed the "version" part to remove a requirement to have a new translated string and simplify it for translations (I'm unsure `{Title} version {number}` is grammatically easily translatable in all the languages).

### 🖼️ Screenshots

image

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppSettingsDialog/NcAppSettingsDialogVersion.vue
+++ b/src/components/NcAppSettingsDialog/NcAppSettingsDialogVersion.vue
@@ -1,0 +1,33 @@
+<!--
+  - SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<script setup lang="ts">
+import { inject } from 'vue'
+import { APP_VERSION, getLocalizedAppName } from '../../utils/appName.ts'
+import { APP_SETTINGS_LEGACY_DESIGN_KEY } from './useAppSettingsDialog.ts'
+
+const legacy = inject(APP_SETTINGS_LEGACY_DESIGN_KEY)!
+
+const localizedAppName = getLocalizedAppName()
+</script>
+
+<template>
+	<div :class="[$style.appSettingsDialogVersion, { [$style.appSettingsDialogVersion__legacy]: legacy }]">
+		{{ localizedAppName }} {{ APP_VERSION }}
+	</div>
+</template>
+
+<style lang="scss" module>
+.appSettingsDialogVersion {
+    --form-element-label-offset: calc(var(--border-radius-element) + var(--default-grid-baseline));
+    color: var(--color-text-maxcontrast);
+    margin-block-end: calc(8 * var(--default-grid-baseline));
+    margin-inline: var(--form-element-label-offset);
+}
+
+.appSettingsDialogVersion__legacy {
+    margin-inline: 0;
+}
+</style>
--- a/src/utils/appName.ts
+++ b/src/utils/appName.ts
@@ -3,14 +3,30 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
+import { loadState } from '@nextcloud/initial-state'
 import { logger } from './logger.ts'
+import { once } from './utils.ts'
 
 let realAppName = 'missing-app-name'
-
 try {
 	realAppName = appName
 } catch {
 	logger.error('The `@nextcloud/vue` library was used without setting / replacing the `appName`.')
 }
-
 export const APP_NAME = realAppName
+
+let realAppVersion = ''
+try {
+	realAppVersion = appVersion
+} catch {
+	logger.error('The `@nextcloud/vue` library was used without setting / replacing the `appVersion`.')
+}
+export const APP_VERSION = realAppVersion
+
+/**
+ * Get localized app name from the initial state
+ */
+export const getLocalizedAppName = once(() => {
+	const activeApp = loadState<string>('core', 'active-app', APP_NAME)
+	return loadState<{ id: string, name: string }[]>('core', 'apps', []).find(({ id }) => id === activeApp)?.name ?? APP_NAME
+})
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -1,0 +1,22 @@
+/*!
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+/**
+ * Single tone function decorator
+ *
+ * @param func - Function
+ */
+export function once<T extends unknown[], K>(func: (...args: T) => K): ((...args: T) => K) {
+	let wasCalled = false
+	let result: K
+
+	return (...args: T): K => {
+		if (!wasCalled) {
+			wasCalled = true
+			result = func(...args)
+		}
+		return result
+	}
+}
--- a/src/components/NcAppContent/NcAppContent.vue
+++ b/src/components/NcAppContent/NcAppContent.vue
@@ -153,12 +153,11 @@ export default {
 import { getBuilder } from '@nextcloud/browser-storage'
 import { getCapabilities } from '@nextcloud/capabilities'
 import { emit } from '@nextcloud/event-bus'
-import { loadState } from '@nextcloud/initial-state'
 import { useSwipe } from '@vueuse/core'
 import { Pane, Splitpanes } from 'splitpanes'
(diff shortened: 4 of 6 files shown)


src/components/NcAppSettingsDialog/NcAppSettingsDialogVersion.vue
```diff
@@ -0,0 +1,45 @@
...
+import { APP_SETTINGS_LEGACY_DESIGN_KEY } from './useAppSettingsDialog.ts'
+
+const legacy = inject(APP_SETTINGS_LEGACY_DESIGN_KEY)!
+
+const localizedAppName = getLocalizedAppName()
+</script>
+
+<template>
+	<div :class="[$style.appSettingsDialogVersion, { [$style.appSettingsDialogVersion__legacy]: legacy }]">
+		{{ localizedAppName }} version {{ APP_VERSION }}
```
Antreesy: Image

Since app name is localised, I'd make the whole string translated. At least that's what I see in some other apps

src/components/NcAppSettingsDialog/NcAppSettingsDialogVersion.vue
```diff
@@ -0,0 +1,45 @@
...
+<script lang="ts">
+import { loadState } from '@nextcloud/initial-state'
+import { APP_NAME } from '../../utils/appName.ts'
+import { once } from '../../utils/utils.ts'
+
+const getLocalizedAppName = once(() => {
+	const activeApp = loadState<string>('core', 'active-app', APP_NAME)
+	return loadState<{ id: string, name: string }[]>('core', 'apps', []).find(({ id }) => id === activeApp)?.name ?? APP_NAME
+})
+</script>
```
Antreesy: Can it be completely moved in utils? IMO it makes understanding of `once` usage complicated, if it's a part of component.

ShGKme: > it makes understanding of `once` usage complicated

Do you mean the `once` pattern in general, or the `<script>` block?

src/components/NcAppSettingsDialog/NcAppSettingsDialogVersion.vue
ShGKme: I was sure, we do not really translate the application name 👀

src/components/NcAppSettingsDialog/NcAppSettingsDialogVersion.vue
Antreesy: I mean the second script block.
I see `once` can be reused and thus make sense to keep it separate js file in utils, but can tell the same about localized name.

src/components/NcAppSettingsDialog/NcAppSettingsDialogVersion.vue
ShGKme: `{appName} version {version}` just feels weird to me... Also dramatically because it is not a complete sentence to me.

What about just `{appName} {version}` (no translation needed).

@nextcloud-libraries/designers May I ask for your opinion here?

src/components/NcAppSettingsDialog/NcAppSettingsDialogVersion.vue
ShGKme: Pushed fixup

src/components/NcAppSettingsDialog/NcAppSettingsDialogVersion.vue
susnux: I would go with `{appName} {version}`

ShGKme: Push `{appName} {version}`

kra-mo:
> ⚠️ I've removed the "version" part to remove a requirement to have a new translated string and simplify it for translations (I'm unsure {Title} version {number} is grammatically easily translatable in all the languages).

Makes sense.

ShGKme:
> 1. In the sidebar (and at the bottom as it is now when there is only one section and no sidebar is visible).

I'd put it to the navigation, yes.

src/utils/utils.ts
```diff
@@ -0,0 +1,28 @@
...
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+export function once<F extends () => any>(func: F): F
+export function once<F extends (...args: any[]) => void>(func: F): F
+/**
+ * Single tone function decorator
+ *
+ * @param func - Function
+ */
+export function once<F extends (() => any) | ((...args: any[]) => void)>(func: F): F {
```
susnux: does `unknown` not work here?

ShGKme: For which part?

src/utils/appName.ts
```diff
@@ -3,14 +3,30 @@
...
+}
+export const APP_VERSION = realAppVersion
+
+/**
+ * Get localized app name from the initial state
+ */
+export const getLocalizedAppName = once(() => {
+	const activeApp = loadState<string>('core', 'active-app', APP_NAME)
+	return loadState<{ id: string, name: string }[]>('core', 'apps', []).find(({ id }) => id === activeApp)?.name ?? APP_NAME
+})
```
susnux, line 32: Do you think we need the `once` util for something else?
Otherwise we could also just inline it like


```suggestion
let localizedAppName: string

/**
 * Get localized app name from the initial state
 */
export function getLocalizedAppName(): string {
	if (!localizedAppName) {
		const activeApp = loadState<string>('core', 'active-app', APP_NAME)
		localizedAppName = loadState<{ id: string, name: string }[]>('core', 'apps', []).find(({ id }) => id === activeApp)?.name ?? APP_NAME
	}
	return localizedAppName
})
```

src/utils/utils.ts
susnux: Ah no I reversed the mapping of the type, its fine :)

src/utils/appName.ts
ShGKme: Yes, I do:
1. Decomposition to simplify it, saying "I'm applying the `once` decorator pattern for getting the name", instead of implementing it again in many places inside the function
2. We have many places where we either implemented the same, or we miss using `once` (probably because we didn't want to complicate it)
3. It is a classic JS util, like debounce/throttle

src/utils/utils.ts
susnux: Ah no its possible to properly type this (without using `any` ):

```suggestion
export once<ArgumentsType extends unknown[], ReturnType>(func: (...ArgumentsType) => ReturnType): ((...ArgumentsType) => ReturnType)
```

src/utils/appName.ts
susnux: Ok, just wanted to ensure we not adding things for one time use the make the code base more complex then they help with decomposing. (e.g. if this would stay the only use case then this util is much more complex then just inline)

But if we plan to reuse it more then its fine and a nice refactoring! :)

src/utils/utils.ts
ShGKme:
```
Parameter has a name but no type. Did you mean 'arg0: ArgumentsType[]'?
```

Could you share a typescriptlang play with the proposal and test cases?

Is there any problem with `any` here? `any` is a normal and a valid type for a library code, it has different purpose from `unknown`.

susnux: https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABAiBTAPAQQE4HMQC2qYUAzgCoCeADqoqgB5TEAmpi4A1mHAO5gBtALoAaRACVUUENjBVaAPgAUoSAC5ESgHQ6AhnlIac+IiQo1UASkQBeBRKky5FyxqXa9Bo3kLEy8q1t7SWlZAOsAbwAoRFjEABspRF5dUgBhXXjElltEYEzSVCiASESoRGxUUhB4qA0QpwCoksrQpHcdLX1cQ0RjXzNw+scwiyDEaOLimGBNAEIU9MzsyJKpxYys1BybRChsECKp4srq2tzVCA8ug0s1gF811qcKqpqoEvu7x6jL2AQ9lUoEoYGBqCA6ohSPtQbhXIgwIQAEaobATGJxZ6yRAAOWRqK01H0hQAYvE4LpgaDwVBvs0IAhoYgkbkUKglMxoXcogywEyILkkUoAOQADi0AEYJcLLEA

ShGKme: Thanks, pushed a fixup, but used different generic param names to void conflicting with a built-in type util

susnux (approved):

ShGKme:
- Rebased and squashed

PR #7838, opened 2025-11-10
[DRAFT] feat: add NcFormBoxSelectNative

ShGKme:
## ☑️ Resolves

- Ref: https://github.com/nextcloud/server/issues/55667
- The simplest version of select -- Native Select
- Note: `.showPicker()` is not supported on Safari https://caniuse.com/wf-show-picker-select
  - Instead, rendering Select on top with opacity: 0;
- Alternatives:
  - Full `vue-select` (I'd avoid it, it is complicated, heavy and too much)
  - Custom select based on `NcActions` using `NcFormBoxButton` as the trigger

## 🖼️ Screenshots

image

### Chromium - Windows

image

### Firefox Windows

image

### Chromium macOS

image

### Safari macOS

image

## 🏁 Checklist

### DIFF
--- a/src/components/NcFormBoxSelectNative/NcFormBoxSelectNative.vue
+++ b/src/components/NcFormBoxSelectNative/NcFormBoxSelectNative.vue
@@ -4,9 +4,11 @@
   -->
 
 <script setup lang="ts" generic="T extends string">
-import { computed, useId, useTemplateRef } from 'vue'
-import IconUnfoldMoreHorizontal from 'vue-material-design-icons/UnfoldMoreHorizontal.vue'
+import { mdiUnfoldMoreHorizontal } from '@mdi/js'
+import { computed, useTemplateRef } from 'vue'
 import NcFormBoxItem from '../NcFormBox/NcFormBoxItem.vue'
+import NcIconSvgWrapper from '../NcIconSvgWrapper/NcIconSvgWrapper.vue'
+import { createElementId } from '../../utils/createElementId.ts'
 
 /** Selected value */
 const modelValue = defineModel<T>({ required: true })
@@ -27,9 +29,29 @@ const {
 	disabled?: boolean
 }>()
 
-const selectId = useId()
+const selectId = createElementId()
 const selectElement = useTemplateRef('select')
+
 const selectedLabel = computed(() => options.find((option) => option.value === modelValue.value)?.label)
+
+// .showPicker() is not available some browsers (e.g. Safari)
+// When the method is not available, we keep select overlay clickable not invisible
+// When the method is available, hidden select is not directly clickable but opens programmatically
+// The last approach looks slightly better without focusing select by click
+const isShowPickerAvailable = 'showPicker' in HTMLSelectElement.prototype
+
+/**
+ * Handle label click to open the native select picker if possible.
+ *
+ * @param event - Click event
+ */
+function onLabelClick(event: MouseEvent) {
+	if (!isShowPickerAvailable) {
+		return
+	}
+	event?.preventDefault()
+	selectElement.value!.showPicker()
+}
 </script>
 
 <template>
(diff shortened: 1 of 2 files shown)


kra-mo (approved):
Really nice, thank you :)

src/components/NcFormBoxSelectNative/NcFormBoxSelectNative.vue
```diff
@@ -0,0 +1,135 @@
...
+		</template>
+	</NcFormBoxItem>
+</template>
+
+<style lang="scss" module>
+.hiddenSelect {
+	position: absolute;
+	inset: 0;
+	margin: 0;
+	height: auto;
```
kra-mo, line 90: The only thing is that this makes it so that on macOS, the currently selected item's label is aligned to the top of the component, but in reality, the label is on the bottom.

Not sure if there is an easy fix for this though.

ShGKme: Could you share a screenshot? I'm not sure I get what you mean.

kra-mo: Your screenshot in the issue description also exhibits this issue. Basically, `select` on macOS (and to an extent iOS) moves the list up/down depending on the selected item to align the currently selected list item with the component.

Essentially, the label in the list and the label in the component shouldn't move when opening the list and should be perfectly (or at least closely) aligned.

Example of how it should look, seeing "Goldfish" only once:

image

whereas for us, comparatively, it is shifted up and you see "Never" twice:

image

ShGKme: > Example of how it should look, seeing "Goldfish" only once

But in the example the select is small, and our button is 40px.

If the entire button just were a select on the exact same place, it would look exactly the same.

ShGKme: It is not shifted relative to macOS's select's picker position

kra-mo: I think it's better if I explain it this way: since the component is taller than list items, right now the list is basically aligned as if there was a less tall invisible select that was top-aligned.

But since our label is at the bottom, it should act like there is a less tall invisible select that is bottom-aligned instead.

kra-mo: > If the entire button just were a select on the exact same place, it would look exactly the same.

And yes, but it should basically be as if only the bottom half were a select, since that is where our label is, not the top half. Not sure if this is easily doable though.

(Maybe if the text in the invisible select could be also bottom-aligned? But maybe it's not smart enough to pick up on that even if it is possible.)

susnux:
Is this the new way to go with components?
Asking because currently we try to use always `NcSelect` or its children for consistency and this will add different looking experience.

Especially because we did have special components for special sections before, but they got unmaintained and caused weird design differences between apps (when to use what and why does it look like this in one place but directly next to it it looks different). So we unified the components to generic components.

Here for example I do not really understand why a select in app settings or other form like locations should look and behave very different from e.g. the main app where the NcSelect would be used.

Also in the past it was the goal to have a consistent webui experience independent from the browser or OS, but with such native components we will loose the Nextcloud consistent design and get different look if you e.g. just switch devices.

kra-mo:
> Also in the past it was the goal to have a consistent webui experience independent from the browser or OS, but with such native components we will loose the Nextcloud consistent design and get different look if you e.g. just switch devices.

I don't think that this is really an issue. Many websites use the native select component even when they don't have browser-specific adjustments. I see this as the same as using the native font or not overriding the native context menu, which basically has the same styling as select.

kra-mo:
> Here for example I do not really understand why a select in app settings or other form like locations should look and behave very different from e.g. the main app where the NcSelect would be used.

I do understand what you mean, but to me, there is a pretty clear semantic difference that I can see:

- `NcFormBoxSelectNative` would be for small, fixed lists. Simple choices.
- `NcSelect` would then only act as more of an aid to input rather than a dropdown, for more complex cases. Same as the date picker for example, where you can type in the date, but there is also a UI to help you visualize it.

And I don't think that it being form-specific should be an issue. I can't think of many non-form places where selection is needed, including in the main app.

Maintenance is of course another concern, and not my area.

nimishavijay:
Picking this up again as we need it for Mail account settings! @susnux I've got to agree that the NcSelect is quite elaborate, and most of the use cases for this component in settings are rather simple (maximum 5-6 options, no search required in pretty much all cases). If we do want to go ahead with using the NcSelect component itself, I would at least try to make sure that it's presented in a seamless way in the FormBox, so no outline or white background, what do you think? also looping in @kra-mo and @ShGKme again for input :)

kra-mo:
I had ideas about this that I wanted to present (that I talked about before), my day is packed with meetings and I'll need to make other mockups for today, but I want to also write that up when I have the time maybe on Thursday.

susnux:
After some work in other apps it maybe really makes sense to have a NcSelectNative for simple selects.
I had to write my own wrapper over standard `<select>` quite some often lately.
So maybe we should create a simple wrapper which provides some common props (design should already be enforced by server styles) like consistent label layout.
And the wrap it also up for FormBox compatibility (while my use cases were mostly generic selects outside of FormBox)

kra-mo:
I'd like to hear your opinions about #8436 :)

PR #7841, opened 2025-11-11
feat(NcRichText): add 'Copy to clipboard' button for code blocks

Antreesy:
### ☑️ Resolves

- Attempt to upstream clipboard functionality from Talk
- Logic based on NcFormBoxCopyButton
- Works with plain text and highlighted syntax

### 🖼️ Screenshots

🏡 After
image

### 🚧 Tasks

- [x] Move `useCopy` to composables?
- [x] Extract CopyButton as standalone component?
- [ ] Need a prop to enable/disable?

### 🏁 Checklist

### DIFF
--- a/src/components/NcRichText/NcRichText.vue
+++ b/src/components/NcRichText/NcRichText.vue
@@ -314,6 +314,7 @@ import { Fragment, h, ref, resolveComponent } from 'vue'
 import { RouterLink } from 'vue-router'
 import NcCheckboxRadioSwitch from '../NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue'
 import NcReferenceList from './NcReferenceList.vue'
+import NcRichTextCopyButton from './NcRichTextCopyButton.vue'
 import { createElementId } from '../../utils/createElementId.ts'
 import { getRoute, parseUrl, remarkAutolink } from './autolink.ts'
 import { remarkPlaceholder } from './remarkPlaceholder.ts'
@@ -561,8 +562,10 @@ export default {
 					}
 
 					if (String(type) === 'pre' && children && String(children.type) === 'code') {
+						const id = this.parentId + '-code-block-' + createElementId()
 						return h('p', { class: 'rich-text__code-block' }, [
-							h(type, props, children),
+							h(type, { ...props, id }, children),
+							h(NcRichTextCopyButton, { class: 'rich-text__code-block-button', contentId: id }),
 						])
 					}
 
@@ -799,6 +802,19 @@ a:not(.rich-text--component) {
 	& pre {
 		width: 100%;
 		overflow-x: auto;
+	}
+
+	.rich-text__code-block-button {
+		position: absolute;
+		top: var(--default-grid-baseline);
+		inset-inline-end: var(--default-grid-baseline);
+		opacity: 0;
+	}
+
+	&:hover .rich-text__code-block-button,
+	&:focus-within .rich-text__code-block-button,
+	& .rich-text__code-block-button:focus {
+		opacity: 1;
 	}
 }
 </style>
--- a/src/components/NcRichText/NcRichTextCopyButton.vue
+++ b/src/components/NcRichText/NcRichTextCopyButton.vue
@@ -1,0 +1,34 @@
+<!--
+  - SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<script setup lang="ts">
+import NcButton from '../NcButton/NcButton.vue'
+import NcIconSvgWrapper from '../NcIconSvgWrapper/NcIconSvgWrapper.vue'
+import { useCopy } from '../../composables/useCopy.ts'
+
+const { contentId } = defineProps<{
+	/** Node representing code block content */
+	contentId: string
+}>()
+
+const { copy, icon, altText } = useCopy(() => document.getElementById(contentId)!.textContent)
+</script>
+
+<template>
+	<NcButton
+		variant="tertiary"
+		size="small"
+		:aria-label="altText"
+		:title="altText"
+		@click="copy">
+		<template #icon>
+			<NcIconSvgWrapper :path="icon" inline />
+		</template>
+	</NcButton>
+</template>
+
+<docs>
+An internal component
+</docs>


src/components/NcRichText/NcRichText.vue
```diff
@@ -559,6 +564,55 @@ export default {
...
+									title: title.value,
+									onClick: copy,
+								}, {
+									icon: () => h(NcIconSvgWrapper, {
+										path: icon.value,
+										inline: true,
+									}),
+								})
+							}
+						})
```
ShGKme: It creates a new component type for every node rendered. Let's split it into a real component (or at least define it outside the rendering in a module scope).

src/components/NcRichText/NcRichText.vue
```diff
@@ -559,6 +564,55 @@ export default {
...
+
+						return h(type, {
+							...props,
+							class: [props.class, 'rich-text--code'],
+						}, [
+							children,
+							h(CopyButton, {}, []),
```
ShGKme:
```suggestion
							h(CopyButton),
```

src/components/NcRichText/NcRichText.vue
```diff
@@ -559,6 +564,55 @@ export default {
...
+								return node.map(extractRawText).join('')
+							}
+							if (node?.children) {
+								return extractRawText(node.children)
+							}
+							return ''
+						}
+
+						const CopyButton = defineComponent(() => {
+							const { isCopied, copy } = useCopy(() => extractRawText(children.children))
```
ShGKme: We can also use a template ref or a query selector to get `textContent`.

src/components/NcRichText/NcRichText.vue
```diff
@@ -784,4 +838,22 @@ a:not(.rich-text--component) {
...
+
+.rich-text--code {
+	position: relative;
+
+	.rich-text--code-copy-button {
+		position: absolute;
+		top: var(--default-grid-baseline);
+		inset-inline-end: var(--default-grid-baseline);
+		display: none;
+	}
```
ShGKme: We need to reserve a space for the button, otherwise it may overlap the code.
Either by adding `padding-inline-end`, or by using `float`.

src/components/NcRichText/NcRichText.vue
```diff
@@ -559,6 +564,55 @@ export default {
...
+							if (node?.children) {
+								return extractRawText(node.children)
+							}
+							return ''
+						}
+
+						const CopyButton = defineComponent(() => {
+							const { isCopied, copy } = useCopy(() => extractRawText(children.children))
+							const icon = computed(() => isCopied.value ? mdiCheck : mdiContentCopy)
+							const title = computed(() => isCopied.value ? t('Copied') : t('Copy to clipboard'))
```
ShGKme: Makes sense to add it to the composable, as it is the same code we have in another place.
We can also move the composable to `src/composables/.` as it is now shared and not internals of a single component.

src/components/NcRichText/NcRichText.vue
```diff
@@ -784,4 +838,22 @@ a:not(.rich-text--component) {
...
+	.rich-text--code-copy-button {
+		position: absolute;
+		top: var(--default-grid-baseline);
+		inset-inline-end: var(--default-grid-baseline);
+		display: none;
+	}
+
+	&:hover .rich-text--code-copy-button {
+		display: flex;
+	}
```
ShGKme: If it is hidden by `display: none` and visible by `:hover`, it is not keyboard accessible.

src/components/NcRichText/NcRichText.vue
Antreesy: Yeah, might also contribute to memory leak...
Replaces with `document.getElementById`

src/components/NcRichText/NcRichText.vue
Antreesy: Extracted to module scope

src/components/NcRichText/NcRichText.vue
Antreesy: Added a wrapper around `pre`, absolute styles work only that way

src/components/NcRichText/NcRichText.vue
```diff
@@ -336,6 +339,37 @@ async function importRehypeHighlightLibrary() {
...
+}, {
+	props: {
+		/** Node representing code block content */
+		contentId: {
+			type: String,
+			required: true,
+		},
+	},
+})
```
ShGKme: **nitpik:** since it is anyway a full component description now (with props), I'd keep a classic component definition. Functional components don't have any advantages anymore in Vue 3.

```suggestion
const CopyCodeBlockButton = defineComponent({
	name: 'CopyCodeBlockButton',

	props: {
		/** Node representing code block content */
		contentId: {
			type: String,
			required: true,
		},
	},

	setup(props) {
		const { copy, icon, altText } = useCopy(() => document.getElementById(props.contentId).textContent)

		return () => {
			return h(NcButton, {
				class: 'rich-text--code-copy-button',
				variant: 'tertiary',
				size: 'small',
				ariaLabel: altText.value,
				title: altText.value,
				onClick: copy,
			}, {
				icon: () => h(NcIconSvgWrapper, {
					path: icon.value,
					inline: true,
				}),
			})
		}
	},
})
```

ShGKme: **nitpik:** Or we can move it to `.vue` file next to the component to have some micro optimizations from compile-time informed VDOM and Vapor in the future.

src/components/NcRichText/NcRichText.vue
```diff
@@ -784,4 +827,26 @@ a:not(.rich-text--component) {
...
+	& pre {
+		width: 100%;
+		overflow-x: auto;
+	}
+
+	.rich-text--code-copy-button {
+		position: absolute;
+		top: var(--default-grid-baseline);
+		inset-inline-end: var(--default-grid-baseline);
+		visibility: hidden;
```
ShGKme: `visibility: hidden;` also makes it inaccessible via keyboard navigation :(
Could you try `opacity: 0`?

Though, I'd also test in Safari...

Antreesy: ~~`tabindex="0"` seems to make it work in Safari~~

Reverted, as should be explicitly set in Safari

src/components/NcRichText/NcRichTextCopyButton.vue
```diff
@@ -0,0 +1,35 @@
...
+}>()
+
+const { copy, icon, altText } = useCopy(() => document.getElementById(contentId)!.textContent)
+</script>
+
+<template>
+	<NcButton
+		variant="tertiary"
+		size="small"
+		:tabindex="0"
```
ShGKme: Buttons have OS defined tab index by default: focusable via keyboard navigation on Windows/Linux and managed by "Enable keyboard navigation" on macOS. So no tabindex needed to respect OS default behavior.

src/components/NcRichText/NcRichText.vue
```diff
@@ -559,6 +560,15 @@ export default {
 						&& props?.class?.includes('language')) {
 						importRehypeHighlightLibrary()
 					}
+
+					if (String(type) === 'pre' && children && String(children.type) === 'code') {
+						const id = this.parentId + '-code-block-' + createElementId()
+						return h('p', { class: 'rich-text--code-block' }, [
+							h(type, { ...props, id }, children),
+							h(NcRichTextCopyButton, { class: 'rich-text--code-copy-button', contentId: id }),
```
susnux: Should it not be `rich-text_code-copy-button` for proper BEM?

Antreesy: Changed to `rich-text__code-block` and `rich-text__code-block-button`

PR #7847, opened 2025-11-12
fix(NcActions): introduce 'size' prop

Antreesy:
### ☑️ Resolves

- Allows to pass 'size' prop to underlying NcButton components
- Introducing beforehand, if that would be ever required by apps

### 🖼️ Screenshots

image

### 🏁 Checklist

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -1081,7 +1081,6 @@ export default {
 		/**
 		 * Specifies the button variant used for trigger and single actions buttons.
 		 *
-		 * Accepted values: primary, secondary, tertiary, tertiary-no-background, tertiary-on-primary, error, warning, success.
 		 * If left empty, the default button style will be applied.
 		 *
 		 * @since 8.23.0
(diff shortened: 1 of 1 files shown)


src/components/NcActions/NcActions.vue
```diff
@@ -1094,6 +1094,18 @@ export default {
 
 			default: null,
 		},
+
+		/**
+		 * Specify the button size
+		 * Accepted values: `'small'`, `'normal'` (default), `'large'`
+		 */
```
ShGKme, line 1101: Styleguidist parses `includes` to see values. And in TS we'll specify it explicitly. So we don't need to describe it in the documentation.

image


And I'd mention that this is specifically inline buttons and the trigger size, not "entire NcActions menu size".


```suggestion
		/**
		 * Trigger and inline button sizes
		 */
```

Antreesy: Updated
image

ShGKme: What about the second part? :)

Antreesy: Overlooked

PR #7854, opened 2025-11-13
fix(NcFormBox): use default text color in items

ShGKme:
### ☑️ Resolves

- Fix https://github.com/nextcloud-libraries/nextcloud-vue/issues/7853

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image
image | image
image | image
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/NcFormBox/NcFormBoxItem.vue
+++ b/src/components/NcFormBox/NcFormBoxItem.vue
@@ -116,7 +116,7 @@ const hasDescription = () => !!description || !!slots.description
 	border-bottom-width: 2px;
 	border-radius: var(--border-radius-element);
 	background-color: var(--color-primary-element-extra-light);
-	color: var(--color-primary-element-light-text);
+	color: var(--color-main-text);
 	transition-property: color, border-color, background-color;
 	transition-duration: var(--animation-quick);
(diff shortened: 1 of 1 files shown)


susnux:
Do we also want to recolor the background? To the normal `--color-background-dark`?
Seeing this in files app settings it looks like a rainbow :sweat_smile: (so many colors that it distracts a bit from important parts).

ShGKme:
> like a rainbow

Exactly what I like in Spark Mail client settings dialog.

image

> Do we also want to recolor the background? To the normal --color-background-dark?

@susnux I'm implementing what designers design 😶
Please, create an issue or ask in the corresponding proposal or the Design chat 👀

kra-mo:
> Do we also want to recolor the background? To the normal --color-background-dark?

Indeed, it looks odd. But we already tint many parts of the UI so I think the long term solution would be the other way around: tint background colors with the theme color, similar to Material 3 (but also macOS for example).

We talked about this yesterday (for the same reason, that the forms stand out too much) and I think Design would want to go ahead with this if possible. We haven't figured out everything yet, but just letting you know. Do you know whether this would at all be feasible?

PR #7855, opened 2025-11-13
fix(NcFormBox): use `text-wrap: pretty` in items

### LINKED ISSUE #7842: [NcFormBox] Use `text-wrap: pretty` for FormBox* components where supported
This is what it looks like with default wrapping:

Image

And `pretty`:

Image

Support is [low-ish](https://caniuse.com/wf-text-wrap-pretty) right now, but it's literally just to make something more pretty so probably not an issue :)

ShGKme:
### ☑️ Resolves

- Fix https://github.com/nextcloud-libraries/nextcloud-vue/issues/7842
- Thanks @kra-mo for a performance-related research
- Keeping it as draft until know if other form items should have it as well

### 🏁 Checklist

### DIFF
--- a/src/components/NcFormBox/NcFormBoxItem.vue
+++ b/src/components/NcFormBox/NcFormBoxItem.vue
@@ -117,6 +117,7 @@ const hasDescription = () => !!description || !!slots.description
 	border-radius: var(--border-radius-element);
 	background-color: var(--color-primary-element-extra-light);
 	color: var(--color-primary-element-light-text);
+	text-wrap: pretty;
 	transition-property: color, border-color, background-color;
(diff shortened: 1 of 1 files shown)


kra-mo (changes requested):
Copy-pasting from the issue:

> > [@kra-mo](https://github.com/kra-mo) Only the interactive items, or anything else as well?
> >
> > Such us the groups headings/description, shortcuts labels etc.
>
> No strong opinions. I found the wrapping odd for interactive elements only because of the trailing icon/switch.
>
> At least for shortcuts labels, that would also apply because of the keys on the right side. For other things, it's your call.

PR #7862, opened 2025-11-14
docs: Move more icons to outline

nickvergessen:
### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Bildschirmfoto vom 2025-11-14 14-23-54 | Bildschirmfoto vom 2025-11-14 14-23-45
Bildschirmfoto vom 2025-11-14 14-25-04 | Bildschirmfoto vom 2025-11-14 14-25-48
Bildschirmfoto vom 2025-11-14 14-29-36 | Bildschirmfoto vom 2025-11-14 14-29-26
Bildschirmfoto vom 2025-11-14 14-31-00 | Bildschirmfoto vom 2025-11-14 14-31-49
Bildschirmfoto vom 2025-11-14 14-32-22 | Bildschirmfoto vom 2025-11-14 14-32-54
Bildschirmfoto vom 2025-11-14 14-34-38 | Bildschirmfoto vom 2025-11-14 14-37-02
Bildschirmfoto vom 2025-11-14 14-38-25 | Bildschirmfoto vom 2025-11-14 14-38-33
Bildschirmfoto vom 2025-11-14 14-40-13 | Bildschirmfoto vom 2025-11-14 14-40-25
Bildschirmfoto vom 2025-11-14 14-45-15 | Bildschirmfoto vom 2025-11-14 14-47-45
Bildschirmfoto vom 2025-11-14 14-48-30 | Bildschirmfoto vom 2025-11-14 14-49-57

### DIFF
--- a/src/components/NcAppSidebar/NcAppSidebar.vue
+++ b/src/components/NcAppSidebar/NcAppSidebar.vue
@@ -36,19 +36,19 @@ Using `v-show` directly will result in usability issues due to internal focus tr
 			@figure-click="figureClick">
 			<NcAppSidebarTab name="Search" id="search-tab">
 				<template #icon>
-					<Magnify :size="20" />
+					<IconMagnify :size="20" />
 				</template>
 				Search tab content
 			</NcAppSidebarTab>
 			<NcAppSidebarTab name="Settings" id="settings-tab">
 				<template #icon>
-					<Cog :size="20" />
+					<IconCogOutline :size="20" />
 				</template>
 				Settings tab content
 			</NcAppSidebarTab>
 			<NcAppSidebarTab name="Sharing" id="share-tab">
 				<template #icon>
-					<ShareVariant :size="20" />
+					<IconShareVariantOutline :size="20" />
 				</template>
 				Sharing tab content
 			</NcAppSidebarTab>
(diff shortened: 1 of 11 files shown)


ShGKme:
Arr, double-clicking mouse

susnux:
> /backport to stable8

what is the current position on backporting this even though this will change how legacy NC versions will look like in next maintenance releases?

ShGKme:
> what is the current position on backporting this even though this will change how legacy NC versions will look like in next maintenance releases?

I didn't get what you mean. It's documentation, what does it change in Nextcloud versions?

ShGKme:
`@nextcloud/vue v9` is also used in Nextcloud versions with the filled icons.

susnux:
> @nextcloud/vue v9 is also used in Nextcloud versions with the filled icons.

Well but in general v9 is mostly used by newer Nextcloud versions (is there an official app that is Vue 3 on Nextcloud 31?).
But what I mean is that v8 will change Nextcloud core down to basically every version if there is a problem we need to fix and always down to v31.

Basically my concerns are not about apps but Nextcloud core.

ShGKme:
@susnux How are these concerns related to this PR with the documentation update?

Or they are unrelated to the PR and in general about icon design update?

susnux:
> Or they are unrelated to the PR and in general about icon design update?

Unrelated to this PR its more about:
Why do we handle legacy design like in buttons but the most visual change for users are the icons that now would change for stable branches. And a more follow up question to: Should be also have icon legacy handling.

Not necessary in examples but in the real components.

ShGKme:
> Why do we handle legacy design like in buttons but the most visual change for users are the icons that now would change for stable branches. And a more follow up question to: Should be also have icon legacy handling.

So you wanted to have `isLegacy` switch for the icons in components like for the classes/styles, correct? Or is it a hint at the corresponding RFC in standards? 😶

IMO, it is much a less visual change than others:
- It is not a layout change like going from 44px to 34px
- It is not a single thing used everywhere like 1px vs 2px border on EVERY interactive element
- In the past we often had a mix, so even solid icons were not really consistent

susnux:
> So you wanted to have isLegacy switch for the icons in components like for the classes/styles, correct? Or is it a hint at the corresponding RFC in standards? 😶

Both ;)
Bur for the first its more a question if we can live with inconsistent Nextcloud UIs in stable versions.

PR #7875, opened 2025-11-17
chore: refactor deprecated :deep selector

ShGKme:
### ☑️ Resolves

```
[@vue/compiler-sfc] :deep usage as a combinator has been deprecated. Use :deep(<inner-selector>) instead of :deep <inner-selector>.
```

### 🏁 Checklist

### DIFF
--- a/src/components/NcRelatedResourcesPanel/NcResource.vue
+++ b/src/components/NcRelatedResourcesPanel/NcResource.vue
@@ -81,14 +81,12 @@ export default {
 		justify-content: flex-start !important;
 		padding: 0 !important;
 
-		&:deep {
-			.button-vue__wrapper {
-				justify-content: flex-start !important;
+		& :deep(.button-vue__wrapper) {
+			justify-content: flex-start !important;
 
-				.button-vue__text {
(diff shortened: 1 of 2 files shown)


src/components/NcRelatedResourcesPanel/NcResource.vue
```diff
@@ -81,14 +81,12 @@ export default {
...
 
-		&:deep {
-			.button-vue__wrapper {
-				justify-content: flex-start !important;
-
-				.button-vue__text {
-					font-weight: normal !important;
-					margin-inline-start: 2px !important;
-				}
+		&:deep(.button-vue__wrapper) {
```
Antreesy: This should be nested, if I'm not wrong?

```suggestion
		& :deep(.button-vue__wrapper) {
```

ShGKme: `:deep` is always nesting, but we can add a space to make it more obvious

ShGKme: Done

PR #7882, opened 2025-11-18
feat(NcUploadPicker): migrate component from `@nextcloud/upload`

### LINKED ISSUE #1809: [RFC] export of vue components
Currently this library basically consists of two parts:
1. The API library (Uploader and Upload class) which provides framework agnostic functions to upload files to Nextcloud
2. The Vue UploadPicker component which also includes **files**-app specific handling of the "new"-menu

This has the problem that the package is bound to that Vue version used for the upload picker.
Also the upload picker itself is very bound to the files app and only party reusable by other apps.

While you can disable the menu function in the upload picker to use it in your app without files specific functions, it does not make much sense to me that we share this in a library.
For me it would make more sense to implement those files specific functions directly in the files app.

So to make this library  more lightweight and easier to maintain across Vue versions and other dependencies I was thinking about those some options:

We could remove the Vue related things all together, making this library a API only library similar to `@nextcloud/dialogs`. We still could provide the conflict dialog similar to the file picker as API only and without the Vue export. This way this library could use Vue 3 even in Vue 2 apps and we only need to maintain one version.

We could also move the upload picker to something like `NcFilePicker` in `@nextcloud/vue` making it pretty generic and reuse this one then in the files app by extending its menu with the "new"-menu entries.

This also has the benefit that the library becomes much more lightweight regarding its dependencies if you only have a small app or even a non-vue app.

Or we leave everything as is and maintain two different version for some time ;)

---

My personal opinion would be that the cleanest structure would be moving the file conflict dialog to `nextcloud/dialogs` similar to the filepicker. Then move the upload picker as a generic to `nextcloud/vue` and move the "new"-menu handling directly into the files app.
Meaning this package would only provide the API to upload files but not any UI.

What are your opinions? @skjnldsv @artonge and @ShGKme (especially as you are working on the Vue 3 branch).

susnux:
### ☑️ Resolves

- Followup on #7097 
- Resolves https://github.com/nextcloud-libraries/nextcloud-upload/issues/1809

### 🚧 Tasks

- [x] Requires fixes in upload logic first: https://github.com/nextcloud-libraries/nextcloud-files/pull/1722

### 🏁 Checklist


ShGKme:
is something missing?

susnux:
> is something missing?

tests and ts check fails, featurewise this should be done

- [x] https://github.com/nextcloud-libraries/nextcloud-files/pull/1722

src/components/NcUploadPicker/NcUploadPicker.vue
```diff
@@ -0,0 +1,428 @@
...
+	isPaused.value = uploadManager.status === UploaderStatus.PAUSED
+	// While paused the queue is not processed, but the queued uploads are still pending
+	isUploading.value = uploadManager.status === UploaderStatus.UPLOADING
+		|| (isPaused.value && uploadManager.queue.length > 0)
+	hasFailure.value = uploadManager.queue.some((upload: IUpload) => upload.status === UploadStatus.FAILED)
+	isAssembling.value = uploadManager.queue.some((upload: IUpload) => upload.status === UploadStatus.ASSEMBLING)
+	isOnlyAssembling.value = isAssembling.value
+		&& uploadManager.queue.every((upload: IUpload) => (
+			// either a meta upload
+			upload.children.length === 0
```
pringelmann: `children` is empty for file uploads and only populated for folder uploads, right? 

If so, unless I'm missing something, this branch is always true for a file, so `every()` never fails and the first chunked file to reach ASSEMBLING hides the cancel button while the others are still uploading. Did you mean `> 0` here?

src/components/NcUploadPicker/NcUploadPicker.vue
```diff
@@ -0,0 +1,428 @@
...
+}
+
+/**
+ * Handle conflicts during upload
+ *
+ * @param nodes - The nodes that might conflict
+ * @param currentPath - The path of the current directory
+ */
+async function handleConflicts(nodes: string[], currentPath: string): Promise<Record<string, string> | false> {
+	const content = await props.content(currentPath)
```
pringelmann: Should this be wrapped in a catch? If it rejects it looks like we leave the queue non-empty

tests/component/components/NcUploadPicker/conflicts.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+		await pickFiles(page, createFile('file.txt', 1))
+		await conflictDialog(page).getByRole('button', { name: 'Cancel' }).click()
+
+		await expect(page.getByRole('progressbar')).toBeHidden()
+		expect(dav.received('PUT')).toHaveLength(0)
+	})
+
+	test('keeping the existing version skips the conflicting file', async ({ mount, page }) => {
+		// FIXME: Skipped files are currently uploaded anyway: https://github.com/nextcloud-libraries/nextcloud-files/pull/1722
+		test.fail()
```
pringelmann: Can this be removed now that the PR was merged?

package.json
```diff
@@ -164,6 +164,10 @@
     "webpack": "^5.109.2",
     "webpack-merge": "^6.0.1"
   },
+  "optionalDependencies": {
```
pringelmann: Any reason for `optionalDependencies` over `peerDependencies` ?

susnux: Yes because this is optional for one feature (the upload picker) - so if you do not need the upload picker there is no need to install the full files dependencies.
Peer is just to dedupe with app dependencies but still means a real full dependency of the library.

pringelmann: The `optional` flag in `peerDependenciesMeta` is then probably what we want here. Plain peer deps do get auto-installed, optional ones don't: https://docs.npmjs.com/cli/v11/configuring-npm/package-json#peerdependenciesmeta

optionalDependencies still install by default afaik. The "Optional" there only means npm itself won't fail if the fetch fails: https://docs.npmjs.com/cli/v11/configuring-npm/package-json#optionaldependencies, but it will still install. In this case there may even be two instances if the consumer has `3.x.x` and we're installing `4.beta`

src/components/NcUploadPicker/NcUploadPicker.vue
susnux: makes sense!

src/components/NcUploadPicker/NcUploadPicker.vue
susnux: I have added tests for this and resolved the issue here - see last commit added now.
Thank you!

pringelmann (approved):

package.json
pringelmann: @susnux happy to go with optionalDependencies if you think it makes sense, just wanted to raise that it might cause two instances running at the same time

susnux: > it might cause two instances running at the same time

You mean 2 instances of the library in one app?
Should not be a problem as the library handles this through its own global state wrapper.
But also modern npm should dedupe when versions are compatible I think.

skjnldsv (approved):

skjnldsv (commented):
Nice migration, the story plus mockDav approach reads really well and I think we should standardise on it for the other component tests.

Two things before this can go in.

First, conflicts.spec.ts is failing, on my machine and in CI on both shards. It looks like it was written against a ConflictPicker that does not match what @nextcloud/dialogs actually ships. Details in the inline comments.

Second, the retry behaviour from the old Cypress suite is not covered anywhere after the move. `retryCondition` in @nextcloud/files (`lib/upload/utils/upload.ts`) decides that 507 is permanent and 423 should be retried, and that file has no spec at all, so those branches are untested in every repo. I opened a follow-up PR #8922 based on this branch, which you can merge into this one if you want a single PR, or keep as a second step. It adds the missing retry tests and a few of the cleanups below.

Separately, the invalid filename flow (rename, skip, cancel on a forbidden name) has no successor at all. `filename-validation.ts` in @nextcloud/files unit tests the rules but nothing drives the UI. If we are dropping that on purpose it should be called out as a breaking change, otherwise it needs a new home before @nextcloud/upload v2 removes its Vue components. I can open an issue for it.

tests/component/components/NcUploadPicker/conflicts.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+	})
+})
+
+/**
+ * The conflict dialog shown when uploading a file that already exists.
+ *
+ * @param page - The page of the test
+ */
+function conflictDialog(page: Page): Locator {
+	return page.getByRole('dialog', { name: '1 file conflict' })
```
skjnldsv, line 129: This is what makes the whole file fail. There is no `1 file conflict` string in @nextcloud/dialogs. The dialog name comes from `dialogName` in `ConflictPicker.vue` and is `Select file to keep` for a single conflict and `Select files to keep` for several, so this locator never resolves.

Not fixed in #8922, I left conflicts.spec.ts alone since it is yours to decide how the single conflict case should read.

tests/component/components/NcUploadPicker/conflicts.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+		const dav = await mockDav(page)
+		await mount(NcUploadPickerStory, { props: { existingFiles: ['file.txt'] } })
+
+		await pickFiles(page, createFile('file.txt', 1))
+
+		const dialog = conflictDialog(page)
+		await expect(dialog).toBeVisible()
+		const conflict = dialog.getByRole('group', { name: 'file.txt' })
+		await expect(conflict.getByRole('checkbox', { name: /^New version/ })).toBeVisible()
+		await expect(conflict.getByRole('checkbox', { name: /^Existing version/ })).toBeVisible()
```
skjnldsv, line 25: Single conflicts do not render checkboxes. `ConflictPicker` uses `incoming.length === 1` to switch to the single file layout, which shows `Cancel`, `Keep both` and `Replace` buttons instead. So for a one file conflict this should assert those buttons, and `keepVersion()` on line 139 cannot work at all since it waits for a `New version` checkbox.

Worth mirroring what nextcloud-dialogs does in `tests/components/ConflictPicker.spec.ts`, it covers both layouts already. For the single case `Replace` means keep the new one and `Keep both` means rename. The checkbox flow only applies once there is more than one incoming file, and there the submit button is `Continue`.

Not fixed in #8922 either, for the same reason.

tests/component/components/NcUploadPicker/conflicts.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+		// Nothing is uploaded before the conflict is resolved
+		expect(dav.received('PUT')).toHaveLength(0)
```
skjnldsv, line 27: `received()` only looks at the requests that have arrived so far, so this can pass because the PUT has not been sent yet rather than because it will never be sent. Same on lines 38, 51 and 62.

Line 51 is the weakest one, the dialog check and the progressbar check are both satisfied by nothing having happened, so the test would still pass if conflict handling silently uploaded a moment later.

Two options, either wait for something that proves the uploader got far enough first (line 62 does this nicely with `waitFor('HEAD')`), or add a small helper to `mockDav` that snapshots the count, waits, and asserts it did not grow.

The two tests that already pass are hardened in #8922, using `waitFor('HEAD')` as the sync point plus an `expectNoMore()` helper added to `mockDav`. The four that fail are left to this PR.

tests/component/components/NcUploadPicker/progress.spec.ts
```diff
@@ -0,0 +1,181 @@
...
+		await page.getByRole('button', { name: 'Pause uploads' }).click()
+		chunk.respond()
+
+		// The upload of the file and of the destination folder are still queued
+		await expect.poll(() => events).toEqual(['paused:2'])
+		await expect(page.getByText('paused')).toBeVisible()
+		await expect(progress).toBeVisible()
+		// The remaining chunk is not uploaded and the chunks are not assembled
+		expect(dav.received('PUT')).toHaveLength(1)
+		expect(dav.received('MOVE')).toHaveLength(0)
```
skjnldsv, line 173: Same as in conflicts.spec.ts, these two can pass before the remaining chunk would have been sent. Since the point of the test is that pausing stops further uploads, it is worth making sure nothing arrives after a short wait rather than checking the count once.

Done in #8922 with `expectNoMore('PUT')`.

tests/component/components/NcUploadPicker/NcUploadPicker.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+
+import { expect, test } from '@playwright/experimental-ct-vue'
+import { join } from 'node:path'
+import NcUploadPickerStory from './NcUploadPicker.story.vue'
+import { createFile, mockDav, pickFiles } from './upload-helpers.ts'
+
+/** The upload of a file finished successfully */
+const UPLOAD_FINISHED = 4
+/** The upload of a file failed */
+const UPLOAD_FAILED = 6
```
skjnldsv, line 14: These duplicate `UploadStatus` by hand here and in cancel.spec.ts, and will drift silently if the values ever change. I know it cannot be imported directly, @nextcloud/files/upload touches `window` on import while the test bodies run in Node.

What works is importing it as a type only and using `satisfies`, so the numbers stay literal but the type checker fails if upstream changes them:

```ts
import type { UploadStatus as UpstreamUploadStatus } from '@nextcloud/files/upload'

export const UploadStatus = {
  INITIALIZED: 0,
  SCHEDULED: 1,
  UPLOADING: 2,
  ASSEMBLING: 3,
  FINISHED: 4,
  CANCELLED: 5,
  FAILED: 6,
} as const satisfies typeof UpstreamUploadStatus
```

I checked, changing one of the values gives `Type '99' is not assignable to type '4'`.

Done in #8922, the mirror lives in upload-helpers and cancel.spec.ts uses it too.

tests/component/components/NcUploadPicker/NcUploadPicker.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+	test('has no progress bar and no cancel button while idle', async ({ mount, page }) => {
+		await mount(NcUploadPickerStory)
+
+		await expect(page.getByRole('progressbar')).toBeHidden()
+		await expect(page.getByRole('button', { name: 'Cancel uploads' })).toBeHidden()
+	})
+})
+
+test.describe('NcUploadPicker: uploading', () => {
+	test.skip(({ browserName }) => browserName === 'webkit', 'WebKit does not support file pickers in Playwright yet')
```
skjnldsv, line 52: This skip is repeated in five describes with the same reason string, here and on line 177, and in cancel.spec.ts, conflicts.spec.ts and progress.spec.ts. Could go into upload-helpers as a small function that calls `test.skip`, since `test` is a module singleton it still applies to the surrounding describe.

Done in #8922 as `skipWithoutFilePicker()`.

tests/component/components/NcUploadPicker/upload-helpers.ts
```diff
@@ -0,0 +1,179 @@
...
+	buffer: Buffer
+}
+
+/**
+ * Create a fake file of the given size.
+ *
+ * @param name - Name of the file
+ * @param sizeInMiB - Size of the file in MiB
+ */
+export function createFile(name: string, sizeInMiB: number): FakeFile {
```
skjnldsv, line 24: Small thing, `createFile` always uses `text/plain`. Taking the mime type as an optional third argument would let us exercise the `accept` prop from a component test rather than only in unit tests.

Done in #8922.

tests/component/components/NcUploadPicker/conflicts.spec.ts
susnux: Thats due to the dialogs library update - this PR is far older than the changes done for NC35 in the library.
Need to adjust this PR now

susnux:
> Second, the retry behaviour from the old Cypress suite is not covered anywhere after the move

Thats not part of the UI - the retry is part of the library function and thus we do not need to duplicate those tests here.
This is only for the UI and should only check the UI is working properly to the library interface.
The actual upload logic is tested in the library itself.

Antreesy:
I'm hitting error from manual install, assume it originates from this PR?

```sh
>> npm i @nextcloud/vue@9.12.0
npm error code ERESOLVE
npm error ERESOLVE could not resolve
npm error
npm error While resolving: @nextcloud/vue@9.12.0
npm error Found: @nextcloud/files@4.0.0
npm error node_modules/@nextcloud/files
npm error   @nextcloud/files@"^4.0.0" from @nextcloud/dialogs@7.5.0
npm error   node_modules/@nextcloud/dialogs
npm error     @nextcloud/dialogs@"^7.5.0" from the root project
npm error     peerOptional @nextcloud/dialogs@"^7.4.0" from @nextcloud/vue@9.12.0
npm error   optional @nextcloud/files@"^3.12.2 || ^4.0.0" from @nextcloud/sharing@0.4.0
npm error   node_modules/@nextcloud/sharing
npm error     @nextcloud/sharing@"^0.4.0" from @nextcloud/dialogs@7.5.0
npm error     node_modules/@nextcloud/dialogs
npm error       @nextcloud/dialogs@"^7.5.0" from the root project
npm error       1 more (@nextcloud/vue)
npm error     @nextcloud/sharing@"^0.4.0" from the root project
npm error     1 more (@nextcloud/vue)
npm error   1 more (the root project)
npm error
npm error Could not resolve dependency:
npm error peerOptional @nextcloud/files@"^4.1.0-beta.2" from @nextcloud/vue@9.12.0
npm error node_modules/@nextcloud/vue
npm error   @nextcloud/vue@"9.12.0" from the root project
npm error   @nextcloud/vue@"^9.9.0" from @nextcloud/dialogs@7.5.0
npm error   node_modules/@nextcloud/dialogs
npm error     @nextcloud/dialogs@"^7.5.0" from the root project
npm error     1 more (@nextcloud/vue)
npm error
npm error Conflicting peer dependency: @nextcloud/files@4.1.0-beta.2
npm error node_modules/@nextcloud/files
npm error   peerOptional @nextcloud/files@"^4.1.0-beta.2" from @nextcloud/vue@9.12.0
npm error   node_modules/@nextcloud/vue
npm error     @nextcloud/vue@"9.12.0" from the root project
npm error     @nextcloud/vue@"^9.9.0" from @nextcloud/dialogs@7.5.0
npm error     node_modules/@nextcloud/dialogs
npm error       @nextcloud/dialogs@"^7.5.0" from the root project
npm error       1 more (@nextcloud/vue)
```

susnux:
@Antreesy should be fixed with https://github.com/nextcloud-libraries/nextcloud-vue/pull/8972

PR #7889, opened 2025-11-19
[DRAFT] feat: add NcFormListBox

ShGKme:
### ☑️ Resolves

- Ref: https://github.com/nextcloud/mail/issues/11910

I'm very unsure with the approach in this PR:
1. It indirectly overrides `<NcListItem>` styles...
   - Should it be internally modified in the `NcListItem` based on `isInFormBox` instead?
   - Should it be just a new component like `McFormListBoxItem`?
2. Should it be in any way related to the `NcFormBox`?
   - We can make it optionally `ul` and override styles there instead of a new container

### 🖼️ Screenshots

image

### 🏁 Checklist


kra-mo:
So that I understand: Currently, there is no dedicated "list" component, only the standard `<ul>` tag, correct?

And the only thing this does is add separators between list items, correct?

ShGKme:
> So that I understand: Currently, there is no dedicated "list" component, only the standard `<ul>` tag, correct?

No, this PR adds a new dedicated `NcFormListBox` component as an alternative to `NcFormBox` for lists.

> And the only thing this does is add separators between list items, correct?

No, it also adjusts the padding of the existing `NcListItem` to match the other form element labels.

kra-mo:
Ok, well regarding https://github.com/nextcloud-libraries/nextcloud-vue/issues/7900#issuecomment-3558424085, I'm not sure some of these should even be list items then…

PR #7890, opened 2025-11-19
fix(NcActionCheckbox): margin around the checkbox

GretaD:
### ☑️ Resolves

- Ref https://github.com/nextcloud/mail/issues/11328

### 🖼️ Screenshots

Screenshot from 2025-11-19 16-59-49
Screenshot from 2025-11-19 16-59-43

### 🚧 Tasks

- [ ] Some margin from inputs.scss is applied that creates that small space between the label and the checkbox. This fixes it. I cannot think of any better solution

### 🏁 Checklist

### DIFF
--- a/src/components/NcActionCheckbox/NcActionCheckbox.vue
+++ b/src/components/NcActionCheckbox/NcActionCheckbox.vue
@@ -206,6 +206,7 @@ export default {
 		&::before {
 			margin-block: 0 !important;
 			margin-inline: calc((var(--default-clickable-area) - 14px) / 2) !important;
+			margin: 9px !important;
 		}
 	}
 


src/components/NcActionCheckbox/NcActionCheckbox.vue
```diff
@@ -206,6 +206,7 @@ export default {
 		&::before {
 			margin-block: 0 !important;
 			margin-inline: calc((var(--default-clickable-area) - 14px) / 2) !important;
+			margin: 9px !important;
```
Antreesy, line 209: This simply overriding two lines above. Doesn't it conflict with other usecases for component?

Antreesy (commented):
Is is still actual, or you fixed it already in Mail?

src/components/NcActionCheckbox/NcActionCheckbox.vue
GretaD: it comes from here, so we should fix it here. I understand that 9px is very random, so maybe i can think about how not to have an exact number like that. Its broken everywhere where we use this component, at least everywhere on mail

ShGKme:
Playing with margins to move server-styled icon having different Nextcloud versions seems complicated and unreliable. And the current icon design looks off anyway.

- As an alternative, I'd propose to re-implement the component using existing styles from other actions: https://github.com/nextcloud-libraries/nextcloud-vue/pull/8020

GretaD:
closing and continuing here https://github.com/nextcloud-libraries/nextcloud-vue/pull/8020

PR #7928, opened 2025-11-27
revert: chore(package): npm=11.6.2

ShGKme:
### ☑️ Resolves

- Reverts: https://github.com/nextcloud-libraries/nextcloud-vue/pull/7895/
- Not needed after npm 11.6.4 release

### 🏁 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -170,7 +170,7 @@
     "packageManager": [
       {
         "name": "npm",
-        "version": ">=11.3.0 <=11.6.2",
+        "version": "^11.3.0",
         "onFail": "error"
       }
     ],
--- a/netlify.toml
+++ b/netlify.toml
@@ -3,4 +3,4 @@
 
 [build.environment]
 	NODE_VERSION = "24"
-	NPM_VERSION = "11.6.2"
+	NPM_VERSION = "11"


package.json
```diff
@@ -170,7 +170,7 @@
     "packageManager": [
       {
         "name": "npm",
-        "version": ">=11.3.0 <=11.6.2",
+        "version": "^11.3.0",
```
susnux, line 173: Should it not exclude 11.6.3?

```suggestion
        "version": "^11.6.4",
```

susnux (approved):
otherwise

package.json
ShGKme: All the versions before 11.6.3 work

ShGKme: Why break is a developer has 11.6.2 locally?

susnux: easier annotation and still better than e.g. corepack annotation where you even have to match the packageManager hash :sweat_smile: 

You probably mean
 ```suggestion
         "version": "^11.3.0 <= 11.6.2 || ^11.6.4",
```

But as said fine with this as well when 11.6.3 is not published with next Node version.

ShGKme: No, I meant ^11.3.0, because this package doesn't use anything specific requiring 11.6.4+. 11.3.0 still works. 

And I don't see why we need to manually exclude 11.6.3. Is it a package responsibility to exclude a broken npm version on the local setup?

PR #7931, opened 2025-11-28
fix(l10n): Fix spelling of "time zone"

kra-mo:
This was spelled inconsistently throughout the repo. I believe this one is the correct spelling.

### DIFF
--- a/l10n/messages.pot
+++ b/l10n/messages.pot
@@ -427,7 +427,7 @@ msgstr ""
 msgid "Search emoji"
 msgstr ""
 
-msgid "Search for timezone"
+msgid "Search for time zone"
 msgstr ""
 
 msgid "Search results"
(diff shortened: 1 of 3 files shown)


dartcafe:
> This was spelled inconsistently throughout the repo. I believe this one is the correct spelling.

Depends on the origin (US or brit). But both variants additionally to time-zone seem to be correct.

https://en.wiktionary.org/wiki/timezone

kra-mo:
> > This was spelled inconsistently throughout the repo. I believe this one is the correct spelling.
> 
> Depends on the origin (US or brit). But both variants additionally to time-zone seem to be correct.
> 
> https://en.wiktionary.org/wiki/timezone

Interesting. It should be consistent in any case.

susnux:
> Depends on the origin (US or brit)

Source language is always US English (GB is another translation we offer).

> https://en.wiktionary.org/wiki/timezone

This seems to be the only source, otherwise all dictionaries state "time zone" and for British also "time-zone".
I think its from the programming world where `timezone` was introduced and used for quite some time.

PR #7939, opened 2025-12-01
feat(NcAppNavigationItem): use exact route matching

### LINKED ISSUE #7965: [NcAppNavigationItem] has no option for exact route matching
In v8 it was possible to use `exact - Pass in true if you want the matching behaviour to be non-inclusive: https://router.vuejs.org/api/#exact` for route matching. This was removed in v9.

I created a PR #7939 to bring back this option by using `isExactActive: true if the [exact active class](https://v3.router.vuejs.org/api/#exact-active-class) should be applied. Allows to apply an arbitrary class`.

wofferl:
Resolves: #7965
### updated

This PR changes the state to use from `isActive` to `isExactActive` and made `active` an optional prop to overwrite the default behaviour.

### old
exact - `Pass in true if you want the matching behaviour to be non-inclusive: https://router.vuejs.org/api/#exact`

The [exact](https://stable8--nextcloud-vue-components.netlify.app/#/Components/App%20containers/NcAppContent/NcAppNavigation?id=ncappnavigationitem) prop for NcAppNavigationItem was removed in v9 without replacement.

The news app could now use this for a new feature (https://github.com/nextcloud/news/pull/3148), to use /starred  for all items and /starred/:feedId for sub items.

This PR adds the exact prop to control which active state - `isExactActive` or `isActive` - to use.

`isExactActive: true if the [exact active class](https://v3.router.vuejs.org/api/#exact-active-class) should be applied. Allows to apply an arbitrary class`

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
+++ b/src/components/NcAppNavigationItem/NcAppNavigationItem.vue
@@ -315,7 +315,7 @@ Just set the `pinned` prop.
 		class="app-navigation-entry-wrapper">
 		<component
 			:is="isRouterLink ? 'router-link' : 'NcVNodes'"
-			v-slot="{ href: routerLinkHref, navigate, isActive }"
+			v-slot="{ href: routerLinkHref, navigate, isExactActive }"
 			v-bind="{ ...isRouterLink && { custom: true, to } }">
 			<div
 				class="app-navigation-entry"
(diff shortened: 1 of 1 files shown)


ShGKme (changes requested):
In `v8` there was `exact` prop because `NcAppNavigationItem` is a wrapper over `RouterLink` which had the `exact` prop in the past in Vue 2. In other words, removed `NcAppNavigationItem`'s `exact` prop followed removed `<RouterLink>`'s `exact` prop.

This prop was removed in Vue Router v4 (Vue 3) ~5 years ago: https://github.com/vuejs/rfcs/blob/master/active-rfcs/0028-router-active-link.md

In my opinion, we should not bring back a prop removed in the base component.

Also, `isExactActive` is different from old `exact`. It acts as `exact-path`, ignoring query and hash.

Adding a prop with the same name but different behaviour seems even more confusing to me and harder to catch. This means that a prop is not removed but changed between v8 and v9, which is still a breaking change from v8, but different from mentioned in the breaking change notes.

In the current case, I'd use `isExactActive` instead of `isExact` without additional configuration to always the the full path.

In addition, I'd make `active` optional but allow override router's active by `active` prop. So a developer can manually specify when a link is active in a more complex case (for example, with query/hash).

wofferl:
@ShGKme can you review the requested changes
Since the News App has the feature that requires this fix in its beta version, it would be great if this fix could be included in 9.4.0

wofferl:
Rebased PR solving conflict with 377667a25d

PR #7941, opened 2025-12-01
fix(NcModal): scoped styles with teleport

ShGKme:
### ☑️ Resolves

- Regression from: https://github.com/nextcloud-libraries/nextcloud-vue/pull/7514
- Ref: https://github.com/vuejs/core/issues/2047
- Ref: https://github.com/vuejs/core/issues/2669
- Custom styles with scoped styles on `NcModal` are broken when relying on the root class
- `NcDialog` lost some styles

I don't know how to properly solve this problem. This is the solution I came up with, but it seems to work.

- Alternatives:
  - Revert migration to `<teleport>`
  - Introduce a hidden breaking change (only breaks custom styles)

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/NcModal/NcModal.vue
+++ b/src/components/NcModal/NcModal.vue
@@ -16,6 +16,7 @@ import NcActions from '../NcActions/NcActions.vue'
 import NcButton from '../NcButton/NcButton.vue'
 import NcIconSvgWrapper from '../NcIconSvgWrapper/NcIconSvgWrapper.vue'
 import { useHotKey } from '../../composables/index.ts'
+import { useScopeIdAttrs } from '../../composables/useScopeIdAttrs.ts'
 import { t } from '../../l10n.ts'
 import { createElementId } from '../../utils/createElementId.ts'
 import { getTrapStack } from '../../utils/focusTrap.ts'
@@ -187,6 +188,8 @@ defineSlots<{
 	 */
 	default?: Slot
 }>()
+
+const scopeIdAttrs = useScopeIdAttrs()
 
 const modalId = createElementId()
 const maskElement = useTemplateRef('mask')
@@ -401,7 +404,7 @@ function clearFocusTrap() {
 			@before-leave="clearFocusTrap">
 			<div
 				v-show="showModal"
-				v-bind="$attrs"
+				v-bind="{ ...$attrs, ...scopeIdAttrs }"
 				ref="mask"
 				class="modal-mask"
 				:class="{
--- a/src/composables/useScopeIdAttrs.ts
+++ b/src/composables/useScopeIdAttrs.ts
@@ -1,0 +1,82 @@
+/*!
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import type { ComponentInternalInstance } from 'vue'
+
+import { getCurrentInstance, warn } from 'vue'
+
+/**
+ * Get the parent instance of the instance only if the instance is the root node of the parent
+ * - Parent > Child - same root node
+ * - Parent > div > Child - different root node
+ * - Parent > WrapperWithSlot > Child - different root node
+ *
+ * @param instance - Current instance (internal)
+ */
+function getSameNodeParent(instance: ComponentInternalInstance): ComponentInternalInstance | null {
+	if (!instance.parent) {
+		return null
+	}
+
+	if ('vapor' in instance || 'vapor' in instance.parent) {
+		warn('Vapor instances are not supported in useScopeIdAttrs :(')
+		return null
+	}
+
+	if (instance.parent.subTree !== instance.vnode) {
+		return null
+	}
+
+	return instance.parent
+}
+
+/**
+ * Get all ancestor instances of the instance that are on the same root node
+ *
+ * @param instance - Current instance (internal)
+ */
+function getSameNodeAncestors(instance: ComponentInternalInstance): ComponentInternalInstance[] {
+	const ancestors: ComponentInternalInstance[] = [instance]
+	let parent = getSameNodeParent(instance)
+	while (parent) {
+		ancestors.push(parent)
+		parent = getSameNodeParent(parent)
+	}
+	return ancestors
+}
+
+/**
+ * Get a binding object for all data-v-scopeid attributes that are supposed to be on the root node.
+ * It allows to have scoped styles from parents in edge cases not covered by Vue:
+ * - Teleport on the root
+ * - Fragments
+ *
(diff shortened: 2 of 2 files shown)


ShGKme:
> * ⛑️ Tests are included or are not applicable

I'll add tests if we decide to continue with this approach

Antreesy:
NcModal still has incorrect styles in docs, or it's just because the example was written like this?

image

ShGKme:
Works on my machine 👀

image

Antreesy:
Locally also works =/ But not on a deployed preview

ShGKme:
Checking

susnux:
Cant we then just use css modules instead of scoped styles?

ShGKme:
> Cant we then just use css modules instead of scoped styles?

Using CSS Modules fixes `NcDialog`.

It doesn't fix a breaking change of the `NcModal` itself. Overriding modal styles is quite common.

ShGKme:
Pushed just a typo fix: `[instance.vnode.scopeId]` -> `instance.vnode.scopeId` as an attributes object key. Which doesn't change anything, because `[str]` also cast to `'str'` as string anyway.

ShGKme:
> NcModal still has incorrect styles in docs, or it's just because the example was written like this?

It is not related and happens on `main` as well.

I found the problem and will push in a different PR.

susnux:
> It doesn't fix a breaking change of the NcModal itself. Overriding modal styles is quite common.

Still I would argue thats expected behavior if you mess with private API that can change at any point.
You can always set content classes or put just your own content within.

ShGKme:
> You can always set content classes or put just your own content within.

You couldn't when you implemented `NcDialog`. Apart from content styling, entire modal styling is sometimes needed.

ShGKme:
We might say that it's _de jure_ not a breaking change. (Though it doesn't mean it must not be fixed, does it?)

But it's breaking **de facto**: an app bumps the UI library, and suddenly a modal is broken for the end user. The refactoring PR already contained a hidden force-pushed feature not mentioned in the title/label/description, now also contains an unmentioned `<style scoped>` compatibility restriction.

A developer facing this issue in an app has 2 choices:
1. Just move styles override to the global scope
2. Switch to CSS modules or another scoping method

Seeing how scoping issues are usually solved, I'm very confident that most of the developers choose the first option. In this case, it later breaks other apps.

This PR:
- Prevents breaking apps after library bump
- Prevents app developers from accidentally breaking other apps by going from `<style scope>` to `<style>`

@susnux could you elaborate what is the problem with the fix, brining back component compatibility with `<style scoped>`? Does it lead to any problems, or is there any reason not to have the fix? You said you would have argued that it had been a breaking change, but you did not argue with having the fix...

susnux:
> If we are to release a lib soon, that should go in at least as a hotfix

agree!

ShGKme:
I cannot call it a "hot fix", I don't see any "properer" fix with `<Teleport>`. Only if we fix it in Vue or switch to another approach from `<Teleport>`.

susnux:
> could you elaborate what is the problem with the fix

It looks like quite complex for a fix of non official API.
But in this case its probably fine because NcModal does not offer any official documented way to style the content.
Meaning you either have to apply classes to the default slot (first content is then an app specific wrapper) or rely on the scoping to work and apply styles to the root modal mask.

For NcDialog we have `contentClasses` you can assign your styles there - of course there is no scoping but thats known and you can either use a unique class name (in apps this is less of a problem), use css modules or just also put in your own slot content with scoping.

> I cannot call it a "hot fix", I don't see any "properer" fix with <Teleport>.

The proper way would be to provide `contentClasses` prop then there is a proper way to assign classes to the component.
Same as for the dialog.

> The refactoring PR already contained a hidden force-pushed feature

Nothing hidden there, also it was more of a fix which was tackled by adding a new option to some composable.
And then was reviewed again :)

ShGKme:
> The proper way would be to provide `contentClasses` prop then there is a proper way to assign classes to the component.
> Same as for the dialog.

Could you show an example how it will be used with scoped styles?

ShGKme:
> It looks like quite complex for a fix of non official API.

```html
<template>
  <!-- my-modal is a root class of the modal component -->
  <NcModal class="my-modal">
    <div class="my-content" />
  </NcModal>

  <!-- my-button is a root class of the button component -->
  <NcButton class="my-button">
    <div class="my-content" />
  </NcButton>
</template>
```

It seems very unclear that `class="my-button"` is official API and `class="my-modal"` is not

susnux:
> Could you show an example how it will be used with scoped styles?

It cannot just as with dialog you have use unique class names (either craft them yourself or use css modules).

> It seems very unclear that class="my-button" is official API and class="my-modal" is not

Yes but this is also a good example:
- for button its logical where the class is set: its on the component you expect it to be. (because the button is basically one component)
- for dialog or modal: Where do you expect the class? On the real dialog (`div role=dialog`) or on the modal backdrop container (which you do not necessary know to exist).

---

To rephrase: I think this is the proper fix for this major version and I do not necessarily think we should remove this in the next. But in general I think that the proper way to style your dialogs is to either put in content classes as unique classes or scoped styles with your example `class=my-content`.

ShGKme:
The only alternative I see is to completely remove `class` attribute support from `<NcModal>`, only leaving explicit classes for specific parts like `content-classes`, `container-classes` etc.

Otherwise there is a general class on the root element but broken scoping on it

ShGKme:
> It cannot just as with dialog you have use unique class names (either craft them yourself or use css modules).

But we cannot force component users to use CSS Modules

ShGKme:
> * for dialog or modal: Where do you expect the class? On the real dialog (`div role=dialog`) or on the modal backdrop container (which you do not necessary know to exist).

Without any additional documentation or explicit restrictions, I expect a class to be on the root. Props to pass classes deeper are usually named differently.

ShGKme:
We can fix usage in `NcDialog` by migration `NcDialog` to `CSS Modules`.

But for the component users it is more complicated. Even with props for internal classes.

```html
<!-- MyModal.vue -->

<template>
  <NcModal
    class="my-modal"
    container-classes="my-modal-container"
    container-content-classes="my-modal-container-content">
    ...
  </NcModal>
</template>

<!-- We cannot force component user to use CSS Modules -->
<style scoped>
.my-modal,
.my-modal-container,
.my-modal-container-content {
  /* Doesn't work */
}

:deep(.my-modal),
:deep(.my-modal-container),
:deep(.my-modal-container-content) {
  /* Doesn't work */
}
</style>

<!-- Works -->
<!-- Developers use it as the only working option -->
<!-- Styles leak -->
<style>
.my-modal,
.my-modal-container,
.my-modal-container-content {
  /* Unfortunately, works */
}
</style>
```

PR #7966, opened 2025-12-09
fix(useHotKey): do not prevent hotkeys on hidden modal/dialog

skjnldsv:
Found working on Server upgrade. If the viewer is idling waiting to be shown, the element is still present in the dom.
We need to ensure it's also really shown.

### 🏁 Checklist

### DIFF
--- a/src/composables/useHotKey/index.ts
+++ b/src/composables/useHotKey/index.ts
@@ -72,8 +72,11 @@ function shouldIgnoreEvent(event: KeyboardEvent, options: UseHotKeyOptions): boo
 		return false
 	}
 
-	/** Abort if any modal/dialog opened */
-	return document.getElementsByClassName('modal-mask').length !== 0
+	/** Abort if any modal/dialog is opened AND visible */
+	return Array.prototype.filter.call(
+		document.getElementsByClassName('modal-mask'),
+		(el) => el.checkVisibility(),
+	).length > 0
 }
 
 type KeyboardEventHandler = (event: KeyboardEvent) => void
--- a/tests/unit/composables/useHotKey.spec.js
+++ b/tests/unit/composables/useHotKey.spec.js
@@ -40,16 +40,31 @@ describe('useHotKey', () => {
 		expect(mockCallback).not.toHaveBeenCalled()
 	})
 
-	it('should not invoke callback by default, when a modal is shown', () => {
+	it('should not invoke callback by default, when a modal is shown', async () => {
 		const modal = document.createElement('div')
 		modal.className = 'modal-mask'
 		document.body.appendChild(modal)
+		modal.checkVisibility = () => true
 
 		const stop = useHotKey(true, mockCallback)
 		triggerKeyDown({ key: 'a', code: 'KeyA' })
 		stop()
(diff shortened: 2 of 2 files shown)


skjnldsv:
I was a bit too quick, let me adjust the test

src/composables/useHotKey/index.ts
```diff
@@ -72,8 +72,10 @@ function shouldIgnoreEvent(event: KeyboardEvent, options: UseHotKeyOptions): boo
 		return false
 	}
 
-	/** Abort if any modal/dialog opened */
-	return document.getElementsByClassName('modal-mask').length !== 0
+	/** Abort if any modal/dialog is opened AND visible */
+	return [...document.getElementsByClassName('modal-mask')]
+		.filter((el) => el.checkVisibility())
```
ShGKme: This seems very tricky anyway, even with `modal-mask` querying itself.

What about using a focus trap? Saying that a focus trap (trapping the Tab hotkey) is also trapping all the hotkeys.

skjnldsv: I'd say unrelated to this PR. I'm fixing the current code.
I don't have time to dive into tech debt

skjnldsv:
> Fine for a hotfix, but I'd propose to remove `allowInModal` in favour of hotkey scoping similar to the focus trap.

Referenced in new ticket :+1:

src/composables/useHotKey/index.ts
```diff
@@ -72,8 +72,11 @@ function shouldIgnoreEvent(event: KeyboardEvent, options: UseHotKeyOptions): boo
 		return false
 	}
 
-	/** Abort if any modal/dialog opened */
-	return document.getElementsByClassName('modal-mask').length !== 0
+	/** Abort if any modal/dialog is opened AND visible */
+	return Array.prototype.filter.call(
+		document.getElementsByClassName('modal-mask'),
+		(el) => el.checkVisibility(),
+	).length > 0
```
Antreesy, line 79: Maybe a personal preference, but can we avoid using the prototype/call syntax?

```suggestion
	return Array.from(document.getElementsByClassName('modal-mask'))
		.filter((el) => el.checkVisibility())
		.length > 0
```

skjnldsv: Ah yes, I forgot about Array.from :see_no_evil: 
I just used the recommended way from mozilla.org https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementsByClassName

image

skjnldsv: https://github.com/nextcloud-libraries/nextcloud-vue/pull/7972

ShGKme: You can also use the spread operator:
```js
[...document.getElementsByClassName('modal-mask')].filter()
```

`filter.call` might be more performant by using iterator, but it isn't relevant for the modal dialogs list.

skjnldsv: SO, this is what I initially did, as I like the spread operator, but ts complained, forcing me to go back to other approaches :)

PR #7971, opened 2025-12-10
[stable8] fix(useHotKey): do not prevent hotkeys on hidden modal/dialog

skjnldsv:
- Manual backport of https://github.com/nextcloud-libraries/nextcloud-vue/pull/7966

### DIFF
--- a/tests/unit/composables/useHotKey.spec.js
+++ b/tests/unit/composables/useHotKey.spec.js
@@ -19,6 +19,7 @@ describe('useHotKey', () => {
 
 	afterEach(() => {
 		mockCallback.mockReset()
+		document.body.innerHTML = ''
 	})
 
 	it('should register listener and invoke callback', () => {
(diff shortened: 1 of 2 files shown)


skjnldsv:
For some reason cypress won't run because of outdated lockfile, adding it here :shrug: 

EDIT: seems like the stable8 branch is broken
https://github.com/nextcloud-libraries/nextcloud-vue/actions/runs/20019933718/job/57404620238

skjnldsv:
Why is it failing now ?? :cry:

tests/unit/composables/useHotKey.spec.js
```diff
@@ -19,6 +19,7 @@ describe('useHotKey', () => {
 
 	afterEach(() => {
 		mockCallback.mockReset()
+		document.body.innerHTML = ''
```
ShGKme, line 22: Safer to remove the created modal to make sure nothing else is removed by mistake (for example, if it is run in a Vitest web-browser environment or Playwright component tests)

skjnldsv: I'll use the same code as main to ensure no one else get confused like I was

PR #7972, opened 2025-12-10
refactor(useHotKey): modal visibility check using Array.from

skjnldsv:
> Maybe a personal preference, but can we avoid using the prototype/call syntax?
> 
> ```ts
> 	return Array.from(document.getElementsByClassName('modal-mask'))
> 		.filter((el) => el.checkVisibility())
> 		.length > 0
> ```

_Originally posted by @Antreesy in https://github.com/nextcloud-libraries/nextcloud-vue/pull/7966#discussion_r2605884973_


Antreesy:
Already in the backport

PR #5155, opened 2024-01-27
feat(NcPopover): auto return focus to trigger button on close

ShGKme:
### ☑️ Resolves

* Fix #5141

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 Tasks

- [x] Automatically find a button inside the `NcPopover` trigger and use it as fallback for `setReturnFocus`
- [x] Add `noAutoReturnFocus` to optionally disable this behavior
  - Useful when there is a custom focus control, like in `NcActions`
- [ ] Adjust `NcActions`

### 🏁 Checklist

- [ ] ⛑️ Tests are included or are not applicable
- [x] 📘 Component documentation has been extended, updated or is not applicable

PR #7975, opened 2025-12-10
[stable8] fix(NcPopover): Don't warn on hidden button elements

mejo-:
`tabbable(<element>)` doesn't return elements that are hidden. This makes the a11y test warn if the trigger element is hidden.

In Nextcloud Text the menubar gets initialized with `visibility: hidden`.

No need to forward-patch this to `main` as `tabbable` is not used in NcPopover there.

### 🏁 Checklist

### DIFF
--- a/src/components/NcPopover/NcPopover.vue
+++ b/src/components/NcPopover/NcPopover.vue
@@ -287,15 +287,14 @@ export default {
 	watch: {
 		shown(value) {
 			this.internalShown = value
+			if (this.internalShown) {
+				this.checkTriggerA11y()
+			}
 		},
 
 		internalShown(value) {
 			this.$emit('update:shown', value)
 		},
-	},
-
-	mounted() {
-		this.checkTriggerA11y()
 	},
 
 	beforeDestroy() {


src/components/NcPopover/NcPopover.vue
```diff
@@ -310,8 +310,9 @@ export default {
 		 */
 		checkTriggerA11y() {
 			if (window.OC?.debug) {
-				const triggerButton = this.getPopoverTriggerButtonElement()
-				if (!triggerButton || !triggerButton.hasAttributes('aria-expanded', 'aria-haspopup')) {
+				const triggerContainer = this.getPopoverTriggerElement()
+				const requiredTriggerButton = triggerContainer.querySelector('[aria-expanded], [aria-popover]')
+				if (!requiredTriggerButton) {
```
ShGKme: This doesn't check that the trigger is an interactive element.

mejo-: It's how the check looked like before `tabbable` was introduced in `NcPopover`. And apparently it's how it still looks like on `main` branch: https://github.com/nextcloud-libraries/nextcloud-vue/blob/ddf0ee840ed0cde30469faa31598bf7ea325eb97/src/components/NcPopover/NcPopover.vue#L472-L480

@ShGKme do you have another idea how to fix the described issue then? Tabbable has a [displayCheck option](https://github.com/focus-trap/tabbable?tab=readme-ov-file#displaycheck-option), but passing `{displayCheck: 'none'}` to `tabbable()` doesn't solve the problem, as apparently hidden elements are always ignored.

susnux: I wonder if this makes sense at all. Because now you silence the warning and it will never warn you.

Should we not just call `checkTriggerA11y` always when `shown` is set to `true`?
Then the problem would also be solved?

ShGKme: Sounds good to me if we do it once `shown` is set to true, but not every time(to save some performance). Or do it only in development.

mejo-: Good idea. I changed the behaviour accordingly, pleas have a look. Since `checkTrickerA11y()` only runs in dev mode anyway, I didn't implement extra logic to only run it once.

src/components/NcPopover/NcPopover.vue
```diff
@@ -287,17 +287,16 @@ export default {
 	watch: {
 		shown(value) {
 			this.internalShown = value
+			if (!this.triggerA11yChecked) {
```
susnux:
```suggestion
			if (this.internalShown) {
```

mejo-: Ouch, good catch. Was a leftover from introducing the `triggerA11yChecked` flag first, which I reverted later. Fixed now.

ShGKme (approved):

src/components/NcPopover/NcPopover.vue
max-nextcloud: > I wonder if this makes sense at all. Because now you silence the warning and it will never warn you.

@susnux I don't quite understand - how would the check for `requiredTriggerButton` silence the warning?
The check is similar to the way it was before. It still performs the check on a hidden menu - it just also accepts a hidden trigger button.

Only triggering on `show` also means it will only warn you if you actually open the menu. That seems much easier to miss.

susnux: > Only triggering on show also means it will only warn you if you actually open the menu. That seems much easier to miss.

Well but when the menu is opened the menu must be visible thats what it checked.
If the popover is not visible the trigger button also does not need to be visible, but as soon as the popover is shown the trigger needs to be visible.

susnux (approved):

mejo-:
Shall I forward-port this and 2ce15eb24b6b3304873399f2ee692acd4182474c to `main`? @ShGKme was there a particular reason why #5155 was only applied to `stable8` and not to `main`?

ShGKme:
@mejo- I don't know a reason for having #5155, it was 1.5 years old PR I forgot to close.

PR #8009, opened 2025-12-15
feat(NcDateTimePicker): add range limit props and inline option

GVodyanov:
We noticed that some properties that we used of the old datepicker in calendar didn't exist in the new one.

### ☑️ Resolves

- Needed for https://github.com/nextcloud/calendar/pull/7697

### 🖼️ Screenshots

Added:
image

image

image

image

### 🏁 Checklist

### DIFF
--- a/src/components/NcDateTimePicker/NcDateTimePicker.vue
+++ b/src/components/NcDateTimePicker/NcDateTimePicker.vue
@@ -163,6 +163,104 @@ export default {
 }
 </script>
 ```
+
+### Restricting to a specific time range
+
+You can optionally restrict the selectable time range, for example from 08:00 on Monday to 20:00 on Sunday of the current week.
+This example shows how the `min` and `max` properties are applied to different picker types.
+
+```vue
+<template>
+	<div>
+		<h4>Date and time</h4>
+		<NcDateTimePicker
+			v-model="selectedDatetime"
+			type="datetime"
+			:min="min"
+			:max="max" />
+		<p>Selected: {{ selectedDatetime }}</p>
+
+		<h4>Date range</h4>
+		<NcDateTimePicker
+			v-model="selectedDateRange"
+			type="date-range"
+			:min="min"
+			:max="max" />
+		<p>Selected: {{ selectedDateRange }}</p>
+
+		<h4>Date and time range</h4>
+		<NcDateTimePicker
+			v-model="selectedDatetimeRange"
+			type="datetime-range"
+			:min="min"
+			:max="max" />
+		<p>Selected: {{ selectedDatetimeRange }}</p>
+
+		<h4>Time range</h4>
+		<NcDateTimePicker
+			v-model="selectedTimeRange"
+			type="time-range"
+			:min="min"
+			:max="max" />
+		<p>Selected: {{ selectedTimeRange }}</p>
+
+		<h4>Time</h4>
+		<NcDateTimePicker
+			v-model="selectedTime"
+			type="time"
+			:min="min"
+			:max="max" />
+		<p>Selected: {{ selectedTime }}</p>
+	</div>
+</template>
+<script>
+export default {
+	data() {
+		const now = new Date()
+		const day = now.getDay()
+		const daysSinceMonday = (day + 6) % 7
+
+		const startOfWeek = new Date(now)
+		startOfWeek.setDate(now.getDate() - daysSinceMonday)
+		startOfWeek.setHours(8, 0, 0, 0)
+
+		const endOfWeek = new Date(startOfWeek)
+		endOfWeek.setDate(startOfWeek.getDate() + 6)
+		endOfWeek.setHours(20, 0, 0, 0)
+
+		return {
+			selectedDatetime: null,
+			selectedDateRange: null,
+			selectedDatetimeRange: null,
+			selectedTimeRange: null,
+			selectedTime: null,
+			min: startOfWeek,
+			max: endOfWeek,
+		}
+	},
+}
+</script>
+```
+
+### Inline calendar
+
+Renders the calendar inline (without the text input) inside your parent component.
+
+```vue
+<template>
+	<div>
+		<NcDateTimePicker v-model="date" inline />
+		<p>Selected: {{ date }}</p>
+	</div>
+</template>
+<script>
+export default {
+	data() {
+		return { date: new Date() }
+	},
+}
+</script>
+```
 </docs>
 
 <script setup lang="ts">
@@ -260,6 +358,16 @@ const props = withDefaults(defineProps<{
 	locale?: string
 
 	/**
+	 * The maximum date that can be selected.
+	 */
+	max?: Date
+
+	/**
+	 * The minimum date that can be selected.
+	 */
+	min?: Date
+
+	/**
 	 * Default increment step for minutes in the time picker.
 	 *
 	 * @default 10
@@ -306,17 +414,27 @@ const props = withDefaults(defineProps<{
 	 * @default 'date'
 	 */
 	type?: 'date' | 'datetime' | 'time' | 'week' | 'month' | 'year' | 'date-range' | 'time-range' | 'datetime-range'
+
+	/**
+	 * Render the calendar inline (no input field).
+	 *
+	 * @default false
+	 */
+	inline?: boolean
 }>(), {
 	ariaLabel: t('Datepicker input'),
 	ariaLabelMenu: t('Datepicker menu'),
 	format: undefined,
 	locale: getCanonicalLocale(),
+	max: undefined,
+	min: undefined,
 	minuteStep: 10,
 	timezoneId: 'UTC',
 	modelValue: null,
 	// set by fallbackPlaceholder
 	placeholder: undefined,
 	type: 'date',
+	inline: false,
 })
 
 const emit = defineEmits<{
@@ -327,6 +445,10 @@ const emit = defineEmits<{
 	 */
 	'update:modelValue': [Date | [Date, Date] | null]
 	'update:timezoneId': [string]
+	/**
+	 * Input blur
+	 */
+	blur: []
 }>()
 
 const targetElement = useTemplateRef('target')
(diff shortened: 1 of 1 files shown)


SebastianKrupinski (approved):

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -603,18 +697,24 @@ function cancelSelection() {
...
+			:inline="inline"
 			:teleport="appendToBody ? (targetElement || undefined) : false"
 			text-input
 			:week-num-name
 			:week-numbers="showWeekNumber ? { type: 'iso' } : undefined"
 			:week-start
 			v-bind="pickerType"
-			@update:model-value="onUpdateModelValue">
+			@update:model-value="onUpdateModelValue"
+			@blur="() => emit('blur')">
```
ShGKme:
```suggestion
			@blur="emit('blur')">
```

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -603,18 +697,24 @@ function cancelSelection() {
...
+			:min-date="min || undefined"
+			:max-date="max || undefined"
+			:min-time="minTime"
+			:max-time="maxTime"
 			:minutes-increment="minuteStep"
 			:model-value="value"
 			:now-button-label="t('Now')"
 			:select-text="t('Pick')"
 			six-weeks="fair"
+			:inline="inline"
```
ShGKme: Vue 3 :P

```suggestion
			:min-time
			:max-time
			:minutes-increment="minuteStep"
			:model-value="value"
			:now-button-label="t('Now')"
			:select-text="t('Pick')"
			six-weeks="fair"
			:inline
```

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -259,6 +313,16 @@ const props = withDefaults(defineProps<{
...
 
+	/**
+	 * The maximum date that can be selected.
+	 */
+	max?: Date | null
+
+	/**
+	 * The minimum date that can be selected.
+	 */
+	min?: Date | null
```
ShGKme: Why do we need `null` here? It is not different from not passing a prop (passing `undefined`)

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -327,6 +401,11 @@ const emit = defineEmits<{
 	 */
 	'update:modelValue': [Date | [Date, Date] | null]
 	'update:timezoneId': [string]
+	/**
+	 * Forwarded when the internal input loses focus.
+	 * This is a custom event from the inner component with no payload.
+	 */
+	blur: []
```
ShGKme, line 451: For simplicity:
- All custom events are custom events
- The payload is described by the type
- Blur is a known action

```suggestion
	/**
	 * Input blur
	 */
	blur: []
```

GVodyanov:
Thanks a lot for the detailed review @ShGKme! Should be good now

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -259,6 +313,16 @@ const props = withDefaults(defineProps<{
...
 
+	/**
+	 * The maximum date that can be selected.
+	 */
+	max?: Date
+
+	/**
+	 * The minimum date that can be selected.
+	 */
+	min?: Date
```
ShGKme, line 368: Does it check the full date value, including time, or only the date?
If only the date, I'd rename prop to `minDate`.
In any case, I'd mention it in the prop docs.

ShGKme: And the same for the timezone

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -603,18 +696,24 @@ function cancelSelection() {
 			:placeholder="placeholder ?? placeholderFallback"
 			:format="realFormat"
 			:locale
+			:min-date="min || undefined"
+			:max-date="max || undefined"
```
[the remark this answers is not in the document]
ShGKme: Not needed when `min` cannot be `null`.

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -463,6 +541,21 @@ const pickerType = computed(() => ({
...
+	}
+	return { hours: props.min.getHours(), minutes: props.min.getMinutes(), seconds: props.min.getSeconds() }
+})
+
+const maxTime = computed(() => {
+	if (!props.max) {
+		return undefined
+	}
+	return { hours: props.max.getHours(), minutes: props.max.getMinutes(), seconds: props.max.getSeconds() }
+})
```
ShGKme: **Very** nitpick.

```suggestion
const minTime = computed(() => props.min && { hours: props.min.getHours(), minutes: props.min.getMinutes(), seconds: props.min.getSeconds() })
const maxTime = computed(() => props.max && { hours: props.max.getHours(), minutes: props.max.getMinutes(), seconds: props.max.getSeconds() })
```

or

```suggestion
const minTime = computed(() => props.min && {
  hours: props.min.getHours(),
  minutes: props.min.getMinutes(),
  seconds: props.min.getSeconds(),
})
const maxTime = computed(() => props.max && {
  hours: props.max.getHours(),
  minutes: props.max.getMinutes(),
  seconds: props.max.getSeconds(),
})
```

src/components/NcDateTimePicker/NcDateTimePicker.vue
susnux: But thats already date or undefined?


```suggestion
			:min-date="min"
			:max-date="max"
```

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -603,18 +696,24 @@ function cancelSelection() {
...
+			:inline
 			:teleport="appendToBody ? (targetElement || undefined) : false"
 			text-input
 			:week-num-name
 			:week-numbers="showWeekNumber ? { type: 'iso' } : undefined"
 			:week-start
 			v-bind="pickerType"
-			@update:model-value="onUpdateModelValue">
+			@update:model-value="onUpdateModelValue"
+			@blur="emit('blur')">
```
susnux, line 798: For what is the blur event needed?

GVodyanov: In calendar we have a few things that have to be called when the time selected in a datepicker is changed, calling them every time the user updates the time without closing the popover is rather impractical as it means firing everything multiple times. Blur is called when the user finalizes their choice of time and deselects the component.

For example, u don't want the end time of the event updated while you're typing, because while typing like "08:40" first you send the update as "08:00".

susnux: But then just set the `confirm` prop?

src/components/NcDateTimePicker/NcDateTimePicker.vue
GVodyanov: Full date value, the reason for calling it `max` and `min` is that they are what is used in the native date picker and in the HTML element for the same purpose

GVodyanov:
So for people following along: I made a small change to make the min/max function exactly like the native time picker.

With my previous implementation it would apply minTime to every day, so for example you would set minTime to 08:00 and that means that you would not able to select before 08:00 on any given day, now however I made min/max represent a range, so if you set min to 08:00 on Monday and max 18:00 on Wednesday you will be able to select any time in between.

GVodyanov:
Now I also added support for ranges and `type="time"`

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -676,6 +714,41 @@ function cancelSelection() {
...
+			minTime: props.min ? minTime.value : undefined,
+			maxTime: props.max ? maxTime.value : undefined,
+		}
+	}
+
+	return {
+		minDate: props.min,
+		maxDate: props.max,
+	}
+}
```
ShGKme: It looks like it should be a computed

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -587,6 +713,55 @@ function selectDate() {
...
+/**
+ *
+ */
+function calcMinMaxTime() {
+	if (props.type === 'datetime') {
+		return {
+			minDate: props.min,
+			maxDate: props.max,
+			minTime: props.min && value.value ? sameDay(props.min, value.value as Date) ? minTime.value : undefined : undefined,
+			maxTime: props.max && value.value ? sameDay(props.max, value.value as Date) ? maxTime.value : undefined : undefined,
```
ShGKme: To remove nested ternary operator
```suggestion
			minTime: props.min && value.value && sameDay(props.min, value.value as Date) ? minTime.value : undefined,
			maxTime: props.max && value.value && sameDay(props.max, value.value as Date) ? maxTime.value : undefined,
```

ShGKme: Is it guaranteed that `value.value` is a `Date` here?

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -603,18 +778,24 @@ function cancelSelection() {
 			:placeholder="placeholder ?? placeholderFallback"
 			:format="realFormat"
 			:locale
+			:min-date="calcMinMaxTime().minDate"
+			:max-date="calcMinMaxTime().maxDate"
+			:min-time="calcMinMaxTime().minTime as LibraryTimeObject"
+			:max-time="calcMinMaxTime().maxTime as LibraryTimeObject"
```
ShGKme: It is not a very big problem, but here the same object is calculated 4 times every rendering, even when nothing changed around `min`/`max`

src/components/NcDateTimePicker/NcDateTimePicker.vue
GVodyanov: It isn't always `Date` but it should be for the types I am considering

ShGKme: If it is not always `Date` here, we cannot use `as Date`. What happens when it is not `Date`?

GVodyanov: What I mean is if we are talking about the only time selector then we get a `TimeObj`, but those `as` are inside ifs that won't be reached in case of an only time selector

SebastianKrupinski:
Hello all, happy holidays.

Just wanted to bump this PR to get the reaming, piece of the migration solved.

SebastianKrupinski:
Hello all, can we get this PR out of the abyss and in to the light? LOL

This is the remaining piece we need for Calendar 6.2 RC

ChristophWurst (approved):

susnux (approved):

PR #8025, opened 2025-12-19
[stable8] chore(l10n): merge translations from main branch to pot files

Antreesy:
### ☑️ Resolves

- Fix missing translations on stable8 branch

### DIFF
--- a/l10n/af.pot
+++ b/l10n/af.pot
@@ -277,3 +277,6 @@ msgid ""
 "Write message, use \"@\" to mention someone, use \":\" for emoji "
 "autocompletion …"
 msgstr ""
+
+msgid "More items …"
+msgstr ""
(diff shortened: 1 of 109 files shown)


susnux:
* closes https://github.com/nextcloud-libraries/nextcloud-vue/pull/7262

susnux:
> closes https://github.com/nextcloud-libraries/nextcloud-vue/pull/7262

Ah no it does not. This PR does not work.
it seems you just copied the main branch translations, but the source strings are different, so the copied translations do not work.

You have to proper merge translations from main (lookup source string used in stable and copy translation from main).
Like the script I posted in the other PR description.

Antreesy:
It does for some reported strings (e..g. tested with `Write a message`), but yeah, 'Keyboard shortcuts' are missing as well 👀

PR #8042, opened 2025-12-30
fix(NcModal): Adjust max height to leave some space

kra-mo:
This is so that the modal doesn't exactly touch the edges of the header, which looks a bit weird. Sort of a follow-up to 4d9e81dcd3c1a878472d8bf9d0797a13c38a8a67.

Discussed with @jancborchardt and @nimishavijay 

Not sure about backports?

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Before | After

### 🏁 Checklist

### DIFF
--- a/src/components/NcModal/NcModal.vue
+++ b/src/components/NcModal/NcModal.vue
@@ -733,7 +733,7 @@ function clearFocusTrap() {
 
 	// We allow 90% max-height, but we need to ensure the header does not overlap the modal
 	// as the modal is centered, we need the space on top and bottom
-	$max-modal-height: min(90%, calc(100% - 2 * var(--header-height)));
+	$max-modal-height: min(90%, calc(100% - 2 * var(--header-height) - 2 * var(--body-container-margin)));
 
 	// Sizing
 	&--small {


src/components/NcModal/NcModal.vue
```diff
@@ -733,7 +733,7 @@ function clearFocusTrap() {
 
 	// We allow 90% max-height, but we need to ensure the header does not overlap the modal
 	// as the modal is centered, we need the space on top and bottom
-	$max-modal-height: min(90%, calc(100% - 2 * var(--header-height)));
+	$max-modal-height: min(90%, calc(100% - 2 * var(--header-height) - 4 * var(--default-grid-baseline)));
```
ShGKme:
```suggestion
	$max-modal-height: min(90%, calc(100% - 2 * var(--header-height) - 2 * var(--body-container-margin)));
```

kra-mo: I don't know what that means, but I trust you 😅

Antreesy: image

kra-mo: Ah, like that. Ok, I see.

ShGKme: It is the size of the margin around the entire content (orange part on the screen shot, except for the header).

image

PR #8045, opened 2025-12-30
fix: animate toggle switch icon

ShGKme:
### ☑️ Resolves

- Fix: https://github.com/nextcloud-libraries/nextcloud-vue/issues/8043

### 🖼️ Screenshots

switch-1

switch-2

### 🏁 Checklist

### DIFF
--- a/src/components/NcIconToggleSwitch/NcIconToggleSwitch.vue
+++ b/src/components/NcIconToggleSwitch/NcIconToggleSwitch.vue
@@ -1,0 +1,66 @@
+<!--
+  - SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<script lang="ts">
+const svg = `<svg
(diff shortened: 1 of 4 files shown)


src/components/NcIconToggleSwitch/NcIconToggleSwitch.vue
```diff
@@ -0,0 +1,66 @@
...
+
+	circle {
+		cx: v-bind(cx);
+		transition: cx var(--animation-quick) ease;
+	}
+}
+</style>
+
+<docs>
+Private component
```
susnux, line 65: Maybe expose it just like `NcLoadingIcon`?

susnux (approved):

src/components/NcIconToggleSwitch/NcIconToggleSwitch.vue
ShGKme: IMO, existing components should cover all the cases of toggle switches. Loading icon, in contrast, is used directly in most of the cases because we don't have any alternatives like loading states of containers/buttons/inputs.

PR #8061, opened 2026-01-08
chore(deps): pin focus-trap library to v7.6.6

Antreesy:
### ☑️ Resolves

- Fix incompatibility issue coming from focus-trap 7.7.0
- Pin exact version until it's solved
- Add ignore pattern to dependabot.yml

### 🏁 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,7 @@
     "emoji-mart-vue-fast": "^15.0.5",
     "escape-html": "^1.0.3",
     "floating-vue": "^5.2.2",
-    "focus-trap": "^7.7.1",
+    "focus-trap": "7.6.6",
     "linkifyjs": "^4.3.2",
     "p-queue": "^9.1.0",
     "rehype-external-links": "^3.0.0",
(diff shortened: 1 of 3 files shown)


ShGKme:
- Ref: https://github.com/focus-trap/focus-trap/issues/1729

mejo- (approved):

package.json
```diff
@@ -97,7 +97,7 @@
     "emoji-mart-vue-fast": "^15.0.5",
     "escape-html": "^1.0.3",
     "floating-vue": "^5.2.2",
-    "focus-trap": "^7.7.1",
+    "focus-trap": "7.6.6",
```
susnux, line 100: why not


```suggestion
    "focus-trap": "~7.6.6",
```

Antreesy: Was done with `npm install --save-exact`
Also by changelog history for `focus-trap`, I assume there are no plans to release patch versions for 7.6.x

ShGKme: And if there is 7.6.7, I'd not update automatically anyway

ShGKme: There is no guarantee for new versions compatibility with older versions in the same stack

Issue #7946, opened 2025-12-02
v9 regression: selecting an NcSelect option inside NcModal resets focused element


Sector6759:
Hey, in v9, if you include an `NcSelect` inside an `NcModal`/`NcDialog`, there is a weird focus behavior when selecting an option of the `NcSelect`, resulting in no element being focus afterwards.

## Steps to reproduce the faulty behavior

1. Go to https://nextcloud-vue-components.netlify.app/#/Components/NcDialog
2. Under the first "Basic example", click "VIEW CODE" and paste the code below into the editor
3. Click "Show dialog"
4. Observe the `NcTextField` being selected
5. Press `tab` once to move the focus to the `NcSelect`
6. Using the arrow keys, select an option in the list, then press `Enter`
7. Observe how there is no focus on any element
8. Press `tab` again and observe the focus is back on the first focusable element, which is the `NcTextField`

## Observed behavior

After pressing `Enter` in step 6, any element focus is lost

## Expected behavior

After pressing `Enter` in step 6, the focus should still be on the `NcSelect` , like it is in v8

## Reproduction code for faulty behavior in v9

```vue
<template>
  <div>
    <NcButton @click="showDialog = true">Show dialog</NcButton>
    <NcDialog v-model:open="showDialog" name="Confirmation" :buttons="buttons">
      <NcTextField
        v-model="text1"
        label="Leading icon and clear trailing button"
        trailing-button-icon="close"
        :show-trailing-button="text1 !== ''"
      >
        <template #icon>
          <Magnify :size="20" />
        </template>
      </NcTextField>
      <NcSelect
        input-label="Require a selection"
        :options="options"
        v-model="singleValue"
        required
      />
    </NcDialog>
    <p>Last response: {{ lastResponse }}</p>
  </div>
</template>
<script>
import IconCancel from "@mdi/svg/svg/cancel.svg?raw";
import IconCheck from "@mdi/svg/svg/check.svg?raw";

export default {
  data() {
    return {
      text1: "",
      options: ["foo", "bar", "baz", "qux", "quux"],
      singleValue: null,
      showDialog: false,
      lastResponse: "None",
      buttons: [
        {
          label: "Cancel",
          icon: IconCancel,
          callback: () => {
            this.lastResponse = 'Pressed "Cancel"';
          },
        },
        {
          label: "Ok",
          type: "primary",
          icon: IconCheck,
          callback: () => {
            this.lastResponse = 'Pressed "Ok"';
[...]

PR #8093, opened 2026-01-17
fix(NcModal): prevent focus trap race condition 

### LINKED ISSUE #8090: `useFocusTrap`: `fallbackFocus` was specified but was not a node, or did not return a node
I see this sometimes in the browser dev tools but it also regularly (30% chance) breaks server frontend unit tests:

```
 Error: `fallbackFocus` was specified but was not a node, or did not return a node
 ❯ getNodeForOption ../../node_modules/focus-trap/dist/focus-trap.js:284:13
 ❯ updateTabbableNodes ../../node_modules/focus-trap/dist/focus-trap.js:410:46
 ❯ Object.activate ../../node_modules/focus-trap/dist/focus-trap.js:866:9
 ❯ useFocusTrap ../../node_modules/@nextcloud/vue/src/components/NcModal/NcModal.vue:384:1

This error originated in "apps/federatedfilesharing/src/services/dialogService.spec.ts" test file. It doesn't mean the error was thrown inside the file itself, but while it was running.
```

nikhil2297:
* resolves https://github.com/nextcloud-libraries/nextcloud-vue/issues/8090

### ☑️ Resolves

- Fix focus trap race condition where `fallbackFocus` error occurs when `onMounted` calls `useFocusTrap()` before template refs are ready 
- Reference : #8090 

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Focus trap error: `fallbackFocus was specified but was not a node` occurring ~30% of the time during modal mounting | Error prevented by null check - focus trap waits for DOM elements to be ready

### 🚧 Tasks

- [x] Add null check in `useFocusTrap()` to prevent race condition
- [x] Add test cases to verify the fix
- [x] Ensure existing functionality is preserved

### 🏁 Checklist

### DIFF
--- a/src/components/NcModal/NcModal.vue
+++ b/src/components/NcModal/NcModal.vue
@@ -384,6 +384,11 @@ async function useFocusTrap() {
 	// wait until all children are mounted and available in the DOM before focusTrap can be added
 	await nextTick()
 
+	// ensuring the mask element is available before creating the focus trap
+	if (!maskElement.value) {
+		return
+	}
+
 	const options: FocusTrapOptions = {
 		allowOutsideClick: true,
 		fallbackFocus: maskElement.value!,
--- a/src/components/NcModal/NcModal.vue
+++ b/src/components/NcModal/NcModal.vue
@@ -196,8 +196,16 @@ const maskElement = useTemplateRef('mask')
 
 // Set up the focus trap
 let focusTrap: FocusTrap | undefined
-onMounted(() => useFocusTrap())
+
 onUnmounted(() => clearFocusTrap())
+
+// Watch for mask element to be available
+watch(maskElement, (el) => {
+	if (el) {
+		useFocusTrap()
+	}
+})
+
 watch(() => props.additionalTrapElements, (elements) => {
 	if (focusTrap) {
 		focusTrap.updateContainerElements([maskElement.value!, ...elements])
@@ -380,11 +388,11 @@ async function useFocusTrap() {
 	if (!showModal.value || focusTrap) {
 		return
 	}
-
 	// wait until all children are mounted and available in the DOM before focusTrap can be added
 	await nextTick()
 
-	// ensuring the mask element is available before creating the focus trap
+	// When called from @after-enter, the maskElement might not be available yet because of transition timing
+	// This gaurd ensures we only initialize the focus trap when the element is available
 	if (!maskElement.value) {
 		return
 	}
@@ -398,7 +406,6 @@ async function useFocusTrap() {
 		escapeDeactivates: false,
 		setReturnFocus: props.setReturnFocus,
 	}
-
 	// Init focus trap
 	focusTrap = createFocusTrap([maskElement.value!, ...props.additionalTrapElements], options)
 	focusTrap.activate()
--- a/src/components/NcModal/NcModal.vue
+++ b/src/components/NcModal/NcModal.vue
@@ -194,18 +194,11 @@ const scopeIdAttrs = useScopeIdAttrs()
 const modalId = createElementId()
 const maskElement = useTemplateRef('mask')
 
-// Set up the focus trap
+// Set up the focus trap (handled by transition event listeners)
 let focusTrap: FocusTrap | undefined
-
+// Ensure focus trap is cleared when the component is unmounted
 onUnmounted(() => clearFocusTrap())
-
-// Watch for mask element to be available
-watch(maskElement, (el) => {
-	if (el) {
-		useFocusTrap()
-	}
-})
-
+// and update the focus trap container elements when props change
 watch(() => props.additionalTrapElements, (elements) => {
 	if (focusTrap) {
 		focusTrap.updateContainerElements([maskElement.value!, ...elements])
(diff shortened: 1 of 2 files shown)


susnux (commented):
Should we not also remove the `onMounted(() => useFocusTrap())` ?

nikhil2297:
> Should we not also remove the `onMounted(() => useFocusTrap())` ?

I added a null check in `useFocusTrap()` which fixes the issue and the `useFocusTrap()` already has guards for edge cases, so there's no downside to keeping the `onMounted()` call.

susnux:
> so there's no downside to keeping the onMounted() call.

Well its a useless function call as it will always fail :wink: 

But anyway: Can you please fix the failing linter?

nikhil2297:
> > so there's no downside to keeping the onMounted() call.
> 
> Well its a useless function call as it will always fail 😉
> 
> But anyway: Can you please fix the failing linter?

Perfect, I've fixed the linter part and did the `npm run lint ` test everything is good. But I'm skeptical about the `new line` which I have added at EOF but idk why I'm not able to see it in the github commit file changes.

susnux (approved):
Still think that the known-to-fail `onMounted` call should be removed but the fix itself seems to work fine!

tests/component/components/NcModal.spec.ts
```diff
@@ -111,3 +111,31 @@ test('Close button is visible when content is scrolled', async ({ mount, page })
...
+		},
+	})
+
+	const dialog = page.getByRole('dialog', { name: 'Focus trap test modal' })
+	await expect(dialog).toBeVisible()
+
+	const testButton = dialog.getByRole('button', { name: 'Test Button' })
+	const closeButton = dialog.getByRole('button', { name: 'Close' })
+
+	await testButton.focus()
```
Antreesy: testButton is the only element in the modal and should receive the initial focus anyway. So this line isn't needed and test passes as well

```suggestion
```

Antreesy (approved):
Otherwise good. Although I'm also not seeing a point in keeping onMounted hook

tests/component/components/NcModal.spec.ts
nikhil2297: Actually that's right. I will make the suggested change and  push the code again.
@susnux @Antreesy

src/components/NcModal/NcModal.vue
```diff
@@ -368,6 +368,11 @@ async function useFocusTrap() {
 	// wait until all children are mounted and available in the DOM before focusTrap can be added
 	await nextTick()
 
+	// ensuring the mask element is available before creating the focus trap
+	if (!maskElement.value) {
+		return
+	}
+
```
ShGKme: I don't understand how this can be a fix.

The change does not "ensure the mask element is available before creating the focus trap". It completely ignores `useFocusTrap` in case `maskElement` is not available without any "creating the focus trap".

It looks to me that either there is an invalid `useFocusTrap` call that we need to remove, or we need to actually ensure maskElement is available and properly wait for it instead of ignoring.

If I'm wrong in both assumptions, could you correct the comment to add a note about when this is an expected case and why this is valid to stop the function?

ShGKme (changes requested):

src/components/NcModal/NcModal.vue
nikhil2297: `useFocusTrap` is also called at `@after-enter` because that is when the transition is complete and the elements are fully visible and focusable. The `null` check for `maskElement` is a guard to prevent `useFocusTrap` from initializing on a `null` reference during the initial mount.
I’ve kept both calls to ensure the trap initializes correctly regardless of whether the component is shown immediately or via a transition.

Antreesy: I think the point is, there should be a deferred initialisation instead of early return?

Antreesy: I think the point is, there should be a deferred initialisation instead of early return?

nikhil2297: Okay, Got it!. I'll add a watcher for `maskElement`  so the focus trap is initializes as soon as it's available.

nikhil2297:
This change doesn’t skip focus trap initialization if maskElement isn’t available it defers it until the element is present.

**Changes:**

- Removed the `onMounted` call.
- Added a watcher to initialize the focus trap when maskElement becomes available
- Kept the null check for maskElement to handle cases where NcModal is used with v-if

src/components/NcModal/NcModal.vue
```diff
@@ -196,8 +196,16 @@ const maskElement = useTemplateRef('mask')
...
 let focusTrap: FocusTrap | undefined
-onMounted(() => useFocusTrap())
+
 onUnmounted(() => clearFocusTrap())
+
+// Watch for mask element to be available
+watch(maskElement, (el) => {
+	if (el) {
+		useFocusTrap()
+	}
```
susnux: Do we need to clear?


```suggestion
	} else {
		clearFocusTrap()
	}
```

nikhil2297: Cleanup is handle by Vue lifecycle (`onUnmounted()`) and the `@before-leave` transition. So I didn't add a `clearFocusTrap()` call in the watcher.

Sector6759:
Would this fix https://github.com/nextcloud-libraries/nextcloud-vue/issues/7946?

susnux (approved):

susnux:
For @ShGKme I pushed one commit that solves the review comment.
Meaning focus trap is managed by the transition events, no need to custom safe guards or lifecycle events.
Except for unmounting, we need to enforce its removed on unmount.

ShGKme (approved):

Sector6759:
> Would this fix #7946?

This did indeed fix the issue.

PR #8118, opened 2026-01-24
feat(NcRadioGroupButton): add disabled state styling to NcRadioGroupButton 

nikhil2297:
- resolves : https://github.com/nextcloud-libraries/nextcloud-vue/issues/7930

### ☑️ Resolves

- Adds proper disabled state styling to NcRadioGroupButton
- Ensures checked+disabled state is visually indicated

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 Tasks

- [x] Add .radioGroupButton_disabled class and styles

### 🏁 Checklist

### DIFF
--- a/src/components/NcRadioGroup/NcRadioGroup.vue
+++ b/src/components/NcRadioGroup/NcRadioGroup.vue
@@ -159,7 +159,7 @@ The radio group also allows to create a button like styling together with the `N
 		<h4>With icons</h4>
 		<div style="max-width: 250px">
 			<NcRadioGroup v-model="alignment" label="Example alignment with icons" hide-label>
-				<NcRadioGroupButton aria-label="Start" value="start" disabled="true">
+				<NcRadioGroupButton aria-label="Start" value="start" disabled>
 					<template #icon>
 						<NcIconSvgWrapper directional :path="mdiAlignHorizontalLeft" />
 					</template>
--- a/src/components/NcRadioGroupButton/NcRadioGroupButton.vue
+++ b/src/components/NcRadioGroupButton/NcRadioGroupButton.vue
@@ -55,6 +55,7 @@ function onUpdate() {
 	if (props.disabled) {
 		return
 	}
+
 	radioGroup!.value.onUpdate(props.value)
 }
 </script>
@@ -80,7 +81,7 @@ function onUpdate() {
 			class="hidden-visually"
 			:checked="isChecked"
 			type="radio"
-			:disabled="disabled"
+			:disabled
 			:value
 			@input="onUpdate">
 	</div>
@@ -123,7 +124,7 @@ function onUpdate() {
 		padding-inline-start: var(--radio-group-button--padding);
 	}
 
-	&:hover {
+	&:hover:not(.radioGroupButton_disabled) {
 		background-color: var(--radio-group-button--background-color-hover);
 	}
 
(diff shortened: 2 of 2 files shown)


src/components/NcRadioGroup/NcRadioGroup.vue
```diff
@@ -159,7 +159,7 @@ The radio group also allows to create a button like styling together with the `N
 		<h4>With icons</h4>
 		<div style="max-width: 250px">
 			<NcRadioGroup v-model="alignment" label="Example alignment with icons" hide-label>
-				<NcRadioGroupButton aria-label="Start" value="start">
+				<NcRadioGroupButton aria-label="Start" value="start" disabled="true">
```
ShGKme: **chore:** same as in another PR, true is a string here and not needed.
```suggestion
				<NcRadioGroupButton aria-label="Start" value="start" disabled>
```

src/components/NcRadioGroupButton/NcRadioGroupButton.vue
```diff
@@ -47,6 +52,9 @@ const isChecked = computed(() => radioGroup?.value.modelValue === props.value)
  * Handle updating the current model value
  */
 function onUpdate() {
+	if (props.disabled) {
+		return
+	}
```
ShGKme, line 57: **nitpik:** additional check here is not needed. It should never happen as disabled button is not active.

src/components/NcRadioGroupButton/NcRadioGroupButton.vue
```diff
@@ -71,6 +80,7 @@ function onUpdate() {
 			class="hidden-visually"
 			:checked="isChecked"
 			type="radio"
+			:disabled="disabled"
```
ShGKme: **chore:** Shorthand

```suggestion
			:disabled
```

src/components/NcRadioGroupButton/NcRadioGroupButton.vue
```diff
@@ -135,6 +145,21 @@ function onUpdate() {
...
+
+	// Reset the cursor
+	cursor: default;
+	* {
+		cursor: default;
+	}
+
+	&:hover {
+		background-color: var(--radio-group-button--background-color);
+	}
```
ShGKme: To simplify styles a bit, let's add `:not(radioGroupButton_disabled)` to the exisging `:hover`.

In case more styles will be added to the hover effect, there won't be a risk of forgotten reset here.

src/components/NcRadioGroupButton/NcRadioGroupButton.vue
```diff
@@ -55,6 +63,7 @@ function onUpdate() {
 	<div
 		:class="[{
 			[$style.radioGroupButton_active]: isChecked,
+			[$style.radioGroupButton_disabled]: disabled,
```
ShGKme, line 67: Note for reviewers: we could use`:has(:disabled)` here, but adding a class is cheaper (`:has()` is an expensive selector) and simplified styles.

src/components/NcRadioGroupButton/NcRadioGroupButton.vue
nikhil2297: click event is on a `div`. So that's the reason I'm using this check. The alternative is to disabling the `pointer-events`.

ShGKme: True, will fix layout and a11y in another PR.

nikhil2297:
> Thanks

@ShGKme - Can you please confirm this [comment](https://github.com/nextcloud-libraries/nextcloud-vue/pull/8118#discussion_r2729775990). Thanks

PR #8128, opened 2026-01-28
fix(NcInputField): Helper message `word-break` on Chrome

artonge:
Without it, the message does not break and increase the size of its container.
Couldn't test locally because linking `@nc/vue` fails, but tested manually by changing the CSS in the browser.

| Browser | Before | After |
|--------|--------|--------|
| Firefox | image | image |
| Chrome | Screenshot From 2026-01-28 12-11-21 | image |

### DIFF
--- a/src/components/NcInputField/NcInputField.vue
+++ b/src/components/NcInputField/NcInputField.vue
@@ -494,6 +494,7 @@ function handleInput(event: Event) {
 		display: flex;
 		align-items: center;
 		color: var(--color-text-maxcontrast);
+		word-break: break-all;
 
 		&__icon {
 			margin-inline-end: 8px;


susnux:
>  Couldn't test locally because linking @nc/vue fails, but tested manually by changing the CSS in the browser.

You need to use `npm pack` and then use the package in server (use `"@nextcloud/files": "file:/path/to/nc/vue/package.tgz"` in package.json).

> Helper message word-break on Chrome

It seems like everything working fine in Chrome is broken in FF and vice-versa :sweat_smile:

artonge:
Tested, thanks

src/components/NcInputField/NcInputField.vue
```diff
@@ -494,6 +494,7 @@ function handleInput(event: Event) {
 		display: flex;
 		align-items: center;
 		color: var(--color-text-maxcontrast);
+		word-break: break-all;
```
susnux, line 497: I wonder if `break-word` looks better but not really sure

artonge: It would, but is deprecated.

artonge: https://developer.mozilla.org/fr/docs/Web/CSS/Reference/Properties/word-break#break-word

PR #8130, opened 2026-01-28
feat(NcSelectUsers): export `NcSelectUsersModel` type

susnux:
### ☑️ Resolves

So that you can make use of the type inside apps, otherwise its pretty cumbersome to get it.
`type IUserData = InstanceType<typeof NcSelectUsers>['$props']['options'][number]`

### 🏁 Checklist

### DIFF
--- a/src/components/NcSelectUsers/index.ts
+++ b/src/components/NcSelectUsers/index.ts
@@ -3,4 +3,6 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
+export type { NcSelectUsersModel } from './NcSelectUsers.vue'
+
 export { default } from './NcSelectUsers.vue'
--- a/src/components/NcSelectUsers/NcSelectUsers.vue
+++ b/src/components/NcSelectUsers/NcSelectUsers.vue
@@ -169,7 +169,7 @@ import { ref, watch } from 'vue'
 import NcListItemIcon from '../NcListItemIcon/index.js'
 import NcSelect from '../NcSelect/index.js'
 
-export interface IUserData {
+export interface NcSelectUsersModel {
 	id: string
 
 	/**
@@ -219,7 +219,7 @@ export interface IUserData {
  * If the `multiple` property is set then an array of users is emitted,
  * otherwise a single user data object will be emitted.
  */
-const modelValue = defineModel<IUserData | IUserData[]>('modelValue')
+const modelValue = defineModel<NcSelectUsersModel | NcSelectUsersModel[]>('modelValue')
 
 defineProps<{
 	/**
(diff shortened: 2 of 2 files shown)


ShGKme:
It is an internal component type that becomes package exported.

```ts
// Imported like a general package type
// Being a single component specific type
import type { IUserData } from '@nextcloud/vue'
```

For a new usage scope, I'd rename it to something to be component-specific.
Like `NcSelectUsersOptionProp` or `NcSelectUsersUser` or `NcSelectUsersModelValue`.

ShGKme:
Renamed to be a feature (it adds public API, reverting it is a breaking change).

susnux:
> It is an internal component type that becomes package exported.

Well otherwise you can not really work with the component or have to chisel code like in 2011 meaning hope to get the type right.

> `import type { IUserData } from '@nextcloud/vue'`

I would personally never export component types from the main entry point, but only from individual component exports.
This makes scoping much cleaner.

> NcSelectUsersModelValue

Makes sense to have consistent naming

> Renamed to be a feature (it adds public API, reverting it is a breaking change).

Well if we think so - IMHO its a bug because it was intended to be published (even reverting bugfixes can be a breaking change).

ShGKme:
> Well if we think so - IMHO its a bug because it was intended to be published

Even if it was intended, it wasn't a part of public API in the end, and adding it is a change of public API. https://semver.org/#spec-item-7

ShGKme:
> Well otherwise you can not really work with the component or have to chisel code like in 2011 meaning hope to get the type right.

I didn't say it must not be exported. Only that it wasn't. While a type is an internal type, it can be renamed any moment.

Unlike the exported type from the package, that can only be renamed with backward compatibility and should not conflict.

ShGKme:
> I would personally never export component types from the main entry point, but only from individual component exports.
> This makes scoping much cleaner.

Indeed, we don't export any component type from the root package export.
Though, it seems weird that we do it for all the other things.

But even if we don't export, a very generic name makes it more conflict-prune and less explicit in the usage. It is already a problem in other libs

susnux:
@ShGKme resolved by renaming the interface

src/components/NcSelectUsers/index.ts
```diff
@@ -3,4 +3,6 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
+export type { NcSelectUserModel } from './NcSelectUsers.vue'
```
ShGKme: I do apologize, but the component is SelectUser**S** 🫥

```suggestion
export type { NcSelectUsersModel } from './NcSelectUsers.vue'
```

susnux: :+1:

ShGKme (approved):

PR #8342, opened 2026-03-23
fix(NcSelect): announce selected option to screen readers

pringelmann:
### ☑️ Resolves

- Fix #8322

### 🖼️ Screenshots

N/A - behavioural fix, no visual change

### 🚧 Tasks

- [x] Add `aria-describedby` on the search input pointing to a hidden span with the selected value. Screen readers announce it on focus.
- [x] Inject a visually-hidden `<item>, selected` marker into the `#option` slot for the currently selected option. Screen readers pick this up via `aria-activedescendant` while navigating the list.

### 🏁 Checklist

> **Note on `aria-selected`**:
>
> vue-select sets `aria-selected` on the `<li>` elements in its own render function - NcSelect can't control these attributes from the wrapper level without fragile DOM manipulation that gets overwritten on re-renders. The visually-hidden text approach works within Vue's reactivity and doesn't fight the upstream library.

### DIFF
--- a/src/components/NcSelect/NcSelect.vue
+++ b/src/components/NcSelect/NcSelect.vue
@@ -331,18 +331,21 @@ export default {
 		v-bind="propsToForward"
 		@search="search = $event"
 		@update:modelValue="$emit('update:modelValue', $event)">
-		<template v-if="!labelOutside && inputLabel" #header>
+		<template #header>
 			<label
+				v-if="!labelOutside && inputLabel"
 				:for="inputId"
 				class="select__label">
 				{{ inputLabel }}
 			</label>
+			<span :id="selectedDescriptionId" class="hidden-visually">{{ selectedDescription }}</span>
 		</template>
 		<template #search="{ attributes, events }">
 			<input
 				class="vs__search"
 				:class="[inputClass]"
 				v-bind="attributes"
+				:aria-describedby="selectedDescriptionId"
 				:required="inputRequired"
 				dir="auto"
 				v-on="events">
@@ -364,6 +367,8 @@ export default {
 					:name="String(option[localLabel])"
 					:search="search" />
 			</slot>
+			<!-- Visually hidden marker so screen readers can identify the selected option -->
+			<span v-if="isOptionCurrentlySelected(option)" class="hidden-visually">{{ t(', selected') }}</span>
 		</template>
 		<template #selected-option="selectedOption">
 			<!-- @slot Customize how a selected option is rendered -->
@@ -778,10 +783,12 @@ export default {
 		const clickableArea = Number.parseInt(window.getComputedStyle(document.body).getPropertyValue('--default-clickable-area'))
 		const gridBaseLine = Number.parseInt(window.getComputedStyle(document.body).getPropertyValue('--default-grid-baseline'))
 		const avatarSize = clickableArea - 2 * gridBaseLine
+		const selectedDescriptionId = createElementId()
 
 		return {
 			avatarSize,
 			isLegacy,
+			selectedDescriptionId,
 		}
 	},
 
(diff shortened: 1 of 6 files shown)


susnux (approved):
Seems to work 👍 
But I wonder if this should directly implemented in vue-select?
https://github.com/nextcloud-libraries/vue-select/

pringelmann:
> Seems to work 👍 But I wonder if this should directly implemented in vue-select? https://github.com/nextcloud-libraries/vue-select/

Good question: the aria-selected bug in `vue-select` (it uses typeAheadPointer instead of actual selection state) is worth filing upstream - happy to open an issue there. That said, the visually-hidden ", selected" text needs to live here regardless, since it goes through the translation system.

src/components/NcSelect/NcSelect.vue
```diff
@@ -364,6 +367,8 @@ export default {
 					:name="String(option[localLabel])"
 					:search="search" />
 			</slot>
+			<!-- Visually hidden marker so screen readers can identify the selected option -->
+			<span v-if="isOptionCurrentlySelected(option)" class="hidden-visually">{{ t(', selected') }}</span>
```
skjnldsv: We probably need a better way for translations, as this won't work for all languages. You should avoid concatenating two translated strings, the lang could be RTL or LTR 😉

pringelmann:
> Seems to work 👍 But I wonder if this should directly implemented in vue-select? https://github.com/nextcloud-libraries/vue-select/

Turns out this was already fixed in the fork, but seems we can't use that here since it still targets Vue 2

src/components/NcSelect/NcSelect.vue
pringelmann: Good catch - fixed. The hidden span now contains the full string as a single translatable unit (`{option} (selected)`) rather than concatenating two separate pieces. `NcEllipsisedOption` is also hidden from screen readers when selected so only the complete string is announced.

src/components/NcSelect/NcSelect.vue
```diff
@@ -362,8 +365,13 @@ export default {
 			<slot name="option" v-bind="option">
 				<NcEllipsisedOption
 					:name="String(option[localLabel])"
-					:search="search" />
+					:search="search"
+					:aria-hidden="isOptionCurrentlySelected(option) || undefined" />
```
susnux: Why not use native markup like `aria-selected` and `role="option"`?

pringelmann: `vue-select` already puts role="option" and aria-selected on the `<li>` elements, but sets `aria-selected` based on keyboard focus position (typeAheadPointer) rather than actual selection - that's the root bug. The `#option` slot we use only controls the content inside each `<li>`, not the `<li>` attributes, so we can't override `aria-selected` from here without DOM manipulation that gets stomped on re-render.

This is already solved in our fork of `vue-select`, but we can't use that until it is migrated to Vue 3.

susnux: > but sets aria-selected based on keyboard focus position (typeAheadPointer) rather than actual selection - that's the root bug

Why then not fix it in vue-select? We did so in the past too, thats why we forked it:
https://github.com/nextcloud-libraries/vue-select/

susnux: > This is already solved in our fork of vue-select, but we can't use that until it is migrated to Vue 3.

Ah wait too fast... :see_no_evil: 

Then we should push forward the migration instead of tinker with our usage of it, as we will fail the other accessibility issues as well (those we fixed in the fork).
https://github.com/nextcloud-libraries/vue-select/pull/25

pringelmann: I will review effort required to bring it to completion

src/components/NcSelect/NcSelect.vue
skjnldsv: Nice! 
You could even go one step further and make it a full sentence like `option "{option} " is selected`? Not sure it makes sense in that context, just a thought

pringelmann: Good suggestion, although in my experience this is a more common screen reader pattern. (since it is less verbose, which can get annoying for users who frequently submit big forms)

src/components/NcSelect/NcSelect.vue
pringelmann: Not quite a drop-in replacement, but is doable. Will revert the fix and switch to `nextcloud/vue-select`

susnux (approved):

pringelmann:
Closing this - superseded by main's switch to `@nextcloud/vue-select`:

- 6c5a91030 `feat(NcSelect): use @nextcloud/vue-select 4.0.0`
- 164f2a903 `chore(deps): bump @nextcloud/vue-select from 4.0.0 to 4.1.0`

`@nextcloud/vue-select` already handles `aria-selected` correctly on the option `<li>` elements (the bug that #8322 reported in the upstream `vue-select`), so the wrapper-level workaround in this PR is no longer needed.

PR #8397, opened 2026-04-07
fix(NcRichText): do not render invalid relative markdown links

Antreesy:
#### fix(NcRichText)!: render all non-resolved links as external
- this covers:
	- external links (as before)
	- internal absolute links, but from another app (as before)
	- and internal relative links, that do not have a router match for current app (new)
		- possibly a breaking change?
		- this is likely an invalid user input, it shouldn't unload the page and destroy an app

What would be rendered for component `<NcRichText :autolink="true" :useMarkdown="true" />`:
- `https://autolink.me` - `https://autolink.me`
- `[hello](/world)` - `hello` with href: '/world'
- `[hello](https://nextcloud.com)` - `hello` with href: 'https://nextcloud.com'
- `[hello](tel:+49123456789)` - `hello` with href: 'tel:+49123456789'
- `[hello](mailto:+49123456789)` - `hello` with href: 'mailto:+49123456789'
- `[hello](other:proto)` - `hello` as plain text
- `[hello](world)` - `hello` as plain text
- `[hello](?parameters=1)` - `hello` as plain text
- `[hello](#anchor)` - `hello` as plain text

### 🚧 Tasks

- [ ] Consult other teams for using of relative links, whether it's breaking for them

### 🏁 Checklist

### DIFF
--- a/src/components/NcRichText/NcRichText.vue
+++ b/src/components/NcRichText/NcRichText.vue
@@ -532,11 +532,9 @@ export default {
 						return entry
 					}
 					const { component, props } = entry
-					// do not override class of NcLink
-					const componentClass = component.name === 'NcLink' ? undefined : 'rich-text--component'
 					return h(component, {
 						...props,
-						class: componentClass,
+						class: 'rich-text--component',
 					})
 				})
 			}
@@ -683,13 +681,6 @@ export default {
 
 	.rich-text--fallback, .rich-text-component {
 		display: inline;
-	}
-
-	.rich-text--external-link {
-		text-decoration: underline;
-		&:after {
-			content: ' ↗';
-		}
 	}
 }
 
--- a/src/components/NcRichText/NcRichText.vue
+++ b/src/components/NcRichText/NcRichText.vue
@@ -320,6 +320,7 @@ import { RouterLink } from 'vue-router'
 import NcCheckboxRadioSwitch from '../NcCheckboxRadioSwitch/NcCheckboxRadioSwitch.vue'
 import NcReferenceList from './NcReferenceList.vue'
 import NcRichTextCopyButton from './NcRichTextCopyButton.vue'
+import NcRichTextExternalLink from './NcRichTextExternalLink.vue'
 import { createElementId } from '../../utils/createElementId.ts'
 import { getRoute, parseUrl, remarkAutolink } from './autolink.ts'
 import { remarkPlaceholder } from './remarkPlaceholder.ts'
@@ -630,6 +631,7 @@ export default {
 				if (String(type) === 'a') {
 					const route = getRoute(this.$router, props.href)
 					if (route) {
+						// Resolved link to this app; render RouterLink
 						delete props.href
 						delete props.target
 
@@ -637,6 +639,18 @@ export default {
 							...props,
 							to: route,
 						}, { default: () => children })
+					}
+
+					const isAllowedScheme = /^(https?:\/\/|tel:|mailto:)/.test(props.href)
+					if (isAllowedScheme) {
+						// External link; render normally, open in the new tab
+						props.href = props.href.trim()
+						return h(NcRichTextExternalLink, props, children)
+					} else {
+						// Unresolved relative link that does not belong to this app; render only children
+						delete props.href
+						delete props.target
+						return h('span', props, children)
 					}
 				}
 				return h(type, props, children)
--- a/src/components/NcRichText/NcRichTextExternalLink.vue
+++ b/src/components/NcRichText/NcRichTextExternalLink.vue
@@ -1,0 +1,41 @@
+<!--
+  - SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<script setup lang="ts">
+const { href, decorateExternal = false } = defineProps<{
+	/** Link attribute to render */
+	href: string
+	/** Whether to add the appended arrow decoration */
+	decorateExternal?: boolean
+}>()
+</script>
+
+<template>
+	<a
+		:href="href"
+		rel="noopener noreferrer"
+		target="_blank"
+		:class="[$style.externalLink, {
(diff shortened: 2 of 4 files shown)


src/components/NcRichText/NcRichTextExternalLink.vue
```diff
@@ -0,0 +1,39 @@
...
+		target="_blank"
+		class="rich-text--external-link"
+		:class="{ 'no-ext-decoration': noExtDecoration }">
+		<slot name="default">
+			{{ href.trim() }}
+		</slot>
+	</a>
+</template>
+
+<style scoped lang="scss">
```
susnux, line 28: maybe directly use css modules?

susnux (approved):
otherwise fine :)

src/components/NcRichText/autolink.ts
```diff
@@ -46,6 +46,11 @@ export function remarkAutolink({ autolink, useMarkdown, useExtendedMarkdown }) {
 		}
 
 		visit(tree, (node) => node.type === 'text', (node, index, parent) => {
+			// Do not autolink text already inside a link node
+			if (parent?.type === 'link' || parent?.type === 'linkReference') {
+				return
+			}
```
ShGKme, line 52: Can it be a non-direct parent? Like `[This is a **bold http://... link**](href)`

src/components/NcRichText/NcRichTextExternalLink.vue
```diff
@@ -0,0 +1,39 @@
...
+<template>
+	<a
+		:href="href"
+		rel="noopener noreferrer"
+		target="_blank"
+		class="rich-text--external-link"
+		:class="{ 'no-ext-decoration': noExtDecoration }">
+		<slot name="default">
+			{{ href.trim() }}
+		</slot>
```
ShGKme, line 24: Let's trim it from the parent to not trim each re-render.

src/components/NcRichText/NcRichTextExternalLink.vue
```diff
@@ -0,0 +1,39 @@
...
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<script setup lang="ts">
+const { href, noExtDecoration = true } = defineProps<{
+	/** Link attribute to render */
+	href: string
+	/** Whether to remove the appended arrow decoration */
+	noExtDecoration?: boolean
+}>()
```
ShGKme, line 12: Though it is not a public component, let's keep boolean props with `false` default.
E.g. `decorateExternal = false`.

src/components/NcRichText/autolink.ts
ShGKme: Yep, it breaks with additional formating:
image

src/components/NcRichText/NcRichTextExternalLink.vue
ShGKme: If I understand it correctly, this component is never used with this prop (with a non-default value)

Antreesy: Weird, I remember that a default was supposed to be false, so it appears as before in autolink.ts and in NcRichText

ShGKme:
> #### fix(NcRichText)!: render all non-resolved links as external
> * this covers:
>   
>   * external links (as before)
>   * internal absolute links, but from another app (as before)
>   * and internal relative links, that do not have a router match for current app (new)
>     
>     * possibly a breaking change?
>     * this is likely an invalid user input, it shouldn't unload the page and destroy an app

I'd even consider forbidding such links.

F relative link can even be a link to a completely different service if Nextcloud is hosted on a non-empty base URL.

src/components/NcRichText/autolink.ts
Antreesy: Commit and additional changes extracted to https://github.com/nextcloud-libraries/nextcloud-vue/pull/8449

src/components/NcRichText/NcRichTextExternalLink.vue
Antreesy: reworked in fixups. prop is used with `autolink: true, useMarkdown: false`, to copy original behaviour

Antreesy:
> I'd even consider forbidding such links.
> F relative link can even be a link to a completely different service if Nextcloud is hosted on a non-empty base URL.

See last commit for the change 👇🏽

ShGKme (approved):

src/components/NcRichText/NcRichTextExternalLink.vue
```diff
@@ -0,0 +1,41 @@
...
+		</slot>
+	</a>
+</template>
+
+<style module lang="scss">
+.externalLink {
+	text-decoration: underline;
+}
+
+.externalLink_decorated::after {
```
DorraJaouad, line 34:
```suggestion
.externalLink--decorated::after {
```

src/components/NcRichText/NcRichText.vue
```diff
@@ -640,6 +640,18 @@ export default {
 							to: route,
 						}, { default: () => children })
 					}
+
+					const isAllowedScheme = /^(https?:\/\/|tel:|mailto:)/.test(props.href)
+					if (isAllowedScheme) {
+						// External link; render normally, open in the new tab
+						props.href = props.href.trim()
+						return h(NcRichTextExternalLink, props, children)
+					} else {
```
DorraJaouad, line 649: Image

src/components/NcRichText/NcRichTextExternalLink.vue
Antreesy: it needs a snake_case with scss modules, otherwise we'd need this in template:
```diff
- [$style.externalLink_decorated]: decorateExternal,
+ [$style.['externalLink--decorated']]: decorateExternal,
```

src/components/NcRichText/NcRichText.vue
Antreesy: added a regression tests. IMO it should not be supported for SPA applications, that do not have section
if vue-router doesn't match it, component shouldn't allow it either

DorraJaouad (approved):
All good then

PR #8440, opened 2026-04-17
fix(NcAvatar): redirect to guest avatar if avatar doesnt exist

skjnldsv:
### ☑️ Resolves

Avoid this kind of stuff if an avatar is not generated/accessible 

| Before | After |
|:---------:|:------:|
|image|image|

### 🏁 Checklist


ShGKme (approved):
Nice.

We should also add this option to the library, probably, and use the library here: https://github.com/nextcloud-libraries/nextcloud-router/blob/main/lib/avatar.ts

ShGKme:
Maybe we should also handle the error to render a local avatar, when it is not available at all?

susnux (approved):

PR #8443, opened 2026-04-18
feat(NcDateTimePicker): manual date input

### LINKED ISSUE #8635: [NcDateTimePicker]: text input gets discarded since v9
Text input in `NcDateTimePicker` gets discarded.
The date/time can only be edited through the popover elements.

### Related

* Bug report in calendar app https://github.com/nextcloud/calendar/issues/8307
* Attempt to fix it https://github.com/nextcloud-libraries/nextcloud-vue/pull/8443

### Reproduce

* Open Styleguidist
* Go to  `NcDateTimePicker` > General example
  * https://nextcloud-vue-components.netlify.app/#/Components/NcPickers?id=ncdatetimepicker
* Change input `10 Oct 2022` to `10 Oct 2023`
* Click outside of input to submit
* Expected: Date change stays `10 Oct 2023`
* Actual: Date changes back to `10 Oct 2022`

SebastianKrupinski:
### ☑️ Resolves

Fixes #8635

- Fix manual date input
- Re-parses manual date input if VueDatePicker fails
- Submits manual input on enter, tab or blur

### 🖼️ Screenshots

https://github.com/user-attachments/assets/089b84b5-6322-42a7-8c0c-7daefcb85886

### 🏁 Checklist

### DIFF
--- a/src/components/NcDateTimePicker/NcDateTimePicker.vue
+++ b/src/components/NcDateTimePicker/NcDateTimePicker.vue
@@ -288,7 +288,7 @@ import {
 	getFirstDay,
 } from '@nextcloud/l10n'
 import VueDatePicker from '@vuepic/vue-datepicker'
-import { computed, useTemplateRef } from 'vue'
+import { computed, ref, useTemplateRef } from 'vue'
 import NcIconSvgWrapper from '../NcIconSvgWrapper/NcIconSvgWrapper.vue'
 import NcTimezonePicker from '../NcTimezonePicker/NcTimezonePicker.vue'
 import { t } from '../../l10n.ts'
@@ -421,6 +421,13 @@ const props = withDefaults(defineProps<{
 	 * @default false
 	 */
 	inline?: boolean
+
+	/**
+	 * Enable or disable manual text input for the date picker.
+	 *
+	 * @default true
+	 */
+	manualInput?: boolean
 }>(), {
 	ariaLabel: t('Datepicker input'),
 	ariaLabelMenu: t('Datepicker menu'),
@@ -433,6 +440,7 @@ const props = withDefaults(defineProps<{
 	modelValue: null,
 	// set by fallbackPlaceholder
 	placeholder: undefined,
+	manualInput: true,
 	type: 'date',
 	inline: false,
 })
@@ -453,6 +461,7 @@ const emit = defineEmits<{
 
 const targetElement = useTemplateRef('target')
 const pickerInstance = useTemplateRef('picker')
+const manualInputValue = ref<Date | null>(null)
 
 /**
  * Mapping of the model-value prop to the format expected by the library.
@@ -559,12 +568,17 @@ const realFormat = computed<LibraryFormatOptions>(() => {
 	}
 
 	if (formatter) {
-		return (input: Date | [Date, Date]) => Array.isArray(input)
-			? formatter.formatRange(input[0], input[1])
-			: formatter.format(input)
+		return (input: Date | [Date, Date]) => {
+			if (Array.isArray(input)) {
+				return isFiniteDateRange(input)
+					? formatter.formatRange(input[0], input[1])
+					: ''
+			}
+
+			return isFiniteDate(input) ? formatter.format(input) : ''
+		}
 	}
 
-	// fallback to default formatting
 	return undefined
 })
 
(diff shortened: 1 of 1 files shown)


src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -421,6 +421,13 @@ const props = withDefaults(defineProps<{
 	 * @default false
 	 */
 	inline?: boolean
+
+	/**
+	 * Enable or disable manual text input for the date picker.
+	 *
+	 * @default true
+	 */
+	manualInput?: boolean
```
ShGKme, line 430: We follow [no-boolean-default](https://eslint.vuejs.org/rules/no-boolean-default.html) props as more natural for HTML-styled template, allowing using boolean shortcut like in HTML.

```html
<!-- default: true -->
<NcDateTimePicker />
<NcDateTimePicker :manualInput="false" />

<!-- default: false -->
<NcDateTimePicker />
<NcDateTimePicker noInput />
```


```suggestion
	/**
	 * Disable date text input
	 */
	noInput?: boolean
```

or

```suggestion
	/**
	 * Disable date text input
	 */
	noTextInput?: boolean
```

ShGKme:
Changing to `feat`, it introduces a new API.

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -789,13 +867,17 @@ function sameDay(a: Date, b: Date): boolean {
...
 			:weekNumName
 			:weekNumbers="showWeekNumber ? { type: 'iso' } : undefined"
 			:weekStart
 			v-bind="pickerType"
 			@update:modelValue="onUpdateModelValue"
-			@blur="emit('blur')">
+			@blur="onBlur"
+			@keydown.enter.capture="onManualInputKeySubmit"
+			@keydown.tab.capture="onManualInputKeySubmit"
+			@textInput="onTextInput"
```
ShGKme, line 879: This breaks accessibility by preventing using the data picker (the menu popup) via keyboard.

Previously, when the focus was in the input field, the menu could be open by typing or pressing <kbd>Space</kbd>, and then moving the focus to the menu by <kbd>Tab</kbd>

Now, <kbd>Space</kbd> acts as text input, and <kbd>Tab</kbd> goes to the next element, without opening the menu.

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -627,6 +641,50 @@ function onUpdateModelValue(value: LibraryModelValue): void {
...
+ * @param parsedDate The library-parsed date or null if parsing failed
+ */
+function onTextInput(event: Event, parsedDate: Date | null): void {
+	if (parsedDate) {
+		manualInputValue.value = parsedDate
+		return
+	}
+
+	const input = event.target instanceof HTMLInputElement ? event.target.value : ''
+	const tempDate = new Date(input)
```
ShGKme, line 657: It works with the date input and date-time input, but it breaks on the week input, the month input, the year input and the range input.

ShGKme (changes requested):
In general, I'm unsure if it makes sense to have manual input without masking the value or explicitly specifying the format to the user.

It is clear when editing the current value, but quite unclear for new values.

What about disabling manual input at all?

SebastianKrupinski:
> In general, I'm unsure if it makes sense to have manual input without masking the value or explicitly specifying the format to the user.

This implementation supports multiple input formats, the one in the video above in not event a standard input format,

You can input,

YYYY-MM-DD

MMM DD YYYY

DD.MM.YYYY

DD/MM/YYYY

So forcing the user to use a specific format is not feasible, also since users are uses to using their localized format.

src/components/NcDateTimePicker/NcDateTimePicker.vue
SebastianKrupinski: I left this as manualInput with a boolean on purpose, the underlying component actually supports:

- False
- True
- Options {
  enterSubmit = true,
  tabSubmit = true,
}

This way the options can be implemented later without having multiple props and/or breaking functionality.

src/components/NcDateTimePicker/NcDateTimePicker.vue
SebastianKrupinski: Oh, will look in to it.

src/components/NcDateTimePicker/NcDateTimePicker.vue
SebastianKrupinski: If a user can type in a date, this is much easier then trying to navigate a popup menu with a keyboard, so I don't seem this as a problem. 

Alternatively we could disable the tab capture and a user can tab in to the menu.

I was actually thinking of disabling the popup menu on manual input, but thought is was a good visual feed back that the correct date is selected.

ShGKme: But not if the input value is changed significantly.

I enter `10/22/2026`. Then I want to edit it changing the month, for example, 10 to 11. I go back and get `Oct 22, 2026`, where I actually need to change `Oct` to `Nov`.

ShGKme:
> So forcing the user to use a specific format is not feasible, also since users are uses to using their localized format.

The first problem is specifying. Without knowing the supported format, a user just doesn't know in what format they should enter the data. 

> ...in not event a standard input format

Which makes it even harder if a user didn't learn the "JS Date parser" format.
(Just leaving it here: https://jsdate.wtf)

> You can input ...

> ... DD.MM.YYYY

I cannot, this is an `Invalid date` in JS.

image

> DD/MM/YYYY

But it will be parsed as `MM/DD/YYYY`.
If I enter `02/03/2026` I get February 3rd, not March 2nd.

image

src/components/NcDateTimePicker/NcDateTimePicker.vue
susnux: I agree with Grigorii here, we should stick with no boolean default rather than some underlying library.
In general we should not care about the library we wrap as we should define an interface and translate this to the library and not the otherway around.

kra-mo:
From the design side, it is tricky. Being able to manually enter a date is really valuable, but indeed, https://jsdate.wtf.

But I've also seen this done right. There are only so many possible date formats out there, and they are documented. People have wrote parsers before that cover 99.5% of what users will input, aside from typos. I'd like that.

But it's also fine to just have a placeholder in ISO 8601 for new values, like just the string `YYYY-MM-DD`, then the user can be expected to follow a format they see.

odzhychko:
https://github.com/nextcloud-libraries/nextcloud-vue/pull/8657 is a different approach for parsing, which always uses a know format (the same format as used for the preview).

This addresses issues from https://github.com/nextcloud-libraries/nextcloud-vue/pull/8443#issuecomment-4276586550

But has the drawback of changing the preview format to in some cases (more in the description of #8657)

SebastianKrupinski:
@odzhychko can this be closed then?

odzhychko:
> @odzhychko can this be closed then?

Yes, closing in favor of https://github.com/nextcloud-libraries/nextcloud-vue/pull/8657

PR #8447, opened 2026-04-20
feat(NcAppSidebarTab): Redesign active tab as rounded filled pill

### LINKED ISSUE #7520: Redesign sidebar tabs
Mostly just taken from https://github.com/nextcloud-libraries/nextcloud-vue/pull/7516#pullrequestreview-3228713318

The selected item should stay highlighted for consistency with the [list selection changes](https://github.com/nextcloud-libraries/nextcloud-vue/issues/7222).

The bottom edge should get the same border radius as the top one. As a result, the blue line indicating highlight should be rounded and less wide.

The active item changing to be bold is quite distracting to me. I feel like we could drop that behavior as we don't do it everywhere for selection either way.

A nitpick:
When hovering, this thing that happens because of the border radius looks weird:

Image
 
There should be at least a 2px gap here to avoid this. The container itself could also be more rounded.

Here is how it should look like with these changes:

Image

It would also be cool if the blue line could do a little animation, growing in width when selected instead of just suddenly appearing, like so:

https://github.com/user-attachments/assets/5a178c9e-b126-4e9d-a132-236b09d65a94

nfebe:
### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
active-tabs-before-7520 | active-tabs-after-7520

### Screencast

[active-tabs-demo-7520.webm](https://github.com/user-attachments/assets/8c65e19c-0356-422d-8268-3c628cc2df0a)

Closes #7520

### DIFF
--- a/src/components/NcAppSidebar/NcAppSidebarTabs.vue
+++ b/src/components/NcAppSidebar/NcAppSidebarTabs.vue
@@ -13,6 +13,7 @@
 			v-if="hasMultipleTabs || showForSingleTab"
 			role="tablist"
 			class="app-sidebar-tabs__nav"
+			:class="{ 'app-sidebar-tabs__nav--legacy': isLegacy34 }"
 			@keydown.left.exact.prevent.stop="focusPreviousTab"
 			@keydown.right.exact.prevent.stop="focusNextTab"
 			@keydown.tab.exact.prevent.stop="focusActiveTabContent"
@@ -44,6 +45,7 @@
 <script>
 import { getCanonicalLocale } from '@nextcloud/l10n'
 import NcAppSidebarTabsButton from './NcAppSidebarTabsButton.vue'
+import { isLegacy34 } from '../../utils/legacy.ts'
 
 export default {
 	name: 'NcAppSidebarTabs',
@@ -93,6 +95,7 @@ export default {
 			 * Local active (open) tab's ID. It allows to use component without v-model:active
 			 */
 			activeTab: props.active,
+			isLegacy34,
 		}
 	},
 
@@ -255,6 +258,11 @@ export default {
 		justify-content: stretch;
 		margin: 10px 8px 0 8px;
 		border-bottom: 1px solid var(--color-border);
+
+		&:not(&--legacy) {
+			gap: var(--default-grid-baseline);
+			padding-block-end: var(--default-grid-baseline);
+		}
 	}
 
 	&__tab {
--- a/src/components/NcAppSidebar/NcAppSidebarTabsButton.vue
+++ b/src/components/NcAppSidebar/NcAppSidebarTabsButton.vue
@@ -7,6 +7,7 @@
 import type NcAppSidebarTab from '../NcAppSidebarTab/NcAppSidebarTab.vue'
 
 import NcVNodes from '../NcVNodes/NcVNodes.vue'
+import { isLegacy34 } from '../../utils/legacy.ts'
 
 const selected = defineModel<boolean>('selected', { required: true })
 
@@ -23,6 +24,7 @@ defineProps<{
 		class="button-vue"
 		:class="[$style.sidebarTabsButton, {
 			[$style.sidebarTabsButton_selected]: selected,
+			[$style.sidebarTabsButton_legacy]: isLegacy34,
 		}]"
 		role="tab"
 		:aria-selected="selected"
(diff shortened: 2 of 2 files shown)


src/components/NcAppSidebarTab/NcAppSidebarTab.vue
```diff
@@ -114,10 +114,11 @@ export default {
 		/**
 		 * Render tab's icon slot if any
 		 *
+		 * @param {boolean} selected Whether the tab is the active one
 		 * @return {import('vue').VNode[]}
 		 */
-		renderIcon() {
-			return this.$slots.icon?.()
+		renderIcon(selected = false) {
+			return this.$slots.icon?.({ selected })
```
ShGKme: We have few components with scoped slots. Might worth an updated example in the docs.

And we might need to adjust Files API for sidebars later to support this feature in Files.

https://github.com/nextcloud-libraries/nextcloud-files/blob/main/lib/ui/sidebar/SidebarTab.ts#L64-L67

src/components/NcAppSidebar/NcAppSidebarTabsButton.vue
```diff
@@ -41,29 +41,41 @@ defineProps<{
...
+		content: '';
+		position: absolute;
+		bottom: 2px;
+		left: 50%;
+		width: 0;
+		height: 4px;
+		border-radius: 999px;
+		background-color: var(--color-primary-element);
+		transform: translateX(-50%);
+		transition: width var(--animation-quick);
```
ShGKme: On the design animation also has the opacity change from 0 to 1.

src/components/NcAppSidebar/NcAppSidebarTabsButton.vue
```diff
@@ -41,29 +41,41 @@ defineProps<{
 
 <style module lang="scss">
 .sidebarTabsButton {
+	position: relative;
 	border: none;
-	border-bottom: var(--default-grid-baseline) solid transparent !important;
-	border-radius: var(--border-radius-small);
-	background-color: var(--color-main-background);
+	border-radius: var(--border-radius-element);
+	background-color: transparent;
```
ShGKme: Why change `background-color: var(--color-main-background);` to `transparent`?

src/components/NcAppSidebar/NcAppSidebarTabsButton.vue
```diff
@@ -72,14 +84,15 @@ defineProps<{
...
 .sidebarTabsButton_selected {
-	border-bottom-color: var(--color-primary-element) !important;
-	border-bottom-left-radius: 0px;
-	border-bottom-right-radius: 0px;
+	background-color: var(--color-background-hover);
 	cursor: default;
 
+	&::after {
+		width: 80%;
+	}
```
ShGKme, line 128: If I'm not mistaken, from the draft it looks more like `65%`

susnux (commented):
What about backwards compatibility with old servers?
This likely needs legacy workarounds just like we did with the border rework we recently did. So that apps on e.g. NC32 still have the old design.

ShGKme:
(I removed the icon and added font-weight to make it closer to the draft on the screenshot)

Design | Current
-- | --
image | image

- It looks like the highlight is supposed to be on the bottom of the button, without the `2px` gap
- The highlight width looks more like `65%` rather than the current `80%`
- The highlight height looks more like `5px` (or more) rather than the current `4px`. With 4px the border radius is `2px`, which is barely a circle section (but might be different due to a different size)

Another thing that catches my eyes — how it lies on the gray line between the tablist and the tab content. Previously it looked fine to me, but now with the border radius and the gap if feels like the gap is missing between the buttons and the content, or not necessary at all.

Before | image
--- | ---
After | image

Also, I'm not sure what it is supposed to look like with the icon. On the draft there are no icons. With the current size, it looks unbalanced to me, with the icon too close to the top.

image

cc @kra-mo

src/components/NcAppSidebar/NcAppSidebarTabsButton.vue
nfebe: > The highlight width looks more like 65% rather than the current 80%

Yes but the design item's width is about half of what an actual tab looks like.... It should scale down to something similar if there are enough tabs.

ShGKme: But the percent is relative to the tab button size. It scales down in both cases.

nfebe: Screenshot From 2026-04-27 14-45-40

kra-mo:
> Also, I'm not sure what it is supposed to look like with the icon. On the draft there are no icons. With the current size, it looks unbalanced to me, with the icon too close to the top.

Yeah, I see what you mean, some more padding would be nice.

kra-mo:
It would be good to have the `font-weight` just be `500` consistently. See #8469

ShGKme:
@kra-mo What about the other points from above?
- Marker width (80%/50% or ~65%)
- Marker thickness (height, 4px or 5px)
- Marker position (on the bottom or with a gap)
- The border between tabs and the content, or the gap above the border

kra-mo:
> Marker width (80%/50% or ~65%)

I was thinking a bit, and honestly, it being wider works in more cases so maybe it's better to just go with that.

> Marker thickness (height, 4px or 5px)

It does look better if it's a bit thicker, but 4px is a variable, I'd say maybe 6px (1.5x).

> Marker position (on the bottom or with a gap)

On the bottom, no gap.

> The border between tabs and the content, or the gap above the border

Yes, elements with border radius should not hug hard edges. We should add a bit of a gap :)

nfebe:
image

kra-mo (approved):

src/components/NcAppSidebarTab/NcAppSidebarTab.vue
ShGKme: Although the new param's name reflects HTML (`aria-selected`) and the current internal variable name, I'm not sure about the naming:
- Parent's selected tab is defined by `active` prop, which results in 2 names for the same thing (The current tab is `active` and whether the tab is active is `selected`)
- Boolean flag pattern seems more fitting here, e.g. `isSelected/isActive`. And it also would allow in the future having both the boolean flag and the value if needed.

Note: this changes the component API and refactoring it later would be only possible with backward compatibility and migration.

The new prop might not be required for this feature. There is an `active` prop already.

It is slightly more complicated to use, though.

```vue
<NcAppSidebar v-model:active="active">

	<!-- ... -->

	<NcAppSidebarTab name="Settings" id="settings-tab">
		<template #icon>
			<IconCog v-if="active === 'settings-tab'" :size="20" />
			<IconCogOutline v-else :size="20" />
		</template>
	</NcAppSidebarTab>

</NcAppSidebar>
```

cc @susnux

src/components/NcAppSidebar/NcAppSidebarTabs.vue
```diff
@@ -253,7 +253,9 @@ export default {
 	&__nav {
 		display: flex;
 		justify-content: stretch;
+		gap: var(--default-grid-baseline);
 		margin: 10px 8px 0 8px;
+		padding-block-end: var(--default-grid-baseline);
```
ShGKme: This also needs the legacy check

Image

src/components/NcAppSidebarTab/NcAppSidebarTab.vue
susnux: I would stick with `active` for consistency and name it `isActive` as thats the schema its usually named in the ecosystem.

ShGKme: @nfebe Could you adjust the name or remove the slot prop?

susnux (changes requested):
Tested seems to work but please adjust the slot prop or remove

susnux (approved):

PR #8448, opened 2026-04-21
feat: change NcAppNavigationItem/NcListItem active state design and NcAppContent edge style on Nextcloud 34+

### LINKED ISSUE #7222: [NcAppContent + NcAppNavigationItem + NcListItem] Revise styling
The changes here are:

1. Add a slight border between the main app content and the sidebar. The color should be similar to the mockup in intensity. Not sure how to exactly mix a similar one, I'll leave it up to you like the others :)
2. Round the borders of the main app content at the left edge.
3. Change the selection styling (for sidebars in general) to match the mockup. @nimishavijay mentioned that Outlook uses a similar style for unread messages so it looks like the message list in Mail would probably have to be special-cased for the time being to avoid confusion, but other than that, all selected items in sidebars should look similar.
4. It also might be nice to [do an animation](https://github.com/nextcloud-libraries/nextcloud-vue/issues/7222#issuecomment-3144431015), but maybe only as a follow-up, unless it's easy enough.

Image

nfebe:
| Before | After |
|----------- | ------- |
| nc-content-and-nav-before-7222 | nc-content-and-nav-before-after png |

## Screencast

[Screencast From 2026-04-24 20-31-16.webm](https://github.com/user-attachments/assets/78f0083e-5a9c-42f8-b75f-fb9b08129655)

Closes #7222


ShGKme:
Counter in the `NcListItem` is not contrast anymore.

And, if I got the plan correctly, `NcListItem` was supposed to have the same new design for selected (active) items, wasn't it?

image

ShGKme:
What about buttons?

src/components/NcAppContent/NcAppContent.vue
```diff
@@ -466,6 +466,12 @@ export default {
 	&:not(.app-content--has-list) {
 		overflow: auto;
 	}
+
+	@media only screen and (min-width: $breakpoint-mobile) {
+		border-inline-start: 1px solid var(--color-border);
+		border-start-start-radius: var(--body-container-radius);
+		border-end-start-radius: var(--body-container-radius);
+	}
```
ShGKme: The border is visible even on the close navigation, and even when there is no navigation at all.

ShGKme:
It looks a bit different from the design draft. Half of the border is visually outside the element and had half-rounded (harp) edges pointed outside.

Design | Current
--- | ---
image | image

src/components/NcAppContent/NcAppContent.vue
ShGKme: Also, the color is different on the draft

ShGKme:
Do we keep this design on the older servers as well?

nfebe:
Now 
image

src/components/NcListItem/NcListItem.vue
```diff
@@ -854,22 +852,44 @@ export default {
 	&--active,
 	&.active {
 		.list-item {
-			background-color: var(--color-primary-element);
-			color: var(--color-primary-element-text) !important;
+			background-color: color-mix(in srgb, var(--color-primary-element) 16%, transparent);
+			color: var(--color-main-text) !important;
```
susnux: By doing so this is not guaranteed to match accessibility requirements anymore.
Because now foreground color and background color are independent

ShGKme:
> Now image

Now it is rounded, but the border is outside the item while in the design it is inside the item.
(IMO, inside looks better)

src/components/NcAppContent/NcAppContent.vue
ShGKme: > The border is visible even on the close navigation, and even when there is no navigation at all.

This part is now fixed

susnux:
> Now it is rounded, but the border is outside the item while in the design it is inside the item.

From design it should be rounded but within the element not outside.

nfebe:
- Rounded
- Starts inside

image

src/components/NcListItem/NcListItem.vue
nfebe: the new bg is mostly transparent (84% transparent + 16% primary tint), what shows through is `--color-main-background`. So the text we render with `--color-main-text` is still contrasting against  `--color-main-backgroun` mostly which is the designed pair nc guarantees.

jancborchardt (changes requested):
Nice! Some feedback:
- [ ] The down-arrow icon of sections with a dropdown needs to stay text-color (dark by default) when selected. Currently it switches to white, which worked with the dark background color of selected elements we had before, but doesn’t work anymore.
- [ ] The icon of the selected element is not filled. Seems a separate issue as it also doesn’t work on the last version, but it is the case for other Hub apps.

kra-mo:
> The icon of the selected element is not filled. Seems a separate issue as it also doesn’t work on the last version, but it is the case for other Hub apps.

It is, it depends on the app.

nfebe:
@jancborchardt thanks the first item can be resolved.

> The icon of the selected element is not filled. Seems a separate issue as it also doesn’t work on the last version, but it is the case for other Hub apps.

This though I descoped for it to done desperately. We need to introduce a new active icon slot and implement upstream so it will multiple PRs across multiple apps to use the new slot.

kra-mo:
> it will multiple PRs across multiple apps to use the new slot.

It's definitely out of scope for this one.

src/assets/NcAppNavigationItem.scss
```diff
@@ -28,16 +28,42 @@
...
 
-	// When .active class is applied, change color background of link and utils. The
-	// !important prevents the focus state to override the active state.
-	&.active {
+	// !important so :focus-within does not override the active state.
+	// New design (NC34+): glassy tinted bg + inline-start stripe.
+	&:not(.app-navigation-entry--legacy).active {
+		background-color: color-mix(in srgb, var(--color-primary-element) 16%, transparent) !important;
+
+		&:hover {
```
kra-mo, line 36: The hover color should also be updated for non-active elements.

src/assets/NcAppNavigationItem.scss
```diff
@@ -28,16 +28,42 @@
...
+		}
+
+		&:not(.app-navigation-entry--editing) {
+			.app-navigation-entry-link,
+			.app-navigation-entry-button {
+				color: var(--color-main-text) !important;
+			}
+		}
+
+		&:not(.app-navigation-entry--editing)::before {
```
kra-mo, line 47: I'd make this a couple of pixels less tall, like in the mockup.

kra-mo (changes requested):
The rest looks good to me, although I haven't been able to actually test it.

AndyScherzinger:
> We need to introduce a new active icon slot and implement upstream so it will multiple PRs across multiple apps to use the new slot.

Afaik this is already present in the lib since mail and contacts app already have active/files icon state?

-> It is just the files app not implementing it since the sidebar has dynamic elements others app plug into the sidebar in contrast to other apps. Except that is what you meant Fon with the slot, but that one would be specific to the files app, in general the mechanism is there.

kra-mo:
Also linking here, `font-weight` should be consistently `500`: #8469

nfebe:
Screenshot From 2026-04-28 11-49-23


@kra-mo

kra-mo (approved):
Could I see how it looks on hover? Looks good in general now :)

nfebe:
@kra-mo 

image

jancborchardt (approved):

ShGKme:
The design seems different in the navigation and list items.
In the list item, it is slightly outside.

NcAppNavigationItem | NcListItem
-- | --
image | image

ShGKme:
`NcAppNavigationItem` now has a different hover effect from `NcListItem`s and `NcButton`.

nfebe:
Updated NcListItems and NcButton.

src/assets/NcAppNavigationItem.scss
```diff
@@ -28,16 +28,42 @@
...
-	// When .active class is applied, change color background of link and utils. The
-	// !important prevents the focus state to override the active state.
-	&.active {
+	// !important so :focus-within does not override the active state.
+	// New design (NC34+): glassy tinted bg + inline-start stripe.
+	&:not(.app-navigation-entry--legacy).active {
+		background-color: color-mix(in srgb, var(--color-primary-element) 16%, transparent) !important;
+
+		&:hover {
+			background-color: color-mix(in srgb, var(--color-primary-element) 25%, transparent) !important;
```
susnux: So as mentioned by @kra-mo this should then be primary-light and primary-light-hover

susnux (approved):
As far as I tested no regressions on v33

ShGKme:
Renamed for the changelog generation.

src/components/NcAppNavigation/NcAppNavigation.vue
```diff
@@ -354,9 +358,14 @@ function handleEsc(): void {
...
+	// New design (NC34+): transparent so the frosted chrome on NcContent shows through.
+	background-color: transparent;
+
+	&--legacy {
+		background-color: var(--color-main-background-blur, var(--color-main-background));
+		-webkit-backdrop-filter: var(--filter-background-blur, none);
+		backdrop-filter: var(--filter-background-blur, none);
+	}
 
 	&--closed {
```
susnux, line 370: This breaks the settings app in server

susnux: But proper fix is to fix the settings app: https://github.com/nextcloud/server/pull/60236

src/assets/NcAppNavigationItem.scss
susnux, line 1: Image


When using real images for the icons the styles are now wrong with 34 as the background is no longer primary the color should not be inverted.

src/components/NcContent/NcContent.vue
```diff
@@ -190,6 +194,15 @@ function setAppNavigation(value: boolean): void {
...
 }
+
+// New design (NC34+): divider only when an open navigation is present.
+@media only screen and (min-width: $breakpoint-mobile) {
+	.content:not(.content--legacy) .app-navigation:not(.app-navigation--closed):not(.app-navigation--close) ~ .app-content {
+		border-inline-start: 1px solid var(--color-border);
+		border-start-start-radius: var(--body-container-radius);
+		border-end-start-radius: var(--body-container-radius);
+	}
+}
```
susnux, line 205: This breaks on Nextcloud 33:

https://github.com/nextcloud/server/issues/61262

PR #8454, opened 2026-04-22
fix(NcAppNavigationSearch): Show search clear icon only when field contains something

jancborchardt:
### ☑️ Resolves

- Fixes the x clear button being there all the time in the search field
- Now adjusted to the behavior of any common platform: x is only present when content is in the field (no matter if field is focused or not)
- The CSS is to prevent the jump in width of the field when the trailing icon is shown or not (see iteration prompt)

### 🖼️ Screenshots

**🏚️ Before**

[search before.webm](https://github.com/user-attachments/assets/61405559-fd01-4a66-8afb-9e32f33367a9)

**🏡 After**

[search after.webm](https://github.com/user-attachments/assets/3bf8b4fa-05ee-4368-8bbc-bf098a9e9a27)

### 🚧 Tasks

- [ ] Has to be checked with the Files app. I didn’t manage to link it up locally with the npm pack strategy apparently due to the `/build/frontend-legacy` folder taking precedence and being unable to resolve that.

### 🏁 Checklist

---

AI-assisted: GitHub Copilot (OpenAI GPT-5.4)

> **Prompt:**
> In the NcAppNavigationSearch component, the close/delete icon in the search field should not be shown by default. It should only appear when there is content in the field, like at least 1 character typed in.
>
> **Iteration:**
> The change works as intended, except that now the width of the search bar jumps. This is because ".input-field--trailing-icon" adds some "--input-padding-end". Please make sure 
that there is no width jumping caused by showing or not showing the clear icon of the search field.

### DIFF
--- a/src/components/NcAppNavigationSearch/NcAppNavigationSearch.vue
+++ b/src/components/NcAppNavigationSearch/NcAppNavigationSearch.vue
@@ -133,7 +133,7 @@ export default {
 			class="app-navigation-search__input"
 			labelOutside
 			:placeholder="placeholder ?? label"
-			showTrailingButton
+			:showTrailingButton="model.length > 0"
 			:trailingButtonLabel="t('Clear search')"
 			type="search"
 			@trailingButtonClick="onCloseSearch">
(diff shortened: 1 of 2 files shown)


ShGKme (approved):
All good, but ESLint is not happy

jancborchardt (commented):
@ShGKme ESLint errors fixed. :)

ShGKme (approved):

Antreesy (approved):

jancborchardt (commented):
Anything else needed, or could this be merged? :)

PR #8462, opened 2026-04-23
ci(dependabot): bring back Nextcloud command approve

ShGKme:
### ☑️ Resolves

- Ref: https://github.com/nextcloud-libraries/nextcloud-vue/pull/8368/
- Example: https://github.com/nextcloud-libraries/nextcloud-vue/pull/8460

### 🏁 Checklist

### DIFF
--- a/.github/workflows/dependabot-approve-merge.yml
+++ b/.github/workflows/dependabot-approve-merge.yml
@@ -50,6 +50,13 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
+      # Nextcloud bot approve
+      - name: Nextcloud Bot approve
+        uses: hmarr/auto-approve-action@f0939ea97e9205ef24d872e76833fa908a770363 # v4.0.0
+        if: startsWith(steps.branchname.outputs.branch, 'dependabot/')
+        with:
+          github-token: ${{ secrets.DEPENDABOT_AUTOMERGE_TOKEN }}
+
       # Enable GitHub auto merge
       - name: Auto merge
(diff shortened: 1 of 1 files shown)


Antreesy (approved):

.github/workflows/dependabot-approve-merge.yml
```diff
@@ -50,6 +50,13 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
+      # Nextcloud bot approve
+      - name: Nextcloud Bot approve
+        uses: hmarr/auto-approve-action@f0939ea97e9205ef24d872e76833fa908a770363 # v4.0.0
+        if: startsWith(steps.branchname.outputs.branch, 'dependabot/')
+        with:
+          github-token: ${{ secrets.DEPENDABOT_AUTOMERGE_TOKEN }}
```
susnux, line 58: But then this should be part of the templates.
It is not - I guess its replaced with the bot approval directly above this?

Antreesy: Library requires two approvals, maybe that was the reason for duplication?

susnux: Ah yes this one requires two (most only one).
Not sure when it got removed but its in upstream templates, so should also be added back to library templates:

https://github.com/nextcloud-libraries/.github/pull/150

susnux (approved):

.github/workflows/dependabot-approve-merge.yml
ShGKme: Do we need it in a template, if this is only needed in this repo?

PR #8469, opened 2026-04-24
feat: Semantic font weights for interactive elements

kra-mo:
All general text should now be weight 400, interactive controls 500, and headings 600. Also added variables, as suggested by @ShGKme.

Depends on https://github.com/nextcloud/server/pull/59899

### ☑️ Resolves

- Fix #8083

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

- [x] ⛑️ Tests are included or are not applicable
- [x] 📘 Component documentation has been extended, updated or is not applicable
- [x] 2️⃣ Backport to `stable8` for maintained Vue 2 version or not applicable

PR #8477, opened 2026-04-27
feat(NcAppSettingsDialog): new navigation design

### LINKED ISSUE #7641: Left-align title of settings
Actually, while we're talking about alignment, it would be nice if we could align the title of the main settings dialog (which is currently centered) with the sidebar items.

This would both make more sense structurally (since it also titles the sections and it wouldn't be free floating anymore), and a nicer use of space.

When there is a sidebar (which in most cases, there will be), this would also allow us to shift the content up and make more settings visible initially. The close button could be overlaid instead of a bar, the same as other similar buttons.

And if there isn't a sidebar, the title could just be part of the main content and scroll offscreen (since in this case, the contents wouldn't scroll too much anyway as they would be short.)

Example with the current Files settings:

| Current | Proposal |
|--------|--------|
| Image | Image |

nfebe:
The settings dialog title now sits over a tinted navigation column.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
settings-dialog-before-7641 | image

Closes #7641

### DIFF
--- a/src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
+++ b/src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
@@ -13,6 +13,7 @@ import NcVNodes from '../NcVNodes/NcVNodes.vue'
 import NcAppSettingsDialogVersion from './NcAppSettingsDialogVersion.vue'
 import { useIsMobile } from '../../composables/useIsMobile/index.ts'
 import { t } from '../../l10n.ts'
+import { isLegacy34 } from '../../utils/legacy.ts'
 import { APP_SETTINGS_LEGACY_DESIGN_KEY, APP_SETTINGS_REGISTRATION_KEY } from './useAppSettingsDialog.ts'
 
 export interface IAppSettingsSection {
@@ -215,6 +216,7 @@ function unregisterSection(id: string) {
 	<NcDialog
 		v-if="open"
 		class="app-settings"
+		:class="{ 'app-settings--legacy': isLegacy34 }"
 		contentClasses="app-settings__content"
 		navigationClasses="app-settings__navigation"
 		:additionalTrapElements
@@ -225,6 +227,9 @@ function unregisterSection(id: string) {
 		:name
 		@update:open="handleCloseModal">
 		<template v-if="hasNavigation" #navigation="{ isCollapsed }">
+			<div v-if="!isLegacy34" class="app-settings__title">
+				{{ name }}
+			</div>
 			<ul
 				v-if="!isCollapsed"
 				class="navigation-list">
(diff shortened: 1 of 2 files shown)


ShGKme:
image

On the design draft:
- The main content section is rounded with a border, like with the new design on `NcAppContent`
- The dialog title is left aligned with the buttons' text
- The dialog title and the first section title are aligned
- Left/right offset around the buttons are equal

Not sure about the font-size, though...

ShGKme:
Dialog navigation is not only the `NcAppSettingsDialog` feature, but also the `NcDialog` feature. For example, used in the file picker dialog.

It might make sense to update the design from the `NcDialog` side. But I'm unsure how we can do it preventing breaking changes.

susnux (commented):
Same question here: What about backwards compatibility?
Keep in mind this is also used in old server versions.

So we use legacy fallbacks to keep the existing design for old server versions, for example see here: https://github.com/nextcloud-libraries/nextcloud-vue/pull/7351

basically: add new styles by default and add some legacy class if server version is < 34.
Then apply the old styles for the legacy class.

susnux:
> It might make sense to update the design from the NcDialog side. But I'm unsure how we can do it preventing breaking changes.

Maybe it makes sense to strip the navigation completely, so make the navigation a part of the app settings dialog.
Then you can style it like you want, the design of the current dialog navigation was mostly inspired by the filepicker.

Meaning its specific to that component, so we maybe want to just use this design here. So in the future we can deprecate the navigation of the NcDialog or at least make it more generic.

nfebe:
settings-dialog-after-7641

jancborchardt (dismissed):
@nfebe regarding the last screenshot: The heading in the content container should be aligned with the heading on the left, as in the mockup by @kra-mo in https://github.com/nextcloud-libraries/nextcloud-vue/issues/7641

Currently, the "General" text is stuck to the top of the dialog too much.

src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
```diff
@@ -258,68 +263,191 @@ function unregisterSection(id: string) {
 </template>
 
 <style lang="scss" scoped>
-.app-settings {
+$navigation-width: 200px;
+// Opaque tint for the dialog surface so the modal stays a solid panel
+// (translucent over `transparent` would let background content bleed through).
+$nav-tint-solid: color-mix(in srgb, var(--color-primary-element) 8%, var(--color-main-background));
```
kra-mo: I'd make this the same color as form boxes, so for now this: https://github.com/nextcloud-libraries/nextcloud-vue/blob/9cf17afedc1f1f189ae7fa205475870a8000025e/src/components/NcFormBox/NcFormBoxItem.vue#L107

src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
```diff
@@ -258,68 +263,191 @@ function unregisterSection(id: string) {
...
+$nav-tint-strong: color-mix(in srgb, var(--color-primary-element) 16%, transparent);
```
kra-mo: This can also be less strong in this case.

kra-mo (changes requested):
> Currently, the "General" text is stuck to the top of the dialog too much.

To translate this to a more actionable complaint, the top of the content needs more padding :)

kra-mo:
It would be good to reduce the maximum width for this dialog now by a bit as well.

kra-mo:
The font-weight also should not change based on selection. It should always be `500`. See #8469

nfebe:
> @nfebe regarding the last screenshot: The heading in the content container should be aligned with the heading on the left, as in the mockup by @kra-mo in https://github.com/nextcloud-libraries/nextcloud-vue/issues/7641

@jancborchardt those titles are not fixed i.e position is determined by scroll but let me see what can be done.

jancborchardt:
@nfebe right, they shouldn't be fixed. But what I mean is that in the screenshot you showed, the first header is too far up.

Did you scroll slightly down there? If not, would be good to give it a bit of whitespace above.

nfebe:
@jancborchardt  @kra-mo 

image

kra-mo (approved):
Same comment about the height of the pill for the selected item as for the general left navigation. It should be a little less tall. Other than that, looks good to me.

One thing I would still change, if possible, is to reduce the maximum width of the entire dialog by a bit. I raised this in the past, but we agreed that we'd do it once we redesign the rest, so I think now is the time.

src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
```diff
@@ -225,6 +227,9 @@ function unregisterSection(id: string) {
 		:name
 		@update:open="handleCloseModal">
 		<template v-if="hasNavigation" #navigation="{ isCollapsed }">
+			<h2 v-if="!isLegacy34" class="app-settings__title">
+				{{ name }}
+			</h2>
```
ShGKme: This `h2` duplicates existing visually hidden `h2`. For now, let's just make it a general div. Later we can improve `NcDialog` to support this case.

ShGKme (changes requested):
Mobile view is broken.

image

nfebe:
Fixed responsiveness.

ShGKme:
The title and the close button are not visible

image

nfebe:
See a stale version was pushed? updated.

image

nickvergessen:
This is broken on dark mode btw: Instead of getting brighter on hover (?) like the "Blur camera …" setting, it tries to get darker and that's not visible anymore.

grafik

susnux (approved):

ShGKme:
The colors are still different. Hover effect should make the color darker in the light theme, but lighter in the dark one.


https://github.com/user-attachments/assets/8f7dde65-1e74-4b4a-b4b8-b5addf06844d

nfebe:
image

ShGKme (approved):
The color is still different and the title on mobile is not aligned.

But this is not critical. Let's merge as it is and adjust in the follow-up, so apps can start testing the new design.

ShGKme:
Renamed a bit for the changelog to clarify it for library users. There was a significant design change, not only the "alignment"

PR #8479, opened 2026-04-28
fix(NcDateTimePicker): improve some usability issues

GVodyanov:
Fixes a few small issues:

- Selected offset/disabled days now have white text color
- Added icon to submit button
- Fixed scrolling through the months. Previously horizontal scroll made it go to the past, and vertical to the future, which is really weird. Now scrolling to left/right goes past/present.

### 🖼️ Screenshots

|🏚️ Before | 🏡 After|
|---|---|
|image | image|
| image |image |

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/NcDateTimePicker/NcDateTimePicker.vue
+++ b/src/components/NcDateTimePicker/NcDateTimePicker.vue
@@ -288,7 +288,7 @@ import {
 	getFirstDay,
 } from '@nextcloud/l10n'
 import VueDatePicker from '@vuepic/vue-datepicker'
-import { computed, useTemplateRef } from 'vue'
+import { computed, ref, useTemplateRef } from 'vue'
 import NcIconSvgWrapper from '../NcIconSvgWrapper/NcIconSvgWrapper.vue'
 import NcTimezonePicker from '../NcTimezonePicker/NcTimezonePicker.vue'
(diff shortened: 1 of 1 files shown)


SebastianKrupinski (approved):
Looks good

susnux (commented):
it would be easier if you could split the PR into individual changes.
e.g. one for the color and one for the scrolling as they are unrelated

jancborchardt (commented):
> Added icon to submit button

@GVodyanov can you show a screenshot of that? Because the "clock" button already _only_ has an icon but no text, and could use some text – namely "Set time" (right?).
It’s most important for a button to have text. Icons are secondary, and very often not needed.

GVodyanov:
> can you show a screenshot of that? Because the "clock" button already _only_ has an icon but no text, and could use some text – namely "Set time" (right?).

No in this case it was just empty, it's in the screenshot in the PR description at the very bottom of the second row. I can totally add text too though yeah.

ShGKme:
@GVodyanov Could you split into PRs?

Fixing the colors is a bug fix we definitely need to have with no discussion.

However, the scroll is a behavior change and more opinionated.
For example, I don't have a horizontal mouse wheel (as many users) and cannot use the scroll at all now.

PR #8484, opened 2026-04-30
feat(NcPopover): expose distance and skidding props

pringelmann:
floating-vue's `<Dropdown>` accepts `distance` and `skidding` for nudging the popper relative to the trigger. NcPopover hardcodes `distance` to `10` and never passes `skidding` through, so consumers can't shift a popover without CSS hacks that confuse the positioning math. Defaults are preserved, so this is non-breaking.

### Resolves

Needed for the redesigned app launcher in nextcloud/server#59888.

### Screenshots

No visual change with defaults.

### Tasks

- [x] Add `distance` and `skidding` props
- [x] Forward both to `<Dropdown>`

### Checklist

- [ ] Tests included or not applicable: not applicable. Defaults match the previous hardcoded values, so existing tests still cover the default path.
- [x] Component documentation extended/updated or not applicable: prop JSDoc added, styleguidist picks it up automatically.
- [ ] Backport to `stable8` for Vue 2 or not applicable: additive and non-breaking. Happy to backport if needed.

### DIFF
--- a/src/components/NcPopover/NcPopover.vue
+++ b/src/components/NcPopover/NcPopover.vue
@@ -183,7 +183,8 @@ See: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/
 		:boundary="boundary || undefined"
 		:container
 		:delay
-		:distance="10"
+		:distance
+		:skidding
 		handleResize
 		:noAutoFocus="true /* Handled by the focus trap */"
 		:placement="internalPlacement"
@@ -281,6 +282,28 @@ export default {
 		 */
 		delay: {
 			type: [Number, Object],
+			default: 0,
+		},
+
+		/**
+		 * Distance in pixels between the popover and the trigger element.
+		 *
+		 * @type {number}
+		 */
+		distance: {
+			type: Number,
+			default: 10,
+		},
+
+		/**
+		 * Skidding (offset along the trigger axis) in pixels. For a `bottom-start`
+		 * placement, positive values shift the popover toward the end of the trigger,
+		 * negative values toward the start.
+		 *
+		 * @type {number}
+		 */
+		skidding: {
+			type: Number,
 			default: 0,
 		},
 


ShGKme:
> Needed for the redesigned app launcher in [nextcloud/server#59888](https://github.com/nextcloud/server/issues/59888).

Could you add more details?

> consumers can't shift a popover without CSS hacks that confuse the positioning math

I assume this is to align the popover with the app container.

image


With "CSS Hacks" we can say that the menu is exactly on the `--body-container-margin` position from the `inset-inline-start`.

But with `skidding` we'd need to either hardcode a magic number or have some complex computation on JS with DOM API requests if I got it correctly.

Also, with `distance=10` the popover is exactly on the caret size from the trigger. With bigger value it moves away from the trigger, and with smaller - overlaps the trigger. Do we need both options?

pringelmann:
> > Needed for the redesigned app launcher in [nextcloud/server#59888](https://github.com/nextcloud/server/issues/59888).
> 
> Could you add more details?
> 
> > consumers can't shift a popover without CSS hacks that confuse the positioning math
> 
> I assume this is to align the popover with the app container.
> image
> 
> With "CSS Hacks" we can say that the menu is exactly on the `--body-container-margin` position from the `inset-inline-start`.
> 
> But with `skidding` we'd need to either hardcode a magic number or have some complex computation on JS with DOM API requests if I got it correctly.
> 
> Also, with `distance=10` the popover is exactly on the caret size from the trigger. With bigger value it moves away from the trigger, and with smaller - overlaps the trigger. Do we need both options?

For our consumer, a CSS workaround that just shifts the popover (`inset-inline-start: var(--body-container-margin) !important; transform: none !important;` etc.) would visually move the popover where we want it. It would also detach the arrow: `.v-popper__arrow-container` is a child of the popover, positioned by floating-vue's `arrow` middleware to point at the trigger from the popover's *un-shifted* position. CSS-moving the popover moves the arrow along with it, so the arrow now points at empty space ~80 px to the side of the trigger. Fixing that requires another `!important` rule to hide or reposition it, and the reposition value isn't expressible in CSS without runtime DOM measurement.

Going through `:skidding`/`:distance` lets floating-vue's positioning system see the offset, so the arrow stays correctly aligned with the trigger. Defaults preserve existing behavior (distance 10, skidding 0) so the addition is a no-op for current consumers.

`-82` is a magic number, fair. Either way the consumer encodes that delta somewhere (here it's the skidding value, in the CSS approach it would be the arrow override).

pringelmann:
Although I noticed that `skidding` seems to pass through implicitly via `$attr` fall-through, so perhaps not necessary to document as public API. However the hard-coded distance is still an issue.

pringelmann:
Closing for now in favor of CSS workaround. `skidding` already works, and `distance` CSS overrides are not likely to cause issues with the popover flipping at runtime for this specific use case. However, in future I still think it will be useful to expose these as public API props since CSS workarounds do not take runtime repositioning into account.

PR #8503, opened 2026-05-06
[stable8] feat(NcAppSettingsDialog): new navigation design

nfebe:
The settings dialog title now sits over a tinted navigation column on the inline-start edge of the dialog, replacing the centered header. Navigation links use the default button weight with a rounded inline-start stripe on the active item.

Backport of https://github.com/nextcloud-libraries/nextcloud-vue/pull/8477

### DIFF
--- a/src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
+++ b/src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
@@ -14,6 +14,7 @@ import NcVNodes from '../NcVNodes/NcVNodes.vue'
 import NcAppSettingsDialogVersion from './NcAppSettingsDialogVersion.vue'
 import { useIsMobile } from '../../composables/useIsMobile/index.ts'
 import { t } from '../../l10n.js'
+import { isLegacy34 } from '../../utils/legacy.ts'
 import { APP_SETTINGS_LEGACY_DESIGN_KEY, APP_SETTINGS_REGISTRATION_KEY } from './useAppSettingsDialog.ts'
 
 export interface IAppSettingsSection {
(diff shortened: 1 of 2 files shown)


src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
```diff
@@ -79,7 +80,9 @@ provide(APP_SETTINGS_REGISTRATION_KEY, {
 	unregisterSection,
 })
 
-provide(APP_SETTINGS_LEGACY_DESIGN_KEY, toRef(() => props.legacy))
+const useNewDesign = computed(() => !props.legacy && !isLegacy34)
+
+provide(APP_SETTINGS_LEGACY_DESIGN_KEY, toRef(() => !useNewDesign.value))
```
ShGKme: `use*` is a name pattern for composables.

Also, those things are different.
There are design changes that depend on the Nextcloud version.

And there is the `legacy` mode when the component was used not with new settings components but with any custom content. It depends not on the server version, but on the content component's user put inside and controlled by a prop.

ShGKme (dismissed):
In the main the merged version used the new navigation design on older server versions.
In this PR, the navigation design is always new

`main` on v32 | This PR on v32
-- | --
image | image

---

On legacy settings format (`:legacy="true"` - default) there is no title on any version

v34 | v32
-- | --
image | image

---

On the new settings format (`:legacy="false"`):

- No title on v32
- Settings content has legacy styles applied even though the prop says to use the new format

v34 | v32
-- | --
image | image

sorbaugh (approved):

nfebe:
| <34 | 34 |
| -----  | --- |
|   |  |

susnux (approved):
looks good now

PR #8532, opened 2026-05-13
feat(referencePicker): introduce getReferenceWithPicker()

mejo-:
Required for Text and Collectives to pass the link title from the picker (e.g. Collectives PagePicker.vue) to the consumer (menubar in Text).

### 🏁 Checklist

### DIFF
--- a/src/components/NcRichText/NcReferencePicker/NcSearch.vue
+++ b/src/components/NcRichText/NcReferencePicker/NcSearch.vue
@@ -122,7 +122,7 @@ export default {
 
 	emits: [
 		'submit',
-		'submitReference',
+		'pick',
 	],
 
 	data() {
@@ -253,7 +253,7 @@ export default {
 					if (item.title) {
 						reference.title = item.title
 					}
-					this.$emit('submitReference', reference)
+					this.$emit('pick', reference)
 				} else if (item.isMore) {
 					this.searchMoreOf(item.providerId).then(() => {
 						// allow clicking twice on the same "more" item
--- a/src/components/NcRichText/NcReferencePicker/NcSearch.vue
+++ b/src/components/NcRichText/NcReferencePicker/NcSearch.vue
@@ -249,7 +249,11 @@ export default {
 				if (item.resourceUrl) {
 					this.cancelSearchRequests()
 					this.$emit('submit', item.resourceUrl)
-					this.$emit('submitReference', { link: item.resourceUrl, title: item.title })
+					const reference = { link: item.resourceUrl }
+					if (item.title) {
+						reference.title = item.title
+					}
+					this.$emit('submitReference', reference)
 				} else if (item.isMore) {
 					this.searchMoreOf(item.providerId).then(() => {
 						// allow clicking twice on the same "more" item
--- a/src/components/NcRichText/NcReferencePicker/NcSearch.vue
+++ b/src/components/NcRichText/NcReferencePicker/NcSearch.vue
@@ -122,6 +122,7 @@ export default {
 
 	emits: [
 		'submit',
+		'submitReference',
 	],
 
 	data() {
@@ -248,6 +249,7 @@ export default {
 				if (item.resourceUrl) {
 					this.cancelSearchRequests()
 					this.$emit('submit', item.resourceUrl)
+					this.$emit('submitReference', { link: item.resourceUrl, title: item.title })
 				} else if (item.isMore) {
 					this.searchMoreOf(item.providerId).then(() => {
 						// allow clicking twice on the same "more" item
--- a/src/functions/reference/index.ts
+++ b/src/functions/reference/index.ts
@@ -13,7 +13,10 @@ export {
 	renderWidget,
 } from './widgets.ts'
 
-export { getLinkWithPicker } from './referencePickerModal.ts'
+export {
+	getLinkWithPicker,
+	getReferenceWithPicker,
+} from './referencePickerModal.ts'
 
 export {
 	type ReferenceProvider,
--- a/src/functions/reference/referencePickerModal.ts
+++ b/src/functions/reference/referencePickerModal.ts
@@ -44,7 +44,9 @@ export async function getLinkWithPicker(providerId?: string, isInsideViewer?: bo
  * The function guarantees a `link: string` on the resolved object. Custom picker elements may
  * include additional fields (e.g. `title`).
  *
- * @template T - Expected shape of the resolved reference, mus extend `PickerReference`
+ * @experimental The shape of the resolved reference and the signature of this function may still change.
+ *
+ * @template T - Expected shape of the resolved reference, must include `link: string`
  * @param providerId - Optional ID of initial selected provider
  * @param isInsideViewer - Should be true if this function is called while the Viewer is displayed
  */
@@ -63,7 +65,7 @@ export async function getReferenceWithPicker<T extends { link: string }>(provide
 			view.unmount()
 			reject(new Error('User cancellation'))
 		},
-		onSubmitReference(reference: T) {
+		onPick(reference: T) {
 			view.unmount()
 			resolve(reference)
 		},
--- a/src/functions/reference/referencePickerModal.ts
+++ b/src/functions/reference/referencePickerModal.ts
@@ -7,7 +7,7 @@ import { createApp } from 'vue'
 import NcReferencePickerModal from './../../components/NcRichText/NcReferencePicker/NcReferencePickerModal.vue'
 import { getProvider } from './providerHelper.js'
 
-type PickerSubmitResult = {
+type PickerReference = {
 	link: string
 	title?: string
 }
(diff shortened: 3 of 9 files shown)


max-nextcloud (approved):

src/functions/reference/referencePickerModal.ts
```diff
@@ -28,9 +33,40 @@ export async function getLinkWithPicker(providerId?: string, isInsideViewer?: bo
...
+ * Creates a reference picker modal and return a promise which provides the reference object
+ *
+ * @param providerId - Optional ID of initial selected provider
+ * @param isInsideViewer - Should be true if this function is called while the Viewer is displayed
+ */
+export async function getReferenceWithPicker(providerId?: string, isInsideViewer?: boolean): Promise<PickerSubmitResult> {
+	const modalId = 'referencePickerModal'
+	const modalElement = document.createElement('div')
+	modalElement.id = modalId
+	document.body.append(modalElement)
```
ShGKme, line 57: We have `spawnDialog` util to mount dialogs

src/functions/reference/index.ts
```diff
@@ -13,7 +13,11 @@ export {
 	renderWidget,
 } from './widgets.ts'
 
-export { getLinkWithPicker } from './referencePickerModal.ts'
+export {
+	type PickerSubmitResult,
```
ShGKme: Is this type needed in public library export?

src/functions/reference/referencePickerModal.ts
```diff
@@ -28,9 +33,40 @@ export async function getLinkWithPicker(providerId?: string, isInsideViewer?: bo
...
+	return promise
+}
+
+/**
+ * Creates a reference picker modal and return a promise which provides the reference object
+ *
+ * @param providerId - Optional ID of initial selected provider
+ * @param isInsideViewer - Should be true if this function is called while the Viewer is displayed
+ */
+export async function getReferenceWithPicker(providerId?: string, isInsideViewer?: boolean): Promise<PickerSubmitResult> {
```
ShGKme: Having `isInsideViewer` as a position parameter seems unexpected to me. Picker shouldn't know, it is it in Viewer. If it's really needed, I'd make it an option rather than a second position parameter.

ShGKme (commented):
If I understand the change correctly, the following components have changed public API:
- `NcReferencePicker`
- `NcSearch`

src/functions/reference/index.ts
mejo-: Not necessarily, I dropped it now.

src/functions/reference/referencePickerModal.ts
mejo-: As written in 1:1 chat, I kept it that way to keep `getLinkWithPicker()` and `getReferenceWithPicker()` as similar as possible to make it a no-brainer to migrate from one to the other.

mejo-:
https://github.com/nextcloud/text/pull/8615 and https://github.com/nextcloud/collectives/pull/2522 are the respective PRs in Text and Collectives to use this new feature.

ShGKme (changes requested):
This PR changes API for custom picker elements - the `submit` event sends reference as an object instead of a string link.

This change for the `custom picker element` is not backward compatible. An object cannot be used where the string was expected.

All the components using this result are updated to support both the new object result and the old string link result. All the components are **backward compatible** and support old custom picker elements.

However, they are not **forward compatible**. Older `nextcloud-vue` version will break with new custom picker element.

For example, new `Collectives`'s `PagePicker` custom picker element emits `submit` event with `{ link, title }`. Then an old `nextcloud-vue`'s `getLinkWithPicker` handle it as a string link.

For the smart picker it results in creating `'[object Object]'` text node.

---

Proposal 1: instead of changing the existing `submit` event, add a new event, `submit-reference` or just `pick`.

Then custom picker elements can emit both events. Where supported, the new event with extra data will be used, and where not - the old event still works.

Later we can mark the old event as deprecated and remove.

```js
this.$el.dispatchEvent(new CustomEvent('submit', {
	bubbles: true,
	detail: pageLink,
}))
this.$el.dispatchEvent(new CustomEvent('pick', {
	bubbles: true,
	detail: {
		link: pageLink,
		title: page.title,
	},
}))
```

---

Proposal 2: submit a string with additional `link` and `title` properties. So it is still a string, but we can extract the details

```js
const result = 'https://nextcloud.com'
result.link = 'https://nextcloud.com'
result.title = 'Nextcloud'
```

Dirty. Hacky. Should work.

---

Proposal 3: emit an object with `toString()` method. 

```js
const result = {
  link: 'https://nextcloud.com',
  title: 'Nextcloud',
  toString() => 'https://nextcloud.com',
}
```

It is simpler, but will only work where the value is casted to string. For example, it works for the `NcRichText` with `document.createTextNode(result)`. But it may not work for all the `getLinkWithPicker` users...

mejo-:
Thanks for the the thorough review @ShGKme :pray: 

> Proposal 1: instead of changing the existing submit event, add a new event, submit-reference or just pick.

I went with this one as it seemed like the cleanest ones for me.

src/components/NcRichText/NcReferencePicker/NcSearch.vue
```diff
@@ -248,6 +249,7 @@ export default {
 				if (item.resourceUrl) {
 					this.cancelSearchRequests()
 					this.$emit('submit', item.resourceUrl)
+					this.$emit('submitReference', { link: item.resourceUrl, title: item.title })
```
Antreesy: Is `item.title` guaranteed to be defined?

mejo-: Good point. As far as I understand, it can be undefined (`rawLinkEnty` doesn't have a `title` property).  I fixed it now to only add the title attribute if `item.title` is truthy.

src/functions/reference/referencePickerModal.ts
```diff
@@ -7,8 +7,13 @@ import { createApp } from 'vue'
 import NcReferencePickerModal from './../../components/NcRichText/NcReferencePicker/NcReferencePickerModal.vue'
 import { getProvider } from './providerHelper.js'
 
+type PickerReference = {
+	link: string
+	title?: string
+}
```
ShGKme: Is it the type we require from custom picker elements or just the one used in Collectives/Text (bounded to the providerId)?

If it is bounded, I'd use a generic type here like:
```ts
export async function getReferenceWithPicker<T extends { link: string }>(providerId?: string, isInsideViewer?: boolean): Promise<T>
```

If we are specifying a type for everyone, we need something we won't need to change and can extend.

For example, the related reference widget uses `name` for links (`openGraphObject`). 

Though, original OGP calls it a `title`...

ShGKme:
I thought it through again, and I'd stick with a short, simple name of the even. Then we can deprecate and remove the old event later, still having a simple name. What about something like `pick` (because it is a reference picker, and a user _picks_ a reference)?

ShGKme:
Currently all the components have 2 events going through individually.
1. `submit`: NcCustomPickerElement - NcReferencePicker - NcReferencePickerModal - getLinkWithPicker
2. `submitReference`: NcCustomPickerElement - NcReferencePicker - NcReferencePickerModal - getReferenceWithPicker

This way, the new event is not an advanced replacement for the old one, but an independent event:
- The old function `getLinkWithPicker` only works when the old `submit` event is fired (fine for backward compatibility)
- The new function `getReferenceWithPicker` only works with the new `submitReference` event, and never resolves the promise for old picker elements

I'd always use the old event as a fallback event rather than a parallel one:
- `NcCustomPickerElement` listens for both, but prefers the new and emits the new one
- All the intermediate internal components operate in the new event
- User-end functions end components support both formats for compatibility

From `NcReferencePickerModal` it would be something like:
```js
// One handler, always emitting both events
onSubmit(value) {
	this.show = false
	this.$emit('submit', value.link)
	this.$emit('submitReference', value)
},
```

And from `NcCustomElementPicker` a bit more complicated to handle async race:
```js
// On subscribing - only save the submitted result, with preference on the second event

this.renderResult.object.$on('submitReference', (e) => {
	this.submitted = e
})
this.renderResult.object.$on('submit', (e) => {
	// Fallback in case there were no submitReference event
	this.submitted ??= { link: e }
})
this.renderResult.element.addEventListener('submitReference', (e) => {
	this.submitted = e.detail
})
this.renderResult.element.addEventListener('submit', (e) => {
	// Fallback in case there were no submitReference event
	this.submitted ??= { link: e.detail }
})

// And fire an event on submitted change

// Watch is async, scheduler will trigger the effect in microtasks queue with the final submitted value,
// if both events where triggered at the same time (in the same tick)
watch: {
	submitted(() => {
		this.$emit('submitReference', this.submitted)
	}),
	// Or, to be a bit more save
	submitted: {
		handler(() => {
			this.$emit('submitReference', this.submitted)
		}),
		flush: 'post',
	},
},

// If watch feels complicated, it can be replaced with a manually schedule function call
```

This change might be complicated and not critical - we can do it in a follow-up.

src/functions/reference/referencePickerModal.ts
mejo-: Thanks, did this now.

ShGKme (approved):

PR #8533, opened 2026-05-13
fix(NcAppNavigationItem): Make active state darker than hover

### LINKED ISSUE #8526: [NcAppSettingsDialog] [NcAppNavigationItem] Redesign follow-ups
Some issues I noticed with #8477 and #8448 that should be fixed before the final release.

## Navigation

The selected item is lighter, but the hovered one is darker (the other way around in dark mode). The selected item should also be darker, the same as in the settings dialog and like in the mockup.

Image

## Settings Dialog

1. The close button is clipping with the content without a background. It should have one.

Image

<br>

2. :hover effects and the pill indicating selection on the sidebar don't animate.
3. The pill is taller than in the regular navigation sidebar and (thus) clips into the border radii. It should look the same as for the regular navigation.

Image

@nfebe

nfebe:
The selected nav item appeared lighter than the hovered one in the new NC34+ design. Use a semi-transparent darker overlay so the active state sits visually below hover, matching the mockup intent.

Resolves #8526

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot From 2026-05-13 13-44-12 | Screenshot From 2026-05-13 14-19-03

### DIFF
--- a/src/assets/NcAppNavigationItem.scss
+++ b/src/assets/NcAppNavigationItem.scss
@@ -34,10 +34,10 @@
 	// !important so :focus-within does not override the active state.
 	// New design (NC34+): glassy tinted bg + inline-start stripe.
 	&:not(.app-navigation-entry--legacy).active {
-		background-color: var(--color-primary-element-light) !important;
+		background-color: color-mix(in srgb, var(--color-primary-element) 16%, transparent) !important;
 
 		&:hover {
-			background-color: var(--color-primary-element-light-hover) !important;
+			background-color: color-mix(in srgb, var(--color-primary-element) 22%, transparent) !important;
 		}
 
 		&:not(.app-navigation-entry--editing) {


nfebe (commented):
Pointing out this was what already existed before @susnux pushed against it via reviews in #8448 (NcListItem a11y concern) (https://github.com/nextcloud-libraries/nextcloud-vue/pull/8448#discussion_r3146257439) and #8448 (NcAppNavigationItem follow-up) (https://github.com/nextcloud-libraries/nextcloud-vue/pull/8448#discussion_r3180551117).

So reverted as it matches @kra-mo's request in #8526 ("darker and not opaque").

The argument about contrast not being guaranteed is not accurate because the active bg is 84% transparent over --color-main-background; what shows through dominates the visible surface, and the text is rendered with --color-main-text -- the NC-guaranteed contrast pair.

kra-mo (approved):
At least from the screenshot, with these colors, contrast does seem to check out. Is that not always the case @susnux?

Image

ShGKme:
Could you also provide screenshots on:
- Dark theme
- Custom primary color light theme
- Custom primary color dark theme

We now have more and more new/updated UI components with new colors that are dynamically computed with different formulas in different places.

It not only makes it hard for maintenance and keeping colors consistent across different places but also for testing (each time we need to make sure it is accessible and supports Nextcloud themes and Nextcloud theming).

@kra-mo Can we have a new palette with the new colors?

kra-mo:
> Can we have a new palette with the new colors?

I mean, I don't think more and more color variables would make sense.

Computing colors from each other is nicer, but they should be based on constraints and logic, especially when it comes to transparency, which I think we've been underutilizing in general in favor or solid colors.

Like a lot of hover and border colors right now are hardcoded solid ones, which start looking off as soon as you put them over anything other than the default background color (in this case, the sidebar).

I think this may warrant a bigger discussion, possibly also tying into OKLCH?

ShGKme:
> I mean, I don't think more and more color variables would make sense.
> 
> Computing colors from each other is nicer, but they should be based on constraints and logic, especially when it comes to transparency, which I think we've been underutilizing in general in favor or solid colors.
> 
> Like a lot of hover and border colors right now are hardcoded solid ones, which start looking off as soon as you put them over anything other than the default background color (in this case, the sidebar).

Computing colors doesn't conflict with having them in the theming palette. We already do it.

For example, we have `--color-error` as a static value, and `--color-error-hover` computed from the `--color-error`. We can use this `--color-error-hover` without creating a new formula or thinking what to do with it on hover.

What is missing is a specific semantic named value in the theme.

The problem is reinventing the computation/color in many individual places. It results in:
- Inconsistency - different computation in different places for similar elements
- Complexity - no value to use for a new place during development
- Issues - each new place with a new formula must be tested in different themes with different theming params

> I think this may warrant a bigger discussion, possibly also tying into OKLCH?

My experience with OKLCH wasn't good. While in theory it should work better, the colors often looked worse than with simpler HSL manipulation.

But how exactly we compute (PHP, CSS, a-channel, OKLCH) doesn't matter.
If we have a named semantic value to reuse, we can change the computation later.

kra-mo:
Ok, let's set aside the topic of computed colors broadly.

This hover color should be consistent across all elements that are neutral/derived from primary, which means we should find some color to consistently mix with everything.

Primary, secondary, as well as tertiary-colored, including transparent controls should all look similar when hovered, which means in this case, the hover color of tertiary elements (which should be without a background), which should be semitransparent instead of opaque.

Let's say that all hovered elements tend toward some mix of fg + primary. Maybe 70% fg, 30% primary (depends on the color space). Let's call this the hover "shade".

So all elements' hover color would be mixed with this shade. Let's say primary's hover color would be primary + 20% shade, secondary's would be secondary + 15% shade and transparent controls' would be transparent + 10% shade (so shade at 10% opacity).

Does that make sense? I'm not sure I explained it well.

nfebe:
Dark more preview

| Before | After | 
| ---------- | ------- |
| Screenshot From 2026-05-20 14-24-22 | Screenshot From 2026-05-20 14-20-33 \

susnux (approved):
similar as in the forms thing. But still agree with need this to be defined in named styles. Otherwise this will explode sooner than later.

susnux:
> Does that make sense? I'm not sure I explained it well.

Yes but as Grigorii said we need this to be named semantic value to reuse. Otherwise this will blow up quite soon.
Moreover for the semi-transparent: This needs to be done very carefully as we need to consider every valid place a component is used with.
For example a button might not only be used on main background but within a table with a dark background (e.g. the currently active files entry). So for all those possible combinations we need to ensure there is always enough contrast.

PR #8544, opened 2026-05-16
fix(NcAppContent): background blur may be missing after minification

### LINKED ISSUE #17991: Background blur missing on left sidebar
Settings | Talk
---|---
Image | Image

ShGKme:
### ☑️ Resolves

- Fix: https://github.com/nextcloud/spreed/issues/17991
- `-webkit-backdrop-filter` doesn't work in Chromium
- `-webkit-backdrop-filter` should be added by build tools only if needed (according to the browserslist config), we don't need it in the source
- When both prefixed and unprefixed values are present, and prefixed is the second, CSS optimization with LightningCSS (used in Vite and RSPack) keeps the prefixed value only
  - Bug: https://github.com/parcel-bundler/lightningcss/issues/1229

### Screenshots

Before | After
-- | --
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppNavigation/NcAppNavigation.vue
+++ b/src/components/NcAppNavigation/NcAppNavigation.vue
@@ -363,7 +363,6 @@ function handleEsc(): void {
 
 	&--legacy {
 		background-color: var(--color-main-background-blur, var(--color-main-background));
-		-webkit-backdrop-filter: var(--filter-background-blur, none);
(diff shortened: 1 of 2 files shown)


nickvergessen (approved):
Confirmed in Firefox devtools

Antreesy (approved):
Tested in Chromium

PR #8545, opened 2026-05-16
fix(widgets): show description even on smaller embeds

### LINKED ISSUE #8524: Hiding the description for width < 450 negatively impacts rendering in Talk
As implemented in Talk and discussed in this issue https://github.com/nextcloud/spreed/issues/17881, when the width for the embed widget is <450, the description is not shown due to the size of the widget. I am not familiar with this code so I can't estimate the impact of changing this for all downstream consumers this but for the use case in Talk, the description should not be hidden.

Relevant code:

https://github.com/nextcloud-libraries/nextcloud-vue/blob/26567e9b8bef08c9a263df10ecd53e0ce7508a13/src/components/NcRichText/NcReferenceWidget.vue#L75-L92

Thank you and apologies in advance if this issue is not appropriate or if there are any errors as I'm not familiar with this codebase.

tyzbit:
### ☑️ Resolves

- Fix #8524, only observed results in the usage of the Talk app only. Other consumers may see changed behavior with rich widgets that are < 450px.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 Tasks

- [X] ...

### 🏁 Checklist

- [X] ⛑️ Tests are included or are not applicable (NA)
- [X] 📘 Component documentation has been extended, updated or is not applicable (NA)

### DIFF
--- a/src/components/NcRichText/NcReferenceWidget.vue
+++ b/src/components/NcRichText/NcReferenceWidget.vue
@@ -73,8 +73,8 @@ const noAccess = computed(() => {
 })
 
 const numberOfLines = computed(() => {
-	// no description for width < 450, one line until 550 and so on
-	const lineCountOffsets = [450, 550, 650, Infinity]
+	// no description for width < 250, one line until 550 and so on
+	const lineCountOffsets = [250, 550, 650, Infinity]
 	return lineCountOffsets.findIndex((max) => width.value < max)
 })
 
@@ -293,7 +293,8 @@ function destroyReferenceWidget() {
 
 	&--details {
 		padding: calc(var(--default-grid-baseline, 4px) * 3);
-		width: 60%;
+		width: 0;
+		flex-grow: 1;
 
 		p {
 			margin: 0;


ShGKme:
Hi Tyler, and my special greetings to a Vivaldi user. Thanks for your contribution.

Though this change fixes the problem for Talk, it does so by removing a feature rather than the root cause.

We definitely can improve the reference widget rendering, for example:
- Remove 40% empty space preserved for an image that doesn't exist
- Play around the sizes to show more data on a narrow block

But the problem comes from the width determination here...

In Talk, the bubble message tries to be as small as possible (`fit-content`). In case of the link - it is the link width itself. Then the reference widget tries to fit in the bubble, stretching it if necessary.

When there is an image OR the link is long, initial width is wide enough to stretch the bubble. Widget is wide and it renders the long description.

When there is no image, it fits the narrow bubble. It stays narrow, and widget content tries to fit in.

**Narrow message bubble --> narrow widget block --> minimal fitting content**

Instead of fitting into a narrow bubble, the widget should have stretched it.

IMO, we may have to solutions here:
1. On the Talk side, if there is a reference in the message, always render it wide to provide as much space to the widget as possible
2. On the library side, before checking the widget width and narrowing down, try to take as much space as possible. So that the widget stretches the bubble.

ShGKme:
@Antreesy What do you think?

tyzbit:
@ShGKme Thank you for taking the time to provide that informed analysis and info. I have converted this PR to a draft since more discussion will probably produce a better fix.

I like the solutions you proposed. I would like to ponder if they would behave in the best manner in these scenarios:

1. In Talk (out of scope for this PR, but as a consumer of this library), messages can be in bubbles on alternating sides or completely left-justified. If the widget takes up as much space as possible, would it be a good user experience for the widget to span the entire message content viewport? If not, Talk can limit individual message width to 75% (as an example) of the viewport in case of the alternating-sides view, right-justified, or do something similar if it doesn't already.
2. In this library, taking up as much space as possible would provide the most content in the preview but then we are no longer considering the length of the description in the calculation - maybe cases where the description is only a few words might result in extra-wide widgets with empty space.

Antreesy:
> Remove 40% empty space preserved for an image that doesn't exist

This change works for me to allow take the space without an image and keep an ellipsis:
```diff
&--details {
-	width: 60%;
+	width: 0;
+ 	flex-grow: 1;
}
```

> Play around the sizes to show more data on a narrow block

If we're working with hardcoded values, I would just decrease the lower border. 250px is approx. the width of the widget in the Talk during the call (in the narrow chat in sidebar)
```diff
-	const lineCountOffsets = [450, 550, 650, Infinity]
+	const lineCountOffsets = [250, 550, 650, Infinity]
```
With no image and no 60% limit there's plenty space to render a description. Even with a small thumbnail, something still fits
image

Another thing I discovered, that there is a `--compact` styles in the widget, that have been commited, but never used 🙈 
https://github.com/nextcloud-libraries/vue-richtext/commit/831fba7e0a095194c6cf4dcefdc6ec9c68415adf

> On the Talk side, if there is a reference in the message, always render it wide to provide as much space to the widget as possible
> On the library side, before checking the widget width and narrowing down, try to take as much space as possible. So that the widget stretches the bubble.

Talk is limit max width to ~80-90%. But I wouldn't span a widget at all.
Important here is a message text that should fit fully. After all, widget is still a decoration (although I agree it should take at least some minimal width to be readable. e.g. `min-width: min(200px, 100%))`.

Antreesy:
Thanks!
I would still like to have a CSS change, otherwise `&--details` with name, description, other text are still limited to 60% if image is missing

tyzbit:
What CSS change are you referring to and where should it go? Sorry if already discussed. Is that the `min-width: min(200px, 100%))` you mention?

Antreesy:
>> Remove 40% empty space preserved for an image that doesn't exist

> This change works for me to allow take the space without an image and keep an ellipsis:
```diff
&--details {
-	width: 60%;
+	width: 0;
+ 	flex-grow: 1;
}
```

tyzbit:
Ah, I added that change to this PR. Can you check?

Antreesy:
Looks good, updated screenshots in the main description!

Antreesy (approved):
Please rebase onto latest `main` branch and squash commits. Otherwise looks good to me!

tyzbit:
Rebased, commits squashed, title updated

tyzbit:
Rebased again so workflows need approval again.

Antreesy:
You don't have to rebase, if it says that branch is outdated (unless there are conflicts)
We're waiting for a second review currently

susnux (approved):

ShGKme:
@susnux Backport?

PR #8553, opened 2026-05-19
[stable8] fix(NcAppSidebarTab): Reduce pill height

kra-mo:
Backport of #8547

### DIFF
--- a/src/components/NcAppSidebar/NcAppSidebarTabs.vue
+++ b/src/components/NcAppSidebar/NcAppSidebarTabs.vue
@@ -306,7 +306,7 @@ export default {
 					bottom: 0;
 					left: 50%;
 					width: 0;
-					height: 6px;
+					height: 4px;
 					border-radius: 999px;
 					background-color: var(--color-primary-element);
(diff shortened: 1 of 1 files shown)


ShGKme (approved):

src/components/NcAppSidebar/NcAppSidebarTabs.vue
```diff
@@ -306,7 +306,7 @@ export default {
 					bottom: 0;
 					left: 50%;
 					width: 0;
-					height: 6px;
+					height: 4px;
```
susnux, line 309:
```suggestion
					height: var(--default-grid-baseline);
```

?

susnux (approved):

src/components/NcAppSidebar/NcAppSidebarTabs.vue
kra-mo: No variable was used before, so I didn't. And I mean, you probably wouldn't want this scaling with that? It's closer to font weight, which is not derived from the baseline either.

PR #8554, opened 2026-05-19
fix(NcSettingsSection): clear NcAppNavigation toggle overhang

pringelmann:
### Resolves

The `NcAppNavigationToggle` button overlaps the first `NcSettingsSection` heading on Nextcloud settings pages. This PR widen the section's inline-start margin so the heading clears the toggle. Falls back to the existing margin when used without an app navigation.

A companion server [PR](https://github.com/nextcloud/server/pull/60596) covers the legacy markup and Vue 2 line.

### Fix

Widen the inline-start margin using `max()`. With an `NcAppNavigation` present, the formula evaluates to ~60 px and clears the toggle. Embedded without a nav, the `max()` falls back to the existing 28 px and the layout is unchanged.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 Tasks

- [ ] Check with design if growing left gutter is OK

### 🏁 Checklist

### DIFF
--- a/src/components/NcSettingsSection/NcSettingsSection.vue
+++ b/src/components/NcSettingsSection/NcSettingsSection.vue
@@ -108,12 +108,18 @@ const ariaLabel = t('External documentation')
 <style lang="scss" scoped>
 $maxWidth: 900px;
 $sectionMargin: calc(var(--default-grid-baseline) * 7);
+// Clear the NcAppNavigation toggle overhang on the inline-start edge.
+$sectionMarginStart: max(
+	#{$sectionMargin},
+	calc(var(--app-navigation-padding) + var(--default-clickable-area) + var(--default-grid-baseline) * 2)
+);
 
 .settings-section {
 	display: block;
 	padding: 0 0 calc(var(--default-grid-baseline) * 5) 0;
 	margin: $sectionMargin;
-	width: min($maxWidth, 100% - calc($sectionMargin * 2));
+	margin-inline-start: $sectionMarginStart;
+	width: min($maxWidth, 100% - #{$sectionMarginStart} - $sectionMargin);
 
 	&:not(:last-child) {
 		border-bottom: 1px solid var(--color-border);


susnux:
I personally dislike this design wise it looks off to me that there is so much weird spacing on the left.

Why not just add some inline margin to the header?
So that on page load you can fully see the header and also the button (prevent overflow).
But when scrolling I think its ok for the toggle to overflow the content.
This also gets really cramped on small mobile (keep in mind we have to support 300px screen width).


Besides that its also because then we have 3 different "content" types:
- left aligned
- centered
- left aligned with margin

What I meant with adjusting the header is like this:

Bildschirmfoto_20260520_105644

similar like we do it in apps.

src/components/NcSettingsSection/NcSettingsSection.vue
```diff
@@ -108,12 +108,18 @@ const ariaLabel = t('External documentation')
 <style lang="scss" scoped>
 $maxWidth: 900px;
 $sectionMargin: calc(var(--default-grid-baseline) * 7);
+// Clear the NcAppNavigation toggle overhang on the inline-start edge.
+$sectionMarginStart: max(
+	#{$sectionMargin},
+	calc(var(--app-navigation-padding) + var(--default-clickable-area) + var(--default-grid-baseline) * 2)
```
susnux, line 114: If we want this then this at least looks a bit uneven to me, most apps I know handle it with:


```suggestion
	calc(2 * var(--app-navigation-padding) + var(--default-clickable-area))
```

pringelmann:
> I personally dislike this design wise it looks off to me that there is so much weird spacing on the left.
> 
> Why not just add some inline margin to the header? So that on page load you can fully see the header and also the button (prevent overflow). But when scrolling I think its ok for the toggle to overflow the content. This also gets really cramped on small mobile (keep in mind we have to support 300px screen width).
> 
> Besides that its also because then we have 3 different "content" types:
> 
>     * left aligned
> 
>     * centered
> 
>     * left aligned with margin
> 
> 
> What I meant with adjusting the header is like this:
> Bildschirmfoto_20260520_105644
> 
> similar like we do it in apps.

That's how I solved it initially, but decided it looks quite bad having the button overlap content on scroll. Sticky headers or banners make sense but a random floating button looks like a bug to me. But fair point about the margin: it does look strange and wastes space. I can change it to header-only and push that to my preview env so we can check it across different pages.

kra-mo:
I do think overlapping generally is undesirable, so how about this:

Remove the button's background and hover colors, and then we have more leeway with margins.

image

We still need to make it look OK when focused, but we could special case that by making it circular for example:

image

Not a great long-term solution, but longer term, we talked about having floating toolbars in general, which this would be a part of. For the short term, this should be an easy fix… if you think it makes sense?

kra-mo:
And actually, if it's smaller and circular, the hover/bg color could also stay.

kra-mo:
Ok, on second thought, my suggestion is kind of stupid. We discussed this with the other designers, I think the best option is to just have a proper toolbar component for overlapping controls in general.

kra-mo:
Until then though, we could just either add padding to the top or do what @susnux suggested.

pringelmann (commented):
Thanks for all the feedback. I will go with the overlapping content solution for now. While not great, its definitely better than a huge margin on the left

kra-mo (approved):
Proactive approval from design unless something changes :)

nfebe (approved):

pringelmann:
Opted to add padding to all settings pages instead. Much simpler and only requires one PR: https://github.com/nextcloud/server/pull/60596

Closing this

PR #8570, opened 2026-05-26
fix(NcSelect): floating label design using NcTextField

skjnldsv:
### ☑️ Resolves

- Relates to nextcloud-libraries/nextcloud-vue#8XXX (NcSelect visual alignment with other form fields)

### Summary

Replace the plain `<input>` in NcSelect's `#search` slot with `NcTextField` so the component shares the same floating label, border, and focus styling as `NcTextField` and `NcTextArea`.

**Single mode:**
- NcTextField owns the border and floating label
- Selected value overlays the input (position: absolute, z-index: 2)
- Label floats on focus or when value selected via `:has(.vs__selected)`
- Extra padding when clearable (clear + chevron need space)

**Multi mode:**
- Border on `.vs__dropdown-toggle` (contains tags + search input)
- Floating label via `#header` slot, centered when empty, floats on border when tags present or dropdown open
- Search input inline with tags (flex-grow, no layout jump on open)
- Tags clipped before actions via `max-width` in no-wrap mode

**Drop-up:**
- Border/radius inverted (transparent top, rounded bottom)
- Label moves to bottom border

**CSS scoping:**
- `data-v-new-select` attribute on root element for manual scoping
- CSS variables scoped to `[data-v-new-select].v-select.select`
- Dropdown menu variables set directly on `.vs__dropdown-menu` (teleported to body by floating-ui)
- `inputLikeBorder` mixin inlined to fix `@media #{&}` doubled-selector bug
- Dark/light theme border overrides with correct selectors

**Other:**
- Forward vue-select events (keydown, blur, focus, IME) via `filterEvents()`
- `v-bind` before explicit props (Vue 3 last-binding-wins)
- No placeholder when multi or value selected (avoids overlap)
- Disabled state: transparent backgrounds, lighter border via `--color-border-dark`

**Tests added:**
- 17 unit tests (vitest): label, placeholder, disabled, inputClass, modelValue, required, single/multi mode, labelOutside, accessibility warnings, event forwarding
- 24 component tests (Playwright): open/close, keyboard nav, search/filter, select, multi tags, deselect, Tab/Escape/blur close, e2e flows

**Docs added:**
- Pre-selected and clearable example
- Disabled state example (single + multi)
- Label outside example (single + multi)

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image
image | image
|

### 🚧 Tasks

- [x] NcTextField integration in search slot
- [x] Event forwarding (keydown, blur, focus, IME)
- [x] Single mode selected value positioning
- [x] Multi mode tags inside bordered container
- [x] Floating label when value selected
- [x] Open state / drop-up border-radius + label repositioning
- [x] CSS scoping via `data-v-new-select`
- [x] Dark theme border fix (inlined mixin)
- [x] Disabled state styling
- [x] No-wrap overflow clipping
- [x] Unit tests (17)
- [x] Component/e2e tests (24)
- [x] Documentation examples (3 new sections)

### 🏁 Checklist

- [x] ⛑️ Tests are included (17 unit + 24 Playwright component tests)
- [x] 📘 Component documentation has been extended (3 new example sections)

### DIFF
--- a/src/components/NcSelect/NcSelect.vue
+++ b/src/components/NcSelect/NcSelect.vue
@@ -484,7 +484,7 @@ export default {
 		</template>
 		<template #search="{ attributes, events }">
 			<!--
-				v-bind MUST come before explicit props — Vue 3 last-binding-wins.
+				v-bind MUST come before explicit props, Vue 3 last-binding-wins.
 				We spread vue-select attributes but null out placeholder/value/modelValue
 				so our explicit bindings below take precedence.
 			-->
@@ -963,7 +963,11 @@ export default {
 				const addClass = {
 					name: 'addClass',
 					fn(/* middlewareArgs */) {
-						dropdownMenu.classList.add('vs__dropdown-menu--floating')
+						// `nc-select__dropdown` scopes our menu styles: the menu is
+						// teleported to <body>, so it cannot be reached by the
+						// `.nc-select` root scope and would otherwise style every
+						// vue-select dropdown on the page (cross-version conflicts).
+						dropdownMenu.classList.add('vs__dropdown-menu--floating', 'nc-select__dropdown')
 						return {}
 					},
 				}
@@ -1296,7 +1300,7 @@ export default {
 	&.vs--multiple {
 		.vs__dropdown-toggle {
 			// Use transparent border in closed state to reserve the same space
-			// as the open state's real border — prevents layout jump on click
+			// as the open state's real border, prevents layout jump on click
 			border: var(--border-width-input-focused, 2px) solid transparent;
 			box-shadow: var(--input-border-box-shadow);
 			// Top padding for floating label clearance, no bottom padding
@@ -1324,13 +1328,13 @@ export default {
 			outline: 2px solid var(--color-main-background);
 		}
 
-		// Remove NcTextField border in multi mode — border is on dropdown-toggle
+		// Remove NcTextField border in multi mode, border is on dropdown-toggle
 		.vs__search .input-field__input {
 			box-shadow: none !important;
 			background: transparent;
 		}
 
-		// NcTextField shares flex row with tags — always inline, no absolute hiding.
+		// NcTextField shares flex row with tags, always inline, no absolute hiding.
 		// No min-width: it may shrink fully so it never forces an empty wrap row
 		// (matches vue-select's own search sizing)
 		.vs__search {
@@ -1356,7 +1360,7 @@ export default {
 			}
 		}
 
-		// No label margin needed — label is via #header
+		// No label margin needed, label is via #header
 		.vs__search .input-field {
 			margin-block-start: 0;
 		}
@@ -1469,10 +1473,11 @@ export default {
 	}
 }
 
-// Dropdown menu is teleported to body by floating-ui, so it must be
-// styled at root level (cannot be nested under .v-select.select).
-// CSS variables must also be set here since it can't inherit from .v-select.select.
-.vs__dropdown-menu {
+// Dropdown menu is teleported to <body>, so it can't live under the
+// `.nc-select` root scope. The `.nc-select__dropdown` class (added in
+// calculatePosition) scopes it instead, so these styles and CSS variables
+// only affect our menus and never leak to other vue-select instances.
+.nc-select__dropdown.vs__dropdown-menu {
 	--vs-border-color: var(--color-border-maxcontrast);
 	--vs-border-style: solid;
 	--vs-border-radius: var(--border-radius-element);
@@ -1524,7 +1529,7 @@ export default {
 	}
 }
 
-// Border for .vs__dropdown-toggle — inlined from inputLikeBorder mixin
+// Border for .vs__dropdown-toggle, inlined from inputLikeBorder mixin
 // because the mixin's @media #{&} produces doubled selectors when nested.
 .nc-select.v-select.select .vs__dropdown-toggle {
 	--input-border-box-shadow-light: 0 -1px var(--vs-border-color),
--- a/src/components/NcSelect/NcSelect.vue
+++ b/src/components/NcSelect/NcSelect.vue
@@ -1014,7 +1014,7 @@ export default {
 							addClass,
 							togglePlacementClass,
 							// On top placement, leave a gap for the floating label overhang
-							// (half the 13px/1.5 label line-height) instead of connecting seamlessly
+							// (about half the floated label line-height) instead of connecting seamlessly
 							offset(({ placement }) => (placement.startsWith('top') ? 10 : -1)),
 							shift({ limiter: limitShift() }),
 						],
@@ -1178,7 +1178,7 @@ export default {
 	// Force floating label when a value is selected (search input is empty but a value exists)
 	.vs__selected-options:has(.vs__selected) .vs__search {
 		.input-field__input + .input-field__label {
-			--input-label-font-size: 13px;
+			--input-label-font-size: var(--font-size-small, 13px);
 			line-height: 1.5;
 			inset-block-start: calc(-1.5 * var(--input-label-font-size) / 2);
 			font-weight: 500;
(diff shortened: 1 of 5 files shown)


jancborchardt (dismissed):
Very nice! :) 2 details:
- [x] Both "multiple" inputs now have inverted shadow effect. It should seem inset (dark shadow from the top, light on the bottom)
- [ ] ~~The border of the select when open is too strong. Our `--color-border-maxcontrast` is confusing as unlike `--color-text-maxcontrast` it is not less contrasting but much more. What would look better is just using `--color-border` here.~~

skjnldsv:
> * The border of the select when open is too strong

You mean the entire border? This is the same it currently is.
Or do you mean the separator border ?

It is also the same as the Input field
image

kra-mo (dismissed):
Really nice, thank you!

But there seem to be issues with alignment and clipping:

Image

Image

And I think Jan means the entire border, which this indeed did not touch, so I'd say it's fine to keep for this pull request.

kra-mo:
Hmm, the alignment issue seems to be based on where in the viewport the fields are. It's only an issue when in the bottom half of the screen, if I see it correctly.

kra-mo:
There also seems to be clipping here:

image

skjnldsv:
Ok @kra-mo @jancborchardt seems I fixed it all! Please update your reviews :bow:

kra-mo:
Looks mostly OK, but

1. user select seems to still clip for me
image

and

2. I'm not loving the label shifting to the bottom:

image

Any chance to just make some room for it and keep it at the same spot even when it expands above?

kra-mo:
Oof, and also, the previous selection remains as a placeholder and you can type over it?

image

And in general the selected value is misaligned with the actual thing you're typing so the position marker clips:

image

skjnldsv:
I'm definitely missing my last commit because those are 3 issues I fixed thhis morning :see_no_evil: 
Il'l have a look at my laptop

skjnldsv:
All fixed @kra-mo and @jancborchardt

kra-mo (changes requested):
Mostly looks good, but this happens when the multiselect pops up at the top:

Image

skjnldsv:
> Mostly looks good, but this happens when the multiselect pops up at the top:

yes, you asked me to remove the bottom floating label. That was my alternative :)
Please suggest a different one :+1:

kra-mo:
@skjnldsv that's not the problem, it's that the bottom border radius is set to 0 even when the popover is at the top.

susnux:
Question should the menu still be shown floating in that case?
Or should the label be hidden and other behavior preserved as before?
(so the "one list" look is preserved)

kra-mo:
I think the floating is fine, it's a good enough solution.

skjnldsv:
> @skjnldsv that's not the problem, it's that the bottom border radius is set to 0 even when the popover is at the top.

Thanks, somehow I did not even saw it! :see_no_evil:

kra-mo (approved):

skjnldsv:
image

src/components/NcSelect/NcSelect.vue
```diff
@@ -895,19 +1055,30 @@ export default {
...
 <style lang="scss">
 @use '../../assets/input-border.scss' as border;
 
-body {
-	/**
-	 * Set custom vue-select CSS variables.
-	 * Needs to be on the body (not :root) for theming to apply (see nextcloud/server#36462)
-	 */
-
+[data-v-new-select].v-select.select {
```
ShGKme: Any specific reason to use attribute selector instead of a class?

It is less performant and harder to scope.

For example, today this is "new select", later we have "slightly new version of new select", and then we need "new-new-select". With classes we can easily switch to scoped classes.

Or was it to increase the specificity?

src/components/NcSelect/NcSelect.vue
```diff
@@ -994,129 +1158,333 @@ body {
...
+		margin: 0 !important;
+		padding: 0 !important;
 	}
 
-	.vs__search, .vs__search:focus {
-		margin: 0;
+	// Force floating label when a value is selected (search input is empty but a value exists)
+	.vs__selected-options:has(.vs__selected) .vs__search {
+		.input-field__input + .input-field__label {
+			--input-label-font-size: 13px;
```
ShGKme: Not `--font-size-small`?

src/components/NcSelect/NcSelect.vue
```diff
@@ -994,129 +1158,333 @@ body {
...
+	--vs-border-style: solid;
+	--vs-border-radius: var(--border-radius-element);
+	--vs-dropdown-bg: var(--color-main-background);
+	--vs-dropdown-color: var(--color-main-text);
+	--vs-dropdown-option-padding: 8px 20px;
+	--vs-dropdown-option--active-bg: var(--color-background-hover);
+	--vs-dropdown-option--active-color: var(--color-main-text);
+	--vs-dropdown-option--kb-focus-box-shadow: inset 0px 0px 0px 2px var(--color-border-maxcontrast);
+	--vs-dropdown-option--deselect-bg: var(--color-error);
+	--vs-dropdown-option--deselect-color: #fff;
```
ShGKme: In the PR description there is:

> CSS Scoping: Dropdown menu variables set directly on .vs__dropdown-menu (teleported to body by floating-ui)

But for me it isn't clear how it is scoped here...
If it is scoped, could you add a comment elaborating it?
If it isn't older and newer selects will conflict...

src/components/NcSelect/NcSelect.vue
```diff
@@ -331,21 +476,33 @@ export default {
...
 		</template>
 		<template #search="{ attributes, events }">
-			<input
+			<!--
+				v-bind MUST come before explicit props — Vue 3 last-binding-wins.
+				We spread vue-select attributes but null out placeholder/value/modelValue
+				so our explicit bindings below take precedence.
+			-->
+			<NcTextField
+				v-bind="{ ...attributes, placeholder: undefined, value: undefined, modelValue: undefined }"
```
ShGKme: The comment describes expected Vue behavior (JS-style v-bind overrides, "last-binding-wins"). But then says, we need null properties out. I didn't get, why do we need to null them out if "last-binding-wins" takes precedence anyway?

ShGKme: Also, it's hard for me to see from the code that all the `...attributes` are actual `NcTextField`'s attributes, and it is guaranteed, we don't pass any extra (invalid) attributes here. Because `attributes` is NcSelect's interface, not HTML pass-though attributes.

Can we explicitly bind them one-by-one? Or add a comment, why it is guaranteed to be valid.

Ans the same for the events.

ShGKme (commented):
Looks very well in the vue-styleguidist, but didn't test yet in apps with multiple select versions.

src/components/NcSelect/NcSelect.vue
skjnldsv: It's to increase the strength of the seelctor. There is too much conflicts on other apps's code otherwise. Scoping is a very strong override. Using a similar approach was the onyl way I could have it work as well :)

src/components/NcSelect/NcSelect.vue
skjnldsv: Damn, very fair point, I forgot about how teleport works :face_exhaling: 
Let me think :thinking:

src/components/NcSelect/NcSelect.vue
skjnldsv: But I think a class could also work, worth a try if you think it's cleaner :+1:

src/components/NcSelect/NcSelect.vue
skjnldsv: sure, it will be cleaner, let's me adjust :+1:

skjnldsv:
Fixing a few tiny design regressions and we should be good :+1:

susnux (approved):
Code looks good, styleguide also looks good!

But please fix DCO

ShGKme:
Slightly broken in `NcActionInput` (inside NcPopover)

image

susnux:
> Slightly broken in `NcActionInput` (inside NcPopover)

Also broken in server in the app store for selecting group membership (ncdialog).
Problem is that the `z-index` variable has been moved from `body` to `.nc-select.v-select.select` and this breaks now as this limits it to the `select` instead of the full component (the popover / list is rendered on body)

PR #8579, opened 2026-05-28
feat(styling): Clicky controls

kra-mo:
Part of https://github.com/nextcloud/server/issues/47739. Let's just start with something really subtle, see how it is in action and we'll go from there.

### 🏁 Checklist

cc @jancborchardt


kra-mo:
PS: I think the buttons themselves we may want to keep subtle in general. It's a very common component, so we shouldn't distract the user too much. But there are certainly opportunities for more fancy animations.

src/components/NcFormBox/NcFormBoxItem.vue
ShGKme, line 1: If we add it to the base component, it's also added to the switch. Is it intended?

src/components/NcButton/NcButton.vue
```diff
@@ -747,7 +747,7 @@ function onClick(event: MouseEvent) {
 	display: flex;
 	align-items: center;
 	justify-content: center;
-	transition-property: color, border-color, background-color;
+	transition-property: color, border-color, background-color, transform;
```
ShGKme, line 751: I'm a bit unsure about `transition` here. With just 1px movement on a normal DPI screen, it feels slow, viscous to me...

ShGKme:
What about the sidebar tabs?

src/components/NcButton/NcButton.vue
kra-mo: Sure, but on HiDPI/zoom, it looks bad without a transition. I do know what you mean, I wanted to do 2px before, it just felt like a bit too much for smaller buttons.

src/components/NcFormBox/NcFormBoxItem.vue
kra-mo: I think so, yeah.

kra-mo:
> What about the sidebar tabs?

I'd say fine without for now because of their appearance (not raised). They get a different type of animation. We'll see how they actually feel in practice.

ShGKme:
> I'd say fine without for now because of their appearance (not raised). They get a different type of animation.

Doesn't a switch also have a different appearance? At least semantically.

Currently, it has to orthogonal animations simultaneously.

src/components/NcButton/NcButton.vue
ShGKme: It it looks bad without animation on high DPI, maybe add it only for high DPI screens?

```css
@media (min-resolution: 2dppx) {
  // ...
}
```

kra-mo:
> Doesn't a switch also have a different appearance? At least semantically.

I chewed on this for 2 years now: https://mastodon.online/@danirabbit/112137211258686683

Initially, I thought the same thing, but at this point, I think it's fine. It's more about consistency then it is about semantics.

kra-mo:
I think we are in the "it's magic" era of UI design :)

ShGKme:
> I chewed on this for 2 years now: https://mastodon.online/@danirabbit/112137211258686683

This one looks good to me. It is "pressed" inside, not moved up orthogonally to the switch animation.

kra-mo:
I don't know. What we have now also looks good to me, better than if it was inconsistent. Pressing down anywhere on a control to flip a switch inside of it doesn't really make sense in the first place, it's already "magic" with or without an animation. At least I don't think the animation changes it, and it would look bad if it was not consistent with the rest.

ShGKme:
> it's already "magic" 

What do you mean by "magic"?

> and it would look bad if it was not consistent with the rest.

Is the vertical slide animation the only one we consider?
Or we can make it consistent and fitting more the cases?

kra-mo:
> What do you mean by "magic"?

I mean the semantics don't really matter because it's not like you are directly flipping a switch anyway.

> Is the vertical slide animation the only one we consider?
> Or we can make it consistent and fitting more the cases?

What do you mean?

I just think raised things clicking down when active makes sense. The switch flipping when (de)activated also makes sense. The relation between the two doesn't really have to be considered, because the switch honestly acts as more of an icon than an actual control itself, since the whole thing is clickable. It might as well just be the text labels "On" and "Off", or a ❌ and a ✔️ or anything else. The switch just is the most common and what we use when it is a standalone control, so it's familiar.

kra-mo:
I changed it to a scale animation which @jancborchardt prefers. It's also more noticeable in general.

The downside is that scale is dependent on the size of the element, meaning it will scale more or less based on size, so I just made FormBoxItem specifically scale a bit less since it can get quite large with a description and everything.

@ShGKme what do you think?

src/components/NcButton/NcButton.vue
```diff
@@ -795,6 +795,9 @@ function onClick(event: MouseEvent) {
 	// TODO: add ripple effect
 	&:active:not(:disabled) {
 		background-color: var(--color-primary-element-light);
+		@media not (prefers-reduced-motion) {
+			transform: scale(0.98);
+		}
```
ShGKme: `prefers-reduced-motion` should be handled by `--animation-quick` variable

src/components/NcFormBox/NcFormBoxItem.vue
```diff
@@ -139,6 +139,12 @@ const hasDescription = () => !!description || !!slots.description
 		background-color: var(--color-primary-element-extra-light-hover);
 	}
 
+	&:active:not(:disabled) {
+		@media not (prefers-reduced-motion) {
+			transform: scale(0.99);
```
ShGKme: Having two values feels complicated.
It is unclear which value to use in new elements, and I feel like different developers will use different or just random values. Also, the same component may even have different sizes depending on the content size or available space.

Would it be possible to stick with one specific value (added to the theming)? For example, `0.985` as a compromise.

On what DPI did you test? Only 1x, 2x or both?

kra-mo: Tested on both, 0.985 makes sense as a compromise, I'll see how it looks.

ShGKme:
Looks a bit slow and weird on the text...
Like it moves independently from the icon/button, and at a moment a distance between `nm` and `p` changes

scale

kra-mo:
> Looks a bit slow and weird on the text...

Yeah, I mean that's gonna be a browser-specific rendering quirk. I'll have to see what can be done about that if anything.

kra-mo:
(on my personal website I just cope with it :P)

kra-mo:
The release will have a broader animation initiative across components, I sort of see this as step one. I wouldn't add something to theming until we start animating more components, but I want to do that soon. I added TODOs for now. Does that sound okay?

Not sure what to do about the text rendering, it looks different on different browsers, devices, and scaling. Since the animations are quick *enough*, I'd say it's fine enough?

ShGKme:
> Not sure what to do about the text rendering, it looks different on different browsers, devices, and scaling. Since the animations are quick _enough_, I'd say it's fine enough?

For me it doesn't look fine.

Could you add `will-change: transform`?

ShGKme:
> Could you add `will-change: transform`?

It may have a negative performance impact, though.
Maybe we can add it only on focus/hover...

kra-mo:
> Maybe we can add it only on focus/hover…

That won't work for touch, though.

kra-mo:
Added `will-change: transform` generally, it does seem to improve rendering significantly on LoDPI here, I hope the performance impact is not too significant.

ShGKme (approved):
Performance-wise it looks fine for me on smoke testing. Let's see if it causes issues in specific places.

On my personal taste, I'd like to have quicker click animation, like `50ms`, to feel like UI is reacting on me faster, and because the animation itself is tiny pixel-wise.

kra-mo:
I just tested with `50ms` and to me on my 60Hz display, I cannot even see the animation. It only gets ~3 frames, so around half of the time, depending on where those frames fall, it just looks like it's not working.

As I said in DMs, `:active` animations are always going to depend on the timing the input type and OS communicate. `100ms` looks good to me, but I know it will look less responsive elsewhere.

jancborchardt (approved):
Feels super juicy, very nice! :) Works super nicely as immediate feedback, and also looks good.

jancborchardt:
FYI @marcoambrosini @nimishavijay if you also want to test (open e.g. "Mail settings"): https://deploy-preview-8579--nextcloud-vue-components.netlify.app/#/Components/App%20containers/NcAppContent/NcAppSettings?id=ncappsettingsdialog

nimishavijay:
Looks good! IMO for wide elements it's a bit distracting because the scale just computes to more pixel values for wider elements, WDYT? I'm wondering if it's possible to have 2 animations, one for "narrow" elements like regular buttons etc, and one for "wide" elements like wide buttons and formbox elements, etc. But that's a follow up for sure!

kra-mo:
@nimishavijay see https://github.com/nextcloud-libraries/nextcloud-vue/pull/8579#discussion_r3459860795, this is what I originally did but I agree that these heuristics are confusing and don't apply very well, since there are wide buttons (due to layouting only) and small screen sizes.

It is not cleanly possible with CSS only since `scale()` only accepts fractions and adding custom calculations for every time a button is pressed is probably a bad idea. The current value we arrived at is the compromise that works for both larger and smaller elements.

Antreesy (commented):
Looks nice =)
Failing test sounds related though. Let's rebase on latest main and see if it still happens

kra-mo:
@Antreesy yeah, even after rebasing, it seems to be failing. Do you know how I can regenerate these tests?

Antreesy:
`npm run update:snapshots`, locally it touches exactly three snapshots from this test

kra-mo:
I'm on Mac so it generates different images, is there something else I can do, or could you update them for me?

Antreesy:
I just remembered we have a GH action 🙈

Antreesy (approved):

PR #8593, opened 2026-06-02
[stable8] fix(NcAppSettingsDialog): Polish dialog navigation and close button

nfebe:
Addresses the remaining Settings Dialog follow-ups from #8526:

  - Hovering an item and switching the selected item now animate smoothly instead of snapping.
  - The selection pill matches the regular navigation height, so it no longer clips into the rounded corners of the item.
  - The close button sits on a tinted background, so content scrolling underneath no longer bleeds through behind it.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
close-b4 | close-After
| before-chevron | after-chevron

### DIFF
--- a/src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
+++ b/src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
@@ -278,6 +278,15 @@ $content-inset: calc(3 * var(--default-grid-baseline));
 		max-width: 900px;
 	}
 
+	:deep(.modal-container__close) {
+		background-color: var(--color-primary-element-light) !important;
+
+		&:hover,
+		&:focus-visible {
+			background-color: var(--color-primary-element-light-hover) !important;
+		}
+	}
+
 	:deep(.dialog__name) {
 		position: absolute;
 		width: 1px;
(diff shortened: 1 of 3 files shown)


susnux:
What about main?

nfebe:
> What about main?

We can backport from here once approved.

jancborchardt (commented):
Looks good, but I would say the close button should have color-main-background rather than being tinted, as the tint calls too much attention to the button.

src/components/NcAppSettingsDialog/NcAppSettingsDialog.vue
```diff
@@ -278,6 +278,15 @@ $content-inset: calc(3 * var(--default-grid-baseline));
 		max-width: 900px;
 	}
 
+	:deep(.modal-container__close) {
+		background-color: var(--color-primary-element-light) !important;
```
susnux, line 282: This is a problem of the modal itself, so should be fixed there.

kra-mo (approved):
Looks good from the screenshots aside from the close button background. I would also like to see how the animations look though. Could you attach a screen recording?

PR #8618, opened 2026-06-08
chore: add Development container

### LINKED ISSUE #8617: Provide devcontainer configuration for contributors
Maybe we can provide a Devcontainer for this repository.

It helped me and will probably help others:

- to get started faster contributing to this lib
- to have a reproducible env for Playwright (that matches CI closely)

Downside: It will need some maintenance in the future.

odzhychko:
### ☑️ Resolves

- Resolves #8617

### 🏁 Checklist

  - Do you want to backport this?

### Info

You can try it out for example with:

- https://github.com/features/codespaces
- https://code.visualstudio.com/docs/remote/remote-overview
- https://devpod.sh/

### DIFF
--- a/.devcontainer/devcontainer-lock.json
+++ b/.devcontainer/devcontainer-lock.json
@@ -1,0 +1,9 @@
+{
+  "features": {
+    "ghcr.io/devcontainers/features/node:2": {
+      "version": "2.0.0",
+      "resolved": "ghcr.io/devcontainers/features/node@sha256:fedd4c11f7adfb64283b578dddc7da906728daa25fa293351c9d913231acf12f",
+      "integrity": "sha256:fedd4c11f7adfb64283b578dddc7da906728daa25fa293351c9d913231acf12f"
+    }
+  }
+}
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,22 @@
+/*
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: CC0-1.0
+ */
+
+{
+	"name": "NextcloudVue",
+	// Do not use mcr.microsoft.com/devcontainers/javascript-node because it only uses Debian.
+	// Playwright snapshots are rendered slightly differently on Debian than on Ubuntu.
+	// Using Ubuntu is consistent with .github/workflows/playwright.yml
+	//
+	// Do not use mcr.microsoft.com/devcontainers/universal because it does not support arm64.
+	"image": "mcr.microsoft.com/devcontainers/base:noble",
+	"features": {
+		"ghcr.io/devcontainers/features/node:2": {}
+	},
+	// Work around for localhost:9323 (Playwright report) not beeing properly forwared due to issues with IPv6.
+	// https://github.com/microsoft/vscode-remote-release/issues/7029
+	"containerEnv": {
+		"NODE_OPTIONS": "--dns-result-order=ipv4first"
+	}
+}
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -90,6 +90,25 @@ You can also test if the design still works with a legacy Nextcloud version by s
...
+npm run test
+```
+
+#### Component tests
+
+
+> [!TIP]
+> We use Playwright for component tests.
+>
+> If you do not meet the [system requirements](https://playwright.dev/docs/intro#system-requirements) or encounter flacky snapshots
```
susnux: flaky?

odzhychko: Yeah, I meant 'flaky' originally. But it is more correct to say that snapshots are just consistently different if the environment is not like in CI. I change the text now accordingly.

.devcontainer/devcontainer-lock.json
susnux, line 1: needs to be added to REUSE.toml or add a .license file

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,22 @@
+/*
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: CC0-1.0
+ */
+
```
susnux, line 5: Comments are not allowed in json files.
So is this is a JSONC file you need to exclude if from eslint

.devcontainer/devcontainer-lock.json
odzhychko: Done

.devcontainer/devcontainer.json
odzhychko: Done. Added config in ESLint to lint it as JSONC.

odzhychko: I probably will add `devcontainer.json`s to other repos in the future...

So would you be fine, if I later contribute it to to the shared config?

(e.g. `GLOB_FILES_JSONC` for https://github.com/nextcloud-libraries/eslint-config/blob/main/lib/configs/json.ts#L41-L48)

susnux: Sure makes sense.

susnux (approved):
I still do not know the use case, as you likely need to test this with your app or at least with styleguide so you will do it with your local development setup anyways.

But if there is a real demand on this and people using this feature ok for me.

odzhychko:
> I still do not know the use case, as you likely need to test this with your app or at least with styleguide so you will do it with your local development setup anyways.

One can still test the changes with an app by running `npm pack` inside the dev container and installing the package in the app.
Or, if you run the dev container locally and therefore your code is only mounted into the dev container, you can do `npm link` etc.


Styleguide can also be run as usual inside a dev container. In that case, the tool running your devcontainer (VSCode, DevPod etc.) usually automatically forwards the port to your host.


My use case is:

- run Playwright without having to set up a VM 
  - (my Linux distro is not supported by Playwright)
- do not have one local development setup
  - so I can safely use different tool chain versions (for Node, PHP etc.) 

BTW the server repo also offers such a config https://github.com/nextcloud/server/tree/master/.devcontainer

> But if there is a real demand on this and people using this feature ok for me.

Yeah, that is hard to track. From my side, we could make this a draft/reject this PR and just wait, whether someone else  voices interest in #8617.I'm not particularly opinionated about this :smile:

.github/dependabot.yml
```diff
@@ -33,6 +33,16 @@ updates:
...
+  - package-ecosystem: devcontainers
+    directory: "/"
+    schedule:
+      interval: weekly
+      timezone: Europe/Paris
+    open-pull-requests-limit: 10
+    labels:
+    - 3. to review
+    - dependencies
+
```
ShGKme, line 45: Probably makes sense to add to the source organization file.

ShGKme (approved):
Thanks!

.github/dependabot.yml
odzhychko: What is the source organization file?

If you mean something in https://github.com/nextcloud-libraries/.github : I didn't find anything about centralizing Depedabot configs with `.github` repos (or any other way).

susnux:
> (my Linux distro is not supported by Playwright)

BTW you do not need support - my is also not supported. Just run the tests only for chromium works fine as that only downloads a prebuilt binary without dynamic dependencies.

PR #8640, opened 2026-06-15
fix(NcDateTimePicker): respect `locale` prop for default date formatting

### LINKED ISSUE #8639: [NcDateTimePicker]: Locales passed as props are ignored for default date formatting (v9)
The `NcDateTimePicker` component accepts `locale` as a prop:

https://github.com/nextcloud-libraries/nextcloud-vue/blob/65a6f32b0887dc20263515bfdcf22c6f835fcca0/src/components/NcDateTimePicker/NcDateTimePicker.vue#L428

But the the default formatting (with `Intl.DateTime`) ignores it by using `getCanoncialLocale()` directly:

https://github.com/nextcloud-libraries/nextcloud-vue/blob/65a6f32b0887dc20263515bfdcf22c6f835fcca0/src/components/NcDateTimePicker/NcDateTimePicker.vue#L548-L559

### Notes

Currently `props.locale` is only passed to `VueDatePicker`.
`VueDatePicker` uses it to extract month and weekday names for the picker components.[^1]

https://github.com/nextcloud-libraries/nextcloud-vue/blob/65a6f32b0887dc20263515bfdcf22c6f835fcca0/src/components/NcDateTimePicker/NcDateTimePicker.vue#L780

[^1]:  https://v11.vue3datepicker.com/props/localization/#locale

odzhychko:
### ☑️ Resolves

- Fix #8639 

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

- [x] ⛑️ Tests **are included** or are not applicable
- [x] 📘 Component documentation has been extended, updated or is **not applicable**

### DIFF
--- a/src/components/NcDateTimePicker/NcDateTimePicker.vue
+++ b/src/components/NcDateTimePicker/NcDateTimePicker.vue
@@ -20,10 +20,23 @@ For those cases this component can be used.
 			<NcCheckboxRadioSwitch v-model="type" type="radio" value="month">Month</NcCheckboxRadioSwitch>
 			<NcCheckboxRadioSwitch v-model="type" type="radio" value="year">Year</NcCheckboxRadioSwitch>
 		</fieldset>
-		<NcDateTimePicker
-			v-model="time"
-			:type />
-		<span>{{ time }}</span>
+		<fieldset>
+			<legend>Picker locale</legend>
+			<NcInputField v-model="locale" labelOutside label="Locale" list="sample-locales"/>
+			<datalist id="sample-locales">
+				<option value="en-US"></option>
+				<option value="en-CA"></option>
+				<option value="es"></option>
+				<option value="de"></option>
+				<option value="ja"></option>
+				<option value="zh"></option>
+			</datalist>
+		</fieldset>
+		<fieldset>
+			<legend>Picker</legend>
+			<NcDateTimePicker v-model="time" :type :locale />
+			<span>{{ time }}</span>
+		</fieldset>
 	</div>
 </template>
 <script>
@@ -32,6 +45,7 @@ export default {
 		return {
 			type: 'date',
 			time: new Date('2022-10-10 10:10:10'),
+			locale: 'en-US'
 		}
 	},
 }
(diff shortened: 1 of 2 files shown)


ShGKme:
As far as I understand, it was intentional – developers are forced to use the user-defined setting. I'd consider removing the unused `locale` prop.

The linked issue doesn't describe why we need to ignore the user setting, what is the use case?

odzhychko:
@ShGKme  The change was primarily motivated by the documentation for the `locale` prop. The component did not behave as described there.

 https://github.com/nextcloud-libraries/nextcloud-vue/blob/65a6f32b0887dc20263515bfdcf22c6f835fcca0/src/components/NcDateTimePicker/NcDateTimePicker.vue#L354-L358

> The linked issue doesn't describe why we need to ignore the user setting, what is the use case?

I found no actual use case but can think of the following: The app developer wants to limit the supported locales for consistency with other elements in their app.
But this is far-fetched, and even then developers would probably do a global override for `getCanonicalLocale()`.

(Also, a nice side effect was that I could try out how the component behaved for different locales directly in Styleguide.)

> I'd consider removing the unused locale prop.

I support that.
Do we have to deprecate it first or can we just remove it?

Impact should be minimal:

 - I found no usages of the `locale` prop in `nextcloud/*` repos
 - Currently it is only passed to https://v11.vue3datepicker.com/props/localization/#locale and influences month and weekday names in the popup
   > Set datepicker locale. Datepicker will use built in javascript locale formatter to extract month and weekday names

ShGKme:
> (Also, a nice side effect was that I could try out how the component behaved for different locales directly in Styleguide.)

I agree, but that should be the documentation feature, not the UI Kit feature.
I'm already adding a huge config block for the documentation, will send a PR shortly.
We can add the locale switch there later. It's just changing the HTML attributes and reloading.

> I support that.
> Do we have to deprecate it first or can we just remove it?

I'd remove. This prop never had an actual effect on the component. It is removing something that had (almost) no effect anyway, so removing it also has no effect.

It changed the Vue3Datepicker behavior, but only partially, because we forced current locale anyway.

odzhychko:
Closing in favor of #8643

odzhychko:
Decided to deprecate `locale` and make it an no-op.
See https://github.com/nextcloud-libraries/nextcloud-vue/pull/8643#issuecomment-4779120249

PR #8643, opened 2026-06-16
fix(NcDateTimePicker): use same locale for formatting and UI elements

### LINKED ISSUE #8639: [NcDateTimePicker]: Locales passed as props are ignored for default date formatting (v9)

odzhychko:
No use case was found for overriding user-selected locales.

Setting the prop had minimal effects because it only influenced some UI elements in `@vuepic/vue-datepicker` but not our formatting.

See https://github.com/nextcloud-libraries/nextcloud-vue/pull/8640#issuecomment-4716837766

### ☑️ Resolves

- Resolves #8639

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🏁 Checklist

- [x] ⛑️ Tests are included or are **not applicable**
- [x] 📘 Component documentation has been extended, **updated** or is not applicable

### DIFF
--- a/src/components/NcDateTimePicker/NcDateTimePicker.vue
+++ b/src/components/NcDateTimePicker/NcDateTimePicker.vue
@@ -288,7 +288,7 @@ import {
 	getFirstDay,
 } from '@nextcloud/l10n'
 import VueDatePicker from '@vuepic/vue-datepicker'
-import { computed, useTemplateRef } from 'vue'
+import { computed, useTemplateRef, warn, watchEffect } from 'vue'
 import NcIconSvgWrapper from '../NcIconSvgWrapper/NcIconSvgWrapper.vue'
 import NcTimezonePicker from '../NcTimezonePicker/NcTimezonePicker.vue'
 import { t } from '../../l10n.ts'
@@ -353,7 +353,9 @@ const props = withDefaults(defineProps<{
 
 	/**
 	 * The locale to use for formatting the shown dates.
-	 * By default the users current Nextcloud locale is used.
+	 *
+	 * @deprecated This property is no longer used and will be ignored because this component will always use the current Nextcloud locale.
+	 * @see https://github.com/nextcloud-libraries/nextcloud-vue/issues/8639
 	 */
 	locale?: string
 
@@ -425,7 +427,7 @@ const props = withDefaults(defineProps<{
 	ariaLabel: t('Datepicker input'),
 	ariaLabelMenu: t('Datepicker menu'),
 	format: undefined,
-	locale: getCanonicalLocale(),
+	locale: undefined,
 	max: undefined,
 	min: undefined,
 	minuteStep: 10,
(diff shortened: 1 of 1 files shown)


ShGKme (approved):
Looks good to me.
I'd even not title it as a breaking change or feature for the change log (PR title). API changed, but not the behavior except a broken case (different locale for different parts of the picker).

odzhychko:
Maybe then `fix` instead of `feat`?

susnux (commented):
Thats a breaking change, no?
Especially even if it was broken before and never worked a user of this API would now get weird HTML as the previous prop would be moved as a attr to the div now.

odzhychko:
> Thats a breaking change, no? Especially even if it was broken before and never worked a user of this API would now get weird HTML as the previous prop would be moved as a attr to the div now.

Ah yeah, I forgot about fallthrough attribute behaviour

The impact of this change besides the prop:

> Impact should be minimal:
>
> - I found no usages of the locale prop in nextcloud/* repos
> - Currently it is only passed to https://v11.vue3datepicker.com/props/localization/#locale and influences month and weekday names in the popup
>   > Set datepicker locale. Datepicker will use built in javascript locale formatter to extract month and weekday names
>
> Wrote in https://github.com/nextcloud-libraries/nextcloud-vue/pull/8640#issuecomment-4716710298



I'm not too particular about what to do:
 - remove and mark as breaking
 - remove and do not mark as breaking
 - do not remove but adapt docs
 - do not remove and do not touch docs
   - the feature is basically not used and a bug report #8639 exists
​

susnux:
A safe way would be to keep the prop, deprecate it and mark it as no-op.

So its basically removed but will not fall-through. Then I agree this can be just a fix

odzhychko:
@susnux Thx. I changed the code and commit message accordingly.

susnux (approved):

PR #8644, opened 2026-06-16
fix(NcDateTimePicker): use translated words in dates formatted with a format string

odzhychko:
Given:
  - user with locale `"de-DE"`
 - `NcDateTimePicker`
   - with `format`  prop `"yyyy MMM dd"`

Expect:
  - Text content is "2000 Okt. 02"

Actual:
  - Input content is "2000 Okt. 02"
  
### ☑️ Resolves

If a format string is provided, the component now uses the user defined locale to use translated names of month, weekday etc. 

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image  |  image

### 🚧 Tasks

- [x] ...

### 🏁 Checklist

- [x] ⛑️ Tests are **included** or are not applicable
- [x] 📘 Component documentation has been extended, updated or is **not applicable**

### Notes

This fix is a byproduct of loading  locales from `data-fns/locale`, which I intent to use in fixing #8639 (in PR #8657)
I opened this PR to make the review easier and get feedback faster.

The loading of `date-fns/locales` is the bulk of the PR.
Loading those locales seemed to be a blocker in https://github.com/nextcloud-libraries/nextcloud-vue/pull/7767#issuecomment-3518611710

### DIFF
--- a/build/date-fns-locales-plugin.mts
+++ b/build/date-fns-locales-plugin.mts
@@ -1,0 +1,62 @@
+/*!
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: CC0-1.0
+ */
+import type { Plugin } from 'vite'
+
+import { exactRegex } from '@rolldown/pluginutils'
+import { glob } from 'fs/promises'
+import { createRequire } from 'module'
+import { dirname, join } from 'path'
+
+/**
+ * This plugin generates a module that can be used to dynamically load locales from `date-fns`.
+ * It generates it in a way that it works with the externalization of the `date-fns` locales
+ * and with Vite and Webpack apps.
+ *
+ * A plugin is used because Vite's dynamic imports and glob imports cannot generate imports in such a way.
+ */
+export default () => {
+	let localeFiles: string[] = []
+	const virtualModuleId = 'virtual:date-fns-locales'
+	const resolvedVirtualModuleId = '\0' + virtualModuleId
+	return {
+		name: 'date-fns-locales-plugin',
+		enforce: 'pre',
+		async buildStart() {
+			this.info('Collectioning locales in date-fns/locales')
+			const require = createRequire(import.meta.url)
+			const pkgJsonPath = require.resolve('date-fns/package.json')
+			const localeDir = join(dirname(pkgJsonPath), 'locale')
+			localeFiles = await Array.fromAsync(glob('*.js', {
+				cwd: localeDir,
+				exclude: ['types.js', 'en-US.js', 'cdn.js', 'cdn.min.js'],
+			}))
+		},
+		resolveId: {
+			filter: { id: /dateFnsLocaleLoader\.ts$/ },
+			handler() {
+				return resolvedVirtualModuleId
+			},
+		},
+		load: {
+			filter: { id: exactRegex(resolvedVirtualModuleId) },
+			handler() {
+				let content = ''
+				content += 'const loader = {}\n'
+				content += 'export default loader\n'
+				for (const localeFile of localeFiles) {
+					// e.g., "pt-BR.js" -> "pt-BR"
+					const localeCode = localeFile.slice(0, -3)
+					let exportName = localeCode.replaceAll('-', '')
+					// One locale is named inconsistently...
+					if (localeCode === 'be-tarask') {
+						exportName = 'beTarasak'
+					}
+					content += `loader['${localeCode}'] = async () => (await import('date-fns/locale/${localeCode}')).${exportName}\n`
+				}
+				return content
+			},
+		},
+	} as Plugin
+}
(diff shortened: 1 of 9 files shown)


build/date-fns-locales-plugin.mts
```diff
@@ -0,0 +1,62 @@
...
+				content += 'export default loader\n'
+				for (const localeFile of localeFiles) {
+					// e.g., "pt-BR.js" -> "pt-BR"
+					const localeCode = localeFile.slice(0, -3)
+					let exportName = localeCode.replaceAll('-', '')
+					// One locale is named inconsistently...
+					if (localeCode === 'be-tarask') {
+						exportName = 'beTarasak'
+					}
+					content += `loader['${localeCode}'] = async () => (await import('date-fns/locale/${localeCode}')).${exportName}\n`
```
odzhychko, line 56: Generated module looks like this.

```js
const loader = {};
loader["af"] = async () => (await import("date-fns/locale/af")).af;
loader["ar-DZ"] = async () => (await import("date-fns/locale/ar-DZ")).arDZ;
...
loader["zh-HK"] = async () => (await import("date-fns/locale/zh-HK")).zhHK;
loader["zh-TW"] = async () => (await import("date-fns/locale/zh-TW")).zhTW;
```

odzhychko: It tested it with Vite and Rspack apps.

vite.config.ts
```diff
@@ -71,7 +73,7 @@ export default defineConfig((env) => {
 		// By default all dependencies are external, but no path imports
 		nodeExternalsOptions: {
 			// Packages with paths imports should be added here to mark them as external as well
-			include: [/^@nextcloud\/.+\//, /^@mdi\/svg\//],
+			include: [/^@nextcloud\/.+\//, /^@mdi\/svg\//, /^date-fns\/locale\//],
```
odzhychko, line 76: Externalized as suggested in https://github.com/nextcloud-libraries/nextcloud-vue/pull/7767#issuecomment-3518611710

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -766,7 +770,11 @@ function sameDay(a: Date, b: Date): boolean {
 
 <template>
 	<div class="vue-date-time-picker__wrapper">
+		<!-- Setting :key="dateFnsLocaleCode" forces the component to rerender when `:formatLocale` changes.
```
susnux: Is there an upstream bug report for this? If not please create one when adding such a work-around

odzhychko: Done. I Created an issue upstream and added a link.
Also simplified the workaround by removing an intermediate computed var.

(...I got lazy there because we are behind a few major versions, and I just assumed it's probably fixed. But no, the issue persists in the most recent version too.)

odzhychko:
@susnux @ShGKme Would you be able to review this again when you have time?

GVodyanov (approved):
Tested with Italian and code looks good!

susnux (approved):
seems to work fine locally

susnux:
One note here: This will increase bundle sizes of such apps a lot.
We might need to consider to reduce list of locales to those provided by server

odzhychko:
> We might need to consider to reduce list of locales to those provided by server

The problem with that is that the server allows ~750 locales. 
https://github.com/nextcloud/server/blob/master/resources/locales.json

Almost any dateFns-locale has a corresponding server locale.

---

One idea i had was to load it for the set locale as we load translations only for the set language.

https://github.com/nextcloud/server/blob/1c8037712d3dc6a15d6125169b92a68bd562bc02/lib/public/Util.php#L207-L222

But that would require new server code...

---

My preferred solution to reduce bundle size is to construct a dateFns-locale at runtime by extracting the required data by poking at the formatting outputs of `Intl.DateTime`

For example, https://tempo.formkit.com/#format-str can extract the format rather reliably from some calls to `Intl.DateTime`.
 And https://tempo.formkit.com/#range can extract month and weekday names.

But I thought of that as being too complex for now. I keep it in mind if bundle size becomes a concern for a concrete app.

susnux:
> The problem with that is that the server allows ~750 locales.

No not really, because this is not about locale itself its translated words.
So this adds translation not localization so server has far less options.

e.g. its common to configure
- language: English
- locale: German

The expected formatting would be to have `DD. MM. YYYY` instead of `MM / DD / YYYY` but still get e.g. `3rd July` instead of `3. Juli`.

odzhychko:
> No not really, because this is not about locale itself its translated words. So this adds translation not localization so server has far less options.

Uh, I haven't thought about that.

Locale objects directly provided by `dateFns` assume the language is the same as the locale. 
Our current default localization with `Intl.DateTimeFormat` also just uses the language from the locale and not the actually configured language.

It should be possible to combine the `dateFns`-locale object for the nextcloud-locale with the `dateFns`-locale object for the nextcloud-language. Something like `{ ...de, localize: enUS.localize, match: enUS.match }`.

I would consider trying that out after https://github.com/nextcloud-libraries/nextcloud-vue/pull/8657

odzhychko:
> e.g. its common to configure
> 
> * language: English
> * locale: German
> 
> The expected formatting would be to have `DD. MM. YYYY` instead of `MM / DD / YYYY` but still get e.g. `3rd July` instead of `3. Juli`.


I created enhancement issue https://github.com/nextcloud-libraries/nextcloud-vue/issues/8930 for this. I didn't just follow up with a PR because (a) it seems to we never respected the user language was never respected in the first place for date formatting and (b) we have related date formatting code (`useFormatTime`) that we might want to change at the same time to be consistent.

PR #8646, opened 2026-06-17
fix(NcAppNavigation): deactivate focus on resize

Antreesy:
### ☑️ Resolves

- Fix case when `isMobile` changed `true -> false` with `open: true` was never considered, and trap stayed active

Reproduction steps:
1. Load the app on ≥ 1024px — isMobile = false, open = true
2. Resize window to below 1024px — isMobile = true → watch(isMobile) fires → open changes from true to false → watch(open) fires → toggleFocusTrap() called, trap deactivates
3. Open the navigation on mobile — open = true → watch(open) fires → toggleFocusTrap() called → isMobile=true && open=true → trap activates
4. Resize back to ≥ 1024px — isMobile = false → watch(isMobile) fires → sets open.value = !false = true.
  - open was already true → Vue sees no change → watch(open) does NOT fire.
  - toggleFocusTrap() is never called.
  - Focus-trap remains active on desktop. ← BUG

### 🚧 Tasks

- [ ] Manual tests:
	Bug is happening sometimes on production, so far it's the only reliable steps to reproduce it

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppNavigation/NcAppNavigation.vue
+++ b/src/components/NcAppNavigation/NcAppNavigation.vue
@@ -140,7 +140,7 @@ import type { Slot } from 'vue'
 
 import { emit, subscribe, unsubscribe } from '@nextcloud/event-bus'
 import { createFocusTrap } from 'focus-trap'
-import { inject, onMounted, onUnmounted, ref, useTemplateRef, warn, watch, watchEffect } from 'vue'
+import { computed, inject, onMounted, onUnmounted, ref, useTemplateRef, warn, watch, watchEffect } from 'vue'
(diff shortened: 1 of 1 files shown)


Antreesy:
stable8 has this handler in both watchers, so backports are not needed:
```js
	watch: {
		isMobile(value) {
			this.open = !value
			this.toggleFocusTrap()
		},

		open(value) {
			this.toggleFocusTrap()
			emit('navigation-toggled', {
				open: value,
			})
		},
	},
	```

ShGKme (approved):

DorraJaouad (approved):

PR #8657, opened 2026-06-19
fix(NcDateTimePicker): handle text input for default formatting

### LINKED ISSUE #8635: [NcDateTimePicker]: text input gets discarded since v9

odzhychko:
Handle text input when default formatting is used.

I tried not to change the existing formatting.
But the default formatting needed to change to accommodate the used parsing solution (locale information available in `date-fns/locale`)

Notable changes in the formatting are:

- changing separator between date and time for some languages
  e.g., "02.01.2000, 03:04" becomes "02.01.2000 03:04" for "de" (comma is removed)
- no shortened form for duration 
  e.g., "Jan 1 – 7, 2000" becomes "Jan 1, 2000 - Jan 7, 2000" for "en-US"
- "-" becomes the delimiter for duration instead of "–"
- some languages use simplified/fallback local forms
  - e.g.,"2000年1月2日 03:04" becomes "Jan 2, 2000, 3:04 AM" for Chinese  "zh"
  - e.g., "١‏/١‏/٢٠٠٠ – ٧‏/١‏/٢٠٠٠" becomes "1 يناير 2000 - 7 يناير 2000" for Arabic (Bahrain) "ar-BH"
  
This could be considered a breaking change, because things are now displayed differently.
But it could also be considered not to be breaking, because we still display the same information but in a new format, that is improved to accommodate text input. I consider it non-breaking and better than not allowing text input.
  
More details on how we could accommodate parsing without changing the preview format at all can be found in https://github.com/nextcloud-libraries/nextcloud-vue/pull/8657#discussion_r3457515384

### ☑️ Resolves

- Fix #8635

#### Before

[Screencast from 2026-06-03 16-17-41.webm](https://github.com/user-attachments/assets/f9f15340-1866-48e2-81ad-1ece1c0dc3a1)

#### After

[Screencast from 2026-06-03 16-16-14.webm](https://github.com/user-attachments/assets/f07638cb-b46a-4fdd-9673-76c347d1cfb6)

### 🏁 Checklist

- [x] ⛑️ Tests are **included** or are not applicable
- [x] 📘 Component documentation has been extended, **updated** or is not applicable


src/components/NcDateTimePicker/useDateFnsLocale.ts
```diff
@@ -0,0 +1,42 @@
...
+import { computed, toValue } from 'vue'
+import loader from './dateFnsLocaleLoader.ts'
+
+/**
+ * Given a locale try to load the corresponding locale from date-fns.
+ * Fall back to en-US while loading, if the request locale does not exist or could not be loaded.
+ *
+ * @param locale locale code (e.g., 'de-DE')
+ */
+export default function useDateFnsLocale(locale: MaybeRefOrGetter<string>): Ref<Locale> {
```
odzhychko: Instead of loading locales, we could construct them by:

- using tokens from https://github.com/nextcloud-libraries/nextcloud-l10n/blob/main/lib/date.ts
  - translation for tokens not yet in nextcloud-l10n (e.g., quarters and genitive case for month names) could have some fallback logic or be added to be translated
  - also we might want to disable implicit support for tokens that are specific to `date-fns`
    - > These patterns are not in the Unicode Technical Standard <span>#</span>5:
      > - ...
      > - o: ordinal number modifier
      > - ...

     https://date-fns.org/v4.4.0/docs/format
- and extracting format strings with `parseFormatForOpts` for `{ dateStyle: 'medium' }` and `{ dateStyle: 'short' }`

But that would require more code to get right, so I went with the build in locales for now.

tests/component/components/NcDateTimePicker/format.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+import FormattingSnapshotStory from './FormattingSnapshot.story.vue'
+
+// Snapshots testing of formatting for all possible locales.
+//
+// Part of "test:component" because tested code relies on `Intl.DateTime`,
+// which is browser dependend.
+//
+// One big snapshot test to not infalte number of test cases.
+// This keeps reporting clean and runs faster.
+test('all locales format', async ({ mount }) => {
```
odzhychko, line 16: Do you want to keep this snapshots tests?

I found them useful when developing to understand the:

- edge cases related to different locales (and sometimes browsers)
- impact of formatting changes
  - e.g.,  by diffing them between the two commits of this PR

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -530,44 +536,81 @@ const placeholderFallback = computed(() => {
...
+		// - replace input value when user focuses the input by ourselves
+
+		// Disable text input if user provided a format function.
+		// We do not know how to parse such values.
+		return false
+	}
+
+	if (typeof realFormatVal === 'string') {
+		return {
+			// This should not be required in Vuepic because `format` is automatically used for parsing.
```
odzhychko, line 603: Check, if reproducible in latest version.
If not, add and link issue.
Else add comment, when it was fixed.

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -530,44 +536,81 @@ const placeholderFallback = computed(() => {
...
 	// fallback to default formatting
 	return undefined
 })
 
+const textInput = computed<LibraryTextInputOptions>(() => {
+	const realFormatVal = realFormat.value
+	if (typeof realFormatVal === 'function') {
+		// NOTE We could provide a format string through `textInput.format`.
+		// Then Vuepic uses it when the user focuses the input.
+		// But this feature has a bug in v11 because it does not respect the provided locale when formatting.
```
odzhychko: Also check, if reproducible in latest version.
If not, add and link issue.
Else add comment, when it was fixed.

src/components/NcDateTimePicker/NcDateTimePicker.vue
```diff
@@ -530,44 +536,81 @@ const placeholderFallback = computed(() => {
...
-		formatter = new Intl.DateTimeFormat(getCanonicalLocale(), { year: 'numeric', month: '2-digit' })
-	} else if (props.type === 'year') {
-		formatter = new Intl.DateTimeFormat(getCanonicalLocale(), { year: 'numeric' })
 	}
 
-	if (formatter) {
-		return (input: Date | [Date, Date]) => Array.isArray(input)
-			? formatter.formatRange(input[0], input[1])
-			: formatter.format(input)
+	switch (props.type) {
```
odzhychko, line 562: If we get https://github.com/Vuepic/vue-datepicker/issues/1286 fixed update to a newer versions of Vuepic, we can use `textInput.format` to specify a format that only applies when the user focuses the input.

With that, we could keep using `Intl.DateTimeFormat` for formatting the text that is shown by default. And with that, do not change the behavior at all.

So the decision would be between:

1. Using same format for preview and input
  - Drawbacks
     - for some languages, the format changes (compared to impl. on `main`)
         - e.g., "02.01.2000, 03:04" becomes "02.01.2000 03:04" for "de" (comma is removed)
     - no shortend form for durations 
         - e.g., "Jan 1 – 7, 2000" becomes "Jan 1, 2000 - Jan 7, 2000" for "en-US"
2. Using different (but mostly similar) for formats for preview and input
  - Drawbacks
    - Text input changes/flickers when use focuses/leaves the text input

src/components/NcDateTimePicker/NcDateTimePicker.vue
odzhychko: Done. It would be fixed after upgrading to at least v12.1.0.

src/components/NcDateTimePicker/useDateFnsLocale.ts
odzhychko: Fow now out of scope.

src/components/NcDateTimePicker/NcDateTimePicker.vue
odzhychko: This PR implements solution 1..
Solution 2. would be blocked by https://github.com/Vuepic/vue-datepicker/issues/1286

kra-mo (approved):
Not an expert in localization specifically, it looks good from what I understand, so unless someone else has feedback, consider it approved from the design side.

hamza221 (approved):

odzhychko:
@hamza221 Sorry, merged without noticing that you requested a review from @Antreesy .

Let me know, if I'm to rollback this merge.

Antreesy:
I'll check now, maybe it's okay)

tests/component/components/NcDateTimePicker/format.spec.ts
Antreesy: we probably expect them to fail for certain languages? so we can maybe limit snapshots to these one, to not commit 15K lines?

tests/component/components/NcDateTimePicker/allLocales.ts
Antreesy, line 1: we can maybe keep it  as .json file - Vite/Playwright should support importing and using it fine?

Antreesy (commented):
code change itself it fine IMO

tests/component/components/NcDateTimePicker/allLocales.ts
odzhychko: I thought about it again. I think we can limit it now to the locales supported by `date-fns`. Including all locales supported by Nextcloud was only interesting from the migration from `Intl.DateFormat` to localization by `date-fns`.

Applied in #8825 (removes ~20k lines from snapshots)

odzhychko:
> we probably expect them to fail for certain languages? so we can maybe limit snapshots to these one, to not commit 15K lines?

@Antreesy No, none of them is expected to fail immediately. They should only fail if the formatting changes. This can happen if (a) the ` date-fns` package is update and changes locales or (b) we start reworking local formatting again. In both cases, it helps with tracking what actually changed in between versions. 

---

But I did what is described in https://github.com/nextcloud-libraries/nextcloud-vue/pull/8657#discussion_r3735874593

Antreesy:
> They should only fail if the formatting changes.

What I meant under 'language that fails' is more 'language that caused issues on parsing'. But agree in general that limiting to date-fns available locales sounds ok to me. By the diff, I see that some secondary locales were dropped from snapshots, they would probably fall back to more generic locales anyway

PR #8665, opened 2026-06-23
[stable8] feat(NcReferenceWidget): allow to enable resizable widget height

mejo-:
### ☑️ Resolves

Allows widgets to flag themselves as resizable, which adds a resize drag handle to resize them horizontally. Particularly useful for embedded office documents, tables and deck boards in a Text file.

Required to fix https://github.com/nextcloud/text/issues/8279

Developed against stable8 on purpose as Text and Deck apps are still on Vue 2, but already forward-ported locally to `main` branch.

Intending to set `isResizable: true` at least in files and deck board widget for a starter.

### 🖼️ Screencasts

| 🏡 After |
| ---|
| recording |

### 🏁 Checklist

- [] ⛑️ Tests are included or are not applicable

### DIFF
--- a/src/components/NcRichText/NcReferenceWidget.vue
+++ b/src/components/NcRichText/NcReferenceWidget.vue
@@ -9,7 +9,21 @@
 			v-if="reference && hasCustomWidget"
 			ref="customWidget"
 			class="widget-custom"
-			:class="{ 'full-width': hasFullWidth }" />
+			:class="{
+				'full-width': hasFullWidth,
+			}"
+			:style="customWidgetStyle">
+			<div ref="customWidgetContent" class="widget-custom__content" />
+			<div
+				v-if="isResizable"
+				class="widget-custom__resize-handle"
+				role="slider"
+				tabindex="0"
+				aria-orientation="vertical"
+				:aria-label="t('Resize widget height')"
+				@pointerdown.stop.prevent="startResize"
+				@keydown="onResizeKeydown" />
+		</div>
 
 		<component
 			:is="referenceWidgetLinkComponent"
@@ -42,10 +56,11 @@ import { nextTick, ref } from 'vue'
 import { RouterLink } from 'vue-router'
 import NcButton from '../../components/NcButton/index.js'
 import { t } from '../../l10n.js'
-import { destroyWidget, hasFullWidth, hasInteractiveView, isWidgetRegistered, renderWidget } from './../../functions/reference/widgets.ts'
+import { destroyWidget, hasFullWidth, hasInteractiveView, isResizable, isWidgetRegistered, renderWidget } from './../../functions/reference/widgets.ts'
 import { getRoute } from './autolink.js'
 
 const IDLE_TIMEOUT = 3 * 60 * 1000 // 3 minutes outside of viewport before widget is removed from the DOM
+const RESIZE_KEYBOARD_STEP = 10
 
 export default {
 	name: 'NcReferenceWidget',
@@ -96,6 +111,12 @@ export default {
 			showInteractive: false,
 			rendered: false,
 			idleTimeout: null,
+			isResizing: false,
+			resizedHeight: null,
+			resizeStartY: 0,
+			resizeStartHeight: 0,
+			resizeMinHeight: 100,
+			resizeMaxHeight: null,
 		}
 	},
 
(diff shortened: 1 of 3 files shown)


src/components/NcRichText/NcReferenceWidget.vue
```diff
@@ -9,7 +9,14 @@
...
 			ref="customWidget"
 			class="widget-custom"
-			:class="{ 'full-width': hasFullWidth }" />
+			:class="{ 'full-width': hasFullWidth }"
+			:style="customWidgetStyle">
+			<div ref="customWidgetContent" class="widget-custom__content" />
+			<div
+				v-if="isResizable"
+				class="widget-custom__resize-handle"
+				@pointerdown.stop.prevent="startResize" />
```
ShGKme: For a11y it should be focusable, have an accessible label and keyboard activation (ArrowUp/ArrowDown).

ShGKme: Why `pointerdown`?

mejo-: Which other event would you prefer?

mejo-: Good point, added now.

src/components/NcRichText/NcReferenceWidget.vue
```diff
@@ -9,7 +9,21 @@
...
+				'full-width': hasFullWidth,
+			}"
+			:style="customWidgetStyle">
+			<div ref="customWidgetContent" class="widget-custom__content" />
+			<div
+				v-if="isResizable"
+				class="widget-custom__resize-handle"
+				role="slider"
+				tabindex="0"
+				aria-orientation="vertical"
```
ShGKme, line 23: It should also have:
- `aria-valuenow`
- `aria-valuemin`
- `aria-valuemax`

Maybe also `aria-controls` referring to the widget.

ShGKme: For the reference: https://www.w3.org/WAI/ARIA/apg/patterns/windowsplitter/

src/components/NcRichText/NcReferenceWidget.vue
```diff
@@ -9,7 +9,21 @@
...
+				:aria-label="t('Resize widget height')"
+				@pointerdown.stop.prevent="startResize"
+				@keydown="onResizeKeydown" />
```
ShGKme: I'd use Vue to set the event params and keys.
And then reuse a single handler, just providing the step, something like:

```suggestion
				@keydown.up.stop.prevent="onResize(-RESIZE_KEYBOARD_STEP)"
				@keydown.down.stop.prevent="onResize(RESIZE_KEYBOARD_STEP)" />
```

src/components/NcRichText/NcReferenceWidget.vue
ShGKme: `mousedown` to have the delay, but I didn't test on a touch device yet.

src/components/NcRichText/NcReferenceWidget.vue
```diff
@@ -9,7 +9,21 @@
...
+			}"
+			:style="customWidgetStyle">
+			<div ref="customWidgetContent" class="widget-custom__content" />
+			<div
+				v-if="isResizable"
+				class="widget-custom__resize-handle"
+				role="slider"
+				tabindex="0"
+				aria-orientation="vertical"
+				:aria-label="t('Resize widget height')"
```
ShGKme: We can simplify to a value name
```suggestion
				:aria-label="t('Widget height')"
```
Or even just height, if aria-controls works correctly
```suggestion
				:aria-label="t('Height')"
```

src/components/NcRichText/NcReferenceWidget.vue
mejo-: Changed now.

src/components/NcRichText/NcReferenceWidget.vue
mejo-: Thanks, good point. implemented now.

src/components/NcRichText/NcReferenceWidget.vue
ShGKme: Sorry, just tested on Android, and it should be pointer even in all the 3 places here...

src/components/NcRichText/NcReferenceWidget.vue
```diff
@@ -73,10 +95,14 @@ export default {
 	},
 
 	setup() {
+		const widgetId = `nc-reference-widget-${widgetCounter++}`
```
ShGKme: We have `createElementId` utility for ID generation to make sure, the ID is valid and unique. Autoincrement doesn't work here, because every library copy starts its counter from 0 resulting in a duplication.

mejo-: Good point, fixed now.

ShGKme (approved):
Tested on Windows Vivaldi and Android Firefox, works and looks well with mouse, finger, keyboard and screen reader.

susnux (approved):

PR #8670, opened 2026-06-24
fix(NcEllipsisedOption): make sure accessible name is properly computed

susnux:
### ☑️ Resolves

The component uses `display: flex` and then splits a string, e.g. "longvaluename" is split like "'longva' 'luname'". Because of `display: flex` this is considered two different strings instead of one, so the accessible name is now: "longva luname"

If `display: inline` is used its calculated correctly, but even with `inline-flex` it does not work as browsers still consider this now two elements instead of one inline.

The solution is to hide the visual separated elements from the accessibility tree and use a non-visual element that just contains the full name.

### 🏁 Checklist

### DIFF
--- a/src/components/NcEllipsisedOption/NcEllipsisedOption.vue
+++ b/src/components/NcEllipsisedOption/NcEllipsisedOption.vue
@@ -68,13 +68,16 @@ export default {
 
 <template>
 	<span dir="auto" class="name-parts" :title="name">
+		<span class="hidden-visually" v-text="name" />
 		<NcHighlight
+			aria-hidden="true"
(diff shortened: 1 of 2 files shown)


ShGKme (commented):
Does it mean, the highlighting itself is not accessible anymore (hidden)?

susnux:
> Does it mean, the highlighting itself is not accessible anymore (hidden)?

From accessibility tree yes.
We could replace the span with another nchighlight if thats important.

---

This is more of a RFC, I am not sure if:
- we want to add more DOM elements just to fix the broken accessible name
- or accept the name is broken
- or there is another better solution

PR #8685, opened 2026-06-26
feat(NcAppNavigation): add animation effect for navigation items

jancborchardt:
### ☑️ Resolves

It would be nice to have an animation of the hover feedback highlight of the NcAppNavigation. This makes it feel nicer and smoother, like you are actually interacting with something.

This is a prototype to see how that could look like. Has been tested in the styleguide.
@susnux @ShGKme please check if this would this be good as a base, or should take a different approach.

#### 🏚️ Before

[Screencast From 2026-06-27 01-40-20.webm](https://github.com/user-attachments/assets/30489d44-fdab-4fe2-b63b-d329f60789b2)

#### 🏡 After

[Screencast From 2026-06-27 01-34-40.webm](https://github.com/user-attachments/assets/c8d94f73-00dd-4dec-ad5c-ed3a27e4469f)

### 🏁 Checklist

Assisted-by: ClaudeCode:claude-opus-4-8

### DIFF
--- a/src/components/NcAppNavigationList/NcAppNavigationList.vue
+++ b/src/components/NcAppNavigationList/NcAppNavigationList.vue
@@ -8,6 +8,16 @@
 
 List wrapper for use in NcAppNavigation.
 
+The list renders a single hover/focus highlight that slides between entries
+instead of every entry painting its own hover background. When it slides onto
+the active entry it turns transparent, so the active entry keeps its own
+static highlight.
+
+Entries ask for the highlight themselves - `NcAppNavigationItem` does this out
+of the box. The list never inspects the DOM to find out what is hovered, so it
+also works when entries are rendered by a virtual scroller. If no entry ever
+asks for it the per-entry hover background is used as a fallback.
+
 #### Example
 
 Usage with NcAppNavigationCaption as a heading.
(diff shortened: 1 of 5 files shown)


susnux:
I do not see anything in the "after" screencast that requires JS changes, that should be possible with pure CSS so no heavy mutation observer is required, no?

jancborchardt:
@susnux adjusted the PR to use minimal needed JS, also rebased and tests fixed.

susnux:
what is the difference between this JS implementation and a simple CSS on hover animation?

jancborchardt:
@susnux the CSS :hover fades the bg of each entry in place. The JS moves that single highlight that slides from the old entry to the new one. That travel between separate elements can’t be done by CSS alone cause a transition can only work on the properties of an element itself but not on siblings.

jancborchardt:
(Maybe it’s not super obvious on the screencast, especially in the top part where  there’s an active element. But there is a background highlight which looks like it travels across the different entries.)

marcoambrosini:
I think it looks really nice!

jancborchardt:
@susnux @ShGKme what's your updated take on this with the latest changes? Anything specific you would like to see done differently?

susnux (approved):

sorbaugh (approved):

src/components/NcAppNavigationList/NcAppNavigationList.vue
```diff
@@ -30,15 +36,148 @@ Usage with NcAppNavigationCaption as a heading.
 </docs>
 
 <template>
-	<ul class="app-navigation-list">
+	<ul
+		ref="list"
+		class="app-navigation-list"
+		:class="{ 'app-navigation-list--animated-highlight': enabled }"
+		@pointerover="handle"
+		@pointerleave="hide"
```
Antreesy: Pointer events are fired rapidly, so it probably makes sense to throttle them (say 50-100ms, or `requestAnimationFrame` is enough to fire less and still feel like it's not lagging behind the movement)?

src/components/NcAppNavigationList/NcAppNavigationList.vue
```diff
@@ -30,15 +36,148 @@ Usage with NcAppNavigationCaption as a heading.
...
+
+		/**
+		 * Find the entry element a given event target belongs to, or null if the
+		 * event did not land on an eligible entry of this list.
+		 *
+		 * @param event the pointer or focus event
+		 */
+		entryFromEvent(event: Event): HTMLElement | null {
+			const target = event.target as HTMLElement | null
+			const entry = target?.closest<HTMLElement>('.app-navigation-entry')
```
Antreesy: Follow-up: what do we need to extend it to other elements used in app navigations, e.g. all other NcAppNavigation* or NcListItem?

src/components/NcAppNavigationList/NcAppNavigationList.vue
```diff
@@ -30,15 +36,148 @@ Usage with NcAppNavigationCaption as a heading.
...
+		 *
+		 * @param event the pointer or focus event
+		 */
+		handle(event: Event) {
+			const entry = this.entryFromEvent(event)
+			// Not over an entry (e.g. the gap between entries): keep the current
+			// state so the highlight can slide across to the next entry. The
+			// highlight slides onto every entry, including the active one (where it
+			// becomes transparent), so the motion stays continuous.
+			if (entry) {
```
Antreesy: If it's the same entry (or child of the same entry) as currently highlighted one, we should not fire anything for it again

src/components/NcAppNavigationList/NcAppNavigationList.vue
```diff
@@ -30,15 +36,148 @@ Usage with NcAppNavigationCaption as a heading.
...
+		 * visible, otherwise it snaps into place to avoid sliding in from a
+		 * previously hovered entry. Over the active entry it turns transparent
+		 * so the active entry keeps its own static highlight.
+		 *
+		 * @param entry the entry element to cover
+		 */
+		showOn(entry: HTMLElement) {
+			const list = this.$refs.list as HTMLElement
+			const entryRect = entry.getBoundingClientRect()
+			const listRect = list.getBoundingClientRect()
```
Antreesy: These IMO are main points of concerns - might be expensive and causes browsers to rearrange the rendering work on the page to be executed

Antreesy (commented):
Visually looks great, left my concerns regarding performance in the comments

jancborchardt:
@Antreesy your concerns are addressed, could you re-review? :)

ShGKme:
> (Maybe it’s not super obvious on the screencast, especially in the top part where there’s an active element. But there is a background highlight which looks like it travels across the different entries.)

The effect is just slightly visible, but for me it "lags". I don't know if this is intentional, but I can see how late the background is moving to my mouse, which feels like a slow, non-performant UI.

At the same time, it is quite a huge code using JS for styling, which has maintenance cost and is more fragile than more simply implemented animations.

Do you want only this specific effect?
The current is "not obvious" but expensive and gives a low-performance feeling.

---

It is inconsistent with other navigations. Do we plan to do the same for?
- Settings dialog navigation
- List items like in Talk or Mail

ShGKme:
I guess this will be incompatible with virtual lists (virtual scrolling).

susnux (changes requested):
This does not work with virtual scrolling navigations.
E.g. imagine on files the folder tree with 1000 folders or on Talk with hundreds of chats.
With virtual scrolling the list itself needs to be aware of the current state for the animation.

So the animation should not inject or try to guess anything from HTML but instead there needs to be a way to communicate the requested animation between child components and the list itself.

jancborchardt:
@susnux @ShGKme I went with a new approach that should be much more performant and is visually the same. I had too many requirements initially to try and make it perfect but that overloaded it.

@ShGKme Also reduced the time from 100ms to 50ms and now it feels quite snappy.


> It is inconsistent with other navigations. Do we plan to do the same for?
>
>- Settings dialog navigation
>- List items like in Talk or Mail

@ShGKme yes, just like for the sidebar tabs at https://github.com/nextcloud-libraries/nextcloud-vue/pull/8793, it would also look nice for settings dialog nav and list items as you say, as well as action menu items (and probably more places).

susnux (approved):
as discussed we will have this externalized later as a composable

kra-mo:
I noticed that this is quite buggy with scrolling lists, so it should definitely not ship in this state. @susnux is there capacity to reimplement this properly before the release?

kra-mo:
It also seems to be way too fast to even be perceptible for me, at least at 60Hz on Safari.

susnux:
> is there capacity to reimplement this properly before the release?

Not on our side. Maybe @ShGKme has some capacity?
Otherwise revert for now?

kra-mo:
Could we revert, then?

It really is quite buggy. In Collectives and Talk especially because they do custom stuff with their navigation items, but also in general when scrolling.

PR #8691, opened 2026-06-29
fix(NcAppNavigation, NcAppSidebar): More graceful animations

kra-mo:
Since these animations can affect the view quite significantly because of their size, giving them some more time is a good idea. cc @jancborchardt 

### 🏁 Checklist

### DIFF
--- a/src/components/NcAppNavigation/NcAppNavigation.vue
+++ b/src/components/NcAppNavigation/NcAppNavigation.vue
@@ -253,7 +253,7 @@ function toggleNavigation(state?: boolean): void {
 
 	open.value = state === undefined ? !open.value : state
 	const bodyStyles = getComputedStyle(document.body)
-	const animationLength = parseInt(bodyStyles.getPropertyValue('--animation-quick')) || 100
+	const animationLength = parseInt(bodyStyles.getPropertyValue('--animation-slow')) || 200
 
 	setTimeout(() => {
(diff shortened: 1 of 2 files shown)


jancborchardt (commented):
It does look much nicer when it’s a bit slower than 100ms! :) When trying some more values, 300ms seems a tad slow though. 200ms might be the sweet spot, what do you think @kra-mo?

kra-mo:
Yeah, I thought so too, but we would need to either introduce a new variable for it or change `slow` since it’s rarely used. Not sure which @ShGKme ?

ShGKme:
> I thought so too, but we would need to either introduce a new variable for it or change `slow` since it’s rarely used.

Depends on the semantics and more complete view on the animation theming.

If we are ok with all the existing slow animations to be `200ms`, then we can change.
If we can define and describe a new value, then we can add a new variable.

jancborchardt:
@kra-mo I think we talked about adjusting `-slow` to 200ms, right? Could you adjust that, then we can merge this. :)

jancborchardt (approved):

jancborchardt:
@susnux @ShGKme seems to be because of the setTimeout of 1.5 * animationLength. The PR changes 100ms to 200ms, so it goes from 150ms to 300ms. Should anything be adjusted?

susnux (approved):

PR #8698, opened 2026-06-30
feat(NcActions): add wide prop for full-width menus

### LINKED ISSUE #8582: [NcActions] enh: add `wide` setting like for NcButton
We're using NcActions in Forms between two questions to insert a new question at that place. For better accessibility Nimisha proposed to make the button the whole width of the column, especially if it's an icon-only button without `menuName` set.

Currently we added a menuName so that the button is larger, but icon-only would be our preferred way.

Chartman123:
### ☑️ Resolves

- Fix #8582

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -545,6 +545,86 @@ export default {
 </script>
 ```
 
+### Wide buttons
+
+```vue
+<template>
+	<NcActions menu-name="Object management" wide>
+		<template #icon>
+			<IconPencilOutline :size="20" />
+		</template>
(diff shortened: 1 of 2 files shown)


Chartman123:
I wasn't able to add screenshots yesterday, Github didn't let me. I'll try again this evening

susnux:
Can you add a case to the docs? Otherwise this looks fine and in sync with NcButton

susnux (approved):
makes sense from my point of view

ShGKme (approved):

PR #8730, opened 2026-07-13
fix(actions): top-align icons with multi-line action labels

### LINKED ISSUE #5284: [NcActions] Inconsistent icon position
`#icon` slot is centered and `icon-class` is on top.

image

jancborchardt:
The icons in the action menu are vertically center-aligned for multi-line entries (especially visible in Mail). This looks a bit unorganized and they should be top-aligned.

The fix is overriding the global `.material-design-icon { align-self: center; }`.

### ☑️ Resolves

- Fix https://github.com/nextcloud-libraries/nextcloud-vue/issues/5284

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot From 2026-07-13 13-45-56|Screenshot From 2026-07-13 13-46-01

### 🏁 Checklist

Assisted-by: ClaudeCode:claude-opus-4-8

### DIFF
--- a/src/components/NcActionInput/NcActionInput.vue
+++ b/src/components/NcActionInput/NcActionInput.vue
@@ -128,6 +128,7 @@ For the `NcSelect` component, all events will be passed through. Please see the 
 			:class="{
 				'action-input-picker--disabled': disabled,
 				'action-input--visible-label': labelOutside && label,
+				'action-input--tight-label': isNativePicker || isMultiselectType,
 			}"
 			@mouseleave="onLeave">
 			<span class="action-input__icon-wrapper">
@@ -535,7 +536,8 @@ $input-margin: 4px;
 
 	&__icon-wrapper {
 		display: flex;
-		align-self: center;
+		align-self: flex-start;
+		margin-top: var(--default-grid-baseline);
 		align-items: center;
 		justify-content: center;
 
@@ -548,6 +550,12 @@ $input-margin: 4px;
 				vertical-align: middle;
 			}
 		}
+	}
+
+	// NcSelect and the native date picker render their own label with a tighter
+	// line-height. Extra margin drops the icon too low, so omitting it there.
+	&--tight-label &__icon-wrapper {
+		margin-top: 0;
 	}
 
 	& > span {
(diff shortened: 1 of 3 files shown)


ShGKme:
Input is missing intentionally?

image

kra-mo (approved):
Makes sense, I wonder where else we do this and shouldn't. FormBoxItems probably.

jancborchardt:
@ShGKme not missing intentionally, thanks. Also checking for NcActionInput. The placement is a bit more fickle for both but I’ll try some stuff.

ShGKme:
> Makes sense, I wonder where else we do this and shouldn't. FormBoxItems probably.

Yes, but that was by design.

image

jancborchardt:
Fixed now for both NcActionInput and NcActionTextEditable as well.
- Only aligning to top puts it too far up as the entries are taller due to the input fields. So adding some `margin-top: var(--default-grid-baseline);`
- This failed for NcSelect and the native datepicker which have their own label. Omitting the extra margin there works fine.

Just btw, the NcActionInput dropdown is quite a hefty example and it basically doesn’t fit on my screen, we should probably separate it into a few separate more sensical examples.

Before | After
-|-
Screenshot From 2026-07-13 15-10-02|Screenshot From 2026-07-13 15-10-07
Screenshot From 2026-07-13 15-11-25|Screenshot From 2026-07-13 15-10-41

susnux (approved):

src/components/NcActionInput/NcActionInput.vue
```diff
@@ -550,6 +551,13 @@ $input-margin: 4px;
 		}
 	}
 
+	// NcSelect and the native date picker render their own label with a tighter
+	// line-height. Extra margin drops the icon too low, so omitting it there.
+	&:has(.action-input__multi) &__icon-wrapper,
+	&:has(.native-datetime-picker__input) &__icon-wrapper {
+		margin-top: 0;
+	}
```
ShGKme, line 559: `:has` is known to be pretty expensive selector. Is it necessary here?

jancborchardt: Yeah good point, switched to a class being set beforehand.

ShGKme (approved):

marcoambrosini (approved):
very nice

PR #8731, opened 2026-07-13
chore: make prerelease script reusable, ignore bots

Antreesy:
### ☑️ Resolves

- Allows to call function from other scripts (currently wip: extended release automation)
- Ignore bots handlers from formatted changelog (from hardcoded list)

### 🏁 Checklist

- [x] AI-assisted

### DIFF
--- a/build/format-changelog.mjs
+++ b/build/format-changelog.mjs
@@ -10,6 +10,16 @@ import process from 'node:process'
 import { fileURLToPath } from 'node:url'
 
 /**
+ * GitHub handles whose authorship reference is omitted from the changelog,
+ * e.g. bots that only forward other people's work (backports, translations).
+ */
+const IGNORED_AUTHORS = [
+	'backportbot[bot]',
+	'nextcloud-bot',
+	'transifex-integration[bot]',
+]
+
+/**
  * Run against CHANGELOG.md if executed as:
  * - package.json script | `npm run prerelease:format-changelog`
  * - Node.js script | `node build/format-changelog.mjs`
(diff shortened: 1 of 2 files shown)


build/format-changelog.mjs
```diff
@@ -6,21 +6,62 @@
...
-if (formatted !== content) {
-	console.info('✏️ fixing format')
-	await writeFile(file, formatted)
-	console.info('🎉 done')
-} else {
-	console.info('✅ no formatting needed - done.')
+// Run against CHANGELOG.md only when executed from scripts / with `npm run`
+if (process.argv[1] === fileURLToPath(import.meta.url)) {
+	await formatChangelogFile()
 }
```
susnux: I would prefer this on top to have the side effects directly visible

susnux (approved):

build/format-changelog.mjs
Antreesy: Above both functions, you mean?

susnux: yes but fine with it anyways

ShGKme (approved):

susnux:
Shall we put this into https://github.com/nextcloud-libraries/.github to reuse it from other apps instead of always copy it?
At least to have a template? Or maybe we need some other way of distribution such tools?

Antreesy:
> to reuse it from other apps instead of always copy it?

I started to script another steps as well (patching version in package.json, read git log / gh api for generating changelog, prepare changes to commit). Shall we try to finish it, before we start reusing this flow?

PR #8746, opened 2026-07-16
fix(NcAppNavigationToggle): Use more common left panel toggle icon

kra-mo:
The old icon is used more so for non-persistent sidebars. Ones that close after you interact with them.

For example, non-persistent on GitHub and our own website, persistent on Safari or Firefox.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

PS: I hope I did the RTL thing correctly? I looked at how it's done elsewhere.

### DIFF
--- a/src/components/NcAppNavigation/NcAppNavigationToggle.vue
+++ b/src/components/NcAppNavigation/NcAppNavigationToggle.vue
@@ -4,7 +4,7 @@
   -->
 
 <script setup lang="ts">
-import { mdiMenu, mdiMenuOpen } from '@mdi/js'
+import { mdiDockLeft } from '@mdi/js'
 import { computed } from 'vue'
 import NcIconSvgWrapper from '../NcIconSvgWrapper/NcIconSvgWrapper.vue'
 import { t } from '../../l10n.ts'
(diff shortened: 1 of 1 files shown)


jancborchardt (approved):
While previously I was kinda religious about that icon, your reasoning totally makes sense. Also especially with the recent push by MacOS and iPadOS for this icon it is more of a standard. 👍 

And as discussed, I still think the icons for the right sidebar being different makes sense, as they are icons for the specific most used items in there: Sharing for Files, Chat while in a Talk call (we use the generic "right sidebar" icon while not in a call actually).

src/components/NcAppNavigation/NcAppNavigationToggle.vue
```diff
@@ -31,7 +32,7 @@ const title = computed(() => open.value ? t('Close navigation') : t('Open naviga
 			variant="tertiary"
 			@click="open = !open">
 			<template #icon>
-				<NcIconSvgWrapper :path="open ? mdiMenuOpen : mdiMenu" />
+				<NcIconSvgWrapper :path="isRtl ? mdiDockRight : mdiDockLeft" />
```
ShGKme: The icon can be inverted automatically with the `directional` prop

susnux:
```suggestion
				<NcIconSvgWrapper directional :path="mdiDockLeft" />
```

ShGKme:
```suggestion
				<NcIconSvgWrapper :path="mdiDockLeft" directional />
```

kra-mo: Thanks, fixed :)

susnux (approved):

PR #8768, opened 2026-07-22
feat(NcDateTimePickerNative): floating label design

skjnldsv:
aFollowup https://github.com/nextcloud-libraries/nextcloud-vue/pull/8570

Unfiying the designs.

here is a render example of the three components next to each others:
1. input
3. select
4. native datetime

image

### DIFF
--- a/src/components/NcInputField/NcInputField.vue
+++ b/src/components/NcInputField/NcInputField.vue
@@ -61,9 +61,13 @@ export interface NcInputFieldProps {
 	labelOutside?: boolean
 
 	/**
-	 * The type of the input element
+	 * The type of the input element.
+	 *
+	 * The date/time types (`date`, `datetime-local`, `month`, `time`, `week`) render a native
+	 * picker; their label always floats as such inputs have no empty placeholder state.
 	 */
 	type?: 'text' | 'password' | 'email' | 'tel' | 'url' | 'search' | 'number'
+		| 'date' | 'datetime-local' | 'month' | 'time' | 'week'
 
 	/**
 	 * The placeholder of the input.
--- a/src/components/NcSelect/NcSelect.vue
+++ b/src/components/NcSelect/NcSelect.vue
@@ -1147,7 +1147,9 @@ export default {
 	/* Override default vue-select styles */
 	min-height: calc(var(--default-clickable-area) - 2 * var(--border-width-input, 2px));
 	min-width: 260px;
-	margin: 0 0 var(--default-grid-baseline);
+	// 6px block-start reserves space for the floated label, matching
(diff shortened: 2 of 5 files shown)


src/components/NcInputField/NcInputField.vue
```diff
@@ -61,9 +61,13 @@ export interface NcInputFieldProps {
...
 
 	/**
-	 * The type of the input element
+	 * The type of the input element.
+	 *
+	 * The date/time types (`date`, `datetime-local`, `month`, `time`, `week`) render a native
+	 * picker; their label always floats as such inputs have no empty placeholder state.
 	 */
 	type?: 'text' | 'password' | 'email' | 'tel' | 'url' | 'search' | 'number'
+		| 'date' | 'datetime-local' | 'month' | 'time' | 'week'
```
ShGKme, line 70: This is inherited by `NcTextField`. We probably should re-define the type there so developers don't use `NcTextField` with these types.

src/components/NcSelect/NcSelect.vue
```diff
@@ -1147,7 +1147,9 @@ export default {
 	/* Override default vue-select styles */
 	min-height: calc(var(--default-clickable-area) - 2 * var(--border-width-input, 2px));
 	min-width: 260px;
-	margin: 0 0 var(--default-grid-baseline);
+	// 6px block-start reserves space for the floated label, matching
+	// NcInputField (.input-field margin-block-start) so fields line up in a row.
+	margin: 6px 0 var(--default-grid-baseline);
```
ShGKme, line 1152: Accidentelly commited?

src/components/NcInputField/NcInputField.vue
skjnldsv: Nice catch, I thought we already did that! let me indeed overule this :+1:

src/components/NcSelect/NcSelect.vue
skjnldsv: No, this is a design related change. Following the NcSelect floating thingy, I discovered it had the wrong margin used in one of the NcDateTimePickerNative docs which uses an NcSelect (see screenshot inPR desc). Thus doing it here. I understand it can be seen as a bit underrelated :grin: 

I can do it in a different pr if you want :+1:

src/components/NcInputField/NcInputField.vue
skjnldsv: Done

src/components/NcSelect/NcSelect.vue
skjnldsv: I did put it out in its own commit for cleaner history 👍

ShGKme (approved):

susnux (approved):

PR #8769, opened 2026-07-22
feat(NcPopover): animate open/close and remove arrow, fix #8690

### LINKED ISSUE #8690: [NcPopover] Animate opening/closing popovers
When opening or closing a popover, it should animate the following properties:

- Scale 96% → 100%
- Opacity 0% → 100%

Their `transform-origin` should be set to where the arrow is currently. The arrow should be removed (we discussed this in the past and decided that it would be a good idea to do it once we animate the popovers).

jancborchardt:
### ☑️ Resolves

- Fix #8690

#### 🏚️ Before

[Screencast From 2026-07-22 15-52-36.webm](https://github.com/user-attachments/assets/fda0bba8-5b75-497c-a797-e917d85bda5c)

#### 🏡 After

[Screencast From 2026-07-22 15-53-16.webm](https://github.com/user-attachments/assets/dee4b3fc-2972-429a-beae-cdb8bbfcb9e6)

### 🏁 Checklist

### DIFF
--- a/src/components/NcPopover/NcPopover.vue
+++ b/src/components/NcPopover/NcPopover.vue
@@ -178,12 +178,11 @@ See: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/
 	<Dropdown
 		ref="popover"
 		v-model:shown="internalShown"
-		:arrowPadding="10"
 		:autoHide="!noCloseOnClickOutside && closeOnClickOutside"
 		:boundary="boundary || undefined"
 		:container
 		:delay
-		:distance="10"
+		:distance="4 /* One --default-grid-baseline (4px). No arrow to bridge a larger gap anymore. */"
 		handleResize
 		:noAutoFocus="true /* Handled by the focus trap */"
 		:placement="internalPlacement"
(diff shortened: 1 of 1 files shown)


susnux:
Looks good, but removing the caret is intended? There could be many element next to each other. Does it not make it confusing for the user to remember which popup is open?

jancborchardt:
@susnux yeah, at least as per @kra-mo it is intentional, see: https://github.com/nextcloud-libraries/nextcloud-vue/issues/8690

jancborchardt:
Reduced the gap of the popover to the action button from 10 to 4px, as per --default-grid-baseline.

susnux (approved):
Fine if removing the arrow is a design decision.
I still think this can get confusing with multiple actions next to each over and when NcPopover is used as a tooltip

kra-mo (approved):
It should be fine without the arrow, many other interfaces (e.g. GitHub here) don't have it. Most of the time, you activated the control, and the animation makes it clear enough. Worst-case scenario, you can close and reopen the popover :)

kra-mo:
But also, the opacity animation seems to not work in WebKit? I don't think this is new to this PR, but it makes it so that the scale animation is basically not visible. Would be good to get that fixed too, although I guess not a blocker.

susnux:
> Worst-case scenario, you can close and reopen the popover :)

But its also used as popover / tooltip on hover. Where its not directly clear what it means (you did not explicitly click on it).

kra-mo:
Yeah, of course. Still, it's in close enough proximity that I think it should be fine. Native hover tooltips in the browser for example also don't have arrows.

kra-mo:
> But also, the opacity animation seems to not work in WebKit?

Except sometimes it does… probably optimization shenanigans

susnux:
I do not want to argue about the design just raise awareness :)

grafik

PR #8787, opened 2026-07-24
feat(NcSelect): add helperText support

skjnldsv:
Similar to the NcInput

image

### DIFF
--- a/src/components/NcSelect/NcSelect.vue
+++ b/src/components/NcSelect/NcSelect.vue
@@ -38,6 +38,18 @@ const selectArray = [
 
 	{
 		props: {
+			inputLabel: 'With helper text',
+			helperText: 'Choose the option that fits best',
+			options: [
+				'foo',
+				'bar',
+				'baz',
+			],
+		},
+	},
+
+	{
+		props: {
 			inputLabel: 'Simple (top placement)',
 			placement: 'top',
 			options: [
@@ -508,6 +520,7 @@ export default {
 				:readonly="attributes.readonly"
 				:autocomplete="attributes.autocomplete"
 				:aria-label="attributes['aria-label']"
+				:aria-describedby="[attributes['aria-describedby'], helperText ? `${inputId}-helper-text` : null].filter(Boolean).join(' ') || undefined"
 				:aria-autocomplete="attributes['aria-autocomplete']"
 				:aria-controls="attributes['aria-controls']"
 				:aria-owns="attributes['aria-owns']"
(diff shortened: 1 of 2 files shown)


ShGKme:
> WIP

Draft?

ShGKme (commented):
No way to use NcInput's existing helper text?
It also supports the error state.

skjnldsv:
> No way to use NcInput's existing helper text? It also supports the error state.

Unfortunately that would complicate things a bit too much I think, because the NcInput is deep nested into the NcSeelct, so the buolt-in helperText would render at the wrong place. I can add error/success support though

src/components/NcSelect/NcSelect.vue
```diff
@@ -962,6 +1041,16 @@ export default {
...
 	computed: {
+		/**
+		 * Consumer-provided slots forwarded to vue-select, excluding `footer`
+		 * which we render ourselves to host the helper text.
+		 *
+		 * @return {string[]}
+		 */
+		forwardedSlots() {
+			return Object.keys(this.$slots).filter((name) => name !== 'footer')
+		},
```
ShGKme: `this.$slots` is not reactive, it cannot be used in computed for re-computation on change.

It should be a method. Or in-template. Or we don't support slots list update.

src/components/NcSelect/NcSelect.vue
```diff
@@ -552,7 +578,31 @@ export default {
...
+					:path="mdiCheck"
+					inline />
+				<NcIconSvgWrapper
+					v-else-if="error"
+					class="select__helper-text-icon"
+					:path="mdiAlertCircleOutline"
+					inline />
+				{{ helperText }}
+			</p>
+		</template>
```
ShGKme, line 604: Is it intentional that passed footer slot is rendered right before built-in footer, not instead?

src/components/NcSelect/NcSelect.vue
skjnldsv: Ah, fair enough! Let me adjust :+1:

src/components/NcSelect/NcSelect.vue
skjnldsv: No, that's a mistake on my part, let me switch the order :+1:

src/components/NcSelect/NcSelect.vue
```diff
@@ -1076,13 +1155,50 @@ export default {
...
+	font-size: var(--default-font-size);
+	overflow-wrap: anywhere;
+
+	&--error {
+		color: var(--color-error);
+	}
+
+	&--success {
+		color: var(--color-success);
+	}
```
ShGKme, line 1195:
```suggestion
	&--error {
		color: var(--color-text-error);
	}

	&--success {
		color: var(--color-text-success);
	}
```

A | B
-- | --
Image | Image

ShGKme (approved):

susnux (approved):

PR #8837, opened 2026-08-10
fix(NcActions): ignore redundant separators inside popover

Antreesy:
### ☑️ Resolves

Offload app logic of keeping track, when a separator should be shown, to library
- do not render separator as a first/last item, or consequent duplicate
- do not construct complex conditions inside of app template

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 Tasks

More a POC drafted for discussion, current TODO:
- [ ] Manual test (in styleguide and app)
- [ ] Add screenshots

### 🏁 Checklist

- [x] AI assisted

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -1360,6 +1360,26 @@ export default {
 		},
 
 		/**
+		 * `Array.filter` callback to drop redundant `NcActionSeparator` entries:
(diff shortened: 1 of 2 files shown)


susnux (commented):
Seems to work but also looks more like an app bug to render multiple separators?
Doing this within the vnodes also looks expensive

PR #8845, opened 2026-08-13
feat: profile popover card

### LINKED ISSUE #30868: User Profile - 5 - Hovercard
Hovercard for Avatars

- https://github.com/nextcloud/server/issues/207 (with Hovercard examples)

kristian-zendato:
## ☑️ Resolves

- Fix: https://github.com/nextcloud/server/issues/30868

### Desktop
image

### Mobile
image

## 🏁 Checklist


kra-mo (changes requested):
From the screenshots, the "mobile" version looks quite odd with the buttons being centered. They should probably be left-aligned.

Also, the card (and by extension the full profile as well) should have all features that the menu that it replaces currently does as we discussed:

1. They should show the person's local time:

Image

2. And also have a button to show their availability next to the other ones (mail, website, etc.)

src/components/NcProfileHoverCard/NcProfileHoverCard.vue
```diff
@@ -0,0 +1,834 @@
...
+		<template #trigger="{ attrs }">
+			<span class="profile-hover-card__trigger" v-bind="attrs">
+				<slot />
+			</span>
+		</template>
+
+		<div
+			class="profile-hover-card"
+			role="dialog"
+			:aria-label="t('Profile of {user}', { user: profileData?.displayname || user || '' })">
```
susnux: This belongs to `NcPopover`, set instead the `popupRole = dialog` on the `NcPopover`

src/components/NcProfileHoverCard/NcProfileHoverCard.vue
```diff
@@ -0,0 +1,834 @@
...
+		:placement
+		:delay
+		:triggers="['hover', 'focus']"
+		:popoverTriggers="['hover']"
+		noFocusTrap
+		popupRole="dialog">
+		<template #trigger="{ attrs }">
+			<span class="profile-hover-card__trigger" v-bind="attrs">
+				<slot />
+			</span>
```
ShGKme: These attributes connect the popover dialog with the trigger. `<span>` cannot be that trigger. It either must be a button or we need to pass `attrs` to the slot, where the end user will have to provide the button and set the attrs.

src/components/NcProfileHoverCard/NcProfileHoverCard.vue
```diff
@@ -0,0 +1,834 @@
...
+		return {}
+	}
+}
+</script>
+
+<template>
+	<!-- Without a user id, render the slot only (no popover) -->
+	<div v-if="!user" class="profile-hover-card__passthrough">
+		<slot />
+	</div>
```
ShGKme: Could you add an example (use-case) for this case?

src/components/NcProfileHoverCard/NcProfileHoverCard.vue
```diff
@@ -0,0 +1,834 @@
...
+onBeforeUnmount(() => {
+	if (!mobileMediaQuery) {
+		return
+	}
+	if (mobileMediaQuery.removeEventListener) {
+		mobileMediaQuery.removeEventListener('change', onMobileChange)
+	} else {
+		mobileMediaQuery.removeListener(onMobileChange)
+	}
+})
```
ShGKme, line 315: We have the `useIsMobile` composable for the mobile state, which also shares the state.

src/components/NcProfileHoverCard/NcProfileHoverCard.vue
```diff
@@ -0,0 +1,1054 @@
+<!--
+  - SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<docs>
```
susnux: I would move this to the end of the file behind the `<style>` section

susnux (commented):
Looks good, but please use the existing `useIsMobile` instead of custom mobile handling

src/components/NcProfileHoverCard/NcProfileHoverCard.vue
```diff
@@ -0,0 +1,996 @@
...
+		<template v-else-if="profileData && !error">
+			<!-- Header band: name, pronouns, status (mirrors ProfileApp.vue) -->
+			<div class="profile-hover-card__header">
+				<NcButton
+					v-if="isCurrentUser"
+					class="profile-hover-card__header__edit"
+					variant="tertiary"
+					:aria-label="t('Edit profile')"
+					:title="t('Edit profile')"
+					:href="settingsUrl">
```
Antreesy: Should have `target="_blank"`, to not unload the current app?

src/components/NcProfileHoverCard/NcProfileHoverCard.vue
```diff
@@ -0,0 +1,996 @@
...
+ * @param action - Hover card action
+ */
+function otherActionComponent(action: IProfileHoverCardAction) {
+	if (action.to) {
+		return NcActionRouter
+	}
+	if (action.onClick && !action.href) {
+		return NcActionButton
+	}
+	return NcActionLink
```
Antreesy, line 254: This shouldn't be an actionable link
Image
Maybe even exclude it from the list, if you're adding 'Local time' anyway

src/components/NcProfileHoverCard/NcProfileHoverCard.vue
```diff
@@ -0,0 +1,996 @@
...
+							:preloadedUserStatus="avatarUserStatus"
+							:isNoUser="!profileData.isUserAvatarVisible" />
+
+						<div class="user-actions">
+							<div
+								v-if="actionsLoading && resolvedActions.length === 0"
+								class="user-actions__loading">
+								<NcLoadingIcon :size="20" :name="t('Loading actions')" />
+							</div>
+							<template v-else>
```
Antreesy, line 483: Does it make sense to show actions for your own profile card?

src/components/NcAvatar/NcAvatar.vue
```diff
@@ -703,6 +656,12 @@ export default {
 			this.isMenuLoaded = false
 			this.loadAvatarUrl()
 		},
+
+		contactsMenuOpenState(shown) {
```
Antreesy, line 825: This is instant and doesn't get delayed by 400ms. So hovering over 20 items quickly will spawn 20 /findOne request.

Maybe use debounce with slightly lower timeout?
```js
// define function before
const debounceFetchContactsMenu = debounce(this.fetchContactsMenu, 300)

...

if (shown) {
	this.debounceFetchContactsMenu()
} else {
	this.debounceFetchContactsMenu?.flush()
}
```

src/components/NcProfileHoverCard/NcProfileHoverCard.vue
```diff
@@ -0,0 +1,996 @@
...
+	const name = profileData.value?.displayname || props.user || ''
+	return t('{user} has not added any info yet', { user: name })
+})
+
+/**
+ * Pass status to NcAvatar so the badge matches the loaded status
+ */
+const avatarUserStatus = computed(() => {
+	if (!userStatus.value.status && !userStatus.value.icon && !userStatus.value.message) {
+		return undefined
```
Antreesy: undefined here mean NcAvatar will try to fetch it on its own, and you fire two requests at once
```suggestion
		return {}
```

src/components/NcProfileHoverCard/NcProfileHoverCard.vue
```diff
@@ -0,0 +1,996 @@
...
+		}
+		return
+	}
+
+	loading.value = true
+	error.value = false
+
+	try {
+		const [{ data }, status] = await Promise.all([
+			axios.get(generateOcsUrl('/profile/{userId}', { userId: props.user })),
```
Antreesy: Need to collect data during RC phase, how many requests are triggered during working hours - and maybe cache results locally for 15-60 minutes

src/components/NcAvatar/NcAvatar.vue
```diff
@@ -210,42 +209,30 @@ export default {
...
-				<template v-if="item.iconSvg" #icon>
-					<NcIconSvgWrapper :svg="item.iconSvg" />
-				</template>
-				{{ item.text }}
-			</component>
-			<template v-if="contactsMenuLoading" #icon>
-				<NcLoadingIcon />
+			:delay="{ show: 400, hide: 0 }"
+			:popoverTriggers="['hover']"
+			:triggers="['hover', 'focus']"
```
Antreesy: Does it work on mobile touchscreens? IIRC focus won't work with span without tabindex, and I think NcAvatar is no longer focusable at all
Check for 'click' trigger

src/components/NcProfileHoverCard/NcProfileHoverCard.vue
kra-mo: It should be displayed next to title and location anyway, like so:

image

kra-mo: > Maybe even exclude it from the list, if you're adding 'Local time' anyway

Unless that's already what you meant :)

kristian-zendato: > This shouldn't be an actionable link Image Maybe even exclude it from the list, if you're adding 'Local time' anyway

Removing the local time action from contact menu would require remove LocalTimeProvider from ActionProivderStore on server.
I am wondering if it would have any side effect on other parts of the app.

susnux: @kristian-zendato does it help if you just ignore entries without onClick and without href? Maybe thats a quick way to solve this for now given that we need this for RC.

kristian-zendato: You mean, filtering actions on frontend? @susnux

Antreesy: API returns following:
```json
        {
            "title": "2:21\u202fPM \u2022 same time",
            "icon": "https:\/\/nextcloud.local\/core\/img\/actions\/recent.svg",
            "hyperlink": "#",
            "appId": "timezone"
        },
```

What about this?
```js
// Render these
actions.filter((action) => action.appId === timezone') 
// Instead of client computations, take server response
actions.find((action) => action.appId === timezone')?.title
```

kristian-zendato: Yeah, that's the easiest way, but I found that filtering it on frontend would not be good choice.
Ok, let me filter the actions onfrontend.

src/components/NcAvatar/NcAvatar.vue
```diff
@@ -901,41 +858,23 @@ export default {
...
+
+	.avatar-profile-popover {
+		position: absolute;
+		inset: 0;
+		width: 100%;
+		height: 100%;
+		z-index: 1;
+
+		&__trigger {
 			cursor: pointer;
```
Antreesy: cursor flickers to default for a moment, when rendering of popover content starts. Check, if avoidable

src/components/NcAvatar/NcAvatar.vue
```diff
@@ -210,42 +209,31 @@ export default {
...
-				<NcLoadingIcon v-if="contactsMenuLoading" />
-				<IconDotsHorizontal v-else :size="20" />
-			</template>
-		</NcButton>
-		<NcActions
-			v-else-if="hasMenu"
-			v-model:open="contactsMenuOpenState"
-			:aria-label="avatarAriaLabel"
+		<!-- Profile hover card (contacts menu actions are shown on the card) -->
+		<NcPopover
```
Antreesy, line 254: We should probably keep legacy NcActions menu
Otherwise stable32-34 branches can no longer receive library update

Antreesy (commented):
Otherwise looks fine to start testing for RC

Please don't forget to move `<docs>` block

src/components/NcAvatar/NcAvatar.vue
kristian-zendato: @susnux Please check my update for the legacy and let me know I am seeing the right place.

susnux: yes makes sense!

susnux (approved):
@kristian-zendato please move the `<doc>` block to the very bottom (below `<style>`).
Then this should be ready to go

src/components/NcProfileHoverCard/NcProfileHoverCard.vue
```diff
@@ -0,0 +1,1027 @@
...
+	profile: null,
+	open: false,
+	actions: null,
+	actionsLoading: false,
+})
+
+const loading = ref(false)
+const error = ref(false)
+const profileData = ref<IProfileData | null>(props.profile)
+const userStatus = ref<IUserStatus>({})
```
Antreesy: IMO we should also pass `preloadedUserStatus` as prop to this component, and prefill this `userStatus`; so if NcAvatar already fetched a status (or also got it prefilled from the app logic), we don't make a duplicate request

src/components/NcProfileHoverCard/NcProfileHoverCard.vue
```diff
@@ -0,0 +1,1027 @@
...
+	}
+})
+
+/**
+ * Ensure user status is loaded when missing.
+ *
+ * @param userId - User id
+ */
+async function ensureUserStatus(userId: string) {
+	if (!userStatus.value.icon && !userStatus.value.message && !userStatus.value.status) {
```
Antreesy: Then we save a request here

src/components/NcProfileHoverCard/NcProfileHoverCard.vue
Antreesy: Tested with given patch, looks enough:

```diff

diff --git a/src/components/NcAvatar/NcAvatar.vue b/src/components/NcAvatar/NcAvatar.vue
--- a/src/components/NcAvatar/NcAvatar.vue	(revision a8869f4842e92c9849f9647614c070990de74564)
+++ b/src/components/NcAvatar/NcAvatar.vue	(date 1787559498250)
@@ -273,7 +273,8 @@
 				:user="user"
 				:open="contactsMenuOpenState"
 				:actions="hoverCardActions"
+				:preloadedUserStatus="userStatus"
				:actionsLoading="contactsMenuLoading" />

 		</NcPopover>
 

===================================================================

diff --git a/src/components/NcProfileHoverCard/NcProfileHoverCard.vue b/src/components/NcProfileHoverCard/NcProfileHoverCard.vue
--- a/src/components/NcProfileHoverCard/NcProfileHoverCard.vue	(revision a8869f4842e92c9849f9647614c070990de74564)
+++ b/src/components/NcProfileHoverCard/NcProfileHoverCard.vue	(date 1787559488596)
@@ -105,18 +105,24 @@
 	 * True while parent actions (e.g. contacts menu) are still loading.
 	 */
 	actionsLoading?: boolean
+
+	/**
+	 * Preloaded user status from the parent component
+	 */
+	preloadedUserStatus?: IUserStatus | null
 }>(), {
 	user: null,
 	profile: null,
 	open: false,
 	actions: null,
 	actionsLoading: false,
+	preloadedUserStatus: null,
 })

-const userStatus = ref<IUserStatus>({})
+const userStatus = ref<IUserStatus>(props.preloadedUserStatus ?? {})
```

Antreesy (approved):
Tested on 34-35. Good from my POV

susnux (approved):

src/components/NcProfileHoverCard/index.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+export { default } from './NcProfileHoverCard.vue'
+export type { IProfileAction, IProfileData, IProfileHoverCardAction, IUserStatus } from './NcProfileHoverCard.vue'
```
ShGKme, line 7: Why do we need all of this as the library export? Especially with `I` prefix...

ShGKme: If we don't really need it as a public API, I'd remove it before the release.

If we do want to have it, can we have names more consistent with other components?
- For Props type - `<ComponentName>Props`
- Without `I` prefix - we always use complete word in naming and hungarian notation is not actual in TS language (it is not .NET, TS itself doesn't follow it, and it leak implementation details into naming, like adding `Ref` to all `ref()`s or `Array` to all arrays)

For example, `IUserStatus` sounds like a generic UserStatus type, however, it is exported from a popover card and is not equal to the real Nextcloud's user status type.

If we want to export tpe props - we can just import the component props. All the details can be extracted from the props.

kristian-zendato: It is exported for the cases of using the types from other places.
Of course, at the moment, it is not a must.

susnux: It is usually used to distinguish types from actual classes, e.g. you try `new UserStatus` which does not work. So its clear it does not mean a proper JS type but just its contracts / interface.

But for your question: I do not think it is needed. So yes should remove it

susnux: Well having a second look into this: we actually expose those as props so yes we need them

ShGKme: > It is exported for the cases of using the types from other places.

I understand that types are exported to be imported in other places. But this is the library API, we can always add something just in case, but we cannot easily remove or rename it.

> Well having a second look into this: we actually expose those as props so yes we need them

If this is the only reason, and the types are about the component properties (not generic types), let's then export `NcProfileHoverCardProps`? Or at least add a common prefix instead of `I`, for example, `ProfileHoverCardUserStatus` if this is not about Nextcloud's user status (all the User Status API types are different from the exported here)

ShGKme: Currently this introduced a new style for component exports.

Antreesy: Removed in https://github.com/nextcloud-libraries/nextcloud-vue/pull/8877

PR #8859, opened 2026-08-21
fix: modal height in small screen

kristian-zendato:
### ☑️ Resolves

- Fix: https://github.com/nextcloud-gmbh/customer-feature-requests/issues/1595

### Summary
Remove the header and use the full screen height for dialogs when there is no modal title and modal actions in the header and the screen height is lower than 400px

#### Before
image

#### After
image

### 🏁 Checklist

### DIFF
--- a/src/components/NcModal/NcModal.vue
+++ b/src/components/NcModal/NcModal.vue
@@ -270,6 +270,13 @@ const numHeaderActions = computed(() => {
 	return actions
 })
 
+/**
+ * Whether the modal header is needed.
+ * Only show it when there is a title, outside close button, or header actions —
+ * otherwise the area can be used for the modal content (full height on mobile / full size).
+ */
+const hasHeader = computed(() => props.name.trim() !== '' || numHeaderActions.value > 0)
+
 // for developers we should add a warning if used with invalid props combination
 onMounted(() => {
 	if (!props.name && !props.labelId) {
@@ -431,9 +438,10 @@ function clearFocusTrap() {
 				:aria-labelledby="labelId || `modal-name-${modalId}`"
 				:aria-describedby="'modal-description-' + modalId"
 				tabindex="-1">
-				<!-- Header -->
+				<!-- Header (only when there is a title, outside close button, or actions) -->
 				<transition name="fade-visibility" appear>
 					<div
+						v-if="hasHeader"
 						class="modal-header"
 						:data-theme-light="lightBackdrop"
 						:data-theme-dark="!lightBackdrop">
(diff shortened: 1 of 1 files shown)


src/components/NcModal/NcModal.vue
```diff
@@ -796,6 +803,35 @@ function clearFocusTrap() {
 	}
 }
 
+// Without a header, only keep a small margin from the viewport edges
+$max-modal-height-no-header: min(90%, calc(100% - 2 * var(--body-container-margin)));
```
susnux: Maybe simplify to always `calc(100% - 2 * var(--body-container-margin))` (but did not test that)

src/components/NcModal/NcModal.vue
```diff
@@ -688,12 +699,18 @@ function clearFocusTrap() {
...
+	// Space reserved for the modal header; 0 when the header is not shown
+	--modal-header-offset: var(--header-height);
 	display: flex;
 	align-items: center;
 	justify-content: center;
 	width: 100%;
 	height: 100%;
 
+	&--no-header {
+		--modal-header-offset: 0px;
```
susnux: I guess a safeguard would be using body margins which will be 0px on small devices anyways but to make this future proof:

```suggestion
		--modal-header-offset: var(--body-container-margin);
```

src/components/NcModal/NcModal.vue
```diff
@@ -788,9 +805,9 @@ function clearFocusTrap() {
 			max-width: initial;
 			width: 100%;
 			max-height: initial;
-			height: calc(100% - var(--header-height));
+			height: calc(100% - var(--modal-header-offset));
 			position: absolute;
-			top: var(--header-height);
+			top: var(--modal-header-offset);
```
susnux, line 810: Needs @nextcloud-libraries/designers input here if we really want to *always* remove the header if it is not used.

Or if we want the modal to keep within the main content area and *only* when the device is small hide the header.

In the second case I would just move the CSS rule for `*--no-header` into this media query.

kra-mo (approved):
Makes sense from the design side :)

src/components/NcModal/NcModal.vue
kristian-zendato: @susnux I checked for this but on small devices, `--body-container-margin` is `calc(var(--default-grid-baseline) * 2);` = 8px, which is not what we aim for.

susnux: https://github.com/nextcloud/server/blob/chore/vue-dialogs/core/css/apps.scss#L11-L17

susnux (approved):

PR #8870, opened 2026-08-24
fix(NcActions) Listen on afterHide instead of non-existing afterClosed.

### LINKED ISSUE #8869: NcActions listenes on non-existing 'afterClose' event which is not emitted by NcPopover, please use 'afterHide' instead
It appears that NcPopover does not emit 'afterClosed', but 'afterHIde', see following lines:

https://github.com/nextcloud-libraries/nextcloud-vue/blob/610a5c6146b1ace5bbef644ae58ab708c154b5bc/src/components/NcPopover/NcPopover.vue#L376-L380

However, NcActions erroneously listens on 'afterClose':

https://github.com/nextcloud-libraries/nextcloud-vue/blob/610a5c6146b1ace5bbef644ae58ab708c154b5bc/src/components/NcActions/NcActions.vue#L1846-L1847

Consequently NcActions never emits its 'closed' event. The bug affects (at least) v9.0.0 up to master. The v8.x.y series does not seem to be affected.

rotdrop:
It appears that NcPopover does emit an 'afterHide' but not an 'afterClose' event, so NcActions should listen on 'afterHide' instead of on 'afterClose'.

Affects v9.0.0 and up, v8.x.y seems not to be affected.

### ☑️ Resolves

- Fix #8869 
 
### 🖼️ Screenshots

not applicable

### 🚧 Tasks

- not applicable

### 🏁 Checklist

- [x] 2️⃣ Backport to `stable8` for maintained Vue 2 version not needed, bug is only present for v9.0.0 and up

### DIFF
--- a/src/components/NcActions/NcActions.vue
+++ b/src/components/NcActions/NcActions.vue
@@ -1844,7 +1844,7 @@ export default {
 					noFocusTrap: !this.config.withFocusTrap,
 					'onUpdate:shown': this.toggleMenu,
 					onAfterShow: this.onOpened,
-					onAfterClose: this.onClosed,
+					onAfterHide: this.onClosed,
 				},
 				{
 					trigger: () => h(NcButton, {


susnux (approved):
https://github.com/nextcloud-libraries/nextcloud-vue/blob/610a5c6146b1ace5bbef644ae58ab708c154b5bc/src/components/NcPopover/NcPopover.vue#L378

Antreesy (approved):

PR #8872, opened 2026-08-24
[stable8] feat: profile hovercard

kristian-zendato:
### ☑️ Resolves

- Fix: https://github.com/nextcloud/server/issues/30868

### Summary
Backport to stable8 of #8845


src/utils/legacy.ts
Antreesy:
```suggestion
const version = window.OC?.config?.version?.split('.')[0] || '35'
```
But since there are no tests, should be fine to keep as-is

Antreesy (approved):
Didn't test, only verified diff against original. Looks correctly backported

susnux (approved):

PR #8874, opened 2026-08-24
fix: update role of NcHeaderMenu

kristian-zendato:
### ☑️ Resolves

- Fix: https://github.com/nextcloud-gmbh/customer-feature-requests/issues/1591
- Related PR: https://github.com/nextcloud/server/pull/63239

### Summary
Update role of `NcHeaderMenu` to `menuitem`

### 🚧 Tasks

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/NcHeaderMenu/NcHeaderMenu.vue
+++ b/src/components/NcHeaderMenu/NcHeaderMenu.vue
@@ -261,6 +261,7 @@ function clearFocusTrap() {
 		<NcButton
 			:id="isNav ? triggerId : null"
 			ref="triggerButton"
+			role="menuitem"
 			:aria-controls="`header-menu-${id}`"
 			:aria-expanded="isOpened.toString()"
 			:aria-label


src/components/NcHeaderMenu/NcHeaderMenu.vue
ShGKme: `role="menuitem"` is not valid here.

1. It breaks `isNav` prop, because `<nav>` cannot have `menuitem` role.
2. `menuitem` is an interactive element like a button, and cannot have another button inside. Like with a button, its content must be presentational.

ShGKme (changes requested):

src/components/NcHeaderMenu/NcHeaderMenu.vue
susnux: Meaning `isNav` must be deprecated and the `role = menuitem` must be applied to the `<NcButton>` instead (?)

susnux: Checked it again @kristian-zendato 
- basically `isNav` for Nextcloud 35+ can be ignored (always false)
- `role = menuitem` belongs to the inner `NcButton`

ShGKme (changes requested):
1. With `role="menuitem"` here, `isNav` is still not valid. If we deprecated `isNav` prop as @susnux proposed, all the `isNav` handling must be removed. Otherwise we have an invalid state, where `menuitem` are items of the navigation instead if the menu.
2. This PR is compatible with the linked server PR for Nextcloud 35. However, it breaks all the previous versions semantics, where the header is not a manubar. We need to either add a new prop here, or check the server version, assuming, this is server version defined.

See also: https://github.com/nextcloud/server/pull/63239#issuecomment-5424877391

PR #8877, opened 2026-08-24
fix: remove unnecessary exports from hovercard

kristian-zendato:
### Summary
Remove unnecessary exports from NcProfileHoverCard

### 🏁 Checklist


susnux (approved):
hm honestly we expose them as props so we should likely also export them as otherwise this can cause Typescript warnings about no public types in components props.

but for now we can remove this here and discuss later how to properly expose the types.

ShGKme (approved):

PR #8885, opened 2026-08-25
fix: update profile hovercard content data

### LINKED ISSUE #63703: Profile hovercard follow-ups
@kristian-zendato there are some issues with the profile hovercard, which should be fixed before next release.

1. If someone's profile is disabled, the hovercard currently displays "Failed to load profile" inside of the card. We should either still display the limited information we have, or just not display the card at all. Maybe @jancborchardt can tell us what the motivation was for being able to disable your profile and decide based on that.
2. The user status takes a bit of time to appear after the rest is shown. This should be blocking, only show the card once the status is loaded, to avoid this small flash.
3. "View profile" should not be a separate action. Instead, the person's name/profile picture/biography should be clickable directly and they should all take you to the full profile.
4. The empty content pages for "They/you have not added any info yet" should not be visible on the hovercard, only on the full profile page, as it takes up a lot of space.
5. For the profile in general, but including the hovercard, the "Call [name]" and "Chat with [name]" buttons should just be replaced with "Start call" and "Send message", since the name is redundant and often ellipsizes.

kristian-zendato:
### Resolves
Resolves: https://github.com/nextcloud/server/issues/63703

### Summary
Update NcProfileHovercard
- Display initial information for disabled users.
- Remove `View Profile` button and link the Avatar and Display to the profile page
- Optimize user status loading
- Update action button label for Call and Message

### Screenshots
image

#### Without any information
image

#### Disabled Profile
- Desktop
image

- Mobile
image

### 🏁 Checklist

### DIFF
--- a/src/components/NcAvatar/NcAvatar.vue
+++ b/src/components/NcAvatar/NcAvatar.vue
@@ -271,6 +271,7 @@ export default {
 			</template>
 			<NcProfileHoverCard
 				:user="user"
+				:displayName="hoverCardDisplayName"
 				:open="contactsMenuOpenState"
 				:actions="hoverCardActions"
 				:actionsLoading="contactsMenuLoading"
@@ -682,6 +683,15 @@ export default {
 				return null
 			}
 			return this.menu.length > 0 ? this.menu : null
+		},
+
+		/**
+		 * Best available display name for the hover card (prop or contacts menu).
+		 *
+		 * @return {string|undefined}
+		 */
+		hoverCardDisplayName() {
+			return this.displayName || this.contactsMenuData?.fullName || undefined
 		},
 
 		/**
--- a/src/components/NcProfileHoverCard/NcProfileHoverCard.vue
+++ b/src/components/NcProfileHoverCard/NcProfileHoverCard.vue
@@ -82,6 +82,12 @@ const props = withDefaults(defineProps<{
 	user?: string | null
 
 	/**
+	 * Optional display name when already known (skips `/displaynames` for limited profiles).
+	 * When omitted and the full profile is unavailable, the card looks the name up itself.
+	 */
+	displayName?: string | null
+
+	/**
 	 * Optional preloaded profile data (skips the profile API request when set).
 	 */
 	profile?: IProfileData | null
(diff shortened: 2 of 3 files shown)


src/components/NcAvatar/NcAvatar.vue
```diff
@@ -253,28 +253,35 @@ export default {
 		<!-- Profile hover card (Nextcloud 35+) -->
 		<NcPopover
 			v-else-if="hasMenu"
-			v-model:shown="contactsMenuOpenState"
+			:shown="contactsMenuOpenState"
 			class="avatar-profile-popover"
 			:container="menuContainer"
-			:delay="{ show: 400, hide: 0 }"
```
susnux, line 259: I wonder if its not just enough to adjust the hide delay to allow some cursor moving to the new popover

e.g.
```diff
-		:delay="{ show: 400, hide: 0 }"
+		:delay="{ show: 400, hide: 600 }"
```

kristian-zendato: @susnux I am now using custom open/close timer to allow cursor moving to the popover.

susnux: What is the benefit of having the custom timers?

jancborchardt (commented):
Looks really nice! How does it look like for people who haven’t set any custom details yet?

And for the disabled profile, we could reduce the height / margin below the avatar/name a bit (from the bottom only, and keep width the same so it’s uniform with the filled profile).

kra-mo (approved):
An improvement, but I do wonder how much info we can display for people whose profile is disabled. Like maybe certain actions could still be shown, not sure. Same with guest users.

Also, the C is not capital :)

Image

src/components/NcProfileHoverCard/NcProfileHoverCard.vue
susnux, line 1: I wonder if its not better to fetch the displayname instead as otherwise this requires adjusting every NcAvatar usage.
So if profile is disabled use something like

```ts
const displayNames = await axios.post(generateUrl('/displaynames'), { users: [...userIds] })
```

kra-mo (approved):

susnux (approved):

PR #8948, opened 2026-09-10
fix(NcAvatar): pick the avatar variant a display actually needs

pringelmann:
### Resolves

Related to https://github.com/nextcloud/server/issues/61726

`NcAvatar` described the 64px avatar as `1x` and the 512px one as `8x` regardless of the size it renders at. Browsers pick the lowest density at or above their own, and there's nothing in between, so any display above 1x fetched the 512px file.

Measured on a real Talk call, that's 267 KiB per avatar instead of 7 KiB, for participants rendered at 40px.

Both descriptors now come from the rendered size, with a 1.5x stretch allowed on the small variant since the backend only serves 64 and 512. Verified in Chromium and Firefox.

Visual change: worst case is a 1.5x upscale, most visible on a 64px avatar at DPR 1.5.

### Checklist

- [x] Tests included or not applicable
- [x] Component documentation extended/updated or not applicable
- [x] Backport to `stable8` for Vue 2 or not applicable

### AI

- [x] The content of this PR was partly generated using AI (tests, reviewed)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved avatar image selection for small display sizes, providing more appropriate image densities for sharper rendering.
  * Preserved direct use of the large avatar variant for larger display sizes.

* **Tests**
  * Added coverage for responsive avatar image selection across multiple sizes.

### DIFF
--- a/src/components/NcAvatar/NcAvatar.vue
+++ b/src/components/NcAvatar/NcAvatar.vue
@@ -299,6 +299,25 @@ import NcUserStatusIcon from '../NcUserStatusIcon/index.js'
 const browserStorage = getBuilder('nextcloud').persist().build()
 
 /**
+ * How far to stretch the 64px avatar before switching to the 512px one.
+ *
+ * Only those two sizes exist: AvatarController clamps everything else. Without
+ * a tolerance, any display above 1x jumps straight to the 512px file, which is
+ * a few hundred KB for an uploaded photo.
+ */
+const SMALL_VARIANT_UPSCALE = 1.5
+
+/**
+ * Round a density to keep 4.3636... out of the markup.
+ *
+ * @param {number} value the density
+ * @return {number} the rounded density
+ */
+function density(value) {
+	return Number(value.toFixed(3))
+}
+
+/**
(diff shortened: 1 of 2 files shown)


cristianscheid (approved):

susnux (approved):
makes sense from my point of view ✔️

ShGKme (approved):

src/components/NcAvatar/NcAvatar.vue
```diff
@@ -949,9 +814,10 @@ export default {
 
 			if (this.size <= 64) {
 				const avatarUrl = this.avatarUrlGenerator(this.user, 64)
```
Antreesy, line 816: Note: this function hardcodes `v` query, so we're getting two questionmarks:
`/index.php/avatar/{userid}/64?guestFallback=true?v=1`
Should be like this?
`/index.php/avatar/{userid}/64?guestFallback=true&v=1`

pringelmann: Yes, I plan to fix this in the next PR which introduces versioning

PR #8955, opened 2026-09-12
feat(NcModal): expose the slideshow state as a model

skjnldsv:
`NcModal` keeps its slideshow state in a private `runSlideshow` ref, so a parent can neither start the slideshow nor know it is running. The viewer used to reach in with `this.$refs.modal.togglePlayPause()` for its `startSlideshow` open option (Photos starts a slideshow from the Memories and Timeline views this way), and `@nextcloud/viewer` 2.x has no way to do the same on v9.

This turns the ref into `defineModel('slideshowRunning')`, so `v-model:slideshow-running` starts and stops the slideshow and follows the play / pause button. Unbound, the behaviour is unchanged. The end-of-slides reset still writes `false`, which the model now reports too.

Relates to nextcloud-libraries/nextcloud-viewer (the `startSlideshow` open option port).

Unit tests cover start, stop, the button round-trip and the last-slide reset. The docs example got a second button that opens the modal with the slideshow running.

*👾 This pull request was assisted by Claude Code, commits carry an `Assisted-by` trailer.*

### DIFF
--- a/src/components/NcModal/NcModal.vue
+++ b/src/components/NcModal/NcModal.vue
@@ -29,6 +29,16 @@ defineOptions({ inheritAttrs: false })
  */
 const showModal = defineModel<boolean>('show', { default: true })
 
+/**
+ * Whether the slideshow is running.
+ *
+ * Bind it to start or stop the slideshow from outside, or to follow the
+ * play / pause button. It resets to `false` when the last slide is reached.
+ *
+ * @since 9.13.0
+ */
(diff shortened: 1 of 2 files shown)


susnux (approved):

skjnldsv:
Ping anyone ?

susnux (approved):

skjnldsv:
Thanks but that doesn't count @susnux :rofl:

artonge (approved):

PR #8956, opened 2026-09-13
fix(dialog): specificity of mobile selector

pulsejet:
This override doesn't apply because it's not specific enough, and causes dialogs to look horribly large (full screen) on mobile.


src/components/NcDialog/NcDialog.vue
```diff
@@ -545,7 +545,7 @@ const modalProps = computed(() => ({
 <style lang="scss">
 /** When having the small dialog style we override the modal styling so dialogs look more dialog like */
 @media only screen and (max-width: $breakpoint-small-mobile) {
```
susnux, line 547: BTW problem here is order of definition - that why it does not always trigger.

susnux (approved):

skjnldsv (approved):

PR #8959, opened 2026-09-14
feat(NcAvatar): add a version prop for longer avatar caching

pringelmann:
Adds a `version` prop to `NcAvatar`. When set it goes into the avatar URL as `?v=`, which earns a much longer cache lifetime server-side. Talk will pass it from the participant list.

Also fixes the existing current-user handling. The version was appended as `'?v=' + version` onto a URL that already ends in `?guestFallback=true`, giving `/avatar/me/64?guestFallback=true?v=3`. PHP reads that as `guestFallback=true?v=3` with no `v` parameter, so cache-busting your own avatar has never worked. Going through `generateUrl` gives a proper `&v=`.

Needs nextcloud/server#64295 to have any effect. Harmless without it.

### ☑️ Resolves

- Groundwork for nextcloud/server#64295

### 🖼️ Screenshots

Nothing visual. The avatar renders the same, only the URL changes.

### 🏁 Checklist


src/utils/getAvatarUrl.ts
Antreesy, line 39: `@nextcloud/router` provides `generateAvatarUrl`, which is an exact copy of this. Maybe we migrate to it instead?
I can open a PR, but it would be conflicting. Waiting for opinions. IMO should be beneficial, as function can be reused by apps, and not only locked by NcAvatar

Antreesy: Went ahead for nc/router repo: https://github.com/nextcloud-libraries/nextcloud-router/pull/916

pringelmann: thank you!

skjnldsv:
So, is this pr still wanted then ?

pringelmann:
> So, is this pr still wanted then ?

Yes, but I need to update it based on router changes. Will ping when ready

src/utils/getAvatarUrl.ts
Antreesy: Shall we bump `@nextcloud/router` and get this change in? 

```js
export function getAvatarUrl(user: string, options?: AvatarUrlOptions): string {
      return generateAvatarUrl(user, {
              isDarkTheme: options?.isDarkTheme ?? checkIfDarkTheme(document.body),
              isGuestUser: options?.isGuest,
              size: options?.size,
              guestFallback: true,
              version: options?.version,
      })
}
```

src/utils/getAvatarUrl.ts
```diff
@@ -23,6 +23,11 @@ interface AvatarUrlOptions {
 	 * @default 64
 	 */
 	size?: 64 | 512
+
+	/**
+	 * The user's avatar version. Including it earns a much longer cache lifetime.
+	 */
+	version?: string | number | null
```
Antreesy: And i would drop the `null` support, also from component default
```suggestion
	version?: string | number
```

pringelmann: Ugh yeah sorry I thought I committed and pushed when I re-requested :see_no_evil:

Antreesy (approved):
Tested (?v=1 is from top-right menu on old lib version, &v=1 is from app + bumped lib)
Image

skjnldsv (approved):

